PR #332, opened 2014-02-26
A sample nipype interface for fit_tensor

MrBago:
This is what an interface for `dipy_fit_tensor` might look like. It even works with Chris's `nipype_cmd script`!

Please let me know if I made any mistakes implementing the nipype interface, but hopefully there is enough here to get some discussion going.

I think I'll need to re-factor the fit_tensor function again, I'd like to add some IO abstraction so that images can be passed in as either file names or some sort of image objects. Does nipype have anything like this already? Also I'm thinking about how we would best drop the dipy_fit_tensor script, and replace it with nipype_cmd. One way would be for dipy to add nipype as a dependency. Alternatively if our scripts had input and run attributes that might almost be enough to make nipype_cmd work.

Any thoughts @chrisfilo @Garyfallidis @oesteban?

```
fisher ~/.dipy/stanford_hardi$ nipype_cmd dipy.workflows.interfaces DipyFitTensor HARDI150.bvec HARDI150.nii.gz
setting function inputs

bvec = HARDI150.bvec
dwi_data = HARDI150.nii.gz
ignore_exception = False
mask = <undefined>
min_signal = <undefined>
nomask = <undefined>
root = <undefined>
scale = <undefined>


ad_map = HARDI150_ad.nii.gz
dirfa_map = HARDI150_dirFA.nii.gz
dummy_trk_file = HARDI150_dummy.trk
fa_map = HARDI150_fa.nii.gz
mask_file = HARDI150_mask.nii.gz
md_map = HARDI150_md.nii.gz
rd_map = HARDI150_rd.nii.gz
t2di = HARDI150_t2di.nii.gz
tensor = <undefined>
```


arokem:
OK - this one has languished here for a while, but since this is something that came up in conversations at HBM, it seems like a good opportunity to bring this back up. In these conversations, I became convinced that creating nipype interfaces would be a big win, for many different reasons. However, I think that these interfaces should be minimal and rely mostly on a scripting API implemented within dipy.  

First of all, I have a much slimmer proposal for a nipype DTI fitting interface here: 

https://github.com/nipy/nipype/pull/1121

The main idea is that no matter what derived measure you want to calculate from DTI, you will always want to calculate these parameters, so start from calculating and caching the tensor parameters, and generate pipelines with further calculations (e.g. FA and MD) to depend on this interface. This is very different from the "calculate everything" approach proposed here. 

Tests are not passing, but the failures are hard for me to decode. I think that I might be doing the testing wrong somehow. Still examining that.  

I think that the way forward is to create a `workflow` module in dipy, and create many useful and flexible scripting interfaces in there (not enforcing any particular uniform API, and without OO). It will make the work of interfacing with nipype much easier, and will also make it simpler to write CLI based on argparse, without having to duplicate functionality in different places.

Garyfallidis:
Agree with @arokem . Should we close this one?

Garyfallidis:
This has been inactive for long time. Closing ...

PR #390, opened 2014-07-09
WIP: New PIESNO example and small corrections

mdesco:
The PIESNO method was merged in Dipy without an example. Here is an example for the Dipy Gallery. 

At the same time, the PIESNO method was corrected by @samuelstjean to better match the Koay et al 2009 paper. Other examples have also been uniformized.


samuelstjean:
Well, since this seems to only collect dust, should we get the discussion back up? I toyed a bit with some stuff regarding the background on the hcp data, and here are some findings:

1. Since the initial estimation is based on a quantile, if there is more than half of zeros it will give back zero and just bail out of the heavy computations.
2. Even if most background is masked, it still picks up nicely the ring of garbage/noise left around the skull on the HCP data.
3. I think we should return a stack of arrays after all instead of relying on the mode, the user can then choose to use it as is since the interest lies in having a variable noise estimation, which is more robust for diffusion MRI than a single value based on the background like the other method provides currently, which is fine for lightly noisy data like T1 weighted images.

samuelstjean:
@arokem @matthew-brett @Garyfallidis 

So, before this gets forgotten forever, anything else you want to add? I kinda lost track fo what happened after a while, but whatever, the detecting background or not stuff is taken care of built-in by the  quantile code, that is, if the detected quantile value is 0 (so more than half of selected slice amongst volume is zero), the returned value is 0 since it's the first guess.

arokem:
Well, for starters, a rebase on `master` is needed. Second, how is this related to: https://github.com/nipy/dipy/pull/572? I haven't done a thorough review of that one yet, just a couple of clarifying comments. The truth is that I don't feel like I have a full picture yet, but that's probably because I don't have the time to dig into this right this moment.

samuelstjean:
#572 will allow you to supply a volume computed wiht piesno (or any other fancy method for that matter) as opposed to a single volume. This PR, if my memory serves correctly, adds a wrapper function to pass a 4D array, which is then iterated upon slice by slice. The old version was about the user wrapping it in a for loop (as intended in the paper), and for some reason that point did not reach through the masses.

Garyfallidis:
@samuelstjean @arokem is telling you that this needs a rebase. Are you saying that first we should merge #572?

samuelstjean:
They are not related directly in any way, so I don't think that is needed. It's just that #572 enables you to use the result piesno function directly instead of using hackery to supply a variable estimation of the noise.

arokem:
Cool. @mdesco - could you please rebase this one on top of current master?

samuelstjean:
We still need to check everything is in working order as expected before 
someone merges it early, this thing is so old I kinda lost track of it.

arokem:
OK. I would be happy to try this out on some data before merging. Still needs a rebase for me to be able to do that...

dipy/denoise/noise_estimate.py
```diff
@@ -61,6 +67,106 @@ def piesno(data, N=1, alpha=0.01, l=100, itermax=100, eps=1e-5, return_mask=Fals
...
+        sigma = np.zeros(data.shape[-2], dtype=np.float32)
+        mask_noise = np.zeros(data.shape[:-1], dtype=np.bool)
+
+        for idx in range(data.shape[-2]):
+            sigma[idx], mask_noise[..., idx] = piesno_3D(data[..., idx, :], N,
+                                                          alpha=alpha, l=l, itermax=itermax, eps=eps)
+
+        # Take the mode of all the sigmas from each slice as the best estimate,
+        # this should be stable with more or less 50% of the guesses at the same value.
+        sigma, num = mode(sigma, axis=None)
```
samuelstjean: Note to myself - change this to return the full array, as it's just plain misleading to return a single value on real data.

samuelstjean:
I added a check so that if the initial estimation of the quantile is zero (so it means that more than half of the slice is zero), a warning is raised and the function returns a zero in that case. 

I also rebased on master.

arokem:
You sure about that rebase?

I am still getting a "can't merge due to conflict" message on the github
interface

samuelstjean:
yes/no I actually made a PR on Maxime branch, with the rebase and everything, so he need to merge that beforehand in his personal repo, I posted too early and moved to something else XD We will review the code tomorrow. I also thought I didn't send that message, silly me.

fmorency:
I've been working lately with the dipy denoise module and one of my conclusion is that the piesno() noise estimation is (almost) unusable in its current state. The documentation is currently lacking details on how to properly use it with 4D data. 

By chance, I stumbled upon this pull request which seems to be *exactly* what's missing (+ #572) to be able to use piesno easily. Those PRs seem (almost) ready to be merged, but they haven't been updated since half-Feb. Is there anything we can do to help push the merge?

samuelstjean:
It has been superseeded by https://github.com/nipy/dipy/pull/576 since it was becoming a mess as the master branch was updated, but the whole relevant discussion is still here indeed.

arokem:
Hi Felix - thanks for looking at this. We have indeed been very close to
merging this for a while now. The most helpful thing you could do at this
point is to look at #576, try it out, review it, and let me know whether
you think this is missing something, or if you think it works well. If you
think this is good, we can then go ahead and merge that one. Sorry for the
slow progress on this issue, and thanks for your help with this!

arokem:
Closed in favor of #576

PR #404, opened 2014-07-31
A better way to create a response function for CSD

ChantalTax:
This implementation includes the recursive calibration of the response function for CSD, Using an FA threshold is not a very robust method.It is dependent on the dataset (non-informed used subjectivity), and still depends on the diffusion tensor (FA and first eigenvector), which has low accuracy at high b-value. This function recursively calibrates the response function, for more information see Tax et al. NeuroImage 2014.

This implementation is slow only on one line (788) which calls real_sph_harm. I would appreciate any help to optimize this function. Thx!

### DIFF
--- a/dipy/reconst/csdeconv.py
+++ b/dipy/reconst/csdeconv.py
@@ -52,7 +52,7 @@ class ConstrainedSphericalDeconvModel(SphHarmModel):
             ndarray and the second is the signal value for the response
             function without diffusion weighting.  This is to be able to
             generate a single fiber synthetic signal. The response function
-            will be used as deconvolution kernel ([1]_)
+            will be used as deconvolution git pull nipy-dipy masterkernel ([1]_)
         reg_sphere : Sphere (optional)
             sphere used to build the regularization B matrix.
             Default: 'symmetric362'.
--- a/dipy/reconst/csdeconv.py
+++ b/dipy/reconst/csdeconv.py
@@ -873,17 +873,17 @@ def recursive_response(gtab, data, mask=None, sh_order=8, peak_thr=0.01,
 
     no_params = ((sh_order + 1) * (sh_order + 2)) / 2
     response_p = np.ones(no_params)
-    r_sh_all = np.zeros(no_params)
     if mask is None:
         data = data[np.ones(data.shape[0:(data.ndim-1)], dtype=bool)]
     else:
         data = data[mask]
 
-    rot_gradients = np.zeros(gtab.gradients.shape)
+#    rot_gradients = np.zeros(gtab.gradients.shape)
     m, n = sph_harm_ind_list(sh_order)
     where_dwi = lazy_index(~gtab.b0s_mask)
 
     for num_it in range(1, iter):
+        r_sh_all = np.zeros(no_params)
         print(num_it)
         csd_model = ConstrainedSphericalDeconvModel(gtab, response,
                                                     None, sh_order)
--- a/dipy/reconst/csdeconv.py
+++ b/dipy/reconst/csdeconv.py
@@ -906,7 +906,7 @@ def recursive_response(gtab, data, mask=None, sh_order=8, peak_thr=0.01,
         response[sh_mask] = 0
 
         change = abs((response_p[~sh_mask] - response[~sh_mask])/response_p[~sh_mask])
-        if change.all() < convergence:
+        if all(change < convergence):
             break
 
         response_p = response
--- a/dipy/reconst/csdeconv.py
+++ b/dipy/reconst/csdeconv.py
@@ -20,6 +20,13 @@ from dipy.reconst.shm import (sph_harm_ind_list, real_sph_harm,
                               real_sym_sh_basis, sh_to_rh, forward_sdeconv_mat,
                               SphHarmModel)
 
+from dipy.reconst.peaks import peaks_from_model
+from dipy.core.geometry import vec2vec_rotmat
+from dipy.reconst.shm import sh_to_sf
+from dipy.viz import fvtk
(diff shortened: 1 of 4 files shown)


dipy/reconst/tests/test_csdeconv.py
```diff
@@ -7,19 +7,96 @@
...
+                              sph_harm_ind_list, real_sph_harm)
+from dipy.reconst.shm import lazy_index
+from dipy.core.geometry import cart2sphere
+import dipy.reconst.dti as dti
+from dipy.reconst.dti import fractional_anisotropy
+from dipy.core.sphere import Sphere
+
+
+def test_recursive_response_calibration():
+    SNR = 100
```
ChantalTax: @MrBago, yes you are right, I noticed a failure sometimes as well. I think it has to do with the 'fat' response function that I choose for the first iteration, this sometimes doesn't converge to the right solution. I am not sure why this is, in the simulations with our own CSD implementation this never  seems to happen. So either I have to increase the anisotorpy of the initial RF for this test and check if it improves, or we have to look deeper in the code (in real data I did not see this issue).

MrBago: Could we somehow re-work this test so that it is deterministic, ie there is no random contribution to where the test passes or fails?

MrBago:
@ChantalTax I really like this PR, I'd like to help get it ready to merge. If you need help debasing or if some of my comments are not clear please send me an email.

dipy/reconst/tests/test_csdeconv.py
ChantalTax: @MrBago, perhaps I can just use the noiseless case for the test?

matthew-brett: Sorry for my ignorance, but it seems dangerous to me to drop the noise from the test if it is hard to understand why the test is failing.  Is that the case?

ChantalTax: I guess due to the noise CSD with a fat RF sometimes doesnt converge to the right solution in the algorithm (though an SNR of 100 is quite high?). I will investigate this problem.

dipy/reconst/csdeconv.py
```diff
@@ -634,6 +641,119 @@ def auto_response(gtab, data, roi_center=None, roi_radius=10, fa_thr=0.7):
...
+            # for the gradient sphere
+            B_dwi = real_sph_harm(m, n, theta[:, None], phi[:, None])
+            r_sh_all = r_sh_all + np.linalg.lstsq(B_dwi,
+                                                  data[num_vox, where_dwi])[0]
+
+        response = r_sh_all/data.shape[0]
+        response[sh_mask] = 0
+
+        change = abs((response_p[~sh_mask] - response[~sh_mask])/response_p[~sh_mask])
+        if change.all() < convergence:
```
kesshijordan: Hi, @ChantalTax  

Is `change.all() < convergence` what you intended or did you mean `all(change < convergence)`? As it is now, the result of `change.all()` (either True or False) is compared to `convergence` (0.001 by default).

Cheers,
Kesshi

ChantalTax: Hi @kesshijordan 

Thanks for your comment, you are right, I think I mean the second option: to check if all changes were below the convergence threshold. I will replace this by `all(change < convergence)`.

Best,
Chantal

dipy/reconst/tests/test_csdeconv.py
ChantalTax: @MrBago I think the problem with the test is now fixed, I checked it manually by running the test a lot of times, can we check this more thoroughly? I will run some extra tests on real data and I think after this it should be almost ready to merge, what do you think?

arokem:
Hey @ChantalTax - could you please rebase this on `master`? Looks like it won't merge automatically, so you might need to resolve some merge conflicts, before we can merge this. If you need any help with this, let me know - I'd be happy to walk you through this.

arokem:
Hey @MrBago - this is now rebased on master. Could you please take another look?

dipy/reconst/csdeconv.py
```diff
@@ -634,6 +646,117 @@ def auto_response(gtab, data, roi_center=None, roi_radius=10, fa_thr=0.7):
...
+    References
+    ----------
+    .. [1] Tax, C.M.W., et al. NeuroImage 2014. Recursive calibration of
+           the fiber response function for spherical deconvolution of
+           diffusion MRI data.
+    """
+    S0 = 1
+    evals = fa_trace_to_lambdas(init_fa, init_trace)
+    response = (evals, S0)
+    sphere = get_sphere('symmetric724')
```
MrBago: Please use this sphere `from dipy.data import default_sphere`.

dipy/reconst/csdeconv.py
```diff
@@ -47,7 +55,7 @@ def __init__(self, gtab, response, reg_sphere=None, sh_order=8, lambda_=1,
             ndarray and the second is the signal value for the response
             function without diffusion weighting.  This is to be able to
             generate a single fiber synthetic signal. The response function
-            will be used as deconvolution kernel ([1]_)
+            will be used as deconvolution git pull nipy-dipy masterkernel ([1]_)
```
arokem: Whoops. Looks like this somehow got inserted here, while we were rebasing this.

Garyfallidis:
@ChantalTax, @arokem, @MrBago  should we close this PR now?

Garyfallidis:
Isn't that replaced by #588? Sorry, if I missed something.

MrBago:
Thanks all! That's a wrap (We merged this in #588).

PR #463, opened 2014-11-13
DOC: update RESTORE tutorial to use new noise estimation technique

arokem:


samuelstjean:
Found something else : turns out hte cephes library also implements the hyp1f1 function, and is included in scipy. For some reason, they simply chose to not use it, but rely on something else for this particular one, but use cephes for a bunch of other hypergeometric series function.

https://github.com/scipy/scipy/blob/master/scipy/special/generate_ufuncs.py at line 101

Any idea why/if we could easily build the support for a new ufunc? With the hope that it actually works, since they don't use it for some reason.

samuelstjean:
I'm trying to wrap it in cython actually meanwhile, with weird errors at 
runtime alas.

samuelstjean:
Well, as previously mentionned I tried wrapping the files with cython, only to get a bunch of errors on import. Any suggestion as to where I could find someone to help debug that? Maybe the lib os too old to compile without patches on modern system, the last update dates back from 2000.

samuelstjean:
So, after all the cephes version is also somewhat not working in all cases.

# our current gsl version
In [17]: scilpy1f1(-0.5,12.,-800**2/24**2)
Out[17]: array(9.773432897666439)

# scipy
In [19]: h1f1(-0.5,12.,-800**2/24**2)
Out[19]: nan

# cephes 
In [20]: hyp1f1(-0.5,12.,-800**2/24**2)
Out[20]: -2.125043669054867e+212

#wolfram 
hypergeometric1f1(-0.5,12.,-800**2/24**2)
9.773432897666456957821593082159404575552490725623272008191994

Maybe we should move that somewhere else, it's messy to find it back everythiing I try something. Is it possible to fork a topic in two?

And also, so we stay up to date, this actually serves the purpose of implementing both NODDI and adaptive noise estimation, stabilisation algorithms + some other stuff relevant to mic.

samuelstjean:
Oh, there is #468 which does that and I had forgotten about, might be easier to take the discussion there instead of this closed PR.

argriffing:
I've merged that scipy PR into the master branch, so now it looks like this:
```python
>>> import numpy as np
>>> import scipy
>>> scipy.__version__
'0.17.0.dev0+89066e8'
>>> np.__version__
'1.10.0.dev0+9232200'
>>> from scipy.special import hyp1f1
>>> hyp1f1(-0.5,12.,-800**2/24**2)
9.7773014400072249
```
It only agrees with gsl and wolfram to a few digits, but it's better than nan or -inf.

argriffing:
That uses integer division on Python 2; it looks like the accuracy is good with true division.
```python
>>> hyp1f1(-0.5,12.,-800.0**2/24.0**2)
9.7734328976664777
```

samuelstjean:
Yes I saw it on the original PR, thanks! It's surprising it was left unattended for so long, but maybe that will fix our problem back in #675 with any luck. Not having a gsl dependency is easier after all if the scipy version works reliably with this fix.

argriffing:
> It's surprising it was left unattended for so long

It's not so surprising to me, for anyone in academia or industry or government there are pretty much only disincentives to spend time on it :)

Issue #449, opened 2014-10-26
local variable 'ftmp' referenced before assignment


matthew-brett:
I'm getting this warning at build time :

    warning: dipy\tracking\propspeed.pyx:403:20: local variable 'ftmp' referenced before assignment

Sure enough, ``ftmp`` does seem to be referenced before assignment;

```
          #ftmp=floor(tmp+.5)
            if ftmp > qa_shape[i]-1 or tmp < 0.:
                 d=0
                 break
```

I guess this is a bug?

arokem:
Closed through #534

PR #534, opened 2014-12-30
BF: removed ftmp variable

Garyfallidis:
Removed unreferenced ftmp variable. This is in response to issue #449. The bug had no real effect because EuDX has already very hard constrains on the boundaries. It doesn't process seeds on the boundary voxels because of limitations with the trilinear interpolation being used. Test showing that added. Thx to the tracking team with the new tracking API we have a more flexible API to deal with the boundaries.


arokem:
Hmm. Seems best to rerun the matlab code with these new tracks. Sorry for holding this up - I will take a look tomorrow.

arokem:
BTW - is this really for 0.8, or are we planning to do this for 0.9? We do intend to have 0.9 come down before the end of February, so it's not ridiculous to plan to solve it for that, instead of right now.

Garyfallidis:
This is resolving an issue reported for quite some time. Let's fix it now if possible.

arokem:
OK - I am working on it.

arokem:
OK - I am done working on this, and I will make a PR with the new files
against this branch.

For future reference, the code to create the matlab weights and rmse is
here:

https://gist.github.com/arokem/5c6a28707273144404ab

The following IPython notebook takes the resulting files, compares to dipy,
and saves the .npy files:

https://gist.github.com/arokem/8d132c35ee94cb564a73

Garyfallidis:
Can someone merge this?

PR #536, opened 2015-01-02
Update Changelog

samuelstjean:
Stumbled upon this, how is the auto changelog with all the commits and info done? Just curious for a personal project.


arokem:
It's created using this script: 

https://github.com/nipy/dipy/blob/master/tools/github_stats.py

Adapted from IPython's tools

PR #538, opened 2015-01-03
WIP: TEST: relaxed precision

Garyfallidis:
Relaxed test decimal precision after bot error reported here
http://nipy.bic.berkeley.edu/builders/dipy-py3.3-pip/builds/138/steps/shell_4/logs/stdio

### DIFF
--- a/dipy/tracking/tests/test_propagation.py
+++ b/dipy/tracking/tests/test_propagation.py
@@ -86,7 +86,7 @@ def test_eudx_further():
                                         input_space=affine2)
     # Check that the tracks are the same
     for sl1, sl2 in zip(eu1, eu2_to_eu1):
-        assert_array_almost_equal(sl1, sl2)
+        assert_array_almost_equal(sl1, sl2, 4)
 
 
 def test_eudx_bad_seed():


arokem:
This will not help here. The problem is not one of precision. The array shapes are different!

Garyfallidis:
But the bot says : Arrays are not almost equal to 6 decimals

Garyfallidis:
Oh yeah. You are correct. There is a shape mismatch.

arokem:
Weird, huh? It's like move_streamlines sometimes drops a node, or EuDX
produces different numbers of nodes, depending on the affine used.

Garyfallidis:
Which both sound impossible!

Garyfallidis:
And we don't see them happening in other buildbots. This is strange...

arokem:
Agreed - and this was triggered by commits entirely unrelated to this piece of code. I suspect some foul play - anything from setuptools, through pip, to the spirit of Lee Harvey Oswald.

arokem:
Also, why is that bot not pulling the most recent version of the code? It seems that even this morning it was still pulling a version a few days old.

Garyfallidis:
Look here at shell 3
http://nipy.bic.berkeley.edu/builders/dipy-py3.3-pip/builds/139/steps/shell_3/logs/stdio
Why is it picking up dipy 0.7.1? And not dipy 0.8.0.dev or something?

Garyfallidis:
Closing this PR as we saw that the error appears only with old OSX versions (for example 0.6 or 0.8) and old python 3 versions. Specifically only with Python 3.3 not with Python 3.4.

PR #540, opened 2015-01-05
MRG: refactor propspeed

matthew-brett:
Refactoring propspeed to make me feel better about not understanding or being
able to debug strange error on buildbots.

PR #539, opened 2015-01-05
WIP: BF: Catching initial fodf creation of SDT 

Garyfallidis:
Julio Villalon has reported that on some Linux clusters with MKL (if I remember correctly) sometimes and in some voxels of some datasets the initial fodf creation in SDT gives a LinalgError and stops the execution. This PR catches this error and prohibits execution from stopping.

PR #541, opened 2015-01-05
BUG:   plot_2d_diffeomorphic_map

grlee77:
This pull request fixes two minor bugs in plot_2d_diffeomorphic_map that I came across when playing around with the new align functionality (which is great by the way!):

-  If fname is set, but show_figure is false, the call to plt.savefig() should not be performed

- The docstring states that `direct_grid_shape` and `inverse_grid_shape` should be a tuple, but if a tuple is used, the call to `mapping.transform()` raises an exception because the warp functions expect a `numpy.int32` array.

To fix this, I put a conversion to numpy.int32 within the `transform` method itself, because it is also an issue there if the provided `sampling_shape` array is not of type `numpy.int32`. 

I also removed a couple of unused mgrid variables from plot_2d_diffeomorphic_map and did some PEP8 formatting of regtools.py.

Finally, if interested, I can submit a separate pull request with an enhancement where draw_lattice_2d is generalized to draw_lattice_nd and there is a plot_diffeomorphic_map() function that works for either 2d or 3d cases.

### DIFF
--- a/dipy/align/imwarp.py
+++ b/dipy/align/imwarp.py
@@ -801,12 +801,14 @@ class DiffeomorphicMap(object):
         See _warp_forward and _warp_backward documentation for further
         information.
         """
+        if sampling_shape is not None:
+            sampling_shape = np.asarray(sampling_shape, dtype=np.int32)
         if self.is_inverse:
             warped = self._warp_backward(image, interpolation, world_to_image,
-                                       sampling_shape, sampling_affine)
+                                         sampling_shape, sampling_affine)
         else:
             warped = self._warp_forward(image, interpolation, world_to_image,
-                                       sampling_shape, sampling_affine)
+                                        sampling_shape, sampling_affine)
         return np.asarray(warped)
 
     def transform_inverse(self, image, interpolation='linear', world_to_image=None,
--- a/dipy/viz/regtools.py
+++ b/dipy/viz/regtools.py
@@ -204,7 +204,6 @@ def plot_2d_diffeomorphic_map(mapping, delta=10, fname=None,
         world_to_image = np.linalg.inv(inverse_grid_affine)
 
     # Draw the squares on the output grid
-    X1, X0 = np.mgrid[0:inverse_grid_shape[0], 0:inverse_grid_shape[1]]
     lattice_out = draw_lattice_2d(
         (inverse_grid_shape[0] + delta) / (delta + 1),
         (inverse_grid_shape[1] + delta) / (delta + 1),
@@ -222,7 +221,6 @@ def plot_2d_diffeomorphic_map(mapping, delta=10, fname=None,
         world_to_image = np.linalg.inv(direct_grid_affine)
 
     # Draw the squares on the input grid
-    X1, X0 = np.mgrid[0:direct_grid_shape[0], 0:direct_grid_shape[1]]
     lattice_in = draw_lattice_2d((direct_grid_shape[0] + delta) / (delta + 1),
                                  (direct_grid_shape[1] + delta) / (delta + 1),
                                  delta)
--- a/dipy/viz/regtools.py
+++ b/dipy/viz/regtools.py
@@ -3,23 +3,25 @@ from ..utils.optpkg import optional_package
 matplotlib, has_mpl, setup_module = optional_package("matplotlib")
 plt, _, _ = optional_package("matplotlib.pyplot")
 
+
 def overlay_images(img0, img1, title0='', title_mid='', title1='', fname=None):
     r""" Plot two images one on top of the other using red and green channels.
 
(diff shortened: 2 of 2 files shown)


grlee77:
For the 3D case, to keep things simple I only implemented computation of the warped grids, and raise an error if `show_figure` is True.  

As far as visualizing the 3D warp fields, there were two ways I tried:

1.) to look at it interactively, `OrthoSlicer3D` from this nipype pull request works well:
https://github.com/nipy/nibabel/pull/251

Although, ideally it would be nicer to only have grid lines in the two dimensions perpendicular to the axis being shown in the viewer, this would require a customized implementation of the class.

It would also be nice to have a subclass of this that takes two volumes and does red/green color overlays as a 3D version of regtools.overlay_slices()

2.) I have some code that can generate 2D mosaics from a 3D volume, but this is not as visually nice to look at.

grlee77:
I will wait until this is merged before putting the pull request for the nd case.  Here is what the lattice code looks like, though:

here the `slice_array` array is used to enable arbitrary lattice dimensions analagous to how the same sort of things is done in functions such as `numpy.gradient` which support nd inputs.

```python
def draw_lattice_nd(shape, gridspacings):
    r"""Create a regular lattice of black lines on a white background
    
    Parameters
    ----------
    shape : tuple or list
        size of the lattice along each dimension
    gridspacings : tuple or list
        grid spacing to use along each dimension. Along dimension i, black
        lines occur at np.arange(gridspacings[i], shape[i], gridspacings[i])

    Returns
    -------
    lattice : array
        A white volume (nD array) of the specified `shape`.  Black grid lines
        are spaced at the specified gridspacing along each dimension
        
    """
    ndim = len(shape)
    if len(gridspacings) != ndim:
        raise ValueError("shape and gridspacing lists must have the same "
                         "length")

    # white background
    lattice = 127 * np.ones(shape, dtype=np.float64)

    # loop over axes, adding black lines at the requested grid spacings
    slice_array = [slice(None)] * ndim
    for sl in range(ndim):
        slice_array[sl] = slice(gridspacings[sl], shape[sl], gridspacings[sl])
        lattice[slice_array] = 0
        slice_array[sl] = slice(None)
    return lattice
```

Then in the `plot_diffeomorphic_map`, the previous implementation for 2D:
```python
    # Draw the squares on the output grid
    lattice_out = draw_lattice_2d(
        (inverse_grid_shape[0] + delta) / (delta + 1),
        (inverse_grid_shape[1] + delta) / (delta + 1),
        delta)
    lattice_out = lattice_out[0:inverse_grid_shape[0], 0:inverse_grid_shape[1]]
```
is replaced by this nd case:
```python
    #Draw the squares on the output grid
    lattice_out=draw_lattice_nd(inverse_grid_shape, [delta,]*mapping.dim)
```

arokem:
I propose to wait with merging this for until after the current release.
The general plan is to release as soon as we can, without further review of
currently pending PRs, except those pertaining to stuff that has already
made it into the current release. We also plan to have another release at
the end of February, so this will follow the current release quickly,
dealing with odds and ends like this. Thanks for your patience with us.

grlee77:
Hi @arokem 

That is fine by me, although I would consider merging this present bug fix pull request and just wait on the nd implementation until the next release.  There is no new functinality implemented in the present pull request.

Garyfallidis:
Hi @grlee77 thank you for your PR. This is indeed very useful. But we have been delaying the release again and again and again. Now having more than 100 PRs in one release. It's about time to say no to including more and just release what we already have. We want also to change our release circle so we release more often. So, this shouldn't take long to be released. I hope you can understand this. Thank you again for your PR.

matthew-brett:
I would argue that bug-fixes should go in.   If they cause pain we can always release off the previous commit. 

I'm happy to take charge of the release.

arokem:
After some more thought, I have changed my mind here. I think that we
should merge this one before making the release, since it's simple and
fixes an obvious bug. That said, I think that we should make the release
tomorrow after merging this, without any further merges. In particular,
unless it fixes that buildbot bug, I think that #540 should wait. How does
that sound?

samuelstjean:
No #539  fix for the deconv  errors?

arokem:
I vote no. There are many ways in which dipy is not perfect. Only some of
them are addressed in this release. That's why we keep working on this. A
year between releases is way too long, and if we aim to make this perfect
before we ever release, we will never make a release. We should deal with
some of the flaws right now, and keep making it better, starting tomorrow

matthew-brett:
It would be nice to have a test of the sampling_shape fix.

matthew-brett:
Guys - I combined this with #542 and added another fix, did some more work - see https://github.com/nipy/dipy/pull/543

Garyfallidis:
I am closing this PR as this PR has been included into #543.

PR #553, opened 2015-01-09
RF: use cholesky to solve csd

MrBago:
I've implemented the CSD fitting using the Cholesky decomposition. Here are the fitting times I get with the new and old versions:

New:
SH order - 8, gradient directons - 150 :: 1.76 sec
SH order - 8, gradient directons - 65 :: 2.01 sec
SH order - 12, gradient directons - 65 :: 5.77 sec

Old:
SH order - 8, gradient directons - 150 :: 7.81 sec
SH order - 8, gradient directons - 65 :: 8.22 sec
SH order - 12, gradient directons - 65 :: 30.87 sec

I'd like to include the document I got from @Garyfallidis as a reference in this PR somehow, but it's a PDF and I'm not sure the best way to do it. Should we maybe upload the PDF tot he website and include a link?

### DIFF
--- a/dipy/reconst/csdeconv.py
+++ b/dipy/reconst/csdeconv.py
@@ -499,10 +499,10 @@ def csdeconv(dwsignal, X, B_reg, tau=0.1, convergence=50, P=None):
     fodf = np.dot(B_reg[:, :15], fodf_sh[:15])
     # set threshold on FOD amplitude used to identify 'negative' values
     threshold = tau * np.mean(fodf)
-    k_last = (fodf < threshold)
+    where_fodf_small = (fodf < threshold).nonzero()[0]
 
     # If all fodf values are greater than zero
-    if not k_last.any():
+    if len(where_fodf_small) == 0:
         return fodf_sh, 0
 
     for num_it in range(1, convergence + 1):
@@ -512,8 +512,7 @@ def csdeconv(dwsignal, X, B_reg, tau=0.1, convergence=50, P=None):
         # sense, this "adds" a measurement, which can help to better estimate
         # the fodf_sh, even if you have more SH coefficients to estimate than
         # actual S measurements.
-        index = k_last.nonzero()[0]
-        H = B_reg.take(index, axis=0)
+        H = B_reg.take(where_fodf_small, axis=0)
 
         # We use the Cholesky decomposition to solve for the SH coefficients.
         Q = P + np.dot(H.T, H)
@@ -521,11 +520,12 @@ def csdeconv(dwsignal, X, B_reg, tau=0.1, convergence=50, P=None):
 
         # Sample the FOD using the regularization sphere and compute k.
         fodf = np.dot(B_reg, fodf_sh)
-        k = (fodf < threshold)
+        where_fodf_small_last = where_fodf_small
+        where_fodf_small = (fodf < threshold).nonzero()[0]
 
-        if (k == k_last).all():
+        if (len(where_fodf_small) == len(where_fodf_small_last) and
+            (where_fodf_small == where_fodf_small_last).all()):
             break
-        k_last = k
     else:
         msg = 'maximum number of iterations exceeded - failed to converge'
         warnings.warn(msg)
--- a/dipy/reconst/csdeconv.py
+++ b/dipy/reconst/csdeconv.py
@@ -353,9 +353,21 @@ def forward_sdt_deconv_mat(ratio, n, r2_term=False):
     return np.diag(b), np.diag(bb)
 
 
+potrf, = la.lapack.get_lapack_funcs(('potrf',))
+potrs, = la.lapack.get_lapack_funcs(('potrs',))
+
 def _solve_cholesky(Q, z):
-    L = la.cho_factor(Q)
-    f = la.cho_solve(L, z)
+    L, info = potrf(Q, lower=False, overwrite_a=False, clean=False)
+    if info > 0:
+        msg = "%d-th leading minor not positive definite" % info
+        raise la.LinAlgError(msg)
+    if info < 0:
+        msg = 'illegal value in %d-th argument of internal potrf' % -info
+        raise ValueError(msg)
+    f, info = potrs(L, z, lower=False, overwrite_b=False)
+    if info != 0:
+        msg = 'illegal value in %d-th argument of internal potrs' % -info
+        raise ValueError(msg)
     return f
 
 
@@ -500,7 +512,8 @@ def csdeconv(dwsignal, X, B_reg, tau=0.1, convergence=50, P=None):
         # sense, this "adds" a measurement, which can help to better estimate
         # the fodf_sh, even if you have more SH coefficients to estimate than
         # actual S measurements.
-        H = B_reg[k_last]
+        index = k_last.nonzero()[0]
+        H = B_reg.take(index, axis=0)
 
         # We use the Cholesky decomposition to solve for the SH coefficients.
         Q = P + np.dot(H.T, H)
--- a/dipy/reconst/csdeconv.py
+++ b/dipy/reconst/csdeconv.py
@@ -120,18 +120,20 @@ class ConstrainedSphericalDeconvModel(SphHarmModel):
         # scale lambda_ to account for differences in the number of
         # SH coefficients and number of mapped directions
         # This is exactly what is done in [4]_
-        self.lambda_ = (lambda_  * self.R.shape[0] * r_rh[0] /
-                        (np.sqrt(self.B_reg.shape[0]) * np.sqrt(362.))
-                       )
+
+        lambda_ = (lambda_  * self.R.shape[0] * r_rh[0] /
+                   (np.sqrt(self.B_reg.shape[0]) * np.sqrt(362.)))
+        self.B_reg *= lambda_
         self.sh_order = sh_order
         self.tau = tau
+        self._X = X = self.R.diagonal() * self.B_dwi
+        self._P = np.dot(X.T, X)
 
     @multi_voxel_fit
     def fit(self, data):
         dwi_data = data[self._where_dwi]
-        X = self.R.diagonal() * self.B_dwi
-        shm_coeff, num_it = csdeconv(dwi_data, self.sh_order, X, self.B_reg,
-                                     self.lambda_, self.tau)
+        shm_coeff, _ = csdeconv(dwi_data, self.sh_order, self._X, self.B_reg,
+                                self.tau, P=self._P)
         return SphHarmFit(self, shm_coeff, None)
 
 
@@ -357,8 +359,7 @@ def _solve_cholesky(Q, z):
     return f
 
 
-def csdeconv(dwsignal, sh_order, X, B_reg, lambda_=1., tau=0.1,
-             convergence=50):
+def csdeconv(dwsignal, sh_order, X, B_reg, tau=0.1, convergence=50, P=None):
     r""" Constrained-regularized spherical deconvolution (CSD) [1]_
 
     Deconvolves the axially symmetric single fiber response function `r_rh` in
@@ -376,9 +377,7 @@ def csdeconv(dwsignal, sh_order, X, B_reg, lambda_=1., tau=0.1,
         coefficients.
     B_reg : array (N, B)
         SH basis matrix which maps FOD coefficients to FOD values on the
-        surface of the sphere.
-    lambda_ : float
-        lambda parameter in minimization equation (default 1.0)
+        surface of the sphere. B_reg should be scaled to account for lambda.
     tau : float
         Threshold controlling the amplitude below which the corresponding fODF
         is assumed to be zero.  Ideally, tau should be set to zero. However, to
@@ -390,6 +389,10 @@ def csdeconv(dwsignal, sh_order, X, B_reg, lambda_=1., tau=0.1,
         more stable.
     convergence : int
         Maximum number of iterations to allow the deconvolution to converge.
+    P : ndarray
+        This is an optimization to avoid computing ``dot(X.T, X)`` many times.
+        If the same ``X`` is used many times, ``P`` can be precomputed and
+        passed to this function.
 
     Returns
     -------
@@ -408,13 +411,14 @@ def csdeconv(dwsignal, sh_order, X, B_reg, lambda_=1., tau=0.1,
 
     """
     mu = 1e-5
-    P = np.dot(X.T, X)
+    if P is None:
+        P = np.dot(X.T, X)
     z = np.dot(X.T, dwsignal)
 
     try:
         fodf_sh = _solve_cholesky(P, z)
     except la.LinAlgError:
-        P += mu * np.eye(P.shape[0])
+        P = P + mu * np.eye(P.shape[0])
         fodf_sh = _solve_cholesky(P, z)
     # For the first iteration we use a smooth FOD that only uses SH orders up
     # to 4 (the first 15 coefficients.
@@ -427,8 +431,6 @@ def csdeconv(dwsignal, sh_order, X, B_reg, lambda_=1., tau=0.1,
     if not k_last.any():
         return fodf_sh, 0
 
-    lambda2 = lambda_ ** 2
-
     for num_it in range(1, convergence + 1):
         # This is the super-resolved trick.  Wherever there is a negative
         # amplitude value on the fODF, it concatenates a value to the S vector
@@ -439,7 +441,7 @@ def csdeconv(dwsignal, sh_order, X, B_reg, lambda_=1., tau=0.1,
         H = B_reg[k_last]
 
         # We use the Cholesky decomposition to solve for the SH coefficients.
-        Q = P + lambda2 * np.dot(H.T, H)
+        Q = P + np.dot(H.T, H)
         fodf_sh = _solve_cholesky(Q, z)
 
         # Sample the FOD using the regularization sphere and compute k.
--- a/dipy/reconst/csdeconv.py
+++ b/dipy/reconst/csdeconv.py
@@ -1,16 +1,18 @@
 from __future__ import division, print_function, absolute_import
 import warnings
+
 import numpy as np
 from scipy.integrate import quad
 from scipy.special import lpn, gamma
+import scipy.linalg as la
 
-from dipy.reconst.multi_voxel import multi_voxel_fit
 from dipy.data import small_sphere, get_sphere
 from dipy.core.geometry import cart2sphere
 from dipy.core.ndindex import ndindex
 from dipy.sims.voxel import single_tensor
 from dipy.utils.six.moves import range
 
+from dipy.reconst.multi_voxel import multi_voxel_fit
 from dipy.reconst.dti import TensorModel, fractional_anisotropy
 from dipy.reconst.shm import (sph_harm_ind_list, real_sph_harm,
                               sph_harm_lookup, lazy_index, SphHarmFit,
@@ -349,7 +351,14 @@ def forward_sdt_deconv_mat(ratio, n, r2_term=False):
     return np.diag(b), np.diag(bb)
 
 
-def csdeconv(dwsignal, sh_order, X, B_reg, lambda_=1., tau=0.1):
+def _solve_cholesky(Q, z):
+    L = la.cho_factor(Q)
+    f = la.cho_solve(L, z)
+    return f
+
+
+def csdeconv(dwsignal, sh_order, X, B_reg, lambda_=1., tau=0.1,
+             convergence=50):
     r""" Constrained-regularized spherical deconvolution (CSD) [1]_
 
     Deconvolves the axially symmetric single fiber response function `r_rh` in
@@ -379,6 +388,8 @@ def csdeconv(dwsignal, sh_order, X, B_reg, lambda_=1., tau=0.1):
         peaks. Because SDT is based on a q-ball ODF deconvolution, and not
         signal deconvolution, using the max instead of mean (as in CSD), is
         more stable.
+    convergence : int
+        Maximum number of iterations to allow the deconvolution to converge.
 
     Returns
     -------
@@ -396,48 +407,52 @@ def csdeconv(dwsignal, sh_order, X, B_reg, lambda_=1., tau=0.1):
            constrained super-resolved spherical deconvolution.
 
     """
-    # generate initial fODF estimate, truncated at SH order 4
-    fodf_sh = np.linalg.lstsq(X, dwsignal)[0]
-    fodf_sh[15:] = 0
+    mu = 1e-5
+    P = np.dot(X.T, X)
+    z = np.dot(X.T, dwsignal)
 
-    fodf = np.dot(B_reg, fodf_sh)
+    try:
+        fodf_sh = _solve_cholesky(P, z)
+    except la.LinAlgError:
+        P += mu * np.eye(P.shape[0])
+        fodf_sh = _solve_cholesky(P, z)
+    # For the first iteration we use a smooth FOD that only uses SH orders up
+    # to 4 (the first 15 coefficients.
+    fodf = np.dot(B_reg[:, :15], fodf_sh[:15])
     # set threshold on FOD amplitude used to identify 'negative' values
-    threshold = tau * np.mean(np.dot(B_reg, fodf_sh))
+    threshold = tau * np.mean(fodf)
+    k_last = (fodf < threshold)
 
-    k = []
-    convergence = 50
+    # If all fodf values are greater than zero
+    if not k_last.any():
+        return fodf_sh, 0
+
+    lambda2 = lambda_ ** 2
+
     for num_it in range(1, convergence + 1):
+        # This is the super-resolved trick.  Wherever there is a negative
+        # amplitude value on the fODF, it concatenates a value to the S vector
+        # so that the estimation can focus on trying to eliminate it. In a
+        # sense, this "adds" a measurement, which can help to better estimate
+        # the fodf_sh, even if you have more SH coefficients to estimate than
+        # actual S measurements.
+        H = B_reg[k_last]
+
+        # We use the Cholesky decomposition to solve for the SH coefficients.
+        Q = P + lambda2 * np.dot(H.T, H)
+        fodf_sh = _solve_cholesky(Q, z)
+
+        # Sample the FOD using the regularization sphere and compute k.
         fodf = np.dot(B_reg, fodf_sh)
+        k = (fodf < threshold)
 
-        k2 = np.nonzero(fodf < threshold)[0]
+        if (k == k_last).all():
+            break
+        k_last = k
+    else:
+        msg = 'maximum number of iterations exceeded - failed to converge'
+        warnings.warn(msg)
 
-        if (k2.shape[0] + X.shape[0]) < B_reg.shape[1]:
-            warnings.warn(
-            'too few negative directions identified - failed to converge')
-            return fodf_sh, num_it
-
-        if num_it > 1 and k.shape[0] == k2.shape[0]:
-            if (k == k2).all():
-                return fodf_sh, num_it
-
-        k = k2
-
-        # This is the super-resolved trick.
-        # Wherever there is a negative amplitude value on the fODF, it
-        # concatenates a value to the S vector so that the estimation can
-        # focus on trying to eliminate it. In a sense, this "adds" a
-        # measurement, which can help to better estimate the fodf_sh, even if
-        # you have more SH coeffcients to estimate than actual S measurements.
-        M = np.concatenate((X, lambda_ * B_reg[k, :]))
-        S = np.concatenate((dwsignal, np.zeros(k.shape)))
-        try:
-            fodf_sh = np.linalg.lstsq(M, S)[0]
-        except np.linalg.LinAlgError as lae:
-            # SVD did not converge in Linear Least Squares in current
-            # voxel. Proceeding with initial SH estimate for this voxel.
-            pass
-
-    warnings.warn('maximum number of iterations exceeded - failed to converge')
     return fodf_sh, num_it
 
 
--- a/dipy/reconst/csdeconv.py
+++ b/dipy/reconst/csdeconv.py
@@ -496,13 +496,18 @@ def csdeconv(dwsignal, X, B_reg, tau=0.1, convergence=50, P=None):
     # For the first iteration we use a smooth FOD that only uses SH orders up
     # to 4 (the first 15 coefficients).
     fodf = np.dot(B_reg[:, :15], fodf_sh[:15])
-    # set threshold on FOD amplitude used to identify 'negative' values
-    threshold = tau * np.mean(fodf)
+    # The mean of an fodf can be computed by taking $Y_{0,0} * coeff_{0,0}$
+    threshold = B_reg[0, 0] * fodf_sh[0] * tau
     where_fodf_small = (fodf < threshold).nonzero()[0]
 
-    # If all fodf values are greater than zero
+    # If the low-order fodf does not have any values less than threshold, the
+    # full-order fodf is used.
     if len(where_fodf_small) == 0:
-        return fodf_sh, 0
+        fodf = np.dot(B_reg, fodf_sh)
+        where_fodf_small = (fodf < threshold).nonzero()[0]
+        # If the fodf still has no values less than threshold, return the fodf.
+        if len(where_fodf_small) == 0:
+            return fodf_sh, 0
 
     for num_it in range(1, convergence + 1):
         # This is the super-resolved trick.  Wherever there is a negative
--- a/dipy/reconst/csdeconv.py
+++ b/dipy/reconst/csdeconv.py
@@ -353,8 +353,7 @@ def forward_sdt_deconv_mat(ratio, n, r2_term=False):
     return np.diag(b), np.diag(bb)
 
 
-potrf, = la.lapack.get_lapack_funcs(('potrf',))
-potrs, = la.lapack.get_lapack_funcs(('potrs',))
+potrf, potrs = la.get_lapack_funcs(('potrf', 'potrs'))
 
 def _solve_cholesky(Q, z):
     L, info = potrf(Q, lower=False, overwrite_a=False, clean=False)
--- a/dipy/reconst/csdeconv.py
+++ b/dipy/reconst/csdeconv.py
@@ -132,8 +132,8 @@ class ConstrainedSphericalDeconvModel(SphHarmModel):
     @multi_voxel_fit
     def fit(self, data):
         dwi_data = data[self._where_dwi]
-        shm_coeff, _ = csdeconv(dwi_data, self.sh_order, self._X, self.B_reg,
-                                self.tau, P=self._P)
+        shm_coeff, _ = csdeconv(dwi_data, self._X, self.B_reg, self.tau,
+                                P=self._P)
         return SphHarmFit(self, shm_coeff, None)
 
 
@@ -359,7 +359,7 @@ def _solve_cholesky(Q, z):
     return f
 
 
-def csdeconv(dwsignal, sh_order, X, B_reg, tau=0.1, convergence=50, P=None):
+def csdeconv(dwsignal, X, B_reg, tau=0.1, convergence=50, P=None):
     r""" Constrained-regularized spherical deconvolution (CSD) [1]_
 
     Deconvolves the axially symmetric single fiber response function `r_rh` in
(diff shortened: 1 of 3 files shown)


arokem:
First of all - score!

Second - Is the document easily converted to markdown? Or in other words:
what's in it?

MrBago:
The [doc is here](https://drive.google.com/file/d/0BxXijiXAopKPSjVxdENfS2RiUXZkaUJFMUtGcHJXR1NoVUlN/view?usp=sharing). I guess it _could_ be converted to markdown but I'm not sure how to re-enter all the formulas.

samuelstjean:
Ask for the latex source, convert to markdown through something like pandoc I guess.

arokem:
Unless, of course, it was written in the google-docs equation editor (sure
looks like it) => no latex source.

This seems to have a way to do that:

http://tex.stackexchange.com/questions/183358/how-to-convert-a-google-docs-document-to-latex

Of course, by this point, we could already have transcribed it by hand, but
this is much more fun.

arokem:
## Fast CSD

Problem is to minimise per iteration:

$F(f_n)  = ||Mf_n -s||^2 + \lambda||H_{n-1} f_n ||^2$

Where $M$ maps current FOD SH coefficients $f_n$ coefficients to DW signals $s$ and $H_n-1$ maps FOD SH coefficients $f_n$ to amplitudes along set of negative directions identified in previous iteration (i.e. the matrix formed by the rows of $H$ for which $Hf_{n-1}<0$ where $H$ maps FOD to amplitude along a dense set of (300) directions

Solve by differentiating and setting to zero:

$\Rightarrow \frac{\deltaF}{\delta f_n} = 2M^T(Mf_n - s) + 2\lambdaH_{n-1}^TH_{n-1}f_n$ = 0$ 

Define $Q = M^TM + \lambdaH_{n-1}^TH_{n-1}$ , which by construction is a square positive­definite symmetric matrix of size $n_{SH} × n_{SH}$ . If needed, positive definiteness can be enforced with a small minimum norm regulariser (helps a lot with poorly conditioned direction sets and/or super­resolution):

$Q = M^TM + \lambdaH_{n-1}^TH_{n-1} + \muI$ 

Solve using Cholesky decomposition:

$Q = LL^T$

where $L$ is lower triangular. Then problem can be solved by back­substitution:

$L_y = M^Ts$

$L^Tf_n = y$

To speeds things up further, form $P = M^TM + \muI$, and update to form $Q$ by rank­n update with $H_{n-1}$. Form $M^Ts$ initially, and re­use over iterations since it doesn’t change. Algorithm then looks like:

   - form $P = M^T M + \muI$

   - for each voxel:
     - form $z = M^Ts$
     - estimate $f_0$ by solving $Pf_0=z$ (note in practive I use a simplified $l_max=4$
     solution here, but I'm not sure it makes a great deal of difference what the FOD actually is).
     - The iterate until no change in rows of $H$ used in $H_n$
       - form $H_{n−1}$ given $f_{n−1}$
       - form $Q = P + \lambdaH_{n−1}^ H_{n-1}$ by rank­n update
       - solve $Qf_n = z$ using Cholesky decomposition

arokem:
I think you might want to add this to the docs as rst. Maybe we could have a theory section, where we put things like this. I would like to extend the documentation to have some more theoretical comments. I think that would be very helpful

MrBago:
Thanks! Eleftherios has emailed donald tournier to get permission and the
full attribution. I'll add this once we hear from him.

arokem:
Gotcha. Does this come from one of his papers? Or is this just the
algorithm they use under the hood in mrtrix?

MrBago:
It's west they use in mrtrix, but it's not described in the papers.

Bago

stefanv:
Perhaps it would be good to check with Jonathan Taylor if Cholesky is the best approach for this problem?  Whenever you have to form ``M.T.dot(M)``, alarm bells go off (sometimes it's necessary, but can also often be avoided by solving another, equivalent minimization problem).  If you'd like to, I am happy to contact him.

MrBago:
I'm happy have Jonathan's input, feel free to give him my info if he's
willing to help us with this. So far we've used the leastsq approach, which
works fine but is about 4x slower than the cholesky. I've also been looking
at using a Quadratic programming approach,
http://en.wikipedia.org/wiki/Quadratic_programming, to fitting the
coefficients, but based on 1 or 2 cases it does not do as well.

Bago

stefanv:
I've pinged Jonathan.

dipy/reconst/csdeconv.py
```diff
@@ -397,48 +410,51 @@ def csdeconv(dwsignal, sh_order, X, B_reg, lambda_=1., tau=0.1):
...
+        P = np.dot(X.T, X)
+    z = np.dot(X.T, dwsignal)
+
+    try:
+        fodf_sh = _solve_cholesky(P, z)
+    except la.LinAlgError:
+        P = P + mu * np.eye(P.shape[0])
+        fodf_sh = _solve_cholesky(P, z)
+    # For the first iteration we use a smooth FOD that only uses SH orders up
+    # to 4 (the first 15 coefficients.
```
Garyfallidis: closing parenthesis is missing

dipy/reconst/csdeconv.py
```diff
@@ -397,48 +410,51 @@ def csdeconv(dwsignal, sh_order, X, B_reg, lambda_=1., tau=0.1):
...
+        # the fodf_sh, even if you have more SH coefficients to estimate than
+        # actual S measurements.
+        H = B_reg[k_last]
+
+        # We use the Cholesky decomposition to solve for the SH coefficients.
+        Q = P + np.dot(H.T, H)
+        fodf_sh = _solve_cholesky(Q, z)
+
+        # Sample the FOD using the regularization sphere and compute k.
         fodf = np.dot(B_reg, fodf_sh)
```
Garyfallidis: While line takes now more time? 445 or 448?

Garyfallidis:
Hi @MrBago, just to repeat here that you can put online the information from Donald about this factorization. We got his permission. 

The same technique can be used in the odf_deconv am I correct? Can you apply that there too. That would be super! 

Another question is if you have compared the speed between numpy and scipy cholesky factorization. 

I think you need to rebase this PR.

MrBago:
Thanks @Garyfallidis, I believe we should be able to have just one function that's called by both models. I looked at this some time ago, but I'll need to look at at again.

Garyfallidis:
Okay let us know when you think this PR is ready.

dipy/reconst/csdeconv.py
MrBago: Line numbers have changes now, but in a quick test it looks like this:

    fodf_sh = _solve_cholesky(Q, z) - 46%
    Q = P + np.dot(H.T, H) - 19%
    H = B_reg[k_last] - 19%

`fodf = np.dot(B_reg, fodf_sh)` is not even close, less than 2%.

MrBago:
I think this is ready to go unless you guys would like me to change anything.

dipy/reconst/csdeconv.py
Garyfallidis: A bit unexpected for Q and H to be that slower than fodf. Now that we are on it let's see if this can be improved a bit. @stefanv do you think we can improve the performance of those 3 lines?

matthew-brett: Which of those three lines is the slow one?

Garyfallidis: @matthew-brett Bago is giving the timings with percentages for all 3 just above. The fodf_sh takes 46%, the Q 19%, and H 18% of the time. Improving any of those it's a gain because this function runs for each voxel.  I am not proposing cythonizing them but it would be nice if we could improve the speed by refactoring a bit. For example the line with H is doing a simple indexing and copy. It looks too me too slow taking 18% when the fodf line takes 2%. So, we should be able to improve that. No?

Garyfallidis: For example can we maybe use take rather than as it now in line with H as explained here: 

In [16]: A = np.random.rand(724, 724)
In [17]: index = np.random.randint(0, 724, 400)
In [18]: %timeit -n 100 b =  A.take(index, axis=0)
100 loops, best of 3: 305 µs per loop
In [19]: %timeit -n 100 b =  A[index]
100 loops, best of 3: 2.53 ms per loop

Here take is about 10X faster than standard indexing.

Garyfallidis: Okay I think if you replace
H = B_reg[k_last]
with 
k_index = np.where(k_last==True)[0]
H = B_reg.take(k_index, axis=0)
your speed will be improved a bit.

Here is the benchmark timings in my notebook:

As it was

SH order - 8, gradient directons - 150 :: 1.6 sec
SH order - 8, gradient directons - 65 :: 1.88 sec
SH order - 12, gradient directons - 65 :: 5.19 sec

With take

SH order - 8, gradient directons - 150 :: 1.38 sec
SH order - 8, gradient directons - 65 :: 1.6 sec
SH order - 12, gradient directons - 65 :: 4.36 sec

Garyfallidis: Other suggestions/improvements? @matthew-brett ?

arokem: Does this mean we should replace all our indexing operations by `take`?

Garyfallidis: No, I don't think so. Here we are working on speeding up the CSD. This is the main purpose of this PR. Using take  is useful only for optimizing for speed in critical areas. No need to change it everywhere. But it is good to know that it is faster. If you had any slowdowns for SFM or LiFE when indexing arrays you should consider it.

arokem: Sorry for derailing the discussion. I was simply mystified by this.

MrBago:
I've made @Garyfallidis's suggested change and also made `_solve_cholesky` faster. Here is what I get rerunning the timings:

Original:
> All files already in /Users/bagrata/.dipy/stanford_hardi.
> == Benchmarking CSD fit on 1413 voxels ==
> SH order - 8, gradient directons - 150 :: 18.66 sec
> SH order - 8, gradient directons - 65 :: 19.46 sec
> SH order - 12, gradient directons - 65 :: 71.11 sec

Cholesky version 1:
> == Benchmarking CSD fit on 1413 voxels ==
> SH order - 8, gradient directons - 150 :: 3.9 sec
> SH order - 8, gradient directons - 65 :: 4.63 sec
> SH order - 12, gradient directons - 65 :: 13.44 sec

Cholesky with lastest updates:
> == Benchmarking CSD fit on 1413 voxels ==
> SH order - 8, gradient directons - 150 :: 1.73 sec
> SH order - 8, gradient directons - 65 :: 1.98 sec
> SH order - 12, gradient directons - 65 :: 7.08 sec

Now the slowest lines are:
        Q = P + np.dot(H.T, H) - 43%
        fodf_sh = _solve_cholesky(Q, z) - 28%
        H = B_reg.take(index, axis=0) - 4%

Garyfallidis:
I am waiting for Travis to finish to express my enthusiasm!

Garyfallidis:
Woot! Woot! Woot! 10X speedup!

Garyfallidis:
If nobody disagrees I would like to merge this one.

arokem:
+1. Both to the wooting and to the merging.

samuelstjean:
Does it give the same result as the old version? Even if the test passes, since it's only changing the linear algrebra part it should give the same coefficients as the old least square fit right?

arokem:
Hang on with merging. I think there are a couple of errors in the latex in the docs.

dipy/reconst/csdeconv.py
```diff
@@ -379,15 +397,85 @@ def csdeconv(dwsignal, sh_order, X, B_reg, lambda_=1., tau=0.1):
...
+
+    Notes
+    -----
+    This section describes how the fitting of the SH coefficients is done.
+    Problem is to minimise per iteration:
+
+    $F(f_n) = ||Xf_n - S||^2 + \lambda^2 ||H_{n-1} f_n||^2$
+
+    Where $X$ maps current FOD SH coefficients $f_n$ to DW signals $s$ and
+    $H_n-1$ maps FOD SH coefficients $f_n$ to amplitudes along set of negative
```
arokem: I believe this should be H_{n-1}

dipy/reconst/csdeconv.py
```diff
@@ -379,15 +397,85 @@ def csdeconv(dwsignal, sh_order, X, B_reg, lambda_=1., tau=0.1):
...
+
+    $\Rightarrow \frac{\delta F}{\delta f_n} = 2X^T(Xf_n - S) + 2 \lambda^2
+    H_{n-1}^TH_{n-1}f_n=0$
+
+    Or:
+
+    $(X^TX + \lambda^2 H_{n-1}^TH_{n-1})f_n = X^Ts$
+
+    Define $Q = X^TX + \lambda^2 H_{n-1}^TH_{n-1}$ , which by construction is a
+    square positive definite symmetric matrix of size $(n_{SH} n_{SH})$. If
```
arokem: => "of size $(n_{SH} by n_{SH})$"

dipy/reconst/csdeconv.py
```diff
@@ -379,15 +397,85 @@ def csdeconv(dwsignal, sh_order, X, B_reg, lambda_=1., tau=0.1):
...
+    $L^Tf_n = y$
+
+    To speeds things up further, form $P = X^TX + \mu I$, and update to form
+    $Q$ by rankn update with $H_{n-1}$. The dipy implementation looks like:
+
+        form initially $P = X^T X + \mu I$ and $\lambda B_{reg}$
+
+        for each voxel: form $z = X^Ts$
+
+            estimate $f_0$ by solving $Pf_0=z$. We use a simplified $l_max=4$
```
arokem: => "$l_{max} = 4$"

MrBago:
Thanks @arokem for catching my typos. I tried one last thing, using blas's rank k update to make `H.T.dot(H) + P` faster. It actually seems to be slower than leaving it as is. I may have done something wrong, but [here](https://gist.github.com/MrBago/4ef447eb74b45650a1da) is gist of what I did if anyone want's to mess with it.

I think this is ready unless you guys catch anything else.

dipy/reconst/csdeconv.py
```diff
@@ -118,18 +120,20 @@ def __init__(self, gtab, response, reg_sphere=None, sh_order=8, lambda_=1,
         # scale lambda_ to account for differences in the number of
         # SH coefficients and number of mapped directions
         # This is exactly what is done in [4]_
-        self.lambda_ = (lambda_  * self.R.shape[0] * r_rh[0] /
-                        (np.sqrt(self.B_reg.shape[0]) * np.sqrt(362.))
-                       )
+
+        lambda_ = (lambda_  * self.R.shape[0] * r_rh[0] /
+                   (np.sqrt(self.B_reg.shape[0]) * np.sqrt(362.)))
```
samuelstjean: Why the hardcoded 362? I'm guessing it's the number of vertices in the default regularization sphere, so if that's the case it should point to the number of vertices of self.sphere.

dipy/reconst/csdeconv.py
```diff
@@ -379,15 +396,85 @@ def csdeconv(dwsignal, sh_order, X, B_reg, lambda_=1., tau=0.1):
...
+
+    Notes
+    -----
+    This section describes how the fitting of the SH coefficients is done.
+    Problem is to minimise per iteration:
+
+    $F(f_n) = ||Xf_n - S||^2 + \lambda^2 ||H_{n-1} f_n||^2$
+
+    Where $X$ maps current FOD SH coefficients $f_n$ to DW signals $s$ and
+    $H_{n-1}$ maps FOD SH coefficients $f_n$ to amplitudes along set of negative
```
samuelstjean: along the set of

dipy/reconst/csdeconv.py
```diff
@@ -379,15 +396,85 @@ def csdeconv(dwsignal, sh_order, X, B_reg, lambda_=1., tau=0.1):
...
+    directions identified in previous iteration, i.e. the matrix formed by the
```
samuelstjean: iterationS

dipy/reconst/csdeconv.py
```diff
@@ -379,15 +396,85 @@ def csdeconv(dwsignal, sh_order, X, B_reg, lambda_=1., tau=0.1):
...
+    regulariser (helps a lot with poorly conditioned direction sets and/or
+    superresolution):
+
+    $Q = X^TX + (\lambda H_{n-1}^T) (\lambda H_{n-1}) + \mu I$
+
+    Solve $Qf_n = X^Ts$ using Cholesky decomposition:
+
+    $Q = LL^T$
+
+    where $L$ is lower triangular. Then problem can be solved by
```
samuelstjean: Then the problem

stefanv:
Well, this is embarrassing--my new email address caused a bit of havoc with my email setup, so many of my emails for the past week went to ``/dev/null``.  I'll re-ping Jonathan.

dipy/reconst/csdeconv.py
MrBago: This is a historical accident, any number ~20 would do here, but we chose `sqrt(362)` to be consistent with older versions of dipy where the value was implied instead of explicit.

matthew-brett: Some note about this in the comments maybe?
​

samuelstjean: Or why not just fix it instead, if the old historical version is a hardcoded value just because (but is it related to the number of points of the sphere, that's what I was curious about).

MrBago: @samuelstjean I'm not sure what you would fix, here is the historical accident:

The CSD equation is:

 $F(f_n) = ||Xf_n - S||^2 + \lambda^2 ||H_{n-1} f_n||^2$

Turns out this equation will give a different result if you use a more dense sphere (sphere with more points), unless you also change $\lambda$. So we can reformulate the the equation as:

 $F(f_n) = ||Xf_n - S||^2 + (\lambda')^2 / N * ||H_{n-1} f_n||^2$

Now the result is not sensitive to the number of points on the sphere and the $\lambda'$ can be used for all spheres (but the result is still sensitive to how well the points are distributed on the sphere). To make sure that dipy gives reasonable result by default we need to set $\lambda'_{default} \approx \lambda_{default} * 20$. This method would work for code which uses all defaults, but existing user code which set $\lambda$ explicitly would give different results before and after the change. Instead we can write the equation as:

 $F(f_n) = ||Xf_n - S||^2 + (\lambda')^2 * N_0 / N  *||H_{n-1} f_n||^2$

With this formulation when $N = N_0$ $\lambda'$ can simply be set to $\lambda$ and dipy will give the same result. Now the question is how to pick $N_0$. We picked $N_0 = 362$ because that's the number of points on the default sphere, which we assume is the sphere that is most often used by our users.

I'm not sure there is anything to fix here, except maybe adding this blurb as a comment so that I can avoid this conversation in the future :).

MrBago: Also, does anyone know where one can post something like my previous comment and have the math rendered legibly to make it easier for other to read?

samuelstjean: Well, if I change the regularization sphere, without changing lambda, that would defeat the purpose of changing the regularization sphere if $N_0$ is set to 362, and not the actual sphere I passed. And existing code would only be different if someone is passing another sphere in this case, where it would probably make more sense to use something else than 362. Or I don't see why the arbitrary $\lambda' ~ 20 * \lambda$ and it's just a quirk that sqrt(362) ~ 20.

arokem: https://www.authorea.com/users/398/articles/16952

MrBago: On Tue, Jan 27, 2015 at 1:21 PM, Samuel St-Jean <notifications@github.com>
wrote:
>
> Well, if I change the regularization sphere, without changing lambda, that
> would defeat the purpose of changing the regularization sphere if $N_0$ is
> set to 362, and not the actual sphere I passed. And existing code would
> only be different if someone is passing another sphere in this case, where
> it would probably make more sense to use something else than 362. Or I
> don't see why the arbitrary $\lambda' ~ 20 * \lambda$ and it's just a quirk
> that sqrt(362) ~ 20.
>
>
Can you please explain this? I don't understand.

samuelstjean: Maybe I got this all wrong, but what I get from this is : 

1. We need a sqrt(362) to accomodate the old hardcoded regularisation sphere.
2. Instead of changing that, users can play with $lambda'$, which kinda takes into account that hardcoded value.
3. We could change sqrt(362) to actually be the number of point on the regularisation sphere, also defaulting to 362 for the default code path, but also taking into account any other sampled sphere. Granted, anyone using another sphere will get a deffiretn value from now on, but that's correcting a hardcoded value at the same time, which is more sane in the long run.

arokem: I think that what @samuelstjean might be proposing is to change N_0 to be 1.

MrBago: On Tue, Jan 27, 2015 at 2:00 PM, Samuel St-Jean <notifications@github.com>
wrote:

> In dipy/reconst/csdeconv.py
> <https://github.com/nipy/dipy/pull/553#discussion_r23647036>:
>
> > @@ -118,18 +120,20 @@ def __init__(self, gtab, response, reg_sphere=None, sh_order=8, lambda_=1,
> >          # scale lambda_ to account for differences in the number of
> >          # SH coefficients and number of mapped directions
> >          # This is exactly what is done in [4]_
> > -        self.lambda_ = (lambda_  * self.R.shape[0] * r_rh[0] /
> > -                        (np.sqrt(self.B_reg.shape[0]) * np.sqrt(362.))
> > -                       )
> > +
> > +        lambda_ = (lambda_  * self.R.shape[0] * r_rh[0] /
> > +                   (np.sqrt(self.B_reg.shape[0]) * np.sqrt(362.)))
>
> Maybe I got this all wrong, but what I get from this is :
>
>    1. We need a sqrt(362) to accomodate the old hardcoded regularisation
>    sphere.
>
> The sphere was never hard-coded, we just didn't take the number of points
on the sphere into account when we computed lambda. We do now.


>
>    1. Instead of changing that, users can play with $lambda'$, which
>    kinda takes into account that hardcoded value.
>
> Users are free to play with lambda', it's an agrument to the model.

>
>    1. We could change sqrt(362) to actually be the number of point on the
>    regularisation sphere, also defaulting to 362 for the default code path,
>    but also taking into account any other sampled sphere. Granted, anyone
>    using another sphere will get a deffiretn value from now on, but that's
>    correcting a hardcoded value at the same time, which is more sane in the
>    long run.
>
> Thanks what we do, if you look at the code it's
`np.sqrt(self.B_reg.shape[0]) * np.sqrt(362.)`, $N_0$ is just a constant,
it's the first part that's important. We can set $N_0$ to anything. I think
you're suggesting setting it to 1., but I 1. is no better than 20. is no
better than sqrt(362), it's just a constant.

samuelstjean: Actually sqrt(len(self.sphere.vertices)) (or something like that), so the default code would still stay at sqrt(362), and other sphere would scale the value accordingly.

MrBago: On Tue, Jan 27, 2015 at 2:09 PM, Samuel St-Jean <notifications@github.com>
wrote:

> In dipy/reconst/csdeconv.py
> <https://github.com/nipy/dipy/pull/553#discussion_r23647669>:
>
> > @@ -118,18 +120,20 @@ def __init__(self, gtab, response, reg_sphere=None, sh_order=8, lambda_=1,
> >          # scale lambda_ to account for differences in the number of
> >          # SH coefficients and number of mapped directions
> >          # This is exactly what is done in [4]_
> > -        self.lambda_ = (lambda_  * self.R.shape[0] * r_rh[0] /
> > -                        (np.sqrt(self.B_reg.shape[0]) * np.sqrt(362.))
> > -                       )
> > +
> > +        lambda_ = (lambda_  * self.R.shape[0] * r_rh[0] /
> > +                   (np.sqrt(self.B_reg.shape[0]) * np.sqrt(362.)))
>
> Actually sqrt(len(self.sphere.vertices)) (or something like that), so the
> default code would still stay at sqrt(362), and other sphere would scale
> the value accordingly.
>
It's in the code,  `self.B_reg.shape[0] == len(self.sphere.vertices)`.

samuelstjean: Oh, so that would be in agreement with @arokem previous comment then. Which would also mean that the sqrt(362) is just a hardcoded value without any underlying meaning either.

MrBago: I don't see any benefit to setting $N_0=1$ at this stage, at least
$N_0=\sqrt{362.}$ has the benefit of backwards compatibility. If you
disagree with me, maybe we can discuss this in a separate issue.

Bago

samuelstjean: Not really, I was curious as to why this particular value.If it works ,why not after all.

arokem: I think that it's fine to leave this as is (if only for backward compatibility), but it would be good to add a comment here, for future curious souls (as suggested above by @matthew-brett), just to point out that this is not something that needs to be normalized away, and that users should consider how they set their lambda input, depending on their data/sampling sphere

mdesco:
Hi guys, Thanks @MrBago for this PR. Looks great. 

Here are my findings playing around with my compute_fodf scripts on full brains.
Full brain computation on Stanford data with old script (using peaks_from_model) vs with Cholesky
12 minutes | 2 minutes
So, 6x speed up on the Standford data. Nice!

On two other datasets, 2mm isotropic, 64 directions, 1.5T and 3T:
8 minutes (old) | 5 minutes (Cholesky)
So, ~1.5x speed up. Less important. @MrBago, do you have feeling as to why? convergence issue?

Now, fODFs produced do not have the exact same SH representations. Normal... So, I checked the implication on the NuFO (number of fiber orientations) and AFD max (max apparent fiber density).  Nifti files showing differences between old and Cholesky can be found here if you want to check:
https://www.dropbox.com/s/rrkwzxb9m7dgfrg/cholesky_vs_normal.zip?dl=0

My general observations are:
i) AFD_max and NuFO are stable in the white matter. Differences close to 0 or quite small. Especially so if a high-res sphere is used. Good.
ii) AFD_max and NuFO differences are important and quite big in partial volume voxels, cortex and ventricles. 
iii) Differences between AFD_max and NuFO reduce the more points are on the sphere.

So, I wonder if we have issues in partial volume voxels. @MrBago, have you played with convergence? number of iterations? tau parameters? lambda parameters?

MrBago:
There is a bug in the current master (at least I think it's a bug) that prevents the solver from returning in the first iteration. If I run the Cholesky implementation against [this](https://github.com/MrBago/dipy/compare/nipy:master...MrBago:debug_cholesky?expand=1) branch I get exactly the same fix.

Let me know if you want me to change the current implementation to be more like the current master.

MrBago:
Or we could also change the current implementation to return a low-order fit (order 4?) when the positivity constraint is not possible.

MrBago:
@mdesco I looked into the timings a little. I'm getting 15min and 1.5min for fitting the Standford data on a single core. My timings are just fitting, they don't include peak finding as is the case for peaks_from_model so that could be part of the difference. I'm surprised by how little improvement you see in the two other data sets. Is there any chance you could share them so I could take a look.

dipy/reconst/csdeconv.py
```diff
@@ -396,48 +483,52 @@ def csdeconv(dwsignal, sh_order, X, B_reg, lambda_=1., tau=0.1):
...
-    fodf = np.dot(B_reg, fodf_sh)
+    mu = 1e-5
+    if P is None:
+        P = np.dot(X.T, X)
+    z = np.dot(X.T, dwsignal)
+
+    try:
+        fodf_sh = _solve_cholesky(P, z)
+    except la.LinAlgError:
+        P = P + mu * np.eye(P.shape[0])
```
samuelstjean: This could be done inplace.

samuelstjean:
So, since gm/ventricules/partial voume seems to vary between the two implementations, could the cholesky one suffer from some numericla stability issues? 

Also, I'm curious about said stability on underdermined/crappy looking datasets, We have a (really) noisy 40 directions dataset at 1.2mm iso we could try to see if it holds reasonably well compared to the least square approach currently used.

dipy/reconst/csdeconv.py
MrBago: Well ... P is an argument to the function so doing this in place will change P for the caller.

We could handle this outside of the function all together, but than I think we would need to make P a required argument instead of an optional one.

MrBago:
@samuelstjean The difference between the two results is not due to the cholesky. As I said before, if we fix a bug that is in the current master the two methods produce the same result up to the 12th decimal place. The fix to current master is a 3 line fix and I've posted it [here](https://github.com/MrBago/dipy/compare/nipy:master...MrBago:debug_cholesky?expand=1).

samuelstjean:
So, if you find any value of the fod higher than the threshold for small values, this returns the order 4 estimation right? Not I sure I get the point actually... I thought this was used in case of non convergence of the fodf, but this seems like many voxels will have a min fod value higher than the threshold. Mostly inquiring, as my knowledge in this area is mostly reading what's written so far.

MrBago:
@samuelstjean you bring up good points, but sometimes I'm not sure you've carefully looked at what I posted.

    So, if you find any value of the fod higher than the threshold for small values,

No, if the min value of the FOD is less than the threshold that means *all* the FOD values are higher than the threshold. No positivity constraint can be applied because all FOD values are already positive.

    this returns the order 4 estimation right?

No, you should see that on line 407 the coefficients are fit the full `X` matrix so the full order least squares estimate is returned.

I think you're right though in that it's not clear how exactly to handle these voxels. The first question is should we use a low order estimate for the first iteration? This is what Tournier says that they do and what we currently do, but he also says that it should not matter.

Second what should we do if the low-order, first iteration FOD does not have any small values (no constrained dconv is possible). We could:
1) Return the low-order FOD
2) Return a full-order FOD using least squares
3) Use the full-order FOD and check to see if that has any negative values.

My new code does 2, the current master does 3. I think all 3 of these options have their merits and potential issues. I'd like to hear your thoughts @mdesco @Garyfallidis @samuelstjean.

Garyfallidis:
Hi @MrBago very sorry for delaying to give you feedback on this. I need to finish with some other pressing deadlines we will get back to this on Sunday or Monday. Keep it up!

matthew-brett:
Can you put in a test to make sure we don't get this bug again?  Maybe a regression test would be good here, as well?

samuelstjean:
Well, then line 406-408 could be moved just below line 400 since the least square fit is the same, and redone again. 

As for the rest, I clearly have not played enough with the code itself to fully understand what's happening (although I have now a better idea of the internal implementation, I think anyway...) so I'll let you guys discuss about improvements and whatnot.

Also, the old version always had really small noisy peaks on the fods on perfectly fine, noiseless synthetic data, while mrtrix version did not and was butter smooth, so I'll try to have a look as to if that corrected version still does that, we never investigated what was happening internally regarding this peculiar behavior.

MrBago:
Also, the old version always had really small noisy peaks on the fods on
> perfectly fine, noiseless synthetic data, while mrtrix version did not and
> was butter smooth, so I'll try to have a look as to if that corrected
> version still does that, we never investigated what was happening
> internally regarding this peculiar behavior.
>
I really would like to know where this difference between mrtrix and dipy
is coming from. My first guess would be the response function we and they
are using, but I would need to look at it more. Let me know if the
corrected version behaves any differently on the sythetic data. However I
would be surprised if it did, because we're only seeing differences between
the correcte and uncorrected versions in gm and csf, in the WM the two
versions are giving the same result.

Garyfallidis:
Hey @MrBago I run the tracking_quickstart tutorial and it gives me results with the peaks_from_model only in 1 minute!!! Amazing!

MrBago:
:) that's great. Any ideas about the best way to handle voxels where the
low-order FOD has no small values? After some thought I'm warming up to
option 3, but I could still be convinced of option 2 is the best way to go.

Bago

mdesco:
I would go for option 3.

Maxime Descoteaux, PhD
Professor
Centre de Recherche CHUS
Computer Science department
Sherbrooke University
2500 Boul. Université
Sherbrooke, Québec
J1K 2R1, Canada
phone: +1 819 821-8000 x66129
fax: +1 819 821-8200

Scientific director

Garyfallidis:
Hi @MrBago if we go with option 3 will we have an important difference  with mrtrix? And are we going to lose any performance? Any other possible disadvantages?

MrBago:
I don't think you'll see a noticeable difference in performance. I suspect those voxels wil converge quickly anyways. I don't know what mrtrix does, it's possible they do option 3 (or maybe something I haven't thought of). My latest commit implants option 3 and now this branch matches master to machine precision. (There are two voxels in the stanford data set which are effected by the `MKL ERROR: Parameter 4 was incorrect on entry to DLASCL` error which do not match, but everything else does.)

arokem:
This is the same problem described in #539, isn't it? Maybe we can finally
nail this one down, once and for all?

Is there anything unusual about these voxels? Can a `min_signal`-like
mechanism squash this kind of thing upfront?

MrBago:
@arokem after this PR there will be nothing to fix. The failure/success of `lstsq` should not depend on the signal, only the design matrix. The fact that it's failing for a few voxels is an MLK issue, but since this doesn't use `lstsq` and uses the cholesky decomposition instead means we don't need to deal with it.

arokem:
Just to verify that I understand - these two voxels fail on `master`, and
not on this branch. Is that correct?

MrBago:
Yep :)

matthew-brett:
So - we have the data from these voxels somewhere that we can add as a test?

arokem:
We don't need to. Once this PR is merged, the problem will be gone forever.

arokem:
See - I think that I'm catching on :-)

arokem:
Oh - I see what you mean. You want to be sure that it's gone forever. But reading @MrBago's last comments, it seems to me that it would be a bit weird, because this error is not even on the code path in any way shape or form.

Garyfallidis:
Hey if @matthew-brett and @MrBago thinks this is ready. I can go ahead and merge this in a couple of days. Sounds good?

matthew-brett:
Where did we get to on voxel data causing the previous error as a test?

samuelstjean:
Two small things again : 

1. What about the 3 methoda for returning voxels with no negative values, hence no deconv needed?
2. Has it been tested on something else than the stanford dataset? I see that the early tests were with order 6 and 8 on overdetermined systems, I'm curious to see if cholesky decomposition is still stable in the presence of high noise/low number of measurements, since numerically it is faster but indeed more prone to problems than other methods.

I have some datasets with 21 directions and 40 directions, highly noisy I can probably spare if you want.

MrBago:
@matthew-brett I'd like to punt that issue if possible :). I misspoke when I said this would resolve #539, this PR does not update SDT (which in principle could and should call the same csdeconv function as the CSD model, but currently does not). Once this PR is merged, I'd like to work on another that will 

1) have the two models call the same underlying fitting function.
2) Get rid of linalg fitting issues.
3) Write tests for 2.

Does that sound ok?

matthew-brett:
Bago - yes - sounds fine to me.

By the way - I'm confused - does someone have actual data available that causes this error?

samuelstjean:
Elef has some data sent by someone else for said bug, pretty sure I can throw something degenerate at the code also, I'm still in the middle of writing my dicom parser so maybe not today.

arokem:
Yes - see discussion above. Two voxels in the stanford HARDI data-set
seemed to have raised this error on Bago's set up, so it would be a good
idea to store their data somewhere, and put it in the tests, to prevent
future changes from raising this issue again. Especially because it's
fairly subtle to hunt down.

I think it's OK to merge this, and go on with the changes you suggest.

arokem:
What's the deal with that last commit? Is that instead of a rebase?

MrBago:
What commit :)? Sorry about that, I was working on some stuff, didn't mean to push that.

samuelstjean:
I just tested the two versions on an under-determined dataset (30 dirs vs 45 coefficients), and I did not get any convergence warning for the cholesky version, plus I get visually the same thing so far. Seems to work exactly the same, I'm impressed.

arokem:
Thanks for checking that @samuelstjean. Here we go!

samuelstjean:
Erm, what about @mdesco issues with partial volume? This seems to still be a valid, unresolved question.

arokem:
To quote @MrBago from above: "My latest commit implants option 3 and now this branch matches master to machine precision".

PR #557, opened 2015-01-15
NF: geodesic anisotropy

mdesco:
geodesic anisotropy (GA) is from Batchelor et al MRM 2005. It can viewed as the log-Euclidean version of FA.

### DIFF
--- a/dipy/reconst/dti.py
+++ b/dipy/reconst/dti.py
@@ -84,7 +84,7 @@ def fractional_anisotropy(evals, axis=-1):
 
 def geodesic_anisotropy(evals, axis=-1):
     r"""
-    Geodesic anisotropy (FA) of a diffusion tensor.
+    Geodesic anisotropy (GA) of a diffusion tensor.
 
     Parameters
     ----------
@@ -107,12 +107,38 @@ def geodesic_anisotropy(evals, axis=-1):
         GA = \sqrt{\sum_{i=1}^3 \log^2{\left ( \lambda_i/<\mathbf{D}> \right )}}, 
         \quad \textrm{where} \quad <\mathbf{D}> = (\lambda_1\lambda_2\lambda_3)^{1/3}
 
+    Note that the notation, <D>, is often used as the mean diffusivity (MD) of the diffusion tensor 
+    and can lead to confusions in the literature (see [1]_ versus [2]_ versus [3]_ for example).
+
+    Reference [2]_ defines geodesic anisotropy (GA) with <D> as the MD in the denominator of the sum. 
+    This is wrong. The original paper [1]_ defines GA with <D> = det(D)^(1/3), as the 
+    isotropic part of the distance. This might be an explanation for the confusion.
+    The isotropic part of the diffusion tensor in Euclideau space is
+    the MD whereas the isotropic part of the tensor in log-Euclidean space is det(D)^(1/3). 
+    The Appendix of [1]_ and log-Euclidean derivations from [3]_ are clear on this.
+
+    Hence, all that to say that <D> = det(D)^(1/3) here for the GA definition and not MD.
+
     References
     ----------
 
     .. [1] P. G. Batchelor, M. Moakher, D. Atkinson, F. Calamante, A. Connelly, 
         "A rigorous framework for diffusion tensor calculus", Magnetic Resonance 
         in Medicine, vol. 53, pp. 221-225, 2005.
+
+    .. [2] M. M. Correia, V. F. Newcombe, G.B. Williams.
+        "Contrast-to-noise ratios for indices of anisotropy obtained from diffusion MRI: 
+        a study with standard clinical b-values at 3T". NeuroImage, vol. 57, pp. 1103-1115, 2011.
+
+    .. [3] A. D. Lee, etal, P. M. Thompson.  
+        "Comparison of fractional and geodesic anisotropy in diffusion tensor images 
+        of 90 monozygotic and dizygotic twins". 5th IEEE International Symposium on 
+        Biomedical Imaging (ISBI), pp. 943-946, May 2008.
+
+    .. [4] V. Arsigny, P. Fillard, X. Pennec, N. Ayache. 
+        "Log-Euclidean metrics for fast and simple calculus on diffusion tensors."
+        Magnetic Resonance in Medecine, vol 56, pp. 411-421, 2006. 
+
     """
 
     evals = _roll_evals(evals, axis)
--- a/dipy/reconst/dti.py
+++ b/dipy/reconst/dti.py
@@ -113,7 +113,7 @@ def geodesic_anisotropy(evals, axis=-1):
     Reference [2]_ defines geodesic anisotropy (GA) with <D> as the MD in the denominator of the sum. 
     This is wrong. The original paper [1]_ defines GA with <D> = det(D)^(1/3), as the 
     isotropic part of the distance. This might be an explanation for the confusion.
-    The isotropic part of the diffusion tensor in Euclideau space is
+    The isotropic part of the diffusion tensor in Euclidean space is
     the MD whereas the isotropic part of the tensor in log-Euclidean space is det(D)^(1/3). 
     The Appendix of [1]_ and log-Euclidean derivations from [3]_ are clear on this.
 
--- a/dipy/reconst/dti.py
+++ b/dipy/reconst/dti.py
@@ -82,6 +82,61 @@ def fractional_anisotropy(evals, axis=-1):
 
     return fa
 
+def geodesic_anisotropy(evals, axis=-1):
+    r"""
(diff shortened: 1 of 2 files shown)


matthew-brett:
Thanks for this.

Would you mind cleaning up the git history?  Something like:

```
git fetch upstream # where upstream is remote for the nipy/dipy repo
git rebase -i upstream/master
```

You can then squash the WIP commit into a bigger commit, and get rid of the merge commit. 

Also - do you know the commit codes (`doc/devel/commit_codes.rst`) ?

mdesco:
Thanks @matthew-brett, rebased and added codes for the commit.

dipy/reconst/dti.py
```diff
@@ -82,6 +82,61 @@ def fractional_anisotropy(evals, axis=-1):
...
+        "A rigorous framework for diffusion tensor calculus", Magnetic Resonance 
+        in Medicine, vol. 53, pp. 221-225, 2005.
+    """
+
+    evals = _roll_evals(evals, axis)
+
+    all_zero = (evals == 0).all(axis=0)
+    ev1, ev2, ev3 = evals
+
+    md = (ev1 + ev2 + ev3)/3
```
arokem: PEP8 (space around operator)

dipy/reconst/dti.py
```diff
@@ -82,6 +82,61 @@ def fractional_anisotropy(evals, axis=-1):
...
+    
+    log1 = np.zeros(md.shape)
+    log2 = np.zeros(md.shape)
+    log3 = np.zeros(md.shape)
+
+    # to make sure we do not divide by 0
+    idx = np.nonzero(md)
+    log1[idx] = np.log(ev1[idx]/md[idx])
```
arokem: Same in these lines

dipy/reconst/dti.py
```diff
@@ -82,6 +82,61 @@ def fractional_anisotropy(evals, axis=-1):
...
+    log2[idx] = np.log(ev2[idx]/md[idx])
+    log3[idx] = np.log(ev3[idx]/md[idx])
+    
+    ga = np.sqrt(log1**2 + log2**2 + log3**2)
```
arokem: And here

dipy/reconst/tests/test_dti.py
```diff
@@ -185,6 +185,11 @@ def test_fa_of_zero():
     fa = fractional_anisotropy(evals)
     assert_array_equal(fa, 0)
 
+def test_ga_of_zero():
+    evals = np.zeros((4, 3))
```
arokem: Just to be sure I got it: these are 4 voxels that each have the same evals of [0, 0, 0]?

dipy/reconst/dti.py
```diff
@@ -82,6 +82,61 @@ def fractional_anisotropy(evals, axis=-1):
...
+    ----------
+
+    .. [1] P. G. Batchelor, M. Moakher, D. Atkinson, F. Calamante, A. Connelly, 
+        "A rigorous framework for diffusion tensor calculus", Magnetic Resonance 
+        in Medicine, vol. 53, pp. 221-225, 2005.
+    """
+
+    evals = _roll_evals(evals, axis)
+
+    all_zero = (evals == 0).all(axis=0)
```
arokem: I don't think you use the variable `all_zero` for anything.

dipy/reconst/dti.py
```diff
@@ -82,6 +82,61 @@ def fractional_anisotropy(evals, axis=-1):
...
+    ga : array
+        Calculated GA. In the range [0, +infty)
+
+    Notes
+    --------
+    GA is calculated using the following equation [1]_ and [2]_:
+
+    .. math::
+
+        GA = \sqrt{\sum_{i=1}^3 \log^2{\left ( \lambda_i/<\mathbf{D}> \right )},
```
arokem: I think there's a curly brace missing here. I believe this should be: 

A = \sqrt{\sum_{i=1}^3 \log^2{\left ( \lambda_i/<\mathbf{D}> \right )}},\quad \textrm{where} \quad <\mathbf{D}> = \frac{\lambda_1+\lambda_2+\lambda_3}{3}

dipy/reconst/dti.py
arokem: Is MD always equal to det(D)^(1/3)?

mdesco: No. md = (l1+l2+l3)/3 whereas detD^(1/3) = (l1*l2*l3)^(1/3). I have realized that there are conflicting definitions of GA in papers. See [1] vs [2]. And other groups compute matrix_log of the whole tensor to generalize FA to log-Euclidean space.

[1] P. G. Batchelor, M. Moakher, D. Atkinson, F. Calamante, A. Connelly, 
        "A rigorous framework for diffusion tensor calculus", Magnetic Resonance 
        in Medicine, vol. 53, pp. 221-225, 2005.

[2] M. Morgado Correia, V.F.J. Newcombe, Guy B. Williams. "Contrast-to-noise 
       ratios for indices of anisotropy obtained from diffusion MRI: A study with 
       standard clinical b-values at 3 T", NeuroImage, vol. 57, pp. 1103-1115, 2011. 
 
I will keep looking and asking around... I have a feeling [1] should be trusted... 
Anyone knows in the Dipy community?

dipy/reconst/tests/test_dti.py
mdesco: Yes. Same test as for the FA.

dipy/reconst/dti.py
mdesco: Yes. thanks. All other comments have been done.

mdesco:
After double-checking of the literature et re-deriving the equations, the equation given in Batchelor et al MRM 2005 is the good one. The denominator should be the determinant(D)^(1/3) not the mean diffusivity.

dipy/reconst/dti.py
```diff
@@ -82,6 +82,56 @@ def fractional_anisotropy(evals, axis=-1):
 
     return fa
 
+def geodesic_anisotropy(evals, axis=-1):
+    r"""
+    Geodesic anisotropy (FA) of a diffusion tensor.
```
samuelstjean: This should be (GA)

matthew-brett:
Maxime - thanks very much for checking that.   Is there a good place you can record the your researches so that the next person coming to this realizes that you have nailed down the definitive answer?

mdesco:
Matthew, what do you suggest? I can latex or write a little something on my findings.

matthew-brett:
I don't think you need much - maybe a paragraph or two in the Notes section of the docstring?   At least then it's there is someone is looking for it.   I don't know whether you think this is something others not using dipy would run into?  If that's the case, would it be worth having a small doc in the doc/theory section?

mdesco:
Ok. I have added a couple references and a couple paragraphs describing the issue and the solution.

arokem:
@matthew-brett - is this ready to go?

matthew-brett:
Yes, ready to go if y'all do not think more markup for <D> is useful for the docstring.

arokem:
Apparently not? @mdesco - do you prefer to have this format, or do you want to add the latex markup into the docstring description of the equation as well? 

Personally, I find that there's generally a trade-off where more markup would make the html/pdf docs more readable, but the raw docstring less (slightly) less readable.

matthew-brett:
I don't have a strong view either way.  Maybe double backticks would be
better than angle brackets for the markup here, but any decision is fine.
​

mdesco:
I let you guys decide. I do not have much experience with this. No preference. @arokem, make the call.

arokem:
OK - for consistency, let's go with the full math-dollar markup. 

So - this part of the docstring would look like this: 

```

Note that the notation, $<D>$, is often used as the mean diffusivity (MD) of the diffusion tensor 
    and can lead to confusions in the literature (see [1]_ versus [2]_ versus [3]_ for example).
    Reference [2]_ defines geodesic anisotropy (GA) with $<D>$ as the MD in the denominator of the sum. 
    This is wrong. The original paper [1]_ defines GA with $<D> = det(D)^{1/3}$, as the 
    isotropic part of the distance. This might be an explanation for the confusion.
    The isotropic part of the diffusion tensor in Euclidean space is
    the MD whereas the isotropic part of the tensor in log-Euclidean space is $det(D)^{1/3}$. 
    The Appendix of [1]_ and log-Euclidean derivations from [3]_ are clear on this.
    Hence, all that to say that $<D> = det(D)^{1/3}$ here for the GA definition and not MD.

```

arokem:
Hey @mdesco - just a gentle reminder to make this change to the docstring, so that I can merge this

mdesco:
Sorry about the delay. Done.

arokem:
Great - thanks for that addition! Seems like we have the largest collection of tensor-based metrics in the known (software) universe!

PR #559, opened 2015-01-16
SFM for multi b-value data

arokem:


dipy/reconst/sfm.py
```diff
@@ -33,7 +33,64 @@
...
+    def __init__(self, gtab):
+        ReconstModel.__init__(self, gtab)
+
+    def fit(self, data):
+        data_no_b0 = data[..., ~self.gtab.b0s_mask]
+        params = np.mean(data_no_b0, -1)
+        if len(data_no_b0.shape) == 1:
+            n_vox = 1
+        else:
+            n_vox = data_no_b0.shape[0]
```
stefanv: Perhaps slightly clearer to first call

```
data_no_b0 = np.array(data_no_b0, ndmin=2, copy=False)
```

in the beginning of the function, and then this if/else falls away.

dipy/reconst/sfm.py
```diff
@@ -33,7 +33,64 @@
...
+        p = np.polyfit(self.gtab.bvals, log_data.T, 1)
+        params = -p[0]
+        return ExponentialIsotropicFit(self, params, n_vox)
+
+
+class ExponentialIsotropicFit(IsotropicFit):
+    def predict(self, gtab=None):
+        if gtab is None:
+            gtab = self.model.gtab
+        return np.exp(-gtab.bvals[~gtab.b0s_mask] * (np.zeros((self.n_vox,np.sum(~gtab.b0s_mask))) + self.params[..., np.newaxis]))
```
stefanv: Break line

dipy/reconst/tests/test_sfm.py
```diff
@@ -109,3 +109,29 @@ def fit(self, X, y):
...
+
+
+def test_exponential_iso():    
+    fdata, fbvals, fbvecs = dpd.get_data()
+    data_dti = nib.load(fdata).get_data()
+    gtab_dti = grad.gradient_table(fbvals, fbvecs)
+    data_dsi, gtab_dsi = dpd.dsi_voxels()
+    for data, gtab in zip([data_dsi, data_dti], [gtab_dsi, gtab_dti]):
+        sfmodel = sfm.SparseFascicleModel(gtab,
+                                        isotropic=sfm.ExponentialIsotropicModel)
```
stefanv: indentation

arokem: This one's a bit tough: If I properly indent, the line goes to more than 80 characters. I preferred to go this way, because it seems to be less than an eye-sore than having the line wrap down. I could also break it: 

    istropic=
    sfm.ExponentialIsotropicModel

What do you think?

stefanv: I see what you mean!  Here's what is suggested in PEP8.

```
# Hanging indents should add a level.
foo = long_function_name(
    var_one, var_two,
    var_three, var_four)
```

stefanv:
Looks good to me!

dipy/reconst/sfm.py
arokem: This sounds like a good idea, but I don't entirely understand it. Do you mean: 
```
data_no_b0 = data[..., ~self.gtab.b0s_mask]
data_no_b0 = np.array(data_no_b0, ndmin=2, copy=False)
n_vox = data_no_b0.shape
```
I don't think that would work

stefanv: I was thinking:

```
data_no_b0 = data[..., ~self.gtab.b0s_mask]
data_no_b0 = np.array(data_no_b0, ndmin=2, copy=False)
...
n_vox = len(data_no_b0)
```

arokem:
I've removed the WIP here. This is ready for a review, from my point of view. @stefanv - would you mind taking another look?

dipy/reconst/sfm.py
```diff
@@ -33,7 +33,76 @@
     warnings.warn(w)
 
 
-def sfm_design_matrix(gtab, sphere, response, mode='signal'):
+# Isotropic signal models: these are models of the part of the signal that
+# changes with b-value, but does not change with direction. This collection is
+# extensible, by inheriting from IsotropicModel/IsotropicFit below:
+class IsotropicModel(ReconstModel):
+    """
+    A base-class for the representation of isotropic signals.
```
stefanv: Not sure, but think the numpy documentation parser may expect a newline after the first sentence.

dipy/reconst/sfm.py
```diff
@@ -33,7 +33,76 @@
...
+    The default behavior, suitable for single b-value data is to calculate the
+    mean in each voxel as an estimate of the signal that does not depend on
+    direction.
+    """
+    def __init__(self, gtab):
+        ReconstModel.__init__(self, gtab)
+
+    def fit(self, data):
+        data_no_b0 = data[..., ~self.gtab.b0s_mask]
```
stefanv: Can it happen that the `b0s_mask` is a scalar?  It looks that way from the ``if len(...)== 1`` below.  In that case, I would consider standardizing the mask first:

```
mask = np.array(self.gtab.b0s_mask, copy=False, ndmin=2)
```

(Hrm, numpy, we really need a shortcut for that)
```

Then, you can proceed without looking back and without special exceptions.

File: dipy/reconst/sfm.py [hunk 6]
```diff
@@ -33,7 +33,76 @@
...
+
+    def predict(self, gtab=None):
+        if gtab is None:
+            gtab = self.model.gtab
+        return self.params[..., np.newaxis] + np.zeros((self.n_vox,
+                                                        np.sum(~gtab.b0s_mask)))
+
+class ExponentialIsotropicModel(IsotropicModel):
+    def fit(self, data):
+        if len(data.shape) == 1:
```
stefanv: Again, my preference would be: standardize the data and then write a single algorithm without exceptions

dipy/reconst/sfm.py
```diff
@@ -33,7 +33,76 @@
...
+            n_vox = 1
+        else:
+            n_vox = data.shape[0]
+
+        data_no_b0 = data[..., ~self.gtab.b0s_mask]
+        if np.sum(self.gtab.b0s_mask) > 0:
```
stefanv: But, here we have no choice :)

dipy/reconst/sfm.py
```diff
@@ -225,36 +316,36 @@ def fit(self, data, mask=None):
...
         flat_S = flat_data[..., ~self.gtab.b0s_mask] / flat_S0[..., None]
-        flat_mean = np.mean(flat_S, -1)
+        flat_isotropic = self.isotropic(self.gtab).fit(flat_data).predict()
         flat_params = np.zeros((flat_data.shape[0],
                                 self.design_matrix.shape[-1]))
 
         for vox, vox_data in enumerate(flat_S):
             if np.any(np.isnan(vox_data)):
-                flat_params[vox] = (np.zeros(self.design_matrix.shape[-1]))
+                pass
```
stefanv: ?

dipy/reconst/sfm.py
```diff
@@ -265,13 +356,13 @@ def __init__(self, model, beta, S0, mean_signal):
             The parameters of fit to data.
         S0 : ndarray
             The mean non-diffusion-weighted signal.
-        mean_signal : ndarray
-            The mean of the diffusion-weighted signal
+        iso : ndarray
```
stefanv: This was here before, so not expecting you to fix it, but it always helps if there is array shape information available in the documentation.

dipy/reconst/sfm.py
```diff
@@ -337,9 +428,9 @@ def predict(self, gtab=None, response=None, S0=None):
             S0 = self.S0
         if isinstance(S0, np.ndarray):
             S0 = S0[..., None]
-        if isinstance(self.mean_signal, np.ndarray):
-            mean_signal = self.mean_signal[..., None]
-        pre_pred_sig = S0 * (pred_weighted + mean_signal)
+        if isinstance(self.iso, np.ndarray):
```
stefanv: What is this meant to do?  Guarantee 2D-ness?  Then rather use an explicit array constructor with `copy=False`.

stefanv:
I like it!  Some minor nitpicks--let me know what you think, also happy to merge without them.

arokem:
Thanks for taking another look! I'll give this another go-through and ping
you in a few days.

dipy/reconst/sfm.py
arokem: Added a line!

dipy/reconst/sfm.py
arokem: `b0s_mask` is never a scalar, always an array. The len statement is here to deal with the situation where we have only one voxel, but let me think how I can standardize things here.

dipy/reconst/sfm.py
arokem: Turns out this should be an `IsotropicFit` object! Now refactored in several places

dipy/reconst/sfm.py
arokem: I've tried to simplify and clarify by adding a comment and a test (https://github.com/arokem/dipy/commit/ac011dfdfbdc5fb197a9131823d7394ed390848e). This is for cases in which the non diffusion-weighted signal is 0, like in voxels outside the brain. In these cases, `ElasticNet` would throw an error, so we pre-empt that by just leaving the parameters as all-zeros.

arokem:
Still thinking about standardization for the isotropic model input fits, so please don't merge yet. But I have to go to a talk now, so it will probably be later this week, before I get back to this.

arokem:
OK - @stefanv  - the last commit should handle your comments about standardization of the data shape in the isotropic models. I also added one more commit to allow users to pass in an already initialized sklearn solver. Sorry that it's taken me this long to address your comments. Hopefully this is all good to go now?

dipy/reconst/sfm.py
```diff
@@ -147,20 +286,27 @@ def __init__(self, gtab, sphere=None, response=[0.0015, 0.0005, 0.0005],
             The eigenvalues of a canonical tensor to be used as the response
             function of single-fascicle signals.
             Default:[0.0015, 0.0005, 0.0005]
-
-        solver : string or SKLearnLinearSolver object, optional
+        solver : string, dipy.core.optimize.SKLearnLinearSolver object, or
+            sklearn.linear_model.base.LinearModel object, optional.
```
stefanv: Does this render correctly?

stefanv:
Beautifully clean!

dipy/reconst/sfm.py
arokem: Good catch.

arokem: This works.

stefanv:
Thanks, @arokem!  Nicely polished PR

PR #561, opened 2015-01-17
Added option to return the number of voxels fitting the fa threshold

samuelstjean:
As per the title, I added an option to return the number of voxels used in the response function estimation. That way, one can increase the roi or lower the threshold if only a small number of voxel is found.

A warning is also raised if no voxel fitting the threshold is found, which is inherently lower than 0.7 in small animal data, highly noisy data, badly behaved synthetic data, etc.

### DIFF
--- a/dipy/reconst/csdeconv.py
+++ b/dipy/reconst/csdeconv.py
@@ -614,7 +614,7 @@ def odf_sh_to_sharp(odfs_sh, sphere, basis=None, ratio=3 / 15., sh_order=8,
 
 def auto_response(gtab, data, roi_center=None, roi_radius=10, fa_thr=0.7,
                   return_number_of_voxels=False):
-    """ Automatic estimation of response function using FA
+    """ Automatic estimation of response function using FA.
 
     Parameters
     ----------
@@ -638,7 +638,7 @@ def auto_response(gtab, data, roi_center=None, roi_radius=10, fa_thr=0.7,
         (`evals`, `S0`)
     ratio : float
         The ratio between smallest versus largest eigenvalue of the response.
-    number of voxels : int
+    number of voxels : int (optional)
         The number of voxels used for estimating the response function.
 
     Notes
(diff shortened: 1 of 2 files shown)


dipy/reconst/csdeconv.py
```diff
@@ -628,13 +628,18 @@ def auto_response(gtab, data, roi_center=None, roi_radius=10, fa_thr=0.7):
...
+        function.
 
     Returns
     -------
     response : tuple, (2,)
         (`evals`, `S0`)
     ratio : float
-        the ratio between smallest versus largest eigenvalue of the response
+        The ratio between smallest versus largest eigenvalue of the response.
+    number of voxels : int
```
arokem: Do we need to mark this as an optional return value?

dipy/reconst/csdeconv.py
```diff
@@ -654,7 +659,11 @@ def auto_response(gtab, data, roi_center=None, roi_radius=10, fa_thr=0.7):
     the highest and second highest eigenvalues in the ROI with FA higher than
     threshold. We also include the average S0s.
 
-    Finally, we also return the `ratio` which is used for the SDT models.
+    We also return the `ratio` which is used for the SDT models. If requested,
+    the number of voxels used for estimating the response function is also
+    returned, which can be used to judge the fidelity of the response function.
+    As a rule of thumb, at least 300 voxels should be used to estimate a good
```
arokem: Where does this number (300) come from?

samuelstjean: Rule of thumb, actually defined in the appendix for the SDT paper.

http://scil.dinf.usherbrooke.ca/wp-content/papers/descoteaux_tmi_2009.pdf  page 17.

arokem: Please add that reference to the PR, so that we can find it in the future.

mdesco: Actually, comes from the original paper by Tournier et al NeuroImage 2004

samuelstjean: Funny, I always though it was really just a rule of thumb, I'll add the proper ref tomorrow.

mdesco: It is a rule of thumb but it came from Tournier.

Maxime Descoteaux, PhD
Professor
Centre de Recherche CHUS
Computer Science department
Sherbrooke University
2500 Boul. Université
Sherbrooke, Québec
J1K 2R1, Canada
phone: +1 819 821-8000 x66129
fax: +1 819 821-8200

Scientific director

arokem:
Thanks!

PR #571, opened 2015-01-28
Bf cart2sphere

MrBago:
I don't think `cart2sphere(0, 0, 0)` should return any nans. Here is a fix.

### DIFF
--- a/dipy/core/geometry.py
+++ b/dipy/core/geometry.py
@@ -126,6 +126,7 @@ def cart2sphere(x, y, z):
     '''
     r = np.sqrt(x * x + y * y + z * z)
     theta = np.arccos(z / r)
+    theta = np.where(r > 0, theta, 0.)
     phi = np.arctan2(y, x)
     r, theta, phi = np.broadcast_arrays(r, theta, phi)
     return r, theta, phi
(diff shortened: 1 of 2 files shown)


arokem:
Hmm. This is a matter of 'philosophical' point of view. I agree that `r` of the origin is 0, but why would the angles be (0, 0), rather than any other angles?

MrBago:
@arokem you're right, but let me describe my use case. I want to compute some function which is more easily expressed in spherical coordinates, take for example `r**2 * cos(theta) * sin(3 * phi)`. If `cart2sphere` returns `nan` for `theta` than I have to special case `r == 0` (that's what i currently do).

When I saw this, it occurred to me that `sphere2cart(cart2sphere(0, 0, 0)) == [0, 0, 0]` should work and currently it doesn't. I don't feel strongly about this, just thought I would bring it up in case you guys think it's an improvement.

arokem:
For whatever it's worth, matlab agrees with you. I tend to think that you are right and this would be an improvement. At least we would get one of the three right, where previously we had no answer for any of them. If no other objections are raised, I'll merge this in a few days. Thanks!

Issue #441, opened 2014-10-08
Registration - Put the scripts/data to run Rohlfing experiment on some standard place


omarocegueda:
Rohlfing's experiment consists on registering 306 pairs of manually annotated MRI (IBSR) and measure the quality of the registration in terms of the average Jaccard index of each anatomical region. Right now, I am running this test on a cluster in my institution when we need to. We need to put the scripts and the data to easily run the experiment on some standard place. Each registration takes about 15 minutes.

matthew-brett:
I think we should write to the INCF to see if they can provide a service like this.  I think it would be useful for anyone evaluating a registration method.

@jbpoline, @chrisfilo - what do you think?  How should we go about this?

arokem:
Sorry for only picking up on this now. We can host data (and also scripts), with semantic versioning, on the Stanford SDR, if that seems like a good solution. People would still have to download the data/scripts to run them on some local resource, though.

matthew-brett:
Right - I was thinking that we would need some system which would run something like - say 50 jobs at a time? Omar - is your estimate then 15 minutes * 306 for total CPU minutes?  What CPUs are you running on?  What kind of memory do they need?

chrisfilo:
Let me try to understand the problem better: 
a) you need to host the manually annotated data, example scripts to show how to validate an algorithm using the data?
b) you want to build a Registration Evaluation as a Service platform which will allow people to submit their coregistration algorithms (packaged in some way) and get an evaluation of their performance.

I can help with a). b) will be a substantial effort in terms of software development if built ground up. However, it could become a part of a bigger Science as a Service platform.

matthew-brett:
Omar knows this much better than me, but I believe our immediate need is to be able to run scripts, say hosted on github, that will execute a multi-CPU job on the a given standard set of data.

So yes, we would need:

1) somewhere to put the original data where the cluster can see it (and it would be better if this was also public;
2) CPUs on which to run the scripts, perhaps every week or so;
3) short term hosting of the output registrations and registration measures, which we could then download somewhere.

Being able to do this with more than one possible configuration of packages would be a bonus (for example, installing dipy / nipy wheels from some given location as part of the job initiation). 

I can imagine that others will want to do something similar with their own packages, and that it might be useful to extend the configuration in due course to something more general, like a .travis.yml setup.

Omar - is that right on the basic requirements?

omarocegueda:
Hello everyone!
[...]

PR #574, opened 2015-02-05
Affine registration PR1: Transforms.

omarocegueda:
Based on some performance comparisons between ANTS2, Flirt and Nipy for affine registration with mutual information, we have observed that ANTS2 consistently performs better in terms of registration accuracy. The following graph depicts the Jaccard index of 31 manually annotated anatomical regions averaged over 306 registrations (1=perfect overlap, 0=no overlap):
http://www.cimat.mx/~omar/imgshare/mi32_flirt_ants2_nipy.png

This motivated us to implement the same registration strategy in Cython, so we can easily use it with our non-linear registration module and at some point modify the algorithms/metrics more easily for DWMRI. This is the current performance of this Cython implementation:
http://www.cimat.mx/~omar/imgshare/ANTS_vs_Dipy.png
I am investigating the cause of the small differences but at some point we should reach the same performance.
 
In an effort to make the review process (a bit) easier, this part of the code only contains the computation of the affine transforms and their Jacobians.


dipy/align/transforms.pyx
```diff
@@ -0,0 +1,810 @@
...
+    J[0, 3] = 1
+    J[1, 4:7] = x[:3]
+    J[1, 7] = 1
+    J[2, 8:11] = x[:3]
+    J[2, 11] = 1
+    # This Jacobian depends on x (it's not constant): return 0
+    return 0
+
+
+cdef void _dot_prod(double[:,:] A, double[:,:] B, double[:,:] C):
```
stefanv: Not tested or used?

stefanv:
@omarocegueda I like your structured approach to measuring these results.  Can we include some form of the performance (in terms of accuracy) tests in the code base?

omarocegueda:
Thank you @stefanv!,
sure, it is actually issue #441, the problem is that the full evaluation may take a long time (especially for non-linear registration). The way I do it right now is to create one job for each registration and send the (306) jobs to a cluster. Maybe we can start by adding only some evaluation tools (dipy.align.eval?) to the align module (code to compute the Jaccard indices, and creating the graphs), and assume that the users somehow ran all the registrations they need to evaluate.

dipy/align/tests/test_transforms.py
```diff
@@ -0,0 +1,233 @@
+from dipy.align.transforms import *
```
matthew-brett: Please import specific names (best) or import the module (`import dipy.align.transforms as dit`) and use with ``dit.`` namespace.  The `*` makes it difficult to spot bad imports, among other things.

dipy/align/tests/test_transforms.py
```diff
@@ -0,0 +1,233 @@
...
+            get_identity_parameters(t, dim, theta)
+
+            expected = np.eye(dim + 1)
+            actual = np.empty_like(expected)
+            param_to_matrix(t, dim, theta, actual)
+            assert_array_almost_equal(actual, expected)
+
+
+def test_eval_jacobian_function():
+    r""" Compare the analytical Jacobians with their numerical approximations
```
matthew-brett: Comments better than docstrings at the top, because the docstring gets printed out during nostests, and this make it harder to see which function is being tested, and doesn't match the output for functions without docstrings.

dipy/align/transforms.pxd
```diff
@@ -0,0 +1,43 @@
...
+information is used by the optimizer to avoid making unnecessary
+function calls
+
+Note: we need jacobian functions to be declared as nogil because they may
+be called once for each sample static/moving voxel pair. This way we
+can make this call inside a nogil loop
+"""
+
+ctypedef void (*param_to_matrix_function)(double[:], double[:,:]) nogil
+r""" Type of a function that computes the matrix associated to an
```
matthew-brett: associated with

dipy/align/transforms.pxd
```diff
@@ -0,0 +1,43 @@
...
+affine transform in canonical coordinates.
+
+Note: this function should be called only O(1) times per iteration of the
+optimization method, as opposed to once per static/moving voxel pair
```
matthew-brett: just for ease of reading "as opposed ot the jacobian_functions, which are called once per static/moving voxel pair."

dipy/align/transforms.pyx
```diff
@@ -0,0 +1,810 @@
...
+
+#This dictionary allows us to get the appropriate transform index from a string
+transform_type = {'TRANSLATION':TRANSLATION,
+                  'ROTATION':ROTATION,
+                  'RIGID':RIGID,
+                  'SCALING':SCALING,
+                  'AFFINE':AFFINE}
+
+
+def number_of_parameters(int ttype, int dim):
```
matthew-brett: Why not cpdef and combine with function below?

dipy/align/transforms.pyx
```diff
@@ -0,0 +1,810 @@
...
+        the point at which to evaluate the Jacobian
+    J : array, shape (dim, n)
+        the destination matrix for the Jacobian
+    """
+    with nogil:
+        get_jacobian_function(ttype, dim)(theta, x, J)
+
+
+def param_to_matrix(int ttype, int dim, double[:] theta, double[:,:] T):
+    r""" Compute the matrix associated to the given transform and parameters
```
matthew-brett: associated with

dipy/align/transforms.pyx
```diff
@@ -0,0 +1,810 @@
...
+    dim : int
+        the domain dimension of the transformation (either 2 or 3)
+    theta : array, shape (n,)
+        the parameters of the transformation at which to evaluate the Jacobian
+    x : array, shape (dim,)
+        the point at which to evaluate the Jacobian
+    J : array, shape (dim, n)
+        the destination matrix for the Jacobian
+    """
+    with nogil:
```
matthew-brett: already a nogil function below?

dipy/align/transforms.pyx
```diff
@@ -0,0 +1,810 @@
...
+        the type of the transformation (use transform_type dictionary to map
+        transformation name to the associated int )
+    dim : int
+        the domain dimension of the transformation (either 2 or 3)
+    theta : array, shape (n,)
+        the transformation parameters
+    T : array, shape (dim + 1, dim + 1)
+        the buffer to write the transform matrix
+    """
+    with nogil:
```
matthew-brett: ditto - isn't this already defined as nogil?

dipy/align/transforms.pyx
```diff
@@ -0,0 +1,810 @@
...
+        get_param_to_matrix_function(ttype, dim)(theta, T)
+
+
+def get_identity_parameters(int ttype, int dim, double[:] theta):
```
matthew-brett: Ditto cpdef and merge with below?

dipy/align/transforms.pyx
```diff
@@ -0,0 +1,810 @@
...
+            return _translation_jacobian_3d
+        elif ttype == ROTATION:
+            return _rotation_jacobian_3d
+        elif ttype == RIGID:
+            return _rigid_jacobian_3d
+        elif ttype == SCALING:
+            return _scaling_jacobian_3d
+        elif ttype == AFFINE:
+            return _affine_jacobian_3d
+    return NULL
```
matthew-brett: Not error?  Or is this raised elsewhere?

dipy/align/transforms.pyx
```diff
@@ -0,0 +1,810 @@
...
+            theta[:3] = 0
+        elif ttype == RIGID:
+            theta[:6] = 0
+        elif ttype == SCALING:
+            theta[0] = 1
+        elif ttype == AFFINE:
+            theta[0], theta[1], theta[2], theta[3] = 1, 0, 0, 0
+            theta[4], theta[5], theta[6], theta[7] = 0, 1, 0, 0
+            theta[8], theta[9], theta[10], theta[11] = 0, 0, 1, 0
+
```
matthew-brett: Do you need to check for bad inputs?

dipy/align/transforms.pyx
```diff
@@ -0,0 +1,810 @@
...
+        the point at which to compute the Jacobian (the Jacobian does not
+        depend on x, but we receive the buffer so all Jacobian functions
+        receive the same parameters)
+    J : array, shape(2, 2)
+        the buffer in which to write the Jacobian
+    """
+    J[0,0], J[0, 1] = 1.0, 0.0
+    J[1,0], J[1, 1] = 0.0, 1.0
+    # This Jacobian does not depend on x (it's constant): return 1
+    return 1
```
matthew-brett: Document return value in docstring?

dipy/align/transforms.pyx
```diff
@@ -0,0 +1,810 @@
...
+        depend on x, but we receive the buffer so all Jacobian functions
+        receive the same parameters)
+    J : array, shape(3, 3)
+        the buffer in which to write the Jacobian
+    """
+    J[0,0], J[0,1], J[0,2] = 1.0, 0.0, 0.0
+    J[1,0], J[1,1], J[1,2] = 0.0, 1.0, 0.0
+    J[2,0], J[2,1], J[2,2] = 0.0, 0.0, 1.0
+    # This Jacobian does not depend on x (it's constant): return 1
+    return 1
```
matthew-brett: Document return value?

dipy/align/transforms.pyx
```diff
@@ -0,0 +1,810 @@
...
+    '''
+    cdef:
+        double st = sin(theta[0])
+        double ct = cos(theta[0])
+        double px = x[0], py = x[1]
+
+    J[0, 0] = -px * st - py * ct
+    J[1, 0] = px * ct - py * st
+    # This Jacobian depends on x (it's not constant): return 0
+    return 0
```
matthew-brett: ditto document return value

dipy/align/transforms.pyx
```diff
@@ -0,0 +1,810 @@
...
+    J[1, 1] = ( -sc * sb + cc * sa * cb ) * px + ( sc * cb + cc * sa * sb ) * pz
+    J[2, 1] = ( -ca * cb ) * px + ( -ca * sb ) * pz
+
+    J[0, 2] = ( -sc * cb - cc * sa * sb ) * px + ( -cc * ca ) * py + \
+              ( -sc * sb + cc * sa * cb ) * pz
+    J[1, 2] = ( cc * cb - sc * sa * sb ) * px + ( -sc * ca ) * py + \
+              ( cc * sb + sc * sa * cb ) * pz
+    J[2, 2] = 0
+    # This Jacobian depends on x (it's not constant): return 0
+    return 0
```
matthew-brett: document

dipy/align/transforms.pyx
```diff
@@ -0,0 +1,810 @@
...
+    '''
+    cdef:
+        double st = sin(theta[0])
+        double ct = cos(theta[0])
+        double px = x[0], py = x[1]
+
+    J[0, 0], J[0, 1], J[0, 2] = -px * st - py * ct, 1, 0
+    J[1, 0], J[1, 1], J[1, 2] = px * ct - py * st, 0, 1
+    # This Jacobian depends on x (it's not constant): return 0
+    return 0
```
matthew-brett: document

dipy/align/transforms.pyx
```diff
@@ -0,0 +1,810 @@
...
+    J[1, 2] = ( cc * cb - sc * sa * sb ) * px + ( -sc * ca ) * py + \
+              ( cc * sb + sc * sa * cb ) * pz
+    J[2, 2] = 0
+
+    J[0,3:6] = 0
+    J[1,3:6] = 0
+    J[2,3:6] = 0
+    J[0,3], J[1,4], J[2,5] = 1, 1, 1
+    # This Jacobian depends on x (it's not constant): return 0
+    return 0
```
matthew-brett: document

dipy/align/transforms.pyx
```diff
@@ -0,0 +1,810 @@
...
+        but we receive the buffer to make it consistent with other Jacobian
+        functions)
+    x : array, shape (2,)
+        the point at which to compute the Jacobian
+    J : array, shape(2, 1)
+        the buffer in which to write the Jacobian
+    """
+    J[0,0], J[1,0] = x[0], x[1]
+    # This Jacobian depends on x (it's not constant): return 0
+    return 0
```
matthew-brett: ditto

dipy/align/transforms.pyx
```diff
@@ -0,0 +1,810 @@
...
+        but we receive the buffer to make it consistent with other Jacobian
+        functions)
+    x : array, shape (3,)
+        the point at which to compute the Jacobian
+    J : array, shape(3, 1)
+        the buffer in which to write the Jacobian
+    """
+    J[0,0], J[1,0], J[2,0]= x[0], x[1], x[2]
+    # This Jacobian depends on x (it's not constant): return 0
+    return 0
```
matthew-brett: ditto

dipy/align/transforms.pyx
```diff
@@ -0,0 +1,810 @@
...
+    """
+    J[0,:6] = 0
+    J[1,:6] = 0
+
+    J[0, :2] = x[:2]
+    J[0, 2] = 1
+    J[1, 3:5] = x[:2]
+    J[1, 5] = 1
+    # This Jacobian depends on x (it's not constant): return 0
+    return 0
```
matthew-brett: ditto

dipy/align/transforms.pyx
```diff
@@ -0,0 +1,810 @@
...
+    for j in range(3):
+        J[j,:12] = 0
+    J[0, :3] = x[:3]
+    J[0, 3] = 1
+    J[1, 4:7] = x[:3]
+    J[1, 7] = 1
+    J[2, 8:11] = x[:3]
+    J[2, 11] = 1
+    # This Jacobian depends on x (it's not constant): return 0
+    return 0
```
matthew-brett: ditto

matthew-brett:
Sorry - this might be silly - but how about having a set of strategy objects like:

```
cpdef class RegMetric:
    cdef get_jacobian(double[:], double[:], double[:,:]) nogil:
    cdef param_to_matrix_function(double[:], double[:]) nogil
```

Instantiate all the ones you need for 2D and 3D, different transformations e.g:

```
regmetrics = {}
regmetrics[(TRANSLATION, 2)] = Translation2D()
```
etc, and then do ``regmetrics[(ttype, dims)]`` to get the strategy you need?

omarocegueda:
Thanks @matthew-brett! I did all the minor corrections and checked for bad inputs (raise ValueError if invalid), I also added tests to verify that exceptions are raised. Regarding the strategy pattern, I was concerned about the overhead of calling a class' method (we need to call it for each voxel in each iteration), that's why I preferred to pass the pointer to the function. Let me do some tests to see if the class doesn't add a significant overhead, and if it doesn't, I totally agree with you, we should use that pattern instead, it's much cleaner.

dipy/align/transforms.pyx
omarocegueda: Exceptions are now raised from the "def" if the corresponding "cdef" returns an error value

dipy/align/transforms.pyx
omarocegueda: The def's is now in charge of passing correct arrays to the cdef's or raise a ValueError if the request was invalid

dipy/align/transforms.pyx
omarocegueda: All return values are now documented

matthew-brett:
Omar - I think, if you use cpdef methods, you get C calling speed from Cython code - see http://docs.cython.org/src/userguide/early_binding_for_speed.html

omarocegueda:
I see!, thanks for pointing that out!, but what I was concerned about was
the overhead of polymorphism (sorry, I am "thinking in C++" here). If I
understand correctly, for the strategy pattern, I need to declare an
abstract class, say Transform, with two abstract methods:
class Transform{
    virtual void get_jacobian( ... ) = 0;
    virtual void param_to_matrix_function( ... ) = 0
};

and then derive from Transform defining specific behaviors for each
transform so that they can be resolved in run time:
class Translation2D:Transform{
    void get_jacobian(...){...}
    void param_to_matrix_function( ... ){...}
};

Since the specific function must be resolved in run time, there must be
some overhead in comparison to just having a pointer to the function which
can be called immediately. Well, that was what I thought, but I have never
implemented this kind of hierarchies in Cython (I may be totally wrong!) so
I need to try and see how it works.

Thanks Matthew! very educational comments, as always! =)


On Mon, Feb 23, 2015 at 4:09 PM, Matthew Brett <notifications@github.com>
wrote:

> Omar - I think, if you use cpdef methods, you get C calling speed from
> Cython code - see
> http://docs.cython.org/src/userguide/early_binding_for_speed.html
>
> —
> <https://github.com/nipy/dipy/pull/574#issuecomment-75647811>.
>



-- 
-Proverbio chino.
-Chinese proverb.

http://www.cimat.mx/~omar

matthew-brett:
I may be wrong, but I believe that cdef classes in Cython are C structs, so, as long as you declare your strategy object as being of type Transform, and your strategy objects have the same members as Transform, then you will get c struct member lookup speed - or near it.

http://docs.cython.org/src/tutorial/cdef_classes.html

omarocegueda:
Hi Matthew!,
I checked how Cython implements method overriding and I found out that it
implements its own v-tables using structs (amazing!), so we basically have
polymorphism for free, no need to explicitly define abstract base clases
nor virtual functions. The downside would be if we don't want that
polymorphic behavior, but I couldn't think of any practical example of
that, so I think the cython guys made a good decision here. I went ahead
and refactored the transforms as you suggested, the code certainly looks
much cleaner.

Thanks!
-Omar.

On Mon, Feb 23, 2015 at 5:46 PM, Matthew Brett <notifications@github.com>
wrote:

> I may be wrong, but I believe that cdef classes in Cython are C structs,
> so, as long as you declare your strategy object as being of type Transform,
> and your strategy objects have the same members as Transform, then you will
> get c struct member lookup speed - or near it.
>
> http://docs.cython.org/src/tutorial/cdef_classes.html
>
> —
> <https://github.com/nipy/dipy/pull/574#issuecomment-75663772>.
>



-- 
-Proverbio chino.
-Chinese proverb.

http://www.cimat.mx/~omar

dipy/align/transforms.pyx
```diff
@@ -0,0 +1,955 @@
...
+            vector containing the n parameters of this transform
+        x : array, shape (dim,)
+            vector containing the point where the Jacobian must be evaluated
+
+        Returns
+        -------
+        J : array, shape (dim, n)
+            Jacobian matrix of the transform with parameters theta at point x
+        """
+        n = len(theta)
```
matthew-brett: Python call?  Maybe use `theta.shape[0]`

dipy/align/transforms.pyx
```diff
@@ -0,0 +1,955 @@
...
+        if n != self.number_of_parameters:
+            raise ValueError("Invalid number of parameters: %d"%(n,))
+        m = len(x)
```
matthew-brett: `x.shape[0]` to avoid Python call?

Garyfallidis:
Hi @omarocegueda, I see only 3 files in this PR if this is going to be only about the transforms, you should change the title of the PR. Travis is failing also. Can you correct that?

dipy/align/transforms.pyx
omarocegueda: Right, now using .shape[0]. I didn't worry about this one, though, because _jacobian(...) will be called instead during optimization.

omarocegueda:
Hi @Garyfallidis!, yes, this PR only contains the Transforms part. I just changed the title. Travis was failing because "iteritems" was deprecated in Python3, now using "items".

Garyfallidis:
Do you still need the Work In Progress (WIP) flag on the title? When we see WIP it means that we need to wait until this PR is ready for a review.  You can change WIP to MRG if you think this is ready to be merged on your side or just remove the WIP.

dipy/align/transforms.pyx
```diff
@@ -0,0 +1,955 @@
...
+        r""" Parameter values corresponding to the identity
+        Sets in theta the parameter values corresponding to the identity
+        transform
+
+        Parameters
+        ----------
+        theta : array, shape(2,)
+            buffer to write the parameters of the 2D translation transform
+        """
+        theta[:2] = 0
```
Garyfallidis: Is this really applied properly in Cython?
So, this will do theta[0] = 0 and theta[1] = 1? No for loop needed?

matthew-brett: I think so : http://docs.cython.org/src/userguide/memoryviews.html

```
# NumPy-style syntax for assigning a single value to all elements.
narr_view[:, :, :] = 3
```

dipy/align/transforms.pyx
```diff
@@ -0,0 +1,955 @@
...
+
+
+cdef class RotationTransform2D(Transform):
+    def __init__(self):
+        r""" Rotation transform in 2D
+        """
+        self.dim = 2
+        self.number_of_parameters = 1
+
+    cdef int _jacobian(self, double[:] theta, double[:] x, double[:,:] J) nogil:
```
Garyfallidis: Use pep 8 and numpy style when it is possible in Cython.
For example double[:,:] J should be double[:, :] (space missing)

dipy/align/transforms.pyx
Garyfallidis: Cool! I forgot about that. Thank you @matthew-brett.

dipy/align/transforms.pyx
```diff
@@ -0,0 +1,955 @@
...
+        cdef:
+            double sa = sin(theta[0])
+            double ca = cos(theta[0])
+            double sb = sin(theta[1])
+            double cb = cos(theta[1])
+            double sc = sin(theta[2])
+            double cc = cos(theta[2])
+            double px = x[0], py = x[1], pz = x[2]
+
+        J[0, 0] = ( -sc * ca * sb ) * px + ( sc * sa ) * py + ( sc * ca * cb ) * pz
```
Garyfallidis: No need for space next to the round brackets. Your call if you think it is more readable. But you could just write
 (-sc * ca * sb) * px + (sc * sa) * py + (sc * ca * cb) * pz

dipy/align/tests/test_transforms.py
```diff
@@ -0,0 +1,230 @@
...
+                       ('SCALING', 2) : 1,
+                       ('SCALING', 3) : 1,
+                       ('AFFINE', 2) : 6,
+                       ('AFFINE', 3) : 12}
+
+    for ttype, transform in regtransforms.items():
+        assert_equal(transform.get_number_of_parameters(), expected_params[ttype])
+
+
+def test_param_to_matrix_2d():
```
matthew-brett: Consider having separate random number generator here in case you want to seed it later?

```
rng = np.random.RandomState()
```

Then use size for multiple values, eg:

```
angle, dx, dy = rng.normal(size=(3,))
```

matthew-brett:
Omar - do you have a feel for how good the test coverage is for the code?

dipy/align/tests/test_transforms.py
matthew-brett: I personally prefer ``rng.uniform`` to ``rand`` because it's more obvious that it's a number from a uniform distribution.

dipy/align/transforms.pyx
```diff
@@ -0,0 +1,955 @@
...
+        J[1, 1] = ( -sc * sb + cc * sa * cb ) * px + ( sc * cb + cc * sa * sb ) * pz
+        J[2, 1] = ( -ca * cb ) * px + ( -ca * sb ) * pz
+
+        J[0, 2] = ( -sc * cb - cc * sa * sb ) * px + ( -cc * ca ) * py + \
+                  ( -sc * sb + cc * sa * cb ) * pz
+        J[1, 2] = ( cc * cb - sc * sa * sb ) * px + ( -sc * ca ) * py + \
+                  ( cc * sb + sc * sa * cb ) * pz
+        J[2, 2] = 0
+
+        J[0,3:6] = 0
```
Garyfallidis: space missing J[0, 3:6]

dipy/align/transforms.pyx
```diff
@@ -0,0 +1,955 @@
...
+                 [ 1,  0,  0], #derivatives of [T1, T2, T3] w.r.t a3
+                 [ 0, x0,  0], #derivatives of [T1, T2, T3] w.r.t a4
+                 [ 0, x1,  0], #derivatives of [T1, T2, T3] w.r.t a5
+                 [ 0, x2,  0], #derivatives of [T1, T2, T3] w.r.t a6
+                 [ 0,  1,  0], #derivatives of [T1, T2, T3] w.r.t a7
+                 [ 0,  0, x0], #derivatives of [T1, T2, T3] w.r.t a8
+                 [ 0,  0, x1], #derivatives of [T1, T2, T3] w.r.t a9
+                 [ 0,  0, x2], #derivatives of [T1, T2, T3] w.r.t a10
+                 [ 0,  0,  1]] #derivatives of [T1, T2, T3] w.r.t a11
+
```
Garyfallidis: Nice. Thank you for the detailed description in the docstring.

dipy/align/tests/test_transforms.py
```diff
@@ -0,0 +1,230 @@
...
+    factor = np.random.rand()
+    theta = np.array([factor])
+    expected = np.array([[factor, 0, 0, 0],
+                         [0, factor, 0, 0],
+                         [0, 0, factor, 0],
+                         [0, 0, 0, 1]])
+    actual = transform.param_to_matrix(theta)
+    assert_array_almost_equal(actual, expected)
+
+    # Test affine 2D
```
Garyfallidis: Comment says affine 2D but the test is about 3D affine.

dipy/align/tests/test_transforms.py
```diff
@@ -0,0 +1,230 @@
...
+    assert_array_almost_equal(actual, expected)
+
+    # Verify that ValueError is raised if incorrect number of parameters
+    for transform in regtransforms.values():
+        n = transform.get_number_of_parameters()
+
+        # Set incorrect number of parameters
+        theta = np.zeros(n + 1, dtype=np.float64)
+        assert_raises(ValueError, transform.param_to_matrix, theta)
+
```
Garyfallidis: Just a remark here. Have in mind that we have the compose_matrix and decompose_matrix functions in 
dipy.core.geometry which enable to check if your matrix composition is correct.
https://github.com/nipy/dipy/blob/master/dipy/core/geometry.py

However, the parametrization is different for the affine case than yours. We have
the shears and rotations independently. It would be interesting to have a function 
that translates your parametrization to the parametrization used in core.geometry
and SLR accordingly. 
This is only a suggestion not a required remark. But, I was thinking that it could help
us double check the numerical stability of your transforms. I played quite a bit
with the de/compose_matrix functions and I found them very stable numerically.
I think Matthew put quite some thinking and testing in those functions.

Anyway, have it in the back of your mind.

matthew-brett: I think you have the earlier Christoph Gohlke versions of the geometry functions.  I did put some work into refactoring and testing these - see : https://github.com/matthew-brett/transforms3d and specifically https://github.com/matthew-brett/transforms3d/blob/master/transforms3d/euler.py for the Euler angle stuff.

Garyfallidis: Is the new implementation more accurate in some way? Or only a better re-factoring of the version that we have already?

matthew-brett: The decompose / compose code is different from Christophe's and therefore the version you copied - I'm afraid I have not investigated relative accuracy.   The Euler stuff is a combination of refactoring / renaming / adding docstrings / adding tests and a more specific path for a common set of Euler transformations.   I think my code is in better repair than Christoph's but his code has certainly been more used.

Garyfallidis:
Hey @omarocegueda please address mine and Matthew's comments. Apart from that. This PR looks very nice and I believe is very close to be merged. Thank you for all the good work and for separating your larger PR in smaller and easier to follow PRs.

dipy/align/tests/test_transforms.py
Garyfallidis: Accuracy in the rotations/shears is super important for a good optimization. Should we open an issue to look at this at a later point? This is definitely not urgent but it would be good to know. Your call.

dipy/align/transforms.pyx
omarocegueda: Sure!, I did a full style check. Is it ok to remove the spaces when it improves readability, like this?:
https://github.com/nipy/dipy/pull/574/files#diff-7e2c30721bedfedcee9a247ccfc88934R422
If I add spaces I would need to break the line, but the code is much more readable as it is (the matrix structure is easily seen)

dipy/align/transforms.pyx
omarocegueda: Right, I removed the spaces

dipy/align/tests/test_transforms.py
omarocegueda: No problem, now using np.random.RandomState

dipy/align/tests/test_transforms.py
omarocegueda: Agreed, adding these routines was a very difficult decision for me, it feels like I am duplicating a lot of effort. The problem is that we need the Jacobians to be as fast as possible (nogil cdefs are mandatory), and having the transforms and Jacobians together is very natural and easier to follow (even more after the refactor suggested by Matthew, now we have very straightforward classes encapsulating their operations). For optimization, the "decompose" functions are not necessary, at the end we want to obtain a 4x4 matrix. The "compose" functions, IMHO, should be defined by every transform independently: to a parameter vector corresponds a matrix that depends on the transform being applied. That way we don't define a particular order in which the translation/shear/scale/rotation are applied, we just take the parameters of each transform and multiply their 4x4 matrices. Also decomposing the affine in shear, rotation, translation, is good for intuition, but for optimization purposes it is much easier to just optimize the parameters of the matrix directly.

omarocegueda:
Hi @matthew-brett!, I tried to use your find-cpdef-tests, but I am having some weird errors, I need to see what went wrong. However, we have close to full coverage: in the tests I iterate over all transforms calling all of their methods, so all methods are called at least once. I also check the ValueError exceptions. I think the only missing case is what happens when the user attempts to use the base class Transform instead of one of its derived classes (which is undefined), what do you think would be a good way to handle that case?(an abstract class would be ideal, so it couldn't be instantiated, afaik there is no such a thing as an "abstract struct"). Right now it will raise ValueError when attempting to get its Jacobian or its identity parameters or its matrix representation, and it returns -1 when queried about its dimension and number of parameters.

matthew-brett:
Yes, my function was only designed for Cython functions, so won't be much use in this case, with lots of classes.

It sounds as if the error modes for the Transform are OK - maybe a few tests for those just to remind ourselves that there are errors and the reason for them.

omarocegueda:
Alright, we should have full coverage now.
Thanks!

matthew-brett:
Good to go from my point of view - thanks Omar.

Garyfallidis:
Thx Omar. Let's move to the second PR.

PR #575, opened 2015-02-13
Speeding up LiFE

arokem:
Some small changes to reduce the memory footprint of the LiFE implementation and to speed up the model formation.

### DIFF
--- a/dipy/tracking/life.py
+++ b/dipy/tracking/life.py
@@ -345,8 +345,8 @@ class FiberModel(ReconstModel):
         affine : 4 by 4 array
             Mapping from the streamline coordinates to the data
         evals : list (3 items, optional)
-            The eigenvalues of the canonical tensor used as a response function
-
+            The eigenvalues of the canonical tensor used as a response function.
+            Default:[0.001, 0, 0].
         sphere: `dipy.core.Sphere` instance.
             Whether to approximate (and cache) the signal on a discrete
             sphere. This may confer a significant speed-up in setting up the
--- a/dipy/tracking/life.py
+++ b/dipy/tracking/life.py
@@ -15,7 +15,7 @@ import scipy.linalg as la
 
 from dipy.reconst.base import ReconstModel, ReconstFit
 import dipy.core.sphere as dps
-from dipy.utils.six.moves import range
+from dipy.utils.six.moves import range, xrange
 from dipy.tracking.utils import unique_rows
 from dipy.tracking.streamline import transform_streamlines
 from dipy.tracking.vox2track import _voxel2streamline
--- a/dipy/tracking/life.py
+++ b/dipy/tracking/life.py
@@ -290,8 +290,8 @@ def voxel2streamline(streamline, transformed=False, affine=None,
     the unique indices in this model), which fibers pass through it? Shape:
     (n_voxels, n_fibers).
 
-    The second answers the question: Given a voxel, for each fiber, which
-    nodes are in that voxel? Shape: (n_voxels, max(n_nodes per fiber)).
+    The second answers the question: Given a voxel, for each fiber in that
+    voxel, which nodes of that fiber are in that voxel? Shape (total_nodes).
     """
     if transformed:
         transformed_streamline = streamline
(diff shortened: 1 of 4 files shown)


arokem:
OK - this is ready for a review. With these changes, the memory foot-print of the algorithm decreases substantial and we get a roughly 3X speedup. We can now run this on 100's of thousands of streamlines within a reasonable amount of time (approx 40 minutes for 200k), on a reasonably sized machine. 

A future PR will do some more work to decrease the memory foot-print even more, so that we can run this on millions of streamlines without requiring big memory machines.

Garyfallidis:
Having tried this PR with 500K streamlines. I have to say that still the memory needed is massive even for a computer with 16GBytes of memory. We need to think of alternative ways to store large memory chunks (possibly using Cython and structures). Do you want me to start looking at this PR now? Or wait until you have a better idea of how a faster and with much less memory requirements COMMIT/LIFE framework can take place?

arokem:
Yes - if you think you could look at this now, that would be useful. Though it certainly does not remove al barriers, this PR provides a more efficient refactor of the previous implementation, so there's benefit to merging this in, until we find a better way to do this. I suspect that it will require some work, and I will not be able to do this work during this release cycle. Does that make sense?

arokem:
Also, relatively small PR with <100 lines added for <100 lines removed :-)

Garyfallidis:
When you say reasonably sized machine what do you mean? What is the memory necessary for 200K streamlines. Can you add that info in the tutorial too?

Garyfallidis:
@arokem this PR is fine on my side. Except from the what I asked above. Keep it up!

arokem:
I ran this on a machine with 128 GB RAM. That's large, but not unheard of.
I did a memory profiling of the previous version (
http://nbviewer.ipython.org/gist/arokem/f91d436af3f0d3084af4), and this one
should certainly fare better. If you want, I can do a more extensive
exploration of these questions, but not before sometime mid-April.

Garyfallidis:
For now, at least say up front in the docs that your algorithm requires very large amounts of memory. And that it should run in workstations with 128 GB of RAM for 200K streamlines. And it could run in a normal computers only for specific bundles but not for full datasets. But you have to say this to prevent people from crashing their PCs!

The memory profiling you have done so far  is clearly not a good representation of reality. I am sure it is clear to you that the implementation needs to be redone with some serious reduction of memory. So, let's have a meeting on mid-April to work on that. But until then. Express the issue in the tutorial please.  For me LiFE is not ready for prime time. If I had realized that it actually needs a supercomputer to run I wouldn't have let yet in Dipy. We know that other similar algorithms need much less memory. So, there are no excuses.

arokem:
The machine I have been running this on is my desktop, and you can get
machines with much more memory from AWS, so I never thought of this as
requiring a "supercomputer". I have been using this implementation, and
have some collaborators who have already been running this, and it seems to
be working OK. Sure enough, it could be made better, but that's true for
many parts of dipy (no one suggests to pull out fvtk, just because it needs
some improvements!). I would, of course, be very open to constructive
suggestions on how to improve this, and any other parts of dipy.

I have added a comment in the example, as requested.

samuelstjean:
You could specify some memory usage numbers instead of "large amounts of memory", like in the benchmarck you did. I though I had a lot of ram with 18gb of ram(most computers here are on 8gb, 2-3 of them have 16gb), so numbers would help people judge on the size of tractography they can process.

Garyfallidis:
@arokem I agree here with Sam that "large amounts of memory" is very vague. Put some numbers down. I will not continue the discussion about if this implementation needs a supercomputer or not. Or if fixing this memory issue is an improvement or an absolutely necessary fix. 

I think the most important point right now is that we need to be clear to our users.   And your PR is not clear about the memory requirements. I hope I am not asking too much.

arokem:
Working on it. Will have numbers in a short while.

arokem:
This will be the best-characterized piece of code in our whole code-base!
:-)

arokem:
OK - all set. Let me know what you think. BTW - running this entire
profiling notebook with this new version took all of 35 minutes, so it's
going pretty fast now. I agree that we need to work on the memory demands.
I am wide open for suggestions on how to do that.

PR #576, opened 2015-02-19
Piesno example

arokem:
This supersedes https://github.com/nipy/dipy/pull/390

### DIFF
--- a/dipy/denoise/noise_estimate.py
+++ b/dipy/denoise/noise_estimate.py
@@ -92,7 +92,6 @@ def piesno(data, N, alpha=0.01, l=100, itermax=100, eps=1e-5, return_mask=False)
     if data.ndim < 3:
         e_s = "This function only works on datasets of at least 3 dimensions."
         raise ValueError(e_s)
-by iterating slice by slice along the direction of acquisition
 
     if data.ndim == 4:
 
@@ -100,16 +99,21 @@ by iterating slice by slice along the direction of acquisition
         mask_noise = np.zeros(data.shape[:-1], dtype=np.bool)
 
         for idx in range(data.shape[-2]):
-            sigma[idx], mask_noise[idx] = _piesno_3D(data[idx], N, alpha=alpha,
-                                                     l=l, itermax=itermax,
-                                                     eps=eps)
+            sigma[idx], mask_noise[..., idx] = piesno_3D(data[..., idx, :],
+                                                         N,
+                                                         alpha=alpha,
+                                                         l=l,
+                                                         itermax=itermax,
+                                                         eps=eps)
 
         # Take the mode of all the sigmas from each slice as the best estimate,
-        # this should be stable with more or less 50% of the guesses at the same value.
+        # this should be stable with more or less 50% of the guesses at the
+        # same value.
         sigma, num = mode(sigma, axis=None)
 
     else:
-        sigma, mask_noise = piesno_3D(data, N, alpha=alpha, l=l, itermax=itermax, eps=eps)
+        sigma, mask_noise = piesno_3D(data, N, alpha=alpha, l=l,
+                                      itermax=itermax, eps=eps)
 
     if return_mask:
         return sigma, mask_noise
--- a/dipy/denoise/noise_estimate.py
+++ b/dipy/denoise/noise_estimate.py
@@ -1,8 +1,8 @@
 from __future__ import division, print_function
 
+import warnings
 import numpy as np
 
-from scipy.ndimage.filters import convolve
 from scipy.special import gammainccinv
 from scipy.ndimage.filters import convolve
 from scipy.stats import mode
@@ -16,116 +16,7 @@ def _inv_nchi_cdf(N, K, alpha):
 
 def piesno(data, N, alpha=0.01, l=100, itermax=100, eps=1e-5, return_mask=False):
     """
-    Probabilistic Identification and Estimation of Noise (PIESNO)
-    A routine for finding the underlying gaussian distribution standard
-    deviation from magnitude signals.
-
-    This is a re-implementation of [1]_ and the second step in the
-    stabilisation framework of [2]_.
-
-    Parameters
-    -----------
-    data : ndarray
-        The magnitude signals to analyse. The last dimension must contain the
-        same realisation of the volume, such as dMRI or fMRI data.
-
-    N : int
-        The number of phase array coils of the MRI scanner.
-        If your scanner does a SENSE reconstruction, ALWAYS use N=1, as the noise
-        profile is always Rician.
-        If your scanner does a GRAPPA reconstruction, set N as the number
-        of phase array coils.
-
-
-    alpha : float
-        Probabilistic estimation threshold for the gamma function.
-
-    l : int
-        number of initial estimates for sigma to try.
-
-    itermax : int
-        Maximum number of iterations to execute if convergence
-        is not reached.
-
-    eps : float
-        Tolerance for the convergence criterion. Convergence is
-        reached if two subsequent estimates are smaller than eps.
-
-    return_mask : bool
-        If True, return a mask identyfing all the pure noise voxel
-        that were found.
-
-    Returns
-    --------
-    sigma : float
-        The estimated standard deviation of the gaussian noise.
-
-    mask (optional): ndarray
-        A boolean mask indicating the voxels identified as pure noise.
-
-    Note
-    ------
-    This function assumes two things : 1. The data has a noisy, non-masked
-    background and 2. The data is a repetition of the same measurements
-    along the last axis, i.e. dMRI or fMRI data, not structural data like T1/T2.
-
-    This function processes the data slice by slice, but makes a global
-    estimation of the noise. Use piesno_3D to get a slice by slice estimation
-    of the noise, as in spinal cord imaging for example.
-
-    References
-    ------------
-
-    .. [1] Koay CG, Ozarslan E and Pierpaoli C.
-    "Probabilistic Identification and Estimation of Noise (PIESNO):
-    A self-consistent approach and its applications in MRI."
-    Journal of Magnetic Resonance 2009; 199: 94-103.
-
-    .. [2] Koay CG, Ozarslan E and Basser PJ.
-    "A signal transformational framework for breaking the noise floor
-    and its applications in MRI."
-    Journal of Magnetic Resonance 2009; 197: 108-119.
-    """
-
-    # This method works on a 2D array with repetitions as the third dimension,
-    # so process the dataset slice by slice.
-
-    if data.ndim < 3:
-        e_s = "This function only works on datasets of at least 3 dimensions."
-        raise ValueError(e_s)
-
-    if data.ndim == 4:
-
-        sigma = np.zeros(data.shape[-2], dtype=np.float32)
-        mask_noise = np.zeros(data.shape[:-1], dtype=np.bool)
-
-        for idx in range(data.shape[-2]):
-            sigma[idx], mask_noise[..., idx] = piesno_3D(data[..., idx, :],
-                                                         N,
-                                                         alpha=alpha,
-                                                         l=l,
-                                                         itermax=itermax,
-                                                         eps=eps)
-
-        # Take the mode of all the sigmas from each slice as the best estimate,
-        # this should be stable with more or less 50% of the guesses at the
-        # same value.
-        sigma, num = mode(sigma, axis=None)
-
-    else:
-        sigma, mask_noise = piesno_3D(data, N, alpha=alpha, l=l,
-                                      itermax=itermax, eps=eps)
-
-    if return_mask:
-        return sigma, mask_noise
-
-    return sigma
-
-
-def piesno_3D(data, N, alpha=0.01, l=100, itermax=100, eps=1e-5):
-    """
     Probabilistic Identification and Estimation of Noise (PIESNO).
-    This is the slice by slice version.
 
     Parameters
     -----------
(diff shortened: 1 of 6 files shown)


doc/examples/piesno.py
```diff
@@ -0,0 +1,103 @@
...
+import matplotlib.pyplot as plt
+fig, ax = plt.subplots(1, 2)
+ax[0].imshow(axial, cmap='gray', origin='lower')
+ax[0].set_title('Axial slice of the b=0 data')
+ax[1].imshow(axial_piesno, cmap='gray', origin='lower')
+ax[1].set_title('Background voxels from the data')
+for a in ax:
+    a.set_axis_off()
+
+# Uncomment the coming line if you want a window to show the images.
```
arokem: Let's take this out...

doc/examples/piesno.py
```diff
@@ -0,0 +1,103 @@
...
+
+Here, we obtained a noise standard deviation of 7.26. For comparison, a simple
+standard deviation of all voxels in the estimated mask (as done in the previous
+example :ref:`example_snr_in_cc`) gives a value of 6.1.
+
+"""
+
+"""
+
+.. [Koay2009] Koay C.G., E. Ozarslan, C. Pierpaoli. Probabilistic Identification and Estimation of Noise (PIESNO): A self-consistent approach and its applications in MRI. JMR, 199(1):94-103, 2009.
```
arokem: This line needs to be wrapped

dipy/denoise/noise_estimate.py
```diff
@@ -138,60 +250,8 @@ def piesno(data, N=1, alpha=0.01, l=100, itermax=100, eps=1e-5, return_mask=Fals
         # Remember the biggest omega array as giving the optimal
         # sigma amongst all initial estimates from phi
         if omega_size > max_length_omega:
-            pos, max_length_omega, best_mask = num, omega_size, idx
+            pos, max_length_omega = num, omega_size
 
         sigma[num] = sig
         mask[num] = idx
-
-    if return_mask:
```
samuelstjean: does this PR actually remove the old noise estimation stuff? It's still sueful for T1 imaging if that's the case.

samuelstjean: does this PR actually remove the old noise estimation stuff? It's still useful for T1 imaging if that's the case.

arokem:
Best would be to rebase your branch on master and then make a new PR
against my recent PR (https://github.com/nipy/dipy/pull/576). I promise to
be quick about merging in your proposed changes, so we can keep going.

dipy/denoise/noise_estimate.py
```diff
@@ -1,9 +1,11 @@
 from __future__ import division, print_function
 
+import warnings
 import numpy as np
 
 from scipy.special import gammainccinv
 from scipy.ndimage.filters import convolve
+from scipy.stats import mode
```
samuelstjean: the mode import is not used anymore, so maybe some other random stuff got in again :/

arokem: No - that's my bad. Wasn't sure what to do with this line in the rebase.
Keep looking, though. I might've messed up in other places...

dipy/denoise/noise_estimate.py
```diff
@@ -88,20 +187,28 @@ def piesno(data, N=1, alpha=0.01, l=100, itermax=100, eps=1e-5, return_mask=Fals
...
-    # prevent overflow in sum_m2
-    data = data.astype(np.float32)
-
     # Initial estimation of sigma
     denom = np.sqrt(2 * _inv_nchi_cdf(N, 1, q))
     m = np.percentile(data, q * 100) / denom
+
+    # if the percentile is 0, then more than half of the slice is zero and give
+    # up: 
+    if m == 0:
```
samuelstjean: Only thing missing if if we really want to have this check, or instead have a sanity check for when the whole slice is zero (and not only the quantile).

arokem: Sorry, I am not sure I understand.

These lines are identical in this PR, and in the PR you did against this
branch, here:

https://github.com/arokem/dipy/pull/13

Do you have a further suggestion for change? Should we change this line to
the following?

`if m==0 and np.all(data==0)`

samuelstjean: It' an open suggestion let's say. I was playing with a very noisy, background masked dataset, and it cut 2-3 lines of gm/wm at the top of the coronal slice for this reason I suspect. Relaxing this warning for np.all(data==0) would probably work, since the algorithm already exits if no noise voxel is found , but it's computed if we decide to remove this check + warning.

arokem: So you are actually suggesting to replace it by the following?

    np.all(data==0)

samuelstjean: yep, that would be more mathematically sound to me. I don't really remember why people suggested that actually, maybe they found a weird corner case on real data or something, but so far I always get a plausible mask except on cropped datasets where the first quantile estimation is screwed up to be high instead. Maybe that's actually what they meant... So far, as long as there is background (even zeros), it works well. Crop the dataset, and you screw the maths also.

arokem: Maybe we can also get someone else's opinion on this? @mdesco - does this make sense? 

Replacing `if m==0:` with `if np.all(data==0)`?

samuelstjean: We are at the MRI for the full day tomorrow, I can try asking him in between scans.

arokem: Cool. Please do. Happy scanning!

samuelstjean: completely forgot about it, our day was crammed already, , oh well he will probably see the message nevertheless

dipy/denoise/noise_estimate.py
```diff
@@ -76,6 +80,101 @@ def piesno(data, N=1, alpha=0.01, l=100, itermax=100, eps=1e-5, return_mask=Fals
...
+                                       return_mask=True)
+
+    if return_mask:
+        return sigma, mask_noise
+
+    return sigma
+
+
+def _piesno_3D(data, N, alpha=0.01, l=100, itermax=100, eps=1e-5,
+               return_mask=False):
```
fmorency: The return_mask parameter is always True (see above).

dipy/denoise/noise_estimate.py
```diff
@@ -76,6 +80,101 @@ def piesno(data, N=1, alpha=0.01, l=100, itermax=100, eps=1e-5, return_mask=Fals
...
+    itermax : int (optional)
+        Maximum number of iterations to execute if convergence
+        is not reached. Default: 100.
+
+    eps : float (optional)
+        Tolerance for the convergence criterion. Convergence is
+        reached if two subsequent estimates are smaller than eps.
+        Default: 1e-5.
+
+    return_mask : bool (optiona)
```
fmorency: Typo. *optional

dipy/denoise/noise_estimate.py
samuelstjean: Yes, the wrapper always calls the inner function with return_mask=True, then discards the mask if the user did not want it in the end at line 110. The 3D wrapped function does not return it, since it confused some people who mistook that as a brain mask to pass to a reconstruction algorithm for example, while it is instead a noise mask, identifying noisy voxels.

doc/examples/piesno.py
```diff
@@ -0,0 +1,111 @@
...
+"""
+Now that we have fetched a dataset, we must call PIESNO with the right number
+of coils used to acquire this dataset. It is also important to know what
+was the parallel reconstruction algorithm used. Here, the data comes from a
+GRAPPA reconstruction, was acquired with a 12-elements head coil available on
+the Tim Trio Siemens, for which the 12 coil elements are combined into 4 groups
+of 3 coil elements each. The signal is therefore received through 4 distinct
+groups of receiver channels, yielding N = 4. Had we used a GE acquisition, we
+would have used N=1 even if multiple channel coils are used because GE uses a
+SENSE acceleration, which has a Rician noise nature and thus N is always 1.
```
fmorency: SENSE reconstruction

doc/examples/piesno.py
```diff
@@ -0,0 +1,111 @@
...
+
+As a convenience, we will estimate the noise for the whole volume in one go,
+but it is also possible to get a slice by slice estimation of the noise if
+it is more desirable through the :func:`dipy.denoise.noise_estimate.piesno_3D`
```
fmorency: The method `dipy.denoise.noise_estimate.piesno_3D` doesn't exist anymore.

samuelstjean: Good catch, that example need some reworking since the function would return one single value at some point to not confuse people since no function at the time actually supported that, which was later changed to work as intended, and instead patch nlmeans to now accept arrays as input.

Anyhow, one can still get a single value from the function if required (by taking the mode or the median for example), but it's not hidden internally anymore.

dipy/denoise/noise_estimate.py
arokem: Yes - I think this is fine. This is a 'private' function anyhow, so users are discouraged from using it directly.

doc/examples/piesno.py
arokem: OK - I removed this paragraph. It is really obscure right now.

arokem:
Thanks for the friendly nudge, @fmorency, and thanks for the review. If you think this is now ready to go, I will merge this into master (in a couple of days, so that others can chime in here), so that you can easily use this. I will ask you (and others) to keep an eye out, and report back with any problems/issues that you encounter.

samuelstjean:
Before that I would remove the check around line 200 to look for explicitly the sane value, as opposed to a threshold of zeros.

arokem:
Like so? https://github.com/arokem/dipy/commit/00d70074853440cd8301c703160a209ac7f240fe

arokem:
No - that would probably not do. Let me rebase that out. What is the "sane value" here and how would you find it? Specifics, please.

samuelstjean:
That would be having an entire slice made of the same value. Remember there is no sane default, we are hacking togegther a convient wrapper for people, which is not covered in the original paper at all. The only in doing this is I don't really know anymore. The saved computation time is probably meaningless, since the function discards invalid estimate and exits when there are none, so it works fine without such a hack.

arokem:
So, replace the current check for `m==0` with the following check?

    if np.var(slice) < tol:
      ...

samuelstjean:
actually, juste removing that check altogether would probably still work fine. I've yet to find a real dataset to throw at it where it fails. 

The only time I got the function to blatantly lie to me is on synthetic data when there is more data than background, then the first estimate of the quantile is screwed and the function reversely picks the data instead of the background. Of course this is super obvious while looking at the mask since everything is inverted.

The only way to get that on a real dataset would be to mask it and then crop it to the minimal bounding box, in which case the hypothesis for the function is void and I have no idea how to find out this case.

That would be the ideal intention behind those lines in all honesty, which is not really covered currently.

arokem:
OK - it's actually also explicitly stated in the Notes section of the
documentation, and should be obvious to users if they read the paper and
the example.

How's this?

https://github.com/arokem/dipy/commit/bd11490be4d009b98e39b89f09228d7b1af6b17a

samuelstjean:
Well, we are back to the original version, which is fine by me actually, but some people said that we should not assume that people read the doc, so I'm at a loss at how to cover this case. 

Anyway, it's still better with all these usability notes since the current version didn't not explicitely mention one needs to wrap this function, leading to troubles in using it for everyone. So wither we assume people will read those notes, or we try to hack a way to find the case where the data is cropped.

arokem:
OK. Let's take a poll. If you have an opinion about this, please choose among one of the three following options: 

1. Leave this as is (no check)
2. Check for 0 percentile (the state of the code before: https://github.com/arokem/dipy/commit/bd11490be4d009b98e39b89f09228d7b1af6b17a)
3. Check for 0 variance (see comment above)

@fmorency - from your experience using this, what do you think about this issue? Was this a problem for you using the current implementation?

samuelstjean:
I vote for 1, original paper does not cover this case, and I don't want to break stuff by implementing approximate fixes which might not always work.

fmorency:
+1 with what @samuelstjean said.

arokem:
Alright - thanks for the votes!

I will let this lie here for a couple of days, in case anyone has some
strong objections, but barring that, I will merge as is (option #1), in a
few days.


On Thu, Mar 19, 2015 at 12:53 PM, Félix C. Morency <notifications@github.com

samuelstjean:
I'd still like to double check everything to be sure after the voting 
period ends

arokem:
As you wish. I will hold off on merging this until I hear back from you.

samuelstjean:
Ok, since nobody seems to want to add anything, I'll think I'll go for a wrapper + private function that always return the mask. github is dying right now, so I'll make a PR agaisn't this branch with those changes and we can see what people think about this form afterward.

arokem:
I was just waiting for you to test this out, and get your OK to merge this.

I am not sure what you want to address with additional work on this PR. If
you want to add anything, maybe do that on a separate PR, after we merge
this one? As mentioned earlier in this thread, the version of PIESNO on
current master is broken, and this fixes it.

fmorency:
+1 with @arokem 

Would merge the two PR and fine tune on separate PR if needed.

samuelstjean:
I don't see where it's broken actually (people just use it in a random fashion), and it might take 2-3 months to be re-merged again if I do another PR. Since it's just a small definition change I'd change it here before forgetting about it.

arokem:
If it's small, I don't see why it can't be another small PR on top of this.

If you think you might forget to do it, add it to the issues, and assign
yourself to it.

Unless I hear any loud objections, I think that I will merge this tomorrow
or the day after.

PR #578, opened 2015-02-20
WIP: try out Stefan Behnel's cython coverage

matthew-brett:
Trying Stefan's coverage work for Cython; see:

http://blog.behnel.de/posts/coverage-analysis-for-cython-modules.html

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -4,3 +4,4 @@ source = dipy
 include = */dipy/*
 omit =
     */setup.py
+plugins = Cython.Coverage
--- a/setup.py
+++ b/setup.py
@@ -8,6 +8,8 @@ from copy import deepcopy
 from os.path import join as pjoin, dirname
 from glob import glob
 
+LINETRACE = os.environ.get('DIPY_LINETRACE', False)
+
 # BEFORE importing distutils, remove MANIFEST. distutils doesn't properly
 # update it when the contents of directories change.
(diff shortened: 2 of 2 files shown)


arokem:
Happy days!

It looks like you'll need to update the wheelhouse for us to see the
results of this on travis, right?

stefanv:
I think you'll have to change the Travis script to install Coverage directly from the bitbucket repository at https://bitbucket.org/ned/coveragepy

stefanv:
Actually, that's not true; I see an alpha is out: https://pypi.python.org/pypi/coverage/4.0a5

stefanv:
https://github.com/matthew-brett/dipy/pull/2

matthew-brett:
The coverage needs absolute latest Cython master and Coverage master - I think the alpha release of coverage will not work.  There are still some problems even so:

https://groups.google.com/forum/#!topic/cython-users/EgWSRVuNHlA

Garyfallidis:
Any progress here? Should we revisit later?

Garyfallidis:
No progress on this PR. Any ideas? Is it now easier to measure Cython coverage?

stefanv:
According to http://docs.cython.org/en/latest/src/tutorial/profiling_tutorial.html, coverage causes significant slowdown for Cython, so you'll have to add a separate matrix entry.

arokem:
Looks like this is not going anywhere. I'll close for now.

PR #580, opened 2015-02-24
docstring for tensor fit was wrong

samuelstjean:
Small kind of typo according to line 762.

While I'm at it, would it be easier to instead create a mask excluding a non zero voxel like when the scanner masks the background instead of processing all voxels? All submodels could call a subfunction creating a mask where at least one voxel is not zero along the 4th D.

### DIFF
--- a/dipy/reconst/dti.py
+++ b/dipy/reconst/dti.py
@@ -748,7 +748,7 @@ class TensorModel(ReconstModel):
 
         mask : array
             A boolean array used to mark the coordinates in the data that
-            should be analyzed that has the shape data.shape[-1]
+            should be analyzed that has the shape data.shape[:-1]
 
         """
         if mask is None:


arokem:
Thanks for the docstring fix! I don't really understand your suggestion, but maybe if you implement it and put in a pull request it would be easier to discuss?

samuelstjean:
I was thinking kinda of a create a mask super function, akin to the fit method which is the same mechanism for all model, but that's only for the lazies like me who do not always craft a mask for simple stuff I guess.

arokem:
I think that I understand: you mean that this would be a way to filter out voxels that have all-zeros? Not a bad idea, but we're already somewhat dealing with that scenario through the whole `min_signal` business. Might be worth implementing at the level of `base.ReconstModel`. Do you want to have a go at implementing that? Otherwise, could you put in an issue, for when someone looks for something to fix? Thanks!

samuelstjean:
Depends, I'm looking at the dti reconst for example, and it selects the max between mean_signal and data. So does it mean it processes all zeros voxels with a constant signal (probably 1) or it skips them? If it's the former, then we can skip a ton of computations, if it's the latter I'm just derp.

PR #581, opened 2015-02-26
BF: Normalization of GQI2 `gqi_vector`.

arokem:
A long overdue fix, , with thanks to @qytian for fixing this and reminding me to
put in this PR on his behalf.

Follow-up of this conversation: http://mail.scipy.org/pipermail/nipy-devel/2013-November/009574.html

### DIFF
--- a/dipy/reconst/gqi.py
+++ b/dipy/reconst/gqi.py
@@ -104,7 +104,7 @@ class GeneralizedQSamplingFit(OdfFit):
             if self.model.method == 'gqi2':
                 H=squared_radial_component
                 #print self.gqi_vector.shape
-                self.gqi_vector = np.real(H(np.dot(self.model.b_vector,                                         sphere.vertices.T) * self.model.Lambda / np.pi))
+                self.gqi_vector = np.real(H(np.dot(self.model.b_vector,                                         sphere.vertices.T) * self.model.Lambda))
             if self.model.method == 'standard':
                 self.gqi_vector = np.real(np.sinc(np.dot(self.model.b_vector,                                   sphere.vertices.T) * self.model.Lambda / np.pi))
             self.model.cache_set('gqi_vector', sphere, self.gqi_vector)


Garyfallidis:
Hey thanks @qytian and @arokem for looking at this. Can you show as the improved ODFs in a before after figure for a 60 degrees crossing for example?

arokem:
On `master`:

master

One this branch:
qgi-norm-fix

arokem:
http://nbviewer.ipython.org/gist/arokem/f92967b718ead8fa494b

Garyfallidis:
Can you generate the same 2 figures but with sampling_length = 3?

arokem:
master

qgi-norm-fix

Garyfallidis:
Aha!

Garyfallidis:
Okay thank you. I asked for this so we can have a reference on the change of sampling_length after the new pi correction. 

Thank you both :)

qytian:
The abstract "Deconvolution enhanced Generalized Q-Sampling 2 and DSI deconvolution" recommends an integration length lamda between 2 and 3. I am not sure whether here includes the "pi" constant, i.e. [2, 3] / pi.

Garyfallidis:
Yes, because it was using the old code (without your fix). @arokem and @qytian would you mind adding this information to the docstring of the function? I mean say that range of the sampling length has changed? We need also to write this in the api_changes.rst file. Thank you again.

arokem:
Yeah - no problem. @qytian and I will work on another PR for that.

samuelstjean:
Regarding the second screenshot, is that supposed to happen? You seem to have reversed the orientation of the crossings to 90 degress on that stick figure, while it was more like 60 degrees before.

Garyfallidis:
Yeah, it's correct. It is the same as sampling from outside the PDF grid in DSI. Your values will not be correct or very small (in the case of GQI2)  and then you may start seeing effects of the sampling sphere in the results.

PR #582, opened 2015-03-01
DKI project: PR#1 Simulations to test DKI

RafaelNH:
I add some functions on dipy.sims.voxel module for DKI simulations (multi_tensor_dki, compute_Wijkl, single_diffkurt_tensors, dki_design_matrix). This PR was discussed to be the first milestone of the DKI project. To see how to use the functions please give a look to the added script on doc/examples/simulations_dki.py. Basically, the objective of the new functions is to produce ground truth DT and KT tensors to test diffusion fits (you can simulate tensors for any single voxel crossing fiber configuration ;) ). DW signals are also simulated from the computed tensors using the DKI model. Future steps on the DKI project can use these simulations to check if the DKI implementation is correct. DKI simulations modules are also relevant for DKI fit comparison studies.

Implementation was done according to PEP 8 standards and the nomenclature is consistent to previous dipy functions. However, I notice that the order of DT elements is different from previous dipy modules (as dipy.reconst.dti.py), I decided to keep this order since I notice that they are consistent to the in progress DKI implementations. Perhaps is better to transform all new functions to dipy assumed format  - Dxx Dxy Dyy Dxz Dyz Dzz. What do you thing? Please let me know if you need help doing this on the submitted changes after its revision.

Enjoy,
R Neto Henriques


dipy/sims/voxel.py
```diff
@@ -15,6 +15,17 @@
...
 diffusion_evals = np.array([1500e-6, 400e-6, 400e-6])
 
+# DT and KT for a voxel of well aligned fibers 
+# parameters are based on the simulations from:
+#
+#    Neto Henriques et al., "Exploring the 3D geometriy of the diffusion 
+#    kurtosis tensor - Impact on the development of robust tractography 
+#    procedures and novel biomarkers", NeuroImage, 2015; 111:85-99.
+#
+diffusion_tensor = np.array([1638e-6, 444e-6, 444e-6, 0, 0, 0])
```
arokem: Would it be OK to use the same diffusion evals as the ones just above (on L16)? It's starting to get crowded in this namespace...

dipy/sims/voxel.py
```diff
@@ -15,6 +15,17 @@
...
+kurtosis_tensor = np.array([1.7068, 0.8010, 0.8010, 0, 0, 0, 0, 0, 0, 0.3897, 
+                            0.3897, 0.2670, 0, 0, 0])
+
```
arokem: Just for my own understanding: are the kurtosis elements of similar units to the DT evals? Why are they 3 orders of magnitude different?

dipy/sims/voxel.py
```diff
@@ -288,7 +299,265 @@ def multi_tensor(gtab, mevals, S0=100, angles=[(0, 0), (90, 0)],
                                                  snr=None)
 
     return add_noise(S, snr, S0), sticks
+    
+def multi_tensor_dki(gtab, mevals, S0=100, angles=[(0., 0.), (90., 0.)],
+                         fractions=[50, 50], snr=20):
+    
+    r"""Simulate the elements of the diffusion and diffusion kurtosis tensor as
```
arokem: Possible typo, Should this be: "...the diffusion *tensor* and diffusion kurtosis tensor..."?

dipy/sims/voxel.py
```diff
@@ -288,7 +299,265 @@ def multi_tensor(gtab, mevals, S0=100, angles=[(0, 0), (90, 0)],
...
+    well as the diffusion signal S based on the DKI model (i.e. taylor 
+    accumulante larger than the fourth order are ignored). Simulations are
```
arokem: "Taylor expansion"?

dipy/sims/voxel.py
arokem: Would be nice to have a one-line summary at the top, and then a more expansive explanation.

dipy/sims/voxel.py
```diff
@@ -288,7 +299,265 @@ def multi_tensor(gtab, mevals, S0=100, angles=[(0, 0), (90, 0)],
...
 
     return add_noise(S, snr, S0), sticks
+    
+def multi_tensor_dki(gtab, mevals, S0=100, angles=[(0., 0.), (90., 0.)],
+                         fractions=[50, 50], snr=20):
+    
+    r"""Simulate the elements of the diffusion and diffusion kurtosis tensor as
+    well as the diffusion signal S based on the DKI model (i.e. taylor 
+    accumulante larger than the fourth order are ignored). Simulations are 
+    based on multicompartmental models which assumes that tissue are described
```
arokem: "tissue is described" or "tissue compartments are described"

dipy/sims/voxel.py
```diff
@@ -288,7 +299,265 @@ def multi_tensor(gtab, mevals, S0=100, angles=[(0, 0), (90, 0)],
...
+    by impermeable diffusion compartments characterized by their only diffusion
+    tensor.
+
+    Parameters
+    -----------
+    gtab : GradientTable
+    mevals : array (K, 3)
+        eigenvalues of the diffusion tensor for each individual compartment
+    S0 : float
```
arokem: Mark optional arguments as "(optional)" and tell us what the default values are.

dipy/sims/voxel.py
```diff
@@ -288,7 +299,265 @@ def multi_tensor(gtab, mevals, S0=100, angles=[(0, 0), (90, 0)],
...
+    angles = np.array(angles)
+    if angles.shape[-1] == 3:
+        sticks = angles
+    else:
+        sticks = [sphere2cart(1, np.deg2rad(pair[0]), np.deg2rad(pair[1]))
+                  for pair in angles]
+        sticks = np.array(sticks)
+    
+    # computing a 3D matrice containing the individual DT components 
+    mD=np.zeros((len(fractions),3,3))
```
arokem: PEP8: spaces around operators and after comma: 

    mD = np.zeros((len(fractions, 3, 3))

dipy/sims/voxel.py
```diff
@@ -288,7 +299,265 @@ def multi_tensor(gtab, mevals, S0=100, angles=[(0, 0), (90, 0)],
...
+    for i in range(len(fractions)):
+        R = all_tensor_evecs(sticks[i])
+        mD[i] = dot(dot(R, np.diag(mevals[i])), R.T)
+        
+    # compute DT
+    D=np.zeros((3,3))
```
arokem: PEP8

dipy/sims/voxel.py
```diff
@@ -288,7 +299,265 @@ def multi_tensor(gtab, mevals, S0=100, angles=[(0, 0), (90, 0)],
...
+    for i in range(len(fractions)):
+        D=D+fractions[i]*mD[i]
```
arokem: PEP8

dipy/sims/voxel.py
```diff
@@ -288,7 +299,265 @@ def multi_tensor(gtab, mevals, S0=100, angles=[(0, 0), (90, 0)],
...
+    kt[13] = compute_Wijkl(mD, fractions, 0,1,1,2, D)
+    kt[14]=  compute_Wijkl(mD, fractions, 0,1,2,2, D)
+    
+    # compute S based on the DT and KT
+    S=single_diffkurt_tensors(gtab, S0, dt, kt, snr)
+
+    return dt, kt, S
+    
+def compute_Wijkl(Dc,frac,i,j,k,el,DT=None):
+    r""" # function that computes the elements of KT by especifying their
```
arokem: You can remove the '#' at the beginning. Also: "especifying" => "specifying"

dipy/sims/voxel.py
```diff
@@ -288,7 +299,265 @@ def multi_tensor(gtab, mevals, S0=100, angles=[(0, 0), (90, 0)],
...
+    
+    # compute S based on the DT and KT
+    S=single_diffkurt_tensors(gtab, S0, dt, kt, snr)
+
+    return dt, kt, S
+    
+def compute_Wijkl(Dc,frac,i,j,k,el,DT=None):
+    r""" # function that computes the elements of KT by especifying their 
+    four indexes i, j, k and el. Each index can be equal to x, y or z which are 
+    codified respectively by 0, 1 and 2. Simulations are based on
```
arokem: I am not sure I understand the sentence: "x, y, or z which are codified by 0, 1, and 2". Does this have to do with the principal orientation of the kurtosis tensor in 3D space?

dipy/sims/voxel.py
```diff
@@ -288,7 +299,265 @@ def multi_tensor(gtab, mevals, S0=100, angles=[(0, 0), (90, 0)],
...
+
+    wijkl=0
+    MD=(DT[0][0]+DT[1][1]+DT[2][2])/3
+    for f in range(len(frac)):
+        wijkl=wijkl+frac[f]*(Dc[f][i][j]*Dc[f][k][el]+Dc[f][i][k]*Dc[f][j][el]+
+                             Dc[f][i][el]*Dc[f][j][k])
+    wijkl=(wijkl-
+           DT[i][j]*DT[k][el]-DT[i][k]*DT[j][el]-DT[i][el]*DT[j][k])/(MD**2)
+    
+    return wijkl
```
arokem: PEP8 in this entire function: space around operators and around `=`. Make sure to break lines into 80 characters when you add those spaces.

dipy/sims/voxel.py
```diff
@@ -288,7 +299,265 @@ def multi_tensor(gtab, mevals, S0=100, angles=[(0, 0), (90, 0)],
...
+        Elements of the diffusion tensor.
+    kt : (15, ) ndarray
+        Elements of the diffusion kurtosis tensor.  
+    snr : float
+        Signal to noise ratio, assuming Rician noise.  None implies no noise.
+
+    Returns
+    --------
+    S : (N,) ndarray
+        Simulated signal based on the DKI model S=S0*exp(-b*D+1/6*b**2*D**2*K)
```
arokem: You might want to render the equation with the latex engine, by adding `$` signs around it, so: 

    Simulated signal based on the DKI model: $S=S0e^{-b*D+\frac{1}{6} b^2 D^2 K}$

dipy/sims/voxel.py
```diff
@@ -288,7 +299,265 @@ def multi_tensor(gtab, mevals, S0=100, angles=[(0, 0), (90, 0)],
...
+    S0 : double,
+        Strength of signal in the presence of no diffusion gradient (also
+        called the ``b=0`` value).
+    dt : (6,) ndarray
+        Elements of the diffusion tensor.
+    kt : (15, ) ndarray
+        Elements of the diffusion kurtosis tensor.  
+    snr : float
+        Signal to noise ratio, assuming Rician noise.  None implies no noise.
+
```
arokem: Again: optional arguments and their default values.

dipy/sims/voxel.py
```diff
@@ -288,7 +299,265 @@ def multi_tensor(gtab, mevals, S0=100, angles=[(0, 0), (90, 0)],
...
+    if dt is None:
+        dt = diffusion_tensor
+
+    if kt is None:
+        kt = kurtosis_tensor
+        
+    A=dki_design_matrix(gtab)
+    
+    # define vector of DKI parameters
+    MD=sum(dt[0:3])/3
```
arokem: Is this also MD when the last 3 elements are non-zero?

dipy/sims/voxel.py
```diff
@@ -288,7 +299,265 @@ def multi_tensor(gtab, mevals, S0=100, angles=[(0, 0), (90, 0)],
...
+    X=np.concatenate((dt,kt*MD*MD,np.array([np.log(S0)])),axis=0)
```
arokem: How is `b**2` factored into this equation for the KT?

dipy/sims/voxel.py
```diff
@@ -288,7 +299,265 @@ def multi_tensor(gtab, mevals, S0=100, angles=[(0, 0), (90, 0)],
...
+    
+    # Compute signals based on the DKI model
+    S=np.exp(dot(A,X))
+
+    S = add_noise(S, snr, S0)
+
+    return S
+    
+def dki_design_matrix(gtab):
```
arokem: It's fine to put this here for now, but we'll eventually need to have just one implementation of this, preferably in the reconst/dki.py file (which doesn't exist yet on this branch...), so we need to remember to remove this from here, and import it from there when we start working on merging these things together.

dipy/sims/voxel.py
```diff
@@ -288,7 +299,265 @@ def multi_tensor(gtab, mevals, S0=100, angles=[(0, 0), (90, 0)],
...
+    r"""  
+    Constructs B design matrix for DKI
```
arokem: What's `B`?

dipy/sims/voxel.py
```diff
@@ -288,7 +299,265 @@ def multi_tensor(gtab, mevals, S0=100, angles=[(0, 0), (90, 0)],
...
+    Parameters
+    ---------
+    gtab : GradientTable
+        Measurement directions.
+        
+    Returns
+    -------
+    design_matrix : array (N,22)
+          Design matrix or B matrix for the DKI model
+            design_matrix[j, :] = (Bxx, Byy, Bzz, Bxy, Bxz, Byz,
```
arokem: Maybe write this in latex?

dipy/sims/voxel.py
```diff
@@ -288,7 +299,265 @@ def multi_tensor(gtab, mevals, S0=100, angles=[(0, 0), (90, 0)],
...
+    design_matrix : array (N,22)
+          Design matrix or B matrix for the DKI model
+            design_matrix[j, :] = (Bxx, Byy, Bzz, Bxy, Bxz, Byz, 
+                                   Bxxxx, Byyyy, Bzzzz, Bxxxy, Bxxxz,
+                                   Bxyyy, Byyyz, Bxzzz, Byzzz, Bxxyy,
+                                   Bxxzz, Byyzz, Bxxyz, Bxyyz, Bxyzz,
+                                   BlogS0)
+    """
+    b=gtab.bvals
+    bvec=gtab.bvecs
```
arokem: PEP8: spaces around `=` in these two lines.

arokem:
Did you mean to push the `Thumbs.db` file into this PR?

doc/examples/simulate_dki.py
```diff
@@ -0,0 +1,83 @@
+"""
+==========================
+DKI MultiTensor Simulation
+==========================
+
+In this example we show how someone can simulate the diffusion tensor,
+diffusion kurtosis tensor and signals of a single voxel by taking into account 
+different cellular media using a multi-compartment simulatations. Simulations
```
arokem: typo =:> "using a multi-compartment simulation".

doc/examples/simulate_dki.py
```diff
@@ -0,0 +1,83 @@
+"""
+==========================
+DKI MultiTensor Simulation
+==========================
+
+In this example we show how someone can simulate the diffusion tensor,
+diffusion kurtosis tensor and signals of a single voxel by taking into account 
+different cellular media using a multi-compartment simulatations. Simulations 
+are preformed according to:
```
arokem: => "performed"

doc/examples/simulate_dki.py
```diff
@@ -0,0 +1,83 @@
...
+
+import numpy as np
+from dipy.sims.voxel import multi_tensor_dki
+
+"""
+For the simulation we will need a GradientTable with the b-values and b-vectors
+Here we use the one we created in :ref:`example_gradients_spheres`.
+"""
+
+from gradients_spheres import gtab
```
arokem: This will run the `gradients_spheres` example. Is it very fast?

doc/examples/simulate_dki.py
```diff
@@ -0,0 +1,83 @@
...
+
+"""
+In ``mevals`` we save the eigenvalues of each tensor. To simulate crossing 
+fibers with two different media (representing intra and extra-cellular media)
+4 tissue components are taken in to account (the first two compartments
```
arokem: "in to" => "into" (no space).

doc/examples/simulate_dki.py
```diff
@@ -0,0 +1,83 @@
...
+Here we use the one we created in :ref:`example_gradients_spheres`.
+"""
+
+from gradients_spheres import gtab
+
+"""
+In ``mevals`` we save the eigenvalues of each tensor. To simulate crossing 
+fibers with two different media (representing intra and extra-cellular media)
+4 tissue components are taken in to account (the first two compartments 
+correspond to the intra and extra cellular media for the first fiber population
```
arokem: => "...intra- and extra-cellular medium..."

doc/examples/simulate_dki.py
```diff
@@ -0,0 +1,83 @@
...
+"""
+
+mevals = np.array([[0.00099, 0, 0],
+                   [0.00226, 0.00087, 0.00087],
+                   [0.00099, 0, 0],
+                   [0.00226, 0.00087, 0.00087],])
+
+"""
+In ``angles`` we save in polar coordinates (:math:`\theta, \phi`) the principal
+axis of each compartment tensor. To simulate crossing fibres at 70 degrees
```
arokem: Let's go with the US spelling everywhere: "fibers".

doc/examples/simulate_dki.py
```diff
@@ -0,0 +1,83 @@
...
+the compartments corresponding to the first fibre are aligned to the x-axis
```
arokem: Same here.

doc/examples/simulate_dki.py
```diff
@@ -0,0 +1,83 @@
...
+mevals = np.array([[0.00099, 0, 0],
+                   [0.00226, 0.00087, 0.00087],
+                   [0.00099, 0, 0],
+                   [0.00226, 0.00087, 0.00087],])
+
+"""
+In ``angles`` we save in polar coordinates (:math:`\theta, \phi`) the principal
+axis of each compartment tensor. To simulate crossing fibres at 70 degrees
+the compartments corresponding to the first fibre are aligned to the x-axis 
+while the compartments corresponding to the second fibre are aligned to the
```
arokem: Likewise.

doc/examples/simulate_dki.py
```diff
@@ -0,0 +1,83 @@
...
+x-z plane with a angular deviation of 70 degrees from the first one.
+"""
+
+angles = [(90, 0), (90, 0), (20, 0), (20, 0)]
+
+"""
+In ``fractions`` we save the percentage of the contribution of each tensor, 
+which is computed by multipling the percentage of contribution of each fiber
```
arokem: => "multiplying"

doc/examples/simulate_dki.py
```diff
@@ -0,0 +1,83 @@
...
+while the compartments corresponding to the second fibre are aligned to the
+x-z plane with a angular deviation of 70 degrees from the first one.
+"""
+
+angles = [(90, 0), (90, 0), (20, 0), (20, 0)]
+
+"""
+In ``fractions`` we save the percentage of the contribution of each tensor, 
+which is computed by multipling the percentage of contribution of each fiber 
+populatuion and the water fraction corresponding each fiber media
```
arokem: Should that be "medium"?

doc/examples/simulate_dki.py
```diff
@@ -0,0 +1,83 @@
...
+
+angles = [(90, 0), (90, 0), (20, 0), (20, 0)]
+
+"""
+In ``fractions`` we save the percentage of the contribution of each tensor, 
+which is computed by multipling the percentage of contribution of each fiber 
+populatuion and the water fraction corresponding each fiber media 
+"""
+
+fie = 0.49 # intra axonal water fraction
```
arokem: PEP8: at least two spaces before inline comment.

arokem: Is this an arbitrary number, or taken from the literature?

doc/examples/simulate_dki.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+import matplotlib.pyplot as plt
+
+plt.plot(signal, label='noiseless')
+
+plt.plot(signal_noisy, label='with noise')
+plt.legend()
+plt.show()
+plt.savefig('simulated_signal.png')
```
arokem: To render these into the documentation you'll need a `figure` directive in the triple-quoted stuff. See here: https://github.com/nipy/dipy/blob/master/doc/examples/reconst_dti.py#L259

doc/examples/simulate_dki.py
samuelstjean: population is also mistyped

arokem:
Nice work! Just a few comments and clarifications for now. 

In particular, you might want to use the pep8 application (`pip install pep8`) to check the files you are editing for pep8. 

By the way, this is just about the perfect size for a PR that can be reviewed in one sitting (by a human, @Garyfallidis is obviously an exception :-D)

I now also understand your comment about the DT evals/parameters. There's a bit of an inconsistency there now. We think of the tensor as having 3 eigenvalues, and then a direction, whereas you prefer to use the 6 tensor params. Is there a way to formulate the KT in a manner that is consistent with the way things are currently encoded? I am not completely averse to changing this, but want to make sure that it's necessary. 

Thanks!

dipy/sims/voxel.py
RafaelNH: Can we instead change the evals? The parameters diffusion_tensor and kurtosis_tensor are the defaults for a voxel of well aligned fibers and assuming the multi-compartmental model's parameters on Table 1 of Neto Henriques et al. (2015), so I want to keep then consistent to each other. Another option is to remove these global variables at all, and require users always giving the DT and KT as input of function single_diffkurt_tensors.

dipy/sims/voxel.py
RafaelNH: Kurtosis don't have the same units of DT evals! Actually kurtosis is dimensionless - it is an statistical measure quantifying deviations of the water diffusion propagation function from gaussian. Typical values of kurtosis are on the range of [0 3].

dipy/sims/voxel.py
RafaelNH: Yes, you are right it should be diffusion tensor

dipy/sims/voxel.py
RafaelNH: Perhaps is better to say 'Taylor expansion coefficients larger than ...'.

dipy/sims/voxel.py
RafaelNH: tissue is described

dipy/sims/voxel.py
RafaelNH: ok, will do

dipy/sims/voxel.py
RafaelNH: Unfortunately, defining a principal orientation of KT is not trivial (and what will this principal orientation mean on KT? - don't forget that KT is not an ellipsoid). KT have 15 elements: Wxxxx Wyyyy Wzzzz Wxxxy Wxxxz Wxyyy Wyyyz Wxzzz Wyzzz Wxxyy Wxxzz Wyyzz Wxxyz Wxyyz Wxyzz - what I wanted to say for codifying each element I use 0 for x, 1 for y and 2 for z.

dipy/sims/voxel.py
RafaelNH: Yup! This is a matrix trace propriety - trace is invariant to tensor rotations, thus MD=trace(DT)/3=(Dxx+Dyy+Dzz)/3=(e0 + e1 + e2)/3.

dipy/sims/voxel.py
RafaelNH: b**2 is factorized on dki design matrix - see function dki_design_matrix

dipy/sims/voxel.py
RafaelNH: I agree

dipy/sims/voxel.py
RafaelNH: As for DTI, in DKI framework you can define the b-values or the B matrix. This function is computing the B matrix for DKI.

doc/examples/simulate_dki.py
RafaelNH: Yes. I use this example also to keep consistency to the simulations shown in simulate.py.

doc/examples/simulate_dki.py
RafaelNH: Ok

RafaelNH:
Thanks for the review. I will install the pep8 application. To be honest, this is my first work done in python, so apologies for some basic errors on pep8 formatting. Thanks for the comments, I will take them into account on my future work in dipy.  

What should be done next to merge this to the main repository? Do you want me to reformat first the files (if yes, how is the best way to do this at this point)? 

I know that decomposing DT in eigenvalues is very useful on DTI's framework, particularly eigenvalues are the basis to compute standard DTI metrics such as FA, axial diffusivity and radial diffusivity. However, as I mentioned on my previous comments, writing KT in terms of eigenvalues might not the most practical approach. Since KT isn't a second order elliptical tensor, its eigenvalues do not have a direct biological interpretation (as for example the kurtosis in the direction of the white matter fibers). As an alternative we could process KT relative to the principal axis of DT, however, apart from given the values of kurtosis values along the DT eigenvector, this will not be useful to compute more general measures of kurtosis, to apply DKI biological modeling or to estimate the DKI based ODF. Thus, I will strongly recommend you to keep KT on the format (Wxxxx, Wyyyy, Wzzzz, Wxxxy, Wxxxz, Wxyyy, Wyyyz, Wxzzz, Wyzzz, Wxxyy, Wxxzz, Wyyzz, Wxxyz, Wxyyz, Wxyzz).

arokem:
To be clear: the code is really clean, and well-described, and overall
looks great! I make many more PEP8 errors when I submit PRs :-).

Code reviews tend to be terse and to the point, so at first, you might feel
criticized when your code is reviewed, but with time you'll see that it is
a wonderful way to collaborate with others on code. For me, it is still a
great way to learn from other people who read the code I write. Ultimately,
being brief in review is helpful, because it goes faster. We are delighted
that you have decided to take the plunge and contribute your code to an
open-source project (and particularly to this open-source project!), and I
think that I speak on behalf of all of us, that we would love for you to
stick around.

For now, going through the comments and fixing the small things would be
helpful. I think that we would want to hold off on merging this into
master, until we have the DKI fitting part as well, and do that in the next
release cycle, not the upcoming one, currently scheduled for March 14th +/-
2.

I am trying to think what the best workflow here would be. Here are some
options:

1. We wait until after the upcoming release and then merge this one,
followed by work on the other PR.

2. We start working on the other PR, as a branch off of this branch. A
little bit more complicated, especially considering that this is the first
git experience for some of you.

3. Merge this in, when it is ready, and then do the next PR whenever we get
to it, regardless of the release.

Which one do you/others prefer?

arokem:
I realize that left some room for ambiguity: for now keep working on your
files, make more commits, and push these commits to this same branch. They
will automatically appear in this PR, and we can continue the discussion on
the various issues here.

arokem:
As for format of the tensor/kurtosis tensor, it makes sense to me to leave
the kurtosis tensor in the lower diagonal format that you suggested.

I would suggest using your DT eigenvalues everywhere here. They're well
within the range of the values in the Basser and Pierpaoli paper cited
here: https://github.com/nipy/dipy/blob/master/dipy/sims/voxel.py#L13, and
I think that it shouldn't have any unwanted side effects.

Is there any reason to use the DT lower diagonal format? Note that there
are functions in `dti.py` for moving between eigenvalue/eigenvector
representations and the lower diagonal format. Of course, in practice, we
want to represent tensors as eigenvectors/eigenvalues, so that we can
quickly compute things like FA, but this might not be such an important
consideration in simulations.

RafaelNH:
Hello again! I didn't felt criticized, actually I was very happy to receive your comments. As I told you before, this is my first project in python so the more comments I receive the better. 

For now, I will continue working on this branch - I will try to clean it ASAP (not sure if I will have time this week though). For the options that you gave me I will prefer the 3rd one. I have been thinking on the future development of DKI in dipy, and in fact we are not entirely dependent on the DKI fitting modules to follow other DKI progress. We can just use the simulated tensors to carrying on with the development of DKI biological modelling (particularly the models proposed by Fiermans et al. 2011, NeuroImage 2011; 58(1): 177-88) and DKI based fiber estimation. If you agree, after cleaning this branch and merging it, I will carrying with these developments on separate branches.

Regarding to the DT lower diagonal, I think we should leave it for now and wait for the opinion of the rest of the team. For the things that I have in mind, I will need the DT in the same consistent form relative to the KT.  However, if the rest of the DKI team requires DT's eigenvalues/eigenvectors, I will not have any problem in adding this (for the future modules that I have in mind, I would only need to add a small line in the beginning of relevant functions to rewrite DT in the right form, so this isn't a big deal).

arokem:
That all sounds great, and I'd be happy for other people's input on the
tensor representation. The other thing that I just noticed is that you'll
need to write some tests for the new code, before we can merge it. Take a
look at the code in test_voxel.py, to give you a sense of the kinds of
tests you would need to write. We use 'nose' for testing, and you can use
the nose coverage tool to tell you what lines are not being exercised by
the tests. There is also an automated build bot that tests your code on
travis: https://travis-ci.org/nipy/dipy/jobs/52645126. One of the jobs (the
one with COVERAGE=1) contains the coverage report. Try to be as exhaustive
as you can in the tests, not only testing each line of the code, but also
testing different situations.

For example, for the tensor representation issue, I wonder what would
happen if the lower diagonal representation was all non-zeros (i.e. a
tensor not aligned with any of the axes).

doc/examples/simulate_dki.py
```diff
@@ -0,0 +1,83 @@
...
+==========================
+DKI MultiTensor Simulation
+==========================
+
+In this example we show how someone can simulate the diffusion tensor,
+diffusion kurtosis tensor and signals of a single voxel by taking into account 
+different cellular media using a multi-compartment simulatations. Simulations 
+are preformed according to: 
+
+[1] R. Neto Henriques et al., "Exploring the 3D geometriy of the diffusion
```
samuelstjean: there is a typo in geometry

dipy/sims/voxel.py
```diff
@@ -288,7 +299,265 @@ def multi_tensor(gtab, mevals, S0=100, angles=[(0, 0), (90, 0)],
...
+    >>> bvals, bvecs = read_bvals_bvecs(fbvals, fbvecs)
+    >>> gtab = gradient_table(bvals, bvecs)
+    >>> mevals=np.array(([0.0015, 0.0003, 0.0003],[0.0015, 0.0003, 0.0003]))
+    >>> e0 = np.array([1, 0, 0.])
+    >>> e1 = np.array([0., 1, 0])
+    >>> dt, kt, S =  multi_tensor_dki(gtab, mevals)
+    
+    References
+    ----------
+    .. [1] R. Neto Henriques et al., "Exploring the 3D geometriy of the
```
samuelstjean: same typo I reported down below, I'm reading backward/going up cause I'm like that.

dipy/sims/voxel.py
```diff
@@ -288,7 +299,265 @@ def multi_tensor(gtab, mevals, S0=100, angles=[(0, 0), (90, 0)],
...
+    return dt, kt, S
+    
+def compute_Wijkl(Dc,frac,i,j,k,el,DT=None):
+    r""" # function that computes the elements of KT by especifying their 
+    four indexes i, j, k and el. Each index can be equal to x, y or z which are 
+    codified respectively by 0, 1 and 2. Simulations are based on 
+    multicompartment models. Implementation of this function is based on 
+    equation 8 of:
+    
+    [1] R. Neto Henriques et al., "Exploring the 3D geometriy of the diffusion
```
samuelstjean: same typo in geometry

doc/examples/simulate_dki.py
```diff
@@ -0,0 +1,83 @@
+"""
+==========================
+DKI MultiTensor Simulation
+==========================
+
+In this example we show how someone can simulate the diffusion tensor,
```
samuelstjean: I think it's more common to say we show one can instead of someone. Is there actually a native english speaker we can poke for that?

arokem: I'm not a native English speaker, but can't really hold a conversation in
my mother tongue anymore either.

I think "we show how to simulate the diffusion tensor ..." would work well
here.

dipy/sims/voxel.py
```diff
@@ -155,8 +166,8 @@ def sticks_and_ball(gtab, d=0.0015, S0=100, angles=[(0, 0), (90, 0)],
         sticks = np.array(sticks)
 
     for (i, g) in enumerate(gtab.bvecs[1:]):
-        S[i + 1] = f0 * np.exp(-gtab.bvals[i + 1] * d) + \
-            np.sum([fractions[j] * np.exp(-gtab.bvals[i + 1] * d * np.dot(s, g) ** 2)
+        S[i + 1] = f0*np.exp(-gtab.bvals[i + 1]*d) + \
```
arokem: Spaces around operators (PEP8)

dipy/sims/voxel.py
```diff
@@ -155,8 +166,8 @@ def sticks_and_ball(gtab, d=0.0015, S0=100, angles=[(0, 0), (90, 0)],
...
+            np.sum([fractions[j]*np.exp(-gtab.bvals[i + 1]*d*np.dot(s, g)**2)
```
arokem: Here as well

dipy/sims/voxel.py
```diff
@@ -290,8 +302,283 @@ def multi_tensor(gtab, mevals, S0=100, angles=[(0, 0), (90, 0)],
...
+        dt = diffusion_tensor
+
+    if kt is None:
+        kt = kurtosis_tensor
+
+    A = dki_design_matrix(gtab)
+
+    # define vector of DKI parameters
+    MD = sum(dt[0:3])/3
+    X = np.concatenate((dt, kt*MD*MD, np.array([np.log(S0)])), axis=0)
```
arokem: Space around operators

dipy/sims/voxel.py
```diff
@@ -290,8 +302,283 @@ def multi_tensor(gtab, mevals, S0=100, angles=[(0, 0), (90, 0)],
...
+    if dt is None:
+        dt = diffusion_tensor
+
+    if kt is None:
+        kt = kurtosis_tensor
+
+    A = dki_design_matrix(gtab)
+
+    # define vector of DKI parameters
+    MD = sum(dt[0:3])/3
```
arokem: Space around operators

dipy/sims/voxel.py
```diff
@@ -290,8 +302,283 @@ def multi_tensor(gtab, mevals, S0=100, angles=[(0, 0), (90, 0)],
...
+    B[:, 11] = 4*b*b*bvec[:, 1]**3*bvec[:, 0]/6
+    B[:, 12] = 4*b*b*bvec[:, 1]**3*bvec[:, 2]/6
+    B[:, 13] = 4*b*b*bvec[:, 2]**3*bvec[:, 0]/6
+    B[:, 14] = 4*b*b*bvec[:, 2]**3*bvec[:, 1]/6
+    B[:, 15] = b*b*bvec[:, 0]**2*bvec[:, 1]**2
+    B[:, 16] = b*b*bvec[:, 0]**2*bvec[:, 2]**2
+    B[:, 17] = b*b*bvec[:, 1]**2*bvec[:, 2]**2
+    B[:, 18] = 2*b*b*bvec[:, 0]**2*bvec[:, 1]*bvec[:, 2]
+    B[:, 19] = 2*b*b*bvec[:, 1]**2*bvec[:, 0]*bvec[:, 2]
+    B[:, 20] = 2*b*b*bvec[:, 2]**2*bvec[:, 0]*bvec[:, 1]
```
arokem: All of these lines - spaces around operators!

dipy/sims/voxel.py
arokem: OK - turns out I was wrong about this. See (from the PEP8 page): 

```
If operators with different priorities are used, consider adding whitespace around the operators with the lowest priority(ies). Use your own judgment; however, never use more than one space, and always have the same amount of whitespace on both sides of a binary operator.

Yes:

i = i + 1
submitted += 1
x = x*2 - 1
hypot2 = x*x + y*y
c = (a+b) * (a-b)
No:

i=i+1
submitted +=1
x = x * 2 - 1
hypot2 = x * x + y * y
c = (a + b) * (a - b)

```

arokem: That is - disregard this kind of comment where irrelevant

dipy/sims/voxel.py
arokem: Still relevant here, though.

dipy/sims/voxel.py
arokem: And here

dipy/sims/voxel.py
arokem: And likewise here - still relevant!

RafaelNH:
In the last days I have been working on this PR. 
1) As suggested by @arokem, I added some tests for the new DKI simulations which can be found at dipy.sims.tests.test_voxel. With the exception of one line (the line to force an error if total volume fraction is larger than 100%), the testing scripts cover all DKI simulations:

1.1) test_compute_Wijkl - tests the symmetry of the elements of the KT (for example checks if Wxxxy = Wxxyx = Wxyxx = Wyxxx). All 81 non-independent KT elements have to match the 15 KT independent elements.

1.2) test_DKI_simulations_aligned_fibers - this function checks if functions multi_tensor_dki and single_diffkurt_tensors gives the same values of directional diffusivity, kurtosis and DW signals for a well aligned fiber when aligned to different orientations (parallel to axis x, y, and z).

1.3) test_DKI_crossing_fibers_simulations - the previous test did not take into account cases of crossing fibers, therefore the last test checks if all kurtosis elements are non zero (something expected for this crossing fiber simulation). In addition the given DT and KT values are compared to the values expected from previous studies. 

2) I notice that all different types of simulations (sticks_and_balls, multi_tensor, multi_tensor_dki, multi_tensor_odf, multi_tensor_pdf at dipy.sims.voxel) had repeated code for checking if directions ground truth were in Cartesian coordinates. So to take redundancy of code and make the functions easier to read I decided to add the helpers function which I named as _check_directions. I also add a test on test_voxel.py to insure that there is no bug on this function.

3) At last I have a question regarding to the function diff2eigenvectors(dx, dy, dz) at dipy.sims.tests.test_voxel.py - what it is doing there?

dipy/sims/voxel.py
arokem: This is still needed

dipy/sims/voxel.py
```diff
@@ -290,8 +295,273 @@ def multi_tensor(gtab, mevals, S0=100, angles=[(0, 0), (90, 0)],
...
+
+
+def compute_Wijkl(Dc, frac, ind_i, ind_j, ind_k, ind_l):
+    r""" Computes the diffusion kurtosis tensor element (with indexes i, j, k
+    and l) based on the individual diffusion tensor components of a
+    multicompartmental model.
+
+    Parameters
+    -----------
+    Dc : (K,3,3) ndarray
```
arokem: We should probably change this variable name to something more informative. How about `diffusion_tensors`?

doc/examples/simulate_dki.py
```diff
@@ -0,0 +1,79 @@
+"""
+==========================
+DKI MultiTensor Simulation
+==========================
+
+In this example we show how to simulate the diffusion kurtosis imaging (DKI)
+data of a single voxel. DKI captures information of tissue heterogeneity,
```
arokem: => "...about tissue heterogeneity."

doc/examples/simulate_dki.py
```diff
@@ -0,0 +1,79 @@
...
+therefore DKI simulations have to take into account different tissue
+compartments with different diffusion properties. For example, here diffusion
+heterogeneity is taken into account by modeling different compartments for the
+intra- and extra-cellular media. These simulations are performed according to:
+
+[1] R. Neto Henriques et al., "Exploring the 3D geometry of the diffusion
```
arokem: The references should be in a section of their own at the bottom of the example.

doc/examples/simulate_dki.py
arokem: This is still relevant.

dipy/sims/voxel.py
```diff
@@ -16,6 +16,22 @@
 diffusion_evals = np.array([1500e-6, 400e-6, 400e-6])
 
 
+def _check_directions(angles):
+    """
+    Helper function to check if direction ground truth have the right format
+    and are in cartesian coordinates
```
arokem: Even though it's an internal function, it still needs a proper docstring (Parameters, Returns)

dipy/sims/voxel.py
```diff
@@ -290,8 +295,273 @@ def multi_tensor(gtab, mevals, S0=100, angles=[(0, 0), (90, 0)],
...
+        should be equal to 100%.
+    ind_i : int
+        Element's index i (0 for x, 1 for y, 2 for z)
+    ind_j : int
+        Element's index j (0 for x, 1 for y, 2 for z)
+    ind_k : int
+        Element's index k (0 for x, 1 for y, 2 for z)
+    ind_l: int
+        Elements index l (0 for x, 1 for y, 2 for z)
+    DT : (3,3) ndarray (optional)
```
arokem: This is not a parameter

arokem:
There's still a file called `Thumbs.db` that was added into `doc/examples`. I suspect it's some cruft your operating system added in there. Could you please take it out before we merge this?

doc/examples/simulate_dki.py
```diff
@@ -0,0 +1,124 @@
...
+'small_64D' was acquired with one non-zero bvalue we artificialy produce a
+second non-zero b-value.
+"""
+
+bvals = np.concatenate((bvals, bvals * 2), axis=0)
+bvecs = np.concatenate((bvecs, bvecs), axis=0)
+
+"""
+The b-values and gradient directions are then converted to Dipy's Gradient
+Table format.
```
arokem: No space between `Gradient` and `Table`: `GradientTable`

arokem:
And one more last comment, I promise: the example is pretty good, but it would be even better if it helped us understand the DKI model a bit better, for example, by simulating just the DTI model for that same voxel, and maybe saying something about how they differ from each other. But that's it - I think that this is ready for the merge, apart from these tiny things.

arokem:
Having now read your [blog post](http://gsoc2015dipydki.blogspot.com/2015/06/rnh-post-4-first-report-1st-week-of.html), I am worried about this transposition business. You mean to say that `all_tensor_evecs` is doing the wrong thing? We might want to fix that, instead of patching around it. In particular, I might have introduced this potential error here (as you pointed out in our previous conversation): 

https://github.com/nipy/dipy/commit/c6e5838cb393d2285adf35002cecd0e9e617fc72

Is there something wrong with that commit?

arokem:
The only place I could see clear instruction on this is in the docstring of the `single_tensor` function which says that `The eigenvectors needs to be column wise.`. 

That means that https://github.com/nipy/dipy/commit/c6e5838cb393d2285adf35002cecd0e9e617fc72 was in fact wrong. Still investigating.

arokem:
On the other hand, the `multi_tensor` simulation function simply takes the return value from all_tensor_evecs and transposes it on input: https://github.com/nipy/dipy/blob/master/dipy/sims/voxel.py#L287, which might also be a reasonable thing to do. 

To be honest, I would expect `all_tensor_evecs` to return the eigen-vectors on the rows (consistent with the output of `reconst.dti`), not on the columns, so I would rather keep it as is but add some documentation clarifying this is what is happening. Will make a PR for that.

arokem:
A bit more investigation shows that I am wrong: dti does return eigenvectors column-wise, so we should probably change all_tensor_evecs to do that as well.

RafaelNH:
I just add the example comparing DKI simulations with a simulation using only DTI components. I also add a small paragraph explaining the differences. Please let me know if everything is clear.

I fix the function all_tensor_evecs to return the eigenvectors in column-wise and made the necessary changes on the parts of voxel.py that were calling this function.

RafaelNH:
PS: Having the eigenvectors column-wise is also consistent to numpy (np.linalg.eigh).

doc/examples/simulate_dki.py
```diff
@@ -0,0 +1,142 @@
+"""
+==========================
+DKI MultiTensor Simulation
+==========================
+
+In this example we show how to simulate the diffusion kurtosis imaging (DKI)
+data of a single voxel. DKI captures information about the non-Gaussian
+properties of water diffusion which is a consequence of the existence of tissue
+barriers and compartments. In this simulations compartmental heterogeneity is
```
arokem: typo: "this simulations" => "these simulations"

arokem: or "this simulation", if you want the singular

doc/examples/simulate_dki.py
```diff
@@ -0,0 +1,142 @@
...
+.. figure:: simulated_dki_signal.png
+   :align: center
+   **Simulated signals obtain from the DTI and DKI models**.
+
+Non-Gaussian diffusion properties in tissues are responsible to smaller signal
+attenuations for larger bvalues when compared to signal attenuations from free
+gaussian water diffusion. This feature can be shown from the figure above,
+since signals simulated from the DKI models reveals larger DW signal
+intensities than the signals obtained only from the diffusion tensor
+components.
```
arokem: Great!

arokem:
OK - this is ready to go! I'll fix the typo in the example in a subsequent commit, once I merge this one in. 

Congratulations @RafaelNH on your first PR! On to DKI fitting!

PR #584, opened 2015-03-02
DOC: explain the changes made to the integration lenght in GQI2.

qytian:
Following up on https://github.com/nipy/dipy/pull/581


Garyfallidis:
Thank you very much @qytian 
We will be very happy to receive other contributions from you in the future. 
Thanks again! :+1:

PR #585, opened 2015-03-09
NF: Initial file strcuture skeleton for amico implementation

flrgsr:
Hi guys,

This is in the context of the GSoC15 proposal number **2**: ["Efficient model fitting for microstructure modelling"] (https://github.com/nipy/dipy/wiki/Google-Summer-of-Code-2015)

I am not sure if I should request a pull at this stage already.
As far as I understand a first pull request is mandatory for all PSF associated projects.
 @Garyfallidis @arokem 

Best,
Florian

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+*.DS_Store
 *.pyc
 *.pyd
 *.so


Garyfallidis:
Hi @flrgsr no you shouldn't. I think there was a misunderstanding. Sorry for that. There is already someone else working on the amico project. Actually, he is part of the team that developed amico. So, this topic will not be part of the GSoC this year. The document was to give you a generic information of how to contribute to Dipy. We will have to discuss another project with you first. It could be related to microstructure or amico. Let us think about that for a bit and will get back to you.

PR #586, opened 2015-03-10
BF: all_tensor_evecs should rotate from eye(3) to e0.

arokem:
This also entails removing some hackish stuff in geometry.

H/T to @cpoetter for pointing this out.


stefanv:
:+1:

PR #587, opened 2015-03-11
Fvtk 2.0 PR1

Garyfallidis:
This will make you happy @arokem and @stefanv! Moving forward with the visualization capabilities of DIPY!!! The PR is ready for a review.

This pull request introduces a new API for visualization.

Here are the main characteristics:
a) Allows for testable screenshots and renderers. Which is crucial for quality assurance.
b) It contains 3 main modules (window, actor and widget). These are replacing the current fvtk module but fvtk can be used too for backwards compatibility.
c) In window we have everything related to windowing, rendering and saving screenshots.
d) In actor you will find the actors for lines and streamtubes etc. Not everything has moved yet from fvtk.py so that the PR doesn't become too large. But all actors should move in the next PRs.
e) In widget we have buttons and sliders which can be used directly from inside OpenGL/VTK. No new dependencies.
d) You can save and open file dialogues using Tk which is ported together with Python. No new dependencies.
f) Because they are a lot of new concepts and ideas in this PR I wrote 4 tutorials to help the reviewing process. 
 
Let me know what you think.


omarocegueda:
Hi @Garyfallidis,
I just started playing with the new visualization tools. When I run the `viz_hud.py` example, there seems to be a bug with the transparency:
tubes_right
The orange streamtube is always rendered on top of the red one, regardless of their relative position:
tubes_wrong

omarocegueda:
by the way, they are correctly rendered if not using transparency...

doc/examples/viz_bundles.py
```diff
@@ -0,0 +1,224 @@
+"""
+========================================
+Visualize bundles and metrics on bundles
+========================================
+
+Frist, let's download some available datasets. Here we are using a dataset
```
omarocegueda: Typo: 'First'

doc/examples/viz_bundles.py
```diff
@@ -0,0 +1,224 @@
...
+from dipy.viz import window, actor
+from dipy.data import fetch_bundles_2_subjects, read_bundles_2_subjects
+from dipy.tracking.streamline import transform_streamlines
+
+fetch_bundles_2_subjects()
+dix = read_bundles_2_subjects(subj_id='subj_1', metrics=['fa'],
+                              bundles=['cg.left', 'cst.right'])
+
+"""
+Store franctional anisotropy.
```
omarocegueda: typo: 'fractional'

omarocegueda:
Just a question: is vtk5 going to be supported? (I see some commits mentioning 'vtk5 support'). I have vtk 5, and when I run some tutorials I get:
```ImportError: Snapshot is available only for VTK 6+```
If it is going to be supported, can we do something to prevent the tutorial from breaking with vtk5?

File: doc/examples/viz_slice.py [hunk 3]
```diff
@@ -0,0 +1,208 @@
...
+image. We can set the range that we would like to see.
+"""
+
+mean, std = data[data > 0].mean(), data[data > 0].std()
+value_range = (mean - 0.5 * std, mean + 1.5 * std)
+
+"""
+The ``slice`` function will read data and resample the data using an affine
+transformation matrix. The default behavior of this function is to show the
+the middle slice of the last dimension of the resampled data.
```
omarocegueda: typo: "the the"

doc/examples/viz_slice.py
```diff
@@ -0,0 +1,208 @@
...
+"""
+The same actor can show any different slice from the given data using its
+``display`` function. However, if we want to show multiple slices we need to
+copy the actor first.
+"""
+
+slice_actor2 = slice_actor.copy()
+
+"""
+Now we have a new ``slice_actor`` which displays the middle slice of saggital
```
omarocegueda: typo: "saggital" -> "sagital"

doc/examples/viz_advanced.py
```diff
@@ -0,0 +1,182 @@
...
+The following function outputs a dictionary with the required bundles e.g. af
+left and maps, e.g. FA for a specific subject.
+"""
+
+res = read_bundles_2_subjects('subj_1', ['t1', 'fa'],
+                              ['af.left', 'cst.right', 'cc_1'])
+
+"""
+We will use 3 bundles, FA and the affine transformation that brings the voxel
+cordinates to world coordinates (RAS 1mm).
```
omarocegueda: typo: 'cordinates' --> 'coordinates'

doc/examples/viz_advanced.py
```diff
@@ -0,0 +1,182 @@
...
+start the interaction using ``show_m.start()``.
+"""
+
+# show_m.render()
+# show_m.start()
+
+"""
+However, if you change the window size, the slider will not update its position
+properly. The solution to this issue is to update the position of the slider
+using its ``place`` method everytime the window size changes.
```
omarocegueda: typo: "everytime" --> "every time"

Garyfallidis:
Sorry for being late to respond. I will get back to you asap. If snapshot is complaining for you. Just use ``window.record``. Soon with more info and some small bug fixes!

doc/examples/viz_slice.py
Garyfallidis: It's actually. Sagittal! thx

Garyfallidis:
Hi @omarocegueda the default rendering in vtk is object based for translucent objects. But I think it is a good idea to support depth rendered translucency. So, will add this in this PR. Thx for the feedback.

Garyfallidis:
Hi @arokem and all.  I have increased the coverage to all important files of this PR (the viz modules).
I am showing here the coverage I get on my machine Ubuntu with VTK6. 

dipy.viz.actor  97% 
dipy.viz.widget 96%
dipy.viz.utils     94% 
dipy.viz.window 78%

You will not see this in Travis because the visual tests are skipped. These bots do not have visual display or vtk installed. Maybe we can do something about that (using docker and emulating the screen?). To run the tests on your machines you will need a GPU, VTK 5 or 6 and a Linux OS with MESA. 

The dipy.viz.window cannot go higher than 78% because we cannot write tests for opening/closing file_dialogs and windows. We can test only what it can be tested offscreen (using the snapshot command) .
Also the other modules cannot go to 100% because they have if statements that work only with VTK 5 or VTK 6.

I think the only thing missing from this PR is to correct the order of translucent actors, an option suggested by @omarocegueda I can do this in this PR on in an upcoming if this is already merged. Maybe it is better to do this in an upcoming PR because it needs a bit of investigation on performance and it may take some time. It is also a bit off topic for this PR which its purpose is to familiarize ourselves with the new design.

Apart from that I am waiting from feedback from the rest of you so I can move to the second PR on this topic. It has been some incredible amount of work so far! But it does worth it! :)

arokem:
Looking forward to reviewing this! I might not get to this before we wrap up the GSoC, though. In the meanwhile - a comment regarding testing on a headless machine: do you think xvfb would work on Travis?

stefanv:
It works, you just need an apt-get install and:

```
sh -e /etc/init.d/xvfb start export DISPLAY=:99.0
```

Garyfallidis:
Hi @stefanv that is great. Can you help me a bit to change the .travis.yml configuration for one of the python 2.7 bots (with coverage) and allow for xvfb and vtk5 or 6 installation?

dipy/data/fetcher.py
```diff
@@ -591,6 +597,125 @@ def read_syn_data():
 """
 
 
+def fetch_viz_icons():
+    """ Download icons for visualization
+    """
+    url = 'https://dl.dropboxusercontent.com/u/2481924/'
```
arokem: Dropbox is not a good place to put data you want to store for the long run. I am happy to put these data on a permanent URL here at UW, if that's helpful.

dipy/data/fetcher.py
```diff
@@ -591,6 +597,125 @@ def read_syn_data():
...
+    fname = 'icomoon.tar.gz'
```
arokem: BTW - where did you get these icons from?

dipy/data/fetcher.py
```diff
@@ -591,6 +597,125 @@ def read_syn_data():
...
+    for bun in bundles:
+
+        streams, hdr = tv.read(pjoin(dname, subj_id,
+                                     'bundles', 'bundles_' + bun + '.trk'),
+                               points_space="rasmm")
+        streamlines = [s[0] for s in streams]
+        res[bun] = streamlines
+
+    return res
+
```
arokem: You should use the `fetch_data` function here, instead of this older API. We should refactor everywhere to use this, but definitely in new code. 

For an example, you can see this recent commit, refactoring in #691 :  https://github.com/arokem/dipy/commit/e410151fcdea6c332ada1fa00e4ca9b04896ba51

dipy/viz/actor.py
```diff
@@ -0,0 +1,648 @@
...
+
+def slicer(data, affine=None, value_range=None, opacity=1.,
+          lookup_colormap=None):
+    """ Cuts 3D scalar or rgb volumes into images
+
+    Parameters
+    ----------
+    data : array, shape (X, Y, Z) or (X, Y, Z, 3)
+        A grayscale or rgb 4D volume as a numpy array.
+    affine : array, shape (3, 3)
```
arokem: Default values for kwargs

dipy/viz/actor.py
```diff
@@ -0,0 +1,648 @@
...
+        Grid to space (usually RAS 1mm) transformation matrix
+    value_range : None or tuple (2,)
+        If None then the values will be interpolated to (0, 255) from
```
arokem: What is this parameter? Is it the range into which the values will be interpolated? Condensed?

arokem: OK - I think I understand - having read the code below: the data is actually clipped into this range. Maybe say that?

dipy/viz/actor.py
```diff
@@ -0,0 +1,648 @@
...
+    set_input(image_resliced, im)
+    image_resliced.SetResliceTransform(transform)
+    image_resliced.AutoCropOutputOn()
+    image_resliced.SetInterpolationModeToLinear()
+    image_resliced.Update()
+
+    if nb_components == 1:
+        if lookup_colormap is None:
+            # Create a black/white lookup table.
+            lut = colormap_lookup_table((0, 255), (0, 0), (0, 0), (0, 1))
```
arokem: Best to define this function before you use it.

dipy/viz/actor.py
```diff
@@ -0,0 +1,648 @@
...
+        else:
+            lut = lookup_colormap
+
+    x1, x2, y1, y2, z1, z2 = im.GetExtent()
+    ex1, ex2, ey1, ey2, ez1, ez2 = image_resliced.GetOutput().GetExtent()
+
+    class ImageActor(vtk.vtkImageActor):
```
arokem: Nice one - good shimming. But maybe implement outside of the function? Might this be reusable?

dipy/viz/actor.py
```diff
@@ -0,0 +1,648 @@
...
+            im_actor.input_connection(self.output)
+            im_actor.SetDisplayExtent(*self.GetDisplayExtent())
+            im_actor.opacity(opacity)
+            return im_actor
+
+        def set_position(self, wx, wy, wz):
+            self.SetPosition(wx, wy, wz)
+
+        def get_position(self):
+            return self.GetPosition()
```
arokem: I must be missing something: why are you implementing `set_position` and `get_position` instead of calling out to the the parent class?

dipy/viz/actor.py
```diff
@@ -0,0 +1,648 @@
...
+        If an array (K, ) is given, where K is the number of points of all
+        lines then these are considered as the values to be used by the
+        colormap.
+        If an array (L, ) is given, where L is the number of streamlines then
+        these are considered as the values to be used by the colormap per
+        streamline.
+        If an array (X, Y, Z) or (X, Y, Z, 3) is given then the values for the
+        colormap are interpolated automatically using trilinear interpolation.
+
+    opacity : float
```
arokem: kwarg default values.

dipy/viz/actor.py
```diff
@@ -0,0 +1,648 @@
...
+    -----
+    Streamtubes can be heavy on GPU when loading many streamlines and
+    therefore, you may experience slow rendering time depending on system GPU.
+    A solution to this problem is to reduce the number of points in each
+    streamline. In Dipy we provide an algorithm that will reduce the number of
+    points on the straighter parts of the streamline but keep more points on
+    the curvier parts. This can be used in the following way
+
+    from dipy.tracking.distances import approx_polygon_track
+    lines = [approx_polygon_track(line, 0.2) for line in lines]
```
arokem: Maybe: 

```
... can be used in the following way::
   
    from dipy.tracking.distances import approx_polygon_track
```
etc.

This will probably make this render better in sphinx.

dipy/viz/actor.py
```diff
@@ -0,0 +1,648 @@
...
+
+    Alternatively we suggest using the ``line`` actor which is much more
+    efficient.
+
+    See Also
+    --------
+    dipy.viz.actor.line
```
arokem: => `:func:`dipy.viz.actor.line``

dipy/viz/actor.py
```diff
@@ -0,0 +1,648 @@
...
+    """
+    # Poly data with lines and colors
+    poly_data, is_colormap = lines_to_vtk_polydata(lines, colors)
```
arokem: Again - please put the implementation of the function above the call to the function

dipy/viz/actor.py
```diff
@@ -0,0 +1,648 @@
...
+    tube_filter.Update()
+    next_input = tube_filter.GetOutputPort()
+
+    # Poly mapper
+    poly_mapper = set_input(vtk.vtkPolyDataMapper(), next_input)
+    poly_mapper.ScalarVisibilityOn()
+    poly_mapper.SetScalarModeToUsePointFieldData()
+    poly_mapper.SelectColorArray("Colors")
+    poly_mapper.GlobalImmediateModeRenderingOn()
+    # poly_mapper.SetColorModeToMapScalars()
```
arokem: ?

dipy/viz/actor.py
```diff
@@ -0,0 +1,648 @@
...
+    # Color Scale with a lookup table
+    if is_colormap:
+        if lookup_colormap is None:
+            lookup_colormap = colormap_lookup_table()
+        poly_mapper.SetLookupTable(lookup_colormap)
+        poly_mapper.UseLookupTableScalarRangeOn()
+        poly_mapper.Update()
+
+    # Set Actor
+    if lod:
```
arokem: What's the difference between vtkActor and vtkLODActor? We would want to explain this in the docstring, so that people know when to use which

dipy/viz/actor.py
```diff
@@ -0,0 +1,648 @@
...
+        actor.GetProperty().SetPointSize(lod_points_size)
+    else:
+        actor = vtk.vtkActor()
+
+    actor.SetMapper(poly_mapper)
+    actor.GetProperty().SetAmbient(0.1)  # .3
+    actor.GetProperty().SetDiffuse(0.15)  # .3
+    actor.GetProperty().SetSpecular(0.05)  # .3
+    actor.GetProperty().SetSpecularPower(6)
+    # actor.GetProperty().SetInterpolationToGouraud()
```
arokem: ?

dipy/viz/actor.py
```diff
@@ -0,0 +1,648 @@
...
+    # Set Actor
+    if lod:
+        actor = vtk.vtkLODActor()
+        actor.SetNumberOfCloudPoints(lod_points)
+        actor.GetProperty().SetPointSize(lod_points_size)
+    else:
+        actor = vtk.vtkActor()
+
+    actor.SetMapper(poly_mapper)
+    actor.GetProperty().SetAmbient(0.1)  # .3
```
arokem: What are the comments here (`#.3`)?

dipy/viz/actor.py
```diff
@@ -0,0 +1,648 @@
...
+        colormap are interpolated automatically using trilinear interpolation.
+
+    opacity : float, optional
+
+    linewidth : float, optional
+        Line thickness.
+    spline_subdiv : int, optional
+        number of splines subdivision to smooth streamtubes
+    lookup_colormap : bool, optional
+        add a default lookup table to the colormap
```
arokem: Not all arguments are mentioned.

dipy/viz/actor.py
```diff
@@ -0,0 +1,648 @@
...
+        Line.
+
+    Examples
+    ----------
+    >>> from dipy.viz import fvtk
+    >>> r=fvtk.ren()
+    >>> lines=[np.random.rand(10,3), np.random.rand(20,3)]
+    >>> colors=np.random.rand(2,3)
+    >>> c=fvtk.line(lines, colors)
+    >>> fvtk.add(r,c)
```
arokem: PEP8: whitespace in this entire example.

dipy/viz/actor.py
```diff
@@ -0,0 +1,648 @@
...
+        spline_filter.SetSubdivideToSpecified()
+        spline_filter.SetNumberOfSubdivisions(spline_subdiv)
+        spline_filter.Update()
+        next_input = spline_filter.GetOutputPort()
+
+    poly_mapper = set_input(vtk.vtkPolyDataMapper(), next_input)
+    poly_mapper.ScalarVisibilityOn()
+    poly_mapper.SetScalarModeToUsePointFieldData()
+    poly_mapper.SelectColorArray("Colors")
+    # poly_mapper.SetColorModeToMapScalars()
```
arokem: ?

dipy/viz/actor.py
```diff
@@ -0,0 +1,648 @@
...
+        streamline.
+        If an array (X, Y, Z) or (X, Y, Z, 3) is given then the values for the
+        colormap are interpolated automatically using trilinear interpolation.
+
+    Returns
+    -------
+    poly_data :  vtkPolyData
+    is_colormap : bool, true if the input color array was a colormap
+    """
+
```
arokem: Un-needed Whitespace

dipy/viz/actor.py
```diff
@@ -0,0 +1,648 @@
...
+    nb_lines = len(lines)
+    nb_points = len(points_array)
+
+    lines_range = range(nb_lines)
+
+    # Get lines_array in vtk input format
+    lines_array = []
+    points_per_line = np.zeros([nb_lines], np.int64)
+    current_position = 0
+    for i in lines_range:
```
arokem: You can iterate over lines instead. Then you wouldn't need to allocate `lines_range`

dipy/viz/actor.py
```diff
@@ -0,0 +1,648 @@
...
+
+    lines_array = np.array(lines_array)
+
+    # Set Points to vtk array format
+    vtk_points = numpy_to_vtk_points(points_array)
+
+    # Set Lines to vtk array format
+    vtk_lines = vtk.vtkCellArray()
+    vtk_lines.GetData().DeepCopy(numpy_support.numpy_to_vtk(lines_array))
+    vtk_lines.SetNumberOfCells(len(lines))
```
arokem: use `nb_lines`

dipy/viz/actor.py
```diff
@@ -0,0 +1,648 @@
...
+    is_colormap = False
+    # Get colors_array (reformat to have colors for each points)
+    #           - if/else tested and work in normal simple case
+    if colors is None:  # set automatic rgb colors
+        cols_arr = line_colors(lines)
+        colors_mapper = np.repeat(lines_range, points_per_line, axis=0)
+        vtk_colors = numpy_to_vtk_colors(255 * cols_arr[colors_mapper])
+    else:
+        cols_arr = np.asarray(colors)
+        if cols_arr.dtype == np.object:  # colors is a list of colors
```
arokem: This could be an `elif` clause

dipy/viz/actor.py
```diff
@@ -0,0 +1,648 @@
...
+    scale_range : tuple
+        It can be anything e.g. (0, 1) or (0, 255). Usually it is the mininum
+        and maximum value of your data.
+    hue_range : tuple of floats
+        HSV values (min 0 and max 1)
+    saturation_range : tuple of floats
+        HSV values (min 0 and max 1)
+    value_range : tuple of floats
+        HSV value (min 0 and max 1)
+
```
arokem: Default values.

dipy/viz/actor.py
```diff
@@ -0,0 +1,648 @@
...
+    return lookup_table
+
+
+def scalar_bar(lookup_table=None, title=" "):
+    """ Default scalar bar actor for a given colormap
+
+    Parameters
+    ----------
+    lookup_table : vtkLookupTable or None
+        If None then ``colormap_lookup_table`` is called with default options.
```
arokem: => :func:`colormap_lookup_table`

dipy/viz/actor.py
```diff
@@ -0,0 +1,648 @@
...
+    title : str
+
+    Returns
+    -------
+    scalar_bar : vtkScalarBarActor
+
+    """
+    lookup_table_copy = vtk.vtkLookupTable()
+    # Deepcopy the lookup_table because sometimes vtkPolyDataMapper deletes it
```
arokem: This comment refers to the code two lines below?

dipy/viz/actor.py
```diff
@@ -0,0 +1,648 @@
...
+
+def axes(scale=(1, 1, 1), colorx=(1, 0, 0), colory=(0, 1, 0), colorz=(0, 0, 1),
+         opacity=1):
+    """ Create an actor with the coordinate's system axes where
+    red = x, green = y, blue = z.
+
+    Parameters
+    ----------
+    scale : tuple (3,)
+        axes size e.g. (100, 100, 100)
```
arokem: Default value

dipy/data/fetcher.py
```diff
@@ -591,6 +597,125 @@ def read_syn_data():
...
+    folder = pjoin(dipy_home, 'exp_bundles_and_maps')
+
+    url_list = [url]
+    md5_list = ['97756fbef11ce2df31f1bedf1fc7aac7']
+    fname_list = [fname]
+
+    if not os.path.exists(folder):
+        print('Creating new directory %s' % folder)
+        os.makedirs(folder)
+        print('Downloading dataset ...')
```
arokem: We usually tell people how much data they're downloading. Could we do that here as well? I am running the example and  it seems like it's a lot of data coming down, but I don't know exactly how much...

dipy/viz/tests/test_fvtk_actors.py
```diff
@@ -0,0 +1,211 @@
...
+
+import numpy.testing as npt
+from nibabel.tmpdirs import TemporaryDirectory
+from dipy.tracking.streamline import center_streamlines, transform_streamlines
+from dipy.align.tests.test_streamlinear import fornix_streamlines
+
+
+@npt.dec.skipif(not actor.have_vtk)
+@npt.dec.skipif(not actor.have_vtk_colors)
+@npt.dec.skipif(not window.have_imread)
```
arokem: Since this triplet of tests repeats several times here, maybe you want a single:

`run_test = actor.have_vtk and actor.have_vtk_colors and window.have_imread`

And then `skipif(not run_tests)`?

arokem:
OK - that's all I have time to review today. Will follow up during this week. For the time being, if you get a chance, please rebase this one, so we don't fall to out of sync. 

Also - I think that this breaks some of the existing examples. I just ran the reconst_dti example from this branch, and the fvtk figures are not as they should be. We'd need to make an effort to make this transition while still keeping those examples running properly.

doc/examples/viz_advanced.py
```diff
@@ -0,0 +1,187 @@
...
+for backwards compatibility but now there is a more comprehensive way to access
+the main functions using the following modules.
+"""
+
+import numpy as np
+from dipy.viz import actor, window, widget
+
+"""
+In ``window`` we have all the objects that connect what needs to be rendered
+to the display or the disk e.g. for saving screenshots. So, there you will find
```
arokem: => "(e.g., for saving screenhots)" 

Either way "e.g." (almost) always has a comma after it.

doc/examples/viz_advanced.py
```diff
@@ -0,0 +1,187 @@
...
+for opening/saving dialogs and printing screenshots (see ``snapshot``).
+
+In the ``actor`` module we can find all the different primitives e.g.
+streamtubes, lines, image slices etc.
+
+In the ``widget`` we have some other objects which allow to add buttons
+and sliders and these interact both with windows and actors. Because of this
+they need input from the operating system so they can process events.
+
+So, let's get started. In this tutorial, we will visualize some bundles
```
arokem: This sentence doesn't need a "so" at its beginning.

doc/examples/viz_advanced.py
```diff
@@ -0,0 +1,187 @@
...
+First we need to fetch and load some datasets.
+"""
+
+from dipy.data.fetcher import fetch_bundles_2_subjects, read_bundles_2_subjects
+
+fetch_bundles_2_subjects()
+
+"""
+The following function outputs a dictionary with the required bundles e.g. af
+left and maps, e.g. FA for a specific subject.
```
arokem: Might want to expand what "af" stands for.

doc/examples/viz_advanced.py
```diff
@@ -0,0 +1,187 @@
...
+ren = window.Renderer()
+stream_actor = actor.line(streamlines)
+
+if not world_coords:
+    image_actor = actor.slicer(data, affine=np.eye(4))
+else:
+    image_actor = actor.slicer(data, affine)
+
+"""
+For fun let's change also the opacity of the slicer
```
arokem: "For fun let's change also..." => "For demonstration purposes let's also change..."

doc/examples/viz_advanced.py
```diff
@@ -0,0 +1,187 @@
...
+Connect the actors with the Renderer.
+"""
+
+ren.add(stream_actor)
+ren.add(image_actor)
+
+"""
+Now we would like to change the position of the ``image_actor`` using a slider.
+The sliders are widgets which require access to different areas of the
+visualization pipeline and therefore we don't recommend using them with
```
arokem: Do we need a deeper explanation about the visualization pipeline?

doc/examples/viz_advanced.py
```diff
@@ -0,0 +1,187 @@
...
+                       value=shape[2] / 2,
+                       label="Move slice",
+                       right_normalized_pos=(.98, 0.6),
+                       size=(120, 0), label_format="%0.lf",
+                       color=(1., 1., 1.),
+                       selected_color=(0.86, 0.33, 1.))
+
+"""
+Then, we can render all the widget and everything else in the screen and
+start the interaction using ``show_m.start()``.
```
arokem: Maybe add a sentence saying that users will have to uncomment the next couple of lines to see this.

doc/examples/viz_advanced.py
```diff
@@ -0,0 +1,187 @@
...
+global size
+size = ren.GetSize()
+
+
+def win_callback(obj, event):
+    global size
+    if size != obj.GetSize():
+
+        slider.place(ren)
+        size = obj.GetSize()
```
arokem: This is annoying!

doc/examples/viz_advanced.py
```diff
@@ -0,0 +1,187 @@
...
+window.record(ren, out_path='bundles_and_a_slice.png', size=(1200, 900))
+
+"""
+.. figure:: bundles_and_a_slice.png
+   :align: center
+
+   **A few bundles with interactive slicing**.
+"""
+
+del show_m
```
arokem: Out of curiosity: why is this `del` needed?

doc/examples/viz_bundles.py
```diff
@@ -0,0 +1,224 @@
...
+fa = dix['fa']
+
+"""
+Store grid to world transformation matrix.
+"""
+
+affine = dix['affine']
+
+"""
+Store the cingulum bundle. A bundle is a set of streamlines.
```
arokem: Technically it's not really a set - it's a list.

doc/examples/viz_bundles.py
```diff
@@ -0,0 +1,224 @@
...
+"""
+
+bundle = dix['cg.left']
+
+"""
+It happened that this bundle is in world coordinates and therefore we need to
+transform in native image coordinates so that it is in the same coordinate
```
arokem: => "transform it into..."

doc/examples/viz_bundles.py
```diff
@@ -0,0 +1,224 @@
...
+
+stream_actor = actor.line(bundle_native)
+
+renderer.set_camera(position=(-176.42, 118.52, 128.20),
+                    focal_point=(113.30, 128.31, 76.56),
+                    view_up=(0.18, 0.00, 0.98))
+
+renderer.add(stream_actor)
+
+# window.show(renderer, size=(600, 600), reset_camera=False)
```
arokem: Might want to explain what happens when you uncomment this line. Otherwise it might look like cruft that needs to be removed.

doc/examples/viz_bundles.py
```diff
@@ -0,0 +1,224 @@
...
+
+Show every point with a value from a volume with your colormap
+==============================================================
+
+Here we will need to input the ``fa`` map in ``streamtube`` or ``
+"""
+
+renderer.clear()
+
+hue = [0.0, 0.0]  # red only
```
arokem: What's the scale here?

doc/examples/viz_bundles.py
```diff
@@ -0,0 +1,224 @@
...
+
+"""
+.. figure:: bundle6.png
+   :align: center
+
+   **Random colors per points per streamline**.
+
+In summary, we showed that there are many useful ways for visualizing maps
+on bundles.
+
```
arokem: In my hands, these figures all come out with the colorbar obscuring the bundle. Is there any way to prevent that?

doc/examples/viz_hud.py
```diff
@@ -0,0 +1,146 @@
...
+which will change the opacity of these tubes and c) move the streamtubes using
+a slider.
+"""
+
+import numpy as np
+from dipy.viz import window, actor, widget
+from dipy.data import fetch_viz_icons, read_viz_icons
+
+"""
+First, we add the streamtubes to the
```
arokem: to the ... what?

doc/examples/viz_hud.py
```diff
@@ -0,0 +1,146 @@
...
+def button_minus_callback(obj, event):
+    print('- pressed')
+    global opacity
+    if opacity > 0:
+        opacity -= 0.1
+    stream_actor.GetProperty().SetOpacity(opacity)
+
+
+"""
+We need to download some icons to create a face for our buttons ...
```
arokem: Don't need the ellipsis here.

doc/examples/viz_hud.py
```diff
@@ -0,0 +1,146 @@
+"""
```
arokem: What's the 'hud' in the title of this example?

doc/examples/viz_hud.py
```diff
@@ -0,0 +1,146 @@
...
+
+button_png_minus = read_viz_icons(fname='minus.png')
+button_minus = widget.button(show_manager.iren,
+                             show_manager.ren,
+                             button_minus_callback,
+                             button_png_minus, (.98, .9), (50, 50))
+
+
+def move_lines(obj, event):
+
```
arokem: whitespace

doc/examples/viz_hud.py
```diff
@@ -0,0 +1,146 @@
...
+"""
+This callback is used to update the buttons/sliders' position so they can stay
+on the correct side of the window when the window is being resized.
+"""
+
+
+def win_callback(obj, event):
+    global size
+    if size != obj.GetSize():
+
```
arokem: whitespace

doc/examples/viz_hud.py
```diff
@@ -0,0 +1,146 @@
...
+renderer.zoom(0.7)
+renderer.roll(10.)
+
+"""
+Uncomment the following line to start the interaction.
+"""
+
+# show_manager.add_window_callback(win_callback)
+# show_manager.render()
+# show_manager.start()
```
arokem: I uncomment these three lines, but the '+' and '-' buttons can't be pressed, and opacity doesn't change. Any ideas?

doc/examples/viz_hud.py
```diff
@@ -0,0 +1,146 @@
...
+
+renderer.reset_clipping_range()
+
+window.record(renderer, out_path='mini_ui.png', size=(800, 800))
+
+del show_manager
```
arokem: I wonder again about the significance of del-ing at this point in this demo

doc/examples/viz_slice.py
```diff
@@ -0,0 +1,208 @@
...
+# window.show(renderer, size=(600, 600), reset_camera=False)
+
+"""
+Otherwise, you can save a screenshot using the following command.
+"""
+
+window.record(renderer, out_path='slices.png', size=(600, 600))
+
+"""
+.. figure:: slices.png
```
arokem: This is very "pretty". Do you have any sense for what smoothing is happening here?

dipy/viz/actor.py
```diff
@@ -0,0 +1,648 @@
...
+
+def streamtube(lines, colors=None, opacity=1, linewidth=0.01, tube_sides=9,
+               lod=True, lod_points=10 ** 4, lod_points_size=3,
+               spline_subdiv=None, lookup_colormap=None):
+    """ Uses streamtubes to visualize polylines
+
+    Parameters
+    ----------
+    lines : list
+        list of N curves represented as 2D ndarrays
```
stefanv: Clarify this type: list of 2D arrays?

dipy/viz/actor.py
```diff
@@ -0,0 +1,648 @@
...
+        size of points when lod is in effect
+    spline_subdiv : int
+        number of splines subdivision to smooth streamtubes
+    lookup_colormap : bool
+        add a default lookup table to the colormap
+
+    Examples
+    --------
+    >>> from dipy.viz import fvtk
+    >>> r=fvtk.ren()
```
stefanv: Why use static methods like these?  How about:

```
r = fvtk.ren()
r.add(...)
```

stefanv: It looks like that API is implemented.  Perhaps update all the examples to use it?

doc/examples/viz_hud.py
stefanv: Heads-up display, I guess?  But not sure how it fits in.

stefanv:
This looks like a big step in the right direction to me.

doc/examples/viz_hud.py
MarcCote: Yes, HUD stands for Heads-Up Display (https://en.wikipedia.org/wiki/HUD_(video_gaming)).

arokem: Hmm. How about `viz_widgets.py` instead?

dipy/data/fetcher.py
Garyfallidis: icomoon

dipy/data/fetcher.py
Garyfallidis: I have an non-free account in Dropbox and I can store big files. We also have space in Nitrc for putting up files. We should start using. Not sure the UW is the most secure.

arokem: NITRC seems like a good solution

Garyfallidis: Okay sure. But can you tell me why Dropbox is bad? I think we never had any problem with Dropbox. And I suppose if we move all the Dropbox files to NITRC we should do that all in one PR and not here. What do you think?

arokem: Services like that tend to come and go. Management of resources on dropbox
has a bus factor of 1. If someone else needs to deal with these files, you
have to give them your dropbox password.

arokem: Also - both Stanford and UW guarantee the existence of permanent url
pointing to these files that will *never* change. I don't know that's true
for dropbox

dipy/viz/actor.py
Garyfallidis: sure

dipy/viz/actor.py
Garyfallidis: Thx. Precise explanation added.

dipy/viz/actor.py
Garyfallidis: Sorry what do you mean?

dipy/viz/actor.py
Garyfallidis: Okay this is a good question. The reasons why often the classes are inside the functions are because a) vtk classes do not allow for multiple inheritance and that makes putting the classes out tricky design-wise, b) it makes testing more complicated. Because the visual tests need to be skipped for most of the bots, c) and more importantly I don't have clear in my mind yet an object oriented design for the actors and widgets that could be ready for prime time. From my experience it is better to be conservative in the design in the beginning and be careful on putting too many objects out too quickly. As we will play more with the new viz module it will be clearer (or not) which objects can go out of the functions and which not. However, from experience right now we have to be more careful than usual. Don't forget that vtk is a very messy library in many ways. So, hiding a lot of its complications inside the functions looks like a good strategy for now.
Nonetheless, in the next PR you will see some grid and container objects which can clearly and safely be written as objects.

dipy/viz/actor.py
Garyfallidis: Good point. I was thinking to add some simple functions like this to all our actors. But maybe this should happen at a different stage. And change it everywhere.

Garyfallidis: Ok removed

dipy/viz/actor.py
Garyfallidis: ok

dipy/viz/actor.py
Garyfallidis: thx

dipy/viz/actor.py
Garyfallidis: ok

dipy/viz/actor.py
Garyfallidis: oops

dipy/viz/actor.py
Garyfallidis: sorry

dipy/viz/actor.py
Garyfallidis: Sure.

dipy/viz/actor.py
Garyfallidis: Just a reminder to myself. Will remove. Thx.

dipy/viz/actor.py
Garyfallidis: Not important. Removing.

dipy/viz/actor.py
Garyfallidis: true story

dipy/viz/actor.py
Garyfallidis: Corrected.

dipy/viz/actor.py
Garyfallidis: removed

dipy/viz/actor.py
Garyfallidis: okay dokey

dipy/viz/actor.py
Garyfallidis: It's okay. It's a very small amount of memory in relation to what is in streamlines. I will let @StongeEtienne to decide if he wants to change this code after this PR is merged. Maybe it was a bit faster this way.

dipy/viz/actor.py
Garyfallidis: sure

dipy/viz/actor.py
arokem: In the order of the file, put name definitions (functions, imports) higher up than when the name is used. That way, a reader of the code would know where to look to find where that name came from: was it defined as a function? Was it imported from somewhere else? etc.

Garyfallidis: sure

dipy/viz/actor.py
Garyfallidis: Hmm need to check that.

dipy/viz/actor.py
Garyfallidis: ok

dipy/viz/actor.py
StongeEtienne: As you can see later the integer "i" is use for another array ; "points_per_line[i] = current_len"
and later on lines_range may be use by the coloring ; "np.repeat(lines_range, points_per_line, axis=0)"

if we iterate over line "for line in lines:" we would need to also iterate over an integer "i" for "points_per_line[i]"
but maybe we dont need to allocate "lines_range" and use multiple "range(nb_lines)"

dipy/viz/actor.py
StongeEtienne: yes it refers to "lookup_table_copy.DeepCopy(lookup_table)"

dipy/viz/actor.py
StongeEtienne: If i remeber well, later on wthe the function "lines_to_vtk_polydata()"
points are stacked with "points_array = np.vstack(lines)"

so: this could be a List of 2D arrays [nb_lines][nb_pts_pl x 3]
or a 3D array [nb_lines x nb_pts x 3]

dipy/viz/actor.py
Garyfallidis: sure

dipy/viz/actor.py
Garyfallidis: okay comment moved

dipy/viz/actor.py
Garyfallidis: k

dipy/viz/actor.py
```diff
@@ -0,0 +1,661 @@
...
+colors, have_vtk_colors, _ = optional_package('vtk.util.colors')
+numpy_support, have_ns, _ = optional_package('vtk.util.numpy_support')
+
+if have_vtk:
+
+    version = vtk.vtkVersion.GetVTKSourceVersion().split(' ')[-1]
+    major_version = vtk.vtkVersion.GetVTKMajorVersion()
+
+
+def colormap_lookup_table(scale_range=(0, 1), hue_range=(0.8, 0),
```
MarcCote: This is not an actor, we should move it in another file, maybe `viz/colormap.py` ?

dipy/viz/actor.py
MarcCote: Is there a reason to have to specify the default value in the docstring when it is clearly visible in the function definition? I just checked and it is included in the documentation generated by sphinx. I agree that sometimes it might need some precisions (e.g., when the default value is `None`).

doc/examples/viz_advanced.py
MarcCote: What is? The UpperCamelCase, the global variable `size` or both!

doc/examples/viz_hud.py
MarcCote: Are you using Mac OS ? If so, you should call the script using `pythonw`. (See http://stackoverflow.com/questions/9705982/pythonw-exe-or-python-exe)

dipy/viz/actor.py
```diff
@@ -0,0 +1,661 @@
...
+
+    # actor = vtk.vtkActor()
+    actor.SetMapper(poly_mapper)
+    actor.GetProperty().SetLineWidth(linewidth)
+    actor.GetProperty().SetOpacity(opacity)
+
+    return actor
+
+
+def lines_to_vtk_polydata(lines, colors=None):
```
MarcCote: This is not an actor, we should move it in another file, probably `viz/utils.py`.

dipy/viz/utils.py
```diff
@@ -0,0 +1,96 @@
...
+        else:
+            vtk_object.SetInputData(inp)
+    elif isinstance(inp, vtk.vtkAlgorithmOutput):
+        vtk_object.SetInputConnection(inp)
+
+    vtk_object.Update()
+    return vtk_object
+
+
+def map_coordinates_3d_4d(input_array, indices):
```
MarcCote: I don't think this should be in the module `viz`. I'm sure this could be useful elsewhere in dipy, not only visualization.

dipy/data/fetcher.py
```diff
@@ -666,3 +703,92 @@ def read_cenir_multib(bvals=None):
...
+        Dictionary with data of the metrics and the bundles as keys.
+
+    Notes
+    -----
+    If you are using these datasets please cite the following publications.
+
+    References
+    ----------
+    .. [1] Renaud, E. M. Descoteaux, M. Bernier, E. Garyfallidis,
+    K. Whittingstall, “Morphology of thalamus, LGN and optic radiation do not
```
MarcCote: The non-ASCII characters are: `“` and `”`. Need to be replace or use `# -*- coding: utf-8 -*-`
  File "/mnt/data/research/src/dipy/dipy/data/fetcher.py", line 757
SyntaxError: Non-ASCII character '\xe2' in file /mnt/data/research/src/dipy/dipy/data/fetcher.py on line 758, but no encoding declared; see http://www.python.org/peps/pep-0263.html for details

dipy/viz/actor.py
Garyfallidis: True. It's more of a tradition. I am not sure if we really need to document each time the default value.

dipy/viz/tests/test_fvtk_actors.py
Garyfallidis: okay good to know

doc/examples/viz_advanced.py
Garyfallidis: yeap

Garyfallidis: thx

doc/examples/viz_advanced.py
Garyfallidis: true I have to reduce my "so" usage

doc/examples/viz_advanced.py
Garyfallidis: k

doc/examples/viz_advanced.py
Garyfallidis: k

doc/examples/viz_advanced.py
Garyfallidis: Yes, but not here. We need a tutorial basic introduction to dipy.viz. I can try to make something with the next fvtk_2.0_PR2. And then revisit this tutorial.

doc/examples/viz_advanced.py
Garyfallidis: thx

doc/examples/viz_advanced.py
Garyfallidis: I think he means that it is annoying that you have to remember to place the widgets when the window is resized. True. I haven't found a way to make that easier. But perhaps there is an alternative way that we can discover together in the future.

doc/examples/viz_advanced.py
Garyfallidis: When you call this example from ipython with qt backend the window stays open because the show_m.initialize. So del makes sure it will go away. In python the del is not necessary.

doc/examples/viz_bundles.py
Garyfallidis: list then

doc/examples/viz_bundles.py
Garyfallidis: sure

doc/examples/viz_bundles.py
Garyfallidis: sure

doc/examples/viz_bundles.py
Garyfallidis: It is explained in colormap_lookup_table.

doc/examples/viz_bundles.py
Garyfallidis: Yes. This colorbar is just an actor so you can position it where you want. Also in the next PR we will have a container object that allows you to grid things together. Finally, you can also create a colorbar widget in vtk which you are then able to drag interactively. I think the best option will be the container object because it will align the colorbar and the actors correctly. We should add an example in the container tutorial @MarcCote :)

doc/examples/viz_hud.py
Garyfallidis: lol

Garyfallidis: sorry adding now what is missing.

doc/examples/viz_hud.py
Garyfallidis: removed

doc/examples/viz_hud.py
Garyfallidis: sure

dipy/viz/actor.py
arokem: Seems you're right. The numpy docstring standard, which is what we rely on,
seems to be ambiguous:

It says: "They can also be detailed in the description".

Nevertheless, I find it to be useful, especially in cases where default
value is inferred, rather than specified. In this case, you are right that
it appears in the signature anyway.

doc/examples/viz_advanced.py
MarcCote: I found a way to make it cleaner. I'll make a PR to your branch with the fix. In short, when creating the slider (or any widget) you can add the callback function directly via `iren.GetRenderWindow()`.

doc/examples/viz_bundles.py
MarcCote: Ok, added to my TODO.

doc/examples/viz_hud.py
Garyfallidis: k

doc/examples/viz_hud.py
Garyfallidis: k

doc/examples/viz_bundles.py
Garyfallidis: thx

doc/examples/viz_slice.py
Garyfallidis: Trilinear interpolation is the default used here plus any initial min-max clipping. Look at actor.slicer.
I would like to start a discussion in the list on how people would like to see their images. What histogram corrections, interpolation strategies etc. they prefer using. It would be nice to reach a consensus on how people would like to see their images. This PR is only a first approach.

dipy/viz/actor.py
Garyfallidis: Certainly!

dipy/viz/actor.py
Garyfallidis: Correct!

dipy/viz/utils.py
Garyfallidis: Good point. But the truth is that we need a general consensus for all interpolation functions. We have too many implemented in different files. Maybe they should all go to one file. Let me write an issue about this. So, I can have some feedback from Omar and Bago who have written other similar functions like these.

Garyfallidis:
Done here! Addressed all comments and re-factored as suggested. All viz tests pass in my NVidia card. This can be merged now. And an additional PR is coming shortly packed with more awesomeness.

arokem:
Do the old examples work if we merge this?

Garyfallidis:
I am running the examples as we speak. Will know soon. We may get some issues with the old slicer. Also I got some feedback from Max that fvtk.points doesn't appear. I haven't touched this in this PR. Maybe some versioning issue. Let's see.

arokem:
OK - fwiw - this is what I get for `tensor_odfs.png` running the `reconst_dti` example: 
tensor_odfs

arokem:
`tensor_ellipsoids` comes out all black

Garyfallidis:
Interesting. Same problem with fvtk.point then. Okay will investigate asap.

Garyfallidis:
Problems with other functions?

arokem:
Yeah - just ran tracking_eudx_tensors. I get an all-white image for `tensor_tracks`. Seems like a zooming issue - it looks like I am getting too close to the background in all of these. Could it be something about the camera setting?

Garyfallidis:
Yes, the camera is now much more flexible. Maybe this is the issue.

dipy/viz/widget.py
```diff
@@ -0,0 +1,311 @@
...
+    text_widget.SetInteractor(iren)
+    text_widget.SetTextActor(text_actor)
+    text_widget.SelectableOn()
+    text_widget.ResizableOff()
+    text_widget.GetRepresentation().ProportionalResizeOn()
+
+    text_widget.AddObserver(vtk.vtkCommand.WidgetActivateEvent, callback)
+    text_widget.On()
+
+    # This is a hack for avoiding not plotting the text widget when
```
MarcCote: This hack is not working with VTK 6.

They actually fixed that bug in VTK 5.10 (see http://www.vtk.org/Bug/view.php?id=12588) but only for `vtkActor` not `vtkLODActor`.

MarcCote: I opened a bug report this: http://www.vtk.org/Bug/view.php?id=15775

dipy/data/fetcher.py
```diff
@@ -187,123 +201,146 @@ def fetcher():
...
+fetch_bundles_2_subjects = _make_fetcher(
+    "fetch_bundles_2_subjects",
+    pjoin(dipy_home, 'exp_bundles_and_maps'),
+    'https://dl.dropboxusercontent.com/u/2481924/',
+    ['bundles_2_subjects.tar.gz'],
+    ['bundles_2_subjects.tar.gz'],
+    ['97756fbef11ce2df31f1bedf1fc7aac7'],
+    data_size="234MB",
+    doc="Download 2 subjects from the SNAIL dataset with their bundles",
+    unzip=True)
```
arokem: nice

Garyfallidis:
The examples are good now and also the widgets can be placed automatically. The current Travis fails should be resolved when #731 is merged.

Garyfallidis:
Okay merged #731 now Travis should be happy. Let's see!

arokem:
Why not use `rebase`?

Garyfallidis:
Why? Is this a problem at this stage?

arokem:
Yeah - it clutters the history.

Garyfallidis:
Well now... not so important. But I can start doing more rebasing from now on (and less merging).

arokem:
It's important enough to be part of our development guidelines.

Garyfallidis:
Rebase can be hard to deal with at times. And then you need to play with the interactive version and edit your commit history etc. Which is often not pleasant. So, we should use merge at times too. It's easier and it creates less conflicts. Even if it makes the history a bit messy.

arokem:
I don't think this was that case.

Should we somehow change the instructions in the development guide to
instruct people when they should and shouldn't use `rebase`?

Garyfallidis:
True in this case this could have been a simple rebase. We should update the documentation but first we need to see what is the current general convention about rebase and merge. I see there is still a lot of debate. Have people reached a consensus on this? I mean what is good for long term etc?

Garyfallidis:
I mean that when we looked with the other nipy devs and learned git it's more than 5 years ago. At that time rebase was the new thing. It will be nice to see what is the current trend if there is one. I see for example that some people use git rerere for resolving merge conflicts. Is this something that we consider?

jchoude:
My 2 cents: what we normally try to do in the projects in which I collaborate is to use the following guidelines:

1) If your branch that needs to be updated (either by rebase or merge) has never been pushed to another remote (i.e. only local work), you should definitely try to rebase. IF, and only IF, there are some incredibly complex conflicts, drop the rebase and try with a merge.

2) If the branch has been pushed and might have been seen / cloned by someone else, either:
  2.1) If it's easy for you to coordinate with your collaborators and everyone agrees you can rebase, go ahead, rebase and force push if needed. You need to agree to make sure they didn't do local work on their copy of your branch
  2.2) Else, use a merge.

Using this, point 1) helps in keeping a mostly clean and linear history. Point 2 helps in cases where you don't want to break the workflow of other people using / working on your branch.

Of course, those guidelines are, as everything in code, a reference. There are always cases where they do not fit the situation, or you might need to do differently.

arokem:
Hey! Two more comments: 

1. This will break people's existing code. Could you please write some documentation guidelines on how to translate code from 0.9 to 0.10? Essentially add some notes telling us what you did to make all these examples run properly. 

1. I am currently getting a test failure with this branch: 

```
FAIL: dipy.viz.tests.test_fvtk_widgets.test_text_widget
  File "/Users/arokem/anaconda/envs/py2/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/arokem/anaconda/envs/py2/lib/python2.7/site-packages/numpy/testing/decorators.py", line 146, in skipper_func
    return f(*args, **kwargs)
  File "/Users/arokem/anaconda/envs/py2/lib/python2.7/site-packages/numpy/testing/decorators.py", line 146, in skipper_func
    return f(*args, **kwargs)
  File "/Users/arokem/source/dipy/dipy/viz/tests/test_fvtk_widgets.py", line 178, in test_text_widget
    npt.assert_equal(report.objects, 30)
  File "/Users/arokem/anaconda/envs/py2/lib/python2.7/site-packages/numpy/testing/utils.py", line 354, in assert_equal
    raise AssertionError(msg)
AssertionError: 
Items are not equal:
 ACTUAL: 29
 DESIRED: 30
-------------------- >> begin captured stdout << ---------------------
Data size is approximately 12KB
Dataset is already in place. If you want to fetch it again please first remove the folder /Users/arokem/.dipy/icons 

--------------------- >> end captured stdout << ----------------------

```

dipy/viz/actor.py
arokem: This one still hasn't been changed.

dipy/viz/window.py
```diff
@@ -0,0 +1,787 @@
...
+
+        """ Manages the visualization pipeline
+
+        Parameters
+        ----------
+        ren : Renderer() or vtkRenderer()
+            The scene that holds all the actors.
+        title : string
+            A string for the window title bar.
+        size : (int, int)
```
arokem: Default values for kwargs here.

dipy/viz/window.py
```diff
@@ -0,0 +1,787 @@
...
+
+def show(ren, title='DIPY', size=(300, 300),
+         png_magnify=1, reset_camera=True, order_transparent=False):
+    """ Show window with current renderer
+
+    Parameters
+    ------------
+    ren : Renderer() or vtkRenderer()
+        The scene that holds all the actors.
+    title : string
```
arokem: Default values for kwargs

dipy/viz/window.py
```diff
@@ -0,0 +1,787 @@
...
+    """ This will record a video of your scene
+
+    Records a video as a series of ``.png`` files of your scene by rotating the
+    azimuth angle az_angle in every frame.
+
+    Parameters
+    -----------
+    ren : vtkRenderer() object
+        as returned from function ren()
+    cam_pos : None or sequence (3,), optional
```
arokem: Default values for kwargs

dipy/viz/window.py
```diff
@@ -0,0 +1,787 @@
...
+
+
+def snapshot(ren, fname=None, size=(300, 300)):
+    """ Saves a snapshot of the renderer in a file or in memory
+
+    Parameters
+    -----------
+    ren : vtkRenderer
+        as returned from function renderer()
+    fname : str or None
```
arokem: Default values for kwargs

dipy/viz/window.py
```diff
@@ -0,0 +1,787 @@
...
+                     strel=None):
+    """ Analyze snapshot from memory or file
+
+    Parameters
+    ----------
+    im: str or array
+        If string then the image is read from a file otherwise the image is
+        read from a numpy array. The array is expected to be of shape (X, Y, 3)
+        where the last dimensions are the RGB values.
+    colors: tuple (3,) or list of tuples (3,)
```
arokem: Default values for kwargs

doc/api_changes.rst
```diff
@@ -6,6 +6,15 @@ Here we provide information about functions or classes that have been removed,
 renamed or are deprecated (not recommended) during different release circles.
 
 Dipy 0.10 Changes
+<<<<<<< HEAD
```
arokem: Whoops! This is what happens when you merge...

doc/examples/linear_fascicle_evaluation.py
```diff
@@ -63,13 +63,19 @@
...
 cc_ROI_actor = fvtk.contour(cc_slice, levels=[1], colors=[(1., 1., 0.)],
                             opacities=[1.])
-vol_actor = fvtk.slicer(t1_data, voxsz=(1.0, 1.0, 1.0), plane_i=[40],
-                        plane_j=None, plane_k=[35], outline=False)
+
+vol_actor = fvtk.slicer(t1_data)
+
+vol_actor.display(40, None, None)
+vol_actor2 = vol_actor.copy()
+vol_actor2.display(None, None, 35)
```
arokem: This seems cumbersome relative to the code that was previously here!

doc/examples/viz_slice.py
arokem: This is still a typo. Should be "sagittal"

dipy/viz/window.py
Garyfallidis: k

dipy/viz/window.py
Garyfallidis: k

dipy/viz/window.py
Garyfallidis: k

dipy/viz/window.py
Garyfallidis: k

doc/examples/linear_fascicle_evaluation.py
arokem: As a practical matter, maybe retain the old apone as well? Would that be possible?

arokem: *api

Garyfallidis: This new API for the slicer allows you to copy any slice from the data and move it as you like and wherever you like and that is how for example you can create easily mosaics etc. The previous function was simple but more rigid. We could if we want recreate the same function signature but it will be complicated to change the slices using a slider. So, I would argue to leave this as it is now. If you want to make your tutorial simpler. You can just keep one slice. You don't record the other one. 
So we could just write 
```
vol_actor = fvtk.slicer(t1_data).display(40, None, None)
```
and ignore the saggital slice.
Now after the release I think we should have a slider_widget object which will allow all 3 vertical slices and sliders to move the slices.
Let me know if you are fine to proceed with what we have now or remove the second slice from this tutorial.

arokem: How about an api that mimics the old api, by hiding these operations in the
module code? Could be a new function name, instead of slicer. Or you could
rename this new functionality, to retain backwards compatibility

Garyfallidis:
@jchoude thank you for your git suggestions.
@arokem there is nothing that should break the examples except from the slicer. This API change is included in the api_changes.rst and it is the only thing that it not backwards compatible.

doc/examples/linear_fascicle_evaluation.py
Garyfallidis: Doesn't worth it in this very specific case. The old slicer function was very limited. And even if we replace it with a new one of a different name the signature will be also different. So, no backwards compatibility again. If you want I can make a simple widget to show the 3 slices interactively. But this again will not be exactly backwards compatible because it will be a widget and not an actor. So, I would suggest to merge this PR as is and reflect in the next days if you really want this backwards compatibility with the old function. For me I prefer that this breaks at people's faces as soon as possible so that they start using the new much improved version.

Garyfallidis: We are still not even in version 1. We don't need to backwards support every single small thing.

arokem:
Ok. BTW - there's still some git conflict stuff further down in the api
changes file

Garyfallidis:
Sorry, for that. Will fix in a bit.

arokem:
NP! Keep up the awesome work here. This is some revolutionary stuff!

Garyfallidis:
Thanks @arokem! Indeed revolutionary :+1: 
Done here! Sorry for the merge issues. Will try to start using again rebase more and look into the latest in fancy merging. Will let you know if I find something cool related to that. Do merge this so we can start the next mental viz PRs!

arokem:
I'm still getting that test failure on my machine: 

```
FAIL: dipy.viz.tests.test_fvtk_widgets.test_text_widget
  File "/Users/arokem/anaconda/envs/py2/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/arokem/anaconda/envs/py2/lib/python2.7/site-packages/numpy/testing/decorators.py", line 146, in skipper_func
    return f(*args, **kwargs)
  File "/Users/arokem/anaconda/envs/py2/lib/python2.7/site-packages/numpy/testing/decorators.py", line 146, in skipper_func
    return f(*args, **kwargs)
  File "/Users/arokem/source/dipy/dipy/viz/tests/test_fvtk_widgets.py", line 177, in test_text_widget
    npt.assert_equal(report.objects, 3)
  File "/Users/arokem/anaconda/envs/py2/lib/python2.7/site-packages/numpy/testing/utils.py", line 354, in assert_equal
    raise AssertionError(msg)
AssertionError: 
Items are not equal:
 ACTUAL: 556
 DESIRED: 3
-------------------- >> begin captured stdout << ---------------------
Data size is approximately 12KB
Dataset is already in place. If you want to fetch it again please first remove the folder /Users/arokem/.dipy/icons 

--------------------- >> end captured stdout << ----------------------

```

arokem:
Would be good to get this up and running on travis. I'm losing my battle over on #732

Garyfallidis:
Hey brother, you need to show me how the snapshot array looks. imshow(arr, origin='lower'). And run it also once with interactive = True.
It's cool that we can now see the differences between GPUs and OSs with these new tests. Hard core!!! :dancers:

arokem:
Yep - very handy to be able to test this! 

Here's what I get: 
figure_1

Garyfallidis:
And what show is giving you?

arokem:
Sorry - not sure what you mean. I ran:

    import matplotlib.pyplot as plt
    plt.imshow(arr, origin='lower')
    plt.show()

What would you like me to run?

Garyfallidis:
window.show (our show)

Garyfallidis:
Change the interactive variable and press 's' to save the window when it appears.

arokem:
Strangely, the window didn't stick around enough for me to save it (it flashed on and then disappeared; even after setting `interactive=True`), but here's what I got with `window.record(renderer, out_path='/Users/arokem/Desktop/fig.png')`

fig

Garyfallidis:
But it window appeared. It was only the pressing 's' that didn't work. Right? Or it didn't appear at all?
I think there is a problem with tkinter and the filedialog in new macosx versions. What is your version?

arokem:
OSX 10.10.5

The window appeared briefly, and then disappeared, so there wasn't enough
time to press `s`.

The code I ran in the debugger was this:

    interactive = True

    window.show(renderer)

I have no issues on current master (e.g., windows persist, although they
sometimes get sent to the background, instead of appearing in the fore)

Garyfallidis:
I am disabling the default mesa usage for snapshot. Let see if that solves your snapshot issue. Still looking on the other issue.

Garyfallidis:
Okay some feedback for the window.show issue. It seems is a backend problem in osx.
@arokem can you run the test with ipython
```
ipython --gui=qt -i test_fvtk_window.py
```
And tell me if you see the window. The window should open but it may not necessarily catch the focus. So, it could go to the background.

arokem:
The windows appear briefly and then close. They're no-where to be found in
the background...

Garyfallidis:
Even running them from ipython?

arokem:
Yep - running them using the command you mentioned above.

Garyfallidis:
Damn! Time for extreme measures.

Garyfallidis:
Do you have pyqt installed in your machine?

arokem:
Yes, I do.

Garyfallidis:
Okay expected. I will get back to you asap.

arokem:
Let's also merge #732 and rebase this one on top of it, before we merge
this one.

Garyfallidis:
Sure, in the meantime can you file an issue in vtk (or write an e-mail) to the vtk mailing list saying that the window disappears in your osx version and not in linux? They should find a way to fix this at some point. Or maybe there is already some hack.

arokem:
Yeah. Regardless, there's still the test failure mentioned above, though.
What do you think about that? Seems unrelated to the window disappearing,
unless I am missing something.

Garyfallidis:
You mean the filedialogue issue?

Garyfallidis:
Please cc me when you write the e-mail to vtk.

arokem:
This one: 

```
FAIL: dipy.viz.tests.test_fvtk_widgets.test_text_widget
  File "/Users/arokem/anaconda/envs/py2/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/arokem/anaconda/envs/py2/lib/python2.7/site-packages/numpy/testing/decorators.py", line 146, in skipper_func
    return f(*args, **kwargs)
  File "/Users/arokem/anaconda/envs/py2/lib/python2.7/site-packages/numpy/testing/decorators.py", line 146, in skipper_func
    return f(*args, **kwargs)
  File "/Users/arokem/source/dipy/dipy/viz/tests/test_fvtk_widgets.py", line 177, in test_text_widget
    npt.assert_equal(report.objects, 3)
  File "/Users/arokem/anaconda/envs/py2/lib/python2.7/site-packages/numpy/testing/utils.py", line 354, in assert_equal
    raise AssertionError(msg)
AssertionError: 
Items are not equal:
 ACTUAL: 556
 DESIRED: 3
-------------------- >> begin captured stdout << ---------------------
Data size is approximately 12KB
Dataset is already in place. If you want to fetch it again please first remove the folder /Users/arokem/.dipy/icons 

--------------------- >> end captured stdout << ----------------------

```

Garyfallidis:
Ah yeah. This is strange too. We didn't have such an issue in the other osx that I used today. Can you paste here the figure that comes out from that snapshot? As we did before (using imshow(...)).

arokem:
That is the one I posted before.

Garyfallidis:
Okay so it is the same? Even after I removed the Mesa option in one of my last commits.

arokem:
Yeah - still looks the same.

Garyfallidis:
Okay for the snapshot error I am not worried much because the plan is to have these tests running for Linux for now. To go multi-platform with offscreen testing and widgets needs more work. 
But to fix the show problem that you have is important. And I hope we can do something about that.

Garyfallidis:
What is also strange is that with the current master you have no problem seeing the window.

arokem:
Or when I run the examples from this branch!

arokem:
I wouldn't worry about the show issue and focus on that test

Garyfallidis:
Can you hangout for a bit?

Garyfallidis:
I am not worried about the test failing because it works when you use fvtk.record so we could call fvtk.record from snapshot when os is darwin (osx). The problem with record is that it will start throwing up windows at your face. So, with snapshot I was trying to go completely offscreen. But maybe it is not possible in all osx versions.

arokem:
Cannot hangout - I am doing this in between talks at a symposium I am
attending :-)

But it sounds like we're basically ready to go, once you merge #732 and
rebase on top of it.

Garyfallidis:
Okay np. Tomorrow then.

arokem:
Still no tkinter! Strange.

arokem:
This https://github.com/travis-ci/travis-ci/issues/1654 seems to suggest that this should have been enough!

arokem:
Wait - should these imports be 'import Tkinter', with a capital 'T'? http://stackoverflow.com/questions/7498658/importerror-when-importing-tkinter-in-python

Garyfallidis:
Isn't this what I am doing here? 
https://github.com/nipy/dipy/pull/587/files#diff-802fef6887a85cd6e23c011cb8bd1b72R10

arokem:
Sorry - yes - it looks like you are. But it's still throwing that import
error...

It's not a general problem though. Travis *can* import Tkinter in
principle: https://travis-ci.org/arokem/dipy/builds/85764045

Maybe it's also something about the path settings in the venv? Like on #732?

Garyfallidis:
Yes, that is more likely. Can you do some magic about that?

arokem:
abracadabra!

arokem:
Let me see if I can make sense of it. The annoying thing about debugging
Travis is that you have to wait for a while to know whether what you did
actually helped. It's a bit mind-melting at times.

Garyfallidis:
True story! But when we have this ready and working it will be a real milestone. I think we are really close to that happening at least for the Linux bots. Keep it up and thanks for all the great help.

arokem:
Could you please rebase this on master? It will help me trying out things on travis

arokem:
I see that some of the fvtk tests depend on Pillow. Should we install that on the build-bot as well?

Garyfallidis:
Yes.  Let me know when you find some time to move ahead with the release.

arokem:
I think this is the last major block to cutting the release, but I am a bit stumped. Do you know anything about this error: https://travis-ci.org/arokem/dipy/jobs/86520597#L2426

By the way - installing Pillow did not seem to resolve this.

Garyfallidis:
Somewhere in a forum I found that for this problem
vtkXOpenGLRenderWindow (0x2d2e410): Could not find a decent visual
You need to set a variable 
export LIBGL_ALWAYS_INDIRECT=y
Here is some information about what this variable is doing
http://unix.stackexchange.com/questions/1437/what-does-libgl-always-indirect-1-actually-do

arokem:
That doesn't seem to work for us:
https://github.com/arokem/dipy/commit/ecdc69e06b603755f655934dca68489efd25cdc1

Garyfallidis:
Yes, okay. You are correct! Let's work together on this later tonight. Need to catch up with something else for now.

arokem:
https://travis-ci.org/arokem/dipy/jobs/86322236

arokem:
OK - actually now revisiting that commit - I might have an idea. Let me try
this again. Let's catch up on this later!

Garyfallidis:
@arokem just to make sure here. On top of which branch do I need to rebase this one? Is it #752? Because I see some vtk tests are skipped there. But maybe this is normal for that version of the code. Sorry in advance for the naivety of the question, it's getting late here and I had a long day. :) But hey no excuses! Let's get the new viz in!

arokem:
Yeah - those are other test-conditions that are causing those to be
skipped. I think this is about importing the vtk colors module. Maybe we
can take that separately?

arokem:
Oh no - wait - you're right. There's one: "No VTK for these tests" in there
as well.

Let me double check again...

arokem:
OK - I am about ready to say that we can't test some of these things on
Travis. I propose setting this to skip on Travis.

If you rebase this on master, I can make a PR against your branch with the
proposed changes to make that happen.

arokem:
Congratulations @Garyfallidis! A new era of diffusion MRI visualization!

Garyfallidis:
Oh man it has been a looooong trip! I am so pleased :) Things are getting exciting! @MarcCote first dipy.viz PR is in!!! You know exactly what that means! @arokem many thx for the great help on setting up VTK  on mad Travis. :dancers:

Garyfallidis:
Let's start making some smaller viz PRs from now on. :+1: This one was a bit too big.

arokem:
+1

PR #588, opened 2015-03-13
Nf csd calibration

MrBago:
I re-based #404 to clean it up a little bit. The changes I made are:

1) Took out new functions in `dipy.reconst.shm`, these functions were not being called or tested. (Did this via rebase).
2) I re-worked a little bit how the SH response was being handled. Only axially symmetric SH functions are needed so we don't bother to fit all the SH functions and zero out the non-symmetric coefficients. This should improve accuracy and performance.
3) Changed recursive_response to use `default_sphere`. 
4) Fixed issue with response_scaling being wrong when an SH response was used.

@arokem  @ChantalTax can you guys look over this.

### DIFF
--- a/dipy/reconst/csdeconv.py
+++ b/dipy/reconst/csdeconv.py
@@ -843,7 +843,7 @@ def auto_response(gtab, data, roi_center=None, roi_radius=10, fa_thr=0.7,
 
 def recursive_response(gtab, data, mask=None, sh_order=8, peak_thr=0.01,
                        init_fa=0.08, init_trace=0.0021, iter=8,
-                       convergence=0.001, parallel=True,
+                       convergence=0.001, parallel=True, nbr_processes=None,
                        sphere=default_sphere):
     """ Recursive calibration of response function using peak threshold
 
@@ -873,6 +873,9 @@ def recursive_response(gtab, data, mask=None, sh_order=8, peak_thr=0.01,
     parallel : bool, optional
         Whether to use parallelization in peak-finding during the calibration
         procedure. Default: True
+    nbr_processes: int
+        If `parallel` is True, the number of subprocesses to use
+        (default multiprocessing.cpu_count()).
     sphere : Sphere, optional.
         The sphere used for peak finding. Default: default_sphere.
     
@@ -919,7 +922,8 @@ def recursive_response(gtab, data, mask=None, sh_order=8, peak_thr=0.01,
                                      sphere=sphere,
                                      relative_peak_threshold=peak_thr,
                                      min_separation_angle=25,
-                                     parallel=parallel)
+                                     parallel=parallel,
+                                     nbr_processes=nbr_processes)
 
         dirs = csd_peaks.peak_dirs
         vals = csd_peaks.peak_values
--- a/dipy/reconst/csdeconv.py
+++ b/dipy/reconst/csdeconv.py
@@ -8,23 +8,22 @@ import scipy.linalg as la
 import scipy.linalg.lapack as ll
 
 from dipy.data import small_sphere, get_sphere
+from dipy.reconst.multi_voxel import multi_voxel_fit
+from dipy.reconst.shm import (sph_harm_ind_list, real_sph_harm,
+                              order_from_ncoef, sph_harm_lookup, lazy_index,
+                              SphHarmFit, real_sym_sh_basis, sh_to_rh,
+                              gen_dirac, forward_sdeconv_mat, SphHarmModel,
+                              sh_to_sf)
+from dipy.data import get_sphere
 from dipy.core.geometry import cart2sphere
 from dipy.core.ndindex import ndindex
 from dipy.sims.voxel import single_tensor
 from dipy.utils.six.moves import range
 
-from dipy.reconst.multi_voxel import multi_voxel_fit
 from dipy.reconst.dti import TensorModel, fractional_anisotropy
-from dipy.reconst.shm import (sph_harm_ind_list, real_sph_harm,
-                              sph_harm_lookup, lazy_index, SphHarmFit,
-                              real_sym_sh_basis, sh_to_rh, forward_sdeconv_mat,
-                              SphHarmModel)
 
 from dipy.reconst.peaks import peaks_from_model
 from dipy.core.geometry import vec2vec_rotmat
-from dipy.reconst.shm import sh_to_sf
-from dipy.viz import fvtk
-from dipy.sims.voxel import single_tensor_odf
 
 
 class ConstrainedSphericalDeconvModel(SphHarmModel):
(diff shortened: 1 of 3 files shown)


dipy/reconst/csdeconv.py
```diff
@@ -47,7 +62,7 @@ def __init__(self, gtab, response, reg_sphere=None, sh_order=8, lambda_=1,
             ndarray and the second is the signal value for the response
             function without diffusion weighting.  This is to be able to
             generate a single fiber synthetic signal. The response function
-            will be used as deconvolution kernel ([1]_)
+            will be used as deconvolution git pull nipy-dipy masterkernel ([1]_)
```
arokem: This is a misplaced copy-paste from rebasing. Oops

dipy/reconst/csdeconv.py
```diff
@@ -798,3 +817,117 @@ def auto_response(gtab, data, roi_center=None, roi_radius=10, fa_thr=0.7,
...
+                                     data=data,
+                                     sphere=sphere,
+                                     relative_peak_threshold=peak_thr,
+                                     min_separation_angle=25,
+                                     parallel=parallel)
+
+        dirs = csd_peaks.peak_dirs
+        vals = csd_peaks.peak_values
+        single_peak_mask = (vals[:, 1] / vals[:, 0]) < peak_thr
+        data = data[single_peak_mask]
```
MrBago: @ChantalTax here data can only get small, is this by design? Maybe it should be something like

```
csd_peaks = peaks_from_model(..., data=data_all, ...)
...
data = data_all[single_peak_mask]
```

If this is by design, that's fine but I just wanted to check with you.

dipy/reconst/csdeconv.py
MrBago: Tnx! fixed it!

dipy/reconst/csdeconv.py
ChantalTax: @MrBago indeed data gets smaller at each iteration, we only want to keep voxels that exhibit a single significant peak after CSD with the previous improved RF.

arokem:
This all looks good to me. Should be ready to go. I've added a couple of small fixes to the example, so once @MrBago merges that (https://github.com/MrBago/dipy/pull/14), I will merge this.

dipy/reconst/csdeconv.py
```diff
@@ -798,3 +839,114 @@ def auto_response(gtab, data, roi_center=None, roi_radius=10, fa_thr=0.7,
...
+        peak in order to call it a single fiber population [1]
+    init_fa : float
+        FA of the initial 'fat' response function (tensor)
+    init_trace : float
+        trace of the initial 'fat' response function (tensor)
+    iter : int
+        maximum number of iterations for calibration
+    convergence : float
+        convergence criterion, maximum relative change of SH coefficients
+    sphere : Sphere
```
samuelstjean: This seems to be missing the parallel arg explanation.

dipy/reconst/csdeconv.py
```diff
@@ -798,3 +839,114 @@ def auto_response(gtab, data, roi_center=None, roi_radius=10, fa_thr=0.7,
...
+    else:
+        data = data[mask]
+
+    n = np.arange(0, sh_order + 1, 2)
+    where_dwi = lazy_index(~gtab.b0s_mask)
+    response_p = np.ones(len(n))
+
+    for num_it in range(1, iter):
+        r_sh_all = np.zeros(len(n))
+        csd_model = ConstrainedSphericalDeconvModel(gtab, res_obj,
```
samuelstjean: Does the model actually needs to be re-instantiated at each iteration, since only the peaks finding part changes during those iterations?

arokem: Good catch. Apparently not.

dipy/reconst/csdeconv.py
samuelstjean: It could be also useful to specify the number of cores to use, as peaks_from_model can consider the number of processes to launch.

dipy/reconst/csdeconv.py
arokem: https://github.com/MrBago/dipy/pull/14

dipy/reconst/csdeconv.py
arokem: OK. Did that too, including spelling out default values. See: https://github.com/MrBago/dipy/pull/14

arokem: Yup.

dipy/reconst/csdeconv.py
ChantalTax: I think it does, because the response function res_obj changes in every iteration? And peaks_from_model uses csd_model?

arokem: Hmm - what you say makes sense (promptly putting that back). I will say
that the tests pass just fine when I move this object outside the loop...

samuelstjean: So, either a test is lacking, or something in the CSD model is terribly broken (I realy hope the former though). Good thing Chantal re-read my suggestion, I don't want to accidentally break your stuff.

ChantalTax: The test filters out the single fiber populations already in the first iteration, and if you keep deconvolving with the initial response function, you will keep finding the same response function because you average only the two single fiber voxels. That is why the test works here, I guess. It is just a really simple example, we could add more complicated ones like in our paper in a next PR, if this is desirable.

arokem: Yes - that would probably be good. For now, I think that we understand why
this is the case. I was just looking at the test, and it does look to be
rather rigorous, but because it's based on a simulation, it always hits
things on the head before iterating further, so maybe something a bit more
comprehensive (that needs to actually iterate) might be good.

Since you are already planning a follow up to polish the documentation up,
we can do that there.

ChantalTax: That is a fair point. The reason why I added this here, is that you ideally want to mimic a bit what is present in a real brain, which means that we probably have to make a (significantly) larger test dataset and that the test will take longer. But I defenitely agree that this would be beneficial.

dipy/reconst/csdeconv.py
MrBago: Ya ..., the way that we're handling this parallel stuff is not very good. We really should redesign the parallel stuff for all of dipy. No functions/classes should be explicitly using `parallel` or `number_proc` keywords.

MrBago:
@ChantalTax Thanks for doing this and putting up with all our nagging. Could you look over the 4 commits that @arokem and I have added, I'd like to have your official go ahead before we merge this PR. I'm sorry I made it harder for you to review by re-basing your commits.

ChantalTax:
@MrBago I had a look at the code and it looks fine to me. I guess you tested the changes? I fetched your branch to look at the real data example fODFs (this often gives a pretty good indication if something weird is happening), however I do not seem able to run it: while I was rebasing with @arokem it gave some errors on the file "streamlinespeed", and this is still happening. I would say you can go ahead and we solve this later? Good work!

arokem:
Yes - I just ran this branch on the example, and it all looks good to me.

For what it's worth, the goodness-of-fit (assessed with R^2) to the data is
better in the example data with recursive calibration, but we'll still need
to do some kind of a cross-validation study to verify that it's not just
due to overfitting to the specific characteristics of this sample.

I'm going to go ahead and merge this.

arokem:
Congratulations @ChantalTax on the good work!

PR #594, opened 2015-03-15
DOC + PEP8: Mostly just line-wrapping.

arokem:
And a few small edits on the csd example.

### DIFF
--- a/doc/examples/reconst_csd.py
+++ b/doc/examples/reconst_csd.py
@@ -74,7 +74,7 @@ print(ratio)
 0.21197
 
 We can double-check that we have a good response function by visualizing the
-response's function's ODF. Here is how you would do that:
+response function's ODF. Here is how you would do that:
 """
 
 from dipy.viz import fvtk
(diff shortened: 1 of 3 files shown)


doc/examples/reconst_csd.py
```diff
@@ -70,8 +73,8 @@
 """
 0.21197
 
-We can double check that we have a good response function by visualizing the
-response's function's ODF. Here is how:
+We can double-check that we have a good response function by visualizing the
+response's function's ODF. Here is how you would do that:
```
Garyfallidis: typo: response function's 
And I would remove all from Here ... 
I think it is redundant

doc/examples/reconst_csd.py
```diff
@@ -98,18 +101,17 @@
 fvtk.rm(ren, response_actor)
 
 """
-However, using an FA threshold is not a very robust method. It is dependent on
-the dataset (non-informed used subjectivity), and still depends on the
-diffusion tensor (FA and first eigenvector), which has low accuracy at high
-b-value. Alternatively, one can calibrate the response function directly from
-the data according to [Tax2014]_. 
+Depending on the dataset, FA threshold may not be the best way to divine the
```
Garyfallidis: divine?

arokem: From the Oxford English Dictionary:


"Divine, v. : 2. To make out by sagacity, intuition, or fortunate
conjecture (that is, in some other way than by actual information); to
conjecture, guess."

Garyfallidis: Keep it simple. This use of the word is very rare afaik. 
Even guess is too vague here. 

I would say: find the best possible response function.
Your call!

doc/examples/reconst_csd.py
arokem: Done

doc/examples/reconst_csd.py
arokem: done

PR #595, opened 2015-03-15
BF: csd predict multi

arokem:
Prediction from arrays of models was failing. Specifically, for multi-voxel fits, we need to provide an array of S0, which matches the 3D shape of the data. In the previous implementation, this silently added multiple dimensions to the array. The solution proposed here is to push the prediction in the `MultiVoxelFit` object, so that it calls out to each individual item's prediction, with the right S0 taken from the inputs (or defaulting to the right thing - all 1's).

### DIFF
--- a/dipy/reconst/multi_voxel.py
+++ b/dipy/reconst/multi_voxel.py
@@ -61,6 +61,33 @@ class MultiVoxelFit(ReconstFit):
         else:
             return item
 
+    def predict(self, *args, **kwargs):
+        """
+        Predict for the multi-voxel object using each single-object's
+        prediction API, with S0 provided from an array.
+        """
+        if not hasattr(self.model, 'predict'):
+            msg = "This model does not have prediction implemented yet"
+            raise NotImplementedError(msg)
+
+        S0 = kwargs.get('S0', np.ones(self.fit_array.shape))
+        idx = ndindex(self.fit_array.shape)
+        ijk = idx.next()
+        def gimme_S0(S0, ijk):
+            if isinstance(S0, np.ndarray):
+                return S0[ijk]
+            else:
+                return S0
+
+        kwargs['S0'] = gimme_S0(S0, ijk)
+        first_pred = self.fit_array[ijk].predict(*args, **kwargs)
+        result = np.empty(self.fit_array.shape + (first_pred.shape[-1],))
+        result[ijk] = first_pred
+        for ijk in idx:
+            kwargs['S0'] = gimme_S0(S0, ijk)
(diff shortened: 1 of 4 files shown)


dipy/reconst/tests/test_csdeconv.py
```diff
@@ -386,6 +386,29 @@ def test_csd_predict():
...
+    multi_S = np.array([[S, S], [S, S]])
+    csd_fit_multi = csd.fit(multi_S)
+    S0_multi = np.mean(multi_S[..., gtab.b0s_mask], -1)
+    pred_multi = csd_fit_multi.predict(S0=S0_multi)
+    npt.assert_array_almost_equal(pred_multi, multi_S)
+
+
 def test_sphere_scaling_csdmodel():
     """Check that mirroring regulization sphere does not change the result of
     csddeconv model"""
```
samuelstjean: regulization is mistyped, and csddeconv is probably meant to be csd or csdeconv, since the d in CSD stands for deconv already.

arokem: That's a good catch, but actually not part of this PR at all. It would be good to do a documentation clean up in a separate PR, though.

samuelstjean: That's funny, good thing they show lines before and after, I always see random stuff like that for some reason.

arokem: https://github.com/nipy/dipy/pull/596

MrBago:
+1, should I merge?

arokem:
It's ready from my point of view.

Garyfallidis:
Sorry, a quick question here. Does this allows for all models with the multi_voxel_fit decorator to have a predict method?

Garyfallidis:
If I understand correctly if a Model has a predict method then it will be used? And if not nothing happens. Am I correct?

dipy/reconst/multi_voxel.py
```diff
@@ -61,6 +61,33 @@ def __getitem__(self, index):
...
             return item
 
+    def predict(self, *args, **kwargs):
+        """
+        Predict for the multi-voxel object using each single-object's
+        prediction API, with S0 provided from an array.
+        """
+        if not hasattr(self.model, 'predict'):
+            msg = "This model does not have prediction implemented yet"
+            raise NotImplementedError(msg)
```
MrBago: @Garyfallidis If a model does not have a predict method a NotImplementedError will be raised here.

Garyfallidis: okay great

PR #598, opened 2015-03-16
Fixed undetected compilation errors prior to Cython 0.22

### LINKED ISSUE #597: tracking module not building on cython 0.22
Error compiling Cython file:
------------------------------------------------------------
...
        def __get__(Feature self):
            return bool(self.is_order_invariant)
        def __set__(self, int value):
            self.is_order_invariant = bool(value)

    cdef Shape c_infer_shape(Feature self, Data2D datum) nogil:
        ^
------------------------------------------------------------

dipy/segment/featurespeed.pyx:39:9: Signature not compatible with previous declaration

Error compiling Cython file:
------------------------------------------------------------
...

cdef class Feature(object):
    cdef int is_order_invariant

    cdef Shape c_infer_shape(Feature self, Data2D datum) nogil except *
                           ^
------------------------------------------------------------

dipy/segment/featurespeed.pxd:7:28: Previous declaration is here

+++++++++++++++++

Seems like those except * are causing trouble for whatever reason. My fork has problems with the tracking module (while it has always been working before of course), but the latest master has a ton more issues. No buildbot error or anything else for anyone?

MarcCote:
Fixes compilation errors in Cython code.

Cython's version 0.22 fixes the following bug "* Mismatching 'except' declarations on signatures in .pxd and .pyx files failed to produce a compile error".

Closes #597

### DIFF
--- a/dipy/segment/metricspeed.pyx
+++ b/dipy/segment/metricspeed.pyx
@@ -44,12 +44,12 @@ cdef class Metric(object):
         def __get__(Metric self):
             return bool(self.is_order_invariant)
 
-    cdef int c_are_compatible(Metric self, Shape shape1, Shape shape2) nogil:
+    cdef int c_are_compatible(Metric self, Shape shape1, Shape shape2) nogil except -1:
         """ Cython version of `Metric.are_compatible`. """
         with gil:
             return self.are_compatible(shape2tuple(shape1), shape2tuple(shape2))
 
-    cdef double c_dist(Metric self, Data2D features1, Data2D features2) nogil:
+    cdef double c_dist(Metric self, Data2D features1, Data2D features2) nogil except -1.0:
         """ Cython version of `Metric.dist`. """
         with gil:
             return self.dist(np.asarray(features1), np.asarray(features2))
@@ -74,7 +74,7 @@ cdef class Metric(object):
         """
         raise NotImplementedError("Metric's subclasses must implement method `are_compatible(self, shape1, shape2)`!")
 
-    cpdef double dist(Metric self, features1, features2):
+    cpdef double dist(Metric self, features1, features2) except -1.0:
         """ Computes a distance between two data points based on their features.
 
         Parameters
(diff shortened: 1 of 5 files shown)


dipy/segment/metricspeed.pyx
```diff
@@ -44,12 +44,12 @@ cdef class Metric(object):
...
             return bool(self.is_order_invariant)
 
-    cdef int c_are_compatible(Metric self, Shape shape1, Shape shape2) nogil:
+    cdef int c_are_compatible(Metric self, Shape shape1, Shape shape2) nogil except -1:
         """ Cython version of `Metric.are_compatible`. """
         with gil:
             return self.are_compatible(shape2tuple(shape1), shape2tuple(shape2))
 
-    cdef double c_dist(Metric self, Data2D features1, Data2D features2) nogil:
+    cdef double c_dist(Metric self, Data2D features1, Data2D features2) nogil except -1.0:
```
matthew-brett: Does this really need to be a float?

MarcCote: I'm not sure. I think it could be a int but will be converted in double anyway.
http://docs.cython.org/src/userguide/language_basics.html#error-return-values

MarcCote: Ok, just tested it, it can be a int. Do you want me to make the change everywhere?

arokem: BTW - ran a try_branch with this:
http://nipy.bic.berkeley.edu/builders/dipy-py2.7-osx-10.8/builds/336

:-)

matthew-brett: Yes, please (convert to int).  It makes it a bit easier to read, because at the moment I'm thinking 'what is the float for' when the answer is just the usual signal that there is an error.

Good job on the 'try_branch' :)

MarcCote:
Travis is not complaining, should be ready to merge.

Garyfallidis:
Thx @MarcCote. Let's hear what the buildbots are saying.

PR #599, opened 2015-03-16
BF: use dpy_rint instead of round for Windows

matthew-brett:
Windows complains for lack of the `round` symbol; use dpy_rint instead
(implemented in the dpy_math.h header).


matthew-brett:
http://nipy.bic.berkeley.edu/builders/dipy-py2.6-32/builds/510

PR #612, opened 2015-03-20
BF: Differences in spherical harmonic calculations wrt scipy 0.15

arokem:
Addresses https://github.com/nipy/dipy/issues/611 ?

### DIFF
--- a/dipy/reconst/shm.py
+++ b/dipy/reconst/shm.py
@@ -24,21 +24,21 @@ keep the data as is and implement the relevant equations rewritten in the
 following form: Y.T = x.T B.T, or in python syntax data = np.dot(sh_coef, B.T)
 where data is Y.T and sh_coef is x.T.
 """
+from functools import wraps
+
 import numpy as np
 from numpy import concatenate, diag, diff, empty, eye, sqrt, unique, dot
 from numpy.linalg import pinv, svd
 from numpy.random import randint
+
 from dipy.reconst.odf import OdfModel, OdfFit
-from scipy.special import lpn, lpmv, gammaln
-from dipy.core.sphere import Sphere
-import dipy.core.gradients as grad
-from dipy.sims.voxel import single_tensor, all_tensor_evecs
 from dipy.core.geometry import cart2sphere
 from dipy.core.onetime import auto_attr
 from dipy.reconst.cache import Cache
 
 from distutils.version import LooseVersion
 import scipy
+from scipy.special import lpn, lpmv, gammaln
 
 if LooseVersion(scipy.version.short_version) >= LooseVersion('0.15.0'):
     SCIPY_15_PLUS = True
@@ -186,14 +186,19 @@ def spherical_harmonics(m, n, theta, phi):
     scipy version < 0.15.0.
 
     """
-    if SCIPY_15_PLUS:
-        return sph_harm(m, n, theta, phi, dtype=complex)
     x = np.cos(phi)
     val = lpmv(m, n, x).astype(complex)
     val *= np.sqrt((2 * n + 1) / 4.0 / np.pi)
     val *= np.exp(0.5 * (gammaln(n - m + 1) - gammaln(n + m + 1)))
     val = val * np.exp(1j * m * theta)
     return val
+
+if SCIPY_15_PLUS:
+    @wraps(sph_harm)
+    def spherical_harmonics(m, n, theta, phi, dtype=complex):
+        """Dipy wraps this function because of a bug in scipy .15 that causes
+        sph_harm to return a complex64 type instead of a complex128"""
+        return sph_harm(m, n, theta, phi, dtype=dtype)
 
 
 def real_sph_harm(m, n, theta, phi):
--- a/dipy/reconst/shm.py
+++ b/dipy/reconst/shm.py
@@ -24,7 +24,6 @@ keep the data as is and implement the relevant equations rewritten in the
 following form: Y.T = x.T B.T, or in python syntax data = np.dot(sh_coef, B.T)
 where data is Y.T and sh_coef is x.T.
 """
-from functools import wraps
 
 import numpy as np
 from numpy import concatenate, diag, diff, empty, eye, sqrt, unique, dot
@@ -186,6 +185,8 @@ def spherical_harmonics(m, n, theta, phi):
     scipy version < 0.15.0.
 
     """
+    if SCIPY_15_PLUS:
+        return sph_harm(m, n, theta, phi, dtype=complex)
     x = np.cos(phi)
     val = lpmv(m, n, x).astype(complex)
     val *= np.sqrt((2 * n + 1) / 4.0 / np.pi)
@@ -193,12 +194,6 @@ def spherical_harmonics(m, n, theta, phi):
     val = val * np.exp(1j * m * theta)
     return val
 
-if SCIPY_15_PLUS:
-    @wraps(sph_harm)
-    def spherical_harmonics(m, n, theta, phi, dtype=complex):
-        """Dipy wraps this function because of a bug in scipy .15 that causes
-        sph_harm to return a complex64 type instead of a complex128"""
-        return sph_harm(m, n, theta, phi, dtype=dtype)
 
 
 def real_sph_harm(m, n, theta, phi):
--- a/dipy/reconst/shm.py
+++ b/dipy/reconst/shm.py
@@ -29,7 +29,7 @@ from numpy import concatenate, diag, diff, empty, eye, sqrt, unique, dot
 from numpy.linalg import pinv, svd
 from numpy.random import randint
 from dipy.reconst.odf import OdfModel, OdfFit
-from scipy.special import sph_harm, lpn, lpmv, gammaln
+from scipy.special import lpn, lpmv, gammaln
 from dipy.core.sphere import Sphere
 import dipy.core.gradients as grad
 from dipy.sims.voxel import single_tensor, all_tensor_evecs
@@ -185,8 +185,6 @@ def spherical_harmonics(m, n, theta, phi):
     scipy version < 0.15.0.
 
     """
-    if SCIPY_15_PLUS:
-        return sph_harm(m, n, theta, phi)
     x = np.cos(phi)
     val = lpmv(m, n, x).astype(complex)
     val *= np.sqrt((2 * n + 1) / 4.0 / np.pi)
--- a/dipy/reconst/shm.py
+++ b/dipy/reconst/shm.py
@@ -42,6 +42,7 @@ import scipy
 
 if LooseVersion(scipy.version.short_version) >= LooseVersion('0.15.0'):
     SCIPY_15_PLUS = True
+    from scipy.special import sph_harm
 else:
     SCIPY_15_PLUS = False
 
--- a/dipy/reconst/shm.py
+++ b/dipy/reconst/shm.py
@@ -185,6 +185,8 @@ def spherical_harmonics(m, n, theta, phi):
     scipy version < 0.15.0.
 
     """
+    if SCIPY_15_PLUS:
+        return sph_harm(m, n, theta, phi, dtype=complex)
     x = np.cos(phi)
     val = lpmv(m, n, x).astype(complex)
     val *= np.sqrt((2 * n + 1) / 4.0 / np.pi)
--- a/dipy/reconst/shm.py
+++ b/dipy/reconst/shm.py
@@ -41,7 +41,7 @@ from scipy.special import lpn, lpmv, gammaln
 
 if LooseVersion(scipy.version.short_version) >= LooseVersion('0.15.0'):
     SCIPY_15_PLUS = True
-    from scipy.special import sph_harm
+    import scipy.special as sps
 else:
     SCIPY_15_PLUS = False
 
@@ -158,7 +158,18 @@ def gen_dirac(m, n, theta, phi):
 
 
 def spherical_harmonics(m, n, theta, phi):
-    r""" Compute spherical harmonics
+    x = np.cos(phi)
+    val = lpmv(m, n, x).astype(complex)
+    val *= np.sqrt((2 * n + 1) / 4.0 / np.pi)
+    val *= np.exp(0.5 * (gammaln(n - m + 1) - gammaln(n + m + 1)))
+    val = val * np.exp(1j * m * theta)
+    return val
+
+if SCIPY_15_PLUS:
+    def spherical_harmonics(m, n, theta, phi):
+        return sps.sph_harm(m, n, theta, phi, dtype=complex)
+
+spherical_harmonics.__doc__ = r""" Compute spherical harmonics
 
     This may take scalar or array arguments. The inputs will be broadcasted
     against each other.
@@ -182,18 +193,9 @@ def spherical_harmonics(m, n, theta, phi):
     Notes
     -----
     This is a faster implementation of scipy.special.sph_harm for
-    scipy version < 0.15.0.
-
+    scipy version < 0.15.0. For scipy 0.15 and onwards, we use the scipy
+    implementation of the function
     """
-    if SCIPY_15_PLUS:
-        return sph_harm(m, n, theta, phi, dtype=complex)
-    x = np.cos(phi)
-    val = lpmv(m, n, x).astype(complex)
-    val *= np.sqrt((2 * n + 1) / 4.0 / np.pi)
-    val *= np.exp(0.5 * (gammaln(n - m + 1) - gammaln(n + m + 1)))
-    val = val * np.exp(1j * m * theta)
-    return val
-
 
 
 def real_sph_harm(m, n, theta, phi):
--- a/dipy/reconst/tests/test_csdeconv.py
+++ b/dipy/reconst/tests/test_csdeconv.py
@@ -451,8 +451,7 @@ def test_default_lambda_csdmodel():
                                                      sh_order=sh_order,
                                                      reg_sphere=sphere)
         B_reg, _, _ = real_sym_sh_basis(sh_order, sphere.theta, sphere.phi)
-        npt.assert_array_almost_equal(model_full.B_reg, expected * B_reg,
-                                      decimal=4)
+        npt.assert_array_almost_equal(model_full.B_reg, expected * B_reg)
 
 
 def test_csd_superres():
--- a/dipy/reconst/tests/test_csdeconv.py
+++ b/dipy/reconst/tests/test_csdeconv.py
@@ -435,7 +435,7 @@ def test_default_lambda_csdmodel():
     the symmetric362 sphere. This value has empirically been found to work well
     and changes to this default value should be discussed with the dipy team.
     """
-    sphere = get_sphere('symmetric362')
+    sphere = default_sphere
 
     # Create gradient table
     _, fbvals, fbvecs = get_data('small_64D')
@@ -451,7 +451,8 @@ def test_default_lambda_csdmodel():
                                                      sh_order=sh_order,
                                                      reg_sphere=sphere)
         B_reg, _, _ = real_sym_sh_basis(sh_order, sphere.theta, sphere.phi)
-        npt.assert_array_almost_equal(model_full.B_reg, expected * B_reg)
+        npt.assert_array_almost_equal(model_full.B_reg, expected * B_reg,
+                                      decimal=4)
 
 
 def test_csd_superres():


MrBago:
This isn't a precision issue, it doesn't match because you changed the sphere :). This test is meant to ensure that in the process of updating the CSD model, we didn't change the default values of lambda and such.

If you use a different sphere these values will change a little, which is normal, but the precomputed values were created using "symmetric362".

matthew-brett:
Canary in coal-mine :)​

arokem:
Well, to continue with the animal metaphors, the selection of sphere turns out to be a red herring (but it's my bad for mixing it in): the test passes on current master with python 2.7 regardless of selection of sphere in this test, and fails on python3.4, also regardless of the sphere (symmetric362 or default_sphere alike).

MrBago:
It means the lambda calculation is (slightly) different between python2.7 and python3.4. Might not be a big deal, but I'd feel better if we tracked it down. Is this happening for all SH orders or only 16?

arokem:
Happens both with 8 and 16

MrBago:
K, will look into this when I have a chance. This PR is fine, but as a separate issue I'd like to know why lambdas are different with python versions.

matthew-brett:
Fails on everything with commit f35702c (0.9.2).  Could this be a numpy version problem?

https://travis-ci.org/MacPython/dipy-wheels

arokem:
I have the same numpy version (1.9.1) on both 2.7 and 3.4

matthew-brett:
Might be conda / MKL thing?

arokem:
I'm on Anaconda. What is this Travis bot running?

matthew-brett:
Python.org Python with standard wheels.

matthew-brett:
Are these problems related ? http://nipy.bic.berkeley.edu/builders/dipy-py3.3-pip/builds/168/steps/shell_4/logs/stdio

arokem:
Probably not. I don't see this one on my laptop/python3

arokem:
Also, any idea why this build worked?

http://nipy.bic.berkeley.edu/builders/dipy-py3.3-pip/builds/167

IIUC, they're pip installing the same release. What changed on this machine
between March 19th and March 20th?

matthew-brett:
Sorry to get off topic.  The other failure went away with a simple rerun - looks like some randomness in the tests - I'll open another issue.

matthew-brett:
https://github.com/nipy/dipy/issues/613

arokem:
Back to this topic: do we know what changed on our 3.3 worker between March 19th and 20th? It seems to explain this issue.

matthew-brett:
See #613 - the test failure was due to memory access violations, unrelated to this issue, now fixed in master.

arokem:
Oh yes, you're right. So we still don't understand this issue, huh? Is Conda/MKL still a plausible culprit?

matthew-brett:
I guess it could be Conda +/- MKL - I'm afraid I don't have conda / MKL installed anywhere.

arokem:
Yeah - I've heard it's really tough to install ;-)

So - should we go ahead with this PR? We do want to support conda/mkl, as
well as non-conda/mkl, and it seems that this is how accurate we can get in
matching the two. Is that a fair assessment?

arokem:
Oh wait, I might not have MKL. Anyone know how to check that? Or is
Anaconda always MKL-powered?

DaveOlg:
*I think* the academic version should come with MKL, but the free version
available to the public does not have it.

Bago

arokem:
OK - I definitely did not have MKL before (but now I do!), so the
difference is not about MKL.

arokem:
Whether or not I have MKL installed, this does fail.

On a side note - MKL is fast!

matthew-brett:
Ariel - did you try using pip installed numpy etc ?

arokem:
No - `pip install scipy` doesn't work for me. I'm getting some compilation
error.

matthew-brett:
Yes, sorry, you'd need to install a standard Python.org or macports or homebrew Python, then `pip install numpy scipy cython` should be a few seconds of download / install.  It won't work with conda because conda has an OSX distutils version string, so won't pick up the wheels.

arokem:
OK - that wasn't too bad actually. I might even see myself recommending
this route for installation for beginners, once we sort out the VTK
situation (e.g. find a way to pip install that). And also, I found it
potentially a bit confusing that the installation only creates the
'python3.4' name in bash (in that weird mac-y /Library/Frameworks
location), but not 'python'. Any particular reason for that? I figure I
could hack an alias or a symlink, but that's not easy to explain to a
beginner, so I would love to have a more straightforward route.

At any rate - I see the same test-failure with the python.org python and
pip-installed everything. So, this is apparently not about MKL/Anaconda.
Should I try downgrading to OS 10.9? ;-)

omarocegueda:
Hi guys!,
this reminds me the problems we had with the regression tests of the
registration module. I think, in general, we will always have this kind of
issues whenever our "oracle" is obtained by running a specific version of
the code, like this:
https://github.com/arokem/dipy/blob/test-lamda-csdeconv/dipy/reconst/tests/test_csdeconv.py#L432

In order to reproduce those values, we need to know if the compiled code
used double or extended precision for the computations at that specific
run. Unfortunately, we not always can ensure that the compiled code will
use the same precision for the computations (I remember a long discussion
with Matthew when we were trying to do exactly that). The 10^-4 tolerance
looks quite good by the way (apparently CSD is very stable across
architectures!). The only alternative I can think of is to make the oracle
depend on the compiler options used, but if the 10^-4 tolerance works well
(what is the largest difference you have seen in this case?), then I think
that is preferable instead of making the test code have more dependencies.

On Mon, Apr 6, 2015 at 7:21 PM, Ariel Rokem <notifications@github.com>
wrote:

> OK - that wasn't too bad actually. I might even see myself recommending
> this route for installation for beginners, once we sort out the VTK
> situation (e.g. find a way to pip install that). And also, I found it
> potentially a bit confusing that the installation only creates the
> 'python3.4' name in bash (in that weird mac-y /Library/Frameworks
> location), but not 'python'. Any particular reason for that? I figure I
> could hack an alias or a symlink, but that's not easy to explain to a
> beginner, so I would love to have a more straightforward route.
>
> At any rate - I see the same test-failure with the python.org python and
> pip-installed everything. So, this is apparently not about MKL/Anaconda.
> Should I try downgrading to OS 10.9? ;-)
>
> On Mon, Apr 6, 2015 at 3:35 PM, Matthew Brett <notifications@github.com>
> wrote:
>
> > Yes, sorry, you'd need to install a standard Python.org or macports or
> > homebrew Python, then pip install numpy scipy cython should be a few
> > seconds of download / install. It won't work with conda because conda has
> > an OSX distutils version string, so won't pick up the wheels.
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/nipy/dipy/pull/612#issuecomment-90266960>.
>
> >
>
> —
> <https://github.com/nipy/dipy/pull/612#issuecomment-90295181>.
>



-- 
-Proverbio chino.
-Chinese proverb.

http://www.cimat.mx/~omar

arokem:
Thanks for the input @omarocegueda! To answer your question: a sensitivity of 10e-4 was chosen here because that was the largest I could make it without failing. That is, the largest error is somewhere between 10e-4 and 10-5.

matthew-brett:
http://nipy.bic.berkeley.edu/builders/dipy-py2.7-osx-10.6/builds/427/steps/shell_6/logs/stdio

arokem:
Interesting - probably doesn't have anything to do with the one-line merge
from earlier today (in fvtk, of all places, so odds of an interaction seem
really low). What else has changed on this machine since April 5th (the
last merge, which came up green)?

arokem:
And is it now more similar to my laptop with python 3 than it was before? :-)

matthew-brett:
I think I accidentally updated scipy from 0.11 to 0.15.1 today.

Now I test, it looks like installing scipy 0.12.0 through 0.14.0 makes the error go away.

pypi appears to be playing up at the moment, but my local copies suggest that 0.15.0 was the first version for which we have a wheel, that used Accelerate instead of ATLAS.

Does that give any clues?

arokem:
Yes - I can confirm that simply changing from scipy 0.14 to 0.15 raises
this error on my machine. Is that true only on Macs? What version of scipy
is on the other bots?

matthew-brett:
Is this also true for conda?  (difference between 0.14 and 0.15)?   Do you know if conda uses Accelerate for both?

matthew-brett:
Also fails on Debian 64-bit with scipy 0.15.1

matthew-brett:
No error on Debian 64-bit with scipy 0.14.0

arokem:
Yes - this is with conda. What is Accelerate ("accelerate scipy" is not
such a great search term...)? And how do I know whether it's being used?

matthew-brett:
Accelerate is Apple's installed BLAS / LAPACK library : 

https://developer.apple.com/library/mac/documentation/Darwin/Reference/ManPages/man7/LAPACK.7.html

You can tell which BLAS / LAPACK library is linked with something like:

```
In [1]: import scipy

In [2]: scipy.__config__.show()
lapack_opt_info:
    extra_link_args = ['-Wl,-framework', '-Wl,Accelerate']
    extra_compile_args = ['-msse3']
    define_macros = [('NO_ATLAS_INFO', 3)]
blas_opt_info:
    extra_link_args = ['-Wl,-framework', '-Wl,Accelerate']
    extra_compile_args = ['-msse3', '-I/System/Library/Frameworks/vecLib.framework/Headers']
    define_macros = [('NO_ATLAS_INFO', 3)]
``` 

But - from the tests on Linux this doesn't look like a BLAS library problem but something to do with scipy 0.15....

arokem:
Something to do with this line, maybe?

https://github.com/nipy/dipy/blob/master/dipy/reconst/shm.py#L188

matthew-brett:
Very likely :)

arokem:
Yes, OK, so the conclusion is that our implementation of the spherical harmonics is somewhat different from scipy's implementation. Which one is the correct one?

matthew-brett:
I'm not qualified on that question - who did the shm implementation on our side?

arokem:
'git blame' says @Garyfallidis wrote those lines of code, so he can probably answer that question

arokem:
Hey @Garyfallidis - any thoughts here about differences between our implementation and scipy implementation of the spherical harmonics?

Garyfallidis:
@MrBago, @stefanv and @mdesco  are better than me to answer this question. I haven't played with SH  for some time. I think git blame mislead you here.

stefanv:
Not my preferred answer, but check for validity against Maxima or
Mathematica.

arokem:
Would you be equally happy with Matlab? I don't have Mathematica, and am
reluctant to install/learn Maxima just for this...

arokem:
OK - through some black magic involving the Wolfram Alpha web API, I managed to calculate spherical harmonics in both scipy, dipy and Matematica (I learn something new every day...): 

http://nbviewer.ipython.org/gist/arokem/2f014ae910d274595a70

The three answers disagree with each other. For what it's worth, dipy is in the middle. For the time being, I propose to use our implementation of the spherical harmonics, instead of scipy. 

I propose to merge this PR as it currently stands.

Garyfallidis:
Agree! Thx man!

matthew-brett:
Ariel - would you mind raising an issue on scipy ?

arokem:
Will do: I already started writing the email to the scipy-dev list. Or
should I raise an issue on the scipy github repo?

matthew-brett:
Either are fine - I didn't realize the differences were so close to `eps`.  Was there just one value where this size of discrepancy came up?  Sorry not to be paying close attention.

arokem:
A little bit of experimentation showed that sometimes dipy and scipy give
exactly the same result. It seems to depend on the order of the spherical
harmonics. This was one case (order=8) for which they didn't match, and the
test error was triggered.

matthew-brett:
So our question is - if the maximum divergence of the spherical harmonic calculation is in the order of 10e-16 - does that matter?  And - is that the maximum divergence?

arokem:
Now that you put it that way, I remember that the maximum divergence is
actually much larger - I had to change the `decimal` kwarg to 4 to get that
test to pass.

arokem:
I should probably try to track down again what the inputs to the SH were in
that specific case and rerun against Mathematica to see what happens then.

matthew-brett:
Would mpmath cover your bases here?

```
In [1]: import numpy as np
In [2]: from scipy.special import sph_harm as sp_sph_harm
In [3]: from sympy.mpmath import spherharm as mp_sph_harm
In [4]: m, n, theta, phi = 8, 2, np.pi/7, np.pi/3
In [5]: sp = sp_sph_harm(n, m, phi, theta)
In [7]: mp = mp_sph_harm(m, n, theta, phi)
In [8]: sp - mp
Out[8]: mpc(real='-1.9428902930940239e-16', imag='3.8857805861880479e-16')
```

Also - scipy / dipy / mpmath here start with a floating point (therefore imprecise) value, and I suppose the Mathematica starts in some more precise representation, so it would be good to separate the effect of the input precision from the output precision.

arokem:
Thanks for the mpmath pointer. It turns out that I was wrong, and the differences are still in the e-15 to e-16 range for these comparisons (http://nbviewer.ipython.org/gist/arokem/d9926589836aafa356d2), but they get amplified in our test to order e-5 to e-6, because of multiplication of these numbers (believe that's the story, see bottom of that notebook). That's sobering. I have had issues posting to scipy dev (I signed up with an old email address, but I will retry).

arokem:
https://github.com/scipy/scipy/issues/4887

arokem:
That all said, should we still merge this PR as it is now? At the very least it preserves compatibility with previous versions of dipy, and it seems that scipy's implementation is the odd man out in all the tests I have done.

matthew-brett:
Hum - so the notebook only tests one set of parameters?

For those sets of parameters, scipy seems closer to mpmath (which should be the standard I suppose).

But - there were other parameters where dipy was closer to mpmath?

How about pumping up the mpath precision, doing a range of parameters and doing something like RMSE with scipy, dipy ?

arokem:
OK. How do I convert one of these guys:

    mpc(real='3.0904650940975318e-13', imag='9.511462442376683e-13'),

Into a floating point (complex) number?

arokem:
Already updated the gist in that direction:
https://gist.github.com/arokem/d9926589836aafa356d2

arokem:
Oh yes - see what you mean. Got my 'small' and 'large' upside down.

stefanv:
```
c = complex(mpc_num)
```

arokem:
Thanks! Well - the gist is updated. I am not sure what to make of any of it.

stefanv:
Where do those expected values come from?

arokem:
From the documentation of this test: "We check that the default value of lambda is the expected value with the symmetric362 sphere. This value has empirically been found to work well and changes to this default value should be discussed with the dipy team." (https://github.com/nipy/dipy/blob/master/dipy/reconst/tests/test_csdeconv.py#L434)

stefanv:
I don't really know how to interpret that.  These values seem to
influence the success/failure of the test, yet their origin seems to be
experimentation?

arokem:
I think that is correct.

These are numbers that were found to work well in the past, and this test
is in place to alert us when they change (a canary in the coal mine, as it
were). Seems to have worked :-)

stefanv:
Now, the question is: did they change for better or worse!

arokem:
That's exactly what I am trying to answer here with all these comparisons.

If I read it correctly, the recent result suggests that, in aggregate, dipy
is closer to sympy on the real part and scipy is closer on the imaginary
part. All these differences are hovering around machine epsilon.

stefanv:
As you mentioned earlier on, it's a bit worrisome that such small
differences result in a 4 decimal difference in the result.  I don't
know what happens underneath the hood, but if there is a pseudo inverse
computed, that would be where I would start searching.

matthew-brett:
Are there any large deviations in there where scipy or dipy are more significantly off from the mpmath result?

argriffing:
> I don't know what happens underneath the hood, but if there is a pseudo inverse computed, that would be where I would start searching.

These explicit constructions of gram matrices are red flags to me:
https://github.com/nipy/dipy/blob/maint/0.9.x/dipy/reconst/csdeconv.py#L173
https://github.com/nipy/dipy/blob/maint/0.9.x/dipy/reconst/csdeconv.py#L529
If that matrix is numerically singular according to `_solve_cholesky` then the code path with the `+ mu*I` is taken and the cholesky solver tries again.  This is especially suspicious in light of the second warning at the bottom of
http://nbviewer.ipython.org/gist/arokem/d9926589836aafa356d2
which suggests that there is a matrix somewhere with an unexpected shape (e.g. tall when it is expected to be wide or vice versa) which could cause a gram matrix to be structurally singular.

argriffing:
Sorry, after looking again at the dipy code I see that this probably doesn't have anything to do with the cholesky decomposition.  Instead my new guess is that it is related to https://github.com/scipy/scipy/issues/4887#issuecomment-103679206

argriffing:
After having the added the most recent commit to this PR, the `sph_harm` import should be added back and the sphere should be changed back to `symmetric362`, right?

arokem:
Yes - thanks for the catch. At least if you are using scipy >0.15. Added back this import conditionally for that case.

arokem:
Hmph. I am now getting this on python 2.7 (but not 3.4):

```

ERROR: Failure: AttributeError ('numpy.ufunc' object has no attribute
'__module__')

  File
"/Users/arokem/anaconda/envs/py2/lib/python2.7/site-packages/nose/loader.py",
line 420, in loadTestsFromName

    addr.filename, addr.module)

  File
"/Users/arokem/anaconda/envs/py2/lib/python2.7/site-packages/nose/importer.py",
line 47, in importFromPath

    return self.importFromDir(dir_path, fqname)

  File
"/Users/arokem/anaconda/envs/py2/lib/python2.7/site-packages/nose/importer.py",
line 94, in importFromDir

    mod = load_module(part_fqname, fh, filename, desc)

  File "/Users/arokem/source/dipy/dipy/reconst/tests/test_csdeconv.py",
line 13, in <module>

    from dipy.reconst.csdeconv import (ConstrainedSphericalDeconvModel,

  File "/Users/arokem/source/dipy/dipy/reconst/csdeconv.py", line 19, in
<module>

    from dipy.reconst.shm import (sph_harm_ind_list, real_sph_harm,

  File "/Users/arokem/source/dipy/dipy/reconst/shm.py", line 198, in
<module>

    def spherical_harmonics(m, n, theta, phi, dtype=complex):

  File "/Users/arokem/anaconda/envs/py2/lib/python2.7/functools.py", line
33, in update_wrapper

    setattr(wrapper, attr, getattr(wrapped, attr))

AttributeError: 'numpy.ufunc' object has no attribute '__module__'


Ran 1 test in 0.001s
```

I think that has to do with @mrBago's recent suggestion, because when I
back off from that, everything seems fine.

arokem:
Concerning the sphere: I think that we want to move ahead with the default
sphere.

MrBago:
Sorry I didn't test it before I pushed it, apparently @wraps chokes on
ufuncs,

```
@wraps(numpy.add)
def foo():
    pass
```

Feel free to take out the @wraps line, or my whole change.

stefanv:
@MrBago Can you explain?  This works, e.g.:

```
import numpy as np
from functools import wraps

def my_decorator(f):
    @wraps(f)
    def wrapper(*args, **kwds):
        print('Calling decorated function')
        return f(*args, **kwds)
    return wrapper

f = my_decorator(np.add)
print(f(1, 2))

```

DaveOlg:
@Stefan your code above fails on my machine: ubuntu 14.1, python2.7.8,
numpy 1.8.2.

```
  File "testWraps.py", line 11, in <module>
    f = my_decorator(np.add)
  File "testWraps.py", line 5, in my_decorator
    @wraps(f)
  File "/usr/lib/python2.7/functools.py", line 33, in update_wrapper
    setattr(wrapper, attr, getattr(wrapped, attr))
AttributeError: 'numpy.ufunc' object has no attribute '__module__'

```

stefanv:
I have numpy 1.10 with Python 3.4.

arokem:
I think it's a python 3 <=> 2 thing. That snippet failed on 2.7.9, and
passed on 3.4.3, on my machine, same numpy (1.9.2) on both.

arokem:
I just wanted to point out that this is potentially the right way to deal with this issue, but am still waiting to hear back on #652.

MrBago:
The problem with #652 is that we still have functions like
`shm.real_sh_basis` that will exhibit the bad behaviour if our users are
unlucky enough to use them with the exact wrong set of arguments. Ideally
we would fix this at the `sph_harm` level.

arokem:
Oh yeah - I am not proposing to merge #652 instead of this one.

Just pointing out that the discussion there about the right way to handle this issue is still ongoing. So far, it seems like the way implemented here is the right way to handle this issue for now.

MrBago:
The only reason I used @wrap was so I didn't have to write a docstring, we
can just drop @wrap and make the docstring better and it should work.

Bago

arokem:
How about this then? Docstring slightly improved to indicate that we use scipy in cases where version >= 0.15

DaveOlg:
I like it! +1

MrBago:
I seem to have done something odd to my github account...

but +1 for the docstring!

arokem:
No further discussions on whether this is really the ultimate best-practice for handling this. It seems like this is at the very least an acceptable solution. Any reason not to merge this fix?

argriffing:
For what it's worth, I just merged the scipy-level workaround in https://github.com/scipy/scipy/pull/4896, and the underlying numpy bug https://github.com/numpy/numpy/issues/5895 is still open.

Garyfallidis:
Hey @arokem I am lost a bit in the discussion of this PR. Is this ready to be merged now? Do we continue using the scipy functions after scipy 0.15 or stick to using the Dipy version?

arokem:
I think that it is ready to be merged. 

We are using scipy from 0.15 onwards, but at least for 0.15 itself, we are taking into account that we need to specify the `dtype` kwarg. We might be able to do away with that workaround in future versions, and eventually also drop our implementation altogether.

arokem:
Are there any remaining objections here? Just rebased this.

Issue #591, opened 2015-03-14
Allow seed_from_mask to generate random seeds 


Garyfallidis:
seeds_from_mask is currently generating evenly distributed points in every voxel. Which is very nice. It would be nice also to have random points too. E.g. determine 10 random points in each voxel.

skoudoro:
I close this one, it seems to be ok @Garyfallidis

PR #614, opened 2015-03-24
Seeds from mask random

wenderen:
Code + tests for #591


wenderen:
I stripped trailing spaces before committing, so there are some extra changes in there. The relevant changes are confined to `seeds_from_mask()` and `test_seeds_from_mask()`

MrBago:
Thanks @wenderen this looks nice. Can I suggest that we add the feature as a separate function? I don't think we gain much by having it be in `seeds_from_mask`. For example, I don't know if using len 3 density input makes sense for random seeds, maybe instead of density a `random_seeds_from_mask` function could have a `seeds_per_voxel` argument. We could add the new in a few examples and in the see also section of `seeds_from_mask` to help users find it.

wenderen:
I agree, a separate function makes sense. I'll make the changes and push again.

wenderen:
The build check for `random_seeds_from_mask` is failing because the random outputs from the Travis job are different from the example outputs I got - as they should be :)

Should I remove the examples from `random_seeds_from_mask` or keep them there, but comment them out?

dipy/tracking/utils.py
```diff
@@ -364,31 +364,41 @@ def seeds_from_mask(mask, density=[1, 1, 1], voxel_size=None, affine=None):
...
-    array([[ 0.5,  0.5,  0.5]])
-
-    >>> seeds_from_mask(mask, [1,2,3], [1,1,1])
-    array([[ 0.5       ,  0.25      ,  0.16666667],
-           [ 0.5       ,  0.75      ,  0.16666667],
-           [ 0.5       ,  0.25      ,  0.5       ],
-           [ 0.5       ,  0.75      ,  0.5       ],
-           [ 0.5       ,  0.25      ,  0.83333333],
-           [ 0.5       ,  0.75      ,  0.83333333]])
+    >>> random_seeds_from_mask(mask, seeds_per_voxel=1, voxel_size=[1,1,1])
```
MrBago: Why is the example for `seeds_from_mask` calling `random_seeds_from_mask`? Shouldn't this be in the latter function?

dipy/tracking/utils.py
```diff
@@ -420,6 +431,75 @@ def seeds_from_mask(mask, density=[1, 1, 1], voxel_size=None, affine=None):
...
+    where = np.repeat(where, seeds_per_voxel, axis=0)
+    seeds = where + grid - .5
+    seeds = asarray(seeds)
+
+    # Apply the spacial transform
+    if affine is not None:
+        # Use affine to move seeds int real world coordinates
+        seeds = np.dot(seeds, affine[:3, :3].T)
+        seeds += affine[:3, 3]
+    elif voxel_size is not None:
```
MrBago: We don't need to worry about voxel size here, we left it in the other functions for backwards compatibility, but we don't need to add it to new functions.

wenderen: Removed

dipy/tracking/utils.py
wenderen: Typo, sorry. Fixed.

dipy/tracking/utils.py
```diff
@@ -420,6 +431,77 @@ def seeds_from_mask(mask, density=[1, 1, 1], voxel_size=None, affine=None):
...
+    num_voxels = len(where)
+
+    # Generate as many random triplets as the number of seeds needed
+    grid = np.random.random([seeds_per_voxel * num_voxels, 3])
+    # Repeat elements of 'where' so that it can be added to grid
+    where = np.repeat(where, seeds_per_voxel, axis=0)
+    seeds = where + grid - .5
+    seeds = asarray(seeds)
+
+    # Apply the spacial transform
```
arokem: typo: "spacial" => "spatial"

dipy/tracking/utils.py
```diff
@@ -420,6 +431,77 @@ def seeds_from_mask(mask, density=[1, 1, 1], voxel_size=None, affine=None):
...
+    if affine is not None:
+        # Use affine to move seeds int real world coordinates
```
arokem: "int" => "into"

wenderen: Done

dipy/tracking/utils.py
wenderen: Done

wenderen:
(ping)

dipy/tracking/utils.py
```diff
@@ -420,6 +431,77 @@ def seeds_from_mask(mask, density=[1, 1, 1], voxel_size=None, affine=None):
...
+    Raises
+    ------
+    ValueError
+        When ``mask`` is not a three-dimensional array
+
+    Examples
+    --------
+    >>> mask = np.zeros((3,3,3), 'bool')
+    >>> mask[0,0,0] = 1
+    >>> random_seeds_from_mask(mask, seeds_per_voxel=1, voxel_size=[1,1,1])
```
arokem: This doctest is currently failing, because the function does not have a `voxel_size` kwarg. Is knowing the voxel size important for this function? If it's not, can you remove this kwarg from the doctest?

arokem: Otherwise - if it is important to know the voxel size (seems like it might be), you'll need to implement handling of voxel size into the function itself.

wenderen: `voxel_size` isn't really important for this function. I'll remove the kwarg in the doctest.

arokem: The tests are still failing (see:
https://travis-ci.org/nipy/dipy/jobs/56414190), because of the randomness.
You might want to set the random seed (as in this question:
http://stackoverflow.com/questions/16341484/doctest-for-a-function-using-a-randomly-generated-variable),
to make sure you always get the same answer.

wenderen: Thank you for the link, I wasn't aware.
Hopefully the doctests will pass now.

arokem:
They're still failing, but it's getting really close. I think that the failures are now only because of formatting (white-space)

Check the travis builds by clicking on the icon next to the commit message (red 'x' for failed builds, green 'v' for successful builds, orange 'o' for builds in progress) on the PR conversation page, or go to https://travis-ci.org/nipy/dipy to find the relevant build. Also - you can run the tests with a '--with-doctest' to run the doctests as well as the unit tests.

arokem:
Yep - that seems to have done the trick.

dipy/tracking/tests/test_utils.py
```diff
@@ -410,12 +411,25 @@ def test_seeds_from_mask():
     assert_equal(in_444.sum(), 3 * 4 * 5)
 
 
+def test_random_seeds_from_mask():
+
+    mask = mask = np.random.random_integers(0, 1, size=(4, 6, 3))
```
arokem: I think this is a typo: `mask = mask = ...`

dipy/tracking/utils.py
```diff
@@ -420,6 +431,79 @@ def seeds_from_mask(mask, density=[1, 1, 1], voxel_size=None, affine=None):
...
+    >>> mask[0,1,2] = 1
+    >>> random_seeds_from_mask(mask, seeds_per_voxel=2)
+    array([[ 0.46826158, -0.18657582,  0.19232262],
+           [ 0.37638915,  0.39460666, -0.41495579],
+           [-0.46094522,  0.66983042,  2.3781425 ],
+           [-0.40165317,  0.92110763,  2.45788953]])
+
+    """
+    mask = np.array(mask, dtype=bool, copy=False, ndmin=3)
+    if mask.ndim != 3:
```
samuelstjean: Why are 2D mask not accepted?

arokem:
This all seems fine to me. @Garyfallidis - could you please take a look and confirm that it properly addresses the original issue?

MrBago:
Looks good to me, I'll merge this soon if there are no objections.

Garyfallidis:
Yes Bago please take the lead on merging this. Busy week in Sherbrooke. Many thx @wenderen :)

wenderen:
Thanks :)

wenderen:
(ping)

Issue #387, opened 2014-07-07
References do not render correctly in SHORE example


Garyfallidis:
@maurozucchelli have a look here
http://nipy.org/dipy/examples_built/reconst_shore.html#example-reconst-shore
and the word workshop is repeated twice.

skoudoro:
Should we close this one @Garyfallidis  ?

skoudoro:
seems to be fixed. Closing

PR #615, opened 2015-03-25
bugfix for double word in doc example issue #387

arybinski:
getting familiar with github workflow and starting with this simple fix


arokem:
Good catch. Welcome on board!

arokem:
For future reference, please use the following codes for commit messages: 

http://nipy.org/dipy/devel/commit_codes.html

Also, it's good practice to write commit messages in the imperative. For example, for this commit, a good message might be something like: "DOC: Remove erroneous repetition"

PR #619, opened 2015-04-01
DOC: Added missing contributor to developer list

jchoude:
Simply readd a missing developer from the contributor list. It seems to have been lost when updating the developers list during one of the last releases.

I also saw that there were other missing contributors (with low commit counts, as I do). @Garyfallidis said to wait before adding them.

### DIFF
--- a/doc/developers.rst
+++ b/doc/developers.rst
@@ -30,8 +30,17 @@ And here is the rest of the wonderful contributors:
 - **Emmanuel Caruyer**, University of Pensylvania, USA
 - **Sylvain Merlet**, INRIA, Sophia-Antipolis, FR
 - **Erick Ziegler**, Université de Liège, BE
-- **Kimberly Chan**, Stanford University, CA, US
+- **Kimberly Chan**, Stanford University, CA, USA
 - **Chantal Tax**, University Medical Center, Utrecht, NL
+- **Demian Wassermann**, INRIA, Sophia Antipolis, FR
+- **Gregory R. Lee**, Cincinnati Children's Hospital Medical Center, Cincinnati, OH, US
+- **Endolith**, New-York, NY, USA
+- **Matthias Ekman**, Donders Institute for Brain, Cognition and Behaviour, Nijmegen, NL
+- **Andrew Lawrence**
+- **Kesshi Jordan**, University of California, San Francisco, CA, USA
+- **Maria Luisa Mandelli**, University of California, San Francisco, CA, USA
+- **Adam Rybinski**, Jagiellonian University, Krakow, PL
+- **Qiyuan Tian**, Stanford University, Stanford, CA, USA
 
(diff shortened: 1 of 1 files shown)


Garyfallidis:
Had more thoughts on this. Let's add them all. The more the merrier! Thx man!

Garyfallidis:
There should be 31 contributors in total.

arokem:
+1. 

Maybe we should somehow consolidate and have this list only in one place? We have another list in `AUTHOR`

jchoude:
Ok, I'll update this this afternoon with all names. However, I don't know if I'll easily be able to find the affiliation of everyone. I'll give it a go and let you know.

Also, as suggestion by @arokem, should we consolidate it only in one place, and if so, which one would be the best?

Garyfallidis:
I like that in this page we can give credit to the core devs who have contributed much more in the project. I would like here also to add links to their pages or even add information about their area of expertise. So, this is different from AUTHOR.

Garyfallidis:
No, sorry. There is no need to find the affiliations now. I will do that at a later stage. No rush on this.

arokem:
Sorry - this confuses me. I thought that you want to put all the
contributors on this page, not only core devs, as well as in the AUTHOR
file? I am worried that we'll need to constantly make sure that these are
in sync.

Garyfallidis:
http://nipy.org/dipy/developers.html

arokem:
Gotcha. Shouldn't be too hard to replicate this structure in the AUTHORS
file as well, right? Why do we need two different files with the same
information?

Garyfallidis:
Because it is not the same information? The authors file is only the list of all authors. And the developers.rst gives more information about who the authors are and what they do. And should have more information in the future too. This file should be like the "Who are we" found in commercial websites. We could make it automatic in some way. But not sure if it worth the effort right now.

arokem:
OK.

Garyfallidis:
But hey point taken. It will be nice to have a script that updates developers.rst automatically.

Garyfallidis:
Let's think of a form first. Should have links to websites and short bios or summaries of interests for all the core devs at least? Anyway... let's think of it for a bit.

arokem:
Yeah - I think that it wouldn't be too hard to extend AUTHOR a bit, to be a
CSV table with some basic information about each contributor: email
address, but also affiliation, commit number, etc. Might make for useful
information about the project that we can use to tell people (e.g. funding
agencies...) about the project. One huge strength of the project which I
always emphasize when telling people about it, is that it's distributed
over many different labs/institutions/countries. We can demonstrate that
through a more careful collection of data. I've made an issue of it #620

jchoude:
Hey guys, just to let you know I added all missing contributors (at least, those that were listed with a git shortlog -ns).

I couldn't find any affilition for Andrew Lawrence, so if any of you know him, please let me know.

Garyfallidis:
Okay no worries. Let's update the website.

PR #629, opened 2015-04-13
Quickbundles 2.1

MarcCote:
This PR mainly adds new `Feature` classes (`ResampleFeature`, `MidPointFeature`, `ArcLengthFeature` and `VectorBetweenEndpointsFeature`) and a new `Metric` class (`CosineMetric`).

Moreover, a new tutorial has been added explaining how to make new `Feature` and new `Metric` for Quickbundles.

This PR also fixes some bugs and adds more sanity checks.


dipy/segment/metricspeed.pyx
```diff
@@ -290,6 +303,49 @@ cdef class MinimumAverageDirectFlipMetric(AveragePointwiseEuclideanMetric):
...
+    -----
+    The distance between two vectors $v_1$ and $v_2$ is equal to
+    $\frac{1}{\pi} \arccos\left(\frac{v_1 \cdot v_2}{\|v_1\| \|v_2\|}\right)$
+    and is bounded within $[0,1]$.
+    """
+    def __init__(CosineMetric self, Feature feature):
+        super(CosineMetric, self).__init__(feature=feature)
+
+    cdef int c_are_compatible(CosineMetric self, Shape shape1, Shape shape2) nogil except -1:
+        return same_shape(shape1, shape2) and shape1.dims[0] == 1
```
omarocegueda: if same_shape(shape1, shape2)!=0 and shape1.dims[0] == 1:
    return 1
else:
    return 0
to avoid Cython trying to create a boolean object:
https://travis-ci.org/nipy/dipy/jobs/58217418#L775

MarcCote: Thanks @omarocegueda

dipy/segment/tests/test_metric.py
```diff
@@ -64,6 +65,84 @@ def test_metric_minimum_average_direct_flip():
...
+
+    class CosineMetric(dipymetric.Metric):
+        def __init__(self, feature):
+            super(CosineMetric, self).__init__(feature=feature)
+
+        def are_compatible(self, shape1, shape2):
+            return shape1 == shape2 and shape1[0] == 1
+
+        def dist(self, v1, v2):
+            norm = lambda x: np.sqrt(np.sum(x**2, dtype=np.float64))
```
arokem: You might want to use np.linalg.norm

dipy/segment/clustering.py
```diff
@@ -416,7 +417,7 @@ class QuickBundles(Clustering):
     metric : str or `Metric` object (optional)
         The distance metric to use when comparing two streamlines. By default,
         the Minimum average Direct-Flip (MDF) distance [Garyfallidis12]_ is
-        used and requires streamlines to have the same number of points.
+        used and streamlines are automatically resampled so they have 18 points.
```
arokem: I assume there's some reason 18 is used here?

dipy/segment/featurespeed.pyx
```diff
@@ -222,6 +249,89 @@ cdef class CenterOfMassFeature(CythonFeature):
             out[0, d] /= N
 
 
+cdef class MidpointFeature(CythonFeature):
+    """ Extracts features from a sequential datum.
+
+    A sequence of N-dimensional points is represented as a 2D array with
+    shape (nb_points, nb_dimensions).
+
+    The feature being extracted consists in one N-dimensional point representing
```
arokem: "consists of"

dipy/segment/featurespeed.pyx
```diff
@@ -222,6 +249,89 @@ cdef class CenterOfMassFeature(CythonFeature):
...
+        out[0, 0] = c_length(datum)
+
+
+cdef class VectorBetweenEndpointsFeature(CythonFeature):
+    """ Extracts features from a sequential datum.
+
+    A sequence of N-dimensional points is represented as a 2D array with
+    shape (nb_points, nb_dimensions).
+
+    The feature being extracted consists in one N-dimensional point representing
```
arokem: IIUC, this feature is not really a point in the same way that the inputs are sequences of points, but rather a vector in the N-dimensional space pointing from one end-point to the other.

dipy/segment/featurespeed.pyx
```diff
@@ -222,6 +249,89 @@ cdef class CenterOfMassFeature(CythonFeature):
...
+        shape.size = datum.shape[1]
+        return shape
+
+    cdef void c_extract(VectorBetweenEndpointsFeature self, Data2D datum, Data2D out) nogil:
+        cdef:
+            int N = datum.shape[0], D = datum.shape[1]
+            int d
+
+        for d in range(D):
+            out[0, d] = datum[N-1, d] - datum[0, d]
```
arokem: Can you index using `datum[-1, d]`?

dipy/segment/metric.py
```diff
@@ -1,15 +1,23 @@
...
                                       AveragePointwiseEuclideanMetric,
-                                      MinimumAverageDirectFlipMetric)
+                                      MinimumAverageDirectFlipMetric,
+                                      CosineMetric)
 
 from dipy.segment.metricspeed import (dist,
                                       distance_matrix)
 
+# Creates aliases
+EuclideanMetric = SumPointwiseEuclideanMetric
```
arokem: Is this here for backwards compatibility?

dipy/segment/metricspeed.pyx
```diff
@@ -191,6 +204,9 @@ cdef class SumPointwiseEuclideanMetric(CythonMetric):
     is equal to $a+b+c$ where $a$ is the Euclidean distance between s1[0] and
     s2[0], $b$ between s1[1] and s2[1] and $c$ between s1[2] and s2[2].
     """
+    #def __init__(SumPointwiseEuclideanMetric self, Feature feature=ResampleFeature(18)):
```
arokem: If this is not needed, you can delete it.

dipy/segment/metricspeed.pyx
```diff
@@ -290,6 +303,49 @@ cdef class MinimumAverageDirectFlipMetric(AveragePointwiseEuclideanMetric):
         return min(dist_direct, dist_flipped)
 
 
+cdef class CosineMetric(CythonMetric):
+    r""" Computes the cosine distance between two vectors.
+
+    A vector of N-dimensional points is represented as a 2D array with
+    shape (1, nb_dimensions).
```
arokem: This is just one N-dimensional point?

dipy/segment/clustering.py
MarcCote: No, not really. I used 18 since that is what @Garyfallidis used in the example about streamlines segmentation. I don't know what you guys prefer: force the user to provide the number of points, or set one by default and mention it clearly in the documentation?

dipy/segment/featurespeed.pyx
MarcCote: Yes, you are right, I'll change it.

dipy/segment/featurespeed.pyx
MarcCote: No, since I disabled bounds checking. At the top of the file: `# cython: wraparound=False, cdivision=True, boundscheck=False`

However, using -1 won't fail at compilation time. Instead at runtime it will behave unexpectedly! -1 refers to memory before the beginning of the array.

dipy/segment/metric.py
MarcCote: No. Depending on the context, I find it clearer. Particularly, when comparing scalars or N-dimensional points. I think it makes more sense to use the term `EuclideanMetric` when computing Euclidean distance between two points and use `SumPointwiseEuclideanMetric` when comparing two sequences. Since `SumPointwiseEuclideanMetric` is generic it can easily handles the scalar and single Nd point cases.

For example, somewhere in the tutorial about extending features, we have to compare scalars: `metric = EuclideanMetric(feature=ArcLengthFeature())`.

dipy/segment/metricspeed.pyx
MarcCote: Yep, I'll delete it.

dipy/segment/metricspeed.pyx
MarcCote: Yep, typo.

dipy/segment/clustering.py
arokem: I would prefer this be a kwarg that users can easily change if they want to. Would that be hard to implement?

dipy/segment/featurespeed.pyx
arokem: Cool. I learn new things here every time I do a review :-)

dipy/segment/metric.py
arokem: Would this be confusing if someone were to look for the docstring of `EuclideanMetric` and found the docstring of `SumPointWiseEuclideanMetric`? Maybe better to implement explicitly as a class with its own docstring, but inheriting all of its properties from the parent? 

I am not sure - just thinking out loud here.

MarcCote: What you said makes sense, I'll create a new class that subclass `SumPointWiseEuclideanMetric`. Thanks for the idea.

dipy/segment/clustering.py
MarcCote: It wouldn't be hard to implement. However, I don't know if this would make sense to add a parameter like that in the `QuickBundles` constructor since `QuickBundles` can accept any kind of metric that doesn't necessarily need a `nb_points`.

I would argue that users who just want to use the QuickBundles algorithm as a black box (I'm guessing it would be the majority) won't care about changing the number of points. This parameter has a direct impact on the speed of the  algorithm but also the quality. I'm afraid uninformed users would simply decrease that number to gain speed without necessarily considering the validity of the results.

An informed user can "easily" and explicitly change that parameter as follows
```python
feature = ResampleFeature(nb_points=42)
metric = AveragePointwiseEuclideanMetric(feature)
qb = QuickBundles(threshold=10., metric=metric)
clusters = qb.cluster(streamlines)
```
and I should probably add a note about that in the tutorial. What do you think? @Garyfallidis ?

arokem: Yeah. How about creating a few usage examples in the docstring of this
class? Include simpler examples, but also something like this example. You
can certainly reuse code you have in the tests for that. I think that some
users will look at the docstring of the class, even rather than looking at
the tutorial (for IPython users in particular, this is the natural thing to
do).

This kind of hard-coded parameters needs to be really well documented - why
you chose a certain setting (here, this follows Eleftherios et al.'s paper,
I understand?), and what the implications of changing it would be (here a
speed-accuracy tradeoff). You can certainly state these things in the
docstring, as you show the example. Warn people about changing this and
what effect that will have on the results.

doc/examples/segment_extending_clustering_framework.py
```diff
@@ -0,0 +1,233 @@
+"""
+===========================================================
+How to make new `Feature` and new `Metric` for Quickbundles
```
Garyfallidis: Hi @MarcCote, this title is not very attractive. 
I would say something more general. For example,
Enhancing QuickBundles with different metrics and features.
or
Introducing the advanced capabilities of the clustering framework. A use case
with adding new metrics to Quickbundles.

doc/examples/segment_extending_clustering_framework.py
```diff
@@ -0,0 +1,233 @@
...
+===========================================================
+
+This tutorial shows how to create new `Feature` and new `Metric` classes that
+can be used by QuickBundles.
```
Garyfallidis: For the purpose of ...

Garyfallidis: You should explain a bit why this is important.

doc/examples/segment_extending_clustering_framework.py
```diff
@@ -0,0 +1,233 @@
...
+
+Clustering framework
+====================
+Dipy provides a simple, flexible and fast framework to do clustering of
+sequential data (e.g. streamlines).
+
+A *sequential datum* in Dipy is represented as a numpy array of size
+:math:`(N \times D)` where each row of the array represent a D dimensional
```
Garyfallidis: represents

doc/examples/segment_extending_clustering_framework.py
```diff
@@ -0,0 +1,233 @@
...
+point of the sequence. A set of these sequences is represented as a list of
+numpy arrays of size :math:`(N_i \times D)` for :math:`i=1:M` where $M$ is the
+number of sequences in the set.
+
+This clustering framework is modular and divided in three parts:
+1) features extraction
```
Garyfallidis: feature

doc/examples/segment_extending_clustering_framework.py
```diff
@@ -0,0 +1,233 @@
...
+2) distance computation
+3) data partitioning
```
Garyfallidis: You mean clustering algorithm?

doc/examples/segment_extending_clustering_framework.py
```diff
@@ -0,0 +1,233 @@
...
+
+The features extraction part regroups any preprocessing needed to be done on
```
Garyfallidis: feature
regroups is not clear here. Clarify please. An example of a simple feature would be helpful here.

doc/examples/segment_extending_clustering_framework.py
```diff
@@ -0,0 +1,233 @@
...
+the data before computing distances between them. To define new way of
+extracting features, one has to subclass `Feature` (see below).
+
+The distance computation part regroups any metric capable of evaluating a
```
Garyfallidis: Regroups is not informative here. Part is not needed too.

doc/examples/segment_extending_clustering_framework.py
```diff
@@ -0,0 +1,233 @@
...
+distance between two set of features previously extracted from the data. To
+define new way of extracting features, one has to subclass `Metric` (see below).
```
Garyfallidis: a new way

doc/examples/segment_extending_clustering_framework.py
```diff
@@ -0,0 +1,233 @@
...
+
+The data partitionning part represents the clustering algorithm itself
+(e.g. QuickBundles, K-means, Hierarchical Clustering). More precisely, it
+regroups any algorithms taking as input a list of sequential data and
```
Garyfallidis: Regroups again not clear? What do you want to say?

doc/examples/segment_extending_clustering_framework.py
```diff
@@ -0,0 +1,233 @@
...
+outputting a `ClusterMap` object. To define new clustering algorithms see tutorial###.
+
+
+Extending `Feature`
+===================
+This section will guide you through the creation of a new features extraction
```
Garyfallidis: feature extraction

Garyfallidis: correct everywhere please or say extraction of features

doc/examples/segment_extending_clustering_framework.py
```diff
@@ -0,0 +1,233 @@
...
+        cos_theta = np.dot(v1, v2.T) / (norm(v1)*norm(v2))
+
+        # Make sure it's in [-1, 1], i.e. within domain of arccosine
+        cos_theta = np.minimum(cos_theta, 1.)
+        cos_theta = np.maximum(cos_theta, -1.)
+        return np.arccos(cos_theta) / np.pi  # Normalized cosine distance
+
+"""
+The new distance `CosineMetric` is ready to be used. Let's use
+it to cluster a set of streamlines according the cosine distance of the
```
Garyfallidis: according to

doc/examples/segment_extending_clustering_framework.py
```diff
@@ -0,0 +1,233 @@
...
+
+We start by loading the fornix streamlines.
+"""
+
+import numpy as np
+from nibabel import trackvis as tv
+from dipy.data import get_data
+from dipy.viz import fvtk
+
+fname = get_data('fornix')
```
Garyfallidis: If you want to use a different bundle than the fornix. I have some simple cingulum bundles (cb) in the data. Look in get_data.
Will try to have some more bundles asap.

dipy/tracking/streamlinespeed.pxd
```diff
@@ -0,0 +1,27 @@
...
+#import numpy as np
+#cimport numpy as np
+#import cython
+
+#from libc.math cimport sqrt
+
+#cdef extern from "stdlib.h" nogil:
+#    ctypedef unsigned long size_t
+#    void free(void *ptr)
+#    void *malloc(size_t size)
```
Garyfallidis: If you don't need the commented imports/cimport and cdef can you remove them?

dipy/segment/clustering.py
Garyfallidis: @MarcCote you can use the same default parameter in what was in the initial Quickbundles algorithm. If not as @arokem suggests you have to explain what will be the effect of the changes and give examples. 
I really like this new design :+1: 
```python
feature = ResampleFeature(nb_points=42)
metric = AveragePointwiseEuclideanMetric(feature)
qb = QuickBundles(threshold=10., metric=metric)
clusters = qb.cluster(streamlines)
```

dipy/segment/featurespeed.pyx
```diff
@@ -187,13 +188,39 @@ cdef class IdentityFeature(CythonFeature):
                 out[n, d] = datum[n, d]
 
 
+cdef class ResampleFeature(CythonFeature):
```
Garyfallidis: Resample is very general. We need a more precise name.

dipy/segment/clustering.py
```diff
@@ -2,6 +2,7 @@
 import numpy as np
```
Garyfallidis: I noticed that the coverage in clustering.py is at 86% is that the best it can be?

dipy/segment/clustering.py
MarcCote: Ok, the default is 12 points, I'll change it.

dipy/segment/metric.py
MarcCote: @Garyfallidis Ok, I've been rethinking the framework. I'll make another PR explaining why but the gist of it is: what the hell I was thinking making `Metric` dependent on `Features`!

For now, I will remove the alias and modify the tutorial where it was used (i.e. the only place it was used anyway).

doc/examples/segment_extending_clustering_framework.py
MarcCote: I find the former better, for now. The second one might be goof once we will have more tutorials describing all the potential of the framework.

dipy/tracking/streamlinespeed.pxd
MarcCote: Oups. Done.

dipy/segment/featurespeed.pyx
MarcCote: You when me to put SetNumberOfPointsFeature :/

dipy/segment/clustering.py
MarcCote: I'll check what I can do.

dipy/segment/featurespeed.pxd
```diff
@@ -1,5 +1,5 @@
 from cythonutils cimport Data2D, Shape
-
+cimport numpy as np
```
matthew-brett: I personally prefer 'cimport numpy as cnp' so it is clearer which is Python numpy, which Cython.

MarcCote:
It seems the QuickBundles2 code is now failling with Cython 0.22.1. I think it has something to do with `except *`, I'll try to fix that.

FYI, according to Stefan Behnel from Cython:
"I understand that developers need Cython to work on kivy. I was just 
wondering why users need Cython in order to *install* kivy. Usually, 
shipping the generated (and tested) C sources avoids that. "
See https://groups.google.com/d/msg/cython-users/r3dQSfOKMjI/U8vabWagDQkJ

dipy/segment/featurespeed.pxd
MarcCote: I agree, I'll change it everywhere it applies.

dipy/segment/clustering.py
MarcCote: Done

dipy/segment/featurespeed.pyx
Garyfallidis: No, it's okay. Keep it as it is.

Garyfallidis:
Very well done! Next PR please! :+1:

arokem:
Nice!

dipy/segment/clustering.py
kesshijordan: I want to make a note here for future reference: I found a reason to increase the resolution. In PR #1625 I'm using quickbundles to generate templates based on a set of many bundles. The centroids in aggregate form the bundle template, and the resolution of the centroid is this nb_points so if you want a higher resolution/nicer template there is reason to change this. I wasn't looking at the example initially and found this confusing (maybe it's just me? I've been using quickbundles for awhile so I wasn't as quick to reference the docstring example as I should have been). Does it make sense to add a few words in the input description of metric to make sure the example gets referenced when someone is using quickbundles?

PR #633, opened 2015-04-23
BF - no import of TissueTypes

MrBago:
TissueTypes was in `__all__` but it was never imported.

### DIFF
--- a/dipy/tracking/local/__init__.py
+++ b/dipy/tracking/local/__init__.py
@@ -4,5 +4,5 @@ from .tissue_classifier import (ActTissueClassifier, BinaryTissueClassifier,
 from .direction_getter import DirectionGetter
 from dipy.tracking import utils
 
-__all__ = ["LocalTracking", "ActTissueClassifier",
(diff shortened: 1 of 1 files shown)


arokem:
Thanks for noticing this. Is it ever used in this context? That is, why did you decide to import it, rather than removing it from `__all__`?

MrBago:
I assumed that whoever added it to __all__ wanted it there, but If none
objects I'm just as happy removing it.

Bago

arokem:
Maybe @gabknight knows which way we want it. He added in `TissueType` here:
https://github.com/nipy/dipy/commit/759859d06f920639687feab543bafd06d4ac398d
and it seems that there was some intention behind it, considering that it
was in the commit message. Hey @gabknight - what do you think we should do
here?

gabknight:
'TissueType' can be removed from __all__, there is no need for it. Thanks!

arokem:
Thank you for clarifying!

MrBago:
K, I removed it from all.

arokem:
Cool. I don't expect any issues here, but I'll wait for Travis and then
commit.

Thanks!

PR #635, opened 2015-04-24
Select streamlines based on logical operations on ROIs

arokem:
This PR starts what will be a long series of PRs towards the implementation of the AFQ system in dipy (see http://journals.plos.org/plosone/article?id=10.1371/journal.pone.0049790), work that I will undertake in close collaboration with @jyeatman, who has done much of the development of the system in Matlab, and has been the maintainer of that software for several years. 

For now, this just does a couple of really simple things, as you can see: 

- Determines whether streamlines pass within some distance from a mask (a region of interest)
- Combines the conclusions from this determination based on inclusion and exclusion criteria. For now, only a fairly limited set of logical operations are possible. 

Questions include: 

- Does the `select_by_roi` function makes sense where it currently is? 
- Would we want/need to build out a more extensive set of logical operations over combinations of ROIs? What would that entail?


dipy/tracking/utils.py
```diff
@@ -484,6 +489,45 @@ def target(streamlines, target_mask, affine, include=True):
...
+    -------
+    Boolean array. Set to True in each streamline
+    """
+    if affine is None:
+        affine = np.eye(4)
+    roi_coords = np.array(np.where(target_mask)).T
+    x_roi_coords = apply_affine(affine, roi_coords)
+    out = np.zeros(len(streamlines), dtype=bool)
+    for ii, sl in enumerate(streamlines):
+        for coord in sl:
```
Garyfallidis: Hey bro @arokem! This looks very inefficient. And it is important that we can have this roi-filter functions as fast as possible. So, can I suggest that you write your "first" Cython code to increase the speed of this function? It would be great way to increase your Cython expertise. Are you in for the challenge?

MrBago:
Before I get into the details of the code, is it right that selecting on a list on includes is the same as selecting on the or of all those include and the same on the exclude? In other words are A and B below always the same if in_* and ex_* are boolean arrays?

```
include = [in_a, in_b, in_c]
exclude = [ex_a, ex_b, ex_c]
A = select(streamlines, include + exclude, [1, 1, 1, 0, 0, 0])

in_all = reduce(operator.or_, include)
ex_all = reduce(operator.or_, exclude)
B = select(streamlines, [in_all, ex_all], [0, 1])
```

dipy/tracking/utils.py
arokem: Is "first" in quotes because it's supposed to be ironic? 

Sure - I can try rewriting this in cython and testing to see how much of a win that would really be. I am worried that our code-base has a lot of cython, which is always a bit harder to maintain than python code, but it's certainly not going to be the largest piece of cython we have.

Garyfallidis: The "first" is in quotes because maybe you have already written many cythonized functions and I am not aware of it. Definitely not ironic. This is a critical functionality and it will be a bottleneck if we are not smart. I think Cythonizing those intersections with ROIs is a great idea and will not add any problems on maintaining the code. These are small functions. If you see that the speed improvement is not considerable then we talk again. But I think you will be surprised from the speedup.

arokem: OK - I've started this process, and will provide some more detailed profiling information down the line. I am happy to get feedback on what could be done in the cython code to speed things up even further.

arokem: What I have done so far seems to have resulted in only a very small speed-up. I am using the fornix bundle as streamlines: 

https://gist.github.com/arokem/51239fe55c07da101633

This level of cythonization takes me from about 730 (only python) to about 650 msec (python + cython), so about 10% speedup 

Happy to hear what else should be done in the cython code to really take advantage of it, and increase the speedup more.

jyeatman: Hi Ariel,

A use case to try is a huge ROI. For example take the fornix bundle (or a
few thousand random streamlines) and perform an operation using a gray
matter mask as an ROI. This is the case that will typically bog down the
computation and would hopefully show a substantial speed up.


Jason D. Yeatman, PhD
Research Scientist
Institute for Learning and Brain Sciences (ILABS)
University of Washington
http://jasonyeatman.com

MrBago:
I have 3 main things I would change about your PR:

1) I think that the problem of selecting on multiple ROIs can be reduced to selecting on a pair of ROIs, an include and an exclude. I think it would make the code easier to follow if we had a first function which took an iterable of streamlines, an include ROI and an exclude ROI and preformed the selection. I think we should also have a second function that takes a list of ROIs and either calls the first function (or maybe just returns a pair of "aggregate ROIs" which can be passed to the first, but I like this version less).

2) In terms of performance, comparing every point in every streamline to every point in every ROI is quite computationally expensive, O(P * N) where P is the number of points in all the streamlines and N is the total size of all the ROIs. I believe what we can do here is apply a binary_dilation (scipy.ndimage) to the input ROIs (I think the structure should be a circle with the appropriate radius). Then for each point, we just need to convert to voxel coordinates and look up a single point in the input ROI. The complexity here is roughly O(P + N) and I believe the overhead of the dilation will be very small compared to the benefit, but I'd like to profile it to know for sure.

3) So far in dipy, we've been using generators for streamlines. That is any function that returns streamlines has been returning a generator and any function that accepts streamlines as inputs has been designed and tested to work with generator inputs. I'm not opposed to changing this on principle, but I think selection should actually work very nicely with the generator model. The idea with the generators is that if I want to do the following and my file is very large, ~8GB, the streamlines will be fed into the final consumer (density_map in this case) 1 by 1 and the streamlines will never all be loaded into memory all at once. Here is the example:

streamlines = read_trk_file(filename, as_generator=True)
subset_streamlines = select(streamlines, ...)
dm = density_map(subset_streamlines, shape, affine)

Dipy's tracking files currently return generators and nibable's read function optionally returns a generator. Typically, any function that accepts generators as an argument can instead be passed a list without any changes. Also, any function which returns a generator can be wrapped with `list`, something like  `out = list(select(...))`, in order to force the output to be a list.

dipy/tracking/vox2track.pyx
```diff
@@ -296,3 +296,35 @@ def track_counts(tracks, vol_dims, vox_sizes=(1,1,1), return_elements=True):
...
+
+    Parameters
+    ----------
+    streamlines : list
+    x_roi_coords : array
+        ROI coordinates transformed to the streamline coordinate frame.
+    tol: float
+        Distance (in the units of the streamlines, usually mm). If any
+        coordinate in the streamline is within this distance from any voxel in
+        the ROI, the filtering criterion is set to True for this streamlin,
```
jchoude: Typo: streamlin -> streamline

jchoude:
I just wanted to quickly comment that I agree with @MrBago 's suggestion in point 2). This is what we do in internal tools, and if the dilation radius is correctly set, the result should be equivalent.

arokem:
Thanks for the comments @MrBago! I have no quibbles with points 1 and 3. I thought about number 2 as well, but was thinking in the direction of a spatial smoothing, and distance tolerance implemented through thresholding on the smoothed image. I agree that the order of the computation is more favorable, but not sure how this bears out in practice, with the data we usually analyze (P>>N, typically), so profiling would be good. 

@jchoude - could you share the code you were referring to? Would be useful to see what goes into setting the radius correctly.

MrBago:
@arokem I believe dialation is very close to what you have in this PR. Spacial smoothing might work, but it is going to be different than what you have here. As an example take a U shaped ROI, after smoothing the interior of the U will have higher intensity than the exterior at the same nominal distance.

Dialation using a cirlce structure returns all the regions within some radius of at least 1 true voxel in the input ROI, isn't that what you're going for?

dipy/segment/select.py
```diff
@@ -0,0 +1,77 @@
...
+    rois: list or ndarray
+        A list of 3D arrays, each with shape (x, y, z) corresponding to the
+        shape of the brain volume, or a 4D array with shape (n_rois, x, y,
+        z). Non-zeros in each volume are considered to be within the regio 
+
+    include: array
+        A 1D array of boolean marking inclusion or exclusion criteria. If a
+        streamline is near any of the inclusion ROIs, it should evaluate to
+        True, unless it is also near any of the exclusion ROIs. 
+
```
jchoude: Missing doc for affine and tol

jchoude:
@arokem Sadly, I think I initially read too quickly. I re-read what you're trying to do, and we don't have code that does exactly that. My bad, sorry for the noise.

OTOH, going through your code has raised a question in my head: if you set your `tol` parameter to 0, would you expect a streamline which has at least one point inside one of the voxel of the ROI to still be selected? As for myself, I would say yes. If you disagree, the following may not hold true.

When you transform the voxel coordinates (in your `near_roi` function) back to world coordinates, you effectively get the coordinates of the center of those voxels in world space. Then, when you compute the distance (in `_near_roi`), you compute the point to point distance. In that case, there is no explicit consideration of the voxel size.

That means that, if we set the `tol` to 0, then there must be at least one point of the streamline overlapping with the center of a voxel, else the streamline won't be selected. However, this goes against the intuition that there only needs to be a point of the streamline inside a voxel to be selected...

If you want to have this behavior (full voxel selection), that means that either
1- the `tol` parameter needs to be better documented to explicitly say that the voxel size should be included
2- the voxel size should be internally used in the distance threshold
3- an approach based on something derived from the `target` function might be better suited to this task.

Personally, I don't like 1 and 2. For 1, it might be confusing for users. For 2, it may be complex to handle the case where voxels are anisotropic.

Maybe having an approach based on the `target` function, and possibly a supersampled grid and mask dilation might be another way to go.

I know that this is a long comment (probably my longest up to today ;)). If it's not clear, please let me know. If you disagree, disregard this :)

arokem:
I am happy that I managed to get your interest up enough to review this PR.
I do appreciate the thought everyone is putting into this. It is going to
be an important piece in something that we hope to use a lot, so it's
important to get it right: conceptually, and in terms of implementation.

Concerning the `tol` kwarg: I actually think that the most natural units
for this are not in voxel size, but in mm. I agree that the default value
is somewhat misleading, because we will probably never use this with a
tolerance of 0. How about setting this per default to `None` and then
extracting voxel size from the affine, and including things that are within
the voxels of the ROI? The simplest would be to treat voxels as though they
are spheres with some radius for this purpose. Otherwise, for simplicity,
we can also just set this to some 'typical' number of mm, e.g. 1 mm.

Concerning dilation: I am not sure this would do what we want, unless we
resample the ROI to a fine resolution, that depends somehow on the
tolerance value. Imagine for example that you want to include streamlines
that have nodes that are <=1.5 voxel-size units away from the center of a
voxel, but not those streamlines passing at 1.6 voxel units away. You would
have to somehow resample the ROI to 0.1 of the original size and proceed to
dilate by 5 steps out from the edges of the ROI.

jchoude:
Hey @arokem 

ok, now I clearly understand your definition of the `tol` arg. In that case, it needs to be clearly documented, and a default value as you are suggesting would be interesting.

As for the strategy of considering the voxels as spheres with a radius, that would fit perfectly with your current code. I guess at that point that's more of a "philosophical" choice... Personally, I know that if I have a ROI taken from either an automatic segmentation or drawn in any software, and I load it in a visualization tool and check the streamlines that are selected by my ROI, I would expect to get all streamlines going through voxels included in my ROI. Using a sphere with a radius + the tolerance would get almost the same result, with possibly some false positives or negatives.

At the point, it's mostly a question of perfect precision versus efficiency. I'll let you make the call :smiley:

MrBago:
@arokem I understand your concern and see some of the advantages of your current definition. I believe both definitions have some issues. For example if you think of voxels as spheres, and set that diameter to be the edge length of one voxel, you'll end up with a grid of non-overlapping spheres and you'll miss points that lie in the space between these spheres. The result is that your ROI will have internal gaps. To avoid gaps in a 3D ROI you'll need to set tol to be >= `sqrt(sum((voxel_edges / 2)**2))`.

The dilation approach has two related problems. The first you clearly pointed out, tol ends up being coerced into multiples of the voxel size. The second is that this multiple ends up depending on orientation. For example an roi cented at (0, 0) with tol=1 will reject points at (1.001, 0), (0, 1.001) and (1.001, 1.001). The first two are > 1 from (0, 0) but the last is a distance > sqrt(2) from (0, 0).

I would argue that both the issues with the dilation method are fundamental to representing ROIs as masks. I believe the select method should only be as precise as it's inputs. The ROI input is already "discretized", for example if I want an ROI that follows the freesurfer pial surface, I need to "discretize" it into a mask in order to pass it to the select method. I would argue that once this is done, trying to create artificial precision is not very useful.

That being said, I've wanted to create a way of representing ROIs as closed 3D surfaces for some time, but when I bring it up people tend to run away screaming :).

arokem:
Yes - I think that enforcing a minimal tolerance so that the resulting ROI
covers the entire mask is reasonable. Another possibility is to use the
edges of the voxels as the frame of reference, rather than the centers. Let
me think about this a bit more, and integrate all these comments into
something coherent.

By the way - I am not running away screaming from the notion of
representing ROIs as closed surfaces. I think that we need that for
visualization purposes anyway: creating triangular meshes from ROIs via a
delaunay triangulation. There are some tricky issues to be tackled there,
to be sure, but it's eminently doable. Do you think we should start with
that first, before doing this here? I am not sure how it would solve all
the problems we have here. That is, you would still need to bound on
distance from the mesh somehow.

arokem:
OK - we've decided to take the middle road here: we stick to the tolerance-setting approach (little spheres around the centers of voxels), but we are enforcing a minimal reasonable tolerance that prevents users from using this with gaps in their ROI.

arokem:
We still need to do: 

[] Implement this as a generator 
[] Use only the streamline endpoints for testing 'nearness'
[] Sort out how to use numpy better for the logical operations (`select` and `reduce`)

MrBago:
@arokem I think you missed my point about the logical operations. My suggestion is that you merge all the include rois and exclude rois into two rois (agg_include and agg_exlude). Then you just need to write a function that deals with two arrays instead of two lists of arrays, that'll make the code easier to read and hopefully faster.

arokem:
Hmm. We do aggregate them inside the function into `to_include` and
`to_exclude` arrays. Is that what you meant?

MrBago:
Take a look at this gist, https://gist.github.com/MrBago/9c5e957cf13c7f37a7b9.

arokem:
OK - I think that I have completed the things I was hoping to do in this PR. The last few commits implement the selection functionality proposed by @MrBago, returns a generator of streamlines from the selection function, and then also allows an option where only the endpoints of the streamline are considered as criteria for selection. Finally, that last commit speeds up things by another 30% or so, in the timing benchmarks I tested. Anything more to do here, before this can be merged?

mdesco:
Nice work guys! This really looks awesome and of great need for Dipy.

A related question or request that could go in a separate PR but I was wondering if any of you were already thinking about it. 

One powerful thing that there is in TrackVis right now and track_vis (command line) version is to select streamlines going through a ROI that connects "either_end" or "both_ends" of the streamlines. Right now, you guys have implemented the "any part" or "no part" of a streamline. The either or both ends are quite powerful when you want to select things ending in a region. 

Could that be added as a flag to the current functions we have in this PR?
Also, you might be interested to have a look at the latest tck2connectome (https://github.com/MRtrix3/mrtrix3/wiki/tck2connectome) in Mrtrix 3, they have cool ways to deal with the end points radius, order, etc. This is also very crucial as you approach the cortex or subcortical regions. 

As I said, this might be too much for this PR. I just thought I would bring it up now because it is related to the current PR. Otherwise, I think this PR is ready to be merged.

Cheers

arokem:
Hey @mdesco - thanks for the feedback! 

I believe that we basically already have the `either_end` covered in the last few commits, and I just need to finish the implementation so that `select` also exposes this functionality. I will definitely do this here. It shouldn't be too hard to extend this to have the `both_end` criterion as well, and I can see why that would also be useful. I will do that here in this PR, so please don't merge yet (will change the title to WIP, just to be sure). 

Also: @MrBago - have not tested this with generator input, so need to do that as well, before proceeding with this.

arokem:
TODO list:

- [ ] Implement both `either_end` and `both_end` kwargs all the way from `select` down. 
- [ ] Test `select` with generator inputs.

arokem:
This is now substantially refactored to make it work with generator input and produce generator output. This required some separation of `near_roi` and `select_from_roi`, because creating the boolean value in `near_roi` required consuming the generator. I have instead implemented the inner cython loop on a streamline by streamline basis. As it turns out, this is actually faster than iterating of streamlines in cython. I've also managed to make this go a bit faster, by implementing distance as a c function. 

I have implemented four different modes: "all", "any", "either_end" and "both_end". The selection function `select_from_roi` allows setting these modes, but the same mode should be used for all ROIs. 

I've added some documentation examples showing how to use this.

arokem:
Any further comments here? Or is this ready to go?

arokem:
Re-rebased. Any further comments here?

Garyfallidis:
Yeah, give me a couple of days.

arokem:
Actually, please hold off on that. @rfdougherty just pointed out to me a few tips that will make this go faster. I'm at the scipy conference, so I will only get to this some time next week. Sorry about the false alarm

arokem:
Specifically, we should use the KDTree algorithm to find the near points in each streamline: 

http://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.spatial.KDTree.html

arokem:
Or this: 

http://docs.scipy.org/doc/scipy-0.15.1/reference/generated/scipy.spatial.distance.cdist.html

rfdougherty:
I think refactoring near_roi to use cdist will make a substantial improvement in speed. In my testing with this toy implementation:

def near_roi(sl, x_roi_coords, tol):
    Y = cdist(sl, x_roi_coords, 'euclidean')
    return Y.min()<tol

the speed-up is about 1000x.

rfdougherty:
Also, pre-filtering the streamlines by the ROI bounding box may improve speed (at least for compact ROIs). E.g., in select_by_roi:

    R = affine[0:3,0:3]
    vox_dim = np.diag(np.linalg.cholesky(R.T.dot(R)))
    bb = np.array((x_include_roi_coords.min(axis=0)-vox_dim,    
                   x_include_roi_coords.max(axis=0)+vox_dim))
    streamlines = [s for s in streamlines 
                   if(np.any(s[:,0]>=bb[0,0])&np.any(s[:,0]<=bb[1,0])&
                      np.any(s[:,1]>=bb[0,1])&np.any(s[:,1]<=bb[1,1])&
                      np.any(s[:,2]>=bb[0,2])&np.any(s[:,2]<=bb[1,2]))]

For my compact ROIs (e.g., the amygdala) this results in an additional 3x speed increase. However, if the ROI has a very large bounding box this will probably slow things down a bit. So we may want a heuristic in there to check the bb and skip this if it's big.

arokem:
Thanks for the tip @rfdougherty! Implemented the new `_near_roi` with all the different modes, and tests pass on my end. Still need to chew on the bounding box idea.

rfdougherty:
Also, the dist_to_corner calculation isn't robust. As implemented, it will only work for affines with no rotation component. Try this:

    R = affine[0:3,0:3]
    vox_dim = np.diag(np.linalg.cholesky(R.T.dot(R)))
    dist_to_corner = np.sqrt(np.sum((vox_dim / 2) ** 2))

arokem:
OK - I have implemented the more robust calculation of the distance to the corner. I am working on filtering by a bounding box, but it might be something that we do separately when we put these things together in a more elaborate pipeline.

arokem:
For the time being, I have implemented the filtering. But not sure how to integrate it with the rest. It doesn't quite fit in with the logic of the different modes (except maybe 'any').

dipy/tracking/utils.py
```diff
@@ -568,6 +574,149 @@ def target(streamlines, target_mask, affine, include=True):
             yield sl
 
 
+def sl_near_roi(sl, x_roi_coords, tol, mode='any'):
```
Garyfallidis: Please say streamlines_near_roi not sl_near_roi.

dipy/tracking/utils.py
```diff
@@ -568,6 +574,149 @@ def target(streamlines, target_mask, affine, include=True):
...
+    """
+    Is a streamline near an ROI.
+
+    Implements the inner loops of the :func:`near_roi` function.
+
+    Parameters
+    ----------
+    sl : array, shape (N, 3)
+	    A single streamline
+    x_roi_coords : array, shape (M, 3)
```
Garyfallidis: Why not say roi_coors? why x_ is useful?

dipy/tracking/utils.py
```diff
@@ -568,6 +574,149 @@ def target(streamlines, target_mask, affine, include=True):
...
+
+    dist = cdist(s, x_roi_coords, 'euclidean')
+
+    if mode=="any" or mode=="either_end":
+        return np.min(dist)<=tol
+    else:
+        return np.all(np.min(dist, -1)<=tol)
+
+
+def filter_by_bb(streamlines, x_roi_coords, tol):
```
Garyfallidis: filter_bounding_box

dipy/tracking/utils.py
```diff
@@ -568,6 +574,149 @@ def target(streamlines, target_mask, affine, include=True):
...
+    """
+    Filter streamlines using a bounding box extracted from the coordinates of
+    an ROI.
+
+    Parameters
+    ----------
+    streamlines : list or generator
+        A sequence of streamlines. Each streamline should be a (N, 3) array,
+        where N is the length of the streamline.
+    x_roi_coords :
```
Garyfallidis: description missing

dipy/tracking/utils.py
arokem: Done, but in singular. This only looks at one streamline at a time.

dipy/tracking/utils.py
arokem: It's just to say that the coordinates have been transformed (or 'x-formed'...) to the streamline coordinate frame.

dipy/tracking/utils.py
arokem: Done

dipy/tracking/utils.py
arokem: Whoops. Thanks for noticing. Added.

dipy/segment/select.py
```diff
@@ -0,0 +1,174 @@
+"""
+Select streamlines
```
Garyfallidis: Not sure that we need a new module for this.
I think for standard processing for streamlines like this the right place is 
``dipy.tracking.streamline``

dipy/segment/select.py
```diff
@@ -0,0 +1,174 @@
...
+
+from warnings import warn
+import numpy as np
+from nibabel.affines import apply_affine
+from dipy.tracking.utils import streamline_near_roi
+from dipy.core.geometry import dist_to_corner
+
+
+def reduce_rois(rois, include):
+    """
```
Garyfallidis: Why not start the docstring from the first line? We save one line space.

arokem: Moved the function into tracking.streamline and eliminated the unnecessary white-space.

arokem:
OK - removed the new module, and moved the new function into `tracking.streamline`.

arokem:
Hey @Garyfallidis - I think I've addressed all your comments. Anything more to do here?

Garyfallidis:
Sorry, I haven't finished reviewing this. One problem I see is that it is not linked to any of the relevant tutorials.

arokem:
I think that a tutorial at a later stage would be better. That is, when we
fetchers/loaders for some standard ROIs that we can demonstrate this with.
This will be included in an upcoming PR. Does that seem OK?

arokem:
Anything more here? The ROIs mentioned in the last exchange here are already in #681, and the tutorial is basically already written here: https://github.com/arokem/AFQ-notebooks/blob/master/AFQ-registration-callosum.ipynb

Garyfallidis:
Sorry for the delay. Will look at this asap!

dipy/tracking/utils.py
```diff
@@ -568,6 +573,150 @@ def target(streamlines, target_mask, affine, include=True):
...
+        coordinate in the streamline is within this distance from the center
+        of any voxel in the ROI, this function returns True.
+
+
+    """
+    bb = np.array((np.min(x_roi_coords, 0) - tol,
+                   np.max(x_roi_coords, 0) + tol))
+    for sl in streamlines:
+        cond = np.array([np.any(sl[:, i] >= bb[0, i]) &
+                         np.any(sl[:, i] <= bb[1, i]) for i in [0, 1]])
```
rfdougherty: Shouldn't this be 'for i in [0,1,2]'?

arokem: Yep. Good catch. Fixed!

arokem:
I'm going to punt the filtering-by-bounding-box stuff to another PR. This is now clean, rebased, and hopefully all passing.

arokem:
Ready for your review @Garyfallidis

dipy/core/geometry.py
```diff
@@ -1006,3 +1006,22 @@ def perpendicular_directions(v, num=30, half=False):
...
+    given an affine
+
+    Parameters
+    ----------
+    affine : 4 by 4 array.
+        The spatial transformation from the measurement to the scanner space.
+
+    Returns
+    -------
+    float, the maximal distance to the
```
Garyfallidis:
```
dist : float
  the maximal distance to the ...
```

dipy/core/geometry.py
```diff
@@ -1006,3 +1006,22 @@ def perpendicular_directions(v, num=30, half=False):
                              (v[0]*cosa - v[2]*v[1]*sina) / sq])
 
     return psamples.T
+
+
+def dist_to_corner(affine):
+    """
```
Garyfallidis: Please move Calculate to the first line of the docstring.

dipy/tracking/utils.py
```diff
@@ -568,6 +573,122 @@ def target(streamlines, target_mask, affine, include=True):
             yield sl
 
 
+def streamline_near_roi(sl, x_roi_coords, tol, mode='any'):
+    """
+    Is a streamline near an ROI.
```
Garyfallidis: move one line up please

dipy/tracking/utils.py
Garyfallidis: Please use the full name. transformed_roi or just say roi_coords or roi. Of course the roi has to be in the same space with the streamlines. x is misleading

dipy/tracking/utils.py
```diff
@@ -568,6 +573,122 @@ def target(streamlines, target_mask, affine, include=True):
...
+    ----------
+    sl : array, shape (N, 3)
+	    A single streamline
+    x_roi_coords : array, shape (M, 3)
+        ROI coordinates transformed to the streamline coordinate frame.
+    tol : float
+        Distance (in the units of the streamlines, usually mm). If any
+        coordinate in the streamline is within this distance from the center
+        of any voxel in the ROI, this function returns True.
+	mode : string
```
Garyfallidis: indentation issue

dipy/tracking/utils.py
```diff
@@ -568,6 +573,122 @@ def target(streamlines, target_mask, affine, include=True):
...
 
+def streamline_near_roi(sl, x_roi_coords, tol, mode='any'):
+    """
+    Is a streamline near an ROI.
+
+    Implements the inner loops of the :func:`near_roi` function.
+
+    Parameters
+    ----------
+    sl : array, shape (N, 3)
```
Garyfallidis: Say streamline not sl.

dipy/tracking/utils.py
```diff
@@ -568,6 +573,122 @@ def target(streamlines, target_mask, affine, include=True):
...
+        the target region.
+    affine : ndarray
+        Affine transformation from voxels to streamlines. Default: identity.
+    tol : float
+        Distance (in the units of the streamlines, usually mm). If any
+        coordinate in the streamline is within this distance from the center
+        of any voxel in the ROI, the filtering criterion is set to True for
+        this streamline, otherwise False. Defaults to the distance between
+        the center of each voxel and the corner of the voxel.
+	mode : string, optional
```
Garyfallidis: indent

dipy/tracking/utils.py
```diff
@@ -568,6 +573,122 @@ def target(streamlines, target_mask, affine, include=True):
...
+    """
+    Provide filtering criteria for a set of streamlines based on whether they
+    fall within a tolerance distance from an ROI
+
+    Parameters
+    ----------
+    streamlines : list or generator
+        A sequence of streamlines. Each streamline should be a (N, 3) array,
+        where N is the length of the streamline.
+    target_mask : ndarray
```
Garyfallidis: You mean ROI? You use ROI in the description but then you say ``target_mask``. Maybe chose one of the two?

dipy/tracking/utils.py
```diff
@@ -568,6 +573,122 @@ def target(streamlines, target_mask, affine, include=True):
...
+    if mode=="any" or mode=="either_end":
+        return np.min(dist)<=tol
+    else:
+        return np.all(np.min(dist, -1)<=tol)
+
+
+def near_roi(streamlines, target_mask, affine=None, tol=None,
+             mode="any"):
+    """
+    Provide filtering criteria for a set of streamlines based on whether they
```
Garyfallidis: one line up

dipy/tracking/utils.py
```diff
@@ -568,6 +573,122 @@ def target(streamlines, target_mask, affine, include=True):
...
+        Affine transformation from voxels to streamlines. Default: identity.
+    tol : float
+        Distance (in the units of the streamlines, usually mm). If any
+        coordinate in the streamline is within this distance from the center
+        of any voxel in the ROI, the filtering criterion is set to True for
+        this streamline, otherwise False. Defaults to the distance between
+        the center of each voxel and the corner of the voxel.
+	mode : string, optional
+		One of {"any", "all", "either_end", "both_end"}, where return True if:
+		"any" : any point is within tol from ROI. Default.
```
Garyfallidis: This will probably not render correctly. Because ``:`` is reserved in the docstrings for parameters.

dipy/tracking/tests/test_utils.py
```diff
@@ -248,6 +248,80 @@ def test_target():
...
+    # Test for use of the 'all' mode, also testing that setting the tolerance
+    # to a very small number gets overridden:
+    assert_array_equal(near_roi(x_streamlines, mask, affine=affine, tol=0.1,
+                                mode='all'), np.array([False, True, False]))
+
+    mask[2, 2, 2] = True
+    mask[3, 3, 3] = True
+    assert_array_equal(near_roi(x_streamlines, mask, affine=affine, tol=None,
+                                mode='all'),
+                 np.array([False, True, True]))
```
Garyfallidis: pep 8.

dipy/tracking/tests/test_utils.py
```diff
@@ -248,6 +248,80 @@ def test_target():
...
+                             [3, 3, 3]])
+                  ]
+
+    affine = np.eye(4)
+    mask = np.zeros((4, 4, 4), dtype=bool)
+    mask[0, 0, 0] = True
+    mask[1, 0, 0] = True
+
+    assert_array_equal(near_roi(streamlines, mask, tol=1),
+                 np.array([True, True, False]))
```
Garyfallidis: pep8

dipy/tracking/tests/test_utils.py
```diff
@@ -248,6 +248,80 @@ def test_target():
...
+    assert_array_equal(near_roi(streamlines, mask),
+                 np.array([False, True, False]))
```
Garyfallidis: pep 8

dipy/tracking/streamline.py
```diff
@@ -113,3 +116,165 @@ def select_random_set_of_streamlines(streamlines, select):
...
+        z). Non-zeros in each volume are considered to be within the region
+    include: array or list
+        A list or 1D array of boolean values marking inclusion or exclusion
+        criteria. If a streamline is near any of the inclusion ROIs, it
+        should evaluate to True, unless it is also near any of the exclusion
+        ROIs.
+    mode : string, optional
+		One of {"any", "all", "either_end", "both_end"}, where a
+            streamline is associated with an ROI if:
+		"any" : any point is within tol from ROI. Default.
```
Garyfallidis: Yeah I think this will not render correctly too. Can you check?

dipy/tracking/streamline.py
```diff
@@ -113,3 +116,165 @@ def select_random_set_of_streamlines(streamlines, select):
...
+
+    See also
+    --------
+    :func:`dipy.tracking.utils.near_roi`
+    :func:`reduce_rois`
+
+    Examples
+    --------
+    >>> streamlines = [np.array([[0, 0., 0.9],
+    ...                                  [1.9, 0., 0.]]),
```
Garyfallidis: indent

dipy/tracking/streamline.py
```diff
@@ -113,3 +116,165 @@ def select_random_set_of_streamlines(streamlines, select):
...
+           [ 0.,  1.,  1.],
+           [ 0.,  2.,  2.]])]
+    >>> selection = select_by_roi(streamlines, [mask1, mask2], [True, False],
+    ...                           tol=0.87)
+    >>> list(selection)
+    [array([[ 0.,  0.,  0.],
+           [ 0.,  1.,  1.],
+           [ 0.,  2.,  2.]])]
+    >>> selection = select_by_roi(streamlines, [mask1, mask2],
+    ...                      [True, True],
```
Garyfallidis: indent

dipy/tracking/streamline.py
```diff
@@ -113,3 +116,165 @@ def select_random_set_of_streamlines(streamlines, select):
...
+    ...                      mode="both_end",
+    ...                      tol=1.0)
+    >>> list(selection)
+    [array([[ 0. ,  0. ,  0.9],
+           [ 1.9,  0. ,  0. ]])]
+    >>> mask2[0, 2, 2] = True
+    >>> selection = select_by_roi(streamlines, [mask1, mask2],
+    ...                      [True, True],
```
Garyfallidis: indent

dipy/tracking/streamline.py
```diff
@@ -113,3 +116,165 @@ def select_random_set_of_streamlines(streamlines, select):
...
+    """
+    if affine is None:
+        affine = np.eye(4)
+    # This calculates the maximal distance to a corner of the voxel:
+    dtc = dist_to_corner(affine)
+    if tol is None:
+        tol = dtc
+    elif tol < dtc:
+        w_s = "Tolerance input provided would create gaps in your"
+        w_s += " inclusion ROI. Setting to: %s"%dist_to_corner
```
Garyfallidis: pep8

dipy/tracking/streamline.py
```diff
@@ -113,3 +116,165 @@ def select_random_set_of_streamlines(streamlines, select):
...
+    include_roi_coords = np.array(np.where(include_roi)).T
+    x_include_roi_coords = apply_affine(affine, include_roi_coords)
+    exclude_roi_coords = np.array(np.where(exclude_roi)).T
+    x_exclude_roi_coords = apply_affine(affine, exclude_roi_coords)
+
+    if mode is None:
+        mode = "any"
+    for sl in streamlines:
+        include = streamline_near_roi(sl, x_include_roi_coords, tol=tol,
+                                mode=mode)
```
Garyfallidis: pep8

dipy/tracking/streamline.py
```diff
@@ -113,3 +116,165 @@ def select_random_set_of_streamlines(streamlines, select):
...
+        exclude = streamline_near_roi(sl, x_exclude_roi_coords, tol=tol,
+                                mode=mode)
```
Garyfallidis: pep 8

Garyfallidis:
Hi @arokem. I am done with my review here. Move on with the corrections and will happily merge. 
One problem that I found with this PR is that I don't see any timings. We need those filters to be quite efficient. If they are then I can use them directly in the new viz module. 
Keep it up!

arokem:
OK - I think that addresses all of your comments. 

There's a bit of timing information in this notebook: 

https://github.com/arokem/AFQ-notebooks/blob/master/AFQ-registration-callosum.ipynb

That's filtering ~1M streamlines in about 90 seconds (on my laptop). It's not quite fast enough for online filtering for visualization (I think?), but not too bad for segmentation purposes.

rfdougherty:
I think that's great performance! For faster online filtering, I think we'd need to go to surface-based ROI definitions and fast collision detection algorithms. Not a small feat. We did that in CINCH/Quench (http://white.stanford.edu/newlm/index.php/QUENCH) and were able to get intersections computed for up to ~1M streamlines fast enough for interactive touch-based streamline editing.

arokem:
Yeah - thanks for all the tips on that. The idea of surface-based ROIs did come up earlier in this very PR - where can I find the relevant CINCH/Quench code to refer to (yes, I know it won't be pleasant ;D)?

rfdougherty:
The code is in an old svn repo: svn co http://white.stanford.edu/repos/StanfordTools-DWI and uses an ancient collision detection library (RAPID). I think a modern game engine would be a better choice for doing fast collision detection (e.g., https://www.panda3d.org/).

arokem:
Hey @Garyfallidis - did this address all your comments?

Garyfallidis:
I had some comments saying that some of the docstrings may not render correctly. Have you checked those?

Garyfallidis:
Also an easy way to make this much faster is by having an initial QuickBundles version of the tractography and immediately remove the clusters that are quite far from the rois by a threshold. That would reduce the search space incredibly! QuickBundles for ever baby! ;)

Garyfallidis:
The clusters can be removed by checking only the points/segments of the centroids of the clusters. You will need to set a distance threshold. Usually 15mm will be good enough for what you want.

arokem:
I did not run the doc build until today (see #702 and #701...), but I did change the docstrings where you pointed out that there are issues.

arokem:
For the time being my tex installations seems to have trouble with our docs. Still debugging.

dipy/tracking/streamline.py
arokem: Whoops - missed this one. Just fixed it as well.

arokem:
From what I can tell, The API documentation of `tracking.utils` is not in our current docs either:  http://nipy.org/dipy/reference/dipy.tracking.html

arokem:
OK - I have checked rendering of the docs. There's no problem using the colon (`:`) in the docstrings, so I reverted some of the recent changes suggested by @Garyfallidis. I also made some additional changes in these docstrings so that they are more readable in the html version of the docs.

Is this now ready to go?

arokem:
Yeah - that's a great idea. We'll take that when we write up the more
complete AFQ pipeline.

arokem:
Hi @Garyfallidis - is this ready to go? I think that the recent commits address all your formatting concerns, and I've checked that this renders properly into html pages. Anything else need doing here?

Garyfallidis:
My plan is later today after I finish working with Julio to have these and another two pending PRs merged.

arokem:
Cool. Looking forward to it.

dipy/tracking/streamline.py
```diff
@@ -113,3 +116,169 @@ def select_random_set_of_streamlines(streamlines, select):
...
+
+def select_by_roi(streamlines, rois, include, mode=None, affine=None,
+                  tol=None):
+    """Select streamlines based on logical relations with several regions of
+    interest (ROIs). For example, select streamlines that pass near ROI1,
+    but only if they do not pass near ROI2.
+
+    Parameters
+    ----------
+    streamlines: list
```
Garyfallidis: space missing
```
streamlines : list
```

dipy/tracking/streamline.py
```diff
@@ -113,3 +116,169 @@ def select_random_set_of_streamlines(streamlines, select):
...
+        A list of candidate streamlines for selection
+    rois: list or ndarray
```
Garyfallidis: same here

dipy/tracking/streamline.py
```diff
@@ -113,3 +116,169 @@ def select_random_set_of_streamlines(streamlines, select):
...
+        A list of 3D arrays, each with shape (x, y, z) corresponding to the
+        shape of the brain volume, or a 4D array with shape (n_rois, x, y,
+        z). Non-zeros in each volume are considered to be within the region
+    include: array or list
```
Garyfallidis: same

dipy/tracking/utils.py
```diff
@@ -568,6 +573,128 @@ def target(streamlines, target_mask, affine, include=True):
...
+
+		"any" : any point is within tol from ROI.
+
+        "all" : all points are within tol from ROI.
+
+        "either_end" : either of the end-points is within tol from ROI
+
+        "both_end" : both end points are within tol from ROI.
+
+	Returns
```
Garyfallidis: Indentation problem.

dipy/core/tests/test_geometry.py
```diff
@@ -288,6 +289,34 @@ def test_perpendicular_directions():
...
+                            [-np.sin(r[2]), np.cos(r[2]), 0],
+                            [0, 0, 1]]))
+    R = np.linalg.inv(R)
+    return R
+
 
+def test_dist_to_corner():
+    affine = np.eye(4)
+    # Calculate the distance with the pythagorean theorem:
+    pythagoras = np.sqrt(np.sum((np.diag(affine)[:-1] / 2) ** 2))
```
Garyfallidis: nice guy ;)

dipy/tracking/streamline.py
```diff
@@ -113,3 +116,169 @@ def select_random_set_of_streamlines(streamlines, select):
...
+    """Reduce multiple ROIs to one inclusion and one exclusion ROI
+
+    Parameters
+    ----------
+    rois : list or ndarray
+        A list of 3D arrays, each with shape (x, y, z) corresponding to the
+        shape of the brain volume, or a 4D array with shape (n_rois, x, y,
+        z). Non-zeros in each volume are considered to be within the region.
+
+    include: array or list
```
Garyfallidis: space

dipy/tracking/streamline.py
```diff
@@ -113,3 +116,169 @@ def select_random_set_of_streamlines(streamlines, select):
     len_s = len(streamlines)
     index = np.random.randint(0, len_s, min(select, len_s))
     return [streamlines[i] for i in index]
+
+
+def reduce_rois(rois, include):
```
Garyfallidis: It is not clear to me which functions should be in the ``dipy.tracking.streamline`` and which in ``dipy.tracking.utils``. How did you decide which should go where? Did you have something specific in mind?

Garyfallidis: Definitely ``select_by_roi`` should be in streamline. Not sure about ``reduce_rois``. Feels like an utils... function.

Also I recommend to change the name to ``select_by_rois``. Should also this function replace Bago's ``target`` at some point?

Garyfallidis:
Hi @arokem nearly done here. A few minor points and docstring corrections that you need to address. Let me know when you are done.

dipy/tracking/utils.py
arokem: So strange - I see it here, but not in my text editor. Hopefully fixed now.

dipy/tracking/streamline.py
arokem: OK - moved `reduce_rois` to utils.

Seems logical to have `select_by_rois` in `streamline`, right next to `select_random_set_of_streamlines` (and that's also the comment you made here on July 20th), and it makes sense to have `streamline_near_roi` in `utils`, together with `target`. I would not replace `target` with that other function, considering that they do slightly different things, and have slightly different use-cases.

arokem:
OK - hopefully we got it all this time. I think I've addressed that recent wave of comments (but let me know if you still think functions need to be moved around further), and I've added a couple of tests for reduce_rois on its own.

dipy/tracking/streamline.py
```diff
@@ -114,3 +118,129 @@ def select_random_set_of_streamlines(streamlines, select):
...
+    ...                          [0, 1., 1.],
+    ...                          [0, 2., 2.]]),
+    ...                np.array([[2, 2, 2],
+    ...                          [3, 3, 3]])]
+    >>> mask1 = np.zeros((4, 4, 4), dtype=bool)
+    >>> mask2 = np.zeros_like(mask1)
+    >>> mask1[0, 0, 0] = True
+    >>> mask2[1, 0, 0] = True
+    >>> selection = select_by_rois(streamlines, [mask1, mask2], [True, True],
+    ...                           tol=1)
```
Garyfallidis: Here the tol parameter is not aligned with streamlines.

dipy/tracking/streamline.py
```diff
@@ -114,3 +118,129 @@ def select_random_set_of_streamlines(streamlines, select):
...
+    >>> list(selection) # The result is a generator
+    [array([[ 0. ,  0. ,  0.9],
+           [ 1.9,  0. ,  0. ]]), array([[ 0.,  0.,  0.],
+           [ 0.,  1.,  1.],
+           [ 0.,  2.,  2.]])]
+    >>> selection = select_by_rois(streamlines, [mask1, mask2], [True, False],
+    ...                           tol=0.87)
```
Garyfallidis: same

dipy/tracking/streamline.py
```diff
@@ -114,3 +118,129 @@ def select_random_set_of_streamlines(streamlines, select):
...
+    >>> list(selection)
+    [array([[ 0.,  0.,  0.],
+           [ 0.,  1.,  1.],
+           [ 0.,  2.,  2.]])]
+    >>> selection = select_by_rois(streamlines, [mask1, mask2],
+    ...                           [True, True],
```
Garyfallidis: same

dipy/tracking/streamline.py
```diff
@@ -114,3 +118,129 @@ def select_random_set_of_streamlines(streamlines, select):
...
+    ...                           mode="both_end",
+    ...                           tol=1.0)
+    >>> list(selection)
+    [array([[ 0. ,  0. ,  0.9],
+           [ 1.9,  0. ,  0. ]])]
+    >>> mask2[0, 2, 2] = True
+    >>> selection = select_by_rois(streamlines, [mask1, mask2],
+    ...                           [True, True],
```
Garyfallidis: same

dipy/tracking/utils.py
```diff
@@ -568,6 +573,128 @@ def target(streamlines, target_mask, affine, include=True):
...
+		"any" : any point is within tol from ROI.
+
+        "all" : all points are within tol from ROI.
+
+        "either_end" : either of the end-points is within tol from ROI
+
+        "both_end" : both end points are within tol from ROI.
+
+    Returns
+	-------
```
Garyfallidis: underline is off

Garyfallidis:
A few more things @arokem

dipy/tracking/tests/test_streamline.py
```diff
@@ -544,5 +546,105 @@ def test_compress_streamlines():
...
+    npt.assert_array_equal(list(selection), [streamlines[0],
+                                            streamlines[1]])
+
+    selection = select_by_rois(streamlines, [mask1, mask2], [True, False])
+
+    npt.assert_array_equal(list(selection), [streamlines[1]])
+
+    # Setting tolerance too low gets overridden:
+    selection = select_by_rois(streamlines, [mask1, mask2], [True, False],
+                              tol=0.1)
```
Garyfallidis: tol not aligned (pep8)

dipy/tracking/tests/test_streamline.py
```diff
@@ -544,5 +546,105 @@ def test_compress_streamlines():
...
+    selection = select_by_rois(streamlines, [mask1], [True], tol=0.87)
+    npt.assert_array_equal(list(selection), [streamlines[1]])
+
+    selection = select_by_rois(streamlines, [mask1], [True], tol=1.0)
+    npt.assert_array_equal(list(selection), [streamlines[0],
+                                            streamlines[1]])
+
+    # Use different modes:
+    selection = select_by_rois(streamlines, [mask1, mask2, mask3],
+                              [True, True, False],
```
Garyfallidis: same

dipy/tracking/tests/test_streamline.py
```diff
@@ -544,5 +546,105 @@ def test_compress_streamlines():
...
+                              mode="all",
+                              tol=1.0)
+    npt.assert_array_equal(list(selection), [streamlines[0]])
+
+
+    selection = select_by_rois(streamlines, [mask1, mask2, mask3],
+                              [True, True, False],
```
Garyfallidis: same

dipy/tracking/tests/test_streamline.py
```diff
@@ -544,5 +546,105 @@ def test_compress_streamlines():
...
+                              mode="either_end",
+                              tol=1.0)
+    npt.assert_array_equal(list(selection), [streamlines[0]])
+
+    selection = select_by_rois(streamlines, [mask1, mask2, mask3],
+                              [True, True, False],
```
Garyfallidis: same

dipy/tracking/tests/test_streamline.py
```diff
@@ -544,5 +546,105 @@ def test_compress_streamlines():
...
+                              mode="both_end",
+                              tol=1.0)
+    npt.assert_array_equal(list(selection), [streamlines[0]])
+
+    mask2[0, 2, 2] = True
+    selection = select_by_rois(streamlines, [mask1, mask2, mask3],
+                              [True, True, False],
```
Garyfallidis: same

Garyfallidis:
Hey man. I think the editor you are currently using does not give you correct indications for pep8.

arokem:
OK - done.

dipy/tracking/tests/test_utils.py
```diff
@@ -248,6 +250,79 @@ def test_target():
...
+                                mode="either_end"),
+                        np.array([True, False, False]))
+
+    assert_array_equal(near_roi(streamlines, mask, tol=0.87, mode="both_end"),
+                        np.array([False, False, False]))
+
+    mask[0, 0, 0] = True
+    mask[0, 2, 2] = True
+    assert_array_equal(near_roi(streamlines, mask, mode="both_end"),
+                        np.array([False, True, False]))
```
Garyfallidis: Nope you still have styling issues. Look here.

dipy/tracking/tests/test_utils.py
```diff
@@ -248,6 +250,79 @@ def test_target():
...
+    mask = np.zeros((4, 4, 4), dtype=bool)
+    mask[0, 1, 1] = True
+    mask[3, 2, 2] = True
+
+    assert_array_equal(near_roi(streamlines, mask, tol=0.87,
+                                mode="either_end"),
+                        np.array([True, False, False]))
+
+    assert_array_equal(near_roi(streamlines, mask, tol=0.87, mode="both_end"),
+                        np.array([False, False, False]))
```
Garyfallidis: And here. Please check again your entire PR.

Garyfallidis:
:)

jchoude:
Just a +1 on not replacing `target` with the new functions.

PR #642, opened 2015-05-05
WIP - NF parallel framework

MrBago:
Starting a PR in case we want to discuss any of this stuff.

### DIFF
--- a/dipy/parallel/config.py
+++ b/dipy/parallel/config.py
@@ -1,14 +1,27 @@
 from multiprocessing import cpu_count
 from contextlib import contextmanager
+import multiprocessing
 
 _dipy_num_cpu = 1
+manager = None
 
 def activate_multithreading(num_cpu=None):
     global _dipy_num_cpu
+    global manager
     if num_cpu is None:
         _dipy_num_cpu = cpu_count()
-    elif num_cpu > 0:
+    elif num_cpu <= 0:
+        raise ValueError("num_cpu must be positive.")
+    else:
         _dipy_num_cpu = num_cpu
+
+    if manager is not None:
+        # manager.shut_down()
+        raise NotImplemented()
+    if _dipy_num_cpu > 1:
+        manager = PoolMananger(_dipy_num_cpu)
+    else:
+        manager = None
 
 def deactivate_multithreading():
     global _dipy_num_cpu
@@ -18,6 +31,31 @@ def deactivate_multithreading():
 def multithreading_on(num_cpu=None):
     previous_state = _dipy_num_cpu
     activate_multithreading(num_cpu)
-    yield
-    activate_multithreading(previous_state)
+    try:
+        yield
+    finally:
+        activate_multithreading(previous_state)
 
+class PoolMananger(object):
+
+    active = False
+
+    def __init__(self, n_cpu):
+        self.n_cpu = n_cpu
+        self.pool = multiprocessing.Pool(n_cpu)
+        self.active = True
+
+    def shut_down(self):
+        if self.active:
+            self.pool.close()
+            self.pool.join()
+            self.active = False
+            self.pool = None
+
+def cleanup():
+    if manager is not None:
+        manager.shut_down()
+    print("dipy-parallel-cleanup done")
+
+import atexit
+atexit.register(cleanup)
--- a/dipy/parallel/config.py
+++ b/dipy/parallel/config.py
@@ -16,10 +16,8 @@ def deactivate_multithreading():
 
 @contextmanager
 def multithreading_on(num_cpu=None):
-    print '{} in'.format(num_cpu)
     previous_state = _dipy_num_cpu
     activate_multithreading(num_cpu)
     yield
     activate_multithreading(previous_state)
-    print '{} out'.format(_dipy_num_cpu)
 
--- a/dipy/parallel/config.py
+++ b/dipy/parallel/config.py
@@ -1,0 +1,25 @@
+from multiprocessing import cpu_count
+from contextlib import contextmanager
+
+_dipy_num_cpu = 1
+
+def active_multithreading(num_cpu=None):
+    global _dipy_num_cpu
+    if num_cpu is None:
+        _dipy_num_cpu = cpu_count()
+    elif num_cpu > 0:
+        _dipy_num_cpu = num_cpu
+
+def deactive_multithreading():
+    global _dipy_num_cpu
+    _dipy_num_cpu = 1
+
+@contextmanager
+def multitheading_on(num_cpu=None):
+    print '{} in'.format(num_cpu)
+    previous_state = _dipy_num_cpu
+    active_multithreading(num_cpu)
+    yield
+    active_multithreading(previous_state)
+    print '{} out'.format(_dipy_num_cpu)
+
--- a/dipy/parallel/config.py
+++ b/dipy/parallel/config.py
@@ -3,23 +3,23 @@ from contextlib import contextmanager
 
 _dipy_num_cpu = 1
 
-def active_multithreading(num_cpu=None):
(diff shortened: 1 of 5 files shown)


arokem:
I noticed a few typos, so I will start with that

dipy/parallel/main.py
```diff
@@ -0,0 +1,141 @@
...
+
+import dipy.parallel.config
+
+
+def update_outputs(index, result, outputs):
+    for key, array in outputs.items():
+        array[index] = result[key]
+
+
+class MultiVoxelFuntion(object):
```
arokem: => "Function"

dipy/parallel/main.py
```diff
@@ -0,0 +1,141 @@
...
+            vox = data[ijk]
+            result = self._main(vox, *args, **kwargs)
+            update_outputs(ijk, result, outputs)
+        return outputs
+
+    def __call__(self, data, mask, *args, **kwargs):
+        self._serial(self, data, mask, *args, **kwargs)
+
+
+class UpdateCallback(MultiVoxelFuntion):
```
arokem: Dsmr hrtr

dipy/parallel/main.py
```diff
@@ -0,0 +1,141 @@
...
+
+def _array_split_points(mask, num_chunks):
+    # TODO split input based on mask values so each thread gets about the same
+    # number of voxels where mask is True.
+    chunk_size = mask.size / num_chunks
+    split_points = np.arange(1, num_chunks + 1) * chunk_size
+    return np.round(split_points).astype(int)
+
+
+def call_remotly(parallel_function, *args, **kwargs):
```
arokem: => "remotely"

dipy/parallel/main.py
```diff
@@ -0,0 +1,141 @@
...
+    return parallel_function._serial(*args, **kwargs)
+
+
+class ParallelFunction(MultiVoxelFuntion):
```
arokem: => Function

dipy/parallel/tests/test_main.py
```diff
@@ -0,0 +1,75 @@
...
+    npt.assert_array_almost_equal(a["max"], max)
+    npt.assert_array_almost_equal(a["argmax"], argmax)
+
+    for key in a.keys():
+        npt.assert_array_equal(a[key], b[key])
+
+    b = maxcumsum(data, mask, weights, return_cumsum=False)
+    assert("cumsum" not in b)
+
+class BrokenFucntion(ParallelFunction):
```
arokem: => BrokenFunction

dipy/parallel/tests/test_main.py
```diff
@@ -0,0 +1,75 @@
...
+
+    def _default_values(self, data, mask):
+        return {"out":np.zeros(data.shape[-1], float)}
+
+    def _main(self, data):
+        raise ValueError
+brokenfucntion = BrokenFucntion()
```
arokem: Same here

dipy/parallel/config.py
```diff
@@ -0,0 +1,25 @@
+from multiprocessing import cpu_count
+from contextlib import contextmanager
+
+_dipy_num_cpu = 1
+
+def active_multithreading(num_cpu=None):
```
MarcCote: It think it should be "activate" instead of "active".

dipy/parallel/config.py
```diff
@@ -0,0 +1,25 @@
...
+    global _dipy_num_cpu
+    if num_cpu is None:
+        _dipy_num_cpu = cpu_count()
+    elif num_cpu > 0:
+        _dipy_num_cpu = num_cpu
+
+def deactive_multithreading():
```
MarcCote: It think it should be "deactivate" instead of "deactive".

dipy/parallel/config.py
```diff
@@ -0,0 +1,25 @@
...
+    global _dipy_num_cpu
+    _dipy_num_cpu = 1
+
+@contextmanager
+def multitheading_on(num_cpu=None):
```
MarcCote: "multitheading" => "multithreading"

dipy/parallel/main.py
```diff
@@ -0,0 +1,141 @@
...
+
+def update_outputs(index, result, outputs):
+    for key, array in outputs.items():
+        array[index] = result[key]
+
+
+class MultiVoxelFuntion(object):
+    """A function that can be applied to every voxel of a dMRI data set.
+
+    Subclass MultiVoxelFuntion and define the _main and _default_Values methods
```
MarcCote: _default_Values => _default_values

MrBago:
I made `multi_voxel_fit` use the parallel framework, I'm still working on some of the other things we talked about but wanted to share this part for you all to look over.

@gabknight I realized that making a `peaks_from_fit` using the framework might be a little tricky because the framework expects arrays as inputs. I haven't actually looked at it closely so I'm not sure how much of an issue it is. I'm happy to talk over potential changes if it is an issue.

Garyfallidis:
Ping!  Any progress with this? Should I start having a look or wait a bit more?

Garyfallidis:
Also @MrBago I haven't heard back from you about the tracking API updates and paper. Neither about the odf_deconv with Cholesky. I would appreciate some updates. But hey I know you are busy with PhD thesis. Just give us a quick summary when you can.

sahmed95:
@arokem @Garyfallidis Hi, I pulled this branch and profiled the DSI fitting on my laptop (2 cores) and here are the results : 

On master, without multiprocessing : 10 loops, best of 3: 65.8 ms per loop
Using multiprocessing :                      10 loops, best of 3: 33.4 ms per loop

@MrBago has already implemented this in [multi_voxel_fit](0513c9d) such that any fitting which uses the multi_voxel decorator (such as DSI) will use the multiprocessing framework. I have a slightly better idea about this now and can discuss further.

So where do we go from here ? And what are the issues to address.

Some thoughts : 
1. Get more modules to use multi_voxel. (Currently the following modules use it - csedeconv, dsi, gqi, mapmri, shore). I don't know which other models are suited for voxel by voxel fitting.
2. Find an efficient way to split chunks of voxels based on mask. 
3. Implement a switch for turning multiprocessing 'on' or 'off'.

sahmed95:
Used the taiwan_ntu_dsi dataset
```
from dipy.data import fetch_taiwan_ntu_dsi, read_taiwan_ntu_dsi, get_sphere
from dipy.reconst.dsi import DiffusionSpectrumModel

img, gtab = read_taiwan_ntu_dsi()

data = img.get_data()
dsmodel = DiffusionSpectrumModel(gtab)
dataslice = data[:, :, data.shape[2] / 2]

%%timeit
dsfit = dsmodel.fit(dataslice)
```

Garyfallidis:
Hi @sahmed95 these are all good points. Do you have a new PR to look into? Do look into how chunks are used for DTI fitting. You may get some ideas there about how to use the mask. Are you thinking about using this technique for IVIM too? Or this will be paralleled like DTI?
See here https://github.com/nipy/dipy/blob/master/dipy/reconst/dti.py#L1195

Garyfallidis:
I suggest first make a PR with a current parallelization that you have. And then look into using the mask to improve performance even further. Start first with the things that can obviously use the parallel multivoxel fit and then try to see if also other models e.g. dti can work in the same way. There we use the chunk size but is that the best way? Should we use chunks here too? Let me know what you think about that.

Also we need to make sure that we can save the results after the parallel fit and create a peaks_from_fit function. But hey one step at the time.

sahmed95:
@Garyfallidis I opened a new PR https://github.com/nipy/dipy/pull/1135

Garyfallidis:
I believe this PR needs to close in favor of #1135.

skoudoro:
closing in favor of #1418

PR #648, opened 2015-05-13
Improve check of collinearity in vec2vec_rotmat

oesteban:
- [x] More strict check of collinearity of vectors.
  - [X] Fixed check for nans in the end.

### DIFF
--- a/dipy/core/geometry.py
+++ b/dipy/core/geometry.py
@@ -840,16 +840,17 @@ def vec2vec_rotmat(u, v):
     array([ 1.,  0.,  0.])
 
     """
-    norm_u_v = np.linalg.norm(u - v)
-    # return eye when u is the same with v
-    if np.linalg.norm(u - v) < np.finfo(float).eps:
-        return np.eye(3)
-    # This is the case of two antipodal vectors:
-    if norm_u_v == 2.0:
-        return -np.eye(3)
-
     w = np.cross(u, v)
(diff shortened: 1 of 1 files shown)


dipy/core/geometry.py
```diff
@@ -840,16 +840,17 @@ def vec2vec_rotmat(u, v):
     array([ 1.,  0.,  0.])
 
     """
-    norm_u_v = np.linalg.norm(u - v)
-    # return eye when u is the same with v
-    if np.linalg.norm(u - v) < np.finfo(float).eps:
+    w = np.cross(u, v)
```
stefanv: Documenting the logic behind the code would be helpful

Garyfallidis:
@oesteban thank you for your PR as @stefanv suggests it would be nice to explain a bit the logic of your change. I would like to see a comment in the code or here explaining why your change is a more strict check of collinearity. After that I will merge this PR. Sounds good?

oesteban:
I basically tried to replace workarounds by directly checking the cross product (which is at the core of finding the rotation matrix).

arokem:
I am +1 for this change. The previous implementation would only work for unit vectors, but despite the documentation stating that the inputs should be unit vectors, there is nothing enforcing that they are. I think that this is more robust for distinguishing the antipodal vs. colinear case.

arokem:
Since there were no further objections here, I am merging this.

PR #651, opened 2015-05-17
Added estimate_sigma bias correction + update example

samuelstjean:
So I added the correction factor for computing the std of magnitude images. Since they do not include any negative value, the variance is lower and the mean is shifted as compared ot a gaussian distribution. This factor corrects that fact and returns a closer estimate of the variance.
I also updated the example to show this fact and changed it to show a slice of diffusion instea,d since b0 images are always relatively clean.

Since estimate_sigma is the same as computing the variance over the background, the correction factor needs an estimate of the local magnitude SNR, but it is zero in that specific case. Hence, the value can be precomputed without using hyp1f1. Next step is to have it on a voxelwise basis with said hyp1f1 function in need of love over here : https://github.com/nipy/dipy/issues/650

### DIFF
--- a/dipy/denoise/noise_estimate.py
+++ b/dipy/denoise/noise_estimate.py
@@ -275,8 +275,8 @@ def estimate_sigma(arr, disable_background_masking=False, N=1):
     This function is the same as manually taking the standard deviation of the
     background and gives one value for the whole 3D array.
     It also includes the coil-dependent correction factor of Koay 2006
-    (see [1], equation 18) with theta = 0.
-    Since this function was introduced in [2] for T1 imaging,
+    (see [1]_, equation 18) with theta = 0.
+    Since this function was introduced in [2]_ for T1 imaging,
     it is expected to perform ok on diffusion MRI data, but might oversmooth
     some regions and leave others un-denoised for spatially varying noise profiles.
     Consider using :func:`piesno` to estimate sigma instead if visual inacuracies
@@ -284,11 +284,11 @@ def estimate_sigma(arr, disable_background_masking=False, N=1):
 
     Reference
     -------
-    [1] Koay, C. G., & Basser, P. J. (2006). Analytically exact correction
+    .. [1] Koay, C. G., & Basser, P. J. (2006). Analytically exact correction
     scheme for signal extraction from noisy magnitude MR signals.
     Journal of Magnetic Resonance), 179(2), 317-22.
 
-    [2] Coupe, P., Yger, P., Prima, S., Hellier, P., Kervrann, C., Barillot, C., 2008.
+    .. [2] Coupe, P., Yger, P., Prima, S., Hellier, P., Kervrann, C., Barillot, C., 2008.
     An optimized blockwise nonlocal means denoising filter for 3-D magnetic
     resonance images, IEEE Trans. Med. Imaging 27, 425-41.
     """
--- a/dipy/denoise/noise_estimate.py
+++ b/dipy/denoise/noise_estimate.py
@@ -270,23 +270,28 @@ def estimate_sigma(arr, disable_background_masking=False, N=1):
     sigma : ndarray
         standard deviation of the noise, one estimation per volume.
 
+    Note
+    -------
+    This function is the same as manually taking the standard deviation of the
+    background and gives one value for the whole 3D array.
+    It also includes the coil-dependent correction factor of Koay 2006 (see [1], equation 18)
+    with theta = 0
+    Since this function was introduced in [2] T1 imaging, it is expected to
+    perform ok on diffusion MRI data,  but might oversmooth some regions and
+    leave others un-denoised for spatially varying noise profiles.
+    Consider using :func:`piesno` to estimate sigma instead if visual inacuracies
+    are apparent in the denoised result.
+
     Reference
     -------
     [1] Koay, C. G., & Basser, P. J. (2006). Analytically exact correction
     scheme for signal extraction from noisy magnitude MR signals.
-    Journal of Magnetic Resonance), 179(2), 317–22.
+    Journal of Magnetic Resonance), 179(2), 317-22.
 
-    Note
-    -------
-    This function is the same as manually taking the standard deviation of the
-    background and gives one value for the whole 3D array. Since it was
-    developped for T1 imaging. It is expected to perform ok on diffusion MRI data,
-    but might oversmooth some regions and leave others un-denoised if the noise
-    varies spatially (especially near the middle of the image).
-    Consider using piesno instead if you see visual inacuracies
-    with nlmeans and estimate_sigma.
+    [2] Coupe, P., Yger, P., Prima, S., Hellier, P., Kervrann, C., Barillot, C., 2008.
+    An optimized blockwise nonlocal means denoising filter for 3-D magnetic
+    resonance images, IEEE Trans. Med. Imaging 27, 425-41.
     """
-
     k = np.zeros((3, 3, 3), dtype=np.int8)
 
     k[0, 1, 1] = 1
@@ -311,7 +316,8 @@ def estimate_sigma(arr, disable_background_masking=False, N=1):
     if N in correction_factor:
         factor = correction_factor[N]
     else:
-        factor = correction_factor[1]
+        raise ValueError("N = {0} is not supported! Please choose amongst \
+{1}".format(N, sorted(list(correction_factor.keys()))))
 
     if arr.ndim == 3:
         sigma = np.zeros(1, dtype=np.float32)
--- a/dipy/denoise/noise_estimate.py
+++ b/dipy/denoise/noise_estimate.py
@@ -92,12 +92,10 @@ def piesno(data, N, alpha=0.01, l=100, itermax=100, eps=1e-5, return_mask=False)
         mask_noise = np.zeros(data.shape[:-1], dtype=np.bool)
 
         for idx in range(data.shape[-2]):
-            sigma[idx], mask_noise[..., idx] = _piesno_3D(data[..., idx, :], N,
-                                                          alpha=alpha,
-                                                          l=l,
-                                                          itermax=itermax,
-                                                          eps=eps,
-                                                          return_mask=True)
+            sigma[idx], mask_noise[..., idx] = _piesno_3D(data[..., idx, :],
+                                                          N, alpha=alpha,
+                                                          l=l, itermax=itermax,
+                                                          eps=eps, return_mask=True)
 
     else:
         sigma, mask_noise = _piesno_3D(data, N,
--- a/dipy/denoise/noise_estimate.py
+++ b/dipy/denoise/noise_estimate.py
@@ -248,7 +248,7 @@ def _piesno_3D(data, N, alpha=0.01, l=100, itermax=100, eps=1e-5,
     return sigma[pos]
 
 
-def estimate_sigma(arr, disable_background_masking=False):
+def estimate_sigma(arr, disable_background_masking=False, N=1):
     """Standard deviation estimation from local patches
 
     Parameters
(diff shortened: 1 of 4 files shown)


samuelstjean:
Since I expect the "why did you change the tests" question, an example as to why this way is better than without the correction factor:

```
% Generate white noise N(25,50**2)
In [34]: a=(np.random.randn(10000)*50)+25

In [35]: np.mean(a)
Out[35]: 24.437325656547163

In [36]: np.std(a)
Out[36]: 50.696824939621159

In [37]: b=(np.random.randn(10000)*50)+25

In [38]: np.std(a)
Out[38]: 50.696824939621159

In [39]: np.mean(b)
Out[39]: 25.756600844271365

% Turn that to Rician noise
In [40]: c=np.sqrt(a**2+b**2)

% Variance is lower, since we lost the negative numbers
In [41]: np.std(c)
Out[41]: 36.449713000167094

% The man is also shifted, resulting in the signal dependent bias
In [42]: np.mean(c)
Out[42]: 70.68793843681992

%Magic correction factor, which I precomputed with this
In [43]: from scilpy.denoising.stabilizer import _test_xi

%Since we assume zero mean background in this PR, 
% the factor is not signal dependent, but only N coil(s) dependent 
In [44]: _test_xi(25,50,1)
Out[44]: 0.47990987386190764

%Apply the factor, which return a much more closer estimate, 51 instead of 35, while the real value hovers about 50
In [45]: 35.86981271463975/sqrt(0.47990987386190764)
Out[45]: 51.778476330177732
```

arokem:
Cool. I am going to read the Koay paper and give this a more proper review. For now, just a first question: does it make sense to leave the 'old' functionality as an option here?

dipy/denoise/noise_estimate.py
```diff
@@ -274,6 +296,23 @@ def estimate_sigma(arr, disable_background_masking=False):
...
+                        16: 0.4960339908122364,
+                        20: 0.4968365823718557,
+                        24: 0.49736907650825657,
+                        32: 0.49803177052530145,
+                        64: 0.49901964176235936}
+
+    if N in correction_factor:
+        factor = correction_factor[N]
+    else:
+        factor = correction_factor[1]
```
arokem: If the user provides a non-sense value of `N`, I think that it would make more sense to raise an error, rather than assuming that they meant `1`.

samuelstjean:
Only one would want to compare with an old result, but I take for granted that people don't upgrade libs for fun when doing a huge project. I don't think the old behavior is still needed, since it's only an added division, kind of like when we fixed the nlmeans that was missing a square root.

The real way to do it would be to use the function instead of a precomputed table (that way it will work for all values of N), but that need a working hyp1f1 function, so it's gonna be that for the time being.

dipy/denoise/noise_estimate.py
samuelstjean: I was thinking of raising a warning, but if you don't force any value all the old code will break in your face.

arokem: I think that you are forcing a value, by setting the default in the
function signature.

I just meant that if a user called:

    estimate_sigma(arr, N=3)

The function should raise an error, not silently behave as though the user
entered `N=1`.  Or am I missing something?

samuelstjean: No I am indeed forcing a value. I just precomputed a bunch of mostly plausible values. What I'm not sure of though is what to do if the user does not input any value for N. All the old code won't be working anymore, but I could raise an error if the number is not in the list, 1 for no input, and the value as requested if it's in the list.

arokem: Exactly: raise an error if a number is provided, but it is not a key to the
`correction_factor` dict. Otherwise, default to 1.

I propose to rewrite this logic as:

    if N in correction_factor:
        factor = correction_factor[N]  # Per default, N = 1
    else:
       raise ValueError("Correction factor needs to be one of the
following: %s, but you entered %s" %(list(correction_factor.keys()),  N))

With the necessary line breaks, etc.

arokem: Just to be clear: I believe that the 'default to 1' part is already taken
care of.

arokem:
Concerning the main part of this: do I understand correctly that you are simply applying equation 7 of the Koay and Basser paper, for the case where S=0? Might be worth adding to the documentation of the function. Also, how did you compute these actual numbers? The paper only implicitly includes them as the intercepts of the curves in Figure 1.

samuelstjean:
Yep, that's equation 7, but the factor xi is given by equation 8 for the Rician case and equation 18 for the general cas. I indeed used eta=0 (and thus theta=eta/sigma=0, with theta the magnitude SNR, eta the signal value and sigma the noise std). Line 299 depicts this fact with a comment, and the reference is in the function docstring.

dipy/denoise/noise_estimate.py
```diff
@@ -260,11 +260,33 @@ def estimate_sigma(arr, disable_background_masking=False):
...
+    -------
+    [1] Koay, C. G., & Basser, P. J. (2006). Analytically exact correction
+    scheme for signal extraction from noisy magnitude MR signals.
+    Journal of Magnetic Resonance), 179(2), 317–22.
+
+    Note
+    -------
+    This function is the same as manually taking the standard deviation of the
+    background and gives one value for the whole 3D array. Since it was
+    developped for T1 imaging. It is expected to perform ok on diffusion MRI data,
```
arokem: I think this was meant to be one sentence, and there's a small typo in here ("developped"). It's also a bit vague, so might not help users that much. How about the following: 

```
This function implements equation 7 in [Koay2006]_: it applies a coil-number dependent correction factor to the standard deviation of the background and  produces a single value for the input volume, or 4D volume series. Since this formulation was developed for T1 imaging, it could work well for diffusion MRI, but take note: when used for denoising, this method may oversmooth some regions and leave others un-denoised. This is particularly true for cases in which the noise profile is not uniform across the volume. Consider using :func:`piesno` if visual examination  reveals non-uniform denoising performance.
```
As an aside, the Notes section should precede the "References" section.

dipy/denoise/noise_estimate.py
```diff
@@ -292,6 +331,6 @@ def estimate_sigma(arr, disable_background_masking=False):
     for i in range(sigma.size):
         convolve(arr[..., i], k, output=conv_out)
         mean_block = np.sqrt(6/7) * (arr[..., i] - 1/6 * conv_out)
-        sigma[i] = np.sqrt(np.mean(mean_block[mask]**2))
+        sigma[i] = np.sqrt(np.mean(mean_block[mask]**2) / factor)
```
arokem: PEP8: white-space around operators (I know - it was there before, but let's please clean it up, while we're at it).

dipy/denoise/tests/test_noise_estimate.py
```diff
@@ -46,20 +46,24 @@ def test_estimate_sigma():
...
+    sigma = estimate_sigma(arr, disable_background_masking=False, N=1)
+    assert_array_almost_equal(sigma, np.array([0.10286889997472792 / np.sqrt(0.42920367320510366),
+                                               0.10286889997472792 / np.sqrt(0.42920367320510366),
+                                               0.10286889997472792 / np.sqrt(0.42920367320510366)]))
 
     arr = np.zeros((3, 3, 3))
     arr[0, 0, 0] = 1
-    sigma = estimate_sigma(arr, disable_background_masking=True)
-    assert_array_almost_equal(sigma, 0.46291005)
+    sigma = estimate_sigma(arr, disable_background_masking=True, N=4)
```
arokem: Nice. Good to test for numbers other than 1...

arokem:
Any ideas how this would affect the use of `estimate_sigma` in the RESTORE example?

dipy/denoise/noise_estimate.py
samuelstjean: Wasn't there a note saying you can keep the priority of operation grouped together like in

a**2 + b**2

Plus I really find it counterintuitive to have the exponent not close to the actual number, like in

a ** 2 / b ** 2
Seems messy :/

samuelstjean:
It's almost the same as using the multiplication by sqrt(pi/2) I'd say, but this one is more general since it takes into account the increased noise floor with higher N. Good thing you brought that up, I think it's better to untangle the noise estimation method from the models, else two correction factors would get applied.

dipy/denoise/noise_estimate.py
arokem: Yes: operators with higher precedence should be without white-space. But I
don't think this applies in cases where they are parenthesis-delimited,
such as here.

All the BDFL gave us mere mortals is the following guidance:

Yes:

i = i + 1
submitted += 1
x = x*2 - 1
hypot2 = x*x + y*y
c = (a+b) * (a-b)

No:

i=i+1
submitted +=1
x = x * 2 - 1
hypot2 = x * x + y * y

c = (a + b) * (a - b)

I don't know what to say, except that an automatic tool (like pep8) would
definitely flag this one. I agree with you that it doesn't read as nicely,
but that's the good thing about using standards: no one's individual taste
matters.

samuelstjean:
PEP 8:

But most importantly: know when to be inconsistent -- sometimes the style guide just doesn't apply. When in doubt, use your best judgment.

dipy/denoise/noise_estimate.py
samuelstjean: PEP 8:

But most importantly: know when to be inconsistent -- sometimes the style guide just doesn't apply. When in doubt, use your best judgment.

arokem: And if it wasn't bad enough, he threw that one in for good measure. I
wonder how many programmer-hours have gone to debating that one line over
the years

arokem: But enough fun and games. Let me know when you've addressed the comments
and I'll take another look

samuelstjean:
Fixed in majority

arokem:
OK - I think that we can merge this. Let's plan on some time next week, so that others can take a look as well. Anyone else have any comments on this?

arokem:
Sorry - I just remembered one more question that I had: what effect does this have on the RESTORE example? Did you run that one?

samuelstjean:
nope, but I can absolutely garantee that we should remove the sqrt(pi/2) fix, since that overlaps with this correction factor. Hence the now deeply buried issue I posted with my suggested fix (this is number 1 of them if I recall correctly).

arokem:
OK - since this will have an effect there, please do run it and make sure that it runs, make any corrections to that example that you think are necessary, as part of this PR. Which one is the 'deeply buried' issue you refer to (link)?

samuelstjean:
https://github.com/nipy/dipy/issues/636

arokem:
Oh - that issue - all good suggestions, as discussed. So you think the
first on that list should be a part of this PR? Seems like a good idea to
me. Either way, we can't merge this if it breaks the RESTORE example.

samuelstjean:
So, do you know the specs of the scanner used for the standford dataset?

arokem:
Yes:

http://cni.stanford.edu/wiki/MR_Scanner

We used the 32 channel coil for these measurements.

samuelstjean:
Well, apparently GE can use both ARC (GRAPPA) or SENSE from a quick internet search. Any idea which one is used by any chance?

arokem:
99% sure it was SENSE

samuelstjean:
I took a loo at the code, and apparently everything is finer, I thought the noise estimation was embedded in the method, but you use another estimator, which is fine and completely independent from this implementation. I orignally thoguh two different correction facotr would be applied, but that's not the case.

arokem:
Thanks for all the cleanup in dti.py!

This function is being used in the RESTORE example: 

https://github.com/nipy/dipy/blob/master/doc/examples/restore_dti.py#L162

Could you please check what happens when you run this example?

arokem:
Just checking back in here: did you have a chance to run the RESTORE example on your branch?

samuelstjean:
Not yet, I just learned I had to defend my master thesis by next week (which I finished wednesday), yesterday. So that will probably take my time until then.

arokem:
Oh wow! This can definitely wait... Hopefully more people will take a look
in the meanwhile.

Good luck!

samuelstjean:
Just ran it, doesn't seem to change anything visually (as expected). Fuunnily enough, the fmri guys on our side hardcoded a bunch of values since they were unhappy with this function result, and with the proper correction it gives the same value that they hardcoded, for whatever that is worth.
Well, maybe anyone else could comment on the usefulness of changing this as default, but as far as I am concerned this is the right thing to do. And it's also the first open source implementation that I am aware of ;)

dti_fa_distributions
tensor_ellipsoids_restore_noisy
tensor_ellipsoids_wls
tensor_ellipsoids_wls_noisy

arokem:
Well - if this will make your fMRI people happy, then I am happy too! :-)

Thanks for checking that. This is complete from my point-of-view, but I
would like to give others a chance to take a look. If no-one pipes up, I
will merge this next week.

samuelstjean:
This broke for an out of space error on travis, no need to retrigger it.

s1
s2

arokem:
Interesting. Haven't seen that one before...

Issue #611, opened 2015-03-20
OSX test fail 'we check the default value of lambda ...'


matthew-brett:
Anyone recognize this one?

```
======================================================================
FAIL: We check that the default value of lambda is the expected value with
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/travis/build/MacPython/dipy-wheels/venv/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/Users/travis/build/MacPython/dipy-wheels/venv/lib/python2.7/site-packages/dipy/reconst/tests/test_csdeconv.py", line 454, in test_default_lambda_csdmodel
    npt.assert_array_almost_equal(model_full.B_reg, expected * B_reg)
  File "/Users/travis/build/MacPython/dipy-wheels/venv/lib/python2.7/site-packages/numpy/testing/utils.py", line 800, in assert_array_almost_equal
    header=('Arrays are not almost equal to %d decimals' % decimal))
  File "/Users/travis/build/MacPython/dipy-wheels/venv/lib/python2.7/site-packages/numpy/testing/utils.py", line 636, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Arrays are not almost equal to 6 decimals
(mismatch 23.9533456108%)
 x: array([[  2.32929569e+01,   0.00000000e+00,  -0.00000000e+00, ...,
          0.00000000e+00,   0.00000000e+00,   0.00000000e+00],
       [  2.32929569e+01,  -1.22849454e+00,  -5.02810375e+00, ...,...
 y: array([[  2.32929581e+01,   0.00000000e+00,  -0.00000000e+00, ...,
          0.00000000e+00,   0.00000000e+00,   0.00000000e+00],
       [  2.32929581e+01,  -1.22849461e+00,  -5.02810401e+00, ...,...
```

https://travis-ci.org/MacPython/dipy-wheels/jobs/55117917#L1270

arokem:
Interesting. I see this on python 3.4, but not 2.7, on my (mac) laptop.
Seems to go away with increased precision (#612), but maybe this is the
canary in the coal-mine?

This is not our nipy/dipy travis configuration, right? How/where are you
running this?

arokem:
This one was resolved in https://github.com/nipy/dipy/pull/612

PR #652, opened 2015-05-19
MAINT: work around scipy bug in sph_harm

argriffing:
related bugs/issues/PRs:
https://github.com/scipy/scipy/issues/4887
https://github.com/scipy/scipy/issues/4895
https://github.com/nipy/dipy/issues/611
https://github.com/nipy/dipy/pull/612

### DIFF
--- a/dipy/reconst/shm.py
+++ b/dipy/reconst/shm.py
@@ -118,7 +118,7 @@ def sh_to_rh(r_sh, m, n):
     mask = m == 0
     # The delta function at theta = phi = 0 is known to have zero coefficients
     # where m != 0, therefore we need only compute the coefficients at m=0.
-    dirac_sh = gen_dirac(0, n[mask], 0, 0)
+    dirac_sh = gen_dirac(0, n[mask], 0, 0.)
     r_rh = r_sh[mask] / dirac_sh
     return r_rh
 


argriffing:
I don't think the dipy travis config uses a new enough scipy to test this change.

arokem:
I think you are correct. Let me try this out on my laptop.

arokem:
I can confirm that this works on scipy 0.15 on my laptop, killing #611 and obviating #612.

arokem:
Another option is to wait until this is fixed upstream in scipy, and keep using our implementation until that time (which is what #612 now implements). Any thoughts from anyone else?

arokem:
Also: thanks so much for taking a look!

pv:
This is probably a numpy bug in ufunc dtype selection (which is choosing float32 instead of float64), so the upstream scipy fix will also be a workaround (and you will have to wait for it for a moderately long time as your users would need to use a new enough Scipy version). The problem is fairly restricted --- it occurs only in the special case when all arguments are integers, at least one of the first two are arrays, *and* both of the angle arguments are integer scalars.

arokem:
Thanks for clarifying. I still think that we can wait. We do have a fully
functioning solution in our code. This particular PR seems to solve the
problem for only a small subset of the potential failure modes of this
function (only when used through `sh_to_rh`). I prefer the solution we have
on #612, and waiting.

Anyone else but me have any opinions on this?

argriffing:
> This particular PR seems to solve the problem for only a small subset of the potential failure modes of this function (only when used through `sh_to_rh`)

Sorry this PR was a bit tongue-in-cheek.  If you never want low-precision output from `sph_harm` then I think you could change
`return sph_harm(m, n, theta, phi)`
to
`return sph_harm(m, n, theta, phi, dtype=complex)`
at
https://github.com/nipy/dipy/blob/maint/0.9.x/dipy/reconst/shm.py#L189

arokem:
Oh - you shouldn't mess with me that way. I'm way too slow to get that kind
of humor. And awfully gullible, when it comes to PRs.

Thanks for the new suggestion. This one was serious, right?

Now implemented in #612.

argriffing:
@pv Do you have a suggestion for the best practice for working around this bug in packages downstream from numpy/scipy?  I think the method in this PR would work for `dipy`, and that the less ad-hoc workaround of adding `dtype=complex` in the `dipy` wrapper of `sph_harm` would also work (assuming that they never want `complex64` output under any circumstance).  Do you agree that the second of these two choices would be more appropriate?  Is there an even better workaround?

Garyfallidis:
Ping! How is this PR going on? Still a work in progress?

arokem:
Hi! This one is probably superseded by #612. Closing this one for now.

PR #657, opened 2015-05-26
Fix compilation error caused by inline functions 

omarocegueda:
If inline functions are just declared in the pxd, cython attempts to declare inline pointers to them, which is invalid since variables cannot be declared inline. According to this:
http://docs.cython.org/src/tutorial/pxd_files.html
inline functions should be defined, not just declared in the pxd.
Unfortunately, right now I don't have access to the buildbots to try this, but in theory this should fix the issue.

### DIFF
--- a/dipy/align/vector_fields.pxd
+++ b/dipy/align/vector_fields.pxd
@@ -1,10 +1,52 @@
+#!python
+#cython: boundscheck=False
+#cython: wraparound=False
+#cython: cdivision=True
+
 cdef inline double _apply_affine_3d_x0(double x0, double x1, double x2,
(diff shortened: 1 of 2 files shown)


arokem:
Yup. I can confirm that with this branch in place everything compiles just fine on my (Mac) laptop, and tests all pass (except for the one that should be taken care of once #612 lands).

arokem:
I am going to go ahead and merge this into `master`, since it's small and presumably fixes a current breakage. Let's see what the buildbots do with this.

omarocegueda:
Thanks Ariel! =D

PR #658, opened 2015-05-26
FIX: resaved npy file causing load error for py 33

matthew-brett:
Loading ``dipy/data/t1_coronal_slice.npy`` on Python 3.3 with numpy 1.7.1 was
causing a syntax error, similar to this one:

http://stackoverflow.com/questions/27145273/compatibility-numpy-1-7-vs-numpy-1-8

Load, resave in numpy 1.9.2 in hope that that will fix the errors.


arokem:
Did you run a try_branch with this one? Does it resolve the issue?

PR #661, opened 2015-06-06
Test DTI eigenvectors

arokem:
As pointed out by @RafaelNH here: https://github.com/nipy/dipy/pull/582, there seems to be some inconsistency about whether eigenvectors of tensors should be arranged column-wise or row-wise in 3-by-3 arrays. I am trying to enforce something uniform here. At the moment, this triggers several other test failures, but I wonder whether these are because of patching around this inconsistency, and would just require some additional fixing to make all parts consistent. I am continuing to investigate.

### DIFF
--- a/dipy/reconst/tests/test_dti.py
+++ b/dipy/reconst/tests/test_dti.py
@@ -112,8 +112,8 @@ def test_tensor_model():
     # Check that eigenvalues and eigenvectors are properly sorted through
     # that previous operation:
     for i in range(3):
-        assert_array_equal(np.dot(tensor, evecs[:, i]),
-                           evals[i] * evecs[:, i])
+        assert_array_almost_equal(np.dot(tensor, evecs[:, i]),
+                                  evals[i] * evecs[:, i])
     # Design Matrix
     X = dti.design_matrix(gtab)
     # Signals
--- a/dipy/reconst/tests/test_dti.py
+++ b/dipy/reconst/tests/test_dti.py
@@ -106,7 +106,14 @@ def test_tensor_model():
     tensor = from_lower_triangular(D)
     A_squiggle = tensor - (1 / 3.0) * np.trace(tensor) * np.eye(3)
     mode = 3 * np.sqrt(6) * np.linalg.det(A_squiggle / np.linalg.norm(A_squiggle))
-    evecs = np.linalg.eigh(tensor)[1]
+    evals_eigh, evecs_eigh = np.linalg.eigh(tensor.T)
+    # Sort according to eigen-value from large to small:
+    evecs = evecs_eigh[:, np.argsort(evals_eigh)[::-1]]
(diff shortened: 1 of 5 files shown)


arokem:
OK - I think that I have patched up all the subsequent test failures, but I'll wait to see if Travis comes up with something I've missed. 

If there's one thing I've learned from this, it is how poorly our simulation module is covered. I had to rely on tests from all kinds of other modules (e.g. SHORE) to tell me that the "ground truth" simulation was doing it wrong.

arokem:
Also just confirmed that this will work fine when we merge https://github.com/nipy/dipy/pull/648

arokem:
This PR was made a bit obsolete with the merging of #582, but there is still some useful stuff here. In particular, more testing of DTI. Changing the title.

arokem:
This one's now rebased. This one would be a good opportunity to gain experience reviewing code for someone who's interested in pitching in, but doesn't have a lot of time to go over large PRs.

arokem:
Just rebased again. With 26 lines of added testing, this one should be uncontroversial and a real quick review. Would someone please take a look?

dipy/reconst/tests/test_dti.py
```diff
@@ -106,7 +106,14 @@ def test_tensor_model():
...
     A_squiggle = tensor - (1 / 3.0) * np.trace(tensor) * np.eye(3)
     mode = 3 * np.sqrt(6) * np.linalg.det(A_squiggle / np.linalg.norm(A_squiggle))
-    evecs = np.linalg.eigh(tensor)[1]
+    evals_eigh, evecs_eigh = np.linalg.eigh(tensor)
+    # Sort according to eigen-value from large to small:
+    evecs = evecs_eigh[:, np.argsort(evals_eigh)[::-1]]
+    # Check that eigenvalues and eigenvectors are properly sorted through
+    # that previous operation:
+    for i in range(3):
+        assert_array_equal(np.dot(tensor, evecs[:, i]),
```
omarocegueda: In general, we should avoid comparing floating point values for equality, because even if the operations are equivalent, the result will in general depend on the order of evaluation, for example:
```
x = (np.float64(3.1) / np.float64(2.1)) * (np.float64(2.1) / np.float64(3.1))
y = (np.float64(3.1) * np.float64(2.1)) / (np.float64(3.1) * np.float64(2.1))
x == y
```
Which evaluates to False.
I think `assert_array_almost_equal` with a precision of about `1e-9` would be better.

arokem:
Thanks for the comment! Changed accordingly.

omarocegueda:
Thanks Ariel!, it looks good to me =)

arokem:
Any more comments here?

arokem:
Just rebased this on top of master. Would anyone please give this a review and merge this?

arokem:
Um, hullo? Is this thing on?

Garyfallidis:
Thx @arokem. :+1:

PR #663, opened 2015-06-09
changed samuel st-jean email to the usherbrooke one

samuelstjean:
More professional than a gmail address.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -23,3 +23,6 @@ Omar Ocegueda <jomaroceguedag@gmail.com> omarocegueda <jomaroceguedag@gmail.com>
 Sylvain Merlet <merlet.sylvain@gmail.com> smerlet <merlet.sylvain@gmail.com>
 Mauro Zucchelli <mauro.zucchelli88@gmail.com> Mauro <mauro@Mauro.(none)>
 Andrew Lawrence <lawrenceajl@gmail.com> AndrewLawrence <AndrewLawrence@users.noreply.github.com>
+Samuel St-Jean <samuel.st-jean@usherbrooke.ca> samuelstjean <samuel.st-jean@usherbrooke.ca>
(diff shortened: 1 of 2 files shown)


jchoude:
Which email address do you use for your commits? I would guess your gmail one... It's probably easier to keep the same, for easiness of mapping later on, but I'll let @Garyfallidis or others decide.

samuelstjean:
Apparently I use this one also for commiting stuff

stjs2902 ~ $ [~] git config --global user.email
<snip>@usherbrooke.ca -> you get the idea

arokem:
Hey @samuelstjean - sorry for the delay (HBM, life...). Seems fine to me. There are a few people who commit with a few different email addresses. This is documented in the `.mailmap` file as well, so please change it there as well.

samuelstjean:
Does .mailmap serves a purpose or it is not updated frequently? There seems to be few email addresses  in there, and mine is not there, should I add it?

matthew-brett:
Checkout https://www.google.com/?gws_rd=ssl#q=mailmap 

`.mailmap` serves to identify two or more email addresses and commit names with the same person.  You'll only see an address there if someone commits with more than one name / email pair.

samuelstjean:
Oh ok, I did not know that. Anyway, I don't fall in thta usecase as my commit email is the one I am changing for in this current PR, the gmail one has never been used (yet, it'll probably come into usage when my academic one expire if I don't find another one by then).

arokem:
According to 'git -se', you've used the following emails to make commits:

Samuel St-Jean <Samuel.St-Jean@usherbrooke.ca>

Samuel St-Jean <stjeansam@gmail.com>
Samuel St-Jean <qekchose@gmail.com>

It would be good to have all of them in that .mailmap file.

samuelstjean:
Isn't that a garanteed recipe for tons of spam tough to have clear text emails everywhere online?

arokem:
Spam is a given.

PR #233, opened 2013-08-22
WIP: Dki

arokem:
This is still WIP, but I think that I am about ready for others to also start taking a look at this. 

I am relying mostly on these two papers for the implementation details: 

    [1] Lu, H, Jensen, JH, Ramani, A, Helpern, JA (2006). Three-dimensional
    characterization of non-gaussian water diffusion in  humans using diffusion
    kurtosis imaging. NMR in Biomedicine 19: 236-247

    [2] Jensen, JH and Helpern JA (2010). MRI quantification of non-Gaussian
    water diffusion by kurtosis analysis. NMR in Biomedicine 23: 698-710.

However, I am trying to generalize this to cases in which measurements were conducted in different directions in the different b values. Not sure if that's legit.

PR #664, opened 2015-06-11
DKI fitting (DKI PR2) 

RafaelNH:
I am starting this work from the code that Prof. Maurizio Marrale kindly supplied.

He did a nice work on DKI fitting and estimation of standard DKI rotational invariant measures. Particularly, mean and radial kurtosis are implemented according to the analytical solution proposed by Tabesh et al., 2011 (Magn Reson Med. 65(3):823-36. doi: 10.1002/mrm.22655). 

Nevertheless, there is still plenty of work to be done on this pull request. I already remove some redundancy of code relative to DTI's module. Now I will add some documentation on the poorly documented functions, rewrite functions in PEP8 style, create some missing parts of the script for better consistency with the module for DTI reconstruction (for example signal predicted from DKI is missing). 

As suggested by @Garyfallidis on PR #233, I am also creating an example document for DKI usage (doc/example/reconst_dki). 

Finally, using testing modules, I will check if implementations are correct by comparing the output of the DKI reconstructions with the DKI simulations and comparing the MK/RK estimations to more conservative implementations of DKI invariant measures (Eqs, 11, 12, 13 and 14 of Jensen and Helpern, 2010 - NMR Biomed. 23(7):698-710. doi: 10.1002/nbm.1518)


dipy/reconst/dki.py
```diff
@@ -0,0 +1,1332 @@
...
+#    -------
+#    parameters : array-like
+#        Various parameters for the evaluation of the kurtosis tensor.
+#        
+#        References
+#        ----------
+#           [1] Tabesh, A., Jensen, J.H., Ardekani, B.A., Helpern, J.A., 2011.
+#           Estimation of tensors and tensor-derived measures in diffusional
+#           kurtosis imaging. Magn Reson Med. 65(3), 823-836
+#
```
RafaelNH: After proper documentation on each function, this will be removed

arokem:
Cool. Looks like you have your work cut out for you here. Don't forget to eliminate the duplication of the DKI design matrix function (which also appears here: https://github.com/nipy/dipy/blob/master/dipy/sims/voxel.py#L536). I think that it makes more sense to put the design matrix definition in this module, and then to import it from here to the simulations.

dipy/reconst/dki.py
```diff
@@ -0,0 +1,1326 @@
...
+""" Classes and functions for fitting tensors """
+from __future__ import division, print_function, absolute_import
+
+import warnings
+
+import numpy as np
+
+import scipy.optimize as opt
+
+from dipy.reconst.dti import (fractional_anisotropy, geoderic_anisotropy,
```
omarocegueda: Typo: `geoderic_anisotropy`-->`geodesic_anisotropy`, this is causing the Travis build failure

dipy/reconst/dki.py
```diff
@@ -0,0 +1,1606 @@
...
+
+    # These for loops can be avoid using kt symmetry properties. If this
+    # simplification is done we don't need also to reconstruct the full kt
+    # tensor
+    for il in range(3):
+        for jl in range(3):
+            for kl in range(3):
+                for ll in range(3):
+                    multiplyB = B[il][indi]*B[jl][indj]*B[kl][indk]*B[ll][indl]
+                    Wre = Wre + multiplyB * W4D[il][jl][kl][ll]
```
MrBago: numpy indexing is done by, `W4D[il, jl, kl, ll]`. The other way works sometimes, but can cause bugs.

dipy/reconst/dki.py
```diff
@@ -0,0 +1,1606 @@
...
+        dki_params[vox] = _wls_iter(design_matrix, inv_design, data_flat[vox],
+                                    min_signal, min_diffusivity)
+
+    # Reshape data according to the input data shape
+    dki_params = dki_params.reshape((data.shape[:-1]) + (27,))
+
+    return dki_params
+
+
+def _wls_iter(design_matrix, inv_design, sig, min_signal, min_diffusivity):
```
MrBago: This function seems to be very close to the similar function it dipy.reconst.dti, could this function call the other function to avoid code duplication?

dipy/reconst/dki.py
```diff
@@ -0,0 +1,1606 @@
...
+    MD = mean_diffusivity(evals)
+
+    # Initialize AKC matrix
+    V = sphere.vertices
+    AKC = np.zeros((len(kt), len(V)))
+
+    # loop over all voxels
+    for vox in range(len(kt)):
+        R = evecs[vox]
+        dt = lower_triangular(np.dot(np.dot(R, np.diag(evals[vox])), R.T))
```
MrBago: I believe there is a function in `reconst.dti` that computes the lower triangular elements from the evals/evecs that might be more efficient than this.

dipy/reconst/dki.py
```diff
@@ -0,0 +1,1606 @@
...
+        if mask is None:
+            out_shape = data.shape[:-1] + (-1, )
+            dki_params = params_in_mask.reshape(out_shape)
+        else:
+            dki_params = np.zeros(data.shape[:-1] + (27,))
+            dki_params[mask, :] = params_in_mask
+
+        return DKIFit(self, dki_params)
+
+    def predict(self, dki_params, S0=1):
```
MrBago: The predict method should take an optional gtab argument so that predictions can be made on arbitrary gradient sets.

MrBago:
I like this PR but it's pretty long, I wasn't able to review it in detail. One concern I have is that the math looks very dense. Has anyone looked to see if the long expressions can be vectorized or better implemented using linear algebra operations? When I see `V[:, 0] * V[:, 1] * dt[1] + ...` I think we should be either using broadcasting or linear algebra but I'm not sure in this case.

dipy/reconst/dki.py
```diff
@@ -0,0 +1,1634 @@
...
+from dipy.data import get_sphere
+from ..core.gradients import gradient_table
+from ..core.geometry import vector_norm
+from ..core.sphere import Sphere
+from .vec_val_sum import vec_val_vect
+from ..core.onetime import auto_attr
+from .base import ReconstModel
+
+
+def rdpython(x,y,z):
```
Garyfallidis: rdpython? change name please.

dipy/reconst/dki.py
```diff
@@ -0,0 +1,1634 @@
...
+    r"""
```
Garyfallidis: Hi @RafaelNH and @arokem,

Many of the functions need correction for PEP8 and the math equations will not render as they are in docstrings.
Please you save some good time to correct these.

Also in your functions you usually write WIP in the documentation. This is not a recommended practice. Try to write the documentation of your functions as you go. Otherwise you will end up with bad documentation (quickly written). As you write a function and you still have good memory of what it does it is easier to write the documentation. It makes it easier also for us to give you good reviews.

dipy/reconst/dki.py
```diff
@@ -0,0 +1,1634 @@
...
+    WIP
+    """
+    d1mach=np.zeros(5)
+    d1mach[0]=1.1*10**(-308)
```
Garyfallidis: pep8: space between operators. Please correct everywhere.

dipy/reconst/dki.py
```diff
@@ -0,0 +1,1634 @@
...
+    ec = ea - eb
+    ed = ea - 6.0*eb
+    ef = ed + ec + ec
+    s1 = ed*(-c1+0.250*c3*ed-1.50*c4*zndev*ef)
+    s2 = zndev*(c2*ef+zndev*(-c3*ec+zndev*c4*ea))
+    drd = 3.0*sigma + power4*(1.0+s1+s2)/(mu*np.sqrt(mu))
+    return drd
+
+
+def rfpython(x,y,z):
```
Garyfallidis: This function name is not clear. Is "python" really useful here? And what rf means?

dipy/reconst/dki.py
```diff
@@ -0,0 +1,1634 @@
...
+
+  ### the following condition has to be checked ###
+    indexesxcond3=np.logical_or.reduce(abc<=0)
+    Carray[indexesxcond3]=0   
+    return Carray  
+
+
+def F1m(a,b,c):
+    """
+    Helper function that computes function $F_1$ which is required to compute
```
Garyfallidis: It is recommended to start the documentation from the first line of the comments.
""" Helper function

dipy/reconst/dki.py
```diff
@@ -0,0 +1,1634 @@
...
+    the analytical solution of the Mean kurtosis.
+    
+    Parameters
+    ----------
+    a : ndarray (...)
```
Garyfallidis: Here the parentheses are not informative. Look at other dipy functions for inspirations of how to explain your variables.

dipy/reconst/dki.py
```diff
@@ -0,0 +1,1634 @@
...
+    Returns
+    -------
+    F1 : array
+       Value of the function $F_1$ given the parameters a, b, and c
+
+    Notes
+    --------
+    Function F_1 is defined as [1]_:
+
+    \begin{multline}
```
Garyfallidis: I bet that this equation here does not render correctly.

dipy/reconst/dki.py
Garyfallidis: So, what do you mean here that your variables can have any dimensions? Please clarify.

dipy/reconst/dki.py
```diff
@@ -0,0 +1,1634 @@
...
+    -------
+    mk : array
+        Calculated MK.
+
+    Notes
+    --------
+    The MK analytical solution is calculated using the following equation [2]_:
+
+    .. math::
+
```
Garyfallidis: There is a lot of math here. Make sure that this documentation renders correctly.

dipy/reconst/dki.py
```diff
@@ -0,0 +1,1634 @@
...
+    \end{multline}
+
+    References
+    ----------
+    .. [1] Tabesh, A., Jensen, J.H., Ardekani, B.A., Helpern, J.A., 2011.
+           Estimation of tensors and tensor-derived measures in diffusional
+           kurtosis imaging. Magn Reson Med. 65(3), 823-836
+    """
+    # Float error used to compare two floats, abs(l1 - l2) < er for l1 = l2 
+    er = 1e-10
```
Garyfallidis: An easier and more robust way to get the machine epsilon across different architectures for a 
given float type is to use np.finfo():

print(np.finfo(float).eps)
# 2.22044604925e-16

print(np.finfo(np.float32).eps)
# 1.19209e-07

So, 1e-10 that you use here could be not enough for 64bit and too low for 32bit.

dipy/reconst/dki.py
```diff
@@ -0,0 +1,1634 @@
...
+    """
+    if sphere == None:
+        MK = _MK_analytical_solution(dki_params)
+    else:
+        MK = np.mean(apparent_kurtosis_coef(dki_params, sphere), axis=-1)
+
+    return MK
+
+
+def _MK_analytical_solution(dki_params):
```
Garyfallidis: In python we try to not use capital letters in function names. But we use capital letters in classes.

dipy/reconst/dki.py
```diff
@@ -0,0 +1,1634 @@
...
+    6 * V[:, 0] * V[:, 0] * V[:, 2] * V[:, 2] * kt[10] + \
+    6 * V[:, 1] * V[:, 1] * V[:, 2] * V[:, 2] * kt[11] + \
+    12 * V[:, 0] * V[:, 0] * V[:, 1] * V[:, 2] * kt[12] + \
+    12 * V[:, 0] * V[:, 1] * V[:, 1] * V[:, 2] * kt[13] + \
+    12 * V[:, 0] * V[:, 1] * V[:, 2] * V[:, 2] * kt[14]
+
+    return (MD/ADC) ** 2 * AKC
+
+
+def DKI_prediction(dki_params, gtab, S0=150, snr=None):
```
Garyfallidis: Again here you use capital letters in function names. Please correct everywhere.

dipy/reconst/dki.py
```diff
@@ -0,0 +1,1634 @@
...
+        DT = np.dot(np.dot(fevecs[v], np.diag(fevals[v])), fevecs[v].T)
+        pred_sig[v] = DKI_signal(gtab, lower_triangular(DT), fkt[v], S0, snr)
+
+    # Reshape data according to the shape of dki_params
+    pred_sig = pred_sig.reshape(dki_params.shape + len(pred_sig))
+
+    return pred_sig
+
+
+class DKIModel(ReconstModel):
```
Garyfallidis: Non standard naming for class. Please write
class DiffusionKurtosisModel

dipy/reconst/dki.py
```diff
@@ -0,0 +1,1634 @@
...
+            The gradient table for this prediction
+
+        S0 : float or ndarray (optional)
+            The non diffusion-weighted signal in every voxel, or across all
+            voxels. Default: 1
+        """
+        return DKI_prediction(dki_params, self.gtab, S0)
+
+
+class DKIFit(TensorFit):
```
Garyfallidis: Same as above. Please write
DiffusionKurtosisFit

Using longer and more expressive names rather than shortcuts make the code much easier to read. 
Having as much readable code as we can is one of our missions. Keep it up!

dipy/reconst/dki.py
```diff
@@ -0,0 +1,1634 @@
...
+        Parameters are ordered as follow:
+            1) Three diffusion tensor's eingenvalues
+            2) Three lines of the eigenvector matrix each containing the first
+               second and third coordinates of the eigenvector
+            3) Fifteen elements of the kurtosis tensor 
+
+    See Also
+    --------
+    decompose_tensors
+
```
Garyfallidis: Reduce unneeded blank lines in docstrings.

dipy/reconst/dki.py
```diff
@@ -0,0 +1,1634 @@
...
+        Fifteen elements of the kurtosis tensor
+    """    
+    evals = dki_params[..., :3]
+    evecs = dki_params[..., 3:12].reshape(dki_params.shape[:-1] + (3, 3))
+    kt = dki_params[..., 12:]
+    
+    return evals, evecs, kt
+
+
+def dki_design_matrix(gtab):
```
Garyfallidis: Because you are already at dki.py module there is no reason repeating dki in front of function names. So, you can use design_matrix here rather than dki_design_matrix.
Now someone would say that that could be a problem if you want to call design_matrix from two different models e.g. dti and dki. In that case you can write
```
from dipy.reconst.dti import design_matrix as dti_design_matrix
from dipy.reconst.dki import design_matrix as dki_design_matrix
```
and resolve any naming issues.

dipy/reconst/tests/test_dki.py
```diff
@@ -0,0 +1,315 @@
...
+
+import numpy as np
+
+from nose.tools import (assert_true, assert_equal,
+                        assert_almost_equal, assert_raises)
+from numpy.testing import (assert_array_equal, assert_array_almost_equal,
+                           assert_)
+
+from dipy.sims.voxel import multi_tensor_dki
+
```
Garyfallidis: remove blank line

dipy/reconst/tests/test_dki.py
```diff
@@ -0,0 +1,315 @@
...
+import dipy.reconst.dti as dti
+
```
Garyfallidis: as above

dipy/reconst/tests/test_dki.py
```diff
@@ -0,0 +1,315 @@
...
+                                   axis=0)
+
+# Since KT is spherical, in all directions AKC and MK should be
+f = 0.5
+Dg = f*Di + (1-f)*De
+Kref_sphere = 3 * f * (1-f) * ((Di-De) / Dg) ** 2
+
+
+def test_dki_fits():
+    """DKI fits are tested on noise free simulates"""
```
Garyfallidis: simulations

doc/examples/reconst_dki.py
```diff
@@ -0,0 +1,314 @@
...
+=====================================================================
+
+The diffusion kurtois model is an expansion of the diffusion tensor model. In
+addition to the diffusio tensor (DT), the diffusion kurtosis model quantifies
+the diffusion kurtosis tensor (KT) which measures the degree to which water
+diffusion in biologic tissues is non-Gaussian [Jensen2005]_. Measurements of
+non-Gaussian diffusion are of interest since they can be used to charaterize
+tissue microstructural heterogeneity [Jensen2010]_ and to derive concrete
+biophysical parameters as the density of axonal fibres and diffusion tortuosity
+[Fierem2011]_. Moreover, DKI can be used to resolve crossing fibers on
```
Garyfallidis: Correct english please. I would stop at crossing fibers. On tractography doesn't read well.

doc/examples/reconst_dki.py
```diff
@@ -0,0 +1,314 @@
...
+                    & ... \\
+                    & W_{xyyy} & W_{yyyz} & W_{xzzz} & W_{yzzz} & W_{xxyy}
+                    & ... \\
+                    & W_{xxzz} & W_{yyzz} & W_{xxyz} & W_{xyyz} & W_{xyzz}
+                    & & )\end{matrix}
+
+In the following example we show how to reconstruct your diffusion multi-shell
+datasets using the kurtosis tensor model.
+
+First import the necessary modules:
```
Garyfallidis: There is no need to explain the imports. We did that already in some other tutorials. There is no need to repeat the same information.

arokem: What if this is the first tutorial a person reads? I don't know that we can
assume that users will read other tutorials before this one, if all they
want is to do DKI.

dipy/reconst/dki.py
arokem: Hi @garyfallidis. Thanks for all the comments!

The "WIP" markers here are to tell you that these functions are still work
in progress, and will change substantially before they are really ready for
review. For now, it would be good to focus on the code below line 435 or so
of this file (`def mean_kurtosis`). In particular, we are still working
through the math of the first few functions, and when we sort it all out,
there will also be clearer documentation here.

arokem: More specifically to your point: these functions are taken from Maurizio's
implementation, so they were copy pasted verbatim in here. As mentioned, we
need to still fully understand these implementations (and verify them
through tests), but once we have that under control, we can also document
them. In principle, I absolutely agree with you that documenting as you
implement is much better.

dipy/reconst/dki.py
RafaelNH: The function dipy.recont.dti is implemented specifically for the DTI, so we cannot call this function here. This will be only possible if we generalize the DTI fitting algorithms. For this, we have to remove the decompose_tensor step and change the variable output format. Tensor decomposition have to be done instead in the main. This changes are feasible, however I will recommend doing this in another pull request, this one is getting long.

dipy/reconst/dki.py
RafaelNH: Indeed, I am importing the function lower_triangular from reconst.dti

dipy/reconst/dki.py
RafaelNH: Does this really make sense for the predict function defined in the class KurtosisModel? Why should you define a class model with a gtab just to make a prediction with a given model parameters and a different gradient set, if you can do it directly for function dki_prediction without defining any class? 

If you think this should be done, this change is trivial. However, I will recommend doing this also in reconst.dti to keep consistency.

dipy/reconst/dki.py
RafaelNH: Hi @Garyfallidis!
As @arokem mentioned the functions marked in WIP are still in the version taken from Mauruzio's. Normally, I write the documentation at the same time that I am writing the functions. In this case, I am writing the documentation at the same time that I am revising the articles and the same time that I am correcting the code :).

dipy/reconst/dki.py
RafaelNH: Cool!!!!

RafaelNH:
@MrBago and @Garyfallidis - thanks for the comments! Indeed this PR was to long!

As suggested by @arokem, I removed for now all the functions regarding the estimation of standard kurtosis measures. This PR contains now only the methods for fitting the diffusion and kurtosis tensor from the DKI model (as the PR title suggests). Standard rotational invariant kurtosis measures will be submitted on a third PR after the revision of this one.

Please let me know what further changes should be done to complete this PR. Further comments are more than welcome =).

dipy/reconst/dki.py
```diff
@@ -0,0 +1,743 @@
...
+        cause quite a lot of noise in diffusion based metrics, diffusivity
+        values smaller than `min_diffusivity` are replaced with
+        `min_diffusivity`. defaut = 0
+
+    min_kurtosis : float (optional)
+        Because high amplitude negative values of kurtosis are not physicaly
+        and biologicaly pluasible, and these causes huge artefacts in kurtosis
+        based measures, directional kurtosis values than `min_kurtosis` are
+        replaced with `min_kurtosis`. defaut = -1
+
```
arokem: `Returns`?

dipy/reconst/dki.py
```diff
@@ -0,0 +1,743 @@
...
+        ----------
+        dki_params : ndarray (x, y, z, 27) or (n, 27)
+            All parameters estimated from the diffusion kurtosis model.
+            Parameters are ordered as follow:
+                1) Three diffusion tensor's eingenvalues
+                2) Three lines of the eigenvector matrix each containing the
+                   first, second and third coordinates of the eigenvector
+                3) Fifteen elements of the kurtosis tensor
+
+        gtab : a GradientTable class instance
```
arokem: Whoops. This is not a parameter here.

dipy/reconst/dki.py
```diff
@@ -0,0 +1,743 @@
...
+
+        S0 : float or ndarray (optional)
+            The non diffusion-weighted signal in every voxel, or across all
+            voxels. Default: 1
+        """
+        return dki_prediction(dki_params, self.gtab, S0)
+
+
+class DiffusionKurtosisFit(TensorFit):
+
```
arokem: Top-level docstring for this class

dipy/reconst/dki.py
```diff
@@ -0,0 +1,743 @@
...
+    def __init__(self, model, model_params):
+        """ Initialize a DiffusionKurtosisFit class instance.
+
```
arokem: Docstring should explain the parameters

dipy/reconst/dki.py
```diff
@@ -0,0 +1,743 @@
+#!/usr/bin/python
+""" Classes and functions for fitting tensors """
```
arokem: This top-level docstring seems to be wrong...

dipy/reconst/dki.py
```diff
@@ -0,0 +1,743 @@
...
+        ----------
+        sphere : Sphere class instance
+
+        Returns
+        -------
+        akc : ndarray
+           The estimates of the apparent kurtosis coefficient in every
+           direction on the input sphere
+
+        Notes
```
arokem: This `Notes` section needs to be rewritten

dipy/reconst/dki.py
```diff
@@ -0,0 +1,743 @@
...
+        .. math ::
+
+            ADC = \vec{b} Q \vec{b}^T
+
+        Where Q is the quadratic form of the tensor.
+        """
+        return apparent_kurtosis_coef(self.model_params, sphere)
+
+    def dki_predict(self, gtab, S0=1):
+        r""" Given a DKI model fit, predict the signal on the vertices of a
```
arokem: More accurate to say that the prediction is made for a gradient table, rather than a sphere

dipy/reconst/dki.py
```diff
@@ -0,0 +1,743 @@
...
+    wls_fit_dki
+
+    References
+    ----------
+       [1] Tabesh, A., Jensen, J.H., Ardekani, B.A., Helpern, J.A., 2011.
+           Estimation of tensors and tensor-derived measures in diffusional
+           kurtosis imaging. Magn Reson Med. 65(3), 823-836
+    """
+
+    tol = 1e-6
```
arokem: Could we make the tolerance an argument, with this default value, rather than have it be a hard-coded constant? I understand this is a tolerance that defines the minimal diffusivity allowed in the solution. Is that correct? Maybe give this variable a more informative name?

dipy/reconst/dki.py
```diff
@@ -0,0 +1,743 @@
...
+    if min_signal <= 0:
```
arokem: For dealing with minimum signal, I prefer the solution we eventually implement in the DTI module: 

See this PR for the details: https://github.com/nipy/dipy/pull/498

dipy/reconst/dki.py
```diff
@@ -0,0 +1,743 @@
...
+    -------
+    dki_params : array (27, )
+        All parameters estimated from the diffusion kurtosis model.
+        Parameters are ordered as follow:
+            1) Three diffusion tensor's eingenvalues
+            2) Three lines of the eigenvector matrix each containing the first,
+               second and third coordinates of the eigenvector
+            3) Fifteen elements of the kurtosis tensor
+    """
+    A = design_matrix
```
arokem: Is this just a convenience so that the code is easier to read below?

doc/examples/reconst_dki.py
```diff
@@ -0,0 +1,270 @@
...
+"""
+
+FA = dkifit.fa
+MD = dkifit.md
+AD = dkifit.ad
+RD = dkifit.rd
+
+"""
+Note that these DTI standard measures could also be computed from Dipy's DTI
+module. However, DTI models is only viable for b-values < 1500 $s.mm^{-2}$. For
```
arokem: In what sense is DTI "only viable for b-values < 1500"? It provides a very accurate fit to the data even for 2k and 4k.

doc/examples/reconst_dki.py
```diff
@@ -0,0 +1,270 @@
...
+=====================================================================
+
+The diffusion kurtois model is an expansion of the diffusion tensor model. In
+addition to the diffusion tensor (DT), the diffusion kurtosis model quantifies
+the degree to which water diffusion in biologic tissues is non-Gaussian using
+the kurtosis tensor [Jensen2005]_. Measurements of non-Gaussian diffusion are
+of interest because they can be used to charaterize tissue microstructural
+heterogeneity [Jensen2010]_ and to derive concrete biophysical parameters as
+the density of axonal fibres and diffusion tortuosity [Fierem2011]_. Moreover,
+DKI can be used to resolve crossing fibers on tractography and to obtain
```
arokem: "on tractography" => "in tractography"

doc/examples/reconst_dki.py
```diff
@@ -0,0 +1,270 @@
...
+datasets using the kurtosis tensor model.
+
+First import Dipy's DKI module:
+
+"""
+
+import dipy.reconst.dki as dki
+
+"""
+DKI requires multi shell data, i.e. data acquired from more than one non-zero
```
arokem: "multi shell" => "multi-shell"

doc/examples/reconst_dki.py
```diff
@@ -0,0 +1,270 @@
...
+from dipy.data import read_sherbrooke_3shell
+
+img, gtab = read_sherbrooke_3shell()
+
+data = img.get_data()
+
+"""
+img contains a nibabel Nifti1Image object (with the data) and gtab contains a
+GradientTable object with information about the b-values and b-vectors. The
+b-values used on the loaded dataset are visualized above
```
arokem: Do you mean "visualized below"?

doc/examples/reconst_dki.py
```diff
@@ -0,0 +1,270 @@
...
+fig1.savefig('HARDI193_bvalues.png')
+
+"""
+.. figure:: HARDI193_bvalues.png
+   :align: center
+   **b-values of the loaded dataset**.
+
+From the figure above, we can check that the loaded dataset contains three
+non-zero b-values as required for DKI. However the highest b-value of 3500
+$s.mm^{-2}$ is higher than normally used on DKI. Since DKI negletes diffusion
```
arokem: "negletes" => "neglects"

doc/examples/reconst_dki.py
```diff
@@ -0,0 +1,270 @@
...
+signal components higher than the 4th order KT, a upper bound of b-value < 3000
+$s.mm^{-2}$ is normally implied to insure the fitting viability[Jensen2010]_.
```
arokem: Again, I am not sure what "viability" means here.

dipy/reconst/dki.py
RafaelNH: Yup! Should I avoid this?

arokem: No - it's fine. Maybe just add a comment.

arokem:
Are you all done here? 

I am hoping that the recent failures on Travis were just some weird glitch. If this comes back green, is this ready for a merge? Anyone else have any comments at this point?

RafaelNH:
I am trying to understand why Travis is failing. The problem happens when I try to import the design_matrix function from dipy.reconst.dki to dipy.sims.voxel. I am defining again the function design_matrix on dipy.sims.voxel to overcame the Travis failures. Anyone have a clue on how we can define this function directly on dipy.reconst.dki and importing it to the simulations without having this failure?

RafaelNH:
Does anyone have suggestions on how to sort the problem that I mentioned on my previous comment?

Anyone have any comments before merging this pull request?

arokem:
Oh - sorry -  I missed that last comment. We definitely need to resolve this first. Might  be a circular import, or some-such. Will take a look.

arokem:
Definitely a circular import:

dki => dti => data => sims => dki

arokem:
Still thinking of a solution

arokem:
This is my proposal: https://github.com/RafaelNH/dipy/pull/1

The idea I have is to create a `reconst.utils` module that will contain the implementation of `dki_design_matrix`, and then import from here both in `dki` and in `sims.voxel`

RafaelNH:
I think I answer all your comments regarding this Pull request. Are there any other aspect that I should address at this point? Or is this ready to be merged?

arokem:
This is good to go from my end. Let's give others a few days to take a look. If no one objects, I will merge this on Monday.

arokem:
I'm going ahead and merging this. It will be quickly followed by the next PR in the series, so feel free to comment on any of this for the next PR as well.

PR #666, opened 2015-06-16
Parallelized local tracking branch so now you can actually look at my code :)

amitvakula:


dipy/tracking/local/optimizedlocaltracking.py
```diff
@@ -0,0 +1,420 @@
...
+            continue
+
+        if stepsB == 1:
+            streamline = F[:stepsF].copy()
+        else:
+            parts = (B[stepsB-1:0:-1], F[:stepsF])
+            streamline = np.concatenate(parts, axis=0)
+        return streamline
+
+class OptimizedLocalTracking(object):
```
MrBago: This class seems like it's mostly a copy of the `LocalTracking` class. You should either change the class directly or use inheritance. We're going to need you to do that before we merge it, plus it makes it much easier for me to see what you've changed :).

dipy/tracking/local/optimizedlocaltracking.py
```diff
@@ -0,0 +1,420 @@
...
+        args.append(vs)
+        args.append(dg)
+        args.append(tc)
+        args.append(affine)
+        args.append(ss)
+        args.append(max_cross)
+        args.append(maxlen)
+        args.append(fixed)
+        args.append(return_all)
+        return args
```
MrBago: this is easier to read if you right `args = [inv_A, lin, offset, ..., return_all]`.

dipy/reconst/peaks.py
```diff
@@ -157,8 +157,59 @@ def peak_directions(odf, sphere, relative_peak_threshold=.5,
...
+        self.rebuild_class_state(*state["class_state"])
+        self.initialized = state["is_initialized"]
+
+
+        if self.initialized:
+            self._initialize()
+            current_state = state["current_state"]
+            self._qa = current_state["_qa"]
+            self._ind = current_state["_ind"]
+            self._odf_vertices = current_state["_odf_vertices"]
```
MrBago: These assignments are already done in `self._initialize()` so do we need to do them again?

MrBago:
We should meet and talk about the `set_state` stuff, but here are my thoughts in summery.

1) We don't need `__setstate__`, we can do everything using `__init__` (I think). That should make the code cleaner.
2) We should make the code more object oriented. That means either a) only defining each method once, either in the parent or child as needed or b) having the child method call the parent method as needed.

Garyfallidis:
Ping! @amitvakula  and @MrBago Any progress with this?

Garyfallidis:
@MrBago one of the reasons why tracking is sometimes slow is because some of the algorithms like max odf needs to calculate the ODF on the fly. Could we use some different representation to speedup the process? Do we really need those large matrix multiplications or we can do this with some smaller matrix multiplications like in Cholesky?

Garyfallidis:
What about this PR? Is something better on the way? Otherwise we need to close it.

MrBago:
I think this is on hold, lets close this and we'll make a new PR if need be.

PR #654, opened 2015-05-24
Affine registration PR 3/3

omarocegueda:
This is the last PR of the affine registration series. It also includes a tutorial explaining how to do an ANTS2-like affine registration. The following graph shows the performance of this implementation compared to ANTS2, using sparse sampling (30% of the voxels).
image

EDIT: we have now the same accuracy as ANTS2 in this dataset (image updated). Our implementation is a bit faster too (still a bit slower than ANTS2), the following are the updated times:

(sparse means use 30% of the voxels for PDF estimation)
ANTS2 sparse: 11.0 minutes 
Dipy sparse: 12.0 minutes 
Flirt : 4.41 minutes
Nipy: 49 seconds

PR #669, opened 2015-06-24
Function to reorient gradient directions according to moco parameters

arokem:
Loosely based on @rfdougherty's implementation here: https://github.com/cni/nims/blob/master/scripts/pepolar_unwarp.py 

TODO: 
- [x] More thorough testing.
- [x] Allow inputs of shape (n, 4, 4) if users get this kind of output from their motion correction algorithm.

### DIFF
--- a/dipy/core/gradients.py
+++ b/dipy/core/gradients.py
@@ -252,8 +252,9 @@ def reorient_bvecs(gtab, rots):
 
     When correcting for motion, rotation of the diffusion-weighted volumes
     might cause systematic bias in rotationally invariant measures, such as FA
-    and MD, and cause , unless the gradient directions are appropriately
-    reoriented to compensate for this effect [Leemans2009]_.
+    and MD, and also cause characteristic biases in tractography, unless the
+    gradient directions are appropriatel reoriented to compensate for this
+    effect [Leemans2009]_.
 
     Parameters
     ----------
--- a/dipy/core/gradients.py
+++ b/dipy/core/gradients.py
@@ -261,10 +261,12 @@ def reorient_bvecs(gtab, affines):
     ----------
     gtab : a GradientTable class instance.
         The nominal gradient table with which the data were acquired.
-    affines : list or ndarray of shape (n, 4, 4)
-        Each entry in this list or array contains an affine transformation
-        that encodes the rotation that was applied to the image corresponding
-        to this bvector as part of a motion correction procedure.
+    affines : list or ndarray of shape (n, 4, 4) or (n, 3, 3)
+        Each entry in this list or array contain either an affine
+        transformation (4,4) or a rotation matrix (3, 3).
+        In both cases, the transformations encode the rotation that was applied
+        to the image corresponding to one of the non-zero gradient directions
+        (ordered according to their order in `gtab.bvecs[~gtab.b0s_mask]`)
 
     Returns
     -------
@@ -284,10 +286,15 @@ def reorient_bvecs(gtab, affines):
         raise ValueError(e_s)
 
     for i, aff in enumerate(affines):
-        # Remove the translation component:
-        aff_no_trans = aff[:3, :3]
-        # Decompose into rotation and scaling components:
-        R, S = la.polar(aff_no_trans)
+        if aff.shape == (4, 4):
+            # This must be an affine!
+            # Remove the translation component:
+            aff_no_trans = aff[:3, :3]
+            # Decompose into rotation and scaling components:
+            R, S = la.polar(aff_no_trans)
+        elif aff.shape == (3, 3):
+            # We assume this is a rotation matrix:
+            R = aff
         Rinv = la.inv(R)
         # Apply the inverse of the rotation to the corresponding gradient
         # direction:
--- a/dipy/core/gradients.py
+++ b/dipy/core/gradients.py
@@ -252,8 +252,7 @@ def gradient_table(bvals, bvecs=None, big_delta=None, small_delta=None,
 
 
 def reorient_bvecs(gtab, affines):
-    """
-    Reorient the directions in a GradientTable according
+    """Reorient the directions in a GradientTable.
 
     When correcting for motion, rotation of the diffusion-weighted volumes
     might cause systematic bias in rotationally invariant measures, such as FA
@@ -263,7 +262,7 @@ def reorient_bvecs(gtab, affines):
 
     Parameters
     ----------
-    gtab : a GradientTable class instance.
+    gtab : GradientTable
         The nominal gradient table with which the data were acquired.
     affines : list or ndarray of shape (n, 4, 4) or (n, 3, 3)
         Each entry in this list or array contain either an affine
--- a/dipy/core/gradients.py
+++ b/dipy/core/gradients.py
@@ -89,7 +89,7 @@ class GradientTable(object):
 
 
 def gradient_table_from_bvals_bvecs(bvals, bvecs, b0_threshold=0, atol=1e-2,
-                                  **kwargs):
+                                    **kwargs):
     """Creates a GradientTable from a bvals array and a bvecs array
 
     Parameters
@@ -221,9 +221,9 @@ def gradient_table(bvals, bvecs=None, big_delta=None, small_delta=None,
     # If you provided strings with full paths, we go and load those from
     # the files:
     if isinstance(bvals, string_types):
-          bvals, _ = io.read_bvals_bvecs(bvals, None)
+        bvals, _ = io.read_bvals_bvecs(bvals, None)
     if isinstance(bvecs, string_types):
-          _, bvecs = io.read_bvals_bvecs(None, bvecs)
+        _, bvecs = io.read_bvals_bvecs(None, bvecs)
 
     bvals = np.asarray(bvals)
     # If bvecs is None we expect bvals to be an (N, 4) or (4, N) array.
@@ -239,7 +239,7 @@ def gradient_table(bvals, bvecs=None, big_delta=None, small_delta=None,
                              " array containing both bvals and bvecs")
     else:
         bvecs = np.asarray(bvecs)
-        if (bvecs.shape[1] > bvecs.shape[0])  and bvecs.shape[0]>1:
+        if (bvecs.shape[1] > bvecs.shape[0]) and bvecs.shape[0] > 1:
             bvecs = bvecs.T
     return gradient_table_from_bvals_bvecs(bvals, bvecs, big_delta=big_delta,
                                            small_delta=small_delta,
--- a/dipy/core/gradients.py
+++ b/dipy/core/gradients.py
@@ -3,6 +3,7 @@ from __future__ import division, print_function, absolute_import
 from ..utils.six import string_types
 
 import numpy as np
+import scipy.linalg as la
 
 from ..io import gradients as io
 from .onetime import auto_attr
@@ -246,28 +247,28 @@ def gradient_table(bvals, bvecs=None, big_delta=None, small_delta=None,
                                            atol=atol)
 
 
-def reorient_bvecs(gtab, rots):
+def reorient_bvecs(gtab, affines):
     """
     Reorient the directions in a GradientTable according
 
     When correcting for motion, rotation of the diffusion-weighted volumes
     might cause systematic bias in rotationally invariant measures, such as FA
     and MD, and also cause characteristic biases in tractography, unless the
-    gradient directions are appropriatel reoriented to compensate for this
+    gradient directions are appropriately reoriented to compensate for this
     effect [Leemans2009]_.
 
     Parameters
     ----------
     gtab : a GradientTable class instance.
         The nominal gradient table with which the data were acquired.
-    rots : (n, 3) array.
-        The x, y, z rotations (in radians) associated with the motion
-        parameters for each of the volumes in the experiment.
+    affines : list or ndarray of shape (n, 4, 4)
+        Each entry in this list or array contains an affine transformation
+        that encodes the rotation that was applied to the image corresponding
+        to this bvector as part of a motion correction procedure.
 
     Returns
     -------
     gtab : a GradientTable class instance with the reoriented directions
-
 
     References
     ----------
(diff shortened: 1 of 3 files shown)


dipy/core/gradients.py
```diff
@@ -244,3 +244,55 @@ def gradient_table(bvals, bvecs=None, big_delta=None, small_delta=None,
                                            small_delta=small_delta,
                                            b0_threshold=b0_threshold,
                                            atol=atol)
+
+
+def reorient_bvecs(gtab, rots):
```
omarocegueda: Where is the user expected to obtain the `rots` input from? For example, they may have the rotation matrices themselves instead of Euler angles

arokem:
No worries. This is a good point for discussion! I think that we should at least allow two kinds of input: 

1. An (n, 3) with the x,y,z rotation angles associated with each volume.
1. An (n, 4, 4) with the affine transformation associated with each volume.

Hopefully, we'll get some parameters from a motion-correction algorithm based on your current PR #654. I assume that the output from that would be more like the latter? Would maybe be good to write a little utility function that extracts the former from the latter.

dipy/core/gradients.py
```diff
@@ -244,3 +244,55 @@ def gradient_table(bvals, bvecs=None, big_delta=None, small_delta=None,
...
+    might cause systematic bias in rotationally invariant measures, such as FA
+    and MD, and also cause characteristic biases in tractography, unless the
+    gradient directions are appropriatel reoriented to compensate for this
+    effect [Leemans2009]_.
+
+    Parameters
+    ----------
+    gtab : a GradientTable class instance.
+        The nominal gradient table with which the data were acquired.
+    rots : (n, 3) array.
```
omarocegueda: I am worried that some people may use this as a black box (I'm not sure if we should add a warning here or be more specific about what the `rots` are) because there may be an issue regarding how the rotational parameters are given by different software. For example, flirt (and mcflirt, and I assume all FSL software, like eddy) returns a matrix M that maps points in physical space from the moving image towards the static image  (that is the most intuitive way for most people, because it is the moving image which is being "sent" to the static image). ANTS, on the other hand, returns a matrix that maps points in the static image towards points in the moving image, but this choice makes sense too: when we transform the moving image towards the static image, we actually send each point `x` in the static image towards its corresponding `y` in the moving image (contrary to the common intuition) so we can interpolate the moving image at `y` to obtain the intensity value that must be written at each `x` to form the transformed image.  So, whenever Flirt needs to transform an image, it actually computes the inverse of the matrix it returned, while ants simply uses the returned matrix itself. However, if people used Flirt, they must apply the returned matrix unchanged to rotate the bvecs, but if for some reason they used ANTS they need to apply the inverse of the returned matrix. Another issue is whether the transformation was computed in RAS or LPS (or any other) coordinate system, and if the voxels in the image are given in C order of Fortran order. The dangerous thing is that the transforms are often very close to the identity, which makes it very hard to notice when we are making a mistake.

MrBago: My understanding is that rotation angles are not well defined, different rotation angles can be used to represent the same transformation and different conventions exist for different applications. For example Euler angles vs Tait-Bryan angles.

Is there a way to simplify this, for example would it be easier to use an affine matrix or a "transformation object" to help standardize this?

dipy/core/gradients.py
```diff
@@ -244,3 +244,55 @@ def gradient_table(bvals, bvecs=None, big_delta=None, small_delta=None,
...
+    gtab : a GradientTable class instance with the reoriented directions
+
+
+    References
+    ----------
+    .. [Leemans2009] The B-Matrix Must Be Rotated When Correcting for
+       Subject Motion in DTI Data. Leemans, A. and Jones, D.K. (2009).
+       MRM, 61: 1336-1349
+    """
+    new_bvecs = gtab.bvecs[~gtab.b0s_mask]
```
MrBago: Are you sure you wan to just apply this to only `~gtab.b0s_mask`?

Either way, there should error checking here to make sure the number rotations matches the number of gradients. When the data has b0s mixed in with gradients (for example hc data) it's easy to occidentally pass in too many or too few rotation angles.

arokem:
Thanks for the comments! 

@omarocegueda: I think that the most important would be to be internally consistent within dipy. That is, best would be to be compatible with your registration work, because I think that we will eventually want to use your work for motion correction as well (?). I see that you are discussing implementation of a Transformation class, over on #654. Maybe the input here should be a list of these objects? I saw that you mentioned a Transformation class that is implemented in SymmetricDiffeomorphicRegistration and in StreamlineLinearRegistration. Could you please point out to me where I could find the implementation of this Transformation class? Might be useful to think of this use-case as we implement that.

omarocegueda:
Agreed, @arokem. I think each software should provide its own method for rotating the b-vectors to avoid any confusion. However, I don't know much about the state of the art on registration-based motion correction for Diffusion: as far as I know, the easiest approach is simply register all DW volumes to the B0, but I'm not convinced that this is a good approach. I have been studying FSL's  `eddy` code, which is based on a more sound approach by attempting to model the diffusion signal under the eddy-current and subject motion artifacts (it models the signal as a Gaussian process). I think we should go for something more in that direction, but for now, with the current framework it is very easy to implement the simple approach of registering all DW images to the B0 (maybe we can put that in the tutorial, like "Write your own motion correction method with b-vec toration").

> I saw that you mentioned a Transformation class that is implemented in SymmetricDiffeomorphicRegistration and in StreamlineLinearRegistration

Sure!, here is the one for streamlines:
https://github.com/nipy/dipy/blob/master/dipy/align/streamlinear.py#L395

And here is the one for Diffeomorphisms:
https://github.com/nipy/dipy/blob/master/dipy/align/imwarp.py#L403

Actually, @Garyfallidis, I think we can use one single `AffineMap` for images and streamlines, the only difference is that in the case of images it needs to provide the `grid2world` transform for each image, while for streamlines, it's not necessary because the streamlines are already in physical space (wouldn't it be nice to transform an image with the same transformation you got from streamline registration? =) ), actually, there's nothing that prevents us from defining a `transform` method for streamlines in `DiffeomorphicMap` as well (i.e., how the streamlines are warped by the displacement field?). Let me define the `AffineMap` with both capabilities and we'll see if it is convenient.

arokem:
On Mon, Jun 29, 2015 at 11:07 AM, Omar Ocegueda <notifications@github.com>
wrote:

> Agreed, @arokem <https://github.com/arokem>. I think each software should
> provide its own method for rotating the b-vectors to avoid any confusion.
> However, I don't know much about the state of the art on registration-based
> motion correction for Diffusion: as far as I know, the easiest approach is
> simply register all DW volumes to the B0, but I'm not convinced that this
> is a good approach. I have been studying FSL's eddy code, which is based
> on a more sound approach by attempting to model the diffusion signal under
> the eddy-current and subject motion artifacts (it models the signal as a
> Gaussian process). I think we should go for something more in that
> direction, but for now, with the current framework it is very easy to
> implement the simple approach of registering all DW images to the B0 (maybe
> we can put that in the tutorial, like "Write your own motion correction
> method with b-vec toration").
>
I agree that more advanced methods would be good to have in the future. For
now, I would like to have this functionality (moco by registration to mean
b0) implemented and easy to use in Dipy. My colleagues and I have used this
approach on many data-sets in the past, and it seems to work well for *a
lot* of scenarios. As you point out, the implementation of this approach
flows naturally from this current work.

Ideally, this would end up not as a tutorial, but an actual scripting API
to do this, possibly as part of a `workflows` module that we will have for
users at various levels to use in their scripting of common tasks. But, for
the time being, the tutorial you propose sounds like an excellent idea. I
think that it can be a separate PR, once this is merged.


>  I saw that you mentioned a Transformation class that is implemented in
> SymmetricDiffeomorphicRegistration and in StreamlineLinearRegistration
>
> Sure!, here is the one for streamlines:
> https://github.com/nipy/dipy/blob/master/dipy/align/streamlinear.py#L395
>
> And here is the one for Diffeomorphisms:
> https://github.com/nipy/dipy/blob/master/dipy/align/imwarp.py#L403
>
Thanks.

Upon further thought and reading through these things, I see two options:

1. Use a list of matrices that are the output of `affreg.optimize` in the
example you have here.

2. Implement a uniform API for affine transformations, that has `transorm`,
`transform_inverse` and `invert`, similar to the API you have in `imwarp`.
This could take as input one of the matrices mentioned in 1.

omarocegueda:
>y colleagues and I have used this approach on many data-sets in the past, and it seems to work well for *a lot* of scenarios

That's great @arokem!, by any chance can you share a dataset with noticeable motion that is successfully corrected with this approach? So far I have been using this year's ISMRM tractography challenge data set: http://www.tractometer.org/ismrm_2015_challenge/data (unfortunately it seems that it's going to take a long time to have access to the ground truth, though) but I have not been able to obtain a correction that looks as good as reported in FSL's eddy page: http://fsl.fmrib.ox.ac.uk/fsl/fslwiki/EDDY

matthew-brett:
Forgive me for not really following, but if you are going to go for a simple rotation parameter set, I agree it would be best to avoid Euler angles.   I would personally go for 3x3 rotation matrices, but you could also go for a rotation axis and rotation angle.

@alexis-roche - any views?  

http://matthew-brett.github.io/transforms3d/reference/transforms3d.euler.html

jchoude:
@omarocegueda Just a quick note: contrarily to your beliefs, the ground truth for the Challenge are online: http://www.tractometer.org/ismrm_2015_challenge/data

omarocegueda:
This is fantastic @jchoude! may be we can now quantitatively compare different motion correction approaches under controlled conditions, I'll take a look, thank you very much!

arokem:
I finally need this myself, so I've revisited this and made fixes according to the comments I got in the previous round. In particular, this now assumes a list or array of affine 4 by 4 transformation matrices as input, and tests for a match between the input and the number of non-zero gradients in the input gradient table.

arokem:
Apparently the `polar` transformation was only recently added to scipy, so I added this in a `fixes` file that I called `scipy`. Is this the right way to support use in older scipy?

dipy/core/gradients.py
```diff
@@ -238,9 +243,67 @@ def gradient_table(bvals, bvecs=None, big_delta=None, small_delta=None,
...
+
+    When correcting for motion, rotation of the diffusion-weighted volumes
+    might cause systematic bias in rotationally invariant measures, such as FA
+    and MD, and also cause characteristic biases in tractography, unless the
+    gradient directions are appropriately reoriented to compensate for this
+    effect [Leemans2009]_.
+
+    Parameters
+    ----------
+    gtab : a GradientTable class instance.
```
Garyfallidis: GradientTable is sufficient no need for a...

dipy/core/gradients.py
```diff
@@ -238,9 +243,67 @@ def gradient_table(bvals, bvecs=None, big_delta=None, small_delta=None,
...
+        The nominal gradient table with which the data were acquired.
+    affines : list or ndarray of shape (n, 4, 4) or (n, 3, 3)
```
Garyfallidis:
```
affines : list of ndarrays of shape (4, 4) or (3, 3) or ndarray of shape (n, 4, 4) or (n, 3, 3)
```

dipy/core/tests/test_gradients.py
```diff
@@ -183,6 +186,69 @@ def test_qvalues():
...
+    # The simple case: all affines are identity:
+    affs = np.zeros((6, 4, 4))
+    for i in range(4):
+        affs[:, i, i] = 1
+
+    # We should get back the same b-vectors:
+    new_gt = reorient_bvecs(gt, affs)
+    npt.assert_equal(gt.bvecs, new_gt.bvecs)
+
+    # Now apply some rotations to these suckers:
```
Garyfallidis: ??

dipy/core/tests/test_gradients.py
```diff
@@ -183,6 +186,69 @@ def test_qvalues():
...
+        cos_rot = np.cos(rot_ang)
+        sin_rot = np.sin(rot_ang)
+        rotation_affines.append(np.array([[1, 0, 0, 0],
+                                         [0, cos_rot, -sin_rot, 0],
+                                         [0, sin_rot, cos_rot, 0],
+                                         [0, 0, 0, 1]]))
+        rotated_bvecs[i] = np.dot(rotation_affines[-1][:3, :3],
+                                  bvecs[i])
+
+    # Copy over the rotation affines:
```
Garyfallidis: no need for ``:`` in the comments.

dipy/core/gradients.py
```diff
@@ -238,9 +243,67 @@ def gradient_table(bvals, bvecs=None, big_delta=None, small_delta=None,
...
             bvecs = bvecs.T
     return gradient_table_from_bvals_bvecs(bvals, bvecs, big_delta=big_delta,
                                            small_delta=small_delta,
                                            b0_threshold=b0_threshold,
                                            atol=atol)
+
+
+def reorient_bvecs(gtab, affines):
+    """
+    Reorient the directions in a GradientTable according
```
Garyfallidis: Reduce docstring size by starting from the first line of the docstring

dipy/fixes/scipy.py
```diff
@@ -0,0 +1,95 @@
...
+    Returns the factors of the polar decomposition [1]_ `u` and `p` such
+    that ``a = up`` (if `side` is "right") or ``a = pu`` (if `side` is
+    "left"), where `p` is positive semidefinite.  Depending on the shape
+    of `a`, either the rows or columns of `u` are orthonormal.  When `a`
+    is a square array, `u` is a square unitary array.  When `a` is not
+    square, the "canonical polar decomposition" [2]_ is computed.
+
+    Parameters
+    ----------
+    a : (m, n) array_like
```
Garyfallidis: We usually write the shape after the object like in here
http://docs.scipy.org/doc/scipy-0.14.0/reference/generated/scipy.spatial.Delaunay.html
It would be nice if we could follow a standard for this.

dipy/core/gradients.py
arokem: That doesn't fit within 80 characters.

dipy/fixes/scipy.py
arokem: This is copied straight from the scipy code-base. But will do.

arokem:
Rebased, and addressed all the comments by @Garyfallidis.

PR #670, opened 2015-06-26
Tissue classification using MAP-MRF

villalonreina:
This is an implementation of the MAP-MRF (Maximum a posteriori - Markov Random Field) by using the Iterative Conditional Modes (ICM) minimization algorithm. There are three inputs, the grey level T1 brain image, its corresponding mask and an initial three class segmentation of the T1. 

For now we have hard-coded paths and I can share the images with whoever wants to test the code. 

This code is still experimental and needs to be extended to estimate the partial volumes of the tissue classes.


matthew-brett:
Can you put the images into their own repo, so we can add them as a submodule for testing?

Garyfallidis:
I believe that Julio cannot share these images right now.

dipy/segment/rois_stats.py
```diff
@@ -0,0 +1,37 @@
...
+
+    Returns
+    -------
+    mu, std : float
+        Mean and standard deviation for every class
+
+
+    """
+    mu = np.zeros(3)
+    std = np.zeros(3)
```
MrBago: You have nclass as an argument than you hard code 3 here.

omarocegueda:
For testing purposes, we may stack a few copies of this coronal slice of a T1-weighted image:
https://github.com/nipy/dipy/blob/master/dipy/data/t1_coronal_slice.npy

Garyfallidis:
Thx @omarocegueda :)

dipy/segment/rois_stats.py
MrBago: Try this:
```
import numpy as np

def seg_stats(input_image, seg_image, ddof=0):
    img = input_image.ravel()
    seg = seg_image.ravel()
    if not isinstance(seg.dtype, np.integer):
        seg = seg.astype(np.int64)

    seg_vol = np.bincount(seg)
    sum_img = np.bincount(seg, img)
    sum_sq = np.bincount(seg, img ** 2)

    mean = sum_img / seg_vol
    var = sum_sq / (seg_vol + ddof) - mean ** 2
    std = np.sqrt(var)
    return mean[1:], std[1:]
```

MrBago:
It seems like most of functions are getting their own files, is there a reason most of this isn't in one file?

dipy/segment/threshold.py
```diff
@@ -37,3 +37,46 @@ def otsu(image, nbins=256):
...
     return threshold
+    
+    def otsu_3(image, nbins=256):
+        
+            """
+    Return two threshold values based on Otsu's method.
+    It is intended to work on a masked T1 brain image.
+    It is intended to give threshold values for CSF/WM and WM/GM. 
+    
+    Copied from scikit-image to remove dependency.
```
arokem: Please don't forget to copy over the license from scikit-image as well

dipy/segment/energy_mrf.py
```diff
@@ -0,0 +1,57 @@
...
+def total_energy(masked_image, masked_segmentation,
+                 mu, var, index, label, beta):
+
+    energytotal = log_likelihood(masked_image, mu, var, index, label)
+    energytotal += gibbs_energy(masked_segmentation, index, label, beta)
+
+    return energytotal
+
+
+def log_likelihood(img, mu, var, index, label):
```
omarocegueda: This is actually the negative log-likelihood, right?

dipy/segment/energy_mrf.py
```diff
@@ -0,0 +1,57 @@
...
+
+    loglike = ((img[index] - mu[label]) ** 2) / (2 * var[label])
+
+    loglike += np.log(np.sqrt(var[label]))
+
+    return loglike
+
+
+def gibbs_energy(seg, index, label, beta):
```
omarocegueda: Iterating over the neighbors of a voxel can be coded more compactly by defining arrays of offsets:
https://github.com/nipy/dipy/blob/master/dipy/align/sumsqdiff.pyx#L319
Then you just iterate:
https://github.com/nipy/dipy/blob/master/dipy/align/sumsqdiff.pyx#L357
adding the offsets to the coordinates of the voxel:
https://github.com/nipy/dipy/blob/master/dipy/align/sumsqdiff.pyx#L358
https://github.com/nipy/dipy/blob/master/dipy/align/sumsqdiff.pyx#L361
https://github.com/nipy/dipy/blob/master/dipy/align/sumsqdiff.pyx#L364

dipy/segment/tests/test_icm_map.py
```diff
@@ -0,0 +1,70 @@
...
+nclass = 3
+L = range(1, nclass + 1)
+niter = 1
+beta = 0.05
+totalE = np.zeros(nclass)
+N = 0
+
+segmented = np.zeros(dataimg.shape)
+
+while True:
```
omarocegueda: With `for N in range(niter):` you no longer need the break at lines 60-61

dipy/segment/energy_mrf.py
```diff
@@ -0,0 +1,57 @@
...
+
+    energy = 0
+
+    if label == seg[index[0] - 1, index[1], index[2]]:
```
omarocegueda: I think you need to check if you are at the boundary of the image and do not accumulate the energy in that case, otherwise this will wrap to the other side in Python (not the desired behavior) and may cause a memory access violation in Cython (I think we will definitely need to cythonize)

omarocegueda:
I think we definitely need to code this in cython. Can we define a base class? like `ImageSegmenter`  that has a method `segment(image)` and returns the segmented image. Then we can derive `FASTImageSegmenter(ImageSegmenter)` which contains this particular implementation of the Maximum A-Posteriori estimation of the HMRF with ICM (`ImageSegmenter` doesn't sound good, any ideas?).

I agree with @MrBago, I think you can put most of this code into one single file.

dipy/segment/energy_mrf.py
villalonreina: Yes it is. My understanding is that this the one you minimize. The positive one is the one that you would maximize, is that right?

Thanks

dipy/segment/energy_mrf.py
villalonreina: This is great. Yes I saw this somewhere else, but first wanted to try the simpler version of it to test it out. Thanks a lot.

dipy/segment/energy_mrf.py
omarocegueda: Yes, that's right, I just thought that something like `neg_log_likelihood` would be a better name for this function

samuelstjean:
As a late reply to all the comments at the top for a t1 image (isn't there a way to reply to a specific comment, or this is a linear discussion?) : There is already one from the standford dataset, see this example http://nipy.org/dipy/examples_built/linear_fascicle_evaluation.html. It seems to be at 2 mm iso instead of the more common 1 mm, but at least anyone can play with it easily and get the same results you would have.

dipy/segment/threshold.py
samuelstjean: There is also some stuff you can probably reuse from http://nipy.org/dipy/reference/dipy.segment.html#dipy.segment.mask.median_otsu

dipy/segment/energy_mrf.py
```diff
@@ -0,0 +1,57 @@
...
+        energy = energy - beta
+    else:
+        energy = energy + beta
+
+    if label == seg[index[0] + 1, index[1], index[2]]:
```
omarocegueda: Just curious: here, the Ising/Potts potential is -beta if the labels are equal and +beta if the labels are different. I remember that I have seen this "symmetric" version before, but it is not the standard Ising/Potts potential, it should be -beta*Delta(x_{i}-x_{j}). In other words: it should be zero if the labels are equal instead of -beta, as appears in Zhang's paper, first paragraph page 49. Any reason why we prefer the symmetric version here?

villalonreina:
Hi all, 

I got confused when trying to follow the algorithm proposed in the Zhang 2001 paper (http://www.ncbi.nlm.nih.gov/pubmed/11293691). I am not sure about which equation(s) goes into each of the steps of the EM algorithm. 

So, for step E should I use the function for Q (equation 24) and then I maximize in step M equations 25 and 26?

This is a list of steps of what I am thinking about:

1) Estimate label parameters for L range(1, 3) from input segmentation

2) First Loop of ICM, which will give me a segmentation (\hat{x})  and P(\ell | N_i) (the probability of L given the neighborhood)

3) EM loop to  estimate the parameters mu and sigma 

4) Back to ICM 

Does this sound reasonable? What do you think?

dipy/segment/energy_mrf.py
villalonreina: Hi Omar, 

Thanks for you comment. Can you point me to where exactly in the Zhang paper they mention the "non-symmetric" version?

Thanks

omarocegueda:
yes, Q in equation 24 is the function that needs to be maximized with respect to the parameters (the parameters in this case are the mu and sigma for each class). The good news is that the authors already did it for us! they did the math and found the the mu and sigma that maximize Q are given by equations 25 and 26.

Note that the key quantities you need to compute in preparation for eqs. 25 and 25 are given by eq. 27: for each voxel i (with coordinates (x,y,z), say), and for each label k (in equation 27 they use \ell instead of k, but I will use k because it reads better in plain text) you need to compute $P^{(t)}(k | y_{i})$, let's denote that P[i, k].

Now in order to compute P[i,k] you will need another quantity first, which is the second factor in the numerator of eq. 27, wich is $P^{t}(k, x_{N_{i}})$, let's denote this quantity Q[i, k], this quantity is the probability that the label of voxel i is $k$ given that its neighbors are known (remember that you already know all the labels because that's the output of ICM). You can actually find how to compute Q[i,k] in equation 2.18 of Stan Z. Li book "Markov Random Field Modeling in Image Analysis" (page 27) . In that equation V1 doesn't concern to us and V2 is, in our case, the potential of the Ising/Potts model (exponential of -beta or +beta, depending on the neighboring voxels having the same or different label, respectively). The following pseudo code may be more clear:

The following computes Q[i, k] for only one voxel i, and for all labels k
```
Initialize Q[i, k] = 0 for all labels k
normalization = 0
for each label k:
    for each neighbor j of voxel i:
        Q[i, k] += PottsIssing(k, x_{j}) 
    Q[i,k] = exp(Q[i,k])
    normalization += Q[i,k]
Q[i,k]/=normalization
```
here, PottsIssing(a, b)= -beta if a==b, else beta, and x_{j} is the label of voxel j (remember, the output from ICM). From these Q, you can compute P[i,k] using eq. 27 and then use that to compute the new mus and sigmas from eq. 25 and 26.

In summary, you initialize the mu and sigma for all models using median otsu, and then iterate many times the following 2 steps(we can discuss when to stop later):
1. ICM to find the labels using your current mus and sigmas
2. Update mu and sigma according to eq. 25 and 25, using the current labels

I hope it helps

dipy/segment/energy_mrf.py
omarocegueda: Sure!, it's 6th line, first paragraph $V_{C}(x) = -\delta(x_{i}-x_{j})$ (the Dirac delta)

omarocegueda: Sorry, sixt line, first paragraph, page 49

villalonreina: Yes, you are right. I completely missed that one. Thanks!

villalonreina: Hi Omar, one question regarding the Delta function. Do you know if  anyone has implemented it in Dipy?

omarocegueda: I don't know, but it would be something like
```
return 1 if x == 0 else 0
```

villalonreina:
Hi Omar, 

I trying to implement equation 27 from the same paper and I am not sure about g^t(y_i) and P(y_i). I am assuming that g^t(y_i) comes from equation 11 and that P(y_i) comes from equation 23. Is that right?

Thanks!

omarocegueda:
Hi Julio,
exactly, $g^{(t)}(y_{i}; \theta_{k})$ is in this case the Normal density (eq. 11) with parameters \theta_{k}, which means the mu and sigma corresponding to class k (denoted \ell in the paper). Regarding $p(y_{i})$, please note that it cannot be given by eq. 23, because eq. 23 is conditioned on the knowledge of $\theta$, but $p(y_{i})$ in eq. 27 is not conditioned on $\theta$. Here, $p(y_{i})$ is the probability that intensity $y_{i}$ is observed at voxel $i$ (without knowing anything about the parameters $\theta$). Also note that in theory you could obtain $p(y_{i})$ by integrating $p(y_{i} | \theta)$ (eq. 23) for all possible values of $theta$ (a huuuge amount of possible values, uncountable many) but, a much simpler way is to observe that $p(y_{i})$ is just a normalization constant that ensures that if you sum all the probabilities $p(k | y_{i})$ (eq. 27) for all the classes k, the sum is equal to 1. You can see this for example in Bishop's Book page 61, eq. 2.76 (which is reference [9] in the paper, and the ebook is here: http://cs.du.edu/~mitchell/mario_books/Neural_Networks_for_Pattern_Recognition_-_Christopher_Bishop.pdf).

So, in summary, you can first ignore $p(y_{i})$ in eq 27, compute only the numerator of eq. 27 for all classes $\ell$ and divide all of them by their sum.

dipy/segment/mrf.pyx
```diff
@@ -0,0 +1,493 @@
...
+        double vox_prob
+        cnp.npy_intp* dX = [-1, 0, 0, 0,  0, 1]
+        cnp.npy_intp* dY = [0, -1, 0, 1,  0, 0]
+        cnp.npy_intp* dZ = [0,  0, 1, 0, -1, 0]
+
+    for x in range(nx):
+        for y in range(ny):
+            for z in range(nz):
+
+                vox_prob = P_L_N[x, y, z, l]
```
omarocegueda: Why not ```vox_prob = 0``` ? starting with ```P_L_N[x, y, z, l]``` is an unnecessary risk (it may not be zero).

dipy/segment/tissue.py
```diff
@@ -0,0 +1,97 @@
...
+        ----------
+        image : 3D ndarray, T1 structural image.
+        nclasses : int, number of desired classes.
+        beta : float, smoothing parameter, the higher this number the 
+                smoother the output will be.
+        max_iter : number of desired iterations. Ususally between 0 and 0.5
+        
+        Returns
+        -------
+        final_segmentation : 3D ndarray, contains all tissue types.
```
Garyfallidis: You actually return 3 things not 2.

dipy/segment/tissue.py
```diff
@@ -0,0 +1,97 @@
...
+        approach for segmentation by using the Iterative Conditional Modes and
+        Expectation Maximization to estimate the parameters.
+        
+        Parameters
+        ----------
+        image : 3D ndarray, T1 structural image.
+        nclasses : int, number of desired classes.
+        beta : float, smoothing parameter, the higher this number the 
+                smoother the output will be.
+        max_iter : number of desired iterations. Ususally between 0 and 0.5
```
Garyfallidis: ``max_iter`` docstring is wrong. Please correct. Say int first and then explain below.

dipy/segment/tissue.py
```diff
@@ -0,0 +1,97 @@
...
+        """
+        This method uses the Maximum a posteriori - Markov Random Field
+        approach for segmentation by using the Iterative Conditional Modes and
+        Expectation Maximization to estimate the parameters.
+        
+        Parameters
+        ----------
+        image : 3D ndarray, T1 structural image.
+        nclasses : int, number of desired classes.
+        beta : float, smoothing parameter, the higher this number the
```
Garyfallidis: just keep float in the first line
and then put the description below. Please correct everywhere!

dipy/segment/tissue.py
```diff
@@ -0,0 +1,97 @@
...
+        self.segmentations = []
+        self.pves = []
+        self.energies = []
+        self.energies_sum = []
+        self.verbose = verbose
+
+        pass
+
+    def classify(self, image, nclasses, beta, max_iter):
+        """
```
Garyfallidis: Start the docstring from the first line.

dipy/segment/mrf.pyx
```diff
@@ -0,0 +1,657 @@
...
+
+        for i in range(0, nclass):
+
+            H = input_image[seg_image == i]
+
+            mu[i] = np.mean(H, -1)
+            std[i] = np.std(H, -1)
+            var[i] = np.var(H, -1)
+
+        return mu, std, var
```
samuelstjean: The variance is pretty much un-needed as a parameter since std = sqrt(var), so you can save one parameter everywhere.

dipy/segment/mrf.pyx
```diff
@@ -0,0 +1,657 @@
...
+
+
+    def negloglikelihood(self, image, mu, sigmasq, nclasses):
```
samuelstjean: it is also much more common to define all mathematicla functions using sigma as an input, instead of using directly sigma**2. It would be less confusing to stick to this convention (or else we could let's say set tau = sigma**2 and forget about the square everywhere).

dipy/segment/tissue.py
villalonreina: Do you mean like this?

image : 3D ndarray, 
             T1 structural image.

Garyfallidis:
```
image : ndarray
    3D structural image
```
It doesn't need to be T1

dipy/segment/tissue.py
samuelstjean: typo in usually also

Garyfallidis:
@omarocegueda and @arokem can you have a look at this PR and merge it? We can add bias field correction at a later point. @villalonreina have a look for feedback for the PR and make the necessary updates.

Garyfallidis:
GGT!!!

arokem:
I'll take a look this weekend

omarocegueda:
Sure! me too =)

dipy/segment/mrf.pyx
```diff
@@ -0,0 +1,657 @@
...
+    cdef double NPY_INFINITY
+    double sqrt(double)
+    double log(double)
+    double exp(double)
+    double fabs(double)
+
+
+class ConstantObservationModel(object):
+    
+    r""" Image segmentation with constant models.
```
omarocegueda: Since this PR defines the base framework for future tissue segmentation algorithms, I think we should include a detailed description of what each class is supposed to do. For example, for this class, we may say something like:
class ConstantObservationModel(object):
r"""Observation model assuming that the intensity of each class is constant.
The model parameters are the means $\mu_{k}$ and variances $\sigma_{k}$ associated 
with each tissue class.
According to this model, the observed intensity at voxel $x$ is given by
$I(x) = \mu_{k} + \eta_{k}$
where $k$ is the tissue class of voxel $x$, and $\eta_{k}$ is a Gaussian random
variable with zero mean and variance $\sigma_{k}^{2}$.
The observation model is responsible of computing the negative
log-likelihood of observing any given intensity $z$ at each voxel 
$x$ assuming the voxel belongs to each class $k$.
It also provides a default parameter initialization.
"""

dipy/segment/mrf.pyx
```diff
@@ -0,0 +1,657 @@
...
+                        neglogl[x, y, z, l] = 1 + log(sqrt(2.0 * NPY_PI * sigmasq[l]))
+                    else:
+                        neglogl[x, y, z, l] = NPY_INFINITY
+
+                else:
+                    neglogl[x, y, z, l] = ((image[x, y, z] - mu[l]) ** 2.0) / (2.0 * sigmasq[l])
+                    neglogl[x, y, z, l] += log(sqrt(2.0 * NPY_PI * sigmasq[l]))
+
+
+cdef void _prob_neighb_perclass(double[:, :, :] image, double[:, :, :] seg,
```
omarocegueda: Parameter `image` is unnecessary: this computes the Ising potential $\beta*Issing(S(x), k)$ with weight $\beta$ between the label $S(x)$ at each voxel $x$, and a fixed label $k$ (named `classid`), so the segmentation `seg` is enough. Probably a better name would be something like `_eval_ising`.

dipy/segment/mrf.pyx
```diff
@@ -0,0 +1,657 @@
...
+
+        nloglike = np.zeros(image.shape + (nclasses,), dtype=np.float64)
+
+        for l in range(nclasses):
+            _negloglikelihood(image, mu, sigmasq, l, nloglike)
+
+        return nloglike
+
+
+    def prob_neighborhood(self, image, seg, beta, nclasses):
```
omarocegueda: Note that `beta` is not a parameter of the model. I think this should be a method of the ICM class instead of the observation model. Also, note that parameter `image` is unnecesary (`seg` is enough for computing these probabilities)

dipy/segment/mrf.pyx
```diff
@@ -0,0 +1,657 @@
...
+                      double[:] mu, double[:] sigmasq, int classid,
+                      double[:, :, :, :] P_L_N,
+                      double[:, :, :, :] P_L_Y) nogil:
+                          
+    r""" Conditional probability of the label given the image
+
+    Parameters
+    -----------
+    image : 3D array, gray-scale T1 image
+    nclasses : int, number of tissue classes
```
omarocegueda: `nclasses` is not a parameter of this method

dipy/segment/mrf.pyx
```diff
@@ -0,0 +1,657 @@
...
+
+                    if seg[xx, yy, zz] == l:
+                        vox_prob -= beta
+                    else:
+                        vox_prob += beta
+
+                P_L_N[x, y, z] = vox_prob
+
+
+cdef void _prob_image(double[:, :, :] image, double[:, :, :] gaussian,
```
omarocegueda: Missing documentation for `gaussian`

dipy/segment/mrf.pyx
villalonreina: Hi Omar, 

I don't understand why beta is not a parameter of the. prob_neighborhood calls the cython function that you mentioned in the comment above and that I have already renamed _prob_class_given_neighb, per your suggestion. This function needs beta as a parameter. Or it maybe has to do with something else?

Also, why do you think this method should be moved to the ICM class?

Thanks!

omarocegueda: Hi Julio =)
I totally agree, `prob_neighborhood` requires `beta`, but `beta` is not a parameter of the observation model. What I mean is that the class `ConstantObservationModel` is only responsible for computing things related to the likelihood (in this particular case, it only requires `mu` and `sigma` for each tissue type). `beta` is a parameter of the prior (in this case, the prior is that the label field is a MRF with Ising potential and parameter `beta`). The problem is that for a given observation model, we may have many different priors, and a prior can be used with many different observation models (e.g. using splines instead of constant models) that's why we should decouple the likelihood and the prior.

ICM is in charge of maximizing the posterior distribution with respect to the segmentation (with a fixed log-likelihood, which was provided by the observation model). Therefore, ICM doesn't need to know any detail about the model, it doesn't care about the mu's and sigma's, it just needs likelihoods, which may have been computed assuming constant models, linear models, splines, etc. 

I hope it helps.

villalonreina: Yes it helped!! Thanks for the explanation.

dipy/segment/mrf.pyx
```diff
@@ -0,0 +1,729 @@
...
+            double[:] mu = np.empty((nclasses,), dtype=np.float64)
+            double[:] sigma = np.empty((nclasses,), dtype=np.float64)
+
+        _initialize_param_uniform(image, mu, sigma)
+        return np.array(mu), np.array(sigma)
+
+
+    def seg_stats(self, input_image, seg_image, nclass):
+        
+        r""" Mean and standard variation for N desired  tissue classes
```
samuelstjean: you probably mean standard deviation

Garyfallidis:
ping @villalonreina :)

dipy/segment/mrf.pyx
```diff
@@ -0,0 +1,729 @@
...
+        -------
+        mu, std: ndarrays,
+                    1 x nclasses dimension
+                    Mean and standard deviation for each class
+
+        """
+        mu = np.zeros(nclass)
+        std = np.zeros(nclass)
+
+        for i in range(0, nclass):
```
omarocegueda: This algorithm is Theta(nclass * N^3). We can compute everything in one pass.

Garyfallidis:
Hey guys and especially @villalonreina and @omarocegueda, any progress with this PR? We need to move on for the next release. The clock is ticking!

Garyfallidis:
@villalonreina any progress with the validation? Keep it up!

dipy/segment/mrf.pyx
villalonreina: Hi Omar,

Thanks for your comment. What exactly do you mean with "Theta(N^3)" ? What specific line is it that you are referring to?

Thanks!

omarocegueda: For each label `k=0, 1, ..., K-1` you are extracting all voxel values in `input_image` which are marked as `k` in `seg_image`. This operation needs to iterate over all voxels, which has complexity `Theta(N^3)` for an image of size `NxNxN`. Since you are doing this once for each label, the full complexity is `Theta(K*N^3)`.

omarocegueda: Instead of doing that you can initialize `mu[...]` and `std[...]` to zero. Then iterate over each voxel `(x, y, z)`, grab its label `s = seg_image[x, y, z]` and its value `v = input_image[x, y, z]`. Then update only the entry in `mu`, `std` corresponding to labe `s`: `mu[s] += v`, `std[s] += v*v`. Since you are sweeping the full volume only once and for each voxel you are updating only one entry of the arrays `mu` and `std`, the full complexity is just `Theta(N^3)`.

omarocegueda: Of course, at the end you have the sum of the values and the sum of their squares (not the mean and std), but from that information and the number of voxels for each label you can directly compute the means and the standard deviations.

dipy/segment/tests/test_mrf.py
```diff
@@ -0,0 +1,371 @@
...
+
+
+def test_greyscale_image():
+
+    com = ConstantObservationModel()
+    icm = IteratedConditionalModes()
+
+    mu, sigma = com.initialize_param_uniform(image, nclasses)
+    sigmasq = sigma ** 2
+    npt.assert_equal(mu, np.array([0., 0.25, 0.5, 0.75]))
```
omarocegueda: In general, we should not compare floating point values for equality, we should use `assert_array_almost_equal` instead.

dipy/segment/tests/test_mrf.py
```diff
@@ -0,0 +1,371 @@
...
+    npt.assert_equal(sigma, np.array([1.0, 1.0, 1.0, 1.0]))
+    npt.assert_equal(sigmasq, np.array([1.0, 1.0, 1.0, 1.0]))
+
+    neglogl = com.negloglikelihood(image, mu, sigmasq, nclasses)
+    npt.assert_equal((neglogl[100, 100, 1, 0] != neglogl[100, 100, 1, 1]),
```
omarocegueda: Sorry, I don't understand what is the purpose of these tests. Why is it important to verify that the negative log-likelihood is different at these voxels?

dipy/segment/tests/test_mrf.py
```diff
@@ -0,0 +1,371 @@
...
+    npt.assert_equal((neglogl[100, 100, 1, 1] != neglogl[100, 100, 1, 3]),
+                     True)
+
+    initial_segmentation = icm.initialize_maximum_likelihood(neglogl)
+    npt.assert_equal(initial_segmentation.max(), nclasses - 1)
+    npt.assert_equal(initial_segmentation.min(), 0)
+
+    PLN = icm.prob_neighborhood(initial_segmentation, beta, nclasses)
+    npt.assert_equal(PLN.all() >= 0.0, True)
+    npt.assert_equal(PLN.all() <= 1.0, True)
```
omarocegueda: For example, these tests really are meaningful! but I think the above are unlikely to catch any bugs.

dipy/segment/tests/test_mrf.py
```diff
@@ -0,0 +1,371 @@
...
+
+    if beta == 0.0:
+        npt.assert_equal((PLN[50, 50, 1, 0] == 0.25), True)
```
omarocegueda: We can put all these coordinates in a list and just iterate over it

doc/examples/tissue_classification.py
```diff
@@ -0,0 +1,157 @@
+# -*- coding: utf-8 -*-
+"""
+Created on Wed Apr 20 15:57:42 2016
+
+@author: jvillalo
```
omarocegueda: We should remove this auto-generated code by spyder

doc/examples/tissue_classification.py
```diff
@@ -0,0 +1,157 @@
...
+approach. Similar algorithms have been proposed by Zhang et al. [Zhang2001]_,
+and Avants et al. [Avants2011]_ available in FAST-FSL and ANTS-Artropos,
+respectively.
+
+Here we will use a T1-weighted image, that has been previously skull-stripped
+and bias field corrected.
+"""
+
+import numpy as np
+#import nibabel as nib
```
omarocegueda: Remove import if unnecessary.

doc/examples/tissue_classification.py
```diff
@@ -0,0 +1,157 @@
...
+First we fetch the T1 volume from the Syn dataset and will determine its shape.
+"""
+
+fetch_tissue_data()
+t1_img = read_tissue_data()
+t1 = t1_img.get_data()
+print('t1.shape (%d, %d, %d)' % t1.shape)
+
+"""
+Once we have fetched the T1 volume, we proceed to denoise it with the NLM algorithm.
```
omarocegueda: Why is this necessary? Note that if we denoise the volume first, then the effect of the regularization parameter will be less noticeable.

dipy/segment/tissue.py
```diff
@@ -0,0 +1,122 @@
...
+        sigmasq = sigma ** 2
+
+        zero = np.zeros_like(image) + 0.001
+        zero_noise = add_noise(zero, 10000, 1, noise_type='gaussian')
+        image_gauss = np.where(image == 0, zero_noise, image)
+
+        final_segmentation = np.empty_like(image)
+        initial_segmentation = seg_init.copy()
+
+        for i in range(100):
```
omarocegueda: We should not hard-code the number of iterations. I think users will expect this to be an input parameter.

dipy/segment/tissue.py
```diff
@@ -0,0 +1,122 @@
...
+            final_segmentation, energy = icm.icm_ising(negll, beta, seg_init)
+            energy_sum.append(energy[energy > -np.inf].sum())
+
+            if self.save_history:
+                self.segmentations.append(final_segmentation)
+                self.pves.append(PVE)
+                self.energies.append(energy)
+                self.energies_sum.append(energy[energy > -np.inf].sum())
+
+            if i % 10 == 0 and i != 0:
```
omarocegueda: Why we only check convergence every 10 iterations?

dipy/segment/tissue.py
```diff
@@ -0,0 +1,122 @@
...
+                                         mu_upd, sigmasq_upd, nclasses)
+
+            final_segmentation, energy = icm.icm_ising(negll, beta, seg_init)
+            energy_sum.append(energy[energy > -np.inf].sum())
+
+            if self.save_history:
+                self.segmentations.append(final_segmentation)
+                self.pves.append(PVE)
+                self.energies.append(energy)
+                self.energies_sum.append(energy[energy > -np.inf].sum())
```
omarocegueda: Sorry, I don't understand what the stop criterion is, could you please comment on this?

dipy/segment/tissue.py
```diff
@@ -0,0 +1,122 @@
...
+        self.save_history = save_history
+        self.segmentations = []
+        self.pves = []
+        self.energies = []
+        self.energies_sum = []
+        self.verbose = verbose
+
+        pass
+
+    def classify(self, image, nclasses, beta, tolerance):
```
omarocegueda: `tolerance` documentation missing in the docstring

dipy/segment/tissue.py
```diff
@@ -0,0 +1,122 @@
...
+        Parameters
+        ----------
+        image : ndarray,
+                3D structural image.
+        nclasses : int,
+                    number of desired classes.
+        beta : float,
+                smoothing parameter, the higher this number the smoother the
+                output will be.
+        max_iter : float,
```
omarocegueda: `max_iter` is currently not an input parameter (I think it should be)

dipy/segment/mrf.pyx
```diff
@@ -0,0 +1,729 @@
...
+            PLN[:, :, :, classid] = np.exp(- PLN[:, :, :, classid])
+            PLN_norm += PLN[:, :, :, classid]
+
+        for l in range(nclasses):
+            PLN[:, :, :, l] = PLN[:, :, :, l] / PLN_norm
+
+        return PLN
+
+
+cdef void _initialize_maximum_likelihood(double[:,:,:,:] nloglike, double[:,:,:] seg) nogil:
```
omarocegueda: `seg` should be of an integral type (int or short), not double

dipy/segment/mrf.pyx
```diff
@@ -0,0 +1,729 @@
...
+                best_class = -1
+                for k in range(nclasses):
+                    if (best_class == -1) or (nloglike[x,y,z,k] < min_energy):
+                        best_class = k
+                        min_energy = nloglike[x,y,z,k]
+                seg[x,y,z] = best_class
+
+
+cdef void _icm_ising(double[:,:,:,:] nloglike, double beta, double[:,:,:] seg,
+                     double[:,:,:] energy, double[:,:,:] new_seg) nogil:
```
omarocegueda: Same here, `seg` and `new_seg` don't need to be double, they should be of an integral type

doc/examples/tissue_classification.py
villalonreina: Ok, yes I agree. Got rid of the denoising of the T1. It makes the example complicated and made the run time longer.

dipy/segment/tissue.py
villalonreina: The converge is tested on the total energy given by ICM at every loop. Every 10 iterations (it could be 20, 25 or 30), a percentage of the total range of change (the tolerance) is compared to the range of change of the last 5 iterations. If this range of change of the last 5 iterations is smaller than the full range of change then the loop stops. Note that the range of change of the last 5 iterations is compared to the total range of change as the loop continues in multiples of 10 (10,20,30, etc....). Shall we change it to be tested every 20 iterations?

dipy/segment/tests/test_mrf.py
villalonreina: Hi Omar. So basically I am trying to test if the neg-loglikelihood is different for that one voxel across the different tissue types (the last axis of the array). Do you think that the neg-loglikelihood will invariably be always different across the tissue types?

Garyfallidis:
Hello @villalonreina. Ready with the rebase and the new PR?

Garyfallidis:
How did the squashing go?

Garyfallidis:
This PR is replaced by https://github.com/nipy/dipy/pull/1115

PR #676, opened 2015-07-12
TST: update Travis config to use container infrastructure.

arokem:
See Travis docs for details:
http://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@
 # It's okay for it to be on multiple physical lines, so long as you remember:
 # - There can't be any leading "-"s - All newlines will be removed, so use
 # ";"s
+sudo: false  # To use travis container infrastructure
 language: python
 env:
     global:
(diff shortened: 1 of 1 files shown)


arokem:
Just thought I would remind the others about this. It's a two-line change that will make our travis runs go faster. Pure win. Please review this.

.travis.yml
```diff
@@ -32,7 +33,6 @@ before_install:
     - source venv/bin/activate
     - python --version # just to check
     - retry pip install nose # always
-    - sudo apt-get install libblas-dev liblapack-dev libatlas3gf-base
```
matthew-brett: I think you need to add something like:

```
cache: pip
addons:
  apt:
    packages:
    - ibblas-dev
    - liblapack-dev
    - libatlas3gf-base
```

https://github.com/scikit-image/scikit-image/pull/1607

arokem: Why? As far as I can tell, everything works just fine without these packages.

matthew-brett: OK - maybe these get installed with the numpy installations.

But - I think you will need some sort of pip cacheing: https://github.com/scikit-image/scikit-image/pull/1610

arokem: Like so?

matthew-brett:
OK - let's give it a shot...

PR #674, opened 2015-07-09
NF: Function to sample perpendicular directions relative to a given vector

RafaelNH:
Hello!

I implemented a function to sample perpendicular directions relative to a given vector. I will need this function for the future steps of the development of the diffusion kurtosis module (work in progress at #664), however since this function is quite general and might be useful for other techniques, I decided to submit it in a different pull request. I am suggesting to have this function in dipy.core.geometry. The function is being tested in the relevant testing file.

For the math details of the function, you can see the following post:

http://gsoc2015dipydki.blogspot.it/2015/07/rnh-post-8-computing-perpendicular.html

PR #677, opened 2015-07-13
DKI PR3 - NF: Adding standard kurtosis statistics on module dki.py 

RafaelNH:
This works follows the work done in PR #664.

Now that the functions to estimate the kurtosis tensor are implemented, we can proceed with the estimate of standard kurtosis statistics (or also known as the standard DKI rotational invariant measures). In this PR, the analytical solutions of the following statistics are implemented:

- Mean kurtosis 
- Radial kurtosis
- Axial kurtosis

For the mean kurtosis analytical solution, the implementation of the Carlson's elliptical integrals was required. Comparative to the code that Prof. Maurizio Marrale kindly supplied, I followed a more recent, general and optimized implementation of the Carlson integrals which is described here: http://arxiv.org/pdf/math/9409227v1.pdf . Despite this, estimation of the integrals are still time consuming, so I am planing to convert functions carlson_rf and carlson_rd in cython.

In alternative to the analytical solution, two numerical methods for a faster estimation of the standard DKI measures are implemented. These numerical methods are expected to be less accurate that the analytical solutions, however they can provide alternatives less computationally demanding and provide a simpler mathematical framework which will be used to further validate the analytical solutions.


RafaelNH:
PS: At the moment this pull request only have the implementations of the Carlson's integrals, however in the following hours I will add the missing code.

RafaelNH:
A final version of the codes for the mean, radial and axial kurtosis were now added! 

These are the planed future steps:

- [ ] write a the numerical method for the radial kurtosis. This will depend on #674. This method is not an exact analytical solution, however it is useful since it does not require rotating the kurtosis tensors to a frame of reference that diagonalizes the diffusion tensor, and perpendicular kurtosis can be computed relative to other vectors besides the diffusion tensor principal axis.

- [ ] replace the Sherbrooke's 3 shells data on example by an HCP-like dataset. Sherbrooke's data seems to have insufficient SNR for fitting the diffusion kurtosis model. For more detail on this see the discussion in the [NeuroImage mailing list](https://mail.python.org/pipermail/neuroimaging/2015-July/000101.html).

- [ ] write carlson_rf and carlson_rd in cython.

Other point to discuss:

- When diffusion parameters are zero, kurtosis values are defined as NaN. As pointed by @arokem, having this NaN can be problematic for image analyzing steps. Should I replace this NaN by zeros?

RafaelNH:
Any comments so far? Ignore the last point of my last github comment. From 399cce7, dki module does not produce NaN anymore.

doc/examples/reconst_dki.py
```diff
@@ -238,7 +238,28 @@
...
 
+ax.flat[0].imshow(MK[:, :, axial_middle], cmap='gray')
+ax.flat[0].set_title('MK')
+ax.flat[1].imshow(RK[:, :, axial_middle], cmap='gray')
+ax.flat[1].set_title('RK')
+ax.flat[2].imshow(AK[:, :, axial_middle], cmap='gray')
+ax.flat[2].set_title('AK')
+
+plt.show()
+fig3.savefig('Kurtosis_tensor_standard_measures.png')
```
arokem: You need to add the `.. figure` directive in the text block below for this figure to be included in the webpage

dipy/reconst/tests/test_dki.py
```diff
@@ -129,3 +140,268 @@ def test_dki_predict():
...
+    dkiF = dkiM.fit(signal_cross)
+
+    # MK analytical solution
+    MK_as = dkiF.mk
+
+    # MK numerical method
+    sph = Sphere(xyz=gtab.bvecs[gtab.bvals > 0])
+    MK_nm = mean_kurtosis(dkiF.model_params, sph)
+
+    assert_array_almost_equal(MK_as, MK_nm, decimal=1)
```
arokem: Why would these two computations be so different (`decimal=1` !)? 

They seem to take the same code path, calling `dipy.reconst.dki.mean_kurtosis` on the same set of parameters here, and in line https://github.com/RafaelNH/dipy/blob/DKI_PR3_kurtosis_statistics/dipy/reconst/dki.py#L1257

dipy/reconst/dki.py
```diff
@@ -28,6 +28,459 @@
...
+    # Convergence has to be done voxel by voxel
+    for v in range(len(x)):
+        n = 0
+        xn = x[v]
+        yn = y[v]
+        zn = z[v]
+        An = (xn + yn + zn) / 3.0
+        Q = (3.*errtol) ** (-1/6.) * np.max([np.abs(An - xn), np.abs(An - yn),
+                                            np.abs(An - zn)])
+        # Convergence condition
```
arokem: Looks to me like you can do all the computations up to this point on the entire vector at once. So no need to have them inside the `for` loop.

dipy/reconst/dki.py
```diff
@@ -28,6 +28,459 @@
...
+            ynroot = np.sqrt(yn)
+            znroot = np.sqrt(zn)
+            lamda = xnroot*(ynroot + znroot) + ynroot*znroot
+            xn = (xn + lamda)*0.250
+            yn = (yn + lamda)*0.250
+            zn = (zn + lamda)*0.250
+            An = (An + lamda)*0.250
+            n = n + 1
+
+        # post convergence calculation
```
arokem: Same goes for all the computations below this point

dipy/reconst/dki.py
```diff
@@ -28,6 +28,459 @@
...
+        n = 0
+        xn = x[v]
+        yn = y[v]
+        zn = z[v]
+        A0 = (xn + yn + 3.*zn) / 5.0
+        An = A0.copy()
+        Q = (errtol/4.) ** (-1/6.) * np.max([np.abs(An - xn), np.abs(An - yn),
+                                             np.abs(An - zn)])
+        sum_term = 0
+        # Convergence condition
```
arokem: I think that the same is true in this function

dipy/reconst/dki.py
```diff
@@ -28,6 +28,459 @@
 from .base import ReconstModel
 
 
+def _positive_evals(L1, L2, L3, er=None):
+    """ Helper function that indentifies which voxels in a array have all
+    eigenvalues larger than zero
+
+    Parameters
+    ----------
+    L1 : ndarray (n,)
```
arokem: I think this would work on ndarrays regardless of their shape. This seems to imply that it would only work with 1D arrays.

dipy/reconst/dki.py
```diff
@@ -28,6 +28,459 @@
...
+        A eigenvalues is classified as larger than zero if it is larger than er
+
+    Returns
+    -------
+    ind : boolean (n,)
+        Array that marks the voxels that have all eigenvalues are larger than
+        zero.
+    """
+    if er is None:
+        er = np.finfo(L1[0]).eps * 1e3
```
arokem: Except for this line. You would have write this as: 

    np.finfo(L1.ravel()[0]).eps * 1e3

arokem: Once you do that. I believe that all the other functions below could also be written to operate on arbitrary-shape ndarrays. In particular, check out `from dipy.core.ndindex import ndindex` to iterate over arbitray-shape arrays

dipy/reconst/tests/test_dki.py
RafaelNH: When a sphere is given as input,  MK is estimated as the average of the directional kurtosis of the sphere's vertices instead of the MK's analytical solution (see documentation dipy.reconst.dki.mean_kurtosis). This is why computations can be so different.

dipy/reconst/dki.py
```diff
@@ -28,6 +28,460 @@
...
+    RF = np.zeros(len(x), dtype=x.dtype)
+
+    xn = x.copy()
+    yn = y.copy()
+    zn = z.copy()
+    An = (xn + yn + zn) / 3.0
+    Q = (3.*errtol) ** (-1/6.) * np.max([np.abs(An - xn), np.abs(An - yn),
+                                         np.abs(An - zn)], axis=0)
+    # Convergence has to be done voxel by voxel
+    for v in range(len(x)):
```
arokem: Could you use ndindex here, so as to make these functions work on arbitrary-dimensional arrays?

dipy/reconst/dki.py
```diff
@@ -28,6 +28,460 @@
...
+    zn = z.copy()
+    A0 = (xn + yn + 3.*zn) / 5.0
+    An = A0.copy()
+    Q = (errtol/4.) ** (-1/6.) * np.max([np.abs(An - xn), np.abs(An - yn),
+                                         np.abs(An - zn)], axis=0)
+    sum_term = np.zeros(len(x), dtype=x.dtype)
+    n = np.zeros(len(x))
+
+    # Convergence has to be done voxel by voxel
+    for v in range(len(x)):
```
arokem: Here as well

arokem:
Have you had a chance to profile this code? Might be worth seeing what the performance bottle-necks are, before considering cythonizing some of it (e.g. the internal loop in the Carlson integrals). Some notes on profiling here: https://github.com/arokem/learning-python/tree/master/profiling

dipy/reconst/dki.py
```diff
@@ -175,6 +638,331 @@ def _directional_kurtosis(dt, MD, kt, V, min_diffusivity=0, min_kurtosis=-1):
...
+    Wyyyy = np.zeros((len(kt)))
+    Wzzzz = np.zeros((len(kt)))
+    Wyyzz = np.zeros((len(kt)))
+
+    for vox in range(len(kt)): 
+        Wyyyy[vox] = Wrotate(kt[vox], evecs[vox], [1, 1, 1, 1])
+        Wzzzz[vox] = Wrotate(kt[vox], evecs[vox], [2, 2, 2, 2])
+        Wyyzz[vox] = Wrotate(kt[vox], evecs[vox], [1, 1, 2, 2])
+
+    # Compute MK
```
arokem: You mean "compute RK?"

dipy/reconst/dki.py
```diff
@@ -668,6 +1579,172 @@ def _wls_iter(design_matrix, inv_design, sig, min_diffusivity):
...
+    \begin{matrix} ( & W_{xxxx} & W_{yyyy} & W_{zzzz} & W_{xxxy} & W_{xxxz}
+                     & ... \\
+                     & W_{xyyy} & W_{yyyz} & W_{xzzz} & W_{yzzz} & W_{xxyy}
+                     & ... \\
+                     & W_{xxzz} & W_{yyzz} & W_{xxyz} & W_{xyyz} & W_{xyzz}
+                     & & )\end{matrix}
+    Returns
+    -------
+    W : array(4,4,4,4)
+        Full 4D kutosis tensor
```
arokem: typo: "kutosis" => "kurtosis"

dipy/reconst/dki.py
```diff
@@ -28,6 +28,460 @@
...
+    sum_term = np.zeros(len(x), dtype=x.dtype)
+    n = np.zeros(len(x))
+
+    # Convergence has to be done voxel by voxel
+    for v in range(len(x)):
+        # Convergence condition
+        while 4.**(-n[v]) * Q[v] > abs(An[v]):          
+            xnroot = np.sqrt(xn[v])
+            ynroot = np.sqrt(yn[v])
+            znroot = np.sqrt(zn[v])
```
arokem: math.sqrt might be faster

dipy/reconst/dki.py
arokem: I believe this should work here (above, line 102):

    np.max(np.abs(An - np.concatenate([xn, yn, zn], axis=-1)), -1)

RafaelNH:
A review of what I have done on the last couple of days:

- According to a suggestion made by @arokem, Carlson integrals are now not limited to 'flat vectors'.

-  @arokem, I also try to replace the np.sqrt inside the Carlson integral by the math.sqrt, however at the end, I decided to keep np.sqrt because it automatically deals better with complex numbers. I could had use cmath instead, however when working with real numbers the output variables from cmath will have a 0j imaginary part.

- By profiling the mean_kurtosis function, I figured out that Carlson integrals are not the unique computational demanding steps in MK estimation. Another computational demanding step is the kurtosis rotation done by Wrotate. I tried to use the kurtosis tensor symmetry to improve Wrotate performance by avoiding the computational demanding loops, however after analysis better the code, I noticed that this is not that trivial. The calculations inside the loops also depend on the non-symmetric tensor R. Instead, I factorize Wrotate by removing its dependency to Wcons. In addition, some unnecessary steps on MK and RK functions were removed by calling directly the helpers function _Wrotate_element. Re-processing the data of the [mid-term summary](http://gsoc2015dipydki.blogspot.co.uk/2015/07/rnh-post-6-mid-term-summary.html), procedures are now 1.6 times faster (MK takes around to compute 14 mins, RK 7 mins and AK 1min).

- I add two test to better evaluate the statistical kurtosis measures based on less mathematical demanding approaches. Tests are covering all the lines of dki.py as you can see from nosetest reports.

Please let me know if you have suggestions for the next steps of this PR. Do you think that we could improve performance at this point by cythonizing Wrotate and Carlson integrals? Rather than this, do you have any other comments that I could address?

dipy/reconst/dki.py
```diff
@@ -175,6 +524,360 @@ def _directional_kurtosis(dt, MD, kt, V, min_diffusivity=0, min_kurtosis=-1):
...
+    MK = _F1m(evals[..., 0], evals[..., 1], evals[..., 2])*Wxxxx + \
+         _F1m(evals[..., 1], evals[..., 0], evals[..., 2])*Wyyyy + \
+         _F1m(evals[..., 2], evals[..., 1], evals[..., 0])*Wzzzz + \
+         _F2m(evals[..., 0], evals[..., 1], evals[..., 2])*Wyyzz + \
+         _F2m(evals[..., 1], evals[..., 0], evals[..., 2])*Wxxzz + \
+         _F2m(evals[..., 2], evals[..., 1], evals[..., 0])*Wxxyy
+
+    MK = MK.reshape(outshape)
+
+
```
arokem: Please remove these lines - too much whitespace!

RafaelNH:
I think all the work here is ready to be merged. I already finish all points that were remaining on this PR. In particular:

- I replace the Sherbrooke's 3 shells data on the example by the HCP-like dataset added in @arokem 's PR #691. Considering that we are not using any noise removal approach, kurtosis metrics of the example are now looking nice.

- I further refactored the kurtosis statistics metrics and now procedures take less that 1min to process the data of the [mid-term summary](http://gsoc2015dipydki.blogspot.co.uk/2015/07/rnh-post-6-mid-term-summary.html).  I also tried to accelerate more the codes by cythonizing the Carlson integrals, however, at the end, I decided to keep the codes in python since the gains weren't anymore significant.

Please let me know if you have any other comment to address.

doc/examples/reconst_dki.py
```diff
@@ -3,28 +3,32 @@
...
-the kurtosis tensor [Jensen2005]_. Measurements of non-Gaussian diffusion are
-of interest because they can be used to charaterize tissue microstructural
+The diffusion kurtois model is an expansion of the diffusion tensor model
+(see :ref:`example_reconst_dti`). In addition to the diffusion tensor (DT), the
+diffusion kurtosis model quantifies the degree to which water diffusion in
+biologic tissues is non-Gaussian using the kurtosis tensor (KT) [Jensen2005]_.
+
+Measurements of non-Gaussian diffusion from the diffusion kurtosis model are of
+interest because they can be used to charaterize tissue microstructural
 heterogeneity [Jensen2010]_ and to derive concrete biophysical parameters as
```
arokem: => "...concrete biophysical parameters, such as..."

doc/examples/reconst_dki.py
```diff
@@ -3,28 +3,32 @@
...
+Moreover, DKI can be used to resolve crossing fibers in tractography and to
+obtain invariant rotational measures not limited to well aligned fiber
+populations [NetoHe2015]_.
 
-The diffusion kurtosis model relates the diffusion-weighted signal,
-$S(\mathbf{n}, b)$, to the applied diffusion weighting, $\mathbf{b}$, the
-signal in the absence of diffusion gradient sensitisation, $S_0$, and the
-values of diffusion, $\mathbf{D(n)}$, and diffusion kurtosis, $\mathbf{K(n)}$,
-along the spatial direction $\mathbf{n}$ [NetoHe2015]_:
+The diffusion kurtosis model express the diffusion-weighted signal as:
```
arokem: => "expresses"

doc/examples/reconst_dki.py
```diff
@@ -3,28 +3,32 @@
...
 
 .. math::
+
     S(n,b)=S_{0}e^{-bD(n)+\frac{1}{6}b^{2}D(n)^{2}K(n)}
 
-$\mathbf{D(n)}$ and $\mathbf{K(n)}$ can be computed from the DT and KT using
-the following equations:
+where $\mathbf{b}$ is the applied diffusion weighting (which is dependent on
+the measuremenent parameters measurement), $S_0$ the signal in the absence of
```
arokem: the word "measurement" appears twice here.

doc/examples/reconst_dki.py
```diff
@@ -3,28 +3,32 @@
...
+diffusion gradient sensitisation, $\mathbf{D(n)}$ the value of diffusion along
```
arokem: => "sensitization"

doc/examples/reconst_dki.py
```diff
@@ -48,99 +52,66 @@
...
-
-maskdata, mask = median_otsu(selected_data, 3, 1, True,
-                             vol_idx=range(10, 50), dilate=2)
-
-"""
-Now that we have prepared the datasets we can go forward with the voxel
-reconstruction. This can be done by first instantiate the DiffusinKurtosisModel
-in the following way.
+Now that we have loaded and prepared the datasets we can go forward with the
+voxel reconstruction. This can be done by first instantiate the
```
arokem: => "instantiating"

doc/examples/reconst_dki.py
```diff
@@ -48,99 +52,66 @@
...
+DiffusinKurtosisModel in the following way:
```
arokem: => "DiffusionKurtosisModel"

doc/examples/reconst_dki.py
```diff
@@ -228,16 +186,61 @@
...
-From the figure, we can see that the DT standard diffusion measures from DKI
-are noisier than the DTI measurements. This is a well known pitfall of DKI
-[NetoHe2014]_. Since it involves the estimation of a larger number of
-parameters, DKI is more sensitive to noise than DTI. Nevertheless, DT diffusion
-based measures were shown to have better precision (i.e. less sensitive to
-bias) [Veraa2011]_.
+From the figure, we can see that the standard diffusion measures from DKI has
+similar constrasts than the standard diffusion measures from DTI.
+
+In addition to the standard diffusion statistics, the DiffusinKurtosisFit
```
arokem: => "DiffusionKurtosisFit" (missing the "o"!)

doc/examples/reconst_dki.py
```diff
@@ -228,16 +186,61 @@
...
+"""
+
+MK = dkifit.mk
+AK = dkifit.ak
+RK = dkifit.rk
+
+"""
+Kurtosis measures are susceptible to high amplitude outliers which can corrupt
+the automatic ajusted scale of the matplotlib functions. By assuming that
+typical values of kurtosis lie on a range between 0 and 2, the impact of high
```
arokem: "in the range between..."

doc/examples/reconst_dki.py
```diff
@@ -228,16 +186,61 @@
...
+amplitude kurtosis outliers are removed in the following:
+"""
+
+MK[MK > 2] = 2
+MK[MK < 0] = 0
+AK[AK > 2] = 2
+AK[AK < 0] = 0
+RK[RK > 2] = 2
+RK[RK < 0] = 0
+
```
arokem: Should these be somehow enforced in the module code? Might be overkill

doc/examples/reconst_dki.py
```diff
@@ -228,16 +186,61 @@
...
+ax.flat[1].imshow(AK[:, :, axial_slice], cmap='gray')
+ax.flat[1].set_title('AK')
+ax.flat[2].imshow(RK[:, :, axial_slice], cmap='gray')
+ax.flat[2].set_title('RK')
+
+plt.show()
+fig2.savefig('Kurtosis_tensor_standard_measures.png')
+
+"""
+.. figure:: Kurtosis_standard_measures.png
```
arokem: Whoops. Looks like the file name changed here from just a couple of lines before. This will fail to render!

arokem:
I just ran the example. It is quite fast! One thing that worries me is that the images with the kurtosis measures seem to have a lot of black spaces inside the brain (e.g in the posterior part of the corpus callosum). Do you have a sense for why that is? Are these really places where these measures aren't defined?

doc/examples/reconst_dki.py
RafaelNH: I added this limits as an optional input on functions to compute the standard kurtosis measures. To add this directly on the procedures on DiffusionKurtosisFit instance, mk, ak and rk cannot anymore be defined as @auto_attr. To compute MK, AK and RK from the DiffusionKurtosisFit we have now to type:

MK = DiffusionKurtosisFit.mk()
AK = DiffusionKurtosisFit.ak()
RK = DiffusionKurtosisFit.rk()

Is this ok?

RafaelNH:
Hi @arokem! I think I answered all your comments, please let me know if there something else that I should change.

Regarding to the black spaces inside the brain, these are a consequence of image artefacts. Particularly kurtosis underestimations caused by gibbs ringing and caused by noise in regions where diffusion is low. If you want to know more about gibbs ringing artefacts you can give a look to the following freshly published articles: [Perrone et al., 2015](http://www.ncbi.nlm.nih.gov/pubmed/26142273) and [Veraart et al, 2015](http://www.ncbi.nlm.nih.gov/pubmed/26257388). For example in the first article, you can see that the regions highly corrupted by Gibb artefacts match the black kurtosis regions of our example. For the underestimations of diffusion kurtosis caused by noise in regions where diffusion is low you can see my master thesis [Neto-Henriques et al., 2012](http://repositorio.ul.pt/bitstream/10451/8511/1/ulfc104137_tm_Rafael_Henriques.pdf).

These black regions are common in DKI reconstructions, for example you can see them on the images of the article describing the methods that I implement on this PR (Tabesh et al., 2011)[http://www.ncbi.nlm.nih.gov/pubmed/21337412]. Actually, considering that I practically didn't pre-process the HCP-like data, we can say that our reconstruction are looking very good.

To improve the quality of the reconstruction, we can propose the following pre-processing steps in the example of usage script:

1. A common way to reduce the impacts of gibbs artefacts and MRI noise artefact DKI is to smooth the data using a gaussian kernel of FWHM=2.5 mm before fitting the DKI model. Any one knows if there is already a smoothing procedure already implemented in python? Scipy has scipy.ndimage.filters.gaussian_filter, however this is not compatible with scipy early versions.

2. A more sophisticated pre-processing methods is to denoise data using [non-local mean](http://nipy.org/dipy/examples_built/denoise_nlmeans.html#example-denoise-nlmeans). For this, anyone knows what should be select for N when estimating the standard deviation of the noise, assuming that we are using a multi-band acquisition (full acquisition details can be found [here](https://digital.lib.washington.edu/researchworks/handle/1773/33311)?

samuelstjean:
Well, you can try the nlmeans already in dipy, as gaussian smoothing is more likely to destroy the information you are looking for in my opinion. Secondly, this is where it hurts, unless you know a bit about the reconstruction algorithm that they used, it takes a bit of guesswork... Or let's say I'm still trying to find a sure fire way to know that as people keep asking me that question for the past week and throwing at me tons of weird reconstructions.

Anyway, segmenting the background and computing the magnitude SNR (that is, mean(background)/std(background) once properly segmented) can apparently give you that number according to this nice table from [1]. That being said, I never tried it, but feel free to, if not I can give a look at the dataset tomorrow.

table


[1] Rajan, J., Veraart, J., Van Audekerke, J., Verhoye, M., & Sijbers, J. (2012). Nonlocal maximum likelihood estimation method for denoising multiple-coil magnetic resonance images. Magnetic Resonance Imaging, 30(10), 1512–1518. doi:10.1016/j.mri.2012.04.021

RafaelNH:
Hi @samuelstjean! In short, I computed the SNR of the background voxels using the follow precedure:

- I generated a 'big' mask covering almost all head's voxels: maskdata, mask = median_otsu(data, 4, 3, False, vol_idx=[0, 1], dilate=12) 
- I removed in data all the b-values smaller than 2000 to reduce the intensity of the residual head voxels not covered by the mask.
- SNR was then computed as: SNR = np.mean(data_sel[mask==False, None])/np.std(data_sel[mask==False, None])

This give me a background SNR magnitude of 3.83. From the table of your last comment this corresponds to L=4. I use this L to compute the sigma for each brain volume of the example data:

 sigma = estimate_sigma(data, N=4)

I notice that the computed sigma have a large dependence with the b-value:

- b-value = 0 => sigma around 810
- b-value = 200 => sigma around 510
- b-value = 400 => sigma around 390
- b-value = 1000 => sigma around 268
- b-value = 2000 => sigma around 175

Is this normal? Is the sigma for the low b-values overestimated?

Anyway, after processing the non-local filter I had some nice reconstruction. See the figures [here](http://gsoc2015dipydki.blogspot.co.uk/). It is only unfortunate that denoising the 446 volumes of the example took more than 2 hours :(.

Given the nice results, I added the denoising procedure in the example of usage. Please let me know that do you think!

arokem:
You didn't really mean to close this, right?

RafaelNH:
Hi @arokem! Thanks, I was already trying to figure out how someone can I re-open an accidentally closed pull request =P

arokem:
When that happens there is a button at the bottom saying "re-open pull
request" :-)

RafaelNH:
Well, I think this pull request it not WIP anymore. Does anyone have further comments?

samuelstjean:
Well, regarding your comment, that estimation is purely intensity based, so that's why you see a change change with the b-value. I suggest using piesno instead, which is diffusion based as it estimates only the noise component along the 4D axis, as the original version of this estimator is for T1 data (same for nlmeans), so I advise against using it. Honestly I think we should plaster everywhere to not use it for diffusion data, as everyone seems to do, but anyway... 

As for the runtime, it works seperately on 3D volumes, so the mlore you have the more time it takes. Then again I have been runnign spmething something on a cluster for 3 days on an abusive dataset, so it's not too bad in comparison.

RafaelNH:
Hi @samuelstjean! Actually I tried to use piesno, however this computes a sigma for each slice instead for each volume as required by nlmeans =S. How can we make nlmean campatible with piesno?

samuelstjean:
Yes I know, I actually got you there ;) You can actually print the whole array, and see how does the noise varies. If it's rather a small difference (on our 1.5T we get values from 13 to 15, so I just heuristically take the median), or you can request that feature by complaining over here https://github.com/nipy/dipy/pull/572

Anyway, was the estimation more stable than the other one?

RafaelNH:
Hi @samuelstjean! Piesno is more stable, i.e. for all axial slices sigma is around 160 (median= 153, max=238, min=142). However, I decided to give a close look to the performances of nlmean with piesno and I detected some problems:

- First, piesno only detected a small amount of background voxels.
- Second, nlmean seems to work worst when the same sigma is given on all volumes. This perhaps is a consequence that nlmean was initially not proposed diffusion-weighted images.

For our record, I decided to create a new [post](http://gsoc2015dipydki.blogspot.co.uk/2015/08/rnh-post-12-attempt-to-further-improve.html) with details and figure of these issues.

Since nlmean seems more compatible to the previous nose estimate I decided to keep it for the example, at least for now.

samuelstjean:
The noise estimation was indeed developped in conjunction with nlmeans, so it might be more stable performance-wise because of that. Well, regarding the mask of background identified, it is indeed weird looking and too small for it's own good. Did you run it on the motion corrected, eddy current corrected data? It probably screwed the background if that's the case, if not, well, I'd need to look into that dataset at a later time then to find out what is happening.

RafaelNH:
Oh! True! The data was corrected from motion and eddy current! That should indeed be the reason why piesno is not working properly! Oh well, it seems anyway that nlmeans work better if it has a sigma per volume, so I will keep the noise estimation technique. This show that we need to develop a new denoising procedure for diffusion-weighted data.

@arokem do you have any thoughts on this? I think all the steps on this PR were accomplished. Is this ready to be merged?

doc/examples/reconst_dki.py
```diff
@@ -3,28 +3,32 @@
...
-DKI can be used to resolve crossing fibers in tractography and to obtain
-invariant rotational measures not limited to well aligned fiber populations
-[NetoHe2015]_.
-
-The diffusion kurtosis model relates the diffusion-weighted signal,
-$S(\mathbf{n}, b)$, to the applied diffusion weighting, $\mathbf{b}$, the
-signal in the absence of diffusion gradient sensitisation, $S_0$, and the
-values of diffusion, $\mathbf{D(n)}$, and diffusion kurtosis, $\mathbf{K(n)}$,
-along the spatial direction $\mathbf{n}$ [NetoHe2015]_:
+The diffusion kurtois model is an expansion of the diffusion tensor model
```
arokem: typo: "kurtosis"

doc/examples/reconst_dki.py
```diff
@@ -3,28 +3,32 @@
...
+(see :ref:`example_reconst_dti`). In addition to the diffusion tensor (DT), the
+diffusion kurtosis model quantifies the degree to which water diffusion in
+biologic tissues is non-Gaussian using the kurtosis tensor (KT) [Jensen2005]_.
```
arokem: => "biological"

doc/examples/reconst_dki.py
```diff
@@ -3,28 +3,32 @@
...
+
+Measurements of non-Gaussian diffusion from the diffusion kurtosis model are of
+interest because they can be used to charaterize tissue microstructural
+heterogeneity [Jensen2010]_ and to derive concrete biophysical parameters, such
+as the density of axonal fibres and diffusion tortuosity [Fierem2011]_.
+Moreover, DKI can be used to resolve crossing fibers in tractography and to
+obtain invariant rotational measures not limited to well aligned fiber
```
arokem: => "well-aligned"

doc/examples/reconst_dki.py
```diff
@@ -48,99 +52,99 @@
...
+from dipy.segment.mask import median_otsu
+from dipy.denoise.noise_estimate import estimate_sigma
+from dipy.denoise.nlmeans import nlmeans
 
 """
 DKI requires multi-shell data, i.e. data acquired from more than one non-zero
-b-value. Here, we use fetch to download the raw HARDI of a dMRI dataset
-cotaining three non-zero b-values. The size of the dataset is 188 MBytes,
-however you only need to fetch it once.
+b-value. Here, we use fetch to download a multi-shell dataset which parameters
```
arokem: "which" => "with"

arokem: => "... a multi-shell dataset with parameters that are similar to the data collected in the Human Connectome Project..."

doc/examples/reconst_dki.py
```diff
@@ -48,99 +52,99 @@
...
+are similar to the used on the Human Connectome Project (HCP). The total size
+of the dowloaded data is 1760 MBytes, however you only need to fetch it once.
```
arokem: typo: "downloaded"

doc/examples/reconst_dki.py
```diff
@@ -48,99 +52,99 @@
...
 
-from dipy.data import fetch_sherbrooke_3shell
-
-fetch_sherbrooke_3shell()
+fetch_cenir_multib(with_raw=False)
 
 """
-Next, we read the saved dataset
+Next, we read the saved dataset. To decrease the influence of diffusion signal
+taylor approximation componets larger than the fourth order (componets not
```
arokem: typo: "components"

doc/examples/reconst_dki.py
```diff
@@ -48,99 +52,99 @@
...
-GradientTable object with information about the b-values and b-vectors. The
-b-values used on the loaded dataset are visualized below
-"""
+Function ``read_cenir_multib`` return img and gtab which contains respectively
+a nibabel Nifti1Image object (where the data can be extracted) and a
+GradientTable object with information about the b-values and b-vectors.
 
-import matplotlib.pyplot as plt
+Before fitting the data, we preform some data pre-processing. We first create a
+preliminary mask and avoid pre-processing unnecessary voxels:
```
arokem: => "... preliminary mask to avoid preprocessing... "

doc/examples/reconst_dki.py
```diff
@@ -48,99 +52,99 @@
...
-From the figure above, we can check that the loaded dataset contains three
-non-zero b-values as required for DKI. However the highest b-value of 3500
-$s.mm^{-2}$ is higher than normally used on DKI. Since DKI neglects diffusion
-signal components higher than the 4th order KT, a upper bound of b-value < 3000
-$s.mm^{-2}$ is normally implied to decrease the influence of higher order terms
-[Jensen2010]_. Following this, we discard the b-value shell of 3500 $s.mm^{-2}$
-before DKI fitting.
+Since the diffusion kurtosis models involves the estimation of a large number
+of parameters [TaxCMW2015]_ and since the non-Gaussian components of the
+diffusion signal are more sensitivity to artefacts [NetoHe2012]_, a fundamental
```
arokem: => "... more sensitive..."

doc/examples/reconst_dki.py
```diff
@@ -48,99 +52,99 @@
...
+data pre-processing step for diffusion kurtosis fitting is to denoise our data.
+For this, we use Dipy's non-local mean filter (see
+:ref:`example-denoise-nlmeans`). Note that, since the HCP-like has a large
```
arokem: => "...since the HCP-like data..."

doc/examples/reconst_dki.py
```diff
@@ -48,99 +52,99 @@
...
+number of diffusion-weigthed volume, this procedure can take some minutes to
```
arokem: "volume" => "volumes"

arokem: "some minutes"? Is "some" more like 10 minutes or more like 100 minutes?

RafaelNH: A couple of hours =S

doc/examples/reconst_dki.py
```diff
@@ -48,99 +52,99 @@
...
+voxel reconstruction. This can be done by first instantiating the
+DiffusionKurtosisModel in the following way:
 """
 
 dkimodel = dki.DiffusionKurtosisModel(gtab)
 
 """
-Fitting the data is very simple. We just need to call the fit method of the
-DiffusinKurtosisModel in the following way:
+To fitting the data using the defined ``dkimodel``, we just need to call the
```
arokem: => "To fit"

arokem: I would drop the word "just".

arokem: => "To fit the data using the defined model object, we call the fit function of this object:"

doc/examples/reconst_dki.py
```diff
@@ -148,9 +152,11 @@
 """
 The fit method creates a DiffusionKurtosisFit object which contains all the
 diffusion and kurtosis fitting parameters and other DKI attributes. For
-instance, all diffusion standard tensor statistics can be computed from the
-DiffusionKurtosisFit instance as the fractional anisotropy (FA), the mean
-diffusivity (MD), the axial diffusivity (AD) and the radial diffusivity (RD).
+instance, since the diffusion kurtosis model estimates the diffusion tensor,
+all diffusion standard tensor statistics can be computed from the
+DiffusionKurtosisFit instance. As example, we show below how to extract the
```
arokem: => "For example, we show below..."

doc/examples/reconst_dki.py
```diff
@@ -161,26 +167,11 @@
 """
 Note that these four standard measures could also be computed from Dipy's DTI
 module. For comparison purposes, we calculate below the FA, MD, AD, and RD
-using Dipy's DTI module. For this, we first extract the data of the lower
-b-value to decrease the effects of the signal components lager than the 2nd
-order which are not taken into account by DTI.
+using Dipy's TensorModel.
```
arokem: Worth saying that the values of these parameters are different in DKI and DTI (and why)

doc/examples/reconst_dki.py
```diff
@@ -189,38 +180,38 @@
 
 """
 The DT based measured obtain from DKI and DTI can be easly visualized using
-matplotlib. For example, we show below the middle axial slices of FA, MD, AD,
-and RD obtain from the DKI model (upper panels) and the DTI model (lower
-panels).
+matplotlib. For example, the FA, MD, AD, and RD obtain from the DKI model
+(upper panels) and the DTI model (lower panels) are ploted for an arbitary
```
arokem: => "are plotted for an axial slice."

"Plotted" has two t's and also no need for the "arbitrary selected"

doc/examples/reconst_dki.py
```diff
@@ -228,19 +219,59 @@
...
    DKI (upper panels) and DTI (lower panels).**.
 
-From the figure, we can see that the DT standard diffusion measures from DKI
-are noisier than the DTI measurements. This is a well known pitfall of DKI
-[NetoHe2014]_. Since it involves the estimation of a larger number of
-parameters, DKI is more sensitive to noise than DTI. Nevertheless, DT diffusion
-based measures were shown to have better precision (i.e. less sensitive to
-bias) [Veraa2011]_.
+From the figure, we can see that the standard diffusion measures from DKI has
+similar constrasts than the standard diffusion measures from DTI.
```
arokem: => "...has similar contrasts to the standard diffusion measures from DTI."

doc/examples/reconst_dki.py
```diff
@@ -228,19 +219,59 @@
...
+
+In addition to the standard diffusion statistics, the DiffusionKurtosisFit
+instance can be used to estimate standard non-Gaussian measures as the mean
```
arokem: => "...can be used to estimate the non-Gaussian measures of mean kurtosis (MK), axial kurtosis (AK) and radial kurtosis (RK)"

doc/examples/reconst_dki.py
```diff
@@ -228,19 +219,59 @@
...
+
+MK = dkifit.mk(0, 3)
+AK = dkifit.ak(0, 3)
+RK = dkifit.rk(0, 3)
+
+"""
+Kurtosis measures are susceptible to high amplitude outliers. The impact of
+high amplitude kurtosis outliers were removed on the above lines of codes by
+introducing as an optional input the extremes of the typical values of kurtosis
+(assumed here as the values on the range between 0 and 3)
```
arokem: Yes - this works for me.

dipy/reconst/dki.py
```diff
@@ -19,13 +19,352 @@
...
+        F1[cond1] = ((L1+L2+L3) ** 2) / (18 * (L1-L2) * (L1-L3)) * \
+                    (np.sqrt(L2*L3) / L1 * RFm +
+                     (3 * L1**2 - L1*L2 - L1*L3 - L2*L3) /
+                     (3 * L1 * np.sqrt(L2*L3)) * RDm - 1)
+
+    # Resolve possible sigularity a==b
+    cond2 = np.logical_and(cond0, np.logical_and(abs(a - b) < er,
+                                                 abs(b - c) > er))
+    if np.sum(cond2) != 0:
+        L1 = a[cond2]
```
arokem: These lines are not currently covered by tests (see here: https://travis-ci.org/nipy/dipy/jobs/75456098)

dipy/reconst/dki.py
```diff
@@ -19,13 +19,352 @@
...
+        L3 = c[cond2]
+        F1[cond2] = _F2m(L3, L1, L1) / 2
+
+    # Resolve possible sigularity a==c
+    cond3 = np.logical_and(cond0, np.logical_and(abs(a - c) < er,
+                                                 abs(a - b) > er))
+    if np.sum(cond3) != 0:
+        L1 = a[cond3]
```
arokem: Same here.

arokem:
This is very close, I think. Please look at the coverage report of the tests (https://travis-ci.org/nipy/dipy/jobs/75456098), and see whether the tests can be written to cover more of the code. In particular, I noticed a few special cases that are not being tested at all.

arokem:
But yes - I think that the example is fine as-is, apart from the few typos I mentioned. We can use nlmeans the way you are using it. No need to go full bore on PIESNO here.

doc/examples/reconst_dki.py
arokem: How about replacing "some minutes" with "a couple of hours"? It's fine that
it takes time, considering how nice the results are following this.

samuelstjean:
I though about something, the acquisition you link to is seperated into smaller nifti files for each b-value. Do you know if the full acquisition was made in one block or the dicom themselves contains seperate sequence for each bvalue? If that's the case, your scanner might have rescaled all your data, making it impossible to compare the rescaled intensity between various acquisitions. Our 3T does that and it's a bit of a pain since we have to manually rescale everything back to comparable intensities...That could explain the weird behavior of nlmeans.

arokem:
Hey @samuelstjean - could you be more specific? What "weird behavior" are you talking about?

RafaelNH:
Hi @samuelstjean! I didn't acquired this data. This was supplied by [Romain Valabregue](https://digital.lib.washington.edu/researchworks/handle/1773/33311). Nevertheless, I will bet that this acquisition was made in one block. If the intensities were re-scaled, we wouldn't never have obtained the nice diffusion kurtosis reconstructions shown in the example.

RafaelNH:
Hi @arokem! I think I answered all your comments! Please check if you are happy with my description of the differences between the diffusion measures from DKI and from DTI (lines 169 of the example usage script). 

Thanks for the comment regarding the test coverage. I don't know why when I type "nosetests -v test_dki.py --with-coverage --cover-package=dipy" the percentages of coverage reported on my local pc are different from the ones reported on Travis [https://travis-ci.org/nipy/dipy/jobs/75456098](https://travis-ci.org/nipy/dipy/jobs/75456098). Anyway, looking to the lines of code not covered with my tests, I detected some bugs :). Also, I discovered that the singularities on the MK analytical solution were still affecting values out of the regions that I was defining as problematic, i.e. classifying two different eigenvalues as different based on three order of magnitude larger than the system epslon was not enough. Now, for two eigenvalues being classified as different they have to different by 2.5% (this value was adjusted according to the analysis that I reported in another [GSoC post](http://gsoc2015dipydki.blogspot.co.uk/2015/08/rnh-post-13-start-wrapping-up-test.html)).

Any other comments, before merging this? Or are we ready to move on to PR #685?

samuelstjean:
I was speaking about the variable noise estimation for eahc b-values. Turns out our scanner maximize the intensity range (between 0 and 2888 if memory serves) for each acquisition separately. So if you acquire let's say in the same session 1 b0 +  40 directions, then a reversed phase encode b0 right afterward, the scaling of the intensity is screwed up between these two different sequence because they were made separately. 

It's true it would break everything modeling wise so you are probably fine anyway.

arokem:
Nice work on hunting down those bugs! I also sometimes see discrepancies between the Travis coverage report and local runs. I am not sure how to explain that - anyone else see this/know what this means?

arokem:
Concerning the example: I think that it's very close, but let's spend a couple of minutes on that when we chat tomorrow over skype. Other than that, this is ready to go from my point of view. Anyone else have any comments here? If not, I will merge this at the end of this week.

arokem:
Yes  - ready on my side of things. Unless someone complains, I will merge this on Friday.

RafaelNH:
Is anyone looking to this PR?

arokem:
Is it ready to go on your side? I am ready to push "merge"

RafaelNH:
yup =)

Garyfallidis:
Hi guys. Sorry I haven't got back to this yes. It would be nice if Haz-Edine or Maurizio would give us a thumbs up for the DKI PRs before they get merged.

RafaelNH:
Hi @Garyfallidis! I will send him an e-mail ;)

Garyfallidis:
Hi @RafaelNH yes please let Maurizio know about this. Also e-mail @hassemlal and @ChantalTax if they don't see this.

arokem:
OK - let me know when they expect to be able to take a look. Since this is
holding up other things (e.g. #685), please encourage them to take a look
by the end of next week. If that seems reasonable to everyone, I can merge
it then.

doc/examples/reconst_dki.py
```diff
@@ -48,99 +53,99 @@
...
-The selected b-values and gradient directions are then converted to Dipy's
-GradientTable format.
+To avoid future reprocessing of the non-local mean field, below we save the
+denoised version of the HCP-like data.
 """
 
-from dipy.core.gradients import gradient_table
-
-gtab = gradient_table(selected_bvals, selected_bvecs)
+nib.save(nib.Nifti1Image(den, affine), 'denoised_cenir_multib.nii.gz')
```
ChantalTax: Perhaps we can save this denoised dataset online too so that it can be fetched?

doc/examples/reconst_dki.py
```diff
@@ -160,27 +167,16 @@
...
 Note that these four standard measures could also be computed from Dipy's DTI
-module. For comparison purposes, we calculate below the FA, MD, AD, and RD
-using Dipy's DTI module. For this, we first extract the data of the lower
-b-value to decrease the effects of the signal components lager than the 2nd
-order which are not taken into account by DTI.
+module. Theoretically, computing these measures from both models should be
+analogous. However, since the diffusion kurtosis model involves a larger number
+of parameters, diffusion statistics maps can look more noisy. On the other
+hand, the diffusion statistics from the kurtosis model are expected to have
+better precision [[Veraar2011]]_. For comparison purposes, we calculate below
```
ChantalTax: change precision to accuracy

doc/examples/reconst_dki.py
```diff
@@ -228,19 +223,62 @@
...
-From the figure, we can see that the DT standard diffusion measures from DKI
-are noisier than the DTI measurements. This is a well known pitfall of DKI
-[NetoHe2014]_. Since it involves the estimation of a larger number of
-parameters, DKI is more sensitive to noise than DTI. Nevertheless, DT diffusion
-based measures were shown to have better precision (i.e. less sensitive to
-bias) [Veraa2011]_.
+From the figure, we can see that the standard diffusion measures of the
+HCP-like data obtain from the diffusion kurtosis model have similar contracts
+to the standard diffusion measures obtain from the tensor model. This can be
+explained by the high quality of the dataset and the high preformance of the
```
ChantalTax: check spelling: obtain --> obtained, contracts --> contrasts (?), preformance --> performance

dipy/reconst/dki.py
```diff
@@ -19,13 +19,353 @@
...
+    References
+    ----------
+    .. [1] Tabesh, A., Jensen, J.H., Ardekani, B.A., Helpern, J.A., 2011.
+           Estimation of tensors and tensor-derived measures in diffusional
+           kurtosis imaging. Magn Reson Med. 65(3), 823-836
+    """
+    # Eigenvalues are considered equal if they are not 2.5% different to each
+    # other. This value is ajusted according to the analysis reported in:
+    # http://gsoc2015dipydki.blogspot.co.uk/2015/08/rnh-post-13-start-wrapping-up-test.html
+    er = 2.5e-2
```
ChantalTax: should this be hardcoded or an option?

dipy/reconst/dki.py
```diff
@@ -175,6 +522,382 @@ def _directional_kurtosis(dt, MD, kt, V, min_diffusivity=0, min_kurtosis=-1):
...
 
+def mean_kurtosis(dki_params, min_kurtosis=0, max_kurtosis=3):
+    r""" Computes mean Kurtosis (MK) from the kurtosis tensor.
+
+    Parameters
+    ----------
+    dki_params : ndarray (x, y, z, 27) or (n, 27)
+        All parameters estimated from the diffusion kurtosis model.
+        Parameters are ordered as follow:
+            1) Three diffusion tensor's eingenvalues
```
ChantalTax: check spelling: follow --> follows, eingenvalues --> eigenvalues

dipy/reconst/dki.py
```diff
@@ -175,6 +522,382 @@ def _directional_kurtosis(dt, MD, kt, V, min_diffusivity=0, min_kurtosis=-1):
...
+
+    .. math::
+        G_2(\lambda_1,\lambda_2,\lambda_3)=
+        \frac{(\lambda_1+\lambda_2+\lambda_3)^2}{(\lambda_2-\lambda_3)^2}
+        \left ( \frac{\lambda_2+\lambda_3}{\sqrt{\lambda_2\lambda_3}}-2\right )
+    """
+    # Flat parameters. For numpy versions more recent than 1.6.0, this step
+    # isn't required
+    outshape = dki_params.shape[:-1]
+    dki_params = dki_params.reshape((-1, dki_params.shape[-1]))
```
ChantalTax: could comments like flatten or ravel help here to reshape?

dipy/reconst/dki.py
```diff
@@ -19,13 +19,353 @@
...
-from ..core.gradients import gradient_table
-from ..core.geometry import vector_norm
-from ..core.sphere import Sphere
-from .vec_val_sum import vec_val_vect
 from ..core.onetime import auto_attr
 from .base import ReconstModel
+from dipy.core.ndindex import ndindex
+
+
+def _positive_evals(L1, L2, L3, er=2e-7):
```
ChantalTax: How do we currently handle voxels that don't have this?

doc/examples/reconst_dki.py
```diff
@@ -48,99 +53,99 @@
...
-$s.mm^{-2}$ is normally implied to decrease the influence of higher order terms
-[Jensen2010]_. Following this, we discard the b-value shell of 3500 $s.mm^{-2}$
-before DKI fitting.
+Since the diffusion kurtosis models involves the estimation of a large number
+of parameters [TaxCMW2015]_ and since the non-Gaussian components of the
+diffusion signal are more sensitive to artefacts [NetoHe2012]_, a fundamental
+data pre-processing step for diffusion kurtosis fitting is to denoise our data.
+For this, we use Dipy's non-local mean filter (see
+:ref:`example-denoise-nlmeans`). Note that, since the HCP-like data has a large
+number of diffusion-weigthed volume, this procedure can take a couple of hours
```
arokem: "volume" => "volumes"

doc/examples/reconst_dki.py
```diff
@@ -228,19 +223,62 @@
...
+Therefore, as the figure above shows, white matter kurtosis values are smaller
+along the axial direction of fibers (smaller amplitudes shown in the AK map)
+than for the radial directions (larger amplitudes shown in the RK map).
 
 References:
 
+.. [TaxCMW2015] Tax CMW, Otte WM, Viergever MA, Dijkhuizen RM, Leemans A
+                (2014). REKINDLE: Robust extraction of kurtosis INDices with
+                linear estimation. Magnetic Resonance in Medicine 73(2):
+                794–808.
```
arokem: The dash on this line is a non-ascii character, which raises this error on some machines: 

```
File "reconst_dki.py", line 281
SyntaxError: Non-ASCII character '\xe2' in file reconst_dki.py on line 282, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details
```

arokem: Easy to fix: simply replace the dash ('-') with an ascii dash. This is probably happening because you copied this reference from somewhere.

doc/examples/reconst_dki.py
```diff
@@ -251,10 +289,11 @@
 .. [Fierem2011] Fieremans E, Jensen JH, Helpern JA (2011). White matter
                 characterization with diffusion kurtosis imaging. NeuroImage
                 58: 177-188
-.. [NetoHe2014] Neto Henriques R, Ferreira HA, Correia MM (2012). Diffusion
-                kurtosis imaging of the healthy human brain. Master
-                Dissertation Bachelor and Master Program in Biomedical
-                Engineering and Biophysics, Faculty of Sciences.
+.. [NetoHe2012] Neto Henriques R, Ferreira H, Correia M, (2012). Diffusion
+                kurtosis imaging of the healthy human brain.Master Dissertation
```
arokem: Spaces in this reference are a bit weird:

"...human brain.Master..." => "...human brain. Master..."
"...Bachelor andMaster Programin..." => "Bachelor and Master Program in..."

doc/examples/reconst_dki.py
ChantalTax: FYI: I indeed had this (specific to Windows?), solved it by putting 
```
#!/usr/bin/env python
# -*- coding: utf-8 -*-
```
at the start of the file, but indeed better to prevent it :)

doc/examples/reconst_dki.py
RafaelNH: well spotted!

dipy/reconst/dki.py
RafaelNH: It is important to keep this at 2.5% for a correct identification of the mean kurtosis analytical solution singularities ([see here](http://gsoc2015dipydki.blogspot.co.uk/2015/08/rnh-post-13-start-wrapping-up-test.html)). Therefore, I think we should keep this hardcoded to avoid users setting this parameter in a wrong way.

dipy/reconst/dki.py
RafaelNH: I think I cannot use here flatten or ravel, because these creates 1D arrays. Instead, I want to create a 2D array with dimension of (N, 27) where N is the number of voxels and 27 the number of parameters estimated from DKI.

dipy/reconst/dki.py
RafaelNH: They are not proceeded. On the final maps, these voxels will be zero.

RafaelNH:
Hi @ChantalTax and @arokem! Thanks for your comments!

@ChantalTax - see my responses to your individual comments. Let me know if there is something else that I could address. I think we should keep all command lines for denoising the data on the example so that users are able to follow the same steps to process their data. Therefore, I don't see any advantage of adding the denoised data on fetcher. Also, don't forget that downloading this data can also be time consuming. What do you think?

Does anyone have further comments?

RafaelNH:
Hi Guys! Does anyone have any clue why travis is failing in a test not relevant to the work of this pull request? This is happening since my commit 83e5cbf, where I just changed some documentation according to the comments made by @ChantalTax and @arokem.

doc/examples/reconst_dki.py
```diff
@@ -3,28 +3,33 @@
...
+The diffusion kurtosis model expresses the diffusion-weighted signal as:
 
 .. math::
+
     S(n,b)=S_{0}e^{-bD(n)+\frac{1}{6}b^{2}D(n)^{2}K(n)}
 
-$\mathbf{D(n)}$ and $\mathbf{K(n)}$ can be computed from the DT and KT using
-the following equations:
+where $\mathbf{b}$ is the applied diffusion weighting (which is dependent on
+the measuremenent parameters), $S_0$ is the signal in the absence of diffusion
```
samuelstjean: typo in measuremenent

doc/examples/reconst_dki.py
samuelstjean: Or since it takes hours like you mentionned to use nlmeans, you could cheat and work on a crop of the data and mention that this is done for the sake of speedier runtime in the example (many others are doing that so it can run on modest computers under a minute).

RafaelNH: Hi @samuelstjean! I didn't work on a crop of the data because this will affect the performance of the nlmeans, right? Regarding to @ChantalTax suggestion of saving this dataset online, see the comments done below (particularly the comment done at September 6).

samuelstjean: You can use a mask to define a roi, it should only process those voxels and
give the same thing as the full one.

doc/examples/reconst_dki.py
```diff
@@ -48,99 +53,99 @@
...
+Taylor approximation components larger than the fourth order (componets not
+taken into account by the diffusion kurtosis tensor), we only select the
+b-values up to 2000 $s.mm^{-2}$:
 """
 
-from dipy.data import read_sherbrooke_3shell
+bvals = [200, 400, 1000, 2000]
 
-img, gtab = read_sherbrooke_3shell()
+img, gtab = read_cenir_multib(bvals)
```
arokem: Does this actually work for you? It seems that read_cenir_multib returns (gtab, img) and not the other way around

RafaelNH: Hi @arokem! See my commit d2a5c93 - I changed the order of these outputs to keep consistency with other fetcher's functions!

arokem: yup - that makes sense.

RafaelNH:
Hi guys! I've updated the example to only process the slice of the data that I am showing in Figures. In this way, the example can be processed in few minutes. I think now I only need someone to proofread the changes and then this pull request is ready to be merged.

@arokem, @samuelstjean and @ChantalTax - once more, thanks for your last comments on this =)

arokem:
Looks good to me. I am trying this out on my laptop just to see what it feels like. Would you mind rebasing this in the meanwhile?

doc/examples/reconst_dki.py
```diff
@@ -48,109 +53,105 @@
...
 
-from dipy.data import fetch_sherbrooke_3shell
-
-fetch_sherbrooke_3shell()
+fetch_cenir_multib(with_raw=False)
 
 """
-Next, we read the saved dataset
+Next, we read the saved dataset. To decrease the influence of diffusion signal
+Taylor approximation components larger than the fourth order (componets not
```
arokem: componets => "components"

arokem:
Needs some refactoring! The example fails to run with the recent merge of #572 !

arokem:
Maybe use `np.median(sigma)` as input to `nlmeans` on line 129, instead of `sigma`?

That seems to run for me, and seems to mostly work (with a small region still looking funky right behind the splenium)

samuelstjean:
This seems more like a hack than a solution, especially if people mimic the
example on their data. What are the values you get for the median and sigma?
On Sep 23, 2015 7:54 AM, "Ariel Rokem" <notifications@github.com> wrote:
>
> Maybe use np.median(sigma) as input to nlmeans on line 129, instead of
sigma?
>
> That seems to run for me, and seems to mostly work (with a small region
still looking funky right behind the splenium)

arokem:
In [4]: sigma

Out[4]:

array([ 890.96685791,  883.58361816,  512.60180664,  511.10501099,

        506.72167969,  508.29724121,  511.5748291 ,  517.90643311,

        512.1741333 ,  508.64685059,  813.62762451,  814.27606201,

        515.29516602,  510.68450928,  510.36514282,  510.90167236,

        516.67321777,  515.61724854,  516.25268555,  513.27539062,

        891.284729  ,  895.61871338,  398.3638916 ,  397.18017578,

        390.88934326,  388.23690796,  391.59222412,  396.61325073,

        393.40475464,  392.37744141,  816.565979  ,  815.80596924,

        386.79071045,  388.44381714,  382.40264893,  381.09481812,

        384.32745361,  387.51168823,  385.03262329,  384.90518188,

        888.83172607,  268.06240845,  284.24731445,  272.23126221,

        276.35238647,  277.243927  ,  268.83605957,  277.44604492,

        274.34622192,  270.93826294,  866.05078125,  277.15032959,

        276.66064453,  275.82141113,  268.69241333,  271.10217285,

        283.12982178,  272.10650635,  281.77807617,  270.09680176,

        855.80554199,  271.91842651,  278.293396  ,  272.20910645,

        265.2303772 ,  282.33630371,  270.45040894,  278.87579346,

        269.45068359,  270.91323853,  865.19744873,  274.44061279,

        274.52767944,  270.98492432,  279.68988037,  276.58563232,

        272.43136597,  281.45336914,  265.94506836,  271.76779175,

        855.29882812,  281.78921509,  270.59915161,  269.77212524,

        281.10736084,  274.46472168,  270.11584473,  269.90719604,

        273.0737915 ,  281.24002075,  858.37457275,  270.79598999,

        271.99725342,  280.72039795,  270.5453186 ,  272.62384033,

        274.08502197,  266.06588745,  276.70422363,  270.68930054,

        844.55383301,  272.44577026,  280.96694946,  264.27954102,

        265.31576538,  275.82876587,  278.26522827,  268.74237061,

        268.26953125,  268.42114258,  835.05224609,  274.65313721,

        270.78503418,  265.01184082,  274.83807373,  270.359375  ,

        264.44671631,  266.29559326,  278.22860718,  270.22616577,

        833.87695312,  267.20980835,  266.71990967,  273.24435425,

        272.06921387,  266.18814087,  269.07806396,  274.54159546,

        264.46347046,  268.10791016,  845.78552246,  272.89517212,

        262.00253296,  268.00198364,  265.94869995,  277.72140503,

        272.14599609,  274.54162598,  264.0274353 ,  267.29949951,

        274.20993042,  786.91107178,  257.56256104,  275.73831177,

        262.70623779,  268.63232422,  264.97103882,  255.86329651,

        266.25344849,  262.72424316,  261.70071411,  772.6204834 ,

        265.49087524,  265.243927  ,  265.26681519,  256.51708984,

        259.42251587,  271.79867554,  260.52041626,  270.17303467,

        258.23577881,  770.11480713,  261.16366577,  266.25454712,

        261.67858887,  252.88453674,  271.20910645,  260.26428223,

        267.06970215,  257.51400757,  258.79879761,  768.04400635,

        264.56933594,  263.32653809,  259.08139038,  267.67480469,

        264.36352539,  260.95663452,  268.30566406,  251.98558044,

        260.03625488,  770.39385986,  269.22207642,  256.72720337,

        259.23190308,  267.58883667,  262.03875732,  258.44668579,

        262.96740723,  262.68939209,  272.87722778,  769.23291016,

        261.1882019 ,  259.87600708,  268.95739746,  258.76846313,

        261.10028076,  262.48049927,  254.86764526,  268.29708862,

        261.1355896 ,  767.07537842,  264.54293823,  270.90164185,

        254.70582581,  255.64750671,  267.80023193,  269.90505981,

        262.86224365,  259.515625  ,  260.11138916,  762.15612793,

        265.77035522,  261.79946899,  254.49595642,  264.92004395,

        261.01824951,  254.51960754,  259.32537842,  271.86608887,

        260.81210327,  764.43103027,  257.65435791,  258.74963379,

        265.04876709,  262.03295898,  255.75682068,  260.91726685,

        266.13928223,  255.80784607,  260.68392944,  758.90142822,

        266.36520386,  251.71130371,  258.87585449,  258.67181396,

        269.10183716,  264.77249146,  266.37188721,  255.88493347,

        257.28997803,  265.44546509,  919.86621094,  183.22773743,

        178.03735352,  180.33892822,  179.1151886 ,  175.95526123,

        180.82540894,  184.05723572,  178.37634277,  179.46505737,

        880.17907715,  175.72109985,  182.11384583,  181.31365967,

        181.98284912,  176.06207275,  177.85195923,  173.2076416 ,

        177.2711792 ,  179.06376648,  876.41851807,  177.17530823,

        177.21179199,  176.04786682,  178.26031494,  181.49101257,

        175.25514221,  174.56678772,  180.9135437 ,  174.57936096,

        858.88873291,  176.09593201,  180.30599976,  177.4629364 ,

        174.72229004,  177.33384705,  175.9324646 ,  180.6361084 ,

        177.18981934,  173.59675598,  857.52050781,  181.50959778,

        178.7334137 ,  178.98852539,  172.03619385,  176.62878418,

        180.4098053 ,  178.82092285,  177.61653137,  181.44636536,

        848.56488037,  176.0382843 ,  177.10945129,  179.38719177,

        175.24398804,  173.57899475,  178.62730408,  177.11170959,

        178.2230835 ,  177.70310974,  876.54449463,  181.37193298,

        177.45225525,  176.89828491,  176.86459351,  178.05175781,

        176.94807434,  176.94277954,  175.48188782,  179.61872864,

        866.47192383,  179.86961365,  180.7094574 ,  182.35974121,

        177.52568054,  172.36753845,  182.74502563,  178.39541626,

        175.30110168,  180.68534851,  863.31604004,  179.06196594,

        178.08456421,  176.57325745,  177.28886414,  178.72598267,

        177.12129211,  180.18859863,  177.83532715,  173.91107178,

        872.18035889,  177.53970337,  179.57929993,  181.07929993,

        177.07156372,  181.75863647,  173.99113464,  175.78858948,

        172.06248474,  179.16290283,  877.10949707,  178.80479431,

        782.18164062,  177.10420227,  171.69564819,  174.7164917 ,

        172.71487427,  168.73234558,  176.91641235,  179.15422058,

        172.43757629,  172.68095398,  780.15618896,  169.84925842,

        175.63484192,  176.16711426,  178.84980774,  172.38104248,

        172.93711853,  168.17425537,  173.76182556,  176.22735596,

        776.78814697,  173.01530457,  173.0163269 ,  173.12861633,

        176.30619812,  177.65357971,  171.15559387,  171.96783447,

        177.81140137,  169.79882812,  770.70709229,  173.51457214,

        176.96322632,  173.50007629,  168.70559692,  170.67762756,

        171.66473389,  177.28039551,  172.48381042,  169.59623718,

        774.39953613,  176.3037262 ,  174.73483276,  174.09040833,

        166.2749176 ,  172.91542053,  175.53582764,  174.44374084,

        171.74797058,  177.1443634 ,  779.63415527,  172.22007751,

        172.01646423,  175.13975525,  171.35784912,  168.03686523,

        173.76353455,  169.55761719,  173.61846924,  171.72224426,

        773.22338867,  175.98873901,  171.72529602,  173.55076599,

        167.77912903,  171.06825256,  169.42480469,  171.09202576,

        169.44259644,  171.62393188,  768.55664062,  175.44338989,

        172.66252136,  175.45999146,  171.20602417,  164.49226379,

        175.10665894,  170.27085876,  169.35180664,  174.83590698,

        778.43560791,  172.49058533,  171.3049469 ,  167.50914001,

        169.47991943,  169.73129272,  168.32281494,  172.52311707,

        170.41288757,  166.21620178,  771.64343262,  170.00733948,

        171.95800781,  174.77693176,  171.05247498,  175.20524597,

        166.69021606,  168.86456299,  164.33203125,  173.30807495,

        772.42175293,  172.09806824], dtype=float32)

arokem:
Maybe you can also explain why we have 446 values here? I thought that
`estimate_sigma` should produce one value per slice. It looks like it's
producing one value per volume...

arokem:
I guess the 'solution' offered in the relevant example (
http://nipy.org/dipy/examples_built/denoise_nlmeans.html#example-denoise-nlmeans)
is to denoise volume by volume?

samuelstjean:
Well, piesno works slice by slice, and nlmeans and estimate sigma work
volume per volume,  so thats why you have that many estimates.

arokem:
Yeah - got those confused, I guess. What do you think about denoising
volume by volume?

arokem:
Also - this seems to expose a change in the API related to the recent
merge (or am I misunderstanding?). Maybe we should make a note of that
somewhere?

samuelstjean:
Thats why noiw you can supply a local estimation per volume, the doc should
say that now. I guess for dmri or fmri 4d aware method should at least
perform equal or better than their 3d application on each volume.

arokem:
How do you get the *local* estimation per volume? I don't think you can get
that with `estimate_sigma` - right now that gives you a single estimate per
volume.

samuelstjean:
That's right, it's not designed for that. Piesno is a kind of hybrid with
strong assumptions for slicewise estimation and I have a bunch of stuff
lying around for 3d local estimation. At least with the recent
modification, people can come up with any estimation method they prefer and
use it in the exact same way on nlmeans.

arokem:
OK @samuelstjean - I need a little bit more clarification here: what
exactly is the code that we are supposed to run to do nlmeans denoising for
DWI data? The example (
http://nipy.org/dipy/examples_built/denoise_nlmeans.html) states clearly
that "you can denoise 3D or 4D images", but only demonstrates how to do
this with 3D data (one single volume). What *exactly* is this 4D-aware
method you wrote about in your recent response? Is this something that
exists in theory (it it's "lying around somewhere" it may as well be), or
something that is in our code-base. Where is there an example of it, that
we can use to learn how to do this? If the answer is not to do this volume
by volume (you haven't really answered that question!), then this denoising
example needs to be rewritten to include an example of denoising of 4D data
(you seem to suggest, but are not clearly saying, that there is some way to
do that - please clarify). In the lack of a more clear explanation what we
are supposed to do here, I think that we are going to go with denoising
volume by volume here, since at least you didn't really object to that.

samuelstjean:
And that's probably why doing technical stuff on a tablet is a bad idea.

So anyway, what I meant to say with a few words is that nlmeans is designed for 3D denoising, and will ever only work on each volume separately because it's not designed to take into account the 4D awareness of dMRI data, so the example is correct like that. 

The recent thing I added is that instead of supplying a single value, you can estimate a varying noise pattern if you have the necessary function. So PIESNO can do that, and I have also 2-3 functions in my own dev branch to do that. This is most useful in the case of a sense reconstruction, as the reconstructed image exhibits this kind of behavior much more than in the case of a GRAPPA reconstruction (your friendly MR physicist can tell you more about that and what your scanner actually does, depending on if it rescales the signal or suppress the background, different pattern emerges in your dataset. And also if multiband imaging becomes more and more popular, you have a lot of signal leakage and artefacts in the image itself which needs to be taken into account).

And lastly, I also have in that same branch a denoising algorithm designed for 4D dMRI datasets explicit, which considers the whole volumetric data as opposed to what is currently available. So for the time being you can only go with nlmeans and do 3D volume per volume denoising at it's the only available option, but I'm sure we can do better at a later time though.

RafaelNH:
@arokem - I already rebase this PR, and yup the new version of function nlmean cannot accept anymore an array of sigmas for each diffusion-weighted volume!

@samuelstjean - I give a look on the code of the functions nlmean. So basically the new version of nlmeans can accept two types of sigma: 1) single value; 2) sigma profile. Regardless the type of sigma, nlmean will act the same according to the format of the data. For 3D data nlmean calls nlmean_3D once, while for 4D data nlmean calls nlmean_3D for each volume.

For my understanding there isn't no function in dipy that generates this sigma profiles, right? Even piesno will not be compatible with the new version of nlmean since this produces an array of sigmas per slice and not an 3D noise profile.

samuelstjean:
That's correct, so far no publicly, easily available function will do 3D noise estimation. You can still use piesno and broadcast the 2D result to a 3D array though. The default behavior should still be (and should also always be) to call the nlmeans3D version on each volume separately. 

According to your first comment, does something not work as expected with the new format of the nlmeans function?

RafaelNH:
Yes. Previously if an array of sigmas per volume was given (output of estimate_sigma), nlmean was able to process each volume with the respective sigma on the array.

Indeed, we still can use piesno and estimate_sigma by converting their outputs to 3D arrays. My only concern is if these steps are also trivial for new Dipy users or users that are new to python. I think we should make nlmean compatible with both piesno and estimate_sigma outputs without changing the new version of nlmeans3D.

samuelstjean:
That's weird, it should work as-is with the output of estimate_sigma if you call nlmeans directly (or I copy pasted the wrong function, but the tests are passing...), as it wraps nlmeans_3D in the proper way if I did not make any mistake. Afaik, nlmeans3D should not be really called directly unless you know how it works, as the function simply named nlmeans should set everything for you.

arokem:
To be concrete, the test implemented in #721 should pass?

arokem:
I also propose a BF in #721 - should be a quick fix that will make this merge *finally* possible

RafaelNH:
Hi @arokem! Yup, #721 seems to do the magic!

PR #678, opened 2015-07-14
BF: added tolerance for negative streamline coordinates checks

jchoude:
In some case, tracts can be generated exactly on the border of the image volume. When mapped back to voxel coordinates, due to numerical imprecisions, they may have a coordinate with a value like -5.96046447754e-08, which trips up the negative coordinates test in _to_voxel_coordinates.

I added a tolerance to avoid rejecting valid points. Using a very small threshold so that genuinely invalid points still end up raising the error.

### DIFF
--- a/dipy/tracking/_utils.py
+++ b/dipy/tracking/_utils.py
@@ -60,7 +60,7 @@ def _to_voxel_coordinates(streamline, lin_T, offset):
     raises an error for negative voxel values."""
     inds = np.dot(streamline, lin_T)
     inds += offset
-    if inds.min() < 0:
+    if inds.min().round(decimals=6) < 0:
         raise IndexError('streamline has points that map to negative voxel'
                          ' indices')
     return inds.astype(int)


arokem:
Would it be possible to provide a test that fails without this fix?

jchoude:
Of course! Just to make sure, you want a dataset that fails without this, or a new test in the testing framework?

If it's the latter, where should I put it: as a new test function in dipy.tracking.tests.test_utils.py, or as an additional test in, say, test_density_map of the same file?

arokem:
The latter! 

As for your question, that depends: ideally as many atomic tests as possible. On the other hand, if your test would need to use variables that are already defined in this test, it might be more expedient to just add this as a test-case in there.

dipy/tracking/_utils.py
```diff
@@ -60,7 +60,7 @@ def _to_voxel_coordinates(streamline, lin_T, offset):
     raises an error for negative voxel values."""
     inds = np.dot(streamline, lin_T)
     inds += offset
-    if inds.min() < 0:
+    if inds.min().round(decimals=6) < 0:
         raise IndexError('streamline has points that map to negative voxel'
```
Garyfallidis: Start with a capital letter please.

Garyfallidis:
@jchoude add a quick testing grabbing the IndexError and will merge this asap. Thank you very much! :dancer:

arokem:
@jchoude - any chance to revive this PR? A small test, pretty please?

jchoude:
@arokem I haven't forgotten, just got swamped as always. Will try to do in the next few days.

Garyfallidis:
Ping!

dipy/tracking/_utils.py
jchoude: @Garyfallidis just to make sure, you mean to start the error message string? I'll fix it, was already like this before.

Garyfallidis: Ignore this. Not important.

arokem: Oh yeah - this PR is also crucial for the release. We need a test!

Garyfallidis:
Okay @jchoude because this is an important fix. I will go ahead and merge this PR. Can you write a test as @arokem suggests in a new PR?

arokem:
I didn't mean that you should merge this without a test!

Now we're almost guaranteed that this will remain untested - sure way of
getting the bug back in...

jchoude:
Damn... I was just finishing the test :( I have it ready here... Will do the new PR then.

Garyfallidis:
:dancers: now I got your attention :dancers: :8ball:

Garyfallidis:
:)

jchoude:
you already had my attention... we're in the same lab, you could have pigned me.

Garyfallidis:
Yeah, sorry. No worries. :+1:

PR #680, opened 2015-07-15
NF: Fetch and Read the MNI T1 and/or T2 template.

arokem:
While I was there, I did a bit of PEP8 cleaning and functionalized some repeated boiler-plate.

### DIFF
--- a/dipy/data/fetcher.py
+++ b/dipy/data/fetcher.py
@@ -77,6 +77,16 @@ or updating to the newest version of dipy.""" % (fullpath, md5)
         _log("Files successfully downloaded to %s" % (folder))
 
 
+def _already_there_msg(folder):
+    """
+    Prints a message indicating that a certain data-set is already in place
+    """
+    msg = 'Dataset is already in place. If you want to fetch it again '
+    msg += 'please first remove the folder %s ' % folder
+    print(msg)
+
+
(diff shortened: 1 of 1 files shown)


Garyfallidis:
Nice job!

Garyfallidis:
@villalonreina you can now update this file to also download the tissue maps from MNI. 
Also @arokem do we need to add some license info for these?

arokem:
License info is included! But let me know if this seems insufficient, and
we'll fix:

https://github.com/nipy/dipy/blob/master/dipy/data/fetcher.py#L567
https://github.com/nipy/dipy/blob/master/dipy/data/fetcher.py#L586

Garyfallidis:
We need to add a printout to say that by downloading this dataset you agree to this license found there... or something.

arokem:
Note that I have mirrored all the files here:

https://digital.lib.washington.edu/researchworks/bitstream/handle/1773/33312/

so that we can grab individual files, rather than a big archive with all of
them.

arokem:
Read the license - I don't think it requires that.

Garyfallidis:
Okay. Maybe they want us to cite some of their papers?

arokem:
Yes - we should do that. Adding.

PR #682, opened 2015-07-16
Created list of examples for available features and metrics

MarcCote:
This PR adds a new example file that will eventually contains more advanced usages related to QuickBundles and the clustering framework. For now, there is only one example about how to use `ResampleFeature` to manually specify the number of points that streamlines will have when used by the QuickBundles algorithm.

This PR also fixes some typos that were present in the `segment_quickbundles.py` example.


arokem:
Is this work in progress? Are you planning to add more text/code in this example?

MarcCote:
No, it is ready to be merged. I just wanted to fix the current tutorial about QuickBundles ans show how to change the number of points in the resampling.

doc/examples/segment_quickbundles_advanced.py
```diff
@@ -0,0 +1,51 @@
...
+from nibabel import trackvis as tv
+from dipy.segment.clustering import QuickBundles
+from dipy.data import get_data
+
+fname = get_data('fornix')
+streams, hdr = tv.read(fname)
+streamlines = [i[0] for i in streams]
+
+"""
+QuickBundles using `ResampleFeature`
```
Garyfallidis: the

arokem:
OK - my impression is that this would work well as an addition to the clustering framework tutorial, rather than its own thing. As it is, it feels like it requires some context, and that other (excellent) tutorial provides that context. In other words, I doubt that anyone would ever read this on its own anyway, so why not integrate it with the other material more fluidly?

MarcCote:
I had in mind that every extensions of `Feature` and `Metric` in dipy should have a small tutorial/example. I was afraid that it might be too much to put everything in what I consider a basic example of how to use QuickBundles.
Maybe **advanced** was not the right term maybe **additional examples**?

arokem:
A matter of taste, I guess. For now, this tutorial doesn't contain multiple
additional examples, only this one about resampling. That's fine, but then
the title should really be something about resampling. And again, it would
be good if it contained some explanation about when/why you might want to
use resampling.

MarcCote:
Ok, I see your point. I'll try to add more examples this week. I'll also add more context about why one would be interested to use different Features or Metrics.

arokem:
Cool. Feel free to ping me when you're done with that, so I can take a look.

arokem:
Hey @MarcCote - just checking in about this. Have you had a chance to take another look?

MarcCote:
@arokem you can have a look. That is more what I had in mind. Right now, only the `Feature`s examples are done but I'll do the same for `Metric`.

arokem:
Nice! How about combining both of these into one tutorial? Too much?

MarcCote:
@arokem you mean combining features and metrics tutorials? I'm afraid that as we will integrate more features and metrics, it will get too big. What do you think @Garyfallidis ?

arokem:
One option is to list the different features and metrics that can be used (and how they are combined), but not explicitly write out lines of code for each one.

arokem:
Any thoughts here @Garyfallidis ?

MarcCote:
I think it is nice to have a working, independent, example for each feature and metric. Personally, when I follow a tutorial I like to be able to copy-paste the code in a IPython notebook and play with it (that is why each example have its imports).

I don't really mind putting metrics and features examples in one "tutorial", even if I don't think it is necessary. I mean it is not like we are trying to minimize the number of tutorial pages.

I don't have time right now to write the example for the metrics. If you really want to merge this PR, I can remove the metrics part and make another PR later on.

Garyfallidis:
Hello both. I think what @MarcCote suggests having 3 tutorials. A very basic for QB and then one for metrics and one for features makes perfect sense to me. So let's move towards that direction. But of course interconnect the tutorials.

Also on copying the imports on different sections of this tutorials. I think this is an exception of the classic tutorials that we write because here Marc is listing a series of features and he wants to make them copyable at any stage and order. So, I am fine with repeating the imports for this case. @arokem if you see the rendered tutorial it will be really clear. So, no strong feelings here. But I think it's fine to do what Marc proposes.

arokem:
Sure - no problem

doc/examples/segment_clustering_features.py
```diff
@@ -0,0 +1,339 @@
...
+**When:** The QuickBundles algorithm requires streamlines to have the same
+number of points. By default, QuickBundles uses `ResampleFeature` to resample
+streamlines so that they have 12 points each [Garyfallidis12]_. If you want to
+use a different number of points for the resampling, you should provide your
+own instance of `ResampleFeature` (see following example).
+
+**Note:** Resampling streamlines has an impact on clustering results both in
+term of speed and quality. Setting the number of points too low will result in
+a loss of information about the shape of the streamlines. On the contrary,
+setting the number of points to high will slow down the clustering process.
```
Garyfallidis: too

doc/examples/segment_clustering_features.py
```diff
@@ -0,0 +1,339 @@
...
+
+qb = QuickBundles(threshold=5., metric=metric)
+clusters = qb.cluster(streamlines)
+
+# Extract feature of every streamline.
+midpoints = np.asarray(map(feature.extract, streamlines))
+
+# Color each midpoint according to the cluster they belong to.
+rng = np.random.RandomState(42)
+#colormap = rng.rand(len(clusters), 3)
```
Garyfallidis: space after #

Garyfallidis: or remove comment if not needed

doc/examples/segment_clustering_features.py
```diff
@@ -0,0 +1,339 @@
...
+
+   **Showing the middle point of each streamline and colored according to the
+   QuickBundles results**.
+
+.. _clustering-examples-ArcLengthFeature:
+
+ArcLength Feature
+=================
+**What:** Instances of `ArcLengthFeature` compute the length of a streamline.
+More specifically, this feature corresponds to the sum of every streamline
```
Garyfallidis: to the sum of the lengths of ...

doc/examples/segment_clustering_features.py
```diff
@@ -0,0 +1,339 @@
...
+
+# Get some streamlines.
+streamlines = get_streamlines()  # Previously defined.
+
+feature = ArcLengthFeature()
+metric = EuclideanMetric(feature)
+qb = QuickBundles(threshold=2., metric=metric)
+clusters = qb.cluster(streamlines)
+
+#lengths = map(feature.extract, streamlines)
```
Garyfallidis: pep8 or remove

MarcCote:
I guess this will fix #715 and close #719

doc/examples/segment_clustering_metrics.py
```diff
@@ -0,0 +1,211 @@
...
+return the average of those distances. This metric takes as inputs two features
+that are sequences containing the same number of elements.
+
+**When:** By default the `QuickBundles` clustering will resample your
+streamlines on-the-fly so they have 12 points. If for some reason you want
+to avoid this and you made sure all your streamlines have already the same
+number of points, you can manually provide an instance of
+`AveragePointwiseEuclideanMetric` to `QuickBundles`. Since the default
+`Feature` is the `IdentityFeature` the streamlines won't be resampled thus
+saving some computional time.
```
Garyfallidis: computational

doc/examples/segment_clustering_metrics.py
```diff
@@ -0,0 +1,211 @@
...
+import numpy as np
+from dipy.viz import fvtk
+from dipy.segment.clustering import QuickBundles
+from dipy.segment.metric import VectorBetweenEndpointsFeature
+from dipy.segment.metric import CosineMetric
+
+# Get some streamlines.
+streamlines = get_streamlines()  # Previously defined.
+
+feature = VectorBetweenEndpointsFeature()
```
Garyfallidis: I am suggesting an alternative name VectorOfEndpointsFeature

Garyfallidis: It's shorter ...

Garyfallidis:
Nice! Good job!

PR #683, opened 2015-07-18
Fixed initial estimation in piesno

samuelstjean:
Turns out I misread something the first time, this fixes the initial estimation fo sigma for piesno. The current version uses the optimal quantile of the current slice, whereas the original paper suggest to use the full volume instead. This should hopefully fix this little discrepancy.

piesno_fix

### DIFF
--- a/dipy/denoise/noise_estimate.py
+++ b/dipy/denoise/noise_estimate.py
@@ -1,6 +1,5 @@
 from __future__ import division, print_function
 
-import warnings
 import numpy as np
 
 from scipy.special import gammainccinv
@@ -11,6 +10,18 @@ def _inv_nchi_cdf(N, K, alpha):
     """Inverse CDF for the noncentral chi distribution
     See [1]_ p.3 section 2.3"""
     return gammainccinv(N * K, 1 - alpha) / K
+
+
+# List of optimal quantil for PIESNO.
+# Get optimal quantile for N if available, else use the median.
+opt_quantile = {1: 0.79681213002002,
+                2: 0.7306303027491917,
+                4: 0.6721952960782169,
+                8: 0.6254030432343569,
+               16: 0.5900487123737876,
+               32: 0.5641772300866416,
+               64: 0.5455611840489607,
+              128: 0.5322811923303339}
 
 
 def piesno(data, N, alpha=0.01, l=100, itermax=100, eps=1e-5, return_mask=False):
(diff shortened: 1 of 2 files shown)


arokem:
First pass impression: this requires some additional tests. In particular, a test that exercises the new `m` kwarg.

arokem:
Also maybe rename that kwarg to something more meaningful

dipy/denoise/noise_estimate.py
```diff
@@ -13,6 +12,18 @@ def _inv_nchi_cdf(N, K, alpha):
     return gammainccinv(N * K, 1 - alpha) / K
 
 
+# List of optimal quantil for PIESNO.
```
Garyfallidis: e missing

arokem:
Also, out of curiosity, a question: how much of a difference does this make? That is roughly how much do the estimates with this fix differ from estimates without the fix?

dipy/denoise/tests/test_noise_estimate.py
```diff
@@ -29,6 +29,20 @@ def test_piesno():
     sigma = piesno(test_piesno_data, N=8, alpha=0.01, l=1, eps=1e-10, return_mask=False)
     assert_almost_equal(sigma, 0.010749458025559)
 
+    noise1 = (np.random.randn(100, 100, 100) * 50) + 10
+    noise2 = (np.random.randn(100, 100, 100) * 50) + 10
+    rician_noise = np.sqrt(noise1**2 + noise2**2)
+    sigma, mask = piesno(rician_noise, N=1, alpha=0.01, l=1, eps=1e-10, return_mask=True)
+
+    # less than 3% of error?
+    assert_almost_equal(np.abs(sigma - 50) / sigma < 0.03, True)
```
arokem: In particular, do these tests fail without the fix?

arokem:
Not complaining, just curious. Other than that, this looks good to me.

dipy/denoise/tests/test_noise_estimate.py
samuelstjean: they should still work, this test is for the 4D convenience function, the 3D function itself behaves in the same way.

samuelstjean:
From some tests it is exactly the same, it's pretty robust to the initialisation. It should also fix the case where 0 was returned when a low number of noise voxels was found. Since this only provides a upper bound on the highest possible value, this only matters when this highest bound was too low for a single 3D slice.

arokem:
Good. I'll give others the opportunity to take a look, and merge it in a week or so, if no other comments arise.

PR #684, opened 2015-07-18
BF: Fixed memory leak in QuickBundles.

MarcCote:
This PR fixes a memory leak that was present in `ClusterCentroid` used by the QuickBundles algorithm. The `__dealloc__` function was missing a free statement for a variable. In addition to that, references count of `memoryview` objects were not decremented automatically/correctly when freeing the allocated memory. I found that assigning them to `None` (before freeing the memory) works.

This PR also add a testing tool that retrieves the refcount of a given object type. This tool is used in the new unit test function that check there are no memory leaks in QuickBundles.

### DIFF
--- a/dipy/segment/clusteringspeed.pyx
+++ b/dipy/segment/clusteringspeed.pyx
@@ -120,6 +120,9 @@ cdef class ClustersCentroid(Clusters):
         cdef cnp.npy_intp i
         for i in range(self._nb_clusters):
             free(&(self.centroids[i].features[0, 0]))
+            free(&(self._updated_centroids[i].features[0, 0]))
+            self.centroids[i].features = None  # Necessary to decrease refcount
+            self._updated_centroids[i].features = None  # Necessary to decrease refcount
 
         free(self.centroids)
         self.centroids = NULL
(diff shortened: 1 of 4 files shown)


matthew-brett:
Thanks for doing this.  Is the memoryview problem a Cython bug?    Does it need reporting?

MarcCote:
I wouldn't say it is a bug, it is more related to the way I'm using them. I needed a variable number of memoryviews. In Cython, you can't have pointers of memoryviews so I hacked my way around using a `struct` that has a memorview as attribute.

I'm using `malloc` to allocate memory for those struct of memoryviews and use `free` accordingly. This implies there are no refcount inc/decrements. However, as soon as I assign a memoryview to something, a refcount increment is done. At first, I wasn't award of this so I didn't take the necessary precaution in the `__dealloc__` function. Once I realized that a refcount decrement wasn't done correctly I searched a lot for a Cython function allowing me to decrement the refcount of a memoryview (which doesn't required a pointer because we cannot take the address of a memoryview). I finally found that assigning them to `None` does the trick.

matthew-brett:
Thanks for the explanation.

Garyfallidis:
Good catch! :dancers:

MarcCote:
Thanks to @martcous to bringing this memory leak to my attention in the first place.

PR #686, opened 2015-07-20
BF: Make buildbot Pyhon26-32 happy

MarcCote:
Replaced an `assert_equal` with an `assert_almost_equal`.
http://nipy.bic.berkeley.edu/builders/dipy-py2.6-32/builds/551/steps/shell_6/logs/stdio

### DIFF
--- a/dipy/segment/tests/test_metric.py
+++ b/dipy/segment/tests/test_metric.py
@@ -174,7 +174,7 @@ def test_metric_cosine():
             if metric.are_compatible(f1.shape, f2.shape):
                 distance = metric.dist(f1, f2)
                 if np.all(f1 == f2):
-                    assert_equal(distance, 0.)
+                    assert_almost_equal(distance, 0.)
 
                 assert_almost_equal(distance, dipymetric.dist(metric, s1, s2))
                 assert_true(distance >= 0.)


dipy/segment/tests/test_metric.py
```diff
@@ -174,7 +174,7 @@ def dist(self, v1, v2):
             if metric.are_compatible(f1.shape, f2.shape):
                 distance = metric.dist(f1, f2)
                 if np.all(f1 == f2):
-                    assert_equal(distance, 0.)
+                    assert_almost_equal(distance, 0.)
```
matthew-brett: Is it obvious why the distance would not be exactly 0 if the vectors are exactly equal on win32?

MarcCote: There is no check verifying if the two vectors are equal in the distance function. My guess is that there is a loss of precision at lines [141](https://github.com/nipy/dipy/pull/686/files#diff-e5143f1d5de219e5851a71755ddae0b4L141)-[143](https://github.com/nipy/dipy/pull/686/files#diff-e5143f1d5de219e5851a71755ddae0b4L143), `cos_theta` must be less than 1.
```python
v1_normed = v1.astype(np.float64) / norm(v1.astype(np.float64))
v2_normed = v2.astype(np.float64) / norm(v2.astype(np.float64))
cos_theta = np.dot(v1_normed, v2_normed.T)
```
Surely, I could add an equality test but I am afraid that adding an `if` in this critical part would slow down the clustering (I didn't do any benchmarks though).

matthew-brett: Thanks for thinking about this.   Yes, I guess it must be slight differences in the precision of dot, maybe the order or operation or something.   It's a bit surprising to me that a vector dotted with itself in float64 precision would differ by 1e-9 on two platforms, but oh well.

No need to add the special case check in the distance code, I was just curious.

MarcCote: I agree with you, I thought casting them in `float64` would have avoided this precision loss.
Wait! I'm stupid :P the error is probably not with the Python version but rather in the Cython one!

In the following, both `features1` and `features2` are C float whereas `cos_theta` is a C double. If I'm not mistaken, the multiplication would be perform in 32 bits instead of 64, right?
```python
cos_theta += features1[0, d] * features2[0, d]
```

matthew-brett: I did a debug build using `try_branch.py` : http://nipy.bic.berkeley.edu/builders/dipy-py2.6-32/builds/553/steps/shell_6/logs/stdio

The error comes out of the Cython code, yes.  I can't reproduce it with the same code in numpy, I think the same precision.  Still not sure where it comes from.

matthew-brett:
I took a look at this, for a long time, and:

* I was wrong, the error is for the Python code, not the Cython code;
* The problem is caused by slightly weird inaccuracy of `np.dot` on the 32-bit machine.

```
# check_features.py
import hashlib
import numpy as np

fvec = np.array([[10, 1.5021631121635437e-01, 1.2246063538223773e-16]],
                dtype=np.float32)

dvec = fvec.astype(np.float64)
dvec_n = dvec / np.linalg.norm(dvec)
print(hashlib.sha1(dvec_n.tostring()).hexdigest())
print(repr(np.sum(dvec_n ** 2)))
print(repr(np.dot(dvec_n, dvec_n.T)))
```

On a 64-bit OSX machine:

```
In [5]: run check_features.py
335df17574da17f252f7df12998d34b94b072317
1.0
array([[ 1.]])
```

On the 32-bit machine giving the test failure:

```
In [114]: run check_features.py
335df17574da17f252f7df12998d34b94b072317
1.0
array([[ 0.99999999999999988898]])
```

So, this is a 32-bit `dot` implementation problem, giving fairly small error in a checking function, so I think it's OK to sweep it under the carpet with this PR.

MarcCote:
Thanks for having investigated this @matthew-brett.

PR #689, opened 2015-07-28
Streamlines compression

MarcCote:
This PR implements the linearization algorithm found in [Presseau et al. *Neuroimage*, 2015](http://scil.dinf.usherbrooke.ca/wp-content/papers/presseau-etal-ni15.pdf).

Basically, it consists in merging consecutive segments that are near collinear. The merging is achieved by removing the point the two segments share.

It can compress 300k streamlines in 10 secs, on a i7-3820 CPU, 3.60GHz. I didn't compare the speed of the C++ version vs. this Cython version, but I will do it and report the speed soon.
@mdesco when you have some time, any comments about the documentation would be appreciated.

EDIT: I timed the C++ version found in https://github.com/scilus/FiberCompression/ on the same 300k streamlines and it took around 2.5 min... in debug mode! I re-timed it, this time in release mode and obtained 48 sec (this includes loading and writing whereas my python timing does not).

EDIT2: Ok I have timing results for both C++ and Cython version. Both version load streamlines, compress them and write them down on disk. The timings are fairly similar.
C++: 48 sec
Cython: 41 sec


dipy/tracking/streamlinespeed.pyx
```diff
@@ -301,3 +301,174 @@ def set_number_of_points(streamlines, nb_points=3):
...
+            for d in range(D):
+                out[nb_points, d] = streamline[i, d]
+
+            nb_points += 1
+
+    return nb_points
+
+
+def compress_streamlines(streamlines, tol_error=0.5):
+    """ Compress streamlines by linearizing some part of them.
```
mdesco: Compress streamlines by linearization as in Presseau et al NeuroImage 2015.

dipy/tracking/streamlinespeed.pyx
```diff
@@ -301,3 +301,174 @@ def set_number_of_points(streamlines, nb_points=3):
...
+
+    Basically, it consists in merging consecutive segments that are
+    near collinear. The merging is achieved by removing the point the two
+    segments have in common. The linearization process [Presseau15]_. will
+    not remove a point if it causes either an offset of more than
+    `tol_error`mm or a distance between two consecutive points to be more
+    than 10mm [Rheault15]_.
```
mdesco: I would define tol_error as tolerance error (tol_error) somewhere.

dipy/tracking/streamlinespeed.pyx
```diff
@@ -301,3 +301,174 @@ def set_number_of_points(streamlines, nb_points=3):
...
+
+    Parameters
+    ----------
+    streamlines : one or a list of array-like of shape (N,3)
+        Array representing x,y,z of N points in a streamline
+    tol_error : float (optional)
+        Tolerance error in mm. Default is 0.5mm.
```
mdesco: I would not use 0.5 mm as a default. This is too aggressive. We can ask Francois Rheault opinion but I would use 0.1 as a default.

I would also add a comment that recommendations can be found in Presseau et al. But, a general rule of thumb is that 0.01 for deterministic streamline as input and 0.1 for proba are good choices.

dipy/tracking/tests/test_streamline.py
```diff
@@ -406,5 +408,93 @@ def test_select_streamlines():
...
+
+    # Make sure Cython and Python versions are the same.
+    for tol_error in np.linspace(0.001, 1, 100):
+        cstreamline_python = compress_streamlines_python([streamline],
+                                                         tol_error)[0]
+        cstreamline_cython = compress_streamlines(streamline, tol_error)
+        assert_equal(len(cstreamline_python), len(cstreamline_cython))
+        assert_array_almost_equal(cstreamline_python, cstreamline_cython)
+
+
```
mdesco: good job!

MarcCote:
Someone has any idea why the following unit test is failing with Python 3.4: 
https://travis-ci.org/nipy/dipy/jobs/73024092
```
FAIL: dipy.align.tests.test_imaffine.test_mi_gradient
  File "/home/travis/build/nipy/dipy/venv/lib/python3.4/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/build/nipy/dipy/venv/lib/python3.4/site-packages/dipy/align/tests/test_imaffine.py", line 307, in test_mi_gradient
    assert(nprod >= 0.999)
AssertionError
```

dipy/tracking/streamlinespeed.pyx
```diff
@@ -301,3 +301,188 @@ def set_number_of_points(streamlines, nb_points=3):
...
+            nb_points += 1
+
+    return nb_points
+
+
+def compress_streamlines(streamlines, tol_error=0.1, max_segment_length=10):
+    """ Compress streamlines by linearization as in [Presseau15]_.
+
+    The compression consists in merging consecutive segments that are
+    near collinear. The merging is achieved by removing the point the two
```
samuelstjean: point is missing an s (possibly)

dipy/tracking/streamlinespeed.pyx
```diff
@@ -301,3 +301,188 @@ def set_number_of_points(streamlines, nb_points=3):
...
+    segments have in common.
+
+    The linearization process [Presseau15]_ ensures that every points being
+    removed are within a certain margin (in mm) of the resulting streamline.
+    Recommendations for setting this margin can be found in [Presseau15]_
+    (in which they called it tolerance error).
+
+    The compression also ensures that two consecutive points won't be to far
```
samuelstjean: too far

dipy/tracking/streamlinespeed.pyx
```diff
@@ -301,3 +301,188 @@ def set_number_of_points(streamlines, nb_points=3):
...
+    from each other (precisely less or equal than `max_segment_length`mm).
+    This is purely to speed up the linearization process [Rheault15]_. A low
+    value will result in a faster linearization but low compression, whereas
+    a high value will result in a slower linearization but high compression.
+
+    Parameters
+    ----------
+    streamlines : one or a list of array-like of shape (N,3)
+        Array representing x,y,z of N points in a streamline
```
samuelstjean: missing a . at the end

dipy/tracking/streamlinespeed.pyx
```diff
@@ -301,3 +301,188 @@ def set_number_of_points(streamlines, nb_points=3):
...
+    tol_error : float (optional)
+        Tolerance error in mm (default: 0.1). A rule of thumb is to set it
+        to 0.01mm for deterministic streamlines and 0.1mm for probabilitic
```
samuelstjean: so, the default assumes prob tracking, mayb e you should change it to 0.01 since 0.01 < 0.1, as to not crew up det tracking by default, and people finding that their prob tracking is still big can change it.

dipy/tracking/streamlinespeed.pyx
```diff
@@ -301,3 +301,188 @@ def set_number_of_points(streamlines, nb_points=3):
...
+        streamlines.
+    max_segment_length : float (optional)
+        Maximum length of any given segment produced by the compression.
```
samuelstjean: The max length segment default at 10mm (is it in mm? it's not indicated), which is sure to kill off any CST. Jasmeen suggest 220 mm to be safe, so I suggest in turn 250mm to be on the always safe side. Plus tracking should take care of that, so if it's not expensive you could put the maximum length in the tractogram.

dipy/tracking/streamlinespeed.pyx
```diff
@@ -301,3 +301,188 @@ def set_number_of_points(streamlines, nb_points=3):
...
+    [100, 50]
+    >>> [len(s) for s in c_streamlines]
+    [12, 5]
+
+
+    References
+    ----------
+    .. [Presseau15] Presseau C. et al., A new compression format for fiber
+                    tracking datasets, NeuroImage, no 109, 73-83, 2015.
+    .. [Rheault15] Rheault F. et al., In writing, 2015.
```
samuelstjean: you can still put the ismrm thingy in the meantime

dipy/tracking/streamlinespeed.pyx
samuelstjean: oh wait, that's the max allowed between two points in mm? Then my previous comment is useless. Anyway, you could also add as a note to not do stat on compressed fibers or the result won't be valid at all, just in case people start compressing like crazy and they don't check back changed docstrings.

omarocegueda:
Hi @MarcCote,

>Someone has any idea why the following unit test is failing with Python 3.4: 

I wrote this test, thanks for reporting it!, I am investigating

dipy/tracking/streamlinespeed.pyx
MarcCote: No, it is the point between two segments.

dipy/tracking/streamlinespeed.pyx
MarcCote: Good point, it makes sense to set the default at 0.01mm.

dipy/tracking/streamlinespeed.pyx
MarcCote: Your previous comment is still valid. Yes, I'll add a note that says to be careful when computing stats on compressed streamlines.

samuelstjean: previous comment refered ot the one I deleted cause it was nonsense after reading the code, I misread the docstring instead XD

dipy/tracking/streamlinespeed.pyx
```diff
@@ -301,3 +301,196 @@ def set_number_of_points(streamlines, nb_points=3):
...
+    nb_points += 1
+
+    return nb_points
+
+
+def compress_streamlines(streamlines, tol_error=0.01, max_segment_length=10):
+    """ Compress streamlines by linearization as in [Presseau15]_.
+
+    The compression consists in merging consecutive segments that are
+    near collinear. The merging is achieved by removing the point the two
```
jchoude: near -> nearly

dipy/tracking/streamlinespeed.pyx
```diff
@@ -301,3 +301,196 @@ def set_number_of_points(streamlines, nb_points=3):
...
+    segments have in common.
+
+    The linearization process [Presseau15]_ ensures that every points being
```
jchoude: points should by singular (because it follows every)

dipy/tracking/tests/test_streamline.py
```diff
@@ -406,5 +408,95 @@ def test_select_streamlines():
...
+        # Pre-allocation
+        current_tract = np.zeros((len(streamlines[i]), 3))
+        nb_points = 0
+        last_added_point = 0
+        # For all points on one streamline
+        for j in range(len(streamlines[i])):
+            # Distance between last point added and current point
+            in_between = dist_between_points(streamlines[i][last_added_point],
+                                             streamlines[i][j])
+            # Always add the first points
```
jchoude: point (no s)

jchoude:
Thanks @MarcCote for doing this ;) I can now remove it from my TODO list ;)

Everything seems good, except one thing: as you hinted when mentioning the timing of the original C++ code, there is no "max segment length" argument in the original algorithm. It was added in François version in a 3D viewer to make sure it stayed quite fast, but you can have a slight drop in your compression ratio when you use it (especially if the max segment length is set to a small value)... I would at least like to have a way to mimic the behaviour of the original algorithm, by giving a max segment length of "infinity" or something like that, so as to get the maximal compression when needed.

What do you think?

dipy/tracking/streamlinespeed.pyx
MarcCote: Thank you very much, I don't know why I thought otherwise. I was sure a noun following "every" should be plural while noun following "each" should be singular! Seems I was wrong thanks :D.

MarcCote:
@jchoude you mean to make the default behaviour as the C++ version? Otherwise you can called the function like this `compress_streamlines(..., max_segment_length=np.inf)` without any problem.

samuelstjean:
Maybe he meant you should mimic the original behavior  with the default of compress_streamlines(..., max_segment_length=np.inf), instead of having the modified françois's version as the default, especially since you reference the default one. And you can add a link to this implementation for further reference as necessary -> https://github.com/scilus/FiberCompression

jchoude:
@MarcCote I'm not sure what the default behaviour should be, but of course I might have a slight bias towards the "classic" behaviour (i.e. no maximal segment length), just for the sake of uniformity with the original tool.

That being said, the method here doesn't implement the other parts of our paper (Huffman coding, quantization, etc.), which is ok because the linearization phase (the one that is implemented here) is where you get the most gains. So, in a sense, since it's not the "complete" algorithm, I would say that adding this option with a reasonable default makes perfect sense. I would **however** clearly document the way to get the original behavior, else it might not be obvious to users that didn't read the code...

jchoude:
@MarcCote Oh and another question: when you got your timing of 2:30 using the C++ tool, how did you run it? You might have launched the full process and not just the linearization... Also, were you in Debug or release? ;)

MarcCote:
@jchoude I ran the following command:
`time fibercompression streamlines.trk tmp.trk -c`
and I roughly substracted the loading/saving time.

Maybe I missed something but I didn't see any other compression options in `fibercompression -h`!

dipy/tracking/streamlinespeed.pyx
```diff
@@ -301,3 +301,197 @@ def set_number_of_points(streamlines, nb_points=3):
...
+        for d in range(D):
+            dn = streamline[i, d] - streamline[prev, d]
+            segment_length += dn*dn
+
+        segment_length = sqrt(segment_length)
+        next = i+1
+
+        # Check that each point is not offset by more than `tol_error` mm.
+        for curr in range(prev, i):
+            dist = c_dist_to_line(streamline, prev, next, curr)
```
FrancoisRheaultUS: The for loop should start at 'prev' and end at 'i-1', because you are testing the in-between point (between the 'prev' and 'i'

FrancoisRheaultUS: dist = c_dist_to_line(streamline, prev, i, curr)

(will not crash, since the for loop should go from 'prev' to 'i-1'

dipy/tracking/streamlinespeed.pyx
```diff
@@ -301,3 +301,197 @@ def set_number_of_points(streamlines, nb_points=3):
...
+                prev = i-1
+                last_success = i-1
+                break
+            # The current point's offset is ok, check the next point.
+            else:
+                last_success = i
+
+    # Copy last point since it is always kept.
+    for d in range(D):
+        out[nb_points, d] = streamline[N-1, d]
```
FrancoisRheaultUS: Im not sure why, but this line doesn`t seem to work properly.
When I verified if the streamlines had the same end point, the last point of the compressed streamline is in fact the last-1 of the original streamline

MarcCote: @FrancoisRheaultUS  can you send me the code you use? I can't reproduce your error. In the unit tested I explicitly test that the first and last points should be the same as in the original streamline.

dipy/tracking/streamlinespeed.pyx
MarcCote: In Python the `range` function gives you the integers between start and end excluded, i.e. [start, end). So the code will effectively stop at `i-1`.

Regarding your comment about using `i` instead of `next`, I'll check it in more details.

MarcCote:
The python version had a bug hence the Cython version was erroneous too. Some points were removed even though the `dist > tol_error`. I fixed both versions and cleaned the Python one while at it. I added a unit test checking this erroneous behaviour.

@FrancoisRheaultUS can you confirm it works as the C++ version, thanks.

jchoude:
@MarcCote For the timing of the original executable, the default call has an error threshold of 0.5 mm. Is this what you used in your cython timing? Having that "huge" of an error threshold can really lengthen the computations...

Also, did you make sure you had a Release build?

MarcCote:
@jchoude I didn't explicitly specified to build in release :P. I re-timed it and obtain way better performance:
48 sec (including loading and writing).

Garyfallidis:
Is this ready for a review?

Garyfallidis:
I mean second round of review / merge stage.

MarcCote:
Yes, it is ready to be merged. @mdesco also tried it on real data.

mdesco:
For me, this is ready to be merged.

Garyfallidis:
Good job. Thx! Let's start compressing!

PR #690, opened 2015-07-29
BF: nan entries cause segfault

omarocegueda:
This fixes the segmentation fault caused when attempting to interpolate an image at nan entries.

### DIFF
--- a/dipy/align/imaffine.py
+++ b/dipy/align/imaffine.py
@@ -144,6 +144,8 @@ class AffineMap(object):
             self.affine_inv = None
             return
         try:
+            if np.isnan(np.sum(affine)):
+                raise AffineInversionError('Affine contains invalid elements')
             self.affine_inv = npl.inv(affine)
         except npl.LinAlgError:
             raise AffineInversionError('Affine cannot be inverted')
--- a/dipy/align/imaffine.py
+++ b/dipy/align/imaffine.py
@@ -143,9 +143,9 @@ class AffineMap(object):
         if self.affine is None:
             self.affine_inv = None
             return
+        if np.any(np.isnan(affine)):
+            raise AffineInversionError('Affine contains invalid elements')
         try:
-            if np.isnan(np.sum(affine)):
-                raise AffineInversionError('Affine contains invalid elements')
             self.affine_inv = npl.inv(affine)
         except npl.LinAlgError:
             raise AffineInversionError('Affine cannot be inverted')
--- a/dipy/align/tests/test_imaffine.py
+++ b/dipy/align/tests/test_imaffine.py
@@ -172,7 +172,7 @@ def test_align_origins_3d():
 
 def test_affreg_all_transforms():
     # Test affine registration using all transforms with typical settings
-    for ttype in factors.keys():
+    for ttype in sorted(factors):
         dim = ttype[1]
         if dim == 2:
             nslices = 1
@@ -265,7 +265,7 @@ def test_mi_gradient():
     np.random.seed(2022966)
     # Test the gradient of mutual information
     h = 1e-5
-    for ttype in factors:
+    for ttype in sorted(factors):
         transform = regtransforms[ttype]
         dim = ttype[1]
         if dim == 2:
@@ -304,4 +304,4 @@ def test_mi_gradient():
         enorm = np.linalg.norm(expected)
         anorm = np.linalg.norm(actual)
         nprod = dp / (enorm * anorm)
-        assert(nprod >= 0.999)
+        assert(nprod >= 0.99)
--- a/dipy/align/tests/test_imaffine.py
+++ b/dipy/align/tests/test_imaffine.py
@@ -172,6 +172,12 @@ def test_align_origins_3d():
 
 def test_affreg_all_transforms():
     # Test affine registration using all transforms with typical settings
+
+    # Make sure dictionary entries are processed in the same order regardless of
+    # the platform. Otherwise any random numbers drawn within the loop would make
+    # the test non-deterministic even if we fix the seed before the loop.
+    # Right now, this test does not draw any samples, but we still sort the entries
+    # to prevent future related failures.
     for ttype in sorted(factors):
         dim = ttype[1]
         if dim == 2:
@@ -265,6 +271,11 @@ def test_mi_gradient():
     np.random.seed(2022966)
     # Test the gradient of mutual information
     h = 1e-5
+    # Make sure dictionary entries are processed in the same order regardless of
+    # the platform. Otherwise any random numbers drawn within the loop would make
+    # the test non-deterministic even if we fix the seed before the loop:
+    # in this case the samples are drawn with `np.random.randn` below
+
     for ttype in sorted(factors):
         transform = regtransforms[ttype]
         dim = ttype[1]
--- a/dipy/align/tests/test_vector_fields.py
+++ b/dipy/align/tests/test_vector_fields.py
@@ -793,15 +793,19 @@ def test_affine_transforms_2d():
 
     # Test exception is raised when the affine transform matrix is not valid
     invalid = np.zeros((2, 2), dtype=np.float64)
+    invalid_nan = np.zeros((3, 3), dtype=np.float64)
+    invalid_nan[1, 1] = np.nan
     shape = np.array(codomain_shape, dtype=np.int32)
-    # Exceptions from warp_2d
+    # Exceptions from transform_2d
     assert_raises(ValueError, vfu.transform_2d_affine, circle, shape, invalid)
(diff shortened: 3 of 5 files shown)


dipy/align/imaffine.py
```diff
@@ -144,6 +144,8 @@ def set_affine(self, affine):
             self.affine_inv = None
             return
         try:
+            if np.isnan(np.sum(affine)):
```
omarocegueda: Using `sum` is faster than, e.g., `min` for finding nan's, according to this:
http://stackoverflow.com/questions/6736590/fast-check-for-nan-in-numpy

dipy/align/tests/test_vector_fields.py
```diff
@@ -793,15 +793,19 @@ def test_affine_transforms_2d():
...
+    invalid_nan = np.zeros((3, 3), dtype=np.float64)
+    invalid_nan[1, 1] = np.nan
     shape = np.array(codomain_shape, dtype=np.int32)
-    # Exceptions from warp_2d
+    # Exceptions from transform_2d
     assert_raises(ValueError, vfu.transform_2d_affine, circle, shape, invalid)
     assert_raises(ValueError, vfu.transform_2d_affine, circle, shape, invalid)
     assert_raises(ValueError, vfu.transform_2d_affine, circle, shape, invalid)
-    # Exceptions from warp_2d_nn
+    assert_raises(ValueError, vfu.transform_2d_affine, circle, shape, invalid_nan)
```
omarocegueda: Without this fix, this assert will reproduce the segfault.

dipy/align/vector_fields.pyx
```diff
@@ -17,6 +17,8 @@ def is_valid_affine(double[:, :] M, int dim):
         return True
     if M.shape[0] < dim or M.shape[1] < dim + 1:
         return False
+    if np.isnan(np.sum(M)):
```
omarocegueda: We are already calling this check from all the exposed cython functions that use affines, so we can check for nans here.

arokem:
Does this actually address all of the errors mentioned in #654? 

In particular, one of the things reported there was an error in test_imaffine:test_mi_gradient (for example, see https://travis-ci.org/nipy/dipy/jobs/73233253#L2162), which is not a segfault. Is this addressed through these changes?

omarocegueda:
I don't know if this will solve the other bug, but the segmentation fault occurs in several buildbots:
http://nipy.bic.berkeley.edu/builders/dipy-bdist32-33/builds/138/steps/shell_8/logs/stdio
http://nipy.bic.berkeley.edu/builders/dipy-py2.6-32/builds/557/steps/shell_6/logs/stdio
http://nipy.bic.berkeley.edu/builders/dipy-py2.7-osx-10.7/builds/310/steps/shell_6/logs/stdio
http://nipy.bic.berkeley.edu/builders/dipy-py2.7-osx-10.8/builds/378/steps/shell_6/logs/stdio

The other bug appears in this buildbot:
http://nipy.bic.berkeley.edu/builders/dipy-py3.4/builds/192/steps/shell_6/logs/stdio
@matthew-brett, can I have access to that machine too?
Do you guys think it will be better to add both fixes to this PR?

arokem:
OK - just wanted to make sure that I have the full picture.
On Wed, Jul 29, 2015 at 4:38 PM, Omar Ocegueda <notifications@github.com>
wrote:

> I don't know if this will solve the other bug, but the segmentation fault
> occurs in several buildbots:
>
> http://nipy.bic.berkeley.edu/builders/dipy-bdist32-33/builds/138/steps/shell_8/logs/stdio
>
> http://nipy.bic.berkeley.edu/builders/dipy-py2.6-32/builds/557/steps/shell_6/logs/stdio
>
> http://nipy.bic.berkeley.edu/builders/dipy-py2.7-osx-10.7/builds/310/steps/shell_6/logs/stdio
>
> http://nipy.bic.berkeley.edu/builders/dipy-py2.7-osx-10.8/builds/378/steps/shell_6/logs/stdio
>
Yes - I expect this PR will solve these issues. Have you tried a a
try_branch with any of these?

> The other bug appears in this buildbot:
>
> http://nipy.bic.berkeley.edu/builders/dipy-py3.4/builds/192/steps/shell_6/logs/stdio
> @matthew-brett <https://github.com/matthew-brett>, can I have access to
> that machine too?
> Do you guys think it will be better to add both fixes to this PR?
>
Do you have some sense of what is causing it? I don't think it matters much
if this one also gets resolved here: If you think this resolves the
segfaults, I am ready to merge this one, and you can take the other fix on
another PR, but if you prefer to keep going here, that's fine too. Your
call.

omarocegueda:
I didn't run a `try_branch` but I tested the fix manually on the buildbot Matthew gave access to (It didn't brake after the fix). 

I'm not sure about the other bug, it may be as simple as a precission issue (e.g. the result was something like 0.9989, very close to the assertion value but still failing). I need to reproduce the bug and find out why it fails there and not in other platforms.

dipy/align/imaffine.py
matthew-brett: Surely these two lines should go above the `try` block?   I think the `np.sum` trick is to avoid large temporary arrays, for me a simple `np.any(np.isnan(x))` is faster still, and I think it's easier to read.

dipy/align/vector_fields.pyx
matthew-brett: Ditto here for `np.any(np.isnan(M))`

matthew-brett:
This is just option 3?

matthew-brett:
Omar - I'll check with the owner of that buildbot machine and the OSX machine - will get back to you.

matthew-brett:
Now I think I see that this is option 2 and 3.

omarocegueda:
Thanks @matthew-brett!, yes this is option 2 and 3. I just reproduced the bug on the buildbot. The root cause is that iteration over dictionary keys is no longer deterministic in Python 3:
http://stackoverflow.com/questions/14956313/dictionary-ordering-non-deterministic-in-python3

This explains the "intermitent" behavior. The assertion was failing because the inner product between numeric and analytical gradients was about 0.994. I think it is safe to reduce the threshold to 0.99. For the non-deterministic behavior, I guess the way to go is to replace the dictionary with a list. What do you think?

matthew-brett:
Where is the iteration-over-dictionary code?

omarocegueda:
Here is the dictionary:
https://github.com/nipy/dipy/blob/master/dipy/align/tests/test_imaffine.py#L25
and here is the loop that is failing:
https://github.com/nipy/dipy/blob/master/dipy/align/tests/test_imaffine.py#L268

matthew-brett:
How about `for ttype in sorted(factors)` to make it deterministic?

omarocegueda:
sure! that's fine too. I'll do the fix.

omarocegueda:
Alright!, this now addresses both failures (checked directly on the failing buildbots)

dipy/align/tests/test_imaffine.py
```diff
@@ -172,7 +172,7 @@ def test_align_origins_3d():
 
 def test_affreg_all_transforms():
     # Test affine registration using all transforms with typical settings
-    for ttype in factors.keys():
+    for ttype in sorted(factors):
```
matthew-brett: Comment to explain why the factor keys must be sorted (in order to preserve relationship of random numbers to dict key / values)?  Ditto for other instances.

omarocegueda: Comments added. Actually only one of the tests needed the sort, but I think it is ok to still sort the keys in the other three places, just in case we extend the tests in the future.

matthew-brett:
There seems to be a new merge commit here - 33451ca - introducing lots of changes not relevant to this PR?

omarocegueda:
Right, now I see what happened... fortunately there is `cherry-pick`!
Thanks!

dipy/align/imaffine.py
```diff
@@ -143,6 +143,8 @@ def set_affine(self, affine):
         if self.affine is None:
             self.affine_inv = None
```
matthew-brett: Sorry to do small suggestions not relevant to this PR - but how about putting this line after `self.affine = affine` so that `self.affine_inv` is always defined (as None or an affine) even if the inverse raises an error. Otherwise it's set to whatever it was before, which could be confusing.  Probably also worth noting that the method sets `self.affine_inv` in the docstring too.

matthew-brett: Actually, don't worry, I'll do a PR for that.

matthew-brett:
Great - thanks for fixing - merging now.

dipy/align/imaffine.py
matthew-brett: https://github.com/nipy/dipy/pull/693

PR #681, opened 2015-07-16
RF + DOC: Add MNI template reference. Also import it into the dipy.da…

arokem:
…ta namespace.

PR #691, opened 2015-07-29
CENIR 'HCP-like' multi b-value data

arokem:
Needed for #677

### DIFF
--- a/dipy/data/__init__.py
+++ b/dipy/data/__init__.py
@@ -46,7 +46,7 @@ from dipy.data.fetcher import (fetch_scil_b0,
                                read_stanford_t1,
                                fetch_stanford_pve_maps,
                                read_stanford_pve_maps,
-                               fetch_cenir_multib,s
+                               fetch_cenir_multib,
                                read_cenir_multib)
 
 from ..utils.arrfuncs import as_native_array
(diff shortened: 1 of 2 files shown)


MrBago:
@arokem could you use the `fetch_data` method in your fetchers? The way `fetch_data` is written, if the data exists, it confirms the md5 and return. If the md5 doesn't match or the file does not exist it will download the file. `fetch_data` has tests written for it, but we cannot write tests for each individual "fetcher."

Also, I think the "read_data" functions should call the "fetch_data" functions. If the files exist, calling the "fetcher" will simply serve to confirm the md5 is correct and the data has not been changed or corrupted. If the file does not exist, then the "fetcher" will try and download the data. Also if the reader is calling the fetcher, the fetcher can return the filenames to the reader so that we don't need to hard-code the file names in 2 different places.

This model makes for less code duplication, better testing and less hard-coding of file names, directories. Look at `fetch_standford_labels`/`read_stanford_labels` if you need an example.

MrBago:
I think we should at some point apply this change to all the "fetchers", but for now I would settle for using it in new "fetchers".

arokem:
Yeah - good call. I did this on #681 as well.

Once these two are merged (hint, hint), I volunteer to do a full refactor
in this direction on all of the fetchers/readers. This module is indeed a
bit of a mess right now.

Garyfallidis:
@MrBago this looks good to me. If you agree then please go ahead and merge this.

arokem:
Hang on with this one. As pointed out by @RafaelNH, there's currently no good way to pull the affine out from this. Working on it.

arokem:
OK - all set with that now.

PR #695, opened 2015-08-04
Introducing workflows

matthieudumont:
Introducing the workflows structure with the median_otsu demonstration workflow


dipy/workflows/segment.py
```diff
@@ -0,0 +1,44 @@
...
+
+import nibabel as nib
+import numpy as np
+
+from dipy.segment.mask import median_otsu
+
+
+def median_otsu_bet(input_file, out_dir, save_masked=False, median_radius=4,
+                    numpass=4, autocrop=False, vol_idx=None, dilate=None):
+
```
arokem: docstring?

dipy/workflows/segment.py
```diff
@@ -0,0 +1,44 @@
...
+from os.path import join, basename, splitext, dirname, isabs, exists
+from os import makedirs
+
+import nibabel as nib
+import numpy as np
+
+from dipy.segment.mask import median_otsu
+
+
+def median_otsu_bet(input_file, out_dir, save_masked=False, median_radius=4,
```
Garyfallidis: I am suggesting that all the functions in workflows should end with the word ``flow``.
``def median_otsu_flow``
and that all parameters should be strings by default. In that way there will be no need for parsing the arguments outside of the flow. That would make life easier when integrating with nipype and client/server web interfaces.
Look for inspiration here:
https://github.com/Garyfallidis/dipy/blob/slr_tutorials/dipy/workflows/segment.py#L54

Garyfallidis: Of course you will need to write a docstring with an explanation of the parameters and perhaps an example of usage. The good news is that most of this information can then be copied in the dipy/bin file.

dipy/workflows/segment.py
```diff
@@ -0,0 +1,44 @@
...
+                                   vol_idx, dilate)
+
+        fname, ext = splitext(basename(fpath))
+        if(fname.endswith('.nii')):
+            fname, _ = splitext(fname)
+            ext = '.nii.gz'
+
+        mask_fname = fname + '_mask' + ext
+
+        if out_dir == '':
```
Garyfallidis: Okay so here you say that if empty string then use the directory of fpath. Nice.

dipy/workflows/segment.py
```diff
@@ -0,0 +1,44 @@
...
+            out_dir_path = dirname(fpath)
+        elif not isabs(out_dir):
```
Garyfallidis: That you say that if ``out_dir`` is not an absolute path you can still append it in the directory of fpath. Good idea! I will change my workflows to do the same.

dipy/workflows/segment.py
```diff
@@ -0,0 +1,44 @@
...
+import nibabel as nib
+import numpy as np
+
+from dipy.segment.mask import median_otsu
+
+
+def median_otsu_bet(input_file, out_dir, save_masked=False, median_radius=4,
+                    numpass=4, autocrop=False, vol_idx=None, dilate=None):
+
+    for fpath in glob(input_file):
```
Garyfallidis: Because this allows it to run with multiple files it would be nice to have some print out too. Showing what is being currently processed.

bin/dipy_median_ostu
```diff
@@ -0,0 +1,18 @@
...
+
+from dipy.workflows.segment import median_otsu_bet
+from dipy.fixes import argparse as arg
+
+parser = arg.ArgumentParser()
+parser.add_argument('input', action='store')
+parser.add_argument('--out_dir', type=str, default='')
+parser.add_argument('--save_masked', action='store_true')
+parser.set_defaults(save_masked=False)
+parser.add_argument('--radius', type=int, default=4)
```
Garyfallidis: This is nice. Apart from the missing documentation/help/example. 
But here I am suggesting a radical change which I believe will save us from concentrating on the cmds and ArgumentParsers and work more on the actual workflow. So, my suggestion is that we should assume that all inputs of workflows are always strings and therefore we will cast the needer parameters inside the workflow.  I am still playing in my head with this idea but I think it will simplify many things in the relative close to far future. 

So, no casting will happen in the ArgumentParser too. All inputs are strings and are given to the relevant workflow.
@arokem, @matthieudumont  can you think a bit this suggestion? Are the advantages clear or do I need to elaborate more?

Garyfallidis: Hmm... think of this for a bit as a client/server problem. Let's say we had a client that would get some inputs and send them to the server to run the workflow.  The client is only able to communicate with the server by sending the inputs as strings. So, I think that the most generic design which we can have for the workflows would be that all parameters are strings. But hey I may be wrong. Happy to hear other opinions about this.

bin/dipy_median_ostu
```diff
@@ -0,0 +1,18 @@
+#! /usr/bin/env python
```
Garyfallidis: The name is dipy_median_otsu
Also it is not executable. You need to change the permissions of the file. Or we have to make it 
executable when we install it. Not sure.

Garyfallidis:
@chrisfilo, @oesteban and @arokem with this workflow design I think it would be trivial to use nipype to call these workflows from dipy. Can you think of this and let me know your thoughts? @matthieudumont and me and hopefully others too will be working on making many of those available very soon in dipy. We want to have workflows that will be called very easily from other pipelines but which also process multiple files and hopefully parameters. The trick here is to have in dipy.workflows all workflows with string params. Let me know what you think. Thanks in advance.

chrisfilo:
Yes it would be fairly easy to wrap such functions. I'm not sure if "workflow" is the best name (since it suggest many processing steps and long running time), but it's just a minor point.

I know we talked about this numerous times, but I would like to once again encourage you to write a Nipype interface instead of argparse. This way writing code once you get:
- command line functionality
- integration with Nipype
- integration with CBrain (through boutiques)

arokem:
I'm with @chrisfilo on this one. Seems like a duplication of effort to do this here, and if we start down the path, we will have to eventually reinvent all the wheels that nipype has already invented. 

For example, it's clear that there are already emerging patterns here, that appear as boiler-plate in the code in this PR. Such things as checking whether output locations exist, whether the files required exist (and if not, what needs to be done), and so on. 

Seems like a better use of time to write simplified APIs in dipy for processes that require many steps (e.g. registration with the SyN algorithm, see my attempts over on #673), and write workflows in nipype.

Garyfallidis:
Hi @chrisfilo. Thank you for your quick feedback. Workflow is used in different projects with a different meaning. Nonetheless, I think it is a good name exactly because in dipy.workflows many functions/methods are called from the different individual workflows. The workflow introduced here is now one of the simplest but there there will be other more complicated ones. Also there will be some generic ones where you will be able to do a complete dwi analysis by running a single command. 

About the nipype interfaces. As I said before I am very happy to write nipype interfaces. But also I want to be as general as possible. So I would like to support both ArgParser and Nipype and calling this from a web client and from another script etc. 

So what I want you to tell me is what would be the best  way to write an inferface for ``dipy.workflows.segment.median_otsu_flow``.

Is this a good example to start from?
https://github.com/nipy/nipype/blob/master/nipype/interfaces/dipy/tracks.py

If I remember correctly I need to write an 
InputSpec, OutputSpec and a BaseInterface
is this correct?

So assuming that all input and outputs are given as string. Can you give me a recommendation of how this classes would look?

Garyfallidis:
Hi @arokem. I disagree that moving all the workflows in nipype will be a complete solution for dipy. I think we really need functions that handle io without extra dependencies. Also it will be much easier to have our quality assurance workflows tested first from the dipy developers and then submit the best ones to nipype. And these functions can be also called from our visualization module right now without waiting for a nipype release. I would like to play with a new technology before making radical decisions. So please do not rush into conclusions.

Garyfallidis:
As about checking if the files exist. I think these are trivial to write. But our scripts also allow in a beautiful way to run multiple subjects sitting in different directories (using ``*`` and other wild-cards) and possibly even multiple parameters and multiple outputs from only one command. I love this option. For example, you can write

``dipy_median_otsu "Data/subj_1*0/t1.nii.gz" ...``

If I remember correctly from what @chrisfilo said when we met in OHBM this is still experimental or under development in nipype.

chrisfilo:
This is a good starting point: 
http://nipy.org/nipype/devel/interface_specs.html

Python example:
http://nipy.org/nipype/devel/python_interface_devel.html

Real world example from nipy:
https://github.com/nipy/nipype/blob/master/nipype/interfaces/nipy/utils.py#L28

On the subject of command line inputs. Your current syntax is equivalent to how bash is dealing with wildcards, but will only work if users will specify parameters in parentheses. I would recommend leaving wild-cards expansion to bash (it's more intuitive for any command line user) and writing your function in such way that accepts lists of files instead of strings with wildcards.

In nipype input specification list of files are described this way:

	input_files = InputMultiPath(File(exists=True), desc='list of files')
	
this will accept both "/path/file" and ["/path/file"] as well as ["/path/file1", "/path/file2"].

Let me know if you have any specific questions!

Garyfallidis:
Hi @chrisfilo thanks for the links! Cool, I will have something ready for you to see asap. Will do my best.

Aha, here you say
```Your current syntax is equivalent to how bash is dealing with wildcards, but will only work if users will specify parameters in parentheses.```

I suppose you mean quotes not parentheses. In my investigation I realized that if I leave the terminal language to parse for multiple files the commands will not be multiplatform. Yes, the user will have to write those quotes but if he does that makes the dipy/bin commands exactly the same to run across many platforms and no assumption is needed for bash or for operating system. That means that those can be run from Windows too. Which is an advantage for a library. I think it will be easy for them to learn to use quotes. And if they don't they can always use the nipype version or just run the commands with single files.

**Comment by chrisfilo** [comment]
A modified version of this solution: http://stackoverflow.com/a/12501893
would be a more robust way of dealing with the windows problem. This way
you can avoid quotes altogether and use nargs in argparse. We can include
this in nipype_cmd if it will make your life easier.

Best,
Chris

**Comment by Garyfallidis** [comment]
Cool! I need to get hands on with your suggestions. Will get back to you asap! :+1:

**Comment by Garyfallidis** [comment]
Hi just wanted to give a quick feedback here so that the others now how this goes. A few days ago I proposed a new design to @matthieudumont which will allow automatic generation of cmd interfaces and nipype interfaces from the workflow functions using introspection.  Mr. Dumont is currently working and extending this idea first for the cmd interfaces and after we have a first version of this I will also work on the objects that will generate the nipype interfaces automatically too. 
If you want a sneak peek  right here and right now here is an example of this idea implemented on the median_otsu interface
https://github.com/Garyfallidis/dipy/blob/introducing_workflows/bin/dipy_median_otsu
https://github.com/Garyfallidis/dipy/blob/introducing_workflows/dipy/workflows/base.py
https://github.com/Garyfallidis/dipy/blob/introducing_workflows/dipy/workflows/segment.py
Or you can wait until this PR is updated.

**Comment by Garyfallidis** [comment]
@matthieudumont there is some weird fail with python 3. Can you rebase and see if you continue getting the same error?

File: dipy/workflows/base.py [hunk 8]
```diff
@@ -0,0 +1,129 @@
+from dipy.fixes import argparse as arg
+import inspect
+
+
+try:
```
arokem: You might want to use the `optional_package` machinery here. For an example, see: https://github.com/nipy/dipy/blob/master/dipy/viz/fvtk.py#L32

dipy/workflows/base.py
```diff
@@ -0,0 +1,129 @@
...
+                dtype = self._select_dtype(typestr)
+                help_msg = ' '.join(self.doc[i][2])
+                self.add_argument('--' + args[i], action='store', type=dtype,
+                                  metavar=dtype.__name__, help=help_msg)
+            else:
+                self.add_argument('--' + args[i])
+
+    def _select_dtype(self, text):
+        text = text.lower()
+        if text.find('str') != -1:
```
arokem: Optionally:
    'str' in text

dipy/workflows/segment.py
```diff
@@ -0,0 +1,76 @@
...
+    """ Workflow wrapping the median_otsu segmentation method.
+
+    It applies median_otsu segmentation on each file found by 'globing'
+    ``input_files`` and saves the results in a directory specified by
+    ``out_dir``.
+
+    Parameters
+    ----------
+    input_files : string
+        Path to the input volumes. This path may contain wildcards to process
```
arokem: Only nifti files, right?

arokem:
Sorry - just a few things I noticed along the way. Looks like it's going in a good direction, with much less boilerplate code than before - that's good. Could you say a little bit more about how this also generates nipype workflows?

Garyfallidis:
Hi @arokem. The nipype generator is not yet in. But the idea is that now that we can get the exact input and output parameters and from the introspectiveparser we can generate python files with the InputSpects and OutputSpecs automatically written. Without you writing a single piece of code. Clear?

arokem:
Yeah. More or less. So the plan is to have some script that goes through
the workflows tree and generates the nipype code?

Garyfallidis:
Yes @arokem. And then we can do the same with GUIs, web interfaces etc. Let the flows generate the interfaces ;) It seems this will work pretty well. But lets not be overenthusiastic and see if it works in practice. But yeah things look very positive right now. It seems we can greatly speedup our development.

Hi @matthieudumont this looks great I would suggest to add a test function to test the IntrospectiveArgParser with a dummy workflow with a large range of different parameters where different action strategies would be required.

Also it would be nice to update the other bin/commands to the workflow design. So, some refactoring will be needed. The tensor_fit should probably change too because right now it is calling fsl's bet etc which is not relevant. I know you have a better version.

Finally, we need to investigate the use of wildcards along different operating systems and how we will deal with that. There were some suggestions from Chris on this. GGT!

arokem:
OK - that's cool. It seems to me that you should be able to generate the command line interfaces as well, just from introspection of the dipy.workflows module. That is `bin/dipy_median_otsu` should also possibly be auto-generated. Presumably they're all going to look the same?

dipy/segment/mask.py
```diff
@@ -8,7 +8,7 @@
 
 from scipy.ndimage.filters import median_filter
 try:
-    from skimage.filter import threshold_otsu as otsu
+    from skimage.filters import threshold_otsu as otsu
```
arokem: Is this a change in the skimage API, or a BF in dipy?

arokem:
And yes: +1 on thorough testing of the IAP!

dipy/workflows/segment.py
matthieudumont: I guess any format supported by Nibabel would do.

chrisfilo:
Looks promising! 

There is one thing I could not figure out from the code: how are you dealing with returning output paths and values? In other words how are the OutputSpecs going to be populated?

dipy/segment/mask.py
matthieudumont: no, idk why I modified and committed that sorry!

Garyfallidis: I did this change. There is a warning that skimage has deprecated skimage.filter and that now we should use skimage.filters.

arokem: Cool. That answers my question.

arokem:
Any progress on writing tests for this one?

Garyfallidis:
@chrisfilo good question. There will be an Outputs description after Parameters. That should do it. I will look into it. Or we just write the outputs manually.

dipy/workflows/base.py
```diff
@@ -0,0 +1,124 @@
...
+                self.add_argument(args[i])
+            cnt += 1
+
+        # Arguments with defaults (Optional)
+        for i in range(cnt, len_args):
+            if has_ndoc:
+                typestr = self.doc[i][1]
+                dtype = self._select_dtype(typestr)
+                help_msg = ' '.join(self.doc[i][2])
+                self.add_argument('--' + args[i], action='store', type=dtype,
```
Garyfallidis: There is a bug here when boolean is used. The argparser is not smart with bool but here is a suggested solution. You can specify 0 or 1. But let's talk about it.
```
                if 'bool' in typestr:
                    self.add_argument('--' + args[i], choices=[0, 1], type=int,
                                      metavar=dtype.__name__, help=help_msg)
                else:
                    self.add_argument('--' + args[i], action='store',
                                      type=dtype,
                                      metavar=dtype.__name__, help=help_msg)
```

dipy/workflows/base.py
```diff
@@ -0,0 +1,124 @@
+from dipy.fixes import argparse as arg
+from dipy.utils.optpkg import optional_package
+
+import inspect
+
+ndoc, has_ndoc, _ = optional_package('numpydoc')
```
Garyfallidis: @matthieudumont it seems we cannot avoid removing the numpydoc dependency. We tried to install dipy_recognize_bundles in another computer (osx yosemite) and it required both numpydoc and sphinx to be installed.
Can you copy only the parts that actually parse the docstring (especially Parameters section) and put them all in a file inside dipy/workflows. Let me know what you think.

Garyfallidis:
Ping!

Garyfallidis:
@matthieudumont any progress with this?

matthieudumont:
Will work on this asap. Sorry for the lack of updates, kind of busy recently.

Garyfallidis:
I see that you had progress with this @matthieudumont Is this ready for review?

matthieudumont:
Yes I think I fixed everything that was mentioned.

setup.py
```diff
@@ -212,7 +212,8 @@ def main(**extra_args):
                           'dipy.sims',
                           'dipy.sims.tests',
                           'dipy.denoise',
-                          'dipy.denoise.tests'],
+                          'dipy.denoise.tests',
+                          'dipy.workflows'],
```
arokem: I think you want to add `dipy.workflows.tests` in here, so that the tests also get installed

dipy/workflows/utils.py
```diff
@@ -0,0 +1,27 @@
...
+    return int(val) if val != 'None' else None
+
+def bool_param(val):
+    return (val == 'True') if val != 'None' else None
+
+def int_list_param(val, sep=' '):
+    if val == 'None':
+        return None
+
+    return val.split(sep)
```
arokem: They are very simple, but it would still be good to write unit tests for these utils.

arokem:
I haven't looked at this closely, but it looks like test coverage in iap is rather low: https://travis-ci.org/nipy/dipy/jobs/87939852#L2250

dipy/workflows/utils.py
matthieudumont: Actually, these were leftovers from the first draft we had with all parameters as string. They are now removed.

arokem: Aha! Great!

dipy/workflows/tests/test_iap.py
```diff
@@ -0,0 +1,89 @@
...
+    sys.argv.extend(inputs)
+    parser = IntrospectiveArgumentParser()
+    parser.add_workflow(dummy_flow)
+    args = parser.get_flow_args()
+    all_keys = pos_keys + opt_keys
+    all_results = pos_results+opt_results
+
+    # Test if types and order are respected
+    for k, v in zip(all_keys, all_results):
+        print k, v
```
arokem: This should be: 

   print(k, v) 

For python 3.

Better yet - you might want to remove this print statement altogether. I believe this is cruft from some round of debugging this?

chrisfilo:
I still cannot find any specification for the location of files generated by procedure (outputs). Without this it will be hard to integrate this into workflows that include other procedures (because a potential command using outputs of median_otsu_flow will not know where to find them).

dipy/workflows/tests/test_iap.py
matthieudumont: Yes I just forgot to remove it

matthieudumont:
Ok, the auto build fails on ./viz/tests/test_fvtk.py
I didnt change anything in that file, has someone ever had this error?

arokem:
Yes - if you rebase on master you should get the fix for that.

matthieudumont:
Thx, will do

dipy/workflows/base.py
```diff
@@ -0,0 +1,131 @@
...
+        self.add_argument(*args, **kargs)
+
+    def show_argument(self, dest):
+
+        for act in self._actions[1:]:
+            if act.dest == dest:
+                print(act)
+
+    def add_epilogue(self):
+        # with citations
```
Garyfallidis: Hola! Can you include the Frontiers Dipy paper as default epilogue here?
Garyfallidis, E., M. Brett, B. Amirbekian, A. Rokem, S. Van Der Walt, M. Descoteaux, and I. Nimmo-Smith. Dipy, a library for the analysis of diffusion MRI data. Frontiers in Neuroinformatics, 1-18, 2014.
So, if no papers are added in the epilogue or description, people should be at least citing Dipy.

dipy/workflows/documentation.py
```diff
@@ -0,0 +1,428 @@
+"""
```
Garyfallidis: I am suggesting to change the name of the file to docparser or docstring or docstring_parser. We should have another file in dipy/doc/ about documentation which should talk about the IAP and how it can be used from developers and from users. But this is for much later (in another PR) when this will start stabilizing.

Garyfallidis:
Hi @chrisfilo,

I had a meeting today with @matthieudumont in Sherbrooke about the outputs and other ideas to improve the dipy workflows. So, what we will do concerning nipype is that under ``Parameters`` in the docstring we will add a section called ``Outputs``. This will give you the output data types from the IAP so we can put them directly into the OutputSpecs class. The last thing missing will be a function which will create the filenames of output files given the input files. This will be something similar with ``_list_outputs`` that you have in this example. 

```
from nipype.interfaces.base import (
    TraitedSpec,
    CommandLineInputSpec,
    CommandLine,
    File
)
import os

class GZipInputSpec(CommandLineInputSpec):
    input_file = File(desc="File", exists=True, mandatory=True, argstr="%s")

class GZipOutputSpec(TraitedSpec):
    output_file = File(desc = "Zip file", exists = True)

class GZipTask(CommandLine):
    input_spec = GZipInputSpec
    output_spec = GZipOutputSpec
    cmd = 'gzip'

    def _list_outputs(self):
            outputs = self.output_spec().get()
            outputs['output_file'] = os.path.abspath(self.inputs.input_file + ".gz")
            return outputs

if __name__ == '__main__':

    zipper = GZipTask(input_file='an_existing_file')
    print zipper.cmdline
    zipper.run()
```
@matthieudumont has already added some similar function in ``workflows.utils``

Right now it looks like this 

```
def choose_create_out_dir(out_dir, root_path):
    if out_dir == '':
        result_path = dirname(root_path)
    elif not isabs(out_dir):
        result_path = join(dirname(root_path), out_dir)
        if not exists(result_path):
            makedirs(result_path)
    else:
        result_path = out_dir

    return result_path
```

We can try to make this more generic and have the same function used in as many workflows as possible.

I think that with proving the inputspecs and outputspecs from the docstrings and giving you also our internal ``list_outputs`` function will complete the design for having automatic nipype interfaces coming out from dipy workflows. I hope this clarification answers your question. Let me know otherwise. Of course we will need your feedback as we go ahead but it seems there is steady progress and we may have something basic ready by Christmas. At least for a few workflows. Thanks again for your feedback about this PR.

chrisfilo:
Yes this seems very reasonable.

I do, however, wonder if it wouldn't just be easier to adopt Nipypy
interface API to describe inputs and outputs instead of creating a new
standard (the Parameters in docstrings). It's an arbitrary design decision,
but it would've saved you some work.

Garyfallidis:
No, introspection which is used here is not an arbitrary design decision but solid software engineering practice. Perhaps it looks strange as it is not usually found in MRI-related software. For our user cases and with the current experience that we have making scripts the coding part is reduced a lot and the general design looks stable. But anyway time will tell. Worst case if the automatic nipype generator doesn't work well. It will still be trivial to wrap the flow functions in nipype. So, it's a win win situation.

Garyfallidis:
Sorry, I misunderstood your answer. Forget my last comment.

Garyfallidis:
@matthieudumont I may have found a small bug. 
When you call ``dipy_median_otsu -h`` do you see the correct names of the positional arguments? I get
```
positional arguments:
  str
```

dipy/workflows/tests/test_iap.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def nargs_flow(variable_ints, optionnal_int=2):
+    """ Workflow used to test the nargs argument.
+
+    Parameters
+    ----------
+    variable_ints : variable int
+        variable number of strings
+    optionnal_int : int
```
Garyfallidis: optional

dipy/workflows/tests/test_iap.py
```diff
@@ -0,0 +1,108 @@
...
+        optionnal int argument
+    """
+    return variable_ints, optionnal_int
+
+def test_nargs():
+    sys.argv = [sys.argv[0]]
+    var_args = ['1', '2', '3', '4', '5', '6', '7', '8']
+    optionnals = ['--optionnal_int', '2']
```
Garyfallidis: optional is with one 'n'

Garyfallidis:
Happy 2016 @matthieudumont I would like to merge this PR the following days.
Can you check my last comments, rebase and add supports for outputs section in the strings? I remember you had that ready in another branch. Let me know what you think. We need to crack on with the interfaces! I can't wait to start using this.

dipy/workflows/utils.py
arokem: They're still here. Did you mean to remove these?

arokem:
Some of the modules here are still only sparsely tested, according to the Travis coverage report: 

https://travis-ci.org/nipy/dipy/jobs/92975862

(e.g., `docstring_parser`). 

Also, please remove `utils.py`, if you don't want it here, or write tests for it, if you are using it.

matthieudumont:
hi everyone, thanks for the comments, sorry for the late answer I was on vacation last week. Ill look at your comments and make this ready for merge asap.

matthieudumont:
(huge changelog because I rebased)

Garyfallidis:
@arokem if you I agree I think I can merge this. 

Next PR should be a cleanup of what is in dipy/bin to actually use the new design. And after that or in parallel start putting in the workflows that @matthieudumont and I have already written.

arokem:
It seems that utils and segment are mostly untested. I think we should have
tests in place for these before we merge.

Garyfallidis:
Okay @matthieudumont added tests as requested. So, I will go ahead and merge this. But something very important to realize with the workflows is that we should have some buildbots which will run the tests of the command lines of the workflows with large and realistic datasets and create some reports. This is going towards a proper quality assurance direction where we do have Travis (and nipy.builders) that check that the internal functions (API) work correctly and then buildbots that check that a specific command line (workflow) or a series of command lines (workflows) generate the required results. This will be quite a task because we will need to generate some kind of QA reports or automated end-to-end tests. I think with the new dipy.viz module we will be able to have some of these things nicely done. But we do need to meet and discuss were we will set the buildbots. I think that @matthieudumont, @jchoude, @matthew-brett, @arokem and me should be in that meeting and share ideas on how to make this done. Can we have for the beginning a linux machine somewhere which will able to run the commands once day and generate some reports? Will the guys in Berkeley or Sherbrooke be willing to set a machine only for that? Or should we look into a cloud service?

PR #696, opened 2015-08-04
Change title of piesno example

samuelstjean:
The previous one was not really saying what it was about, so this one might attract more people to it.

### DIFF
--- a/doc/examples_index.rst
+++ b/doc/examples_index.rst
@@ -39,7 +39,7 @@ Basic SNR estimation
 
 - :ref:`example_snr_in_cc`
 
-PIESNO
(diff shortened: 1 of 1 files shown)


arokem:
What does "robust" mean in this context?

arokem:
Oh - I see this also in the paper. It's because the median is a "robust" statistic (relatively oblivious to outliers). Merging.

samuelstjean:
Exactly, since it's based on statistical estimation, it's deterministic, and the median is a statistic which is robust to outliers. Hence, it's more powerful than the other method, provided you respect the hypothesis of course.

PR #699, opened 2015-08-11
Added gaussian noise option to estimate_sigma

samuelstjean:
Since this is a case that can happen, and was the old behavior, I added an option to disable the correction factor, which is the default as it was before.

### DIFF
--- a/dipy/denoise/noise_estimate.py
+++ b/dipy/denoise/noise_estimate.py
@@ -282,7 +282,7 @@ def estimate_sigma(arr, disable_background_masking=False, N=0):
         Number of coils of the receiver array. Use N = 1 in case of a SENSE
         reconstruction (Philips scanners) or the number of coils for a GRAPPA
         reconstruction (Siemens and GE). Use 0 to disable the correction factor,
-        as for example is the noise is Gaussian distributed. See [1] for more information.
+        as for example if the noise is Gaussian distributed. See [1] for more information.
 
     Returns
     -------
--- a/dipy/denoise/noise_estimate.py
+++ b/dipy/denoise/noise_estimate.py
@@ -266,7 +266,7 @@ def _piesno_3D(data, N, alpha=0.01, l=100, itermax=100, eps=1e-5,
     return sigma[pos]
 
 
-def estimate_sigma(arr, disable_background_masking=False, N=1):
+def estimate_sigma(arr, disable_background_masking=False, N=0):
     """Standard deviation estimation from local patches
 
     Parameters
@@ -278,10 +278,11 @@ def estimate_sigma(arr, disable_background_masking=False, N=1):
         If True, uses all voxels for the estimation, otherwise, only non-zeros
         voxels are used. Useful if the background is masked by the scanner.
 
-    N : int, default 1
+    N : int, default 0
         Number of coils of the receiver array. Use N = 1 in case of a SENSE
         reconstruction (Philips scanners) or the number of coils for a GRAPPA
-        reconstruction (Siemens and GE). See [1] for more information.
+        reconstruction (Siemens and GE). Use 0 to disable the correction factor,
+        as for example is the noise is Gaussian distributed. See [1] for more information.
 
     Returns
     -------
@@ -320,7 +321,8 @@ def estimate_sigma(arr, disable_background_masking=False, N=1):
     k[1, 1, 2] = 1
 
     # Precomputed factor from Koay 2006, this corrects the bias of magnitude image
-    correction_factor = {1: 0.42920367320510366,
+    correction_factor = {0: 1 # No correction
+                         1: 0.42920367320510366,
                          4: 0.4834941393603609,
                          6: 0.4891759468548269,
                          8: 0.49195420135894175,
--- a/dipy/denoise/noise_estimate.py
+++ b/dipy/denoise/noise_estimate.py
@@ -321,7 +321,7 @@ def estimate_sigma(arr, disable_background_masking=False, N=0):
     k[1, 1, 2] = 1
 
     # Precomputed factor from Koay 2006, this corrects the bias of magnitude image
-    correction_factor = {0: 1 # No correction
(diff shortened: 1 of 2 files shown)


arokem:
Would be good to have a test that checks this mode.

arokem:
LGTM. Anyone else have any comments here? If not, I will merge this in a few days.

arokem:
Hey @mdesco - any opinions here? I don't know this stuff well enough, and could really use a second opinion.

samuelstjean:
For future reference, this merely restores the default behavior, but also keeps the suggested enhancements for the fmri people.

mdesco:
Not much to say. I asked @samuelstjean to put the N=0 option for the Gaussian noise case. The MGH group now has a paper to correct the phase and produce DWI with Gaussian-like profiles. Hence, the N=0 case can become important to have.

http://www.ncbi.nlm.nih.gov/pubmed/26241680

Garyfallidis:
@mdesco should that be the default one?

Garyfallidis:
Or N=1 is more likely?

mdesco:
To me, N=1 should be the default one, the rician case.

samuelstjean:
Depends, if you have fancy data, N=1 is probably advisable, N=0 is the previous default, and the one used in the paper. So to keep the original behavior, that is also the default. Since this is a data driven estimation, this keeps the spirit of the original version (well, actually it should do exactly as the original provided we did not introduce a bug or anything in the rewrite).

dipy/denoise/noise_estimate.py
```diff
@@ -278,10 +278,11 @@ def estimate_sigma(arr, disable_background_masking=False, N=1):
         If True, uses all voxels for the estimation, otherwise, only non-zeros
         voxels are used. Useful if the background is masked by the scanner.
 
-    N : int, default 1
+    N : int, default 0
         Number of coils of the receiver array. Use N = 1 in case of a SENSE
         reconstruction (Philips scanners) or the number of coils for a GRAPPA
-        reconstruction (Siemens and GE). See [1] for more information.
+        reconstruction (Siemens and GE). Use 0 to disable the correction factor,
+        as for example is the noise is Gaussian distributed. See [1] for more information.
```
arokem: => "Use 0 to disable the correction factor, for example if the noise is Gaussian distributed"

arokem:
Hi @samuelstjean - what do you mean by "fancy"? 2006's "fancy" is today's extreme low end.

If by "fancy" you mean a multi-coil or accelerated acquisition than I am with @mdesco on this: 1 should be the default.

rfdougherty:
I agree with default=1. I don't think any standard vendor sequence produces real data with Gaussian noise. 

Also, note that the current EPI diffusion sequences on GE use a SENSE-like recon (they call it ASSET) for inplane acceleration. (Siemens is currently the only one of the three that uses GRAPPA by default for inplane acceleration.) On GE, the appropriate N is 1 if ASSET (i.e. SENSE) is used, or num coils if ASSET is disabled (with ASSET disabled, coils are combined with a simple sum-of-squares algorithm). I suspect Philips would work the same (but can't say for sure). 

So the recommended settings need to be more nuanced: "For stock vendor sequences, you probably want to use N=num coils on Siemens. On GE and Philips, you probably want N=1 if inplane acceleration is enabled, otherwise use N=num coils. But it's best to confirm the appropriate N to use by inspecting your actual noise distribution."  (That disclaimer is essential, since any of this can change. E.g., GE is saying they'll release a GRAPPA-like option for dw-epi soon, Siemens may add SENSE coil combination, etc.) 

And maybe even add "If you use non-stock sequences, you should talk to the pulse sequence developers for advice." E.g., at our center, we do GRAPPA followed by SENSE coil combination (a.k.a. "SENSE1"), and phase correction to produce real data with zero-mean gaussian noise. So for us, N=0 would be appropriate, even though we use a GE scanner. (And as @mdesco notes, this will be more common given the recent paper out of MGH.)

samuelstjean:
That seems much more confusing than needed for anyone having no idea of what they are supposed to use. Plus, even if you ask for the appropriate answer, the number people who can help you in any given institute which might be able to help you can be counted on one hand (if they are cooperative and do exist in the first place).

I still vote for restoring the default option as it was before (N=0), and people needing it can still change the appropriate value from N=0 to whatever. Remember I'm only trying to shove something not belonging there in the original design for more accurate estimation, and people complained from that, so hence the "let's return to the default value".

samuelstjean:
By fancy I meant anything that deviates from the standard, stationary Rician/non central chi noise model, so your everyday acquisition would be a not fancy one while the HCP/multiband/etc. and the just released paper on phase smoothing for gaussian distributed noise is considered fancy in my book.

I think that people are confusing two un-related but still close issues here : the rician intensity correction bias from nlmeans and the second order moment correction factor (or variance correction for non gaussian data) from estimate_sigma.

This PR is about the second case, that is the old estimate_sigma estimator is a statistical estimator designed for gaussian noise and I tacked together a correction factor to account for the lower variance arising from not having a gaussian distribution. Of course this is not what the original version, so I just put back the original estimation instead of using that correction factor by default as orginally intended. This is also what the current release does.


Where I think the mix-up arises from, and that is totally untouched and still intact, is that nlmeans use the rician bias correction factor by default by substracting -2*variance to each voxel at the far end of the denoising process to correct the bias intensity. This is still the case and this is also what I think people wrongly refer to as N=1 should be default. You can look at the ~7 lines I changed to convince yourself of the good natured idea of the proposed change of course.

Garyfallidis:
Okay  @rfdougherty and @mdesco and all. If you do not disagree I would like to go ahead and merge this PR. Please let me know otherwise.

rfdougherty:
OK with me.

samuelstjean:
As for detecting the noise distribution, I have an idea on how to do that, maybe, I will need to play a bit with that beforehand, but it should be doable with some maths.

arokem:
@Garyfallidis - did you mean to merge this one?

PR #700, opened 2015-08-13
TEST: add tests for AffineMap

omarocegueda:
This PR just adds tests for the AffineMap class.

### DIFF
--- a/dipy/align/imaffine.py
+++ b/dipy/align/imaffine.py
@@ -147,7 +147,7 @@ class AffineMap(object):
         self.affine_inv = None
         if self.affine is None:
             return
-        if np.any(np.isnan(affine)) or np.any(np.isinf(affine)):
+        if not np.all(np.isfinite(affine))
             raise AffineInversionError('Affine contains invalid elements')
         try:
             self.affine_inv = npl.inv(affine)
--- a/dipy/align/imaffine.py
+++ b/dipy/align/imaffine.py
@@ -563,10 +563,6 @@ class MutualInformationMetric(object):
         H = self.histogram  # Shortcut to `self.histogram`
         grad = None  # Buffer to write the MI gradient into (if needed)
         if update_gradient:
-            # Re-allocate buffer for the gradient, if needed
-            n = params.shape[0]  # Number of parameters
-            if (self.metric_grad is None) or (self.metric_grad.shape[0] != n):
-                self.metric_grad = np.empty(n)
             grad = self.metric_grad
             # Compute the gradient of the joint PDF w.r.t. parameters
             if self.sampling_proportion is None:  # Dense case
--- a/dipy/align/imaffine.py
+++ b/dipy/align/imaffine.py
@@ -147,7 +147,7 @@ class AffineMap(object):
         self.affine_inv = None
         if self.affine is None:
             return
-        if not np.all(np.isfinite(affine))
+        if not np.all(np.isfinite(affine)):
             raise AffineInversionError('Affine contains invalid elements')
         try:
             self.affine_inv = npl.inv(affine)
--- a/dipy/align/imaffine.py
+++ b/dipy/align/imaffine.py
@@ -144,7 +144,7 @@ class AffineMap(object):
         if self.affine is None:
             self.affine_inv = None
             return
-        if np.any(np.isnan(affine)):
+        if np.any(np.isnan(affine)) or np.any(np.isinf(affine)):
             raise AffineInversionError('Affine contains invalid elements')
         try:
             self.affine_inv = npl.inv(affine)
@@ -223,8 +223,9 @@ class AffineMap(object):
         shape = np.array(sampling_grid_shape, dtype=np.int32)
 
         # Verify valid image dimension
-        if dim < 2 or dim > 3:
-            raise ValueError('Undefined transform for dimension: %d' % (dim,))
+        img_dim = len(image.shape)
+        if img_dim < 2 or img_dim > 3:
+            raise ValueError('Undefined transform for dim: %d' % (img_dim,))
 
         # Obtain grid-to-world transform for sampling grid
         if sampling_grid2world is None:
@@ -427,6 +428,8 @@ class MutualInformationMetric(object):
             interpolation artifacts. The default is None, implying no
             pre-alignment is performed.
         """
+        n = transform.get_number_of_parameters()
+        self.metric_grad = np.zeros(n, dtype=np.float64)
         self.dim = len(static.shape)
         if moving_grid2world is None:
             moving_grid2world = np.eye(self.dim + 1)
--- a/dipy/align/imaffine.py
+++ b/dipy/align/imaffine.py
@@ -132,6 +132,9 @@ class AffineMap(object):
     def set_affine(self, affine):
         """ Sets the affine transform (operating in physical space)
 
+        Also sets `self.affine_inv` - the inverse of `affine`, or None if
+        there is no inverse.
+
         Parameters
         ----------
         affine : array, shape (dim + 1, dim + 1)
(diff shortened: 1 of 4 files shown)


dipy/align/tests/test_imaffine.py
```diff
@@ -312,7 +320,232 @@ def test_mi_gradient():
...
+        enorm = npl.norm(expected)
+        anorm = npl.norm(actual)
         nprod = dp / (enorm * anorm)
         assert(nprod >= 0.99)
+
+
+def create_affine_transforms(dim, translations, rotations, scales, rot_axis=None):
+    r""" Creates a list of affine transforms with all combinations of params
+
+    This function is intended to be used for testing only. It generates
```
arokem: Why is this function only for testing? I can see this being useful elsewhere as well.

dipy/align/tests/test_imaffine.py
```diff
@@ -312,7 +320,232 @@ def test_mi_gradient():
...
+        for affine in gt_affines:
+
+            # make both domain point to the same physical region
+            # It's ok to use the same transform, we just want to test
+            # that this information is actually being considered
+            domain_grid2world = affine
+            codomain_grid2world = affine
+            grid2grid_transform = affine
+
+            # Evaluate the transform with vector_fields module (already tested)
```
arokem: What does it mean that it's "already tested". Does that mean that we treat this as ground truth here?

dipy/align/tests/test_imaffine.py
```diff
@@ -312,7 +320,232 @@ def test_mi_gradient():
...
+            assert_array_almost_equal(actual_nn, expected_nn)
+
+            # Test set_affine with valid matrix
+            affine_map.set_affine(affine)
+            if affine is None:
+                assert(affine_map.affine is None)
+                assert(affine_map.affine_inv is None)
+            else:
+                assert_array_equal(affine, affine_map.affine)
+                actual = affine_map.affine.dot(affine_map.affine_inv)
```
arokem: This is just to show that these properties really are the inverses of each other?

dipy/align/tests/test_vector_fields.py
```diff
@@ -798,13 +798,9 @@ def test_affine_transforms_2d():
     shape = np.array(codomain_shape, dtype=np.int32)
     # Exceptions from transform_2d
     assert_raises(ValueError, vfu.transform_2d_affine, circle, shape, invalid)
-    assert_raises(ValueError, vfu.transform_2d_affine, circle, shape, invalid)
-    assert_raises(ValueError, vfu.transform_2d_affine, circle, shape, invalid)
```
arokem: These were just repeated here? Curious.

dipy/align/vector_fields.pyx
```diff
@@ -19,6 +19,8 @@ def is_valid_affine(double[:, :] M, int dim):
         return False
     if np.any(np.isnan(M)):
         return False
+    if np.any(np.isinf(M)):
+        return False
```
arokem: Nice one. You could use `np.isfinite` to catch both of these.

dipy/align/tests/test_imaffine.py
omarocegueda: Oh! really!? I don't imagine an application other than generating some random transforms for testing. Where do you think we should put it? some `utilities` module?

dipy/align/tests/test_imaffine.py
omarocegueda: Yes, exactly, `transform_2d_affine` and `transform_3d_affine` have their own tests, which compare against scipy's `map_coordinates`. In this case, the test is much cleaner by using `transform_2d_affine` and `transform_3d_affine` as oracles instead of creating the full coordinate arrays to call `map_coordinates`.

dipy/align/tests/test_imaffine.py
omarocegueda: Yes. Too paranoid? =)

dipy/align/tests/test_vector_fields.py
omarocegueda: Yes... in Notepad++, Ctrl+d means "duplicate current line" and 'd' is just besides 's'... I guess I accidentally pressed 'd' (twice?! =S) attempting to save...

dipy/align/vector_fields.pyx
omarocegueda: Thanks! fixed.

dipy/align/tests/test_vector_fields.py
arokem: Notepad++?! You are a strange individual :-)

dipy/align/tests/test_imaffine.py
arokem: Yes - I would think something like that. Anyone else have strong opinions about this one?

arokem:
Except for the controversy around whether that one function belongs in the tests, or somewhere else, I am pleased with all this, but I have a hard time determining whether this covers all we need. Would be good if someone else with a knowledge of this part of the codebase (e.g., @matthew-brett ) took a look at this.

dipy/align/imaffine.py
```diff
@@ -140,10 +143,10 @@ def set_affine(self, affine):
             transformation.
         """
         self.affine = affine
+        self.affine_inv = None
         if self.affine is None:
-            self.affine_inv = None
             return
-        if np.any(np.isnan(affine)):
+        if np.any(np.isnan(affine)) or np.any(np.isinf(affine)):
```
matthew-brett: `if not np.all(np.isfinite(affine))` ?

dipy/align/imaffine.py
```diff
@@ -222,8 +225,9 @@ def _apply_transform(self, image, interp='linear', image_grid2world=None,
         shape = np.array(sampling_grid_shape, dtype=np.int32)
 
         # Verify valid image dimension
-        if dim < 2 or dim > 3:
-            raise ValueError('Undefined transform for dimension: %d' % (dim,))
+        img_dim = len(image.shape)
```
matthew-brett: Sorry to be lazy - so this was previously a bug right?  In that you were checking the output dims rather than the input dims?  Do the current tests check for this?

dipy/align/imaffine.py
omarocegueda: Fixed, thanks!

dipy/align/imaffine.py
omarocegueda: Yes, actually I found this bug while writing the tests for the exceptions. Here I verify that the exception is raised when attempting to transform an image of dimension 2 and 4:
https://github.com/nipy/dipy/pull/700/files#diff-5a512c70f85666010a0c8c3d13db8f3aR508

dipy/align/tests/test_vector_fields.py
omarocegueda: Thank you?... I guess... =P

dipy/align/tests/test_imaffine.py
matthew-brett: No strong opinion, but why not leave it in the test module for now, and see if anyone has use for it?

Garyfallidis:
@omarocegueda rebase to remove the rbf error. Then you have some other tests failing. Keep it up!

omarocegueda:
Rebased, thanks!

arokem:
This looks ready to merge. If no one objects, I will go ahead and do that some time next week.

omarocegueda:
Great! =) thanks Ariel!

omarocegueda:
Thanks Ariel! I'm so happy that buildbots didn't complain this time =)

arokem:
Yeah - it's like a small miracle when that happens...

PR #703, opened 2015-08-14
Fix syn-3d example

omarocegueda:
We recently renamed some utility functions (those that apply affine transforms) from 'warp...' to 'transform...', but did not update the tutorial appropriately. Here we use AffineMap instead of directly calling the utility function to make the code a bit cleaner.

### DIFF
--- a/dipy/align/imaffine.py
+++ b/dipy/align/imaffine.py
@@ -54,6 +54,7 @@ from .parzenhist import (ParzenJointHistogram,
                          compute_parzen_mi)
 from .imwarp import (get_direction_and_spacings, ScaleSpace)
 from .scalespace import IsotropicScaleSpace
+from warnings import warn
 
 _interp_options = ['nearest', 'linear']
 _transform_method = {}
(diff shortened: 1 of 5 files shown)


arokem:
I think that these changes break backwards compatibility with the 0.9 release. I think that we should transition more gently: put a deprecation warning for now, but keep the previous functionality, and make these names go away only on the 0.11 release. What do you think?

Garyfallidis:
It's a good procedure to keep backwards compatibility. But it may be an overkill here to worry about a single transform function introduced only in the previous release. So, my take is don't worry about backwards compatibility with this one. But definitely say in doc/api_changes.rst that this change happened.

Garyfallidis:
But also it doesn't cost much to have to remove this function in the 0.11 release. The problem with this specific function is that the name is misleading. It says warp although it is only about affine transformation. I wouldn't want people to get used to warp* terminology. So maybe the sooner we get rid of it the better it will be.

Garyfallidis:
For me this kind of like more like a BF here rather than RF. Good naming is crucial for an API.

arokem:
OK - agreed.

This does seem to be a fix, but maybe make a note in the docs, for (the
few) people who transition from 0.9 to 0.10 and have their code start
breaking...

Garyfallidis:
Sure! :)

arokem:
Hey @omarocegueda - just a gentle reminder to put a note in the docs (here: https://github.com/nipy/dipy/blob/master/doc/api_changes.rst), so that we can merge this.

omarocegueda:
Thank you Ariel! I had forgotten about this! I just added the note. Actually, this also reminds me about these other utility functions from the `imaffine` module:
`align_centers_of_mass`
`align_geometric_centers`
`align_origins`
These have not been released, we are ok. @Garyfallidis suggested to rename them to `transform_*` instead of `align_*`. These functions just return the affine that must be applied to the moving image to `align` their geometric centers, centers of mass and origins, respectively (they are not applying the transform, just returning the matrix). Do you guys agree to rename them to `transform_*`?

arokem:
I think that renaming is a good idea. Could we continue to support the `align_*` functions for the next release and raise a deprecation warning when the old name is used (could probably be done with a deprecation decorator of some sort?)? These are not really bugs, but really just mis-namings that should be corrected.

arokem:
LGTM. What do you think, @Garyfallidis ?

Garyfallidis:
Looks good. Thx!

Issue #688, opened 2015-07-27
dipy.test() fails on centos 6.x / python2.6


hassemlal:
Hi,
I tried running the dipy test suite, but it returns multiple failures (see below). I am running on centos 6.5 with python2.6 and virtualenv.

```
$ python -c "import dipy; dipy.test()"
.................Running unit tests for dipy
NumPy version 1.9.2
NumPy is installed in /home/hassemlal/dipy/venv/lib/python2.6/site-packages/numpy
Python version 2.6.6 (r266:84292, Jan 22 2014, 09:42:36) [GCC 4.4.7 20120313 (Red Hat 4.4.7-4)]
nose version 1.3.7

[clip]

.........SS.........
======================================================================
FAIL: dipy.core.tests.test_sphere.test_interp_rbf
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/hassemlal/dipy/venv/lib/python2.6/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/hassemlal/dipy/dipy/core/tests/test_sphere.py", line 374, in test_interp_rbf
    nt.assert_(np.mean(np.abs(interp_data_a - expected)) < 0.1)
  File "/home/hassemlal/dipy/venv/lib/python2.6/site-packages/numpy/testing/utils.py", line 53, in assert_
    raise AssertionError(smsg)  
AssertionError

======================================================================
FAIL: dipy.reconst.tests.test_csdeconv.test_odfdeconv
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/hassemlal/dipy/venv/lib/python2.6/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/hassemlal/dipy/dipy/reconst/tests/test_csdeconv.py", line 220, in test_odfdeconv
    assert_equal(len(w) > 0, False)
  File "/home/hassemlal/dipy/venv/lib/python2.6/site-packages/numpy/testing/utils.py", line 334, in assert_equal
    raise AssertionError(msg)   
AssertionError:
Items are not equal:
 ACTUAL: True
 DESIRED: False

======================================================================
FAIL: dipy.segment.tests.test_metric.test_metric_cosine
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/hassemlal/dipy/venv/lib/python2.6/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
[...]

PR #707, opened 2015-09-01
BF: Keep up with changes in scipy 0.16

arokem:
Recent changes in scipy use a different heuristic to find the epsilon
parameter used for RBF interpolation, causing for us issue #688. This
should fix it, by using the old heuristic.

### DIFF
--- a/dipy/core/sphere.py
+++ b/dipy/core/sphere.py
@@ -552,7 +552,7 @@ def interp_rbf(data, sphere_origin, sphere_target,
         w_s = "The Eucldian norm used for interpolation is inaccurate "
         w_s += "and will be deprecated in future versions. Please consider "
         w_s += "using the 'angle' norm instead"
-        warnings.warn(w_s)
+        warnings.warn(w_s, DeprecationWarning)
         norm = euclidean_norm
 
     # Workaround for bug in older versions of SciPy that don't allow
--- a/dipy/core/sphere.py
+++ b/dipy/core/sphere.py
@@ -552,16 +552,16 @@ def interp_rbf(data, sphere_origin, sphere_target,
         w_s = "The Eucldian norm used for interpolation is inaccurate "
         w_s += "and will be deprecated in future versions. Please consider "
         w_s += "using the 'angle' norm instead"
-        warnings.warning(w_s)
+        warnings.warn(w_s)
         norm = euclidean_norm
 
     # Workaround for bug in older versions of SciPy that don't allow
     # specification of epsilon None:
     if epsilon is not None:
         kwargs = {'function': function,
-                 'epsilon': epsilon,
-                 'smooth' : smooth,
-                 'norm' : norm}
+                  'epsilon': epsilon,
+                  'smooth' : smooth,
+                  'norm' : norm}
     else:
         kwargs = {'function': function,
                   'smooth': smooth,
--- a/dipy/core/sphere.py
+++ b/dipy/core/sphere.py
@@ -515,7 +515,8 @@ def interp_rbf(data, sphere_origin, sphere_target,
     function : {'multiquadric', 'inverse', 'gaussian'}
         Radial basis function.
     epsilon : float
-        Radial basis function spread parameter. Defaults to approximate average distance between nodes.
+        Radial basis function spread parameter. Defaults to approximate average
+        distance between nodes.
     a good start
     smooth : float
         values greater than zero increase the smoothness of the
--- a/dipy/core/sphere.py
+++ b/dipy/core/sphere.py
@@ -1,7 +1,5 @@
 from __future__ import division, print_function, absolute_import
 
-__all__ = ['Sphere', 'HemiSphere', 'faces_from_sphere_vertices', 'unique_edges']
-
 import numpy as np
 import warnings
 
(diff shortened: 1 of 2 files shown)


arokem:
Note - the first commit (https://github.com/arokem/dipy/commit/7c4e670deb798785c095655dc5813ca959039827) are the substantial changes. The rest is just PEP8 fixes that I did while I was in this file already.

arokem:
Oh dear. Hey @stefanv  - any ideas here? Looks pretty bad.

arokem:
Just to be clear on this one - these tests would have failed with previous versions of scipy as well (this PR basically just patches this up to the previous behavior, but adds more tests). It seems that RBF is just inaccurate in many use-cases, which we weren't thinking of before.

Garyfallidis:
Oy! This is strange. Does that mean that all the functionality with interpolating on sphere is now uncertain?

arokem:
Well - it was always uncertain. It's just that we now know for certain just
how uncertain it was...

Garyfallidis:
Okay so what is your suggestion. How to move forward with this?

arokem:
A few options (in order of preference, I think):

- Spend some time understanding RBF interpolation, and fix it up (might
need to fix the scipy implementation, see links to discussions there - IIUC
that epsilon parameter is pretty much made up there as well...).
- Put a big fat warning on RBF for users to check the results of
interpolation when they use this, and relax test requirements.
- Drop RBF interpolation altogether.

I can spend some time on option 1, but not before this Thursday. Even then,
progress might be slow. What do you think?

Garyfallidis:
Hi @arokem, 
here is some feedback from my investigations.

The problem at least at my scipy 0.14 seems to be only with norm="euclidean_norm"
(norm="angle" works fine) which can be resolved using a small smoothing (here using 0.1).

Check my code snippet here

```
def test_interp_rbf():
    def data_func(s, a, b):
        return a * np.cos(s.theta) + b * np.sin(s.phi)

    from dipy.core.sphere import Sphere, interp_rbf
    from dipy.core.subdivide_octahedron import create_unit_hemisphere
    import numpy as np

    s0 = create_unit_hemisphere(5)
    s1 = create_unit_hemisphere(6)

    from dipy.viz import fvtk
    s0 = Sphere(xyz = np.vstack((s0.vertices, -s0.vertices)))
    s1 = Sphere(xyz = np.vstack((s1.vertices, -s1.vertices)))


    for a, b in zip([1, 2, 0.5], [1, 0.5, 2]):
        ren = fvtk.ren()
        print(a, b)
        data = data_func(s0, a, b)
        expected = data_func(s1, a, b)
        interp_data_en = interp_rbf(data, s0, s1, smooth=.1, norm="euclidean_norm")
        interp_data_a = interp_rbf(data, s0, s1, smooth=.1, norm="angle")

        sf = np.zeros((3, s1.vertices.shape[0]))
        sf[0] = expected
        sf[1] = interp_data_en
        sf[2] = interp_data_a
        fvtk.add(ren, fvtk.sphere_funcs(sf, sphere=s1, norm=False, radial_scale=False))
        nt.assert_(np.mean(np.abs(interp_data_en - expected)) < 0.1)
        nt.assert_(np.mean(np.abs(interp_data_a - expected)) < 0.1)

        fvtk.show(ren)


if __name__ == "__main__":
    # nt.run_module_suite()
    test_interp_rbf()
```

image

Now if you change to the default smooth (zero) you get some weird result looking 

image

Some other important points to improving the tests. I think those tests should use higher sampled spheres. We were using create_unit_hemisphere(2) and (3) which provide only 9 and 33 points respectively. Also you will need to make sure that using hemispheres is a good idea for this type of interpolation. I am afraid that the points in the edges might create problems. So, I would use full spheres to be on the safe side. 

Finally, I noticed some another minor issue. In lines 550 and 552 we use ``is`` for comparing strings which is generally not recommended. I would change that to ``==``.

arokem:
So what do you think? Should we enforce smoothing per default? Might be a
safer way to go(?) And drop the Euclidean distance? In a way, it's not
surprising that doesn't work as well as angle - Euclidean is just an
approximation of angle in this case.

I've made the following changes:

- Smoothing is set to 0.1 per default.
- `norm` is set to `angle` per default. If a user sets it to
`euclidean_norm` a warning is issued. The euclidean will be completely
removed as an option in a subsequent release cycle. I've added a test to
see that this warning is raised.
- I've set the test to use spheres, instead of hemispheres (good call), and
these have more points than before.

arokem:
Thanks for all the feedback, by they way!

Garyfallidis:
Great! All good here. 
I think this is good as it is and it doesn't affect any other parts of the code. 

But put in your TODO list (after Thursday) to update the api_changes.rst file by saying that the default parameter has changed and some norms will be deprecated. You should keep an issue open for scipy 0.16 if necessary.

I will go ahead and merge this as it is now. That will make it easier to know which of the other PRs are failing for a good reason and not because of RBF. 

Oh and best of luck with your deadline. Thank you for looking into this.

stefanv:
I guess it's not too surprising that the Euclidean norm performed badly.  This looks like a reasonable solution!

PR #708, opened 2015-09-01
Check for bval dimensionality on read. 

### LINKED ISSUE #697: dipy.io.gradients read_bvals_bvecs does not check bvals length
Probably stumbled deep into old stuff, but here it is

~~~
from dipy.io.gradients import read_bvals_bvecs
bvals, bvecs = read_bvals_bvecs('bvecs_KS_HCP104_PmRS.txt', 'bvecs_KS_HCP104_PmRS.txt')

In [11]: np.all(bvals == bvecs)
Out[11]: True
~~~

Also, there is a typo in the docstring, 
~~~
    fbvecs : str
             path of file with b-vectorsl, or None if you don't want to read bvecs
~~~
Culprit seems to be the last lines of the function, where the check can be tricked.

~~~
    if max(bvals.shape) != max(bvecs.shape):
            raise IOError('b-values and b-vectors shapes do not correspond')
~~~

arokem:
This should fix #697. 

As with the previous one: the first two commits are fixes to the issue, and the last commit is a bit of cleanup (PEP8, etc).

### DIFF
--- a/dipy/io/gradients.py
+++ b/dipy/io/gradients.py
@@ -68,6 +68,9 @@ def read_bvals_bvecs(fbvals, fbvecs):
     if bvals is None:
         return bvals, bvecs
 
+    if len(bvals.shape) > 1:
+        raise IOError('bval file should have one row')
+
     if max(bvals.shape) != max(bvecs.shape):
             raise IOError('b-values and b-vectors shapes do not correspond')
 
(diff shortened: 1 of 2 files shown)


jchoude:
Had a quick look, seems good. I'm no Travis expert, so I'll let @arokem or someone else have a look at why Travis is complaining.

arokem:
Seems like it's (finally?) complaining about #688. Maybe we merge #707 first, and then rebase this one on top of that, just to be on the safe side.

arokem:
But it does look like all our PRs from now on are going to fail until we deal with #688. See also: https://travis-ci.org/arokem/dipy/jobs/78445866

PR #709, opened 2015-09-02
DOC: Use the `identity` variable in the resampling transformation.

arokem:
One-word fix (which is probably a no-op anyway), but would help clarify the example a tiny bit.


omarocegueda:
I think it's totally safe to merge this fix, the Travis failure is unrelated:
https://travis-ci.org/nipy/dipy/jobs/78445965#L2114
Merging this and opening #710 to keep track of the issue.

PR #714, opened 2015-09-18
RF: use masks in predictions and cross-validation

arokem:
This is an improvement on the previous API, which did not allow using masks for multi-voxel predictions, making predictions for cross-validation un-necessarily slow.

### DIFF
--- a/dipy/reconst/dti.py
+++ b/dipy/reconst/dti.py
@@ -1601,13 +1601,17 @@ def restore_fit_tensor(design_matrix, data, sigma=None, jac=True):
                     this_tensor, status = opt.leastsq(_nlls_err_func,
                                                       start_params,
                                                       args=(clean_design,
-                                                            clean_sig),
+                                                            clean_sig,
+                                                            'sigma',
+                                                            this_sigma),
                                                       Dfun=_nlls_jacobian_func)
                 else:
                     this_tensor, status = opt.leastsq(_nlls_err_func,
                                                       start_params,
                                                       args=(clean_design,
-                                                            clean_sig))
+                                                            clean_sig,
+                                                            'sigma',
+                                                            this_sigma))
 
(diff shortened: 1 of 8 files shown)


arokem:
Changed to WIP, while I try out a few more things.

arokem:
OK - this is ready for review. Note that many of the changes are PEP8 cleanups. 

If you are interested in the functional changes this is the relevant diff: 

https://github.com/nipy/dipy/compare/master...315c0be04e71860b31179eb2bc6eaabdcb42fcfe

arokem:
I think I've found the source of the previous test failures: the multi-voxel prediction function was allocating an `empty` array for the results, and this came up with a nan ever so often, in a random way. In general, `empty` sure is a dangerous function!

dipy/reconst/dti.py
```diff
@@ -1017,7 +1024,8 @@ def linearity(self):
 
         .. math::
 
-            Linearity = \frac{\lambda_1-\lambda_2}{\lambda_1+\lambda_2+\lambda_3}
+            Linearity =
+            \frac{\lambda_1-\lambda_2}{\lambda_1+\lambda_2+\lambda_3}
```
Garyfallidis: Question. Have you checked that breaking the line here renders fine?

Garyfallidis:
Okay this looks good to me. Can someone have another look?

dipy/reconst/dti.py
arokem: Yes - the docs render just fine. I have found a few typos in the meanwhile, so adding the fixes for that.

arokem:
This is now rebased on current master. @MrBago or @stefanv - any chance to get one of you to take a quick look?

stefanv:
This looks good to me, Ariel.

PR #719, opened 2015-09-23
Fixes #715

### LINKED ISSUE #715: In current segment_quickbundles tutorial there is no example for changing number of points
This makes it difficult for people who were used to the previous version of the algorithm. And they may not that the default is 12. I suggest to add the ResampleFeature in this first tutorial.

MarcCote:
Fixes #715, plus fixes some typos.

### DIFF
--- a/doc/examples/segment_quickbundles.py
+++ b/doc/examples/segment_quickbundles.py
@@ -84,7 +84,9 @@ print("Centroid of the last cluster:\n", clusters[-1].centroid)
            [  66.74534607,   86.00262451,   78.82582092],
            [  64.02451324,   88.43942261,   75.0697403 ]], dtype=float32)
 
+"""
 
+"""
 `clusters` has also attributes like `centroids` (cluster representatives), and
 methods like `add`, `remove`, and `clear` to modify the clustering result.
 


doc/examples/segment_quickbundles.py
```diff
@@ -36,10 +37,13 @@
...
 the same number of points, the clustering algorithm will internally use a
 representation of streamlines that have been automatically downsampled/upsampled
-so they have only 18 points (See tutorial### to manually set that number of points).
+so they have only 12 points (to manually set the number of points one can use
+the `ResampleFeature`).
 """
 
-qb = QuickBundles(threshold=10.)
+feature = ResampleFeature(nb_points=12)
+metric = AveragePointwiseEuclideanMetric(feature=feature)  # a.k.a. MDF
```
arokem: Regarding the comment at the end of this line: wouldn't it be better to change the references above to `MDF` to talk about this metric instead? It's not like 'MDF' is a very commonly used acronym.

arokem:
Yep - overall a good change to make, I think. With the one comment I made above about 'MDF'. My other comment is that it is now not entirely clear that this is emulating the default behavior, so you might want to point out which lines of code here would happen automatically, and what it would look like if you chose another setting (e.g. `nb_points=18`). One possibility is to do this with and without the `ResampleFeature` explicitly mentioned, and show the (small, I assume) differences.

arokem:
I see that this has some overlap with #682 ;-) 

@Garyfallidis - what do you think about these two? In particular, where do you stand on #682?

MarcCote:
Personally I prefer how `segment_quickbundles.py` tutorial is written in #682. That is to only show the default behaviour and mention in the comment above that the number of points can be changed. Then refer the user to the other tutorial that has more explanation.

Another reason I prefer not showing explicitly `ResampleFeature` here is that the main goal of this tutorial is to show how simple it is to use the plain `QuickBundles`. Also, the default behaviour should be alright for the common users.

@Garyfallidis when you say "this makes it difficult for people who were used to the previous version of the algorithm." are you only talking about you :P  (not a common user ;)) ? If you want we could make a function `quickbundles(streamlines, threshold, nb_points=12)` that returns a list of clusters (like the `cluster` method of a `QuickBundles` object). What do you think?

@arokem if this one gets in, I'll rebase to update #682 .

Garyfallidis:
Hi @MarcCote, can you remove lines 
+feature = ResampleFeature(nb_points=12)
+metric = AveragePointwiseEuclideanMetric(feature=feature)  # a.k.a. MDF
and I will merge the rest of your corrections.
Make sure this tutorial has a link to the next.

MarcCote:
This PR is now outdated since #682 is merged.

PR #723, opened 2015-10-01
WIP: Assign streamlines to an existing cluster map via QuickBundles

MarcCote:
When there are just too many streamlines, you might want to only cluster a subset of those then assign the rest to their nearest cluster. To do so, we needed a way to assign streamlines to an existing `cluster_map` object (i.e. what QuickBundles.cluster returns).

### DIFF
--- a/dipy/segment/benchmarks/bench_quickbundles.py
+++ b/dipy/segment/benchmarks/bench_quickbundles.py
@@ -92,3 +92,40 @@ def bench_quickbundles():
     assert_equal(len(clusters), expected_nb_clusters)
     assert_array_equal(sizes3, sizes1)
     assert_arrays_equal(indices3, indices1)
+
+
+def bench_quickbundles_assignation():
+    dtype = "float32"
+    repeat = 10
+    nb_points = 18
+
+    streams, hdr = nib.trackvis.read(get_data('fornix'))
+    fornix = [s[0].astype(dtype) for s in streams]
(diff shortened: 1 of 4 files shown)


Garyfallidis:
Hi @MarcCote can you please add an option for the assignments to be direct? Without updating the centroids? And/or update the centroids only at the end (if you think that this is useful)?

MarcCote:
@Garyfallidis  we have some decisions to make regarding this PR.
[] should streamlines assignation modifies the clusters i.e. updates the centroids?
[] when assigning streamlines to some clusters should we modified the `ClusterMap` object in-place?

I don't like the way streamlines assignation is done right now.
```python
qb = QuickBundles(threshold=12)
clusters = qb.cluster(streamlines)

# Use existing clusters to group `other_streamlines`
# Note: `new_clusters` won't contain any element from `streamlines`.
new_clusters = qb.assign(clusters, other_streamlines)
```

Maybe something like this, would be more user-friendly?
```python
qb = QuickBundles(threshold=12, metric=some_metric)
clusters = qb.cluster(streamlines)

# Use existing clusters to assign `other_streamlines` to existing `clusters`.
# Note: `new_clusters` *will* contain all elements from both `streamlines` and `other_streamlines`.
clusters.assign(streamlines, metric=some_metric)
```

Garyfallidis:
Let's go for option 1 and have an option to include initial streamlines or not.

Garyfallidis:
This PR is not necessary any more. There will be a new PR which speeds up QB a lot.

PR #724, opened 2015-10-02
@sinkpoint's power map - refactored

arokem:
Hi - here are a few suggestions on top of @sinkpoint's implementation of the Dell'Acqua power map (#672). I think that this would be very useful to have, but I have proposed a few changes relative to @sinkpoint's previous implementation that simplify the code and potentially lead to (marginally) faster performance. See here for a diff relative to #672 (ignore the top of this diff in `denspeed` - that has to do with the fact that #672 needs a rebase): 

https://github.com/sinkpoint/dipy/pull/1/files


arokem:
OK - here's what I've come up with: I am now testing this with a very simplified set of coefficients, that are all equal to 1. In this case, it is easy to calculate what AP should be equal to: it's the log of the number of even coefficients (because in this case the 2l+1 factors cancel out). Then, the normalization factor has to be taken into account, which I do here by varying that between two different quantities. 

This seems like a better solution to me, because it tests the equation, rather than testing some other implementation of the same equation, on a specific test-set. I wouldn't mind having both in here, but I think this is sufficient.

dipy/reconst/shm.py
```diff
@@ -965,3 +966,111 @@ def sh_to_sf_matrix(sphere, sh_order, basis_type=None, return_inv=True, smooth=0
         return B.T, invB.T
 
     return B.T
+
+
+def calculate_max_order(n_coeffs):
+        """
+        Calculate the maximal harmonic order, given that you know the
```
Garyfallidis: push one line up

dipy/reconst/tests/test_shm.py
```diff
@@ -419,6 +418,37 @@ def test_faster_sph_harm():
 
     assert_array_almost_equal(sh, sh2, 8)
 
+def test_anisotropic_power():
+    for n_coeffs in [6, 15, 28, 45, 66, 91]:
+        for norm_factor in [0.0005, 0.00001]:
+
+            # Create some really simple case:
```
Garyfallidis: typo: cases

dipy/reconst/tests/test_shm.py
```diff
@@ -419,6 +418,37 @@ def test_faster_sph_harm():
...
+            assert_array_almost_equal(apvals, answers)
+            # Test that this works for single voxel arrays as well:
+            assert_array_almost_equal(
+                anisotropic_power(coeffs[1], norm_factor=norm_factor),
+                answers[1])
+
+
+def test_calculate_max_order():
+    """
+    Based on the table in:
```
Garyfallidis: push one line up please

Garyfallidis:
Good job! :+1:  Many thanks to @sinkpoint, @arokem and @deflavio!

PR #725, opened 2015-10-02
WIP: TST: Install vtk on travis with conda.

arokem:
This is a PR to help #587 along. To test that one, we need vtk installed. Hopefully this does it. 

After the initial installation, I will try to see whether we can also get a virtual frame-buffer going.


arokem:
Will try again on a different branch

PR #727, opened 2015-10-07
Optimize tensor fitting

dimrozakis:
Speed up certain tensor fitting functions by vectorizing calculations.

### DIFF
--- a/dipy/reconst/dti.py
+++ b/dipy/reconst/dti.py
@@ -55,7 +55,7 @@ def _min_positive_signal(data):
     data: array ([X, Y, Z, ...], g)
         Data or response variables holding the data. Note that the last
         dimension should contain the data.
-    
+
     Returns
     -------
     min_signal : float
@@ -1239,15 +1239,6 @@ def _wls_iter(ols_fit, design_matrix, sig, min_diffusivity):
     return decompose_tensor(tensor, min_diffusivity=min_diffusivity)
 
 
-def _ols_iter(inv_design, sig, min_diffusivity):
-    ''' Helper function used by ols_fit_tensor.
-    '''
-    log_s = np.log(sig)
-    D = np.dot(inv_design, log_s)
-    tensor = from_lower_triangular(D)
-    return decompose_tensor(tensor, min_diffusivity=min_diffusivity)
-
-
 def ols_fit_tensor(design_matrix, data):
     r"""
     Computes ordinary least squares (OLS) fit to calculate self-diffusion
@@ -1293,27 +1284,23 @@ def ols_fit_tensor(design_matrix, data):
     """
     tol = 1e-6
 
-    data = np.asarray(data)
-    data_flat = data.reshape((-1, data.shape[-1]))
-    evals = np.empty((len(data_flat), 3))
-    evecs = np.empty((len(data_flat), 3, 3))
-    dti_params = np.empty((len(data_flat), 4, 3))
-
     #obtain OLS fitting matrix
     #U,S,V = np.linalg.svd(design_matrix, False)
     #math: beta_ols = inv(X.T*X)*X.T*y
     #math: ols_fit = X*beta_ols*inv(y)
     #ols_fit =  np.dot(U, U.T)
 
-    min_diffusivity = tol / -design_matrix.min()
-    inv_design = np.linalg.pinv(design_matrix)
+    evals, evecs = decompose_tensor(
+        from_lower_triangular(
+            np.einsum('...ij,...j',
+                      np.linalg.pinv(design_matrix),
+                      np.log(np.asarray(data)))
+        ),
+        min_diffusivity=tol / -design_matrix.min()
+    )
 
-    for param, sig in zip(dti_params, data_flat):
-        param[0], param[1:] = _ols_iter(inv_design, sig, min_diffusivity)
-
-    dti_params.shape = data.shape[:-1] + (12,)
-    dti_params = dti_params
-    return dti_params
+    dti_params = np.concatenate((evals[..., None, :], evecs), axis=-2)
+    return dti_params.reshape(data.shape[:-1] + (12,))
 
 
 def _ols_fit_matrix(design_matrix):
--- a/dipy/reconst/dti.py
+++ b/dipy/reconst/dti.py
@@ -1722,7 +1722,7 @@ def decompose_tensor(tensor, min_diffusivity=0):
 
     Parameters
     ----------
-    tensor : array (3, 3)
+    tensor : array (..., 3, 3)
         Hermitian matrix representing a diffusion tensor.
     min_diffusivity : float
         Because negative eigenvalues are not physical and small eigenvalues,
@@ -1732,22 +1732,37 @@ def decompose_tensor(tensor, min_diffusivity=0):
 
     Returns
     -------
-    eigvals : array (3,)
+    eigvals : array (..., 3)
         Eigenvalues from eigen decomposition of the tensor. Negative
         eigenvalues are replaced by zero. Sorted from largest to smallest.
-    eigvecs : array (3, 3)
+    eigvecs : array (..., 3, 3)
         Associated eigenvectors from eigen decomposition of the tensor.
-        Eigenvectors are columnar (e.g. eigvecs[:,j] is associated with
-        eigvals[j])
+        Eigenvectors are columnar (e.g. eigvecs[..., :, j] is associated with
+        eigvals[..., j])
 
     """
     #outputs multiplicity as well so need to unique
     eigenvals, eigenvecs = np.linalg.eigh(tensor)
 
     #need to sort the eigenvalues and associated eigenvectors
-    order = eigenvals.argsort()[::-1]
-    eigenvecs = eigenvecs[:, order]
-    eigenvals = eigenvals[order]
+    if eigenvals.ndim == 1:
+        # this is a lot faster when dealing with a single voxel
+        order = eigenvals.argsort()[::-1]
+        eigenvecs = eigenvecs[:, order]
+        eigenvals = eigenvals[order]
+    else:
+        # temporarily flatten eigenvals and eigenvecs to make sorting easier
+        shape = eigenvals.shape[:-1]
+        eigenvals = eigenvals.reshape(-1, 3)
+        eigenvecs = eigenvecs.reshape(-1, 3, 3)
+        size = eigenvals.shape[0]
+        order = eigenvals.argsort()[:, ::-1]
+        xi, yi = np.ogrid[:size, :3, :3][:2]
+        eigenvecs = eigenvecs[xi, yi, order[:, None, :]]
+        xi = np.ogrid[:size, :3][0]
+        eigenvals = eigenvals[xi, order]
+        eigenvecs = eigenvecs.reshape(shape + (3, 3))
+        eigenvals = eigenvals.reshape(shape + (3, ))
 
     eigenvals = eigenvals.clip(min=min_diffusivity)
     # eigenvecs: each vector is columnar
--- a/dipy/reconst/dti.py
+++ b/dipy/reconst/dti.py
@@ -877,6 +877,11 @@ class TensorFit(object):
         return fractional_anisotropy(self.evals)
 
     @auto_attr
+    def color_fa(self):
+        """Color fractional anisotropy of diffusion tensor"""
+        return color_fa(self.fa, self.evecs)
+
+    @auto_attr
     def ga(self):
         """Geodesic anisotropy (GA) calculated from cached eigenvalues."""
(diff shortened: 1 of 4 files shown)


Garyfallidis:
Hi @dimrozakis and thank you for your contribution. Can you give us an idea of what is the actual speedup before and after your change?

dimrozakis:
With a given set of dMRI dicom files (128, 128, 38, 39) I get the following measurements in master:

```
OLS fitting completed in 40.53 secs.
WLS fitting completed in 135.74 secs.
```

and the following in the branch of this PR (diff measurements is the absolute difference between tensor calculated in the master and the PR branch for the same exam with the same fitting method):

```
OLS fitting completed in 4.72 secs.
Diff min/mean/max: 0.0, 2.0502642549085253e-19, 1.474514954580286e-17
WLS fitting completed in 45.56 secs.
Diff min/mean/max: 0.0, 1.2172205038436504e-16, 7.6679114446864816e-15
```

Garyfallidis:
Oh nice! :) That's a good speedup :+1:

Garyfallidis:
The travis bot with numpy 1.6.0 and scipy 0.9.0 is failing. Could it be that some of the vectorization that you are using is available with more recent numpy versions?
https://travis-ci.org/nipy/dipy/builds/84175253

dimrozakis:
This:
```
  File "/home/travis/build/nipy/dipy/venv/lib/python2.7/site-packages/dipy/utils/arrfuncs.py", line 38, in pinv_vec

    u, s, v = np.linalg.svd(a, full_matrices=False)

  File "/home/travis/build/nipy/dipy/venv/lib/python2.7/site-packages/numpy/linalg/linalg.py", line 1271, in svd

    _assertRank2(a)

  File "/home/travis/build/nipy/dipy/venv/lib/python2.7/site-packages/numpy/linalg/linalg.py", line 155, in _assertRank2

    two-dimensional' % len(a.shape)

numpy.linalg.linalg.LinAlgError: 3-dimensional array given. Array must be two-dimensional
```

seems to indicate that `numpy.linalg.svd` can't accept arrays with more than 2 dimensions in that numpy version. @Garyfallidis how do you suggest we handle this?

dimrozakis:
I guess an option would to either set a condition based on installed numpy version or catch a `numpy.linalg.LinAlgError` exception and fall back to iterating `numpy.linalg.svd` over each 2D array contained in the ND array. I don't think that trying to implement a vectorized version of `numpy.linalg.svd` in the code base of dipy just so it can work properly with numpy 1.6.0 would be a great idea (or a practical one). @Garyfallidis, your thoughts?

Garyfallidis:
What you suggest seems reasonable (two first options). Because the dti part is a core part of the reconstruction module it would be great to hear what other people think about this change. Also we should look into using the svd part from scipy not numpy maybe this was already resolved there (just brainstorming here). Finally, maybe this vectorization strategy will be interesting in other parts of Dipy too. More tomorrow... keep hacking :)

dimrozakis:
Finally all tests pass!

The main issue was that both `numpy.linalg.eigh` and `numpy.linalg.svd` where vectorized (able to run on arrays with more than 2 dimensions) in numpy version 1.8.0. I worked around this by iterating them for every nested 2D array, if installed numpy version is less than 1.8.0. Obviously in that case there's no increase in speed. @Garyfallidis I looked into the scipy implementations but it seems that none of [scipy.linalg.svd](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.svd.html), [scipy.linalg.pinv](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.pinv.html) or [scipy.linalg.pinv2](https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.pinv2.html) are vectorized.

Garyfallidis:
Bravo Dimitri! :) Can the other core devs have a look at this and give some thumbs up? @MrBago, @arokem ?

arokem:
Will look, but probably not before early next week!

Garyfallidis:
Okay if @MrBago can help here and give me some thumbs up I can go ahead and merge.

dipy/reconst/dti.py
```diff
@@ -740,11 +741,12 @@ def __init__(self, gtab, fit_method="WLS", *args, **kwargs):
 
         if not callable(fit_method):
             try:
-                self.fit_method = common_fit_methods[fit_method]
+                fit_method = common_fit_methods[fit_method]
             except KeyError:
                 raise ValueError('"' + str(fit_method) + '" is not a known fit '
                                  'method, the fit method should either be a '
                                  'function or one of the common fit methods')
+        self.fit_method = fit_method
```
arokem: Nice catch! Would you mind adding a test that fails without this change, so that we don't end up reverting without noticing?

dipy/utils/arrfuncs.py
```diff
@@ -28,3 +28,45 @@ def as_native_array(arr):
         return arr
     return arr.byteswap().newbyteorder()
 
+
+def pinv_vec(a, rcond=1e-15):
+    """Vectorized version of numpy.linalg.pinv"""
+
+    a = np.asarray(a)
+    if np.version.version.split('.') < ['1', '8']:
```
arokem: Use `LooseVersion` to compare versions. Like here: https://github.com/nipy/dipy/blob/master/dipy/core/optimize.py#L14

dipy/utils/arrfuncs.py
```diff
@@ -28,3 +28,45 @@ def as_native_array(arr):
         return arr
     return arr.byteswap().newbyteorder()
 
+
+def pinv_vec(a, rcond=1e-15):
+    """Vectorized version of numpy.linalg.pinv"""
```
arokem: Documentation should follow the numpy docstring standard: https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt

dipy/utils/arrfuncs.py
```diff
@@ -28,3 +28,45 @@ def as_native_array(arr):
...
+    mask = s > cutoff
+    s[mask] = 1. / s[mask]
+    s[~mask] = 0
+    return np.einsum('...ij,...jk',
+                     np.transpose(v, swap) * s[..., None, :],
+                     np.transpose(u, swap))
+
+
+def eigh(a, UPLO='L'):
+    """Iterate over eigh if it doesn't support vectorized operation"""
```
arokem: Same here about the documentation - please follow the numpy docstring standard (Parameters, Returns, etc.)

dipy/utils/arrfuncs.py
```diff
@@ -28,3 +28,45 @@ def as_native_array(arr):
...
+    a = np.asarray(a)
+    if a.ndim > 2 and np.version.version.split('.') < ['1', '8']:
```
arokem: Use LooseVersion here as well.

dipy/reconst/dti.py
```diff
@@ -1151,6 +1158,28 @@ def predict(self, gtab, S0=1):
         return tensor_prediction(self.model_params[0:12], gtab, S0=S0)
 
 
+def tensor6_to_dtiparams(tensor, min_diffusivity=0):
```
arokem: This function is already implemented as `eig_from_lo_tri`

arokem: But it might be worth refactoring that function. For example, it currently does not take into account `min_diffusivitiy`

dipy/utils/arrfuncs.py
```diff
@@ -28,3 +28,45 @@ def as_native_array(arr):
...
+    if np.version.version.split('.') < ['1', '8']:
+        # numpy 1.8.0 introduced a vectorized version of np.linalg.svd,
+        # if using older numpy fall back to using non vectorized np.linalg.pinv
+        shape = a.shape[:-2]
+        a = a.reshape(-1, a.shape[-2], a.shape[-1])
+        result = np.empty((a.shape[0], a.shape[2], a.shape[1]))
+        for i, item in enumerate(a):
+            result[i] = np.linalg.pinv(item, rcond)
+        return result.reshape(shape + (a.shape[2], a.shape[1]))
+
```
arokem: For clarity, you might want to put the following in an `else` clause

arokem:
Very nice! Just a few comments.

dimrozakis:
@arokem thanks for the comments, I'll look into them.

Btw, I just noticed that both `ols_fit_tensor` and `wls_fit_tensor` claim in their docstrings to be returning an `evals, evecs` tuple, whereas in fact they return a single array with its last dimension equal to 12 (that's true both for the master branch and this PR). Is it the return type or the docstring that should change so that they both match? Perhaps have them return a 2-tuple and have TensorModel/TensorFit concatenate them if given as a 2-tuple?

dimrozakis:
@Garyfallidis, @arokem something is causing the tests for some python versions to fail. I thought it was me after pushing new commits, but that doesn't seem to be the case. After several attempts, I have ended up with the following situation in this branch: The second last commit https://github.com/nipy/dipy/commit/23a62bebe3674e7b25c84c6aa4b6cb5d66faa560 (3 days ago) passes all tests. The last commit (https://github.com/nipy/dipy/commit/2cd37e0fa56d44b90883be4d7620dbe2d72bb5b2) however, which simply touches a file, fails most tests. For the failing tests, there's a dtype error. Any idea as to what may be causing this?

arokem:
I am not sure this has anything to do with it, but did you rebase on
master? By the way - you don't need to make a commit to trigger the testing
on travis. There's a "rerun" button there.

arokem:
This might also have to do with new version of numpy. Testing locally with this new version to see if I get this as well.

dimrozakis:
@arokem no, I didn't rebase on master. Should I try that or would you prefer to leave it as is in order to track down the issue with the tests? Also, it may be cause it's getting late, but I can't seem to find a 'rerun' button in travis, could it be it's only accessible by accounts with write access to the repo? I thought about opening a dummy PR to trigger tests on the master branch, but didn't want to spam everybody.

arokem:
Might be only accessible to some users - I set it to rerun all the failed
runs. I'll try to get it to run one on master as well.

arokem:
But it did run one of those just six hours ago:
https://travis-ci.org/nipy/dipy/builds/84939747, so I wonder.

arokem:
Yep - new numpy - new troubles...

I'm getting this locally on my machine as well.

arokem:
See https://github.com/nipy/dipy/issues/730 and https://github.com/nipy/dipy/pull/731

Garyfallidis:
@dimrozakis please rebase when you get back to this. Thx!

arokem:
If you could also squash away that last commit (the dummy file), that would be great.

arokem:
Sorry - wrong button :-)

dimrozakis:
I have rebased on master and I think I have addressed all the issues raised in the comments above. @arokem care to take another look?

arokem:
Yep - looks good to me. If no one objects, I will merge this in some time this weekend. Thanks for the nice contribution!

dipy/utils/arrfuncs.py
```diff
@@ -28,3 +32,97 @@ def as_native_array(arr):
...
+def pinv(a, rcond=1e-15):
+    """Vectorized version of `numpy.linalg.pinv`
+
+    If numpy version is less than 1.8, it falls back to iterating over
+    `np.linalg.pinv` since there isn't a vectorized version of `np.linalg.svd`
+    available.
+
+    Parameters
+    ----------
+    a : (..., M, N) array_like
```
Garyfallidis: Just a stylistic issue here. In Dipy we first write array_like and then the shape. Please change everywhere.

dipy/utils/tests/test_arrfuncs.py
```diff
@@ -28,3 +28,30 @@ def test_as_native():
...
     assert_equal(narr.dtype.byteorder, NATIVE_ORDER)
 
+
+def test_pinv():
+    arr = np.random.randn(4, 4, 4, 3, 7)
+    _pinv = pinv(arr)
+    for i in range(4):
+        for j in range(4):
+            for k in range(4):
+                assert_array_almost_equal(_pinv[i, j, k],
```
Garyfallidis: Cool! :+1: Good test!

Garyfallidis:
Apart from this minor stylistic change this is ready to be merged.

dimrozakis:
>Just a stylistic issue here. In Dipy we first write array_like and then the shape. Please change everywhere.

@Garyfallidis I fixed this in 6ef3d4ba997832a2f2b27ea84ec2cec935c3df97.

Garyfallidis:
Bravo @dimrozakis ! Your PR is merged. Thank you for discovering and contributing to DIPY. And welcome! :+1:  Hopefully see you in Athens ;)

PR #731, opened 2015-10-12
Np 1.10

arokem:
See https://github.com/nipy/dipy/issues/730


Garyfallidis:
@arokem I can go ahead and merge this. Agree?

arokem:
yup

samuelstjean:
So how about #617 after all? If any other thing breaks because of dependencies, everyone will know to not clone the latest master this way.

PR #732, opened 2015-10-14
WIP: Test with vtk on Travis

arokem:


arokem:
Trying this out again...

.travis.yml
```diff
@@ -31,6 +31,7 @@ matrix:
       env:
         - DEPENDS="cython numpy scipy matplotlib h5py nibabel cvxopt scikit_learn"
 before_install:
+    - apt-get update && apt-get install -y python-vtk
     - source tools/travis_tools.sh
     - virtualenv --python=python venv
```
matthew-brett: I think you'll need `virtualenv --system-site-packages --python=python venv`

arokem: Will this allow me to install with apt-get?

.travis.yml
```diff
@@ -31,8 +31,9 @@ matrix:
       env:
         - DEPENDS="cython numpy scipy matplotlib h5py nibabel cvxopt scikit_learn"
 before_install:
+    - apt-get update && apt-get install -y python-vtk
```
matthew-brett: I believe you'll need sudo for the apt-get.  I think you can use sudo for selected builds - but at the moment the global setting is `sudo: false` - which makes the builds faster.   If you want `python-vtk` for all builds, you can add it to the `apt` package list - e.g 

```
addons:
  apt:
    packages:
      - python-vtk
```

See the matplotlib `.travis.yml` file for an example.

arokem: Yes - I am just reading through the docs about this.

Sadly, it seems that we will also need to ask for travis to white-list vtk:
https://github.com/travis-ci/apt-source-whitelist

matthew-brett: It's a good idea to do that, but maybe using sudo for one build would be enough for now.

matthew-brett:
Try adding a new matrix entry:

```
matrix:
   include:
     - python: 2.7
       sudo: true
       env:
         - VTK=1
```

then lower down:

```
- if [ "${VTK}" == "1" ]; then
       sudo apt-get update && apt-get install -y python-vtk;
       toggleglobalsitepackages;
       fi
```

arokem:
Thanks! Trying that out. We'll still also need to get xvfb running here...

arokem:
What's the `toggleglobalsitepackages;` business all about. Travis didn't seem to take to that at all.

matthew-brett:
Ah, yes, sorry, that's a `virtualenvwrapper` feature. 

I think you'll have to do something like:

```
matrix:
   include:
     - python: 2.7
       sudo: true
       env:
         - VTK=1
         - VENV_ARGS='--system-site-packages'
```

and

```
    - virtualenv $VENV_ARGS --python=python venv
```

arokem:
Hrrm. That's already there:

https://github.com/arokem/dipy/blob/test-vtk/.travis.yml#L39

arokem:
Well - I can get VTK to install, and I can't get the tests to run,
seemingly with xvfb. The only remaining piece in the puzzle (I hope!) is to
get vtk to actually import within this venv. Otherwise, it's back to conda,
possibly?

arokem:
Hey @matthew-brett - do you understand why vtk can't be imported in this build: https://travis-ci.org/arokem/dipy/jobs/85562441 ? It's based on this commit: https://github.com/arokem/dipy/commit/264348d6192b9a16393d98b9b0bdfcf336f5c905 

Am I calling virtualenvwrapper the right way there? 

It seems to run fine, but still vtk can't be imported.

matthew-brett:
The toggle stuff is only get system site packages on the path - which should be covered by your virtualenv creation.

Why not replace everything after the virtualenv creation with `python -c 'import vtk'` to see what's going on?

matthew-brett:
Sorry - specifically I don't think it's useful to install virtualenvwrapper - sorry for the red herring there.

arokem:
But I do think it's about the availability of vtk to be imported.

See: https://travis-ci.org/nipy/dipy/jobs/85562784

VTK gets installed, and is immediately afterwards unavailable to import.

Maybe if we knew where this is installed, we could symlink from within the
venv, or add that to the venv PYTHONPATH?

matthew-brett:
I see the problem - the virtualenv isn't using the system Python, which is where vtk will appear:

Failing thing showing that import vtk doesn't work : 

Succeeding thing showing that import vtk does work: https://travis-ci.org/matthew-brett/dipy/jobs/85600918

Travis.yml : https://github.com/matthew-brett/dipy/blob/c51cccb761ee4b3c07b8c8a897d67a9318003c41/.travis.yml

arokem:
Thanks! I just pushed that change. Let's see what the golem says now.

arokem:
Yep - it's now testing with VTK!

Garyfallidis:
Super!

PR #733, opened 2015-10-14
Add RTOP, RTAP and RTPP and the relative test

maurozucchelli:
This is the second part of the MAPMRI pull request, which add RTOP, RTAP and RTPP, three microstructural EAP indices introduced in  Ozarslan E. et. al, "Mean apparent propagator (MAP) MRI: a novel diffusion imaging method for mapping tissue microstructure", NeuroImage, 2013.

p.s. The example is still missing, which dataset do you think I should use? The DSI data that I use as an example for the SHORE? Or do we have better multi-shell data?

### DIFF
--- a/dipy/reconst/mapmri.py
+++ b/dipy/reconst/mapmri.py
@@ -10,6 +10,7 @@ from ..utils.optpkg import optional_package
 
 cvxopt, have_cvxopt, _ = optional_package("cvxopt")
 
+
 class MapmriModel(ReconstModel):
 
     r"""Mean Apparent Propagator MRI (MAPMRI) [1]_ of the diffusion signal.
@@ -167,7 +168,7 @@ class MapmriModel(ReconstModel):
         if self.eap_cons:
             if not have_cvxopt:
                 raise ValueError(
-                'CVXOPT package needed to enforce constraints')
+                    'CVXOPT package needed to enforce constraints')
             w_s = "The implementation of MAPMRI depends on CVXOPT "
             w_s += " (http://cvxopt.org/). This software is licensed "
             w_s += "under the GPL (see: http://cvxopt.org/copyright.html) "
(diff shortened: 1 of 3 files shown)


dipy/reconst/mapmri.py
```diff
@@ -269,6 +270,65 @@ def odf(self, sphere, s=0):
...
+        References
+        ----------
+        .. [1] Ozarslan E. et. al, "Mean apparent propagator (MAP) MRI: A novel
+        diffusion imaging method for mapping tissue microstructure",
+        NeuroImage, 2013.
+        """
+        Bm = self.model.Bm
+        rtpp = 0
+        const = 1 / (np.sqrt(2 * np.pi) * self.mu[0])
+        for i in range(self.ind_mat.shape[0]):
```
arokem: It should be possible to cythonize this nested loop rather easily (?)

dipy/reconst/mapmri.py
```diff
@@ -269,6 +270,65 @@ def odf(self, sphere, s=0):
...
+        diffusion imaging method for mapping tissue microstructure",
+        NeuroImage, 2013.
+        """
+        Bm = self.model.Bm
+        rtap = 0
+        const = 1 / (2 * np.pi * self.mu[1] * self.mu[2])
+        for i in range(self.ind_mat.shape[0]):
+            if Bm[i] > 0.0:
+                rtap += (-1.0) ** (
+                    (self.ind_mat[i, 1] + self.ind_mat[i, 2]) / 2.0) * self._mapmri_coef[i] * Bm[i]
```
arokem: Same here

dipy/reconst/tests/test_mapmri.py
```diff
@@ -63,15 +67,38 @@ def test_shore_metrics():
...
+                            fractions=[50, 50], snr=None)
+
+    mapm = MapmriModel(gtab, radial_order=radial_order, lambd=lambd)
+    mapfit = mapm.fit(S)
+
+    # RTOP
+    gt_rtop = 1.0 / np.sqrt((4 * np.pi * tau)**3 *
+                            mevals[0, 0] * mevals[0, 1] * mevals[0, 2])
+    rtop = mapfit.rtop()
+    assert_almost_equal(rtop, gt_rtop, 6)
```
arokem: This one seems to be failing in python 3, and also with older versions of numpy/scipy/etc.

Any ideas why this would be the case?

maurozucchelli: It can be some numerical instability of Scipy. I will try to relax a bit the condition, RTOP has very high value ~10^6  so even if it is wrong by 10^(-2) it will be a very good estimation...

dipy/reconst/mapmri.py
maurozucchelli: I think yes, but is it that slow? I mean, ind_mat.shape[0] is usually 22 or 50 for full brain can take some time, but still less than the actual fitting. Are there guidelines for cythonizing code in Dipy right?

arokem: You're right. If this is not a huge bottleneck, it's probably not worth the work and maintenance

Garyfallidis:
@maurozucchelli what would be the best dataset to show the capabilities of this method?
You can have a look at ``dipy.data.fetcher``  for all the lists of datasets that are currently available. It is always nicer to show these methods with a real brain. We should change the SHORE example too with a real brain too.

arokem:
One handy option is the cenir multi b-value data-set that @RafaelNH used in the DKI work: 

https://github.com/nipy/dipy/blob/master/doc/examples/reconst_dki.py#L82

arokem:
Hey @maurozucchelli - have you had a chance to start working on an example? Is the CENIR data-set appropriate for your needs?

maurozucchelli:
Hi I pushed the example, the dataset looks very good. I also fixed the bug
pointed by Rutger in the orientation of the ODF

doc/examples/reconst_mapmri.py
```diff
@@ -12,24 +12,30 @@
...
-with x, y, z and the three axis defining the spatial positions of the voxels.
 """
+For this example we select only the shell with b-values equal to the one of the
+Human Connectome Project (HCP).
+"""
+bvals = [1000, 2000, 3000]
 
-fetch_sherbrooke_3shell()
-img, gtab = read_sherbrooke_3shell()
+gtab, img = read_cenir_multib(bvals)
```
arokem: Please rebase on top of master. I believe this should now be:
    
    img, gtab = read_cenir_multib(bvals)

arokem:
Another comment about the example: since this runs rather quickly, maybe go for a full slice? It might look a bit more compelling. 

Concerning @rutgerfick's additions: I propose that we first merge this work (very close!), and wait until there are some tests on #740 , and then refactor this example with the additions that @rutgerfick proposed.

maurozucchelli:
I don't know there are some issues about using a full slice: the ODF image will be too small and less clear, the voxel near the scalp are very unstable for the metrics (around ten times the max tissue value). This would mean that I will have to find some ad-hoc threshold to visualize the maps correctly, which strongly depends on the data. Maybe, if you prefer, I can change to a saggital slice that better shows the corpus callosum.

arokem:
OK - this is good for now. We can revisit the example in future PR.

PR #735, opened 2015-10-16
Added comment on nlmeans example about selecting one volume

samuelstjean:
In case people take the example too literally.

### DIFF
--- a/doc/examples/denoise_nlmeans.py
+++ b/doc/examples/denoise_nlmeans.py
@@ -25,6 +25,8 @@ data = img.get_data()
 affine = img.get_affine()
 
 mask = data[..., 0] > 80
+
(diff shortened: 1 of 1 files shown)


doc/examples/denoise_nlmeans.py
```diff
@@ -25,6 +25,8 @@
 affine = img.get_affine()
 
 mask = data[..., 0] > 80
+
+# We select only one volume for the example to run quickly.
```
arokem: Maybe write a commented-out version of the line without this? Remember that people are reading the examples because they don't know how to do these things.

arokem: Just noticed that was (ironically) nonsense. Merging.

PR #741, opened 2015-10-19
ENH: it is preferable to use choice rather than randint to not have 

Garyfallidis:
multiple streamlines selected twice.


arokem:
Yep - clearly a win. Will merge when Travis is done looking at this.

PR #743, opened 2015-10-19
BF: memleaks with typed memory views in Cython

### LINKED ISSUE #737: Possibly set_number_of_points doesn't delete memory
You can simply run the function with a large tractogram about (>1M streamlines) and resample at 2 points. You should see your memory going up.

MarcCote:
Refcount is not decreased when converting a numpy array, previously defined in Python space (i.e. not in the .pyx file), into a Cython typed memory view.

The workaround is to create a copy of the numpy array using `astype` while in Cython space. Luckily for us, most of the time no data is copied, only metadata (i.e. when the dtype stays the same).

Fixes #737

### DIFF
--- a/dipy/tracking/streamlinespeed.pyx
+++ b/dipy/tracking/streamlinespeed.pyx
@@ -88,14 +88,12 @@ def length(streamlines):
     if dtype is None:
         # List of streamlines having different dtypes
         for i in range(len(streamlines)):
-            streamline = streamlines[i]
-            dtype = streamline.dtype
+            dtype = streamlines[i].dtype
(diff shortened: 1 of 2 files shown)


Garyfallidis:
Thx! Cool debugging and thank you for discovering another cython bug! I hope your reported this. You have some minor pep8 issues on your tests. Fix them in another PR please. Not crucial for now. Thx much!

MarcCote:
Yes, I made a post explaining the problem and attached some code to reproduce the bug. Last time I did this they confirmed the bug and opened a issue in their bug tracker pretty quickly. However, as of now, it is still not fixed :(

PR #745, opened 2015-10-22
faster version of piesno

samuelstjean:
Some small improvements I had on my version, should be 80 to 100 times faster.


dipy/denoise/noise_estimate.py
```diff
@@ -201,6 +201,12 @@ def _piesno_3D(data, N, alpha=0.01, l=100, itermax=100, eps=1e-5,
     Journal of Magnetic Resonance 2009; 197: 108-119.
     """
 
+    if np.all(data == 0):
```
arokem: Please write a test for this code path.

arokem: Good call though - GIGO

arokem:
in general, testing in this module is rather spotty (76%). Please add tests of corner cases, to be sure that we actually exercise these.

arokem:
Other than that looks good. I tried this with the example, to make sure we don't get any regressions, and it does replicate *exactly* the same results as before, and runs much faster - nice work.

arokem:
Any more thoughts here? @samuelstjean - did you see my comments?

samuelstjean:
Yes, but since the mask is recomputed at each iteration, I dont know if
it's better to replace or reassign the values. And I don't know the
difference between : and ... for this case.

arokem:
Yep. Neither do I, which is why I asked. Does anyone else have an opinion
about this?

@samuelstjean: have you had a chance to get some more tests in there?

Garyfallidis:
It looks good. @samuelstjean increase the coverage and let's merge this asap.

arokem:
If not, it can also wait for after the release.

arokem:
I mean for the next release.

samuelstjean:
Busy ismrm stuff this week.

samuelstjean:
How do you check coverage by the way? Coveralls.io seems ok for reporting that automatically.

Garyfallidis:
You can use 
```
nosetests --with-coverage --cover-package=dipy.denoise
```
You will need to pip install coverage.
Also Travis reports coverage automatically in specific bots which have the flage COVERAGE=1
See here 
https://travis-ci.org/nipy/dipy/jobs/92228117

dipy/denoise/noise_estimate.py
```diff
@@ -217,53 +223,46 @@ def _piesno_3D(data, N, alpha=0.01, l=100, itermax=100, eps=1e-5,
...
-                omega_size = 0
-                break
+        s = sum_m2 / (2 * K * sigma**2)
+        mask[...] = np.logical_and(lambda_minus <= s, s <= lambda_plus)
+        omega = data[mask, :]
 
-            sig_prev = sig
+        # If no point meets the criterion, exit
+        if omega.size == 0:
+            break
```
arokem: This one's also not tested at the moment.

arokem:
Would you mind rebasing this on master? That way, we'll get it also tested on 3.5 on travis. Thanks.

dipy/denoise/noise_estimate.py
Garyfallidis: @samuelstjean  and @arokem is this now tested?

samuelstjean: Yes thats the latest commit I added

PR #747, opened 2015-10-23
TEST: make test faster

Garyfallidis:


dipy/denoise/tests/test_nlmeans.py
```diff
@@ -90,8 +90,15 @@ def test_nlmeans_4d_3dsigma_and_threads():
     sigma = np.ones(data.shape[:3])
     mask = np.zeros(data.shape[:3])
 
-    # mask[25-10:25+10] = 1
-    mask[:] = 1
+    # If you don't provide large enough datasets the OS or OpenMP may not give
+    # all threads to all CPUs
+    # We use False by default to make the test run faster.
+    heavy_duty = False
```
arokem: Maybe better to do this through a global switch (something like "fast" and using `skipif`)?

Garyfallidis: No, that would be a bad idea because the point is to make the test faster in all computers both Travis and ours. But if someone wants to see the effect of his cpus being used he can change the variable. To see if you are using all your cpus you need to check manually your system monitor. This is quite complicated to do automatically.  I hope this is clear. :)

arokem: Just to explain further: my suggestion is to separate off a test that checks all the cpus and only execute that when the global `fast` is set to `False`. As it is, this is hidden a bit out of sight for anyone to take good advantage of (including ourselves 6 months from now).

Garyfallidis:
I am closing this one for now. We will be testing some new PRs with OpenMP mostly coming from @stephanmeesters and some others on openmp in windows with @matthew-brett. It will be good to have a basic test working even if it takes a bit more time. Let's revisit this later.

PR #750, opened 2015-10-27
Updated LICENSE

souravsingh:
Updated License File to add License for Otsu


arokem:
Hey @souravsingh - thanks for doing this! But - I don't think that we want this notice to appear in the main LICENSE file of the repo. Rather, I think this should be placed alongside the Otsu code, so alongside this file: https://github.com/nipy/dipy/blob/master/dipy/segment/threshold.py. I would name it something like LICENSE-skimage, and add a note to the docstring of the otsu function, mentioning that the license for this function is in this file. Does that make sense to you, @stefanv?

stefanv:
Yes, I think that would be fine, as long as you have a line in the license
file that makes it clear to look for other LICENSE-* files.

arokem:
So we should probably add the line that in the skimage license: 

    Unless otherwise specified by LICENSE.txt files in individual directories, all code is

At the top of the dipy main license file as well.

souravsingh:
Alright, I will make the changes

arokem:
Hey @souravsingh - have you had a chance to take a look here?

PR #751, opened 2015-10-28
Reorient tracks according to ROI

arokem:
This is important in order to correctly pull out a core of a bundle based on a median operation on the coordinates. For details, see: 

https://github.com/arokem/AFQ-notebooks/blob/08ac2fc66d1ded0234eec898438d0c7f8380cafc/AFQ-registration-callosum.ipynb

### DIFF
--- a/dipy/tracking/streamline.py
+++ b/dipy/tracking/streamline.py
@@ -9,6 +9,7 @@ import dipy.tracking.utils as ut
 from dipy.tracking.utils import streamline_near_roi
 from dipy.core.geometry import dist_to_corner
 from scipy.spatial.distance import cdist
+from copy import deepcopy
 
 def unlist_streamlines(streamlines):
     """ Return the streamlines not as a list but as an array and an offset
@@ -253,7 +254,7 @@ def select_by_rois(streamlines, rois, include, mode=None, affine=None,
             yield sl
 
 
-def orient_by_rois(streamlines, roi1, roi2, affine=None):
+def orient_by_rois(streamlines, roi1, roi2, affine=None, copy=True):
     """Orient a set of streamlines according to a pair of ROIs
 
     Parameters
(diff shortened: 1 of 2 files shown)


dipy/tracking/streamline.py
```diff
@@ -251,3 +251,63 @@ def select_by_rois(streamlines, rois, include, mode=None, affine=None,
...
             yield sl
+
+
+def orient_by_rois(streamlines, roi1, roi2, affine=None):
+    """Orient a set of streamlines according to a pair of ROIs
+
+    Parameters
+    ----------
+    streamlines : list
+        List of 3d arrays. Coordinates of the
```
jchoude: Missing the end of the comment

dipy/tracking/streamline.py
```diff
@@ -251,3 +251,63 @@ def select_by_rois(streamlines, rois, include, mode=None, affine=None,
...
+    roi1, roi2 : ndarray
+        Binary masks designating the location of the regions of interest, or
+        coordinate arrays (3d coordinates)
+    affine : ndarray
+        Affine transformation from voxels to streamlines. Default: identity.
```
jchoude: Since this is something that pops up in almost all of our internal discussion on streamlines: should the affine documentation specify if the offset should be included in the transformation? By offset, I mean the same offset that is used when mapping streamlines to voxel coordinates to interact with a mask, which is normally 0.5 * the voxel size.

I would say that it is up to the user, and in most cases, it probably won't change the end result, but I'm curious what others would say.

jchoude:
Good, simple work. If there was an approve button, I would click it.

arokem:
Thanks for taking a look, @jchoude - and thanks for the comments. 

I added a couple more tests, to get full coverage, and I added a `copy` kwarg, so that streamlines can be copied or changed in-place (default is to copy).

PR #752, opened 2015-10-28
TST: Install vtk and mesa on Travis to test the fvtk module.

arokem:
Finally! I think that I have found the incantations that will make Travis test #587 without dying. Once this is merged into master, and #587 is rebased, we are essentially ready for the release (?). Going to check on those buildbots now...


arokem:
This was superseded through fixes to #587

PR #753, opened 2015-10-29
TST: Test DSI with b0s.

arokem:
Is this a problem? 

cc: @Garyfallidis

### DIFF
--- a/dipy/reconst/dsi.py
+++ b/dipy/reconst/dsi.py
@@ -1,10 +1,9 @@
 import numpy as np
 from scipy.ndimage import map_coordinates
 from scipy.fftpack import fftn, fftshift, ifftshift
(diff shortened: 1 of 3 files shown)


samuelstjean:
DSI is an acquisition scheme in which you have usually only one b0, not sure it does make sense to have more than one conceptually just because of that, but ask mic, he's the dsi expert.

arokem:
Yep. This is something @garyfallidis mentioned to me in passing as a
potential bug, so I am floating it here.

The problem in this case seems to be that even just one b0 raises an error
(the dsi_voxels don't have a b0. Smallest b-value in that set is 15).

If people do run into this with multiple b0's being the issue, they can
always get around it by 'preprocessing' the data to average all the b0
measurements into one b0 that gets passed in.

Garyfallidis:
Seems working :+1:

PR #754, opened 2015-10-30
DOC: Should we add a side-car gitter chat to the website?

arokem:
For an example of what that looks like see: https://sidecar.gitter.im/

### DIFF
--- a/doc/_templates/layout.html
+++ b/doc/_templates/layout.html
@@ -28,7 +28,7 @@
 {# This block gets put at the top of the sidebar #}
 {% block sidebarlogo %}
 
-  
+
 <h4> Site Navigation </h4>
   <ul>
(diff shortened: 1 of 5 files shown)


Garyfallidis:
We should. And we should also update the front-page with some new cool figure and put the OHBM figure at the old highlights wiki.
Also I think it would be nice to work during or after Christmas on making the website better and supporting multiple documentations from different dipy versions. That is a request that people ask me very often and they are correct. Now we always show what is in the latest dev version but what we need is to have the documentation for the released version and then have links to the dev docs and previous releases. But to do so we need to redesign a bit how the website generation works. It shouldn't be too hard. "Famous last words!" :)

arokem:
OK - I will keep working on website tweaks towards the release on this branch.

PR #757, opened 2015-10-31
Optimize dipy.align.reslice

dimrozakis:
- Using a 1D transformation matrix instead of a 2D diagonal matrix reduces computation time.
- Using the `output` argument in `affine_transform` reduces memory usage.

### DIFF
--- a/dipy/align/reslice.py
+++ b/dipy/align/reslice.py
@@ -32,7 +32,7 @@ def reslice(data, affine, zooms, new_zooms, order=1, mode='constant', cval=0,
     cval : float
         Value used for points outside the boundaries of the input if
         mode='constant'.
-    num_processes: int
+    num_processes : int
         Split the calculation to a pool of children processes. This only
         applies to 4D `data` arrays. If a positive integer then it defines
         the size of the multiprocessing pool that will be used. If 0, then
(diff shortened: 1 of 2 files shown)


arokem:
Looks great to me. If someone else could take a look that would be great. Otherwise, if no one has any comments, I can merge this towards the end of the week. Thanks!

dimrozakis:
In https://github.com/tomotech/dipy/commit/7f7e8a74e9b6395124a00fac095ca2751e8f6d51 I added the option to use `multiprocessing.Pool` in `reslice`. Should I push the commit into this branch as well?

Garyfallidis:
Good idea @dimrozakis. Use ``num_processes`` as a variable name rather than ``mp``. We usually use ``num_processes`` or ``nbr_processes``  when using multiprocessing and ``num_threads`` when using multithreading.

dimrozakis:
@arokem I just fixed the missing space issue in the `num_processes` definition of `dipy.align.reslice` docstring according to your comment

arokem:
Thanks! I will wait for travis to give this a run through, and then merge...

arokem:
I just noticed that test coverage here seems rather spotty (63% for the module). Any chance you could do something about that? Thanks!

dimrozakis:
@arokem I added more tests, now coverage is at 98%

The 8th test case, the one with `python-vtk` fails. I have no idea why. It failed on the previous commit as well, which simply added a space character, so it must involve something merged to master since this branch was forked?

arokem:
Thanks. That's fantastic!

The VTK machine failure is a known (and rather hard!) issue. We'll sort it
out, but maybe not on this PR...

PR #760, opened 2015-11-03
NF - random seeds from mask

gabknight:
- This PR adds the fct ``random_seeds_from_mask_total`` to generate N seeds given a mask. This complements the fct ``random_seeds_from_mask_per_voxel`` that already generates seeds given a mask. This is useful when you want N streamlines per region, with regions of various size.

- It also fixes a small bug. If N seeds were sent to ``LocalTracking`` with ``return_all==True``, the returned streamlines excluded all seeds with no valid initial direction. Now, if a seed has no valid initial direction, it return a streamline containing only the seed position.

### DIFF
--- a/dipy/tracking/utils.py
+++ b/dipy/tracking/utils.py
@@ -370,7 +370,8 @@ def seeds_from_mask(mask, density=[1, 1, 1], voxel_size=None, affine=None):
 
     See Also
     --------
-    random_seeds_from_mask
+    random_seeds_from_mask_per_voxel
+    random_seeds_from_mask_total
 
     Raises
     ------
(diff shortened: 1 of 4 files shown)


arokem:
These function names are really long. Might it make more sense to incorporate the new functionality into the existing function as a kwarg set per default to the 'old' functionality ('voxel'), but allowing use of the new functionality ('total')?

gabknight:
Thx @arokem. I did what you suggested.  let me know what you think.

arokem:
Great! Just one more suggestion - to test the BF that you had there. It seems that we'd want a test to prevent that from creeping back into the code.

dipy/tracking/utils.py
```diff
@@ -503,6 +515,11 @@ def random_seeds_from_mask(mask, seeds_per_voxel=1, affine=None):
     seeds = where + grid - .5
     seeds = asarray(seeds)
 
+    if not seed_count_per_voxel:
+        # Randomize the seeds and select the requested amount
+        np.random.shuffle(seeds)
```
MarcCote: Wouldn't be better to create its own random generator using a seed (ideally provided as an argument to this function) instead of relying on the user having to call `np.random.seed(seed)`. I think changing the seed of the global random generator (i.e. the one in `np.random`) is prone unreproducible results. (e.g., for some reason you decided to call `np.random.rand()` in another part of your code before calling this function).

gabknight: I see your point. Currently there is only one dipy function having a random_seed parameter (`dipy.align.sample_domain_regular`). Elsewhere, it is up to the user to set random.seed before calling functions using a random number generator. 

If we add a random seed parameter to functions, the random seed parameter should be uniform throughout the code, and adding a parameter seed=0 as in `dipy.align.sample_domain_regular` could bring a bit a confusion for this particular function. 

What do you guys think? Should we set a standard parameter name to add to function using np.random, say random_seed=0?

Garyfallidis: My view is that there is no reason to have a parameter for the random state. It will bring more confusion than good.  But if you add in the documentation that you can actually specify the randomness (using np.random.seed...) before calling the function that would be enough.

Garyfallidis:
@gabknight can you rebase this if needed and you think is ready? And I will merge it asap.

gabknight:
Done. @Garyfallidis, all good on my side

dipy/tracking/utils.py
```diff
@@ -436,21 +436,28 @@ def seeds_from_mask(mask, density=[1, 1, 1], voxel_size=None, affine=None):
...
-    which are ``True``. This function is essentially similar to
-    ``seeds_from_mask()``, with the difference that instead of evenly
+    Seeds points are placed randomly distributed in voxels of ``mask``
+    which are ``True``.
+    If ``seed_count_per_voxel`` is ``True``, this function is
+    similar to ``seeds_from_mask()``, with the difference that instead of evenly
     distributing the seeds, it randomly places the seeds within the voxels
-    specified by the ``mask``
+    specified by the ``mask``. The initial random conditions can be set using
+    ``numpy.random.seed(...)``, prior to call this function.
```
arokem: => "prior to calling this function"

arokem:
Just small comment on docs. Otherwise - I am +1 for the merge.

gabknight:
Thanks!

Garyfallidis:
Good job! Let's move to the next frontier! :+1:

Issue #763, opened 2015-11-05
Tensor Fitting Overflows Memory


bcmcpher:
I'm running a updated clone of dipy from current master (0.10.0-dev I believe) to fit a tensor model in python 2.7.10:

```
from dipy.data import read_stanford_hardi
import dipy.reconst.dti as dti

img, gtab = read_stanford_hardi()
dti_model = dti.TensorModel(gtab)
data = img.get_data()
dti_fit = dti_model.fit(data)
```

I have 8GB system RAM and 8GB swap, all of which fills up before the process dies. Something appears to have changed in how memory is utilized, as it didn't used to do this.

This is on a laptop running Linux Debian Stable (8.2) with the kernel from backports (4.2.0-0.bpo.1-amd64 #1 SMP Debian 4.2.5-1~bpo8+1 (2015-11-02) x86_64 GNU/Linux) and Continuum conda 3.18.3. 

Let me know if you need any other information.

arokem:
Could this be related to https://github.com/nipy/dipy/pull/727

arokem:
Any thoughts about this @dimrozakis ?

dimrozakis:
In #727 all fitting operations were vectorized, bringing a great increase in speed. But when all operations are performed on all voxels at the same time, then any temporary arrays created are a lot larger than when iterating over single voxels. To work around this, one can split the large input array into a number of chunks and apply the vectorized fitting function by iterating on those chunks. If the number of chunks is considerably less than the total number of voxels, then there's no significant decrease in speed, but the temporary memory consumption decreases a lot. I've written something like that in https://github.com/tomotech/dipy/commit/c080e8981d76a7f7f6cf4b66d6e5b06070501d82 . @arokem should I open a PR?

dimrozakis:
@bcmcpher would you like to merge https://github.com/tomotech/dipy/commit/c080e8981d76a7f7f6cf4b66d6e5b06070501d82 and tell us if this fixes your out of memory issue?

Garyfallidis:
@dimrozakis indeed you need to use chunks to be memory efficient with this. Dimitri can you provide us with your current memory consumption in your system with and without using chunks and also the speedup gain in comparison to the initial version? It would be great to have all this information so we can decide how to move forward with this. Thx in advance.

dimrozakis:
With the stanford hardi dataset (81, 106, 76, 160):

Before PR #727 (0503c72):
```
Completed in 173.94 secs.
[...]

PR #766, opened 2015-11-06
Add fit_tensor iteration decorator

dimrozakis:
Splits data into a number of chunks of specified size and iterates the
decorated fit_tensor function over them. This is useful to counteract
the temporary but significant memory usage increase in fit_tensor
functions that use vectorized operations and need to store large
temporary arrays for their vectorized operations.

### DIFF
--- a/dipy/reconst/dti.py
+++ b/dipy/reconst/dti.py
@@ -4,6 +4,8 @@ from __future__ import division, print_function, absolute_import
 
 import warnings
 
+import functools
+
 import numpy as np
 
 import scipy.optimize as opt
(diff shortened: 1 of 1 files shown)


dimrozakis:
Closes #763

Garyfallidis:
Thank you for this! :)

Garyfallidis:
I am going to wait for travis and if all good will merge it. Keep it up!

dimrozakis:
thanks

dipy/reconst/dti.py
```diff
@@ -1163,6 +1165,42 @@ def predict(self, gtab, S0=1):
         return tensor_prediction(self.model_params[..., 0:12], gtab, S0=S0)
 
 
+def iter_fit_tensor(step=1e4):
```
Garyfallidis: I am going to go ahead and merge this PR because a lot of people are always using the master. Please add a Parameters section in another PR. And explain what step is. Is this the number of voxels for example?

Garyfallidis:
Thx @dimrozakis. Check my last comment. :+1:

dipy/reconst/dti.py
dimrozakis: Yes `step` is the number of voxels in one chunk. I will push a PR putting a params section, probably during the weekend.

Garyfallidis:
Okay good. While you are on it maybe check also that when you parallelized reslicing there was not memory addition there too. Multiprocessing can duplicate memory at times.

dimrozakis:
@Garyfallidis:

>Please add a Parameters section in another PR. And explain what step is.

Done in #768 

>While you are on it maybe check also that when you parallelized reslicing there was not memory addition there too.

Well, there is *some* memory increase. I think that's inevitable when using multiprocessing and not using shared memory. However, I don't think it's much of an issue. Let me elaborate a little bit on this:

Multiprocessing support in `dipy.align.reslice` was added in https://github.com/tomotech/dipy/commit/7f7e8a74e9b6395124a00fac095ca2751e8f6d51 and the implementation splits the 4D input array to a number of 3D arrays, which it sends to the workers of a multiprocessing pool. When sending over a 3D array to a worker, then that array is duplicated. If one has N workers and G gradients, this will take up at most `N/G` more memory. For 4 workers and 40 gradients, that's a 10% temporary memory increase over the size of the input array. When sending back the results, until they're saved in the output 4D array, each chunk will temporarily take up more memory (since the `output` param of `affine_transform` can't be used). So there's some more memory there. Note however that it was the previous commit that started using the `output` param and that the pool implementation is still better in this aspect than 2 commits before. Also by default `reslice` won't use a Pool. So I'd say there isn't something to fix there. Some things could further be improved by using shared memory, but that also has its own problems (for example if the input array isn't given in shared memory, copying it there will temporarily take up twice as much memory).

Garyfallidis:
Thank you for the detailed explanation.

Issue #517, opened 2014-12-24
TODO: Peaks to be removed from dipy.reconst in 0.10


Garyfallidis:
Peaks will be accessible only from dipy.direction after version 0.10.

arokem:
We need to do this now.

theaverageguy:
This can be closed @arokem

arokem:
Thanks for the reminder. Closing.

PR #770, opened 2015-11-13
RF: Move the peaks module from dipy.reconst to dipy.directions. 

arokem:
Addresses #517

### DIFF
--- a/bin/dipy_peak_extraction
+++ b/bin/dipy_peak_extraction
@@ -6,7 +6,7 @@ import numpy as np
 from dipy.fixes import argparse
 
 from dipy.core.ndindex import ndindex
-from dipy.reconst.peaks import peak_directions
+from dipy.direction.peaks import peak_directions
 from dipy.core.sphere import Sphere
 
 
--- a/dipy/direction/__init__.py
+++ b/dipy/direction/__init__.py
@@ -1,5 +1,4 @@
 
 from .probabilistic_direction_getter import (ProbabilisticDirectionGetter,
                                              DeterministicMaximumDirectionGetter)
-import peaks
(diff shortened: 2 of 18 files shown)


MrBago:
Shouldn't this get moved to `dipy/direction` and relevant imports be adjusted?

arokem:
Clearly this by itself is no good :-)

arokem:
How does this look @MrBago ?

samuelstjean:
Can you add a deprecation warning and import the new module in the old one for compatibility with old stuff?

arokem:
How's that @samuelstjean ?

Garyfallidis:
@MrBago can you merge this PR if you think it is okay? We are in release mode right now and we would appreciate some quick feedback so we can move on. Thx in advance.

MrBago:
This looks good to me. My only question is about the DeprecationWarning, these are silenced by default so most people will not see them. Do we want something stronger? I feel like this question has come up before so if we settled on decision, I'm sorry for raising it again.

Garyfallidis:
I agree that this is an important discussion and I am not sure we had a final decision on this. But because this is a generic design issue and not related with this PR I would suggest to move on with the release and then open an issue or a discussion in gitter to figure out the best warning strategy. Sounds good?

MrBago:
K, I'm merging this. I brought this up partially because we're already using UserWarnings for this kind of thing in some places, /home/bagrata/projects/dipy/dipy/align/aniso2iso.py.

Garyfallidis:
Good reminder. Write an issue please so we can keep track of that.

Garyfallidis:
We should clearly resolve this problem by 0.11 release.

PR #772, opened 2015-11-17
DOC: Use xvfb when building the docs in a headless machine.

arokem:
Sometimes I want to build the docs on a headless machine.


Garyfallidis:
Very nice addition! Headless wins! Let's use the cloud for building the docs. :+1:

PR #773, opened 2015-11-20
Website 0.10

arokem:
Release notes! And a few more small changes to docs.


doc/index.rst
```diff
@@ -16,9 +16,9 @@ Highlights
 **Dipy 0.10** is now available. New features in this release include: 
 
 
-- FVTK 2.0: a new visualization framework. For example, see: 
 - DKI: Thanks to support through the Google Summer of Code, to fund work done by Rafael Henriques, we now have the only open-source implementation of DKI.
```
samuelstjean: Is is really the only open source DKI implementation? A quick google search gives another one on github on the first link. Might be the only one in python though.

Garyfallidis: Yeah, I think there are some Matlab implementations too.

Garyfallidis: I would also add MAPMRI after DKI at this list.

Garyfallidis: And of course Affine registration is an important new addition to complement the nonlinear registration. Big cool addition.

arokem: Yup. Good catch. Previous searches did not surface these. I wonder if they've open-sourced recently...

doc/index.rst
```diff
@@ -16,10 +16,12 @@ Highlights
 **Dipy 0.10** is now available. New features in this release include: 
 
 
-- FVTK 2.0: a new visualization framework. For example, see: 
-- DKI: Thanks to support through the Google Summer of Code, to fund work done by Rafael Henriques, we now have the only open-source implementation of DKI.
-- Compatibility with new versions of scipy (0.16) and numpy (1.10)
-
+- Compatibility with new versions of scipy (0.16) and numpy (1.10).
+- New cleaner visualization API, including compatibility with VTK 6, and functions to create your own interactive visualizations.
+- DKI: Google Summer of Code work by Rafael Henriques.
```
Garyfallidis: I think it's more preferable to write Diffusion Kurtosis Imaging (DKI)

doc/stateoftheart.rst
```diff
@@ -6,7 +6,7 @@ A quick overview of features
 
 Here are just a few of the state-of-the-art :ref:`technologies <examples>` and algorithms which are provided in Dipy_:
 
-- Reconstruction algorithms e.g. CSD, DSI, GQI, DTI, QBI and SHORE.
+- Reconstruction algorithms e.g. CSD, DSI, GQI, DTI, DKI, QBI and SHORE.
```
Garyfallidis: SHORE and MAPMRI

doc/stateoftheart.rst
```diff
@@ -14,6 +14,7 @@ Here are just a few of the state-of-the-art :ref:`technologies <examples>` and a
 - Reslice datasets with anisotropic voxels to isotropic.
 - Calculate distances/correspondences between streamlines.
 - Deal with huge streamline datasets without memory restrictions (.dpy).
+- Visualize streamlines relative to anatomical images.
```
Garyfallidis: in the same space with anatomical images

doc/subscribe.rst
```diff
@@ -7,11 +7,12 @@ Subscribe
...
 community and boost collaboration across neuroimagers. If you
-want to post to the list you will need first to subscribe to `nipy
+want to post to the list you will need first to subscribe to the `nipy
 mailing list`_. We suggest to begin the subject of your e-mail with
 ``[Dipy]``. This will help us respond faster to your questions.
 
-Best wishes,
-Dipy developers
+Additional help can be found on the Neurostars_ website. You can 
+sometimes catch us on the `dipy gitter`_ channel.
```
Garyfallidis: catch us a bit too informal

Garyfallidis: The README file also needs an update with the new mailing list. See issue https://github.com/nipy/dipy/issues/774

Garyfallidis: Thx for correcting the README file. Please don't forget to address the comment above the README. Should we say that gitter is more for developers or we want anyone to be able to drop and ask questions? If we do so people maybe able to have difficulties finding information about previously posted questions.

arokem: I think that it is more than fine for users to find us on any public forum
that has history saved in some form. That includes gitter, neurostars, or
the mailing list. We may direct some questions that call for a broader
discussion to the mailing list. I don't mind mixing user and developer
discussions into the same forum. On the contrary - that makes users feel
like they might also be developers, which is one of the goals of the
project, as I see it - to empower users to make their tools more useful for
them

As for "catch us" - I don't think this sounds "too informal". We're not an
insurance company.

arokem: Just to add: I don't mind revisiting the "ask us for help anywhere you can
find us" policy at some later point, if we find that it doesn't work well
for us. So far, I think that it is working well. See recent conversations
with Pratik, etc.

Garyfallidis: I would still simply write ``Additional help can be found on the Neurostars website or in `dipy gitter` `` Also, I think it makes sense to send an e-mail around and ask if the other dipy developers want to have gitter in the picture for user support. From what I see from other projects, gitter is used mostly as a forum for developers and github users but not for every user. Everyone of course should be able to get in or read the discussions but I am not sure that this is the correct framework for Q&A or user support. We may need to open a new channel only for that if we want to have a forum based support. I suppose you are assuming in all this that the traffic in gitter will stay low for some time. But you never know... Anyway no strong opinions here on my side. Your call.

arokem: Yes - I am proposing to use this release as an experiment to see how that
goes. Following some arguments made by @matthew-brett, I am generally
disinclined to separate "developers" from "users", unless things become
unwieldy. We're all in this together!

The only project I have seen that explicitly separates is IPython, and they
have millions of users, and several dozen developers working together on
dozens of repos. But you are right that gitter channels tend to be places
where development is discussed. It is intentionally the last in this list.
But it will also now be featured as a side-car on the front page of the
website, so we might get a lot of people coming in through there. I think
that giving immediate low-barrier feedback to people who get stuck on
something early on in their experience will be important to sustain a flow
of new users (and eventually more developers) into the project. That's the
hypothesis, and this is the experiment.

Garyfallidis: Sure I understand your point. I suggest that you send what you wrote above in an e-mail so that as many people know about this experiment.

arokem: OK - sent.

PR #779, opened 2015-11-22
Saving DiffeomorphicMap with pickle

omarocegueda:
Pickle fails when attempting to save memory views. Here we simply convert the `forward` and `backward` deformation fields to numpy arrays so we can now save the `DiffeomorphicMap` instance returned from `SymmetricDiffeomorphicRegistration` to disk with pickle.

### DIFF
--- a/dipy/align/imwarp.py
+++ b/dipy/align/imwarp.py
@@ -1434,4 +1434,6 @@ class SymmetricDiffeomorphicRegistration(DiffeomorphicRegistration):
                              static_grid2world, moving_grid2world, prealign)
         self._optimize()
         self._end_optimizer()
+        self.static_to_ref.forward = np.array(self.static_to_ref.forward)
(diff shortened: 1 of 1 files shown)


Garyfallidis:
Hi Omar, thank you for fixing this problem. Just to say here that saving large diffeomorphic maps with pickling may not be a good idea as pickle can generate more memory than needed and have slower loading and saving times. We could alternatively use Nifti files or Numpy files (npz or npy) to save those. But anyway your fix now allows for many possibilities. Thx!

omarocegueda:
Thank you Elef!,
I totally agree, but meanwhile people should have at least one easy way to save the maps. The only thing I don't like about the Niftis is that people would need to have many files for each DiffeomorphicMap (2 deformation fields, and at least one numpy file containing the affines), I would like to save everything within one single file.

PR #780, opened 2015-11-23
Fix anisotropic voxel size use by setting vox_size in vtkImageReslice of slicer actor

Garyfallidis:
If the voxel size was not isotropic the slicer was not showing properly the image in world space. Now this is fixed. The previous version had the potential advantage of allowing to slice in 1mm voxel size rather than in voxel coordinates. This is still available by unsetting the flag voxel size and it should work nicely with isotropic voxel sizes but the results will not be nice with anisotropic voxel sizes therefore it is set to false by default.

### DIFF
--- a/dipy/viz/actor.py
+++ b/dipy/viz/actor.py
@@ -47,8 +47,8 @@ def slicer(data, affine=None, value_range=None, opacity=1.,
         latter may have problems with images with anisotropic voxel size.
     interpolation : string
         If 'linear' (default) then linear interpolation is used on the final
-        texture texture mapping. If 'nearest' then nearest neighbor
-        interpolation is used on the final texture mapping.
+        texture mapping. If 'nearest' then nearest neighbor interpolation is
+        used on the final texture mapping.
 
     Returns
     -------
--- a/dipy/viz/actor.py
+++ b/dipy/viz/actor.py
@@ -195,6 +195,9 @@ def slicer(data, affine=None, value_range=None, opacity=1.,
     else:
         image_actor.SetInterpolate(True)
 
+    if major_version >= 6:
+        image_actor.GetMapper().BorderOn()
+
     return image_actor
 
 
--- a/dipy/viz/actor.py
+++ b/dipy/viz/actor.py
@@ -21,7 +21,8 @@ if have_vtk:
 
 
 def slicer(data, affine=None, value_range=None, opacity=1.,
-           lookup_colormap=None):
+           lookup_colormap=None, force_voxel_size=True,
+           interpolation='linear'):
     """ Cuts 3D scalar or rgb volumes into 2D images
 
     Parameters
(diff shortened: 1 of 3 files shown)


Garyfallidis:
@MarcCote have a look at this and if you think it is good, go ahead and merge. @arokem this fixes a bug in current master and it should be merged before the release.

dipy/viz/actor.py
```diff
@@ -39,6 +40,15 @@ def slicer(data, affine=None, value_range=None, opacity=1.,
...
     lookup_colormap : vtkLookupTable
         If None (default) then a grayscale map is created.
+    force_voxel_size : bool
+        If True (default) then initial voxel sizes of the image will be
+        recovered from the affine and used to show the slices correctly. If
+        False then the output spacing will be considered as (1., 1., 1.). The
+        latter may have problems with images with anisotropic voxel size.
+    interpolation : string
+        If 'linear' (default) then linear interpolation is used on the final
+        texture texture mapping. If 'nearest' then nearest neighbor
```
arokem: The word "texture" is repeated

dipy/viz/tests/test_fvtk_actors.py
```diff
@@ -107,6 +107,40 @@ def test_slicer():
...
+    report = window.analyze_snapshot(arr, find_objects=True)
+    npt.assert_equal(report.objects, 1)
+    npt.assert_equal(data.shape, slicer.shape)
+
+    renderer.clear()
+
+    data = (255 * np.random.rand(50, 50, 50))
+    affine = np.diag([1, 3, 2, 1])
+    slicer = actor.slicer(data, affine,
+                          force_voxel_size=False, interpolation=True)
```
arokem: IIUC, the `interpolation` kwarg should be 'nearest' or 'linear', not True/False, right?

dipy/viz/tests/test_fvtk_actors.py
```diff
@@ -107,6 +107,40 @@ def test_slicer():
     report = window.analyze_snapshot(arr, find_objects=True)
     npt.assert_equal(report.objects, 1)
 
+    renderer.clear()
+
+    data = (255 * np.random.rand(50, 50, 50))
+    affine = np.diag([1, 3, 2, 1])
+    slicer = actor.slicer(data, affine,
+                          force_voxel_size=True, interpolation=False)
```
arokem: Same here as below for the `interpolation` kwarg

dipy/viz/tests/test_fvtk_actors.py
```diff
@@ -107,6 +107,40 @@ def test_slicer():
...
+    slicer.display(None, None, 25)
+
+    renderer.add(slicer)
+    renderer.reset_camera()
+    renderer.reset_clipping_range()
+
+    arr = window.snapshot(renderer, offscreen=False)
+    report = window.analyze_snapshot(arr, find_objects=True)
+    npt.assert_equal(report.objects, 1)
+    npt.assert_equal(slicer.shape, (50, 148, 99))
```
arokem: I can tell that this is related to the scaling by the voxel dimensions, but could you add a comment how these numbers are derived? In particular, why is this not simply vox_size * [50, 50, 50]?

arokem:
Cool stuff. I will wait for @MarcCote to take another look after you have a chance to address the comments, and then merge. We still need to deal with those bots, and see whether the new nibabel release has any effects, before doing the release anyway.

dipy/viz/tests/test_fvtk_actors.py
MarcCote: @arokem I think you are right, it should be `vox_size * [50, 50, 50]`. @Garyfallidis I just tested it with the `standard.nii.gz` and `standard.trk` and the borders are crop!
selection_035

Garyfallidis: Yes, okay I can go ahead and remove the option to have force slicing equal false. To be on the safe side. Let me do that.

Garyfallidis:
@MarcCote this should be ready to be merged.

MarcCote:
Ok, now it is working as expected. Once Travis is done, I'll merge it.

Garyfallidis:
:) Thx! Good catch!

PR #782, opened 2015-11-25
TST: buildbot-proof testing with XVFB, in case the env variable gets crunked.

arokem:
This will hopefully help deal with this business: http://nipy.bic.berkeley.edu/builders/dipy-py2.6/builds/606/steps/shell_9/logs/stdio

### DIFF
--- a/dipy/testing/decorators.py
+++ b/dipy/testing/decorators.py
@@ -10,8 +10,9 @@ import os
 
 SKIP_RE = re.compile("(\s*>>>.*?)(\s*)#\s*skip\s+if\s+(.*)$")
 
+
 def doctest_skip_parser(func):
-    """ Decorator replaces custom skip test markup in doctests
+    """Decorator replaces custom skip test markup in doctests.
 
     Say a function has a docstring::
 
@@ -20,13 +21,13 @@ def doctest_skip_parser(func):
         >>> something # skip if HAVE_BMODULE
 
     This decorator will evaluate the expresssion after ``skip if``.  If this
-    evaluates to True, then the comment is replaced by ``# doctest: +SKIP``.  If
-    False, then the comment is just removed. The expression is evaluated in the
-    ``globals`` scope of `func`.
+    evaluates to True, then the comment is replaced by ``# doctest: +SKIP``.
+    If False, then the comment is just removed. The expression is evaluated in
+    the ``globals`` scope of `func`.
 
     For example, if the module global ``HAVE_AMODULE`` is False, and module
-    global ``HAVE_BMODULE`` is False, the returned function will have docstring::
-
+    global ``HAVE_BMODULE`` is False, the returned function will have
+    docstring::
         >>> something # doctest: +SKIP
         >>> something + else
         >>> something
@@ -52,9 +53,15 @@ def doctest_skip_parser(func):
 # xvfbwrapper) conditioned on an environment variable (that we set in
 # .travis.yml for these cases):
 use_xvfb = os.environ.get('TEST_WITH_XVFB', False)
+if use_xvfb == 'true':
+    use_xvfb = True
+
+
 def xvfb_it(my_test):
+    """Run a test with xvfbwrapper."""
     # When we use verbose testing we want the name:
     fname = my_test.__name__
+
     def test_with_xvfb():
         if use_xvfb:
             from xvfbwrapper import Xvfb


dipy/testing/decorators.py
```diff
@@ -52,9 +53,15 @@ def doctest_skip_parser(func):
 # xvfbwrapper) conditioned on an environment variable (that we set in
 # .travis.yml for these cases):
 use_xvfb = os.environ.get('TEST_WITH_XVFB', False)
+if use_xvfb == 'true':
```
Garyfallidis: Hi @arokem. Questions :) I need some clarifications here. Does that mean that I have to write 'true' in the command line too? What we were using before? TEST_WITH_XVFB=1 right? Or am I missing something?

Garyfallidis: Ah no. I am wrong. I see that in the travis file it says https://github.com/nipy/dipy/blob/master/.travis.yml#L50

Garyfallidis: Okay, if you agree I can go ahead and merge your PR.

arokem: Do it! Please.

Ugh - I have spent all day on these bots, and still no joy. If this doesn't
work, I suggest that we go to skipping these tests on the bots.

Garyfallidis:
We really need to make it easier to use the buildbots. I am afraid that we are losing a lot of time dealing with those for each release. I hope this system will be simplified in the future. Maybe we should consider even paying for a service like Travis but that supports OSX and Windows. Time is precious!

arokem:
Travis has OSX. Appveyor does Windows.

We just need to set these things up.

arokem:
Another thing that might make this easier is to split things up a little
bit. For example, to move all the visualization stuff to another repo, that
can have its own release cycle. Part of the problem here (I think) is that
we are trying to release many features on different modules *at the same
time*.

matthew-brett:
You'd have to set up a separate test repo for OSX on travis, travis hasn't been allowing dual os testing for some time I believe : https://blog.travis-ci.com/2014-05-13-multi-os-feature-available .  Then you'd have to trigger the builds somehow, which is not trivial. 

Also buildbots can test on multiple OSX, Windows, Linux versions.  At least the OSX versions we have had use for in the past.

You can also do things like add tests which are too long or need too much memory for travis-ci.   And you can (if you need to) log into the buildbot machines and debug, which you cannot do on travis (apparently you can for circle-ci, but that's Linux only as well I think).

I think you (Ariel) are right that the main problems are because some of the dipy tests are hard to run across machines, rather than inherent to buildbots, although of course the buildbots do have a learning curve.

arokem:
Thanks for chiming in!

For Travis/Mac OS, Let's see what happens on #785. The blog-post you linked
has an update to the update, so maybe it does work? To be completely
honest, the coverage of Mac OS in the buildbots seems a bit excessive. For
example, we are still testing on 10.6, even though 10.7 was released in
2011. I'd assume it would be quite hard to actually find machines with 10.6
out in the wild (for example, if that machine running the bot were to die,
what would we do?).

matthew-brett:
Have a look at https://www.adium.im/sparkle/#osVersion

OSX 10.6 has been holding steady at around 10% for a while now.

Garyfallidis:
@matthew-brett, me and Ariel have two options either learn exactly how to use the buildbots for which you probably need to give us a more detailed crash course. Or move to some different technology. Otherwise the time put on debugging issues on buildbots is too high to help productivity or releasing faster. Right now releasing is usually slowed down considerably because it takes us time to look and debug the buildbots. So, we need to get better at this. I am not saying that this is your fault in some way but I would like to hear what is your plan on simplifying the current experience.

arokem:
That's probably because of this:
http://www.computerworld.com/article/2476004/mac-os-x/why-mac-users-still-use-os-x-snow-leopard.html

I think that it's OK to tell users to buy a newer computer or use a cloud
computing solution, if all they have access to is a used mac.

matthew-brett:
Oh dear.   I think you will also find that universities have a lot of these machines, because labs are often slow to change machines running custom or scientific software.

Of course, if there comes a point where it becomes too hard to maintain 10.6, then drop it, but I don't think you are anywhere near that point at the moment.

Garyfallidis:
For me, it's fine to support old versions of OS.

Garyfallidis:
But we need to calculate what is the maintenance cost of having and using the buildbots.

arokem:
@garyfallidis : support for old Mac OS is one of the reasons that the
buildbots still don't have any feasible replacement.

BTW: that sample from adium could also be read as evidence that we should
drop support for 10.7 and 1.8 (>4% each).

Garyfallidis:
Yes, I understand what you say. But we could have the buildbots for the long term releases and for let's say the 2-month long releases we could look only on Travis in Linux and OSX... Something like that. I don't have a perfect solution in my mind.

arokem:
<4% each, of course

matthew-brett:
Eleftherios - I think the best option is just to ask for my help when you need it.

I can certainly imagine better interfaces to the buildbot results (something more automatic than looking at the web page) but in terms of maintenance, it's always going to be tuning the parameters that go into the builds.

You may want to drop the buildbots of course, but I imagine that will have a moderate cost in terms of the machines that dipy runs on, and / or the machines dipy gives the correct results.   It seems to me that we found lots of problems in Omar's registration code with the buildbots that we would likely have missed without them, and using the buildbots make it much easier to debug.

There's certainly a better case for dropping 10.7 and 10.8 than for 10.6, but you probably don't want to put on the dipy page 'Known to work on 10.6 but may not work on 10.7 or 10.8' unless there's really a cost for those tests, and I think, at the moment, you've got no extra issues for keeping those working.

arokem:
No - I think that as long as we have buildbots, they should all be green on
any release, and ideally we should work to make them all green, whenever
one goes red.

arokem:
Yes - as long as VTK is impossible to install on Macs without Anaconda, I
think we're safe :-)

Garyfallidis:
I think there is an important point to understand here. Yes, with these buildbots we found some bugs. Most of the bugs though were red herrings. The actual problems were not bugs but missing packages or wrong versions of other packages or some of the buildbots we shouldn't be looking because they are idle or off etc. So, most of the times the buildbots need maintenance rather than bug fixes. So, if I spend a month to figure out a bug in one buildbot I maybe better of not looking at it at all and let the users report it if they do. I definitely not like the option of contacting Matthew every time we have a problem with a builbots because this not efficient and I would like to have more regular releases.

Garyfallidis:
Maybe a temporary solution is to increase the number of people who maintain the buildbots.

matthew-brett:
There are certainly red-herrings, but I don't think these take much time to fix.

We did find substantial bugs with Omar's code, and it's not surprising that testing new visualization code would be difficult.

I don't think you've hit a road-block from waiting for me to fix stuff on the buildbots, I try and keep up to date.

Yes, sure, anyone is welcome to help with maintaining the buildbots.

Garyfallidis:
Matthew, I think what you should do is bring more people to be responsible for the maintenance of the buildbots and reduce the workload from your shoulders. It's generally a bad strategy to have a single person responsible for an important project that needs high maintenance. 

It's interesting that you think that the buildbots have been that important on finding bugs. Actually, for Dipy the number of times where we indeed found important bugs were not so many but I don't have any stats to show you. This is my personal feeling about this. Of course I am not saying here that we shouldn't have buildbots to test our code. What I am saying is that the current setting is not efficient. But maybe that is the best we can have right now. Which is fine. But if there can be other alternatives I would like to know.

matthew-brett:
Check the conversations about Omar's stuff where we spent a rather long time working out how to do cross-platform compilation on Windows and OSX, and tracked down floating point precision problems on Windows in particular.

If the new contributions don't involve new compilation, or don't test very variable parts of the system (like visualization) you will have less need of the buildbots.  

Maybe I wasn't clear - if either of you - or anyone else - would like to help maintain the buildbots - please let me know, I would love to have some help.

Garyfallidis:
Sure, there is no doubt that the buildbots have been helpful debugging both the registration and the tracking modules for example. 

Okay so we all agree then. So, from the existing builbots are they some that can be completely be updated and maintained remotely and some that cannot be?

arokem:
http://tjelvarolsson.com/blog/how-to-continuously-test-your-python-code-on-windows-using-appveyor/

Garyfallidis:
And other some of the buildbots that can be completely move to Travis or Appveyor?

matthew-brett:
Turning on Appveyor certainly a good idea in general, but a) its a small virtual machine with timeout making it hard to run heavy jobs; b) hellish to debug on.

All the buildbot slaves can be reached remotely, yes.   You can also trigger builds of arbitrary uncommitted code from your local machine with the ``try_branch.py`` utility - Omar had good use of that for the Windows problems.

arokem:
Would be good to have these things (try_branch, logging into the machines)
documented on the nibotmi repo. For the time being, I keep having to search
my email for previous conversations with you, to tell me how to do these
things. I don't think that I've ever logged into one of these machines with
any success, and actually managed to debug a test failure. I have run
try_branch with some success, but would have to go find an email
conversation we once had to figure out how that's done.

Garyfallidis:
I see also that travis.ci is planning on having windows bots too. Maybe it will be available sooner than later.

matthew-brett:
Google "try_branch.py" ?  Second hit contains the instructions:

https://github.com/nipy/nibotmi/blob/master/install.rst#trying-a-set-of-changes-on-the-buildbots

arokem:
I agree with @matthew-brett that logging into the machines could be an
advantage.

Does anyone have a good example of circleci configuration for a scipy-ish
project?

arokem:
Aha! So that's where the documentation was hiding! I was looking for
something like "documentation.rst" or a "doc" folder :-)

Garyfallidis:
Okay so I definitely need to reread these instructions. I see they have been updated recently.

Garyfallidis:
Maybe what we should do is give you a list of questions after we have read the doc.

arokem:
We have a perfect opportunity to test the theory that logging into these
machines is helpful! So, @matthew-brett: how do I log into one of these
machines. Specifically, if I wanted to interactively run this failing test
on this machine:

http://nipy.bic.berkeley.edu/builders/dipy-py2.7-wheezy-sparc/builds/225/steps/shell_9/logs/stdio

What are the steps I would need to take?

arokem:
While we're at it, same for this machine:

http://nipy.bic.berkeley.edu/builders/dipy-py2.6-32/builds/646/steps/shell_9/logs/stdio

Thanks!

omarocegueda:
Hi guys! =)
I was just following this discussion, and I wanted to confirm that, as
Matthew said, logging into the buildbots is *extremely* useful. We have
found really weird bugs that happen to occur under very special
circumstances, and logging into the actual failing machine basically gives
us "bug-repro for free". Actually when my friends ask me what is so awesome
about contributing code to Dipy, the buildbots are one of the things that I
show off first =). Anyway, this is another interesting case I need to
investigate:
http://nipy.bic.berkeley.edu/builders/dipy-py2.7-pip/builds/229/steps/shell_5/logs/stdio



On Wed, Nov 25, 2015 at 8:08 PM, Ariel Rokem <notifications@github.com>
wrote:

> We have a perfect opportunity to test the theory that logging into these
> machines is helpful! So, @matthew-brett: how do I log into one of these
> machines. Specifically, if I wanted to interactively run this failing test
> on this machine:
>
>
> http://nipy.bic.berkeley.edu/builders/dipy-py2.7-wheezy-sparc/builds/225/steps/shell_9/logs/stdio
>
> What are the steps I would need to take?
>
>
>
> On Wed, Nov 25, 2015 at 2:54 PM, Eleftherios Garyfallidis <
> notifications@github.com> wrote:
>
> > Maybe what we should do is give you a list of questions after we have
> read
> > the doc.
> >
> > —
> > Reply to this email directly or view it on GitHub
> > <https://github.com/nipy/dipy/pull/782#issuecomment-159749889>.
>
> >
>
> —
> <https://github.com/nipy/dipy/pull/782#issuecomment-159781799>.
>



-- 
-Proverbio chino.
-Chinese proverb.

http://www.cimat.mx/~omar

PR #783, opened 2015-11-25
RF: Enforce integer indices

arokem:
So that we can wade through logs like this:
http://nipy.bic.berkeley.edu/builders/dipy-py2.6-32/builds/640/steps/shell_9/logs/stdio

Also, a bit of PEP8.


Garyfallidis:
For this one. I will have to wait for Travis to finish first.

arokem:
yup - I've only run the test for that one module, and only on one version of python. might have unintended consequences elsewhere...

Garyfallidis:
agreed :)

arokem:
Seems good. I am going to go ahead and merge this, so that we can debug the buildbot that has this issue

PR #784, opened 2015-11-25
TST: If the envrinoment variable is set to 'skip', xvfb_it returns a no-op.

arokem:


arokem:
Might want to wait with this one too, and see that Travis doesn't go nuts over this either.

Garyfallidis:
Okay, will wait.

arokem:
OK - this seems good to go. I want to move ahead with the buildbots, so merging this myself.

PR #785, opened 2015-11-25
WIP: TST: Add OSX to Travis.

arokem:
Maybe to finish after the release... 

To consider: this is only one version of the Mac OS

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,9 @@
 # - There can't be any leading "-"s - All newlines will be removed, so use
 # ";"s
 sudo: false  # To use travis container infrastructure
+os:
+    - linux
+    - osx
 language: python
 cache:
   directories:


theaverageguy:
Can be closed. This is done @arokem

arokem:
Sorry @theaverageguy - I'd like to keep this around until we actually manage to get that working. Haven't had time to get back to this...

skoudoro:
I will look at this next week but I would like to know if we should start to use [CircleCI](https://github.com/marketplace/circleci). Indeed, it will be good if we can split Linux and OSX bots. 

So before starting this, I would like your opinions @arokem @Garyfallidis  @matthew-brett

matthew-brett:
What advantages are there to splitting the builds across CI environments?  It is more complicated to maintain, but will run a little quicker.  Does Circle-CI now have the matrix of testing environments that Travis-CI does?

skoudoro:
My main concern was effectively the time. Indeed, one build/test take approximately 20min, so adding many OSX matrix can increase a lot the total build time. But I agree with the maintenance point. We could use only one OSX matrix but I really think it is not enought (more testing, less surprise 😄).

> Does Circle-CI now have the matrix of testing environments that Travis-CI does?

Yes, CircleCI have matrix of testing environments. Following their website : 

> Note that while CircleCI supports macOS builds, Marketplace plans can only build for Linux.
> 4x build concurrency for open source projects

Any other thought about this ?

PR #786, opened 2015-11-25
DOC: Add the license/copyright to skimage into median_otsu.

### LINKED ISSUE #284: Otsu from scikit-image copied without license
Not a big deal, just need to add the copyright line from

https://github.com/scikit-image/scikit-image/blob/master/LICENSE.txt

arokem:
This supersedes #750 and should finally close #284 (!).

Just cleaning up a bit, while waiting for Travis/buildbots.

### DIFF
--- a/dipy/segment/mask.py
+++ b/dipy/segment/mask.py
@@ -51,10 +51,10 @@ def applymask(vol, mask):
     vol : ndarray
         Array with $V$ dimensions
     mask : ndarray
-        Binary mask.  Has $M$ dimensions where $M <= V$. When $M < V$, we append
-        $V - M$ dimensions with axis length 1 to `mask` so that `mask` will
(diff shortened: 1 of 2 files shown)


dipy/segment/mask.py
```diff
@@ -102,7 +102,7 @@ def bounding_box(vol):
 
 
 def crop(vol, mins, maxs):
-    """ Crops the input volume.
+    """Crops the input volume.
```
Garyfallidis: I think it's fine to have a space after docstring starts. Reads better.

arokem: I set up my editor to enforce PEP257 (https://pypi.python.org/pypi/pep257), and it flagged that.

Garyfallidis: hmm... interesting

PR #789, opened 2015-11-26
BF: Cast input to np.repeat into int32.

arokem:
Only an issue on some systems. For example:
http://nipy.bic.berkeley.edu/builders/dipy-py2.7-wheezy-sparc/builds/225/steps/shell_9/logs/stdio

### DIFF
--- a/dipy/viz/utils.py
+++ b/dipy/viz/utils.py
@@ -159,6 +159,9 @@ def lines_to_vtk_polydata(lines, colors=None):
 
     # Get lines_array in vtk input format
     lines_array = []
+    # Using np.intp (instead of int64), because of a bug in numpy:
+    # https://github.com/nipy/dipy/pull/789
+    # https://github.com/numpy/numpy/issues/4384
     points_per_line = np.zeros([nb_lines], np.intp)
     current_position = 0
     for i in lines_range:
--- a/dipy/viz/utils.py
+++ b/dipy/viz/utils.py
@@ -209,7 +209,7 @@ def lines_to_vtk_polydata(lines, colors=None):
                         np.tile(255 * cols_arr, (nb_points, 1)))
 
             elif cols_arr.ndim == 2:  # map color to each line
-                colors_mapper = np.repeat(lines_range, points_per_line, axis=0)
+                colors_mapper = np.repeat(lines_range, points_per_line.astype(np.int32), axis=0)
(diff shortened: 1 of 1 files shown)


arokem:
This looks like it could potentially be a bug in numpy. Thoughts from anyone?

arokem:
Any thoughts about this one? @matthew-brett - does this look like a numpy bug to you? I believe it's the only remaining error on the bots at the moment

matthew-brett:
Yes, it does look like a bug in numpy, should be so for all 32-bit systems (I can replicate on 32-bit Lintux) - see bug referenced above.

What about initializing `points_per_line` as 32-bit instead?

matthew-brett:
Sorry - better might be to do:

```
points_per_line = np.zeros([nb_lines], int)  # use native index type to avoid casting errors later
```

seberg:
Yeah. You can use `int` (or `np.int_`), but if you want to support 64bits on 64bit platforms (no idea if it matters for this one), you should make it `np.intp`. `np.intp` is the correct type for anything related to array sizes.

matthew-brett:
Are `int` and `intp` different sizes on some platforms?   I'd got used to them being the same, but maybe that isn't always true.

seberg:
No, `int` (the python one or `np.int_`) will map to `long` and on 64 bit windows long is unfortunatly 32bits.

arokem:
OK - using `np.intp` now.

arokem:
Running `try_branch` here: http://nipy.bic.berkeley.edu/builders/dipy-py2.6-32/builds/653

dipy/viz/utils.py
```diff
@@ -159,7 +159,7 @@ def lines_to_vtk_polydata(lines, colors=None):
 
     # Get lines_array in vtk input format
     lines_array = []
-    points_per_line = np.zeros([nb_lines], np.int64)
+    points_per_line = np.zeros([nb_lines], np.intp)
```
matthew-brett: Could do with a comment here explaining why you're using intp.

matthew-brett: I mean why you are using intp instead of np.int64.

arokem: Yup. Linking back here in the comment.

arokem:
That seems to do the trick: http://nipy.bic.berkeley.edu/builders/dipy-py2.6-32/builds/654

PR #790, opened 2015-11-26
BF: Indexing with integers raises warnings on some platforms.

arokem:
This is caught by a test for something else. For example here:
http://nipy.bic.berkeley.edu/builders/dipy-py2.6-32/builds/646/steps/shell_9/logs/stdio

### DIFF
--- a/dipy/reconst/csdeconv.py
+++ b/dipy/reconst/csdeconv.py
@@ -352,11 +352,11 @@ def forward_sdt_deconv_mat(ratio, n, r2_term=False):
         The degree of spherical harmonic function associated with each row of
         the deconvolution matrix. Only even degrees are allowed.
     r2_term : bool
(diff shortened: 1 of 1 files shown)


omarocegueda:
Wow!, there are many places where people is attempting to index with non-integers... I can help with this one, if you are not doing it already, so you can focus on more important ones. Are you already working on this?

arokem:
It's actually not that bad. There are just a few more places where this is
raised. See warnings here:

http://nipy.bic.berkeley.edu/builders/dipy-py2.6-32/builds/646/steps/shell_9/logs/stdio

If you would like to fix these up, that would be a great help. In the
meanwhile, can we merge this one, to see that it helps fix the errors on
the bots?

omarocegueda:
Sure!, I'll start with the other ones

PR #794, opened 2015-11-26
RF: Allow setting the verbosity of the AffineRegistration while running it

arokem:
There might be a better way to do this. Suggestions are most welcome.


omarocegueda:
Thank you Ariel!

Issue #795, opened 2015-11-27
dipy.tracking: Converting an array with ndim > 0 to an index will result in an error


omarocegueda:
This warning is triggered from this test in dipy.tracking:
https://github.com/nipy/dipy/blob/master/dipy/tracking/tests/test_markov.py#L256
here, `self.n` is a list with only one element. Changing to `self.n[0]` passes the tests, but I'm not sure if that is the intended behaviour.

Here is the log:
http://nipy.bic.berkeley.edu/builders/dipy-py2.6-32/builds/651/steps/shell_9/logs/stdio

arokem:
Do we still need this issue? FWIW, this same part of code is also addressed in #858, on https://github.com/arokem/dipy/commit/6a6132c31f8afa172ccb24b7963f7227b3671a10

skoudoro:
This warning does not pop up anymore and `test_markov.py` does not exist anymore. So I'm closing this issue

PR #796, opened 2015-11-27
Fix deprecation warnings.

omarocegueda:
Fixes more deprecation warnings. Here is the resulting log after running `try_branch`:
http://nipy.bic.berkeley.edu/builders/dipy-py2.6-32/builds/651/steps/shell_9/logs/stdio

Most of the remaining warnings are caused by #795


arokem:
Thanks for doing this!

PR #797, opened 2015-11-30
Update csdeconv.py

samuelstjean:
In [2]: iter = 8

In [3]: range(1, iter)
Out[3]: [1, 2, 3, 4, 5, 6, 7]

In [4]: len(range(1, iter))
Out[4]: 7

So now it should do 8 iterations by default, which is what I think is originally intended.

### DIFF
--- a/dipy/reconst/csdeconv.py
+++ b/dipy/reconst/csdeconv.py
@@ -964,7 +964,7 @@ def recursive_response(gtab, data, mask=None, sh_order=8, peak_thr=0.01,
     where_dwi = lazy_index(~gtab.b0s_mask)
     response_p = np.ones(len(n))
 
-    for num_it in range(1, iter):
+    for num_it in range(iter):
         r_sh_all = np.zeros(len(n))
         csd_model = ConstrainedSphericalDeconvModel(gtab, res_obj,
                                                     sh_order=sh_order)
@@ -983,7 +983,7 @@ def recursive_response(gtab, data, mask=None, sh_order=8, peak_thr=0.01,
         data = data[single_peak_mask]
         dirs = dirs[single_peak_mask]
 
-        for num_vox in range(0, data.shape[0]):
+        for num_vox in range(data.shape[0]):
             rotmat = vec2vec_rotmat(dirs[num_vox, 0], np.array([0, 0, 1]))
 
             rot_gradients = np.dot(rotmat, gtab.gradients.T).T


dipy/reconst/csdeconv.py
```diff
@@ -964,7 +964,7 @@ def recursive_response(gtab, data, mask=None, sh_order=8, peak_thr=0.01,
     where_dwi = lazy_index(~gtab.b0s_mask)
     response_p = np.ones(len(n))
 
-    for num_it in range(1, iter):
+    for num_it in range(iter):
```
arokem: While you're at it, would you mind changing the name of this variable? `iter` is a key-word (a function) in the language.

arokem:
@ChantalTax : could you please confirm this was your intention?

dipy/reconst/csdeconv.py
samuelstjean: That would break the backward compatibility of the function (it's an input), including the script I am revamping which lead me to see this. Unless we add deprecation warning and everything to it it might break more stuff than it could fix.

arokem: Yes. Good point. Sigh.

arokem:
Hey @ChantalTax - could you please confirm that this is indeed a bug fix?

ChantalTax:
Hey! This seems indeed a bug fix, thanks!

arokem:
Great. Thanks @ChantalTax for confirming! And thanks @samuelstjean for noticing this! Is there any test you could write that would prevent this bug from recurring?

samuelstjean:
er, not really, this just arose from the fact that matlab indexing starts from 1 and python indexing starts from 0 I think, it's just a small implementation detail overlook, not a real bug per se.

arokem:
I understand that. How did you run into this? Just by running the code, or
were you getting some odd results?

arokem:
Sorry, meant to say "just by reading the code"

samuelstjean:
Reading the cod,e since it returns a whole object and the auto response returns a tuple, it was breaking a bunch of debug prints in the calling script I have to run CSD.

arokem:
OK -- thanks. I think this is fine to merge, and will wait until early next week to do so, in case anyone else has anything to add. Thanks for noticing this!

arokem:
Whoops - forgot to hit the merge button on this one! 

:christmas_tree:

PR #798, opened 2015-12-01
DOC: Installation, added RNH to list of contributors.

arokem:


doc/installation.rst
```diff
@@ -178,52 +172,39 @@ we'll also call it the ``<dipy root>`` directory.
...
 
-  [build]
-  compiler=mingw32
+When adding dipy_ to the ``PYTHONPATH``, we usually add the ``PYTHONPATH`` at
+the end of ``~/.bashrc`` or (OSX) ``~/.bash_profile`` so we don't need to
+retype it every time. This should look something like::
 
-Save this into your system python ``distutils`` directory as ``distutils.cfg``.
-This will be something like ``C:\Python26\Lib\distutils\distutils.cfg``.
+  export PYTHONPATH=/home/user_dir/Devel/dipy:/home/user_dir/Devel/nibabel
```
Garyfallidis: It's not necessary to have nibabel in the PYTHONPATH. But you need to have nibabel installed.

arokem: Good catch. I also figure they should add `$PYTHONPATH` at the end.

doc/installation.rst
```diff
@@ -255,39 +237,76 @@ Same as above but use yum rather than apt-get when necessary.
...
+configuration (usually in ``~/.bash_profile``), the following::
+
+	export PATH="/usr/local/Cellar/gcc/5.2.0/bin/gcc-5:$PATH
+
+If you are already using the Homebrew Python, or the standard python.org Python, 
+you will need to run::
+	
+	brew install clang-omp
+
+And then edit the ``setup.py`` file to include the following line::
```
Garyfallidis: Where to include it? Maybe it will be nice to say at which line or at the top of the file or something.

doc/installation.rst
```diff
@@ -255,39 +237,76 @@ Same as above but use yum rather than apt-get when necessary.
...
+Next, if you are using Anaconda, you will need to run the following::
+    
+	brew reinstall gcc --without-multilib
+	
+This should take about 45 minutes to complete. Then add to your bash 
+configuration (usually in ``~/.bash_profile``), the following::
+
+	export PATH="/usr/local/Cellar/gcc/5.2.0/bin/gcc-5:$PATH
+
+If you are already using the Homebrew Python, or the standard python.org Python,
```
Garyfallidis: The document gives the impression that you need to do both (GCC-part and CLANG-part). Explain the difference between the two options and work on the transition. It's not about if you are using the homebrew python or not. The first option is about using gcc which doesn't need a change of the setup.py file. And the other is about using the new compiler and it needs change of the setup. It would be nice if we could make the second part automatic too. But that is future work... or maybe Apple will fix this for us.

arokem:
Thanks for comments. Changes accordingly.

doc/installation.rst
```diff
@@ -255,39 +237,88 @@ Same as above but use yum rather than apt-get when necessary.
...
+OpenMP with OSX
+---------------
+OpenMP_ is a standard library for efficient multithreaded applications. This 
+is used in Dipy for speeding up many different parts of the library (e.g., denoising 
+and bundle registration). If you do not have an OpenMP-enabled compiler, you can 
+still compile Dipy from source using the above instructions, but it might not take 
+advantage of the multithreaded parts of the code. To be able to compile 
+Dipy from source with OpenMP on Mac OSX, you will have to do a few more things. First 
+of all, you will need to install the Homebrew_ package manager. Next you will need
+to install and configure the compiler. This depends on your python installation:
```
Garyfallidis: Add this after the compiler. 
```
You have two options using the GCC compiler or CLANG compiler. 
```
makes the text flow ...

arokem: Done.

Garyfallidis:
Thx! :+1:

Issue #803, opened 2015-12-05
Windows 7 Pro VM Python 2.7 gives 5 test errors with latest release 0.10.1


Garyfallidis:
After doing in Windows 7 Pro (64 bit) (Python 2.7, Anaconda 2.0 with Numpy 1.8.1, Cython  0.20.1 and Scipy 0.14.0)
```
pip install dipy
import dipy
dipy.test()
```
I get the following errors
```
======================================================================
ERROR: dipy.segment.tests.test_feature.test_feature_extract
Traceback (most recent call last):
File "C:\Users\Eleftherios\Anaconda\lib\site-packages\nose\case.py", line 197, in runTest
self.test(*self.arg)
File "C:\Users\Eleftherios\Anaconda\lib\site-packages\dipy\segment\tests\test_feature.py", line 231, in test_feature_e
features = extract(feature, streamlines)
File "dipy/segment/featurespeed.pyx", line 371, in dipy.segment.featurespeed.extract (dipy\segment\featurespeed.c:6302
File "dipy/segment/featurespeed.pyx", line 394, in dipy.segment.featurespeed.extract (dipy\segment\featurespeed.c:5973
File "dipy/segment/featurespeed.pyx", line 363, in dipy.segment.featurespeed.infer_shape (dipy\segment\featurespeed.c:
File "dipy/segment/featurespeed.pyx", line 46, in dipy.segment.featurespeed.Feature.c_infer_shape (dipy\segment\featur
7)
TypeError: object of type 'long' has no len()
======================================================================
======================================================================
ERROR: dipy.tracking.tests.test_life.test_voxel2streamline
Traceback (most recent call last):
File "C:\Users\Eleftherios\Anaconda\lib\site-packages\nose\case.py", line 197, in runTest
self.test(*self.arg)
File "C:\Users\Eleftherios\Anaconda\lib\site-packages\dipy\tracking\tests\test_life.py", line 89, in test_voxel2stream
v2f, v2fn = life.voxel2streamline(streamline, False, affine)
File "C:\Users\Eleftherios\Anaconda\lib\site-packages\dipy\tracking\life.py", line 313, in voxel2streamline
return _voxel2streamline(transformed_streamline, unique_idx)
File "dipy/tracking/vox2track.pyx", line 20, in dipy.tracking.vox2track._voxel2streamline (dipy\tracking\vox2track.c:1
ValueError: Buffer dtype mismatch, expected 'npy_intp' but got 'long'
======================================================================
======================================================================
[...]

PR #806, opened 2015-12-10
Relax regression tests

omarocegueda:
This is a proposal for relaxing the regression tests of the registration module.
The original idea about the regression tests was to detect subtle changes in the registration module which could affect the overall registration accuracy. Current tests attempt to reproduce the evolution of the energy profile along the full optimization, this is a **very** strict test because it basically means attempting to perform exactly the same sequence of computations with the same precision. However, we have been facing a lot of problems reproducing the full energy profile across platforms, especially in  64 bits windows.

Instead of requiring the full sequence of computations being the same, this change only measures the final registration result, so the optimization may take different paths depending on the platform, but the final result should be of good quality.

### DIFF
--- a/dipy/align/tests/test_imwarp.py
+++ b/dipy/align/tests/test_imwarp.py
@@ -6,7 +6,6 @@ from numpy.testing import (assert_equal,
                            assert_array_equal,
                            assert_array_almost_equal,
                            assert_raises)
-from ...__config__ import USING_VC_SSE2, USING_GCC_SSE2
 from ...data import get_data
 from .. import floating
 from .. import imwarp as imwarp
@@ -15,7 +14,6 @@ from .. import vector_fields as vfu
 from .. import VerbosityLevels
 from ..imwarp import DiffeomorphicMap
 
-NO_SSE2 = not (USING_VC_SSE2 or USING_GCC_SSE2)
 
 def test_mult_aff():
     r""" Registration: test matrix multiplication using None as identity
@@ -370,7 +368,6 @@ def simple_callback(sdr, status):
         sdr.ITER_END_CALLED = 1
 
 
-@npt.dec.skipif(NO_SSE2)
 def test_ssd_2d_demons():
     r''' Registration: test 2D SyN with SSD metric, demons-like optimizer
 
@@ -433,7 +430,6 @@ def test_ssd_2d_demons():
     assert_equal(optimizer.ITER_END_CALLED, 1)
 
 
-@npt.dec.skipif(NO_SSE2)
 def test_ssd_2d_gauss_newton():
     r''' Registration: test 2D SyN with SSD metric, Gauss-Newton optimizer
 
@@ -584,7 +580,6 @@ def get_synthetic_warped_circle(nslices):
     return circle_3d, wcircle_3d
 
 
-@npt.dec.skipif(NO_SSE2)
 def test_ssd_3d_demons():
     r''' Registration: test 3D SyN with SSD metric, demons-like optimizer
 
@@ -626,7 +621,6 @@ def test_ssd_3d_demons():
     assert(reduced > 0.9)
 
 
-@npt.dec.skipif(NO_SSE2)
 def test_ssd_3d_gauss_newton():
     r''' Registration: test 3D SyN with SSD metric, Gauss-Newton optimizer
 
@@ -668,8 +662,6 @@ def test_ssd_3d_gauss_newton():
     assert(reduced > 0.9)
 
 
-
-@npt.dec.skipif(NO_SSE2)
 def test_cc_2d():
     r''' Registration: test 2D SyN with CC metric
 
@@ -706,7 +698,6 @@ def test_cc_2d():
     assert(reduced > 0.9)
 
 
-@npt.dec.skipif(NO_SSE2)
 def test_cc_3d():
     r''' Registration: test 3D SyN with CC metric
 
@@ -751,7 +742,6 @@ def test_cc_3d():
     assert(reduced > 0.9)
 
 
-@npt.dec.skipif(NO_SSE2)
 def test_em_3d_gauss_newton():
     r''' Registration: test 3D SyN with EM metric, Gauss-Newton optimizer
 
@@ -799,7 +789,6 @@ def test_em_3d_gauss_newton():
     assert(reduced > 0.9)
 
 
-@npt.dec.skipif(NO_SSE2)
 def test_em_2d_gauss_newton():
     r''' Registration: test 2D SyN with EM metric, Gauss-Newton optimizer
 
@@ -841,7 +830,6 @@ def test_em_2d_gauss_newton():
     assert(reduced > 0.9)
 
 
-@npt.dec.skipif(NO_SSE2)
 def test_em_3d_demons():
     r''' Registration: test 3D SyN with EM metric, demons-like optimizer
 
@@ -889,7 +877,6 @@ def test_em_3d_demons():
     assert(reduced > 0.9)
 
 
-@npt.dec.skipif(NO_SSE2)
 def test_em_2d_demons():
     r''' Registration: test 2D SyN with EM metric, demons-like optimizer
 
--- a/dipy/align/tests/test_imwarp.py
+++ b/dipy/align/tests/test_imwarp.py
@@ -16,7 +16,7 @@ from ..imwarp import DiffeomorphicMap
 
 
 def test_mult_aff():
-    r""" Registration: test matrix multiplication using None as identity
+    r""" Test matrix multiplication using None as identity
     """
     A = np.array([[1.0, 2.0], [3.0, 4.0]])
     B = np.array([[2.0, 0.0], [0.0, 2.0]])
@@ -36,7 +36,7 @@ def test_mult_aff():
 
 
 def test_diffeomorphic_map_2d():
-    r""" Registration: test 2D DiffeomorphicMap
+    r""" Test 2D DiffeomorphicMap
 
     Creates a random displacement field that exactly maps pixels from an
     input image to an output image. First a discrete random assignment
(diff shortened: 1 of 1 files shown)


arokem:
Another option is to run this on other platforms as well, and save the different energy trajectories, as well as the final results. Would that work? If we do decide to go with these changes, you would also need to at least change the docstrings of the functions to explain what they are checking now (e.g. for `test_ssd_2d_gauss_newton`).

arokem:
Also, have you run a try_branch with this on the bots that were having issues with these tests?

omarocegueda:
That's right, we may add a different profile for each case, but I'm afraid that's not going to be scalable. This guy summarizes very well the difficulties of trying to reproduce the results of floating point operations in different platforms:

https://randomascii.wordpress.com/2012/03/21/intermediate-floating-point-precision/

So the result of the floating point operations depends on too many factors, including of course the architecture, the compiler version, and the particular optimizations and flags we specify.

After a long conversation with @matthew-brett , we finally achieved reproducibility assuming SSE2 instructions were available (in summary, by asking the compiler to use SSE2 instructions, we were forcing intermediate operations to be performed with 64 bits and not 80 bits). This worked very well in Linux, but MSVC makes things particularly problematic, they textually say:
https://msdn.microsoft.com/en-us/library/7t5yh4fd.aspx
>The optimizer chooses when and how to use the SSE and SSE2 instructions when /arch is specified. It uses SSE and SSE2 instructions for some scalar floating-point computations when it determines that it is faster to use the SSE/SSE2 instructions and registers instead of the x87 floating-point register stack. As a result, your code **may** actually use a mixture of both x87 and SSE/SSE2 for floating-point computations. Also, with /arch:SSE2, SSE2 instructions can be used for some 64-bit integer operations.

The crucial word here is "**may**" =(. So, it is unpredictable and we ended up adding a new profile for windows. My concern is that it may take a considerable amount of time to update the profiles each time we add a new buildbot and/or each time we want to make an enhancement to the algorithms and it may unnecessarily delay the releases.

Having said that, this is just a proposal, if you guys thing it is worth the effort adding the extra profiles, I can go ahead and update them. Please let me know what you think.

omarocegueda:
The regression tests pass on `dipy-bdist64-35-win10` after relaxing:

http://nipy.bic.berkeley.edu/builders/dipy-bdist64-35-win10/builds/1/steps/shell_11/logs/stdio

I will try on the other two when they are back online.

arokem:
Is this branch up-to-date with current master? There are some TripWire
errors in there that I thought we already fixed here:

https://github.com/nipy/dipy/pull/801

But it might also be something else...

omarocegueda:
Yes, I re-based the branch just before doing the pull request (and I doube-check right now). The other 64-bit buildbots are back online, I'll run try_branch now.

omarocegueda:
I just updated the docstrings.

These tests pass on `dipy-bdist64-35` too:
http://nipy.bic.berkeley.edu/builders/dipy-bdist64-35/builds/6/steps/shell_11/logs/stdio

but it still reports the Tripwire error.

arokem:
That's weird. Many of these errors do not appear on Windows 8, which is what is currently running on Appveyor: https://ci.appveyor.com/project/arokem/dipy/build/1.0.508/job/4kn2da2a15boeprd

arokem:
But I have zero understanding of the differences between these systems, so I am not sure how to proceed. I did spend a few hours yesterday trying to build dipy on an AWS Windows 7 machine. Unpleasant.

matthew-brett:
Is it possible the extra errors are caused by installing dipy as a wheel rather than `python setup.py install` ?

arokem:
That makes sense. Where/how is the wheel `dipy-0.11.0.dev0-cp35-none-win_amd64.whl` generated?

matthew-brett:
You can follow along the build steps in - say - http://nipy.bic.berkeley.edu/builders/dipy-bdist64-27/builds/5

The wheel gets made with `python setup.py bdist_wheel`.

arokem:
What shell is being used here? In the Windows installation I have access to
there are now (after installing VS 2015), at least three: CMD, PowerShell,
and the MSBuild Command Prompt for VS2015.

matthew-brett:
`cmd` I think - but is that relevant?

arokem:
I don't know. Is it?

matthew-brett:
I don't think it is relevant, but I'm happy to be corrected.

dipy/align/tests/test_imwarp.py
```diff
@@ -516,6 +491,62 @@ def test_ssd_2d_gauss_newton():
...
+    The image is deformed under an invertible field, and a 3D volume is
+    generated as follows:
+    the first and last `nslices`//3 slices are filled with zeros
+    to simulate background. The remaining middle slices are filled with
+    copies of the deformed `image` under the action of the invertible
+    field.
+
+    Parameters
+    ----------
+    image: 2d array shape(r, c)
```
Garyfallidis: add space after image

dipy/align/tests/test_imwarp.py
```diff
@@ -516,6 +491,62 @@ def test_ssd_2d_gauss_newton():
...
+        the image to be deformed
+    nslices: int
```
Garyfallidis: same

dipy/align/tests/test_imwarp.py
```diff
@@ -15,11 +14,9 @@
 from .. import VerbosityLevels
 from ..imwarp import DiffeomorphicMap
 
-NO_SSE2 = not (USING_VC_SSE2 or USING_GCC_SSE2)
 
 def test_mult_aff():
-    r"""mult_aff from imwarp returns the matrix product A.dot(B) considering
-    None as the identity
+    r""" Registration: test matrix multiplication using None as identity
```
Garyfallidis: Please remove ``Registration:`` from the title of the dosctrings.

dipy/align/tests/test_imwarp.py
```diff
@@ -516,6 +491,62 @@ def test_ssd_2d_gauss_newton():
...
+    nslices: int
+        the number of slices in the final volume
+    b, m : float
+        parameters of the harmonic field (as in [1]).
+
+    Returns
+    -------
+    vol : array shape(r, c, `nslices`)
+        the volumed generated using the undeformed image
+    wvol :
```
Garyfallidis: missing dtype

dipy/align/tests/test_imwarp.py
```diff
@@ -586,36 +612,21 @@ def test_ssd_3d_demons():
...
-    Register a stack of circles ('cylinder') before and after warping them with
-    a synthetic diffeomorphism. This test is intended to detect regressions
-    only: we saved the energy profile (the sequence of energy values at each
-    iteration) of a working version of SSD in 3D using the Gauss-Newton step,
-    and this test checks that the current energy profile matches the saved
-    one. The validation of the "working version" was
-    done by registering the 18 manually annotated T1 brain MRI database IBSR
-    with each other and computing the jaccard index for all 31 common
-    anatomical regions.
+    r''' Registration: test 3D SyN with SSD metric, Gauss-Newton optimizer
```
Garyfallidis: Same as before please remove ```Registration:``` from all tests. You are in the alignment module so these tests are of course about registration.

dipy/align/tests/test_imwarp.py
```diff
@@ -953,24 +908,16 @@ def test_em_2d_demons():
...
+    reduced = 1.0 - final_energy/starting_energy
+
+    assert(reduced > 0.9)
+
 
 if __name__=='__main__':
-    test_scale_space_exceptions()
     test_optimizer_exceptions()
     test_mult_aff()
     test_diffeomorphic_map_2d()
```
Garyfallidis: Please replace calling all the functions in the main and just use 
```run_module_suite``` and that will call all available tests in the file.

dipy/align/tests/test_imwarp.py
omarocegueda: Added

dipy/align/tests/test_imwarp.py
omarocegueda: Added

dipy/align/tests/test_imwarp.py
omarocegueda: removed

dipy/align/tests/test_imwarp.py
omarocegueda: fixed

dipy/align/tests/test_imwarp.py
omarocegueda: Removed

dipy/align/tests/test_imwarp.py
omarocegueda: removed

arokem:
The TripWire error should now be fixed with the recent merge of #808.

omarocegueda:
Great!, I just re-based and ran try-branch. The remaining failures are #803 and #810:
http://nipy.bic.berkeley.edu/builders/dipy-bdist64-27/builds/9/steps/shell_11/logs/stdio
http://nipy.bic.berkeley.edu/builders/dipy-bdist64-35-win10/builds/5/steps/shell_11/logs/stdio
thanks!

Garyfallidis:
@MarcCote can you check Omar's last message on top? One of the errors are from the new clustering framework. Thx!

matthew-brett:
Eleftherios - I guess you mean this error : https://github.com/nipy/dipy/issues/813 ?

Garyfallidis:
Yes, I mean this error @matthew-brett Is this already resolved?

matthew-brett:
Not as far as I know.

Garyfallidis:
ok

Garyfallidis:
Thx @omarocegueda :)

PR #808, opened 2015-12-13
MRG: pull in discarded changes from maintenance

matthew-brett:
I was puzzled to see that current master doesn't have the `Tripwire` fixes
Ariel did on the maintenance branch.

After a bit of debugging, I think what happened is you (Ariel) did a ``git
merge -s ours`` on the maintenance branch, at merge commit 6d31467 .  This has
the effect of marking that part of the maintenance history as known, but
throwing away the changes from the maintenance branch.  It's the right thing to
do if you are making changes you *don't* want in mainline, but the wrong thing
to do if you do want the changes.

I don't know how to undo that in a nice way, so I just cherry-picked back the
relevant commits.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ python:
     - 3.2
     - 3.3
     - 3.4
+    - 3.5
 matrix:
   include:
     - python: 2.7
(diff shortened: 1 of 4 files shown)


arokem:
Huh. I thought this would pull anything except for things that cause merge conflicts. Thanks for catching this...

dimrozakis:
@arokem `git merge -s ours` uses the `ours` merging strategy that ignores any changes in all branches except the current head. What you wanted is probably is `git merge -X ours` which uses the default `recursive` strategy with the `ours` option, that keeps changes from all branches and favors the current head in case of conflict between branches.

arokem:
Ah! Thanks @dimrozakis! Clearly this is not something I do every day...

PR #809, opened 2015-12-13
MRG: refactor and test setup.py

matthew-brett:
Refactor setup.py to work better with pip.  Allow setuptools / pip installs
into systems that don't have Cython (Cython gets installed automatically).

Update requirements in requirements.txt file and info.py.

Check installation via sdist and wheels.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,18 +5,12 @@ import numpy as np
 import os
 import sys
 from copy import deepcopy
-from os.path import join as pjoin, dirname
+from os.path import join as pjoin, dirname, exists
 from glob import glob
 
 # BEFORE importing distutils, remove MANIFEST. distutils doesn't properly
 # update it when the contents of directories change.
-if os.path.exists('MANIFEST'): os.remove('MANIFEST')
-
-
-# Get version and release info, which is all stored in dipy/info.py
-ver_file = os.path.join('dipy', 'info.py')
-# Use exec for compabibility with Python 3
-exec(open(ver_file).read())
+if exists('MANIFEST'): os.remove('MANIFEST')
 
 # force_setuptools can be set from the setup_egg.py script
 if not 'force_setuptools' in globals():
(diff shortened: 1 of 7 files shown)


arokem:
Any idea why this hasn't triggered a Travis build?

matthew-brett:
Yes, that's because I turned off the build for this commit because it was a trivial modification of the one tested here :  https://travis-ci.org/matthew-brett/dipy/builds/96530773

Diff:

```
diff --git a/.gitignore b/.gitignore
index bacf6a7..579881e 100644
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,4 @@ __config__.py
 .DS_Store
 .coverage
 .buildbot.patch
+.eggs/
```

matthew-brett:
I pushed up a little refactor which has triggered a build.

setup.py
```diff
@@ -144,39 +138,53 @@ def package_check(*args, **kwargs):
...
+
+for name, min_ver, req_type, heavy in DEPS:
+    found_ver = get_pkg_version(name)
+    ver_err_msg = version_error_msg(name, found_ver, min_ver)
+    if not using_setuptools:
+        if ver_err_msg != None:
+            raise RuntimeError(ver_err_msg)
+    else:  # Using setuptools
+        # Add packages to given section of setup/install_requires
+        if ver_err_msg != None or not heavy:
```
arokem: What does `heavy` mean in this context?

arokem:
Just took a look through everything. It seems good, though I have to admit that the distinctions between setuptools and distutils make my head spin. And the additional checks seem good to me. I had one small question, but otherwise, happy to merge this.

setup.py
matthew-brett: "Heavy" means that it's a pain to install from scratch.

In the code, it means that, if the package is installed already and has a
good-enough version, we don't add it to the setuptools requirements.

This avoids two undesirable cases:

* installing e..g. scipy with pip just because the current installation was
not installed with pip / setuptools;
* upgrading e..g. scipy when upgrading dipy (pip by default upgrades all
dependencies at the same time as the main package).

arokem:
Thanks for the explanation. If no one else pipes up, I will merge this in a couple of days.

PR #811, opened 2015-12-16
Extract values from an image based on streamline coordinates.

arokem:
Still need to test affine transformations

### DIFF
--- a/dipy/tracking/streamline.py
+++ b/dipy/tracking/streamline.py
@@ -366,9 +366,9 @@ def _extract_vals(data, streamlines, affine=None, threedvec=False):
         if affine is not None:
             streamlines = ut.move_streamlines(streamlines,
                                               np.linalg.inv(affine))
+
         vals = []
         for sl in streamlines:
-
             if threedvec:
                 vals.append(list(vfu.interpolate_vector_3d(data,
                                  sl.astype(np.float))[0]))
--- a/dipy/tracking/streamline.py
+++ b/dipy/tracking/streamline.py
@@ -416,7 +416,9 @@ def values_from_volume(data, streamlines, affine=None):
 
     affine : ndarray, shape (4, 4)
         Affine transformation from voxels (image coordinates) to streamlines.
-        Default: identity.
+        Default: identity. For example, if no affine is provided and the first
+        coordinate of the first streamline is ``[1, 0, 0]``, data[1, 0, 0] would
+        be returned as the value for that streamline coordinate
 
     Return
     ------
--- a/dipy/tracking/streamline.py
+++ b/dipy/tracking/streamline.py
@@ -369,16 +369,20 @@ def _extract_vals(data, streamlines, affine=None, threedvec=False):
             isinstance(streamlines, types.GeneratorType)):
         vals = []
         for sl in streamlines:
+
             if threedvec:
-                vals.append(list(vfu.interpolate_vector_3d(data, sl)[0]))
+                vals.append(list(vfu.interpolate_vector_3d(data,
+                                 sl.astype(np.float))[0]))
             else:
-                vals.append(list(vfu.interpolate_scalar_3d(data, sl)[0]))
+                vals.append(list(vfu.interpolate_scalar_3d(data,
+                                 sl.astype(np.float))[0]))
 
     elif isinstance(streamlines, np.ndarray):
         sl_shape = streamlines.shape
         sl_cat = streamlines.reshape(sl_shape[0] * sl_shape[1], 3)
         if affine is not None:
-            sl_cat = np.dot(sl_cat, affine[:3, :3]) + affine[:3, 3]
+            sl_cat = (np.dot(sl_cat, affine[:3, :3]) +
+                      affine[:3, 3]).astype(np.float)
 
         # So that we can index in one operation:
         if threedvec:
--- a/dipy/tracking/streamline.py
+++ b/dipy/tracking/streamline.py
@@ -379,10 +379,10 @@ def _extract_vals(data, streamlines, affine=None, threedvec=False):
 
     elif isinstance(streamlines, np.ndarray):
         sl_shape = streamlines.shape
-        sl_cat = streamlines.reshape(sl_shape[0] * sl_shape[1], 3)
+        sl_cat = streamlines.reshape(sl_shape[0] *
+                                     sl_shape[1], 3).astype(np.float)
         if affine is not None:
-            sl_cat = (np.dot(sl_cat, affine[:3, :3]) +
-                      affine[:3, 3]).astype(np.float)
+            sl_cat = (np.dot(sl_cat, affine[:3, :3]) + affine[:3, 3])
 
         # So that we can index in one operation:
         if threedvec:
--- a/dipy/tracking/streamline.py
+++ b/dipy/tracking/streamline.py
@@ -387,6 +387,11 @@ def _extract_vals(data, streamlines, affine=None, threedvec=False):
             vals = np.array(vfu.interpolate_scalar_3d(data, sl_cat)[0])
         vals = np.reshape(vals, (sl_shape[0], sl_shape[1], -1)).squeeze()
 
+    else:
+        raise RuntimeError("Extracting values from a volume ",
+                           "requires streamlines input as an array, ",
+                           "a list of arrays, or a streamline generator.")
+
     return vals
 
 
--- a/dipy/tracking/streamline.py
+++ b/dipy/tracking/streamline.py
@@ -329,13 +329,16 @@ def orient_by_rois(streamlines, roi1, roi2, affine=None, copy=True):
     return new_sl
 
 
-def scalar_values(data, streamlines, affine=None):
-    """Extract values of a scalar along each streamline.
+def _extract_vals(data, streamlines, affine=None, threedvec=False):
+    """
+    Helper function for use with `values_from_volume`.
 
     Parameters
     ----------
-    data : 3D array
-        Scalar values to be extracted.
+    data : 3D or 4D array
+        Scalar (for 3D) and vector (for 4D) values to be extracted. For 4D
+        data, interpolation will be done on the 3 spatial dimensions in each
+        volume.
 
     streamlines : ndarray or list
         If array, of shape (n_streamlines, n_nodes, 3)
@@ -345,6 +348,12 @@ def scalar_values(data, streamlines, affine=None):
     affine : ndarray, shape (4, 4)
         Affine transformation from voxels (image coordinates) to streamlines.
         Default: identity.
+
+    threedvec : bool
+        Whether the last dimension has length 3. This is a special case in
+        which we can use :func:`vfu.interpolate_vector_3d` for the
+        interploation of 4D volumes without looping over the elements of the
+        last dimension.
 
     Return
     ------
@@ -360,15 +369,75 @@ def scalar_values(data, streamlines, affine=None):
             isinstance(streamlines, types.GeneratorType)):
         vals = []
         for sl in streamlines:
-            vals.append(vfu.interpolate_scalar_3d(data, sl)[0])
+            if threedvec:
+                vals.append(list(vfu.interpolate_vector_3d(data, sl)[0]))
+            else:
+                vals.append(list(vfu.interpolate_scalar_3d(data, sl)[0]))
 
     elif isinstance(streamlines, np.ndarray):
         sl_shape = streamlines.shape
         sl_cat = streamlines.reshape(sl_shape[0] * sl_shape[1], 3)
         if affine is not None:
             sl_cat = np.dot(sl_cat, affine[:3, :3]) + affine[:3, 3]
+
         # So that we can index in one operation:
-        vals = vfu.interpolate_scalar_3d(data, sl_cat)[0]
-        vals = np.reshape(vals, (sl_shape[0], sl_shape[1]))
+        if threedvec:
+            vals = np.array(vfu.interpolate_vector_3d(data, sl_cat)[0])
+        else:
+            vals = np.array(vfu.interpolate_scalar_3d(data, sl_cat)[0])
+        vals = np.reshape(vals, (sl_shape[0], sl_shape[1], -1)).squeeze()
 
     return vals
+
+
+def values_from_volume(data, streamlines, affine=None):
+    """Extract values of a scalar/vector along each streamline from a volume.
+
+    Parameters
+    ----------
+    data : 3D or 4D array
+        Scalar (for 3D) and vector (for 4D) values to be extracted. For 4D
+        data, interpolation will be done on the 3 spatial dimensions in each
+        volume.
+
+    streamlines : ndarray or list
+        If array, of shape (n_streamlines, n_nodes, 3)
+        If list, len(n_streamlines) with (n_nodes, 3) array in
+        each element of the list.
+
+    affine : ndarray, shape (4, 4)
+        Affine transformation from voxels (image coordinates) to streamlines.
+        Default: identity.
+
+    Return
+    ------
+    array or list (depending on the input) : values interpolate to each
+        coordinate along the length of each streamline
+    """
+    data = np.asarray(data)
+    if len(data.shape) == 4:
+        if data.shape[-1] == 3:
+            return _extract_vals(data, streamlines, affine=affine,
+                                 threedvec=True)
+        if isinstance(streamlines, types.GeneratorType):
+            streamlines = list(streamlines)
+        vals = []
+        for ii in range(data.shape[-1]):
+            vals.append(_extract_vals(data[..., ii], streamlines,
+                        affine=affine))
+
+        if isinstance(vals[-1], np.ndarray):
+            return np.swapaxes(np.array(vals), 2, 1).T
+        else:
+            new_vals = []
+            for sl_idx in range(len(streamlines)):
+                sl_vals = []
+                for ii in range(data.shape[-1]):
+                    sl_vals.append(vals[ii][sl_idx])
+                new_vals.append(np.array(sl_vals).T)
+            return new_vals
+
+    elif len(data.shape) == 3:
+        return _extract_vals(data, streamlines, affine=affine)
+    else:
+        raise ValueError("Data needs to have 3 or 4 dimensions")
--- a/dipy/tracking/streamline.py
+++ b/dipy/tracking/streamline.py
@@ -361,12 +361,11 @@ def _extract_vals(data, streamlines, affine=None, threedvec=False):
         coordinate along the length of each streamline
     """
     data = data.astype(np.float)
-    if affine is not None:
-        streamlines = ut.move_streamlines(streamlines,
-                                          np.linalg.inv(affine))
-
     if (isinstance(streamlines, list) or
             isinstance(streamlines, types.GeneratorType)):
+        if affine is not None:
+            streamlines = ut.move_streamlines(streamlines,
+                                              np.linalg.inv(affine))
         vals = []
         for sl in streamlines:
 
@@ -381,8 +380,11 @@ def _extract_vals(data, streamlines, affine=None, threedvec=False):
         sl_shape = streamlines.shape
         sl_cat = streamlines.reshape(sl_shape[0] *
                                      sl_shape[1], 3).astype(np.float)
+
         if affine is not None:
-            sl_cat = (np.dot(sl_cat, affine[:3, :3]) + affine[:3, 3])
+            inv_affine = np.linalg.inv(affine)
+            sl_cat = (np.dot(sl_cat, inv_affine[:3, :3]) +
+                      inv_affine[:3, 3])
 
         # So that we can index in one operation:
         if threedvec:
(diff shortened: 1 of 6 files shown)


arokem:
Fixed up for elderly scipy, rebased and added testing. This is ready for a review.

dipy/tracking/utils.py
```diff
@@ -937,3 +939,50 @@ def reduce_rois(rois, include):
             exclude_roi |= rois[i]
 
     return include_roi, exclude_roi
+
+
+def vals_from_img(img, streamlines, affine=None):
+    """
```
Garyfallidis: no initial description

dipy/tracking/utils.py
```diff
@@ -937,3 +939,50 @@ def reduce_rois(rois, include):
...
+
+    affine : ndarray, shape (4, 4)
+        Affine transformation from voxels (image coordinates) to streamlines.
+        Default: identity.
+
+    Return
+    ------
+    array or list (depending on the input) : values interpolate to each
+        coordinate along the length of each streamline
+
```
Garyfallidis: reduce empty lines please

dipy/tracking/utils.py
Garyfallidis: I prefer ``values_`` than ``vals_``.

Garyfallidis: Also having input an img rather than data and affine is something rarely used in dipy. Do you really need img as input?

Garyfallidis: Hmm... after a bit more thinking  it seems the function needs to be renamed and moved to ``dipy.tracking.streamline``. The name of the function I propose is ``dipy.tracking.streamline.scalar_values``. This function will be useful also for @martcous who is currently preparing some PRs related to tractometry. @martcous give us a review here please and tell us if you have a different preference.

dipy/tracking/utils.py
```diff
@@ -937,3 +939,50 @@ def reduce_rois(rois, include):
...
+        for sl in streamlines:
+            vals.append(vfu.interpolate_scalar_3d(data, sl)[0])
+
+    elif isinstance(streamlines, np.ndarray):
+        sl_shape = streamlines.shape
+        sl_cat = streamlines.reshape(sl_shape[0] * sl_shape[1], 3)
+        if affine is not None:
+            sl_cat = np.dot(sl_cat, affine[:3, :3]) + affine[:3, 3]
+        # So that we can index in one operation:
+        vals = vfu.interpolate_scalar_3d(data, sl_cat)[0]
```
Garyfallidis: Does using this function means that we can get only scalar values? For example can we get rgb values from ``color_fa``?

dipy/tracking/utils.py
arokem: OK - for now, I am moving it into tracking.streamline. I have also addressed the other comments -- I think that you are right about using the data array and an affine, rather than the nibabel image. Have refactored accordingly.

dipy/tracking/utils.py
arokem: For now it does only interpolate scalar values. I guess if you want to interpolate a 4th dimension, you would have to interpolate each coordinate separately and combine? I guess we could implement this here as well (a loop over the 4th dimension, I am thinking) -- what do you think?

arokem:
Any more comments here? @martcous -- do you have any input on any of this?

martcous:
Sorry for not answering earlier. I think it makes sense for this to be in dipy.tracking.streamline. However, I am not sure about the name of the scalar_values() function. I think this operation would be useful to extract vectors from 4D arrays as well. I am not too familiar with Dipy's nomenclature, but I would name it something along the lines of extract_along_streamlines().

arokem:
Thanks for taking a look! That's a comment that @Garyfallidis made as well. Let me try to implement the n-dimensional version of this (or at least the 4D version), and I will also think about a name that makes sense.

arokem:
I've refactored this to work for 4D data as well. I've special-cased the 3D vector case, to use `interpolate_vector_3d`, but if the last dimension has a shape different from 3, it loops over that dimension and stacks the results in the end (using a concatenation for arrays, and another loop for lists). Let me know what you think. Does this work for your use-cases?

martcous:
Looks good to me, awesome work, thanks a lot!

dipy/tracking/streamline.py
```diff
@@ -323,3 +327,117 @@ def orient_by_rois(streamlines, roi1, roi2, affine=None, copy=True):
...
+    if (isinstance(streamlines, list) or
+            isinstance(streamlines, types.GeneratorType)):
+        vals = []
+        for sl in streamlines:
+            if threedvec:
+                vals.append(list(vfu.interpolate_vector_3d(data, sl)[0]))
+            else:
+                vals.append(list(vfu.interpolate_scalar_3d(data, sl)[0]))
+
+    elif isinstance(streamlines, np.ndarray):
```
samuelstjean: So if I don't give it an array, a list of a streamline generator, won't it crash (or do weird stuff) on trying to return vals?

arokem: That is correct. This would've raised a cryptic error ("val undefined" or somesuch), so I've added proper error handling in the `else` block and a test for that.

samuelstjean:
At the same time, this seems to only works on points and not segments, might as well add a scary warning in a note to not run this on compressed datasets or you will most probably get nonsense/very few values.

arokem:
OK - I have added that in the `Notes` section of the docstring

dipy/tracking/streamline.py
```diff
@@ -323,3 +327,130 @@ def orient_by_rois(streamlines, roi1, roi2, affine=None, copy=True):
...
+    if affine is not None:
+        streamlines = ut.move_streamlines(streamlines,
+                                          np.linalg.inv(affine))
+
+    if (isinstance(streamlines, list) or
+            isinstance(streamlines, types.GeneratorType)):
+        vals = []
+        for sl in streamlines:
+            if threedvec:
+                vals.append(list(vfu.interpolate_vector_3d(data, sl)[0]))
```
samuelstjean: The interpolate function only works on float64, so it gives errors like 

  File "/home/samuel/python/dipy/tracking/streamline.py", line 440, in values_from_volume
    affine=affine))
  File "/home/samuel/python/dipy/tracking/streamline.py", line 375, in _extract_vals
    vals.append(list(vfu.interpolate_scalar_3d(data, sl)[0]))
  File "dipy/align/vector_fields.pyx", line 434, in dipy.align.vector_fields.interpolate_scalar_3d (dipy/align/vector_fields.c:21841)
ValueError: Buffer dtype mismatch, expected 'double' but got 'float'

just adding a quick n dirty sl=sl.astype(np.float64) makes it work though, so at this point it might be worth to do the typecast in the cython files directly instead of each upper functions.

arokem: On Fri, Jan 15, 2016 at 7:15 AM, Samuel St-Jean <notifications@github.com>
wrote:

> In dipy/tracking/streamline.py
> <https://github.com/nipy/dipy/pull/811#discussion_r49863546>:
>
> > +    Return
> > +    ------
> > +    array or list (depending on the input) : values interpolate to each
> > +        coordinate along the length of each streamline
> > +    """
> > +    data = data.astype(np.float)
> > +    if affine is not None:
> > +        streamlines = ut.move_streamlines(streamlines,
> > +                                          np.linalg.inv(affine))
> > +
> > +    if (isinstance(streamlines, list) or
> > +            isinstance(streamlines, types.GeneratorType)):
> > +        vals = []
> > +        for sl in streamlines:
> > +            if threedvec:
> > +                vals.append(list(vfu.interpolate_vector_3d(data, sl)[0]))
>
> The interpolate function only works on float64, so it gives errors like
>
> File "/home/samuel/python/dipy/tracking/streamline.py", line 440, in
> values_from_volume
> affine=affine))
> File "/home/samuel/python/dipy/tracking/streamline.py", line 375, in
> _extract_vals
> vals.append(list(vfu.interpolate_scalar_3d(data, sl)[0]))
> File "dipy/align/vector_fields.pyx", line 434, in
> dipy.align.vector_fields.interpolate_scalar_3d
> (dipy/align/vector_fields.c:21841)
> ValueError: Buffer dtype mismatch, expected 'double' but got 'float'
>
> just adding a quick n dirty sl=sl.astype(np.float64) makes it work though,
> so at this point it might be worth to do the typecast in the cython files
> directly instead of each upper functions.
>
Sorry - is this happening somewhere in our CI? Or in some test that you've
written? I am not seeing this anywhere. Could you share some code that
triggers this error with this function?

samuelstjean:
I was testing it on a real use case, but here is one from the fornix dataset. Might be a good idea someday to force output of tracking as float32 since they are just coordinate, and you will save ram for free also.
~~~python 
import numpy as np
fname = get_data('fornix')
from nibabel import trackvis as tv
streams, hdr = tv.read(fname)
streamlines = [i[0].astype(np.float32) for i in streams]
from dipy.tracking.streamline import set_number_of_points, values_from_volume
extracted = values_from_volume(np.ones((10,10,10)), streamlines)
~~~

arokem:
Great. Thanks for the example. Will see what I can do to fix.

arokem:
For the time being, I've put in a test for this (that currently fails).

samuelstjean:
Since cython functions hardcode types, and for testing they have a simple python wrapper, easiest way would be to typecast there and use those. Kind of what I do with my stuff sometimes, then I return the original dtype if needed

arokem:
@omarocegueda : what do you think about adding type-casting to a wrapper
around `interpolate_scalar_3d`? IIUC, that's what @samuelstjean is
proposing in this thread.

omarocegueda:
Do you mean like:

https://github.com/omarocegueda/dipy/blob/cast_float64/dipy/align/vector_fields.pyx#L464 ?

That seems to be ok if we want to pass float32 buffers. 
Regarding forcing the output of tracking to be float32 mentioned above, I would be very cautious. That memory saving is not for free, at some point those tracks might be  filtered using anatomic ROIs and that loss of precision might be important. So, I would do that only if using float64 is definitely not feasible.

samuelstjean:
Arent streamlines only point coordinates? I would be very surprised to see a difference between 1e-8 and 1e-16 mm. Even rounding to 5 digits would probably be fine for roi extraction, with all the interpolation going around it can't have that much impact.

arokem:
Yep. I think that's the right thing for this. I actually just went ahead
and merged this branch into this current branch, so we can see how Travis
fares with that. I imagine we'd want some more testing and to also apply
this same logic in a few more places (vector interpolation and
interpolation in 2d as well), right? Let's move the discussion about
enforcing float32 to a dedicated issue:
https://github.com/nipy/dipy/issues/837

arokem:
Any thoughts about the loss of precision (see
https://travis-ci.org/nipy/dipy/jobs/102821379)? Is that inevitable?

arokem:
OK - I have armed this against different data-types of coordinates. This did result in some loss of precision in the answers up to approximately 4 decimals in the cases used for testing. Anyone has any more thoughts here?

dipy/tracking/streamline.py
```diff
@@ -323,3 +327,134 @@ def orient_by_rois(streamlines, roi1, roi2, affine=None, copy=True):
...
+        data, interpolation will be done on the 3 spatial dimensions in each
+        volume.
+
+    streamlines : ndarray or list
+        If array, of shape (n_streamlines, n_nodes, 3)
+        If list, len(n_streamlines) with (n_nodes, 3) array in
+        each element of the list.
+
+    affine : ndarray, shape (4, 4)
+        Affine transformation from voxels (image coordinates) to streamlines.
```
samuelstjean: streamlines space is world cooordinate or some other space?

arokem: It's whatever you did. If you represented your streamlines in "world
coordinates" (mm, relative to scanner iso-center), then you need an affine
to transform from that to the image.

samuelstjean: So its the streamlines which are moved (according to the function) to the
voxel space of the metrics. I was just wondering if the term used are
enough to guess what is happening, pretty new to all these confusing
conventions, but if people who know what they are doing understands what
you mean, then I'm good (and will read more stuff)

2016-01-25 17:35 GMT+01:00 Ariel Rokem <notifications@github.com>:

MarcCote: Is the coordinate (0,0,0) at the center of the voxel or at its corner? I think it refers to the center, so providing integer coordinates and affine=np.eye(4) to the function, it should output the exact value of `data` indexed at those coordinates, right?

arokem: Yes. That's correct. For example in the test here: https://github.com/arokem/dipy/blob/vals_from_img/dipy/tracking/tests/test_streamline.py#L834, the first value is exactly the value at `data[1, 0, 0]`, because the first coordinate of the first streamline is `[1, 0, 0]`. This uses the default  affine, which is `eye(4)`. Does that answer your question?

arokem: Any suggestions for ways to make the documentation clearer on that? I'll add a couple of words.

arokem: How's that?

dipy/tracking/streamline.py
```diff
@@ -323,3 +327,136 @@ def orient_by_rois(streamlines, roi1, roi2, affine=None, copy=True):
...
+            else:
+                vals.append(list(vfu.interpolate_scalar_3d(data,
+                                 sl.astype(np.float))[0]))
+
+    elif isinstance(streamlines, np.ndarray):
+        sl_shape = streamlines.shape
+        sl_cat = streamlines.reshape(sl_shape[0] *
+                                     sl_shape[1], 3).astype(np.float)
+        if affine is not None:
+            sl_cat = (np.dot(sl_cat, affine[:3, :3]) + affine[:3, 3])
```
samuelstjean: Why is the affine only applied if the input is an array? Besides, the streamlines are moved at the top already, so what does this do?

arokem: Good catch. This line never got run in the tests, because when an affine was provided, the array would be passed through move_streamlines, which always returns a generator. Recent commit fixes that.

samuelstjean: Oh good, but what does it actually do? It was a real question and not a bug report :P I thought if you moved everything to image space with the move streamline function, why do you need to reapply the affine? 

Might explain why my extracted values seems decent, but whenever I try to save back the streamlines they won't fit on anatomy.

arokem: See the recent commit. Now, the affine is not applied upfront. Instead,
each case is handled separately: if you have a list or generator, it gets
moved by move_streamline. If it's an array, we apply the affine to it here.
To answer your question, this is just a simplified way of applying the
affine. Dot product your coordinates with the top left 3x3 part of the
affine to apply the rotation and then add the last column to apply the
translation (see here for details:
http://nipy.org/nibabel/coordinate_systems.html)

I've also added a test to check that I am not changing the streamlines on
you, but I'd be happy if you could take a look at the data you are putting
through this and see that indeed the coordinates in the streamlines you
pass through remain the same, after passing through this function, in case
I missed some cases.

samuelstjean: Seems good, I must do whatever weird when converting from voxel to world since I pass between different voxel size, or something else I'm using in-between does not like that.

dipy/tracking/streamline.py
```diff
@@ -323,3 +327,138 @@ def orient_by_rois(streamlines, roi1, roi2, affine=None, copy=True):
...
+                                 sl.astype(np.float))[0]))
+
+    elif isinstance(streamlines, np.ndarray):
+        sl_shape = streamlines.shape
+        sl_cat = streamlines.reshape(sl_shape[0] *
+                                     sl_shape[1], 3).astype(np.float)
+
+        if affine is not None:
+            inv_affine = np.linalg.inv(affine)
+            sl_cat = (np.dot(sl_cat, inv_affine[:3, :3]) +
```
samuelstjean: You can also do that in only one matrix multiplication, that's why homogeneous coordinates are used.

https://en.wikipedia.org/wiki/Transformation_matrix#Affine_transformations

arokem: Yeah - but then you have to augment the coordinates, adding a '1' to every
coordinate. This is a way to get around that.

samuelstjean: Wouldn't that be faster if you have to move a whole brain tractogram? Computer vision/imaging people are using these since you can express all transformations in a single operation instead, maybe they want to pitch in about if its faster to add another dimension or go this way.

dipy/tracking/streamline.py
MarcCote: @arokem, yes that answered my question.

arokem:
I don't think it would be faster, and it's definitely less memory-demanding. FWIW, this is the same computation that is used by move_streamlines: https://github.com/nipy/dipy/blob/master/dipy/tracking/utils.py#L917

MarcCote:
I tested it a bit and seems to be all good. Is there anything left to check? If not, I can go ahead with the merging it you want.

samuelstjean:
Can't see my message from yesterday, it must have gone under a hidden diff comment. But anyway, its a few seconds faster (2.5 vs 3) on a small set of 200 000 streamlines, if you deal with a few millions it might be useful in the long run, just saying.

arokem:
Must have been a fluke: https://gist.github.com/arokem/db8396eccc9d138175f3

arokem:
To be clear: I don't see any evidence that transforming streamlines by multiplying an augmented streamline (4D) with the full 4 by 4 matrix is faster than multiplying the 3D streamline with the top-left rotation matrix and adding the 3 coordinates of the translation transform. Without evidence, I don't think we should change anything about how we transform streamlines. Either here, or in `move_streamlines`. But I would love to see evidence that there is a potential almost 20% speedup we could get here.

samuelstjean:
I meant on a real example, like shifting a whole tractogram, linear algebra would have to be bad to do something on such a small example.

~~~python
aff=np.eye(4)
aff[3,:3] = 10

import nibabel_streamlines as nib
bund = nib.streamlines.load('tracking.trk')
slines = bund.streamlines

In [86]: %timeit for i in slines: np.dot(i,aff[:3,:3]) + aff[:3,3];

1 loops, best of 3: 6.91 s per loop

In [87]: %timeit for i in slines: a = np.ones((i.shape[0],4));a[:,:3]=i;np.dot(a,aff);
1 loops, best of 3: 6.38 s per loop
~~~

Anyway, it just seems weird to not use something the way it's intended for, that's all.

arokem:
Sorry. Still not seeing any substantial difference, even with 200k
streamlines: https://gist.github.com/arokem/db8396eccc9d138175f3

samuelstjean:
Wow, my computer might be bad then or something compared to yours. Or we are not using the same versions/stuff/lib/whatever, not really important at this point. Still, I guess it's more important for realtime stuff than what we are doing.

matthew-brett:
You'll be doing more operations with the 4x4 multiplication than by doing the 3x3 multiplication and adding the 3, vector, as long as you can assume that the bottom row of the affine is [0, 0, 0, 1].  First case is 4 * (4 multiplications, 3 additions) per coordinate, second is 3 * (3 multiplications, 3 additions) per coordinate.  I think.

So, I'd expect second approach to be faster in general, but I guess will depend on your BLAS - it might be that intense cache optimization could overcome the operation count.

samuelstjean:
So while we are still discussing this, would you mind adding a nearest neighbour interpolation option? The idea is to extract indices of crossed voxels instead of values (whichever might be easier), this way you can get back the original pathway of peaks followed by the tracking easily.

arokem:
Yeah. That wouldn't be too hard. Let me do that here, before this one gets
merged. It will require another round of review here, though...

MarcCote:
Good idea @samuelstjean. @arokem no problem.

samuelstjean:
That's fine, I actually wanted to do something similar, and you actually
made everything for me. So I'm kind of beta testing it instead.

arokem:
Great! Feel free to make a PR against this branch, if you have something
that's reasonably ready to go. We can polish this up together on this
branch.

samuelstjean:
No actually I was about to do it and you had a pr already, so I am just
leeching your stuff as you make it in all honesty. But I was looking this
afternoon at redoing the intersection stuff JC used for ismrm to do it,
might go at it monday as it looked relatively simple.

jchoude:
@samuelstjean You don't need to do it, I'll try to put it in in the coming week.

arokem:
Hey @jchoude -- thanks for pitching in on this. Would it be easier for you to do this as a follow-up PR, or as a PR against this one?

jchoude:
@arokem Easier in a separate PR

arokem:
No problem. Just to be sure -- the idea would be to introduce a key-word argument in this function (something like `interpolate="nearest"` or `interpolation="linear"`, correct?

samuelstjean:
I'd rather not for this case, his version is segment based instead of point
based, which is independent  of the internal interpolation.

arokem:
I meant as key-word arguments to the `values_from_volume` function.

We can do the segment-based thing under the hood.

jchoude:
Well, if I don't see any performance degradation during the integration to Dipy, it could become the default mode. From my tests in standalone code, I was able to reach about the same speed as the density_map (and / or track_counts) functions, which are points-based.

arokem:
Any more comments here? It seems that @jchoude will make a separate PR for nearest-neighbors behavior.

jchoude:
In fact, I'll make a separate PR for the code that works with variable step sizes.

arokem:
Great! Anything more to do here?

MarcCote:
Seems all good. I'll merge it then.

samuelstjean:
I saw some typo still though

PR #817, opened 2015-12-21
RF: Convert nan values in bvectors to 0's

arokem:
This get's rid of warnings of the following type that currently show up
in our tests:

    ./Users/arokem/source/dipy/dipy/core/gradients.py:132: RuntimeWarning: invalid value encountered in less_equal
  bvecs_close_to_1 = abs(vector_norm(bvecs) - 1) <= atol

### DIFF
--- a/dipy/core/gradients.py
+++ b/dipy/core/gradients.py
@@ -134,11 +134,11 @@ def gradient_table_from_bvals_bvecs(bvals, bvecs, b0_threshold=0, atol=1e-2,
                          "respectively, where N is the number of diffusion "
                          "gradients")
 
+    bvecs[np.isnan(bvecs)] = 0
     bvecs_close_to_1 = abs(vector_norm(bvecs) - 1) <= atol
     if bvecs.shape[1] != 3 or not np.all(bvecs_close_to_1[dwi_mask]):
         raise ValueError("bvecs should be (N, 3), a set of N unit vectors")
 
-    bvecs = np.where(bvecs_close_to_1[:, None], bvecs, 0)
     bvals = bvals * bvecs_close_to_1
     gradients = bvals[:, None] * bvecs
 
--- a/dipy/core/gradients.py
+++ b/dipy/core/gradients.py
@@ -243,7 +243,7 @@ def gradient_table(bvals, bvecs=None, big_delta=None, small_delta=None,
                              " array containing both bvals and bvecs")
     else:
         bvecs = np.asarray(bvecs)
-        if (bvecs.shape[1] > bvecs.shape[0]) and bvecs.shape[0] > 1:
+        if (bvecs.shape[1] > bvecs.shape[0])  and bvecs.shape[0] > 1:
             bvecs = bvecs.T
     return gradient_table_from_bvals_bvecs(bvals, bvecs, big_delta=big_delta,
                                            small_delta=small_delta,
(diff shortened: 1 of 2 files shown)


matthew-brett:
This changes the API, no?    What do people use NaN bval values for?

Needs a test.

arokem:
Added a test to clarify the case in which this is happening. In the data set that we commonly use for our own tests, a nan bvector is used to mark a 0 b-value.

Either way, I don't think that this is a change in API, because I think that it's exactly equivalent to what was previously happening on line 137 

I have now also removed that line, and rebased to change the order of things: first the test and then the fix.

arokem:
Just rebased this one on master. Anyone care to review/merge?

dipy/core/gradients.py
```diff
@@ -134,11 +134,11 @@ def gradient_table_from_bvals_bvecs(bvals, bvecs, b0_threshold=0, atol=1e-2,
                          "respectively, where N is the number of diffusion "
                          "gradients")
 
+    bvecs[np.isnan(bvecs)] = 0
     bvecs_close_to_1 = abs(vector_norm(bvecs) - 1) <= atol
     if bvecs.shape[1] != 3 or not np.all(bvecs_close_to_1[dwi_mask]):
         raise ValueError("bvecs should be (N, 3), a set of N unit vectors")
 
-    bvecs = np.where(bvecs_close_to_1[:, None], bvecs, 0)
```
MrBago: @arokem this line is here because in some data sets a low b-value (~5-30) is used for the b-0 directions. There was a little back and forth about this, but I decided to squash these values because they don't come with a direction vector and a gradient with no direction vector doesn't really make sense. I think we should leave this line in, I can make a test for this kind of data.

dipy/core/gradients.py
```diff
@@ -134,11 +134,11 @@ def gradient_table_from_bvals_bvecs(bvals, bvecs, b0_threshold=0, atol=1e-2,
                          "respectively, where N is the number of diffusion "
                          "gradients")
 
+    bvecs[np.isnan(bvecs)] = 0
```
MrBago: Please use something like `bvecs = np.where(np.isnan(bvecs), 0, bvecs)`, `bvecs` is an argument to the function, we don't want to silently modify the function arguments.

dipy/core/gradients.py
```diff
@@ -243,7 +243,7 @@ def gradient_table(bvals, bvecs=None, big_delta=None, small_delta=None,
                              " array containing both bvals and bvecs")
     else:
         bvecs = np.asarray(bvecs)
-        if (bvecs.shape[1] > bvecs.shape[0]) and bvecs.shape[0] > 1:
+        if (bvecs.shape[1] > bvecs.shape[0])  and bvecs.shape[0] > 1:
```
MrBago: what did you change here?

arokem: whitespace

arokem:
Thanks for taking a look @MrBago - fixed according to your suggestions. Would be good to have test-cases for those bvecs_close_to_1 cases you mentioned.

Garyfallidis:
Is this ready for a merge or we are waiting for the close to 1 case?

arokem:
Ready from my point of view, unless @MrBago wants to add the test-cases discussed. He can also do this on a separate PR. At the very least, to my understanding, this now gets rid of those unnecessary warnings, and doesn't change any other behavior.

PR #818, opened 2015-12-24
Round coords life

### LINKED ISSUE #810: Wrong input type for `_voxel2stream` on 64-bit Windows
We're getting several errors of form:

```
ERROR: dipy.tracking.tests.test_life.test_voxel2streamline
  File "C:\buildslaves\mike-win7-64\dipy-bdist64-35\build\venv\lib\site-packages\nose\case.py", line 198, in runTest
    self.test(*self.arg)
  File "C:\buildslaves\mike-win7-64\dipy-bdist64-35\build\venv\lib\site-packages\dipy\tracking\tests\test_life.py", line 89, in test_voxel2streamline
    v2f, v2fn = life.voxel2streamline(streamline, False, affine)
  File "C:\buildslaves\mike-win7-64\dipy-bdist64-35\build\venv\lib\site-packages\dipy\tracking\life.py", line 313, in voxel2streamline
    return _voxel2streamline(transformed_streamline, unique_idx)
  File "dipy\tracking\vox2track.pyx", line 20, in dipy.tracking.vox2track._voxel2streamline (dipy\tracking\vox2track.c:1562)
ValueError: Buffer dtype mismatch, expected 'npy_intp' but got 'long'
```

In fact, `long` and `int` on Windows 64 are length 4 integers.

Here's the relevant part of `voxel2streamline`:

```
    if unique_idx is None:
        all_coords = np.concatenate(transformed_streamline)
        unique_idx = unique_rows(all_coords.astype(int))
    else:
        unique_idx = unique_idx

    return _voxel2streamline(transformed_streamline, unique_idx)
```

So, an obvious fix would be to cast `unique_idx` to `np.intp`, but - is the previous cast to `int` a good idea?  Shouldn't it be `round` instead?

arokem:
Should fix #810

### DIFF
--- a/dipy/tracking/life.py
+++ b/dipy/tracking/life.py
@@ -53,7 +53,8 @@ def gradient(f):
     ----
     This is a simplified implementation of gradient that is part of numpy
     1.8. In order to mitigate the effects of changes added to this
-    implementation in version 1.9 of numpy, we include this implementation here.
+    implementation in version 1.9 of numpy, we include this implementation
+    here.
     """
     f = np.asanyarray(f)
     N = len(f.shape)  # number of dimensions
@@ -306,9 +307,7 @@ def voxel2streamline(streamline, transformed=False, affine=None,
 
     if unique_idx is None:
         all_coords = np.concatenate(transformed_streamline)
-        unique_idx = unique_rows(all_coords.astype(int))
-    else:
-        unique_idx = unique_idx
+        unique_idx = unique_rows(np.round(all_coords).astype(np.intp))
 
     return _voxel2streamline(transformed_streamline, unique_idx)
 
@@ -399,11 +398,8 @@ class FiberModel(ReconstModel):
         range_bvecs = np.arange(n_bvecs).astype(int)
         # In each voxel:
         for v_idx in range(vox_coords.shape[0]):
-            # dbg:
-            #if not np.mod(v_idx, 1000):
-            #    print("voxel %s"%(100*float(v_idx)/n_vox))
             mat_row_idx = (range_bvecs + v_idx * n_bvecs).astype(np.int32)
-            #For each fiber in that voxel:
+            # For each fiber in that voxel:
             for f_idx in v2f[v_idx]:
                 # For each fiber-voxel combination, store the row/column
                 # indices in the pre-allocated linear arrays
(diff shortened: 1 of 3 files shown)


matthew-brett:
Combination of PEP8 and fix is hard to follow.

dipy/tracking/life.py
```diff
@@ -306,9 +307,7 @@ def voxel2streamline(streamline, transformed=False, affine=None,
 
     if unique_idx is None:
         all_coords = np.concatenate(transformed_streamline)
-        unique_idx = unique_rows(all_coords.astype(int))
-    else:
-        unique_idx = unique_idx
+        unique_idx = unique_rows(np.round(all_coords).astype(np.intp))
```
matthew-brett: How about leaving the cast to the later line, to avoid users getting errors from passed `unique_idx` of wrong data type.  As in:

```
        unique_idx = unique_rows(np.round(all_coords))

return _voxel2streamline(transformed_streamline, unique_idx.astype(np.intp))
```

dipy/tracking/life.py
```diff
@@ -399,11 +398,8 @@ def setup(self, streamline, affine, evals=[0.001, 0, 0], sphere=None):
         range_bvecs = np.arange(n_bvecs).astype(int)
         # In each voxel:
         for v_idx in range(vox_coords.shape[0]):
-            # dbg:
-            #if not np.mod(v_idx, 1000):
-            #    print("voxel %s"%(100*float(v_idx)/n_vox))
             mat_row_idx = (range_bvecs + v_idx * n_bvecs).astype(np.int32)
```
matthew-brett: Why `int32` here?

dipy/tracking/life.py
matthew-brett: Also - need test for modified behavior - such as `transformed_streamline` with non-integer indices.

dipy/tracking/life.py
matthew-brett: Also - should `f_matrix_row`, `f_matrix_col` etc be `np.intp`?

arokem:
Sorry about interleaving functional changes together with PEP8 -- the only change (so far) that isn't PEP8 (i.e. white-space) is the rounding and casting to intp. We just otherwise never get to these things....

arokem:
Added testing, and more calls to `round` to accommodate the change more widely. 

I had to also reduce the match to the matlab version, but I am not sure how close we want to stay to that, as we fix up things (The matlab version uses `ceil` on the streamline coordinates in these cases: https://github.com/francopestilli/life/blob/master/utility/fefgGet.m#L201, but I am not convinced this is correct). 

At the very least, the model RMSE is lower here than in the Matlab version, so the model is a more accurate fit to the signal, for the same number of parameters.

I want to make more changes to LiFE to deal with the memory requirement. I will make a branch off of this work and a separate PR, once we resolve things here.

matthew-brett:
Sorry to be dumb - but are the tests such that, if you change the `round` back to casting to `int`, they will fail?

arokem:
Yes - I made sure of that on my laptop, but there's also an explanation:
the float coordinates [1.1, 2.4, 2.9] would be [1, 2, 2] in the
`astype(int)` case, creating a coordinate that is different and unique from
the first coordinate of the next streamline: [1, 2, 3]. In the `round`
case, this goes to the same coordinate => [1, 2, 3].

matthew-brett:
OK - thanks for the explanation - good to go from my PoV.

Garyfallidis:
:+1:

PR #819, opened 2015-12-24
DOC: Add the DKI reconstruction example to the list of valid examples.

arokem:
This is a (very) small change in the documentation, but also a more general question about how we should go about updating the documentation, while keeping it compatible with the 0.10.1 released version. Is the right thing to make the PRs on the documentation against the maint/0.10.x branch and then merge (cherry pick?) from there into master?

### DIFF
--- a/doc/examples/valid_examples.txt
+++ b/doc/examples/valid_examples.txt
@@ -10,6 +10,7 @@
  reconst_dti.py
  reconst_gqi.py
  reconst_dsid.py
+ reconst_dki.py
  kfold_xval.py
  reslice_datasets.py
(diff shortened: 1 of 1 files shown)


Garyfallidis:
I don't know what is the best way to do this. I think we need to experiment a bit.

Garyfallidis:
In the meantime I think because already other important PRs for 0.10.x have been merged in the master we should add that one too. Agreed?

arokem:
Sure. We can also have the discussion on some other channel. Any reason not to merge this?

arokem:
ping!

PR #823, opened 2015-12-31
WIP: "Out-of-core" LiFE

arokem:
Introduces an additional fitting method to the LiFE algorithm. This fitting method performs the computation "out of core", in the sense that the large model matrix is never entirely held in memory at any given time. Instead, the part of the matrix pertaining to each voxel is composed in each iteration through and the contribution of that voxel to the gradient on the parameters beta is calculated. The combined voxel contributions are summed iterating through all the voxels and applied together to the parameters in each iteration through the optimization. 

On my laptop, this uses no more than 6 GB of memory running on 250k streamlines (compare to http://nbviewer.ipython.org/gist/arokem/bc29f34ebc97510d9def, using about 60 GB...). But it also takes slightly longer (about 2.5 hours on my laptop). 

I am still testing this out and will also add a few words about this to the gallery example. Comments and suggestions about the general approach are very welcome at this point. If you dig into the commit history of this branch, you will see that I tried a few other things as well: including both PyTables and memmap to map the arrays to memory. But these approaches were not successful in reducing the memory demand, because access to the arrays (for things like np.where) still required loading the entire thing into memory. Happy to hear suggestions in that direction as well -- we might still be able to make the "standard" approach more memory-efficient as well.


samuelstjean:
What is the runtime for the old version? You seem to use a fixed step gradient descent for the optimization, that could explain why it takes hours.

arokem:
Thanks for the comments. Hard to compare run-time, because I previously ran it on a different machine (much faster, in addition to much more memory), but it was on the order of 15 minutes for that same task (so x6 or so faster). Do you have a good reference or pointer for how to set and change step size in gradient descent?

samuelstjean:
Well, that's where it hurts, its application dependent. There are criterion
for finding the 'best' stepsize for each iteration, but sometimes you end
up using more computing time to find this than do fixed step size.
Although, you can have a quick look at
http://docs.scipy.org/doc/scipy-0.16.0/reference/generated/scipy.optimize.line_search.html
for a basic one and see how it goes. There are a slew of methods,
heuristic,s algorithms and all, as optimization is still a flourishing
branch of applied math.

Anyway, I guess I would need to look at the math in the paper, but isn't
there a closed form solution instead of using an iterative solution for
these problems? You might be able to get a huge speedup (unless you invert
matrices which are ridiculously big, which was happening before if I recall
correctly) in that way instead of iterative optimization.

2016-01-14 16:33 GMT+01:00 Ariel Rokem <notifications@github.com>:

dipy/tracking/life.py
```diff
@@ -488,14 +522,191 @@ def fit(self, data, streamline, affine=None, evals=[0.001, 0, 0],
...
+                    sl_as_coords[sl_idx][:, 1] == vox_coords[v_idx][1]),
+                  sl_as_coords[sl_idx][:, 2] == vox_coords[v_idx][2])
+                nodes_in_vox = np.where(find_vox)[0]
+                s_in_vox[v_idx].append((sl_idx, nodes_in_vox))
+
+        # We no longer need these variables:
+        del v2f, streamline, sl_as_coords
+
+        delta = np.zeros(beta.shape)
+        while 1:
```
samuelstjean: As this part here seem to do fixed stepsize gradient descent, you might want to try http://docs.scipy.org/doc/scipy-0.16.0/reference/generated/scipy.optimize.nnls.html to see if its faster in case. If not, the problem is convex, so you might wan tot try an already made solver also for ease of maintenance.

arokem: None of these will work here for two reasons: the first is that even in the best case, this is a sparse matrix, and these typically do not support sparse matrices. The second is that in this case, not only is this a sparse matrix, but even the matrix definition (the row-column addresses and data that composes the non-zero parts of the matrix) will not comfortably fit into memory on a standard machine. For that reason, the algorithm that I have implemented goes about constructing the gradient in each step *voxel by voxel*. But the idea of changing the step size as we go along is a good one, and might help. I am still working on this, but I need to sketch out something simpler, to see where memory leaks from here, and how we can most effectively incorporate changes in step size.

samuelstjean: I looked at that ref 15 mentioned in the appendix, and they do just that, using a custom nnls with a nesterov step, that should give you your speed boost. Could also use it for shore, might be what cvxopt does under the hood, and actually could be useful for scipy as their version is horribly slow (hence why the guys use cvxopt in the first place).

samuelstjean: Ah and as for the sparse matrix thing, where is it used? I see you call it, but wasn't the first version doing that? It might be reverting back to regular matrix without telling you (yes., it does that through weird broadcasting, and it's nasty), which could explain the original 60+ go of ram required.

arokem: Yes - it would be good to implement the Kim/Sra/Dhillon algorithm, and might be useful for many things. But I am not sure it would help us here.

In the previous implementation of LiFE (still maintained here as as the "fast, but memory hungry" implementation), I implemented a similar gradient descent algorithm, but used the sparse dot product. So, I am pretty sure that was not the point at which memory went up -- it actually went up earlier, when allocating the rows columns and data of the matrix (maybe we should switch this up to a float32 representation too?). 

In this implementation instead the sparse matrix doesn't even ever get defined in full in this implementation -- that's why it doesn't take up as much memory: instead it's defined voxel by voxel, and the gradient contribution from each voxel gets added up (L663), finally to be applied at once at the end of every loop through voxels (L675). I will see if we can do the Nesterov step here as well - that's a good idea.

martcous:
Hi, I will be using LiFE in Dipy and I've been trying out this PR in order to avoid memory problems. When I run the example `doc/examples/linear_fascicle_evaluation.py` by simply changing the line `fiber_model = life.FiberModel(gtab, conserve_memory=True)`, I get an all-zero `fiber_fit.beta` after `fiber_model.fit()`. It works fine if `conserve_memory=False`. Is this normal?

arokem:
No - that seems wrong. What is the code you are running (beyond the call to
the FiberModel)?

martcous:
`doc/examples/linear_fascicle_evaluation.py` with the only modification of the line `fiber_model = life.FiberModel(gtab, conserve_memory=True)` (Setting conserve_memory to True).

arokem:
Oh yeah, that's not good. Let me take a look.

dipy/tracking/life.py
samuelstjean: While Martin reminded me this existed, I think I found out where memory blows up (had a similar thing once), just forgot about it, but you also seem to have fixes it in this new one. Anyway, seems like the sparse dot function (I guess coming coming from sklearn) does not really do what we thought : If one of the matrices is not sparse, the whole thing seems to multiply in memory, thus blowing it up.

~~~python
from dipy.core.optimize import spdot
import scipy.sparse as sps
a=sps.random(100000,1000, 0.001)
b=sps.random(1000,1000, 0.01)
# this works spdot(a,b)
c=b.todense()
# this start to kill your computer
d=spdot(a,c)
e=spdot(a,c)
~~~

Issue #802, opened 2015-12-04
toollib and doc tools need update to 3.5


arokem:
For example: 
https://github.com/nipy/dipy/blob/master/tools/toollib.py#L16
https://github.com/nipy/dipy/blob/master/doc/tools/apigen.py#L428

arokem:
Closed through #825

PR #825, opened 2016-01-03
Updated toollib and doc tools (#802)

smarshy:


matthew-brett:
Great, thanks for doing this.

PR #826, opened 2016-01-04
Return numpy arrays instead of memory views from cython functions

omarocegueda:
Workaround to prevent this kind of memory leaks:
https://github.com/omarocegueda/cython_leak_repro/blob/master/memory_leak.ipynb
which occur when using memory views allocated and returned from cython functions

### DIFF
--- a/dipy/align/crosscorr.pyx
+++ b/dipy/align/crosscorr.pyx
@@ -145,7 +145,7 @@ def precompute_cc_factors_3d(floating[:, :, :] static, floating[:, :, :] moving,
                         Imean * sums[SI] + sums[CNT] * Imean * Imean)
                     factors[s, r, c, 4] = (sums[SJ2] - Jmean * sums[SJ] -
                         Jmean * sums[SJ] + sums[CNT] * Jmean * Jmean)
-    return factors
+    return np.array(factors)
 
 
 @cython.boundscheck(False)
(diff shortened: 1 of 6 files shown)


arokem:
I've pushed a branch with this change and the appveyor config here: 

https://github.com/arokem/dipy/tree/omarocegueda-wrap_memviews-appveyor

It's running an appveyor build over here: 
https://ci.appveyor.com/project/arokem/dipy/build/1.0.576

omarocegueda:
Awesome! let's see what it says =)

arokem:
Still no joy:
https://ci.appveyor.com/project/arokem/dipy/build/1.0.576/job/4j8cy33yvdesmd66

Garyfallidis:
Rather than using np.array(memview) it would be interesting to check if it is more advantageous to use np.asarray see here
http://stackoverflow.com/questions/20978377/cython-convert-memory-view-to-numpy-array

Garyfallidis:
@omarocegueda can you update this PR? It will be nice to get rid of the memory leaks (or any copies) in this release.

omarocegueda:
No problem!, now using np.asarray(...). This will eliminate the memory leaks, but not yet the issue on appveyor.

arokem:
That's OK. We'll figure out the appveyor thing on another branch. For the time being, @Garyfallidis - is this ready to be merged?

Garyfallidis:
Yes!

omarocegueda:
Thank you guys! =)

PR #762, opened 2015-11-05
Contextual enhancements of ODF/FOD fields

stephanmeesters:
Included is the code for contextual enhancement of ODF/FOD fields based on the work of Remco Duits and Jorg Portegies.

Eleftherios has assisted a lot with the integration into DiPy. The enhancements are written in Cython and support OpenMP.

Files:
doc/examples/contextual_enhancement.py *-- describes the method and showcases the enhancement*
dipy/denoise/enhancement_kernel.py *-- computes a lookup-table*
dipy/denoise/shift_twist_convolution.py *-- performes the convolution over the space of positions and orientations*
dipy/denoise/tests/test_kernel.py *-- tests if the kernel values are computed correctly, if the symmetric properties of the kernel are there, and if the convolution does its job.*

If needed, pictures for the demo are here (2 images generated by the example code, cropped; one new image).
https://drive.google.com/folderview?id=0B-yaAUqTzxPBOC1FLWxtekhfSWM&usp=sharing

PR #828, opened 2016-01-07
Fiber to bundle coherence measures

stephanmeesters:
Extends upon the previous pull request #762 with fiber to bundle coherence (FBC) measures used to clean up tractography results. Uses the lookup-table computations of pull request #762 .

Files added:
-- dipy/tracking/fbcmeasures.pyx
        Computes the FBC measures.
-- doc/examples/fiber_to_bundle_coherence.py
        Demonstration and explanation of FBC measures applied to tractography of the optic radiation.


dipy/tracking/fbcmeasures.pyx
```diff
@@ -0,0 +1,334 @@
...
+                           J. Portegies, P. Ossenblok, R. Duits. (2016) Cleaning 
+                           output of tractography via fiber to bundle coherence, 
+                           a new open source implementation. Human Brain Mapping 
+                           conference 2015.
+        [Portegies2015b] J. Portegies, R. Fick, G. Sanguinetti, S. Meesters, 
+                         G.Girard, and R. Duits. (2015) Improving Fiber Alignment 
+                         in HARDI by Combining Contextual PDE flow with 
+                         Constrained Spherical Deconvolution. PLoS One.
+        """
+        self.compute(streamlines, kernel)
```
arokem: I believe you need to pass `num_threads` here as well?

dipy/tracking/fbcmeasures.pyx
```diff
@@ -0,0 +1,334 @@
...
+from dipy.data import get_sphere
+from dipy.denoise.enhancement_kernel import EnhancementKernel
+from dipy.core.ndindex import ndindex
+
+cdef class FBCMeasures:
+
+    cdef int [:] streamline_length
+    cdef double [:, :, :] streamline_points
+    cdef double [:, :] streamlines_lfbc
+    cdef double [:] streamlines_rfbc
```
arokem: Why are these variables defined in this scope? Would it make sense to do these initializations inside the `__init__` instead?

dipy/tracking/fbcmeasures.pyx
```diff
@@ -0,0 +1,334 @@
...
+                           a new open source implementation. Human Brain Mapping 
+                           conference 2015.
+        [Portegies2015b] J. Portegies, R. Fick, G. Sanguinetti, S. Meesters, 
+                         G.Girard, and R. Duits. (2015) Improving Fiber Alignment 
+                         in HARDI by Combining Contextual PDE flow with 
+                         Constrained Spherical Deconvolution. PLoS One.
+        """
+        self.compute(streamlines, kernel)
+        
+    def get_points_rfbc_thresholded(self, threshold, showInfo=False, emphasis=.5):
```
arokem: I prefer `verbose` to `showInfo`. 

Either way, please don't use camelCase.

dipy/tracking/fbcmeasures.pyx
```diff
@@ -0,0 +1,334 @@
...
+
+        if have_openmp:
+            openmp.omp_set_dynamic(0)
+            openmp.omp_set_num_threads(threads_to_use)
+        
+        # if the fibers are too short FBC measures cannot be applied, 
+        # remove these.
+        streamlines_length = np.array([len(x) for x in py_streamlines], 
+                                dtype=np.int32)
+        minLength = min(streamlines_length)
```
arokem: `minLength` => `min_length`

dipy/tracking/fbcmeasures.pyx
```diff
@@ -0,0 +1,334 @@
...
+        if minLength < 10:
```
arokem: Maybe make this threshold a key-word argument (in the `__init__` and in the input to `compute`) with a reasonable default (e.g., 10)

dipy/tracking/fbcmeasures.pyx
```diff
@@ -0,0 +1,334 @@
...
+        else:
+            threads_to_use = all_cores
+
+        if have_openmp:
+            openmp.omp_set_dynamic(0)
+            openmp.omp_set_num_threads(threads_to_use)
+        
+        # if the fibers are too short FBC measures cannot be applied, 
+        # remove these.
+        streamlines_length = np.array([len(x) for x in py_streamlines],
```
arokem: I believe using the `shape` attribute of the streamlines might be (marginally) faster than calling `len` on each one.

dipy/tracking/fbcmeasures.pyx
arokem: If you are using an array, maybe call `np.min` here?

dipy/tracking/fbcmeasures.pyx
```diff
@@ -0,0 +1,334 @@
...
+                                dtype=np.int32)
+        minLength = min(streamlines_length)
+        if minLength < 10:
+            print("The minimum fiber length is 10 points. \
+                    Shorter fibers were found and removed.")
+            py_streamlines = [x for x in py_streamlines if len(x) >= 10]
+            streamlines_length = np.array([len(x) for x in py_streamlines], 
+                                    dtype=np.int32)
+            minLength = min(streamlines_length)
+        numberOfFibers = len(py_streamlines)
```
arokem: `numberOfFibers` => `num_fibers` (or some-such, but no camelCase, please).

dipy/tracking/fbcmeasures.pyx
```diff
@@ -0,0 +1,334 @@
...
+        self.streamline_length = streamlines_length
+        maxLength = max(streamlines_length)
```
arokem: `max_length`?

arokem: Also, might want to use `np.max` instead?

dipy/tracking/fbcmeasures.pyx
```diff
@@ -0,0 +1,334 @@
...
+        
+        # prepare numpy arrays for speed
+        streamlines = np.zeros((numberOfFibers, maxLength, dim), 
+                                dtype=np.float64) * np.nan
+        streamlines_tangents = np.zeros((numberOfFibers, maxLength, dim), 
+                                dtype=np.float64)
+        streamlines_nearestp = np.zeros((numberOfFibers, maxLength), 
+                                dtype=np.int32)
+        streamline_scores = np.zeros((numberOfFibers, maxLength), 
+                                dtype=np.float64) * np.nan
```
arokem: Can any of these arrays be preallocated using `cdef`? Might make this even faster?

dipy/tracking/fbcmeasures.pyx
```diff
@@ -0,0 +1,334 @@
...
+        for lineId in range(numberOfFibers):
+            for pointId in range(streamlines_length[lineId] - 1):
+                streamlines_nearestp[lineId, pointId] = \
+                    tree.query(streamlines[lineId, pointId])[1]
+        
+        # arrays for parallel computing
+        score_mp = np.zeros(numberOfFibers)
+        xd_mp = np.zeros(numberOfFibers, dtype=np.int32)
+        yd_mp = np.zeros(numberOfFibers, dtype=np.int32)
+        zd_mp = np.zeros(numberOfFibers, dtype=np.int32)
```
arokem: Again, I wonder whether these arrays need to be numpy, or can be allocated using cdef instead? I am really not sure, but thought I would ask.

dipy/tracking/fbcmeasures.pyx
```diff
@@ -0,0 +1,334 @@
...
+
+        if have_openmp:
+            print("Running in parallel!")
```
arokem: Print messages only if `verbose` (or whatever it's called) is true.

dipy/tracking/fbcmeasures.pyx
```diff
@@ -0,0 +1,334 @@
...
+                                        lut[streamlines_nearestp[lineId, pointId], 
+                                        streamlines_nearestp[lineId2, pointId2], 
+                                        hn+xd_mp[lineId], 
+                                        hn+yd_mp[lineId], 
+                                        hn+zd_mp[lineId]]  # ang_v, ang_r, x, y, z
+                                    
+                    streamline_scores[lineId, pointId] = score_mp[lineId]
+
+        if have_openmp and num_threads is not None:
+            openmp.omp_set_num_threads(all_cores)
```
arokem: Out of curiosity: what does this do for you here?

dipy/tracking/fbcmeasures.pyx
```diff
@@ -0,0 +1,334 @@
...
+        Contains the local fiber to bundle coherence (LFBC) for each streamline 
+        element.
+
+    Returns
+    ----------
+    output: normalized lowest average LFBC region along the fiber
+    """
+
+    # finds the region of the fiber with minimal length if 7 points in which the
+    # LFBC is the lowest
```
arokem: Where does the number 7 come from? Does that also need to be set as a key-word argument?

dipy/tracking/fbcmeasures.pyx
```diff
@@ -0,0 +1,334 @@
...
+    intLength = min(np.amin(streamlines_length), 7)
+    intValue = np.apply_along_axis(lambda x: min_moving_average(x[~np.isnan(x)], intLength), 
+                                    1, streamline_scores)
+    averageTotal = np.mean(np.apply_along_axis(
+                lambda x:np.mean(np.extract(x[~np.isnan(x)] >= 0, x[~np.isnan(x)])), 1, streamline_scores))
+    if not averageTotal == 0:
```
arokem: I might be missing something, but isn't averageTotal (or whatever it will be named, once it's not camelCased...) a 1D array? Doesn't apply_along_axis reduce it down only by one dimension?

arokem:
Exciting to have this follow-up! I haven't had an opportunity to look closely at the example yet. For now two general comments: 

- I haven't commented everywhere, but there's a lot of camelCased variables. Please follow PEP8, as much as possible on the variable naming. 

- Tests?

Great work!

dipy/tracking/fbcmeasures.pyx
stephanmeesters: I believe that the class variables need to be defined this way in Cython since they require an extension type. Check out this [Cython docpage](http://docs.cython.org/src/userguide/extension_types.html)

dipy/tracking/fbcmeasures.pyx
stephanmeesters: If I recall correctly, apply_along_axis is applied to streamline_scores which is a 2D array. It is applied to the first axis and by doing so loops over each fiber, and then applies the lambda function. This results in a 1D array of moving average values, and finally np.mean is used to get the total average of all moving average values (confusing I know..). In the end average_total is a number.

dipy/tracking/fbcmeasures.pyx
stephanmeesters: This line resets the number of OpenMP threads back to the default (all cores). I will add a comment there.

dipy/tracking/fbcmeasures.pyx
stephanmeesters: Not completely sure, but I think these arrays are already taking advantage of the Cython speedup because all those variables were defined with cdef (see all the cdef declaration in the top of the function). I checked them with ```cython -a fbcmeasures.pyx``` and they appeared OK.

dipy/tracking/fbcmeasures.pyx
arokem: Gotcha : I missed the effect of that call to np.mean! Makes perfect sense.

dipy/tracking/fbcmeasures.pyx
arokem: Awesome. Thanks for the pointer!

stephanmeesters:
Thanks for the review. Based on it I've made a lot of formatting changes and added parameters. A new test is added that runs the FBC code with two fibers and checks if the calculated RFBC value is correct.

dipy/tracking/tests/test_fbc.py
```diff
@@ -0,0 +1,44 @@
...
+
+    # run FBC
+    fbc = FBCMeasures(streamlines, k, verbose=True)
+
+    # get FBC values
+    fbc_sl_orig, clrs_orig, rfbc_orig = \
+        fbc.get_points_rfbc_thresholded(0, emphasis=0.01)
+
+    # check RFBC against tested value
+    npt.assert_almost_equal(np.mean(rfbc_orig), 1.0549502181194517)
```
arokem: Where did that number come from? Is this test passing on your machine (it's not passing on Travis currently: https://travis-ci.org/nipy/dipy/builds/122739850). Could this just be a typo?

stephanmeesters: The test passes on my Mac, so I'm not really sure why the Travis bots have a different outcome...

The Travis bots are running the code in parallel, however I'm testing it single thread. Maybe there's a bug somewhere that causes the values to be a little bit different, I'll have to investigate.

stephanmeesters: Running the code in parallel on my Mac has the same outcome. I've tested it on a Windows machine too with the exact same outcome. So I don't know why the Travis bots return ```1.0441612054810014``` instead of the desired number. Any ideas?

arokem: I can confirm that the test passes on my machine as well. Just to understand: the number in the test (``1.05...`) is based on running the code, or did you do a separate calculation to come up with this number?

arokem:
BTW - for some reason that is unclear to me as of yet, I am having trouble building the Cython on Python 3.5 on my machine, with this error: 

```

cythoning dipy/tracking/fbcmeasures.pyx to dipy/tracking/fbcmeasures.c
  File "setup.py", line 245, in <module>
    main(**extra_setuptools_args)
  File "setup.py", line 238, in main
    **extra_args
  File "/Users/arokem/anaconda/lib/python3.5/distutils/core.py", line 148, in setup
    dist.run_commands()
  File "/Users/arokem/anaconda/lib/python3.5/distutils/dist.py", line 955, in run_commands
    self.run_command(cmd)
  File "/Users/arokem/anaconda/lib/python3.5/distutils/dist.py", line 974, in run_command
    cmd_obj.run()
  File "/Users/arokem/anaconda/lib/python3.5/site-packages/Cython/Distutils/build_ext.py", line 164, in run
    _build_ext.build_ext.run(self)
  File "/Users/arokem/anaconda/lib/python3.5/distutils/command/build_ext.py", line 338, in run
    self.build_extensions()
  File "/Users/arokem/source/dipy/setup_helpers.py", line 194, in build_extensions
    build_ext_class.build_extensions(self)
  File "/Users/arokem/anaconda/lib/python3.5/site-packages/Cython/Distutils/build_ext.py", line 171, in build_extensions
    ext.sources = self.cython_sources(ext.sources, ext)
  File "/Users/arokem/anaconda/lib/python3.5/site-packages/Cython/Distutils/build_ext.py", line 324, in cython_sources
    full_module_name=module_name)
  File "/Users/arokem/anaconda/lib/python3.5/site-packages/Cython/Compiler/Main.py", line 682, in compile
    return compile_single(source, options, full_module_name)
  File "/Users/arokem/anaconda/lib/python3.5/site-packages/Cython/Compiler/Main.py", line 635, in compile_single
    return run_pipeline(source, options, full_module_name)
  File "/Users/arokem/anaconda/lib/python3.5/site-packages/Cython/Compiler/Main.py", line 492, in run_pipeline
    err, enddata = Pipeline.run_pipeline(pipeline, source)
  File "/Users/arokem/anaconda/lib/python3.5/site-packages/Cython/Compiler/Pipeline.py", line 365, in run_pipeline
    data = phase(data)
  File "/Users/arokem/anaconda/lib/python3.5/site-packages/Cython/Compiler/Pipeline.py", line 130, in inject_utility_code_stage
    tree = utilcode.get_tree(cython_scope=context.cython_scope)
TypeError: get_tree() takes no keyword arguments

```

This problem does not appear on master, so seems specific to something here.

arokem:
Oh - scratch that seems to be a general problem (just tried rebuilding with the `-f` flag)

Garyfallidis:
When you say a general problem?

arokem:
I mean I am getting this error on Python 3.5 (but not 2.7) on master.

Garyfallidis:
What is your cython version?

Garyfallidis:
If not latest I would suggest to upgrade.

arokem:
https://github.com/nipy/dipy/issues/1029

dipy/tracking/tests/test_fbc.py
stephanmeesters: Yes the number was obtained from running this test code and saving its output (the mean RFBC for two fibers).

arokem: Out of curiosity -- are you working on Anaconda Python? I am suspecting this could be the source of the difference... Otherwise/in addition, could you insert a few print statements into your code and pushing that, so that we can figure out where the divergence starts happening?

stephanmeesters: I'm not using Anaconda on my Mac and the Windows machine probably isn't either, but I will have to check. Adding print statements is a good idea, I'll look into it.

dipy/tracking/tests/test_fbc.py
```diff
@@ -0,0 +1,47 @@
...
+            fiber[j,0] = j
+            fiber[j,1] = i*0.2
+            fiber[j,2] = 0
+            streamlines.append(fiber)
+
+    # Create lookup table. Set seed for deterministic results.
+    D33 = 1.0
+    D44 = 0.04
+    t = 1
+    np.random.seed(1)
```
arokem: Where is the randomness elicited here?

dipy/tracking/fbcmeasures.pyx
```diff
@@ -0,0 +1,365 @@
...
+            Fibers with fewer points than minimum_length are excluded from FBC computation.
+        max_windowsize : int
+            The maximal window size used to calculate the average LFBC region
+        num_threads : int
+            Number of threads to use for OpenMP.
+        verbose : boolean
+            Enable verbose mode.
+        """
+        cdef:
+            int num_fibers
```
Garyfallidis: too many lines added here
you can do
int num_fibers, max_length, dim

dipy/tracking/fbcmeasures.pyx
Garyfallidis: true story

dipy/tracking/tests/test_fbc.py
```diff
@@ -0,0 +1,49 @@
+from dipy.denoise.enhancement_kernel import EnhancementKernel
```
Garyfallidis: We do not have automated coverage for cython files. Can you make a list of the functions that are being tested by this test? We need to make sure that all functions are tested.

doc/examples/fiber_to_bundle_coherence.py
```diff
@@ -0,0 +1,281 @@
...
+Fiber to bundle coherence measures
+==============================================
+
+This demo presents the fiber to bundle coherence (FBC) quantitative 
+measure of the alignment of each fiber with the surrounding fiber bundles
+[Meesters2016_HBM]_. These measures are useful in “cleaning” the results of 
+tractography algorithms, since low FBCs indicate which fibers are isolated and 
+poorly aligned with their neighbors, see Fig. 1.
+
+.. figure:: fbc_illustration.png
```
Garyfallidis: This figure is not available.

doc/examples/fiber_to_bundle_coherence.py
```diff
@@ -0,0 +1,281 @@
...
+   sum over all such locally aligned kernels. The local fiber to bundle coherence
+   shown on the right color-coded for each fiber, is obtained by evaluating the
+   kernel density estimator along the fibers. One spurious fiber is present which
+   is isolated and badly aligned with the other fibers, and can be identified 
+   by a low LFBC value in the region where it deviates from the bundle. 
+   Figure adapted from [Portegies2015_PLoSOne]_.
+
+Here we implement FBC measures based on kernel density estimation in the 
+non-flat 5D position-orientation domain. First we compute the kernel density 
+estimator induced by the full lifted output of the tractography. Then, the
```
Garyfallidis: lifted?

doc/examples/fiber_to_bundle_coherence.py
```diff
@@ -0,0 +1,281 @@
...
+
+.. figure:: OR_after.png
+   :align: center
+
+   The tractography result is cleaned (shown in bottom) by removing fibers with 
+   a relative FBC (RFBC) lower than the threshold tau=0.2.
+
+Acknowledgements
+~~~~~~~~~~~~~~~~~~~~~~
+Funded by the European Research Council under the European Community's Seventh
```
Garyfallidis: We usually do not put acknowledgements about funding in the tutorials.  People can have a look at the papers for the specifics. If you have especially received money for software development in DIPY then this needs to be added in another page. I hope this is okay.

dipy/tracking/tests/test_fbc.py
stephanmeesters: EnhancementKernel 
-  __init
- create_lookup_table
- estimate_kernel_size
- k2
- coordinate_map
- kernel
- euler_angles
- R
- get_lookup_table
- get_orientations
- get_sphere (not tested here, but tested in test_kernel.py)

FBCMeasures
- __init
- get_points_rfbc_thresholded
- compute
- compute_rfbc
- min_moving_average

All functions are tested

stephanmeesters:
Thanks for checking it out @Garyfallidis I've made some changes and added the figure.

arokem:
There are still a couple of comments outstanding, but I think it's very close to done!

dipy/tracking/tests/test_fbc.py
Garyfallidis: Nice thank you for this.

Garyfallidis:
@stephanmeesters are you on it? Have you seen @arokem's last message?

doc/examples/fiber_to_bundle_coherence.py
stephanmeesters: I have removed the funding acknowledgement here. If we make a future paper about this we can mention it there I guess, but it's already in the abstracts.

doc/examples/fiber_to_bundle_coherence.py
stephanmeesters: I've added some extra details here. Lifted means that it's defined in the space of positions and orientations.

doc/examples/fiber_to_bundle_coherence.py
stephanmeesters: I've included it in the recent commit.

dipy/tracking/tests/test_fbc.py
stephanmeesters: The randomness in this old version was inside EnhancementKernel, in which orientations were randomly chosen and electrostatic repulsion was used to distribute the orientations evenly on the sphere. However this game some issues with the bots, showing different outcomes despite setting the seed. This was fixed by giving the kernel a small set of orientations manually.

dipy/tracking/fbcmeasures.pyx
stephanmeesters: This managed to shorten it a couple of lines

dipy/tracking/fbcmeasures.pyx
stephanmeesters: The number 7 was something we found experimentally a long time ago, unfortunately I don't have good evidency why. However it's now available as a parameter to the user.

dipy/tracking/fbcmeasures.pyx
stephanmeesters: I've added a verbose parameter

dipy/tracking/fbcmeasures.pyx
stephanmeesters: See the previous answer regarding cdef

dipy/tracking/fbcmeasures.pyx
stephanmeesters: The streamlines are not of equal length so unfortunately calling shape on py_streamlines won't work. In order to gain speed I converter the Python array ```py_streamlines``` with variable length into a fixed-size 2D Numpy array based on the longest fiber. The undefined fiber points of the shorter fibers are set to NaN.

dipy/tracking/fbcmeasures.pyx
stephanmeesters: It's been added as a parameter

dipy/tracking/fbcmeasures.pyx
stephanmeesters: Fixed!

dipy/tracking/fbcmeasures.pyx
stephanmeesters: It's been changed to ```verbose```

dipy/tracking/fbcmeasures.pyx
stephanmeesters: For some reason ```np.min``` led to some problems when I tried it, so I stuck with Python's min.

dipy/tracking/fbcmeasures.pyx
stephanmeesters: Fixed

stephanmeesters:
Sorry I forgot to respond to a couple of comments, I think I got them all now. Let me know if I missed something.

dipy/tracking/fbcmeasures.pyx
arokem: Oh - sorry. Should've been more specific. I meant something like:

     [x.shape[0] for x in py_streamlines]

arokem: And I don't think that we want to allocate the fixed-size numpy array. That will be a huge memory hog for large sets of streamlines!

dipy/tracking/fbcmeasures.pyx
```diff
@@ -0,0 +1,358 @@
...
+        
+        # if the fibers are too short FBC measures cannot be applied, 
+        # remove these.
+        streamlines_length = np.array([len(x) for x in py_streamlines], 
+                                dtype=np.int32)
+        min_length = min(streamlines_length)
+        if min_length < min_fiberlength:
+            print("The minimum fiber length is 10 points. \
+                    Shorter fibers were found and removed.")
+            py_streamlines = [x for x in py_streamlines if len(x) >= min_fiberlength]
```
arokem: By the way, here as well: 

    [x for x in py_streamlines if x.shape[0 >= min_fiberlength]

would be faster

stephanmeesters: OK I see, thanks!

dipy/tracking/fbcmeasures.pyx
stephanmeesters: In case of outliers in fiber length that is true. I will have a look at changing it to an array of arrays implementation. Sorry for delayed reactions -- I am currently attending the ISMRM

stephanmeesters:
Regarding the fixed-size array for the streamlines, what I've found so far is that Numpy is really designed for contiguous memory. An array of pointer to ndarray's doesn't appear to be a straightforward thing.

Streamlines in DIPY are generally represented by a Python list of ndarrays, but that doesn't work well for Cython since it requires GIL (basically losing all speed benefits of Cython).

So the options are

- Rewrite everything to c-style arrays, however this will make interfacing with Numpy functions harder
- Rewrite it to use a 1D array and use an index map to reference each fiber. This *may* lead to perforce loss since I'm not sure if selecting a part of an 1D array is expensive.

Let me know what you think.

arokem:
This will probably all become easier once https://github.com/nipy/nibabel/pull/391 is merged, and we start refactoring things here.

For now, my take is that we should punt on this, and use the nipy convention. What are your thoughts, @Garyfallidis ?

stephanmeesters:
Perhaps we can address this issue in a later pull request? The HBM conference is coming up and it would be great if I can put a link to the demo on my poster.

arokem:
This looks really close to ready to me. From my point of view the only thing remaining is the change of those two calls to `len` to references to the array `shape` attributes. @Garyfallidis : was there anything else you wanted addressed here before the merge?

Regarding link on your poster -- you should know that this will not appear on the dipy website until a release is made. You'll be able to link to the documentation examples on the `master` branch though.

stephanmeesters:
I've made the requested changes, please see this new commit. Regarding the link, would I be linking directly to the source code or is there perhaps a parsed HTML page available somewhere of the master branch?

arokem:
This is good to go from my point of view. @Garyfallidis: did you have any other comments here? 

The link would be something like this: https://github.com/nipy/dipy/blob/master/doc/examples/contextual_enhancement.py

For the time being, we don't have a rendered html of the documentation as it is currently on master, but that's probably going to change soon with the GSoC project on that.

doc/examples/fiber_to_bundle_coherence.py
```diff
@@ -0,0 +1,280 @@
+# -*- coding: utf-8 -*-
```
Garyfallidis: Isn't the same figure already available at http://nipy.org/dipy/_images/stochastic_process.jpg??
Please delete the figure from this PR. Although tutorials can generate pictures you cannot add external pictures in the examples folder. Please use the existing picture as you did in the other example!
Thanks in advance!

stephanmeesters: Woops, I uploaded the wrong figure actually. I've now added ```fbc_illustration.png``` to doc/examples_built/_static, which is the same folder where ```stochastic_process.jpg``` was also placed. Let me know if this is OK.

Garyfallidis:
I still cannot merge this because of the many pep8 errors. Look for example at the example fiber_to_bundle_coherence.py

Please correct all pep8 errors asap so we can merge this hopefully before the conference and the release.

stephanmeesters:
I've run it through a PEP8 checker and cleaned up the code. The trailing white spaces in figure captions are kept since those were giving problems before I remember.

stephanmeesters:
Just checking if you noticed the last commit. Do you think it's ready for merging?

arokem:
@Garyfallidis : what are your thoughts?

arokem:
@Garyfallidis : this is waiting for your confirmation.

Garyfallidis:
Thank you Stephan. I am merging this now. But please @stephanmeesters make another small PR after where you remove lines 45, 222 and 223 of the tutorials as they are not used. Also in the same tutorial correct pep 8 for line 156.

stephanmeesters:
Thanks. Will do.

Issue #813, opened 2015-12-16
Windows 64-bit error in segment.featurespeed.extract


matthew-brett:
```
======================================================================
ERROR: dipy.segment.tests.test_feature.test_feature_extract
----------------------------------------------------------------------
Traceback (most recent call last):
  File "C:\buildslaves\mike-win7-64\dipy-bdist64-27\build\venv\lib\site-packages\nose\case.py", line 197, in runTest
    self.test(*self.arg)
  File "C:\buildslaves\mike-win7-64\dipy-bdist64-27\build\venv\lib\site-packages\dipy\segment\tests\test_feature.py", line 231, in test_feature_extract
    features = extract(feature, streamlines)
  File "dipy\segment\featurespeed.pyx", line 371, in dipy.segment.featurespeed.extract (dipy\segment\featurespeed.c:6302)
  File "dipy\segment\featurespeed.pyx", line 394, in dipy.segment.featurespeed.extract (dipy\segment\featurespeed.c:5973)
  File "dipy\segment\featurespeed.pyx", line 363, in dipy.segment.featurespeed.infer_shape (dipy\segment\featurespeed.c:5661)
  File "dipy\segment\featurespeed.pyx", line 46, in dipy.segment.featurespeed.Feature.c_infer_shape (dipy\segment\featurespeed.c:2597)
TypeError: object of type 'long' has no len()
```

See: http://nipy.bic.berkeley.edu/builders/dipy-bdist64-27/builds/9/steps/shell_11/logs/stdio

One error in series reported at gh-803

Garyfallidis:
Hi @MarcCote this is an error that fails only in a few windows machines. You may not have a windows machine or access to a buildbot to play with this right now but I can fix or help you fix this if you give me some feedback with the following. 

Basically, in featurespeed.pyx you check in line 44 if the type(shape) is int. But some times this type may come as long. Because in this buildbot 
http://nipy.bic.berkeley.edu/builders/dipy-bdist64-27/builds/14/steps/shell_11/logs/stdio
the shape comes as long you immediately jump to line 46 where you expect to check something that you can apply a ``len`` to.

Now we could just go and replace in line 44 the existing line ``if type(shape) is int:`` with ``if type(shape) in [int, long]`` but maybe there is a safer/better way. What do you think?

matthew-brett:
How about:

```
try:
    ndim = len(shape)
except TypeError:
    return tuple2shape((1, shape))
```

instead of:

```
[...]

PR #829, opened 2016-01-07
Fixes issue #813 by not checking data type explicitly.

MarcCote:
According to issue #813, on Windows 64bit machines, integer constants are of type `long` instead of `int`. I added a unit test that explicitly test using a `long` as the type of the value returned by `feature.infer_shape`.

This PR also fixes another issue that came up in `metric.dist` while creating the new unit test.

### DIFF
--- a/dipy/segment/tests/test_feature.py
+++ b/dipy/segment/tests/test_feature.py
@@ -286,7 +286,7 @@ def test_using_python_feature_with_cython_metric():
 
     class ArcLengthFeature(dipymetric.Feature):
         def infer_shape(self, streamline):
-            return long(1)
+            return np.int64(1)
 
         def extract(self, streamline):
             return np.sum(np.sqrt(np.sum((streamline[1:] - streamline[:-1]) ** 2)))
--- a/dipy/segment/tests/test_feature.py
+++ b/dipy/segment/tests/test_feature.py
@@ -1,3 +1,4 @@
+import sys
 import numpy as np
 import dipy.segment.metric as dipymetric
 from dipy.segment.featurespeed import extract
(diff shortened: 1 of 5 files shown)


arokem:
Well - this fix seems to work... in Python 2.

See: http://python3porting.com/differences.html#long

MarcCote:
Alright, I return `1` or `long(1)` whether `sys.version_info > (3,)` or not.

arokem:
LGTM. Let's wait and see what Travis says. In the meanwhile, I've started builds on the 64 bit Windows machine that was having trouble before:

http://nipy.bic.berkeley.edu/builders/dipy-bdist64-27/builds/21
http://nipy.bic.berkeley.edu/builders/dipy-bdist64-35/builds/23

arokem:
Yeah - everything's coming back green. Any objections from anyone? I'll wait for a few days for others to take a look and then merge.

dipy/segment/tests/test_feature.py
```diff
@@ -284,6 +285,26 @@ def extract(self, streamline):
     d2 = metric.dist(features1, features2)
     assert_equal(d1, d2)
 
+    class ArcLengthFeature(dipymetric.Feature):
+        def infer_shape(self, streamline):
+            if sys.version_info > (3,):
```
matthew-brett: I find this very ugly.   Is there any way of avoiding it?

arokem: Would this do it? https://github.com/MarcCote/dipy/commit/13687569f666d499f84735c97910cd4312ba41b9

matthew-brett: Less ugly but still hard to understand.  Could y'all remind me why this is necessary?  Isn't the return type dealt with by https://github.com/nipy/dipy/pull/829/files#diff-6102332b6a9436b5c5b18ab9ee552a9eR169 ?

matthew-brett: Oh - sorry - I see you are trying to test the Windows behavior on non-Windows machines - maybe just a comment to that effect.

MarcCote: As @matthew-brett said I'm trying to test the Windows 64 bits behavior, i.e. where it uses `long` instead of `int` for constant integers. I added a comment explaining  why we do such test.

Garyfallidis:
Fine with me too to merge this.

PR #835, opened 2016-01-14
NF: Free water tensor model

RafaelNH:
In this pull request the free water tensor model will be added to Dipy. The model's fit will be based on the method proposed in [http://www.ncbi.nlm.nih.gov/pubmed/25271843](http://www.ncbi.nlm.nih.gov/pubmed/25271843). 

As @arokem suggested this technique will be implemented in a separate reconst module which I named fwdti.py. Since this technique is an expansion of the simple DTI model, its classes will be defined from inheritance of the classes defined on Dipy's DTI module. 

At the moment, I just add the structures for the fwdti class. My very next step will be editing class FreeWaterTensorModel.

### DIFF
--- a/dipy/reconst/fwdti.py
+++ b/dipy/reconst/fwdti.py
@@ -69,9 +69,9 @@ def fwdti_prediction(params, gtab, Diso=3.0e-3):
     index = ndindex(f.shape)
     for v in index:
         if mask[v]:
-            pre_pred_sig = S0[v] * \
-            ((1 - f[v]) * np.exp(-gtab.bvals[~gtab.b0s_mask] * adc[v]) +
-            f[v] * np.exp(-gtab.bvals[~gtab.b0s_mask] * Diso))
+            S_tissue = (1-f[v]) * np.exp(-gtab.bvals[~gtab.b0s_mask] * adc[v])
+            S_water = f[v] * np.exp(-gtab.bvals[~gtab.b0s_mask] * Diso)
+            pre_pred_sig = S0[v] * (S_tissue+S_water)
 
             # Then we need to sort out what goes where:
             pred_s = np.zeros(pre_pred_sig.shape[:-1] + (gtab.bvals.shape[0],))
@@ -220,7 +220,7 @@ class FreeWaterTensorFit(TensorFit):
     def f(self):
         """ Returns the free water diffusion volume fraction f """
         return self.model_params[..., 12]
-        
+
     @property
     def S0(self):
         """ Returns the non-diffusion weighted signal estimate """
@@ -304,7 +304,7 @@ def _wls_iter(design_matrix, inv_design, sig, min_diffusivity, Diso=3e-3,
     params = np.dot(invWTS2W_WTS2, log_s)
 
     # General free-water signal contribution
-    fwsig = np.exp(np.dot(design_matrix, 
+    fwsig = np.exp(np.dot(design_matrix,
                           np.array([Diso, 0, Diso, 0, 0, Diso, 0])))
 
     df = 1  # initialize precision
@@ -316,7 +316,7 @@ def _wls_iter(design_matrix, inv_design, sig, min_diffusivity, Diso=3e-3,
         for p in range(piterations):
             df = df * 0.1
             fs = np.linspace(flow+df, fhig-df, num=ns)  # sampling f
-            SFW = np.array([fwsig,]*ns)  # repeat contributions for all values
+            SFW = np.array([fwsig, ]*ns)  # repeat contributions for all values
             FS, SI = np.meshgrid(fs, sig)
             SA = SI - FS*S0*SFW.T
             # SA < 0 means that the signal components from the free water
@@ -340,7 +340,7 @@ def _wls_iter(design_matrix, inv_design, sig, min_diffusivity, Diso=3e-3,
         for p in range(piterations):
             df = df * 0.1
             fs = np.linspace(flow+df, fhig-df, num=ns)  # sampling f
-            SFW = np.array([fwsig,]*ns)  # repeat contributions for all values
+            SFW = np.array([fwsig, ]*ns)  # repeat contributions for all values
             FS, SI = np.meshgrid(fs, sig)
             S0 = np.exp(-params[6])  # S0 is now taken as a model parameter
             SA = SI - FS*S0*SFW.T
@@ -349,7 +349,7 @@ def _wls_iter(design_matrix, inv_design, sig, min_diffusivity, Diso=3e-3,
             all_new_params = np.dot(invWTS2W_WTS2, y)
 
             # Select params for lower F2
-            S0r = np.exp(-np.array([all_new_params[6],]*nvol))
+            S0r = np.exp(-np.array([all_new_params[6], ]*nvol))
             SIpred = (1-FS)*np.exp(np.dot(W, all_new_params)) + FS*S0r*SFW.T
             F2 = np.sum(np.square(SI - SIpred), axis=0)
             Mind = np.argmin(F2)
@@ -362,7 +362,7 @@ def _wls_iter(design_matrix, inv_design, sig, min_diffusivity, Diso=3e-3,
 
     evals, evecs = decompose_tensor(from_lower_triangular(params),
                                     min_diffusivity=min_diffusivity)
-    fw_params = np.concatenate((evals, evecs[0], evecs[1], evecs[2], 
+    fw_params = np.concatenate((evals, evecs[0], evecs[1], evecs[2],
                                 np.array([f]), np.array([S0])), axis=0)
     return fw_params
 
@@ -452,8 +452,7 @@ def _nlls_err_func(tensor_elements, design_matrix, data, Diso=3e-3,
         compartment. Note that if cholesky is set to true, tensor elements are
         assumed to be written as Cholesky's decomposition elements. If
         f_transform is true, volume fraction f has to be converted to
-        ft = arcsin(2*f - 1) + pi/2 
-
+        ft = arcsin(2*f - 1) + pi/2
     design_matrix : array
         The design matrix
 
@@ -471,7 +470,7 @@ def _nlls_err_func(tensor_elements, design_matrix, data, Diso=3e-3,
         noise in each diffusion-weighting direction (if an array is
         provided). If 'gmm', the Geman-Mclure M-estimator is used for
         weighting.
-    cholesky : bool, optional 
+    cholesky : bool, optional
         If true, the diffusion tensor elements were decomposed using cholesky
         decomposition. See fwdti.nlls_fit_tensor
         Default: False
@@ -494,8 +493,7 @@ def _nlls_err_func(tensor_elements, design_matrix, data, Diso=3e-3,
     # This is the predicted signal given the params:
     y = (1-f) * np.exp(np.dot(design_matrix, tensor[:7])) + \
         f * np.exp(np.dot(design_matrix,
-                          np.array([Diso, 0, Diso, 0, 0, Diso, 
-                                    tensor[6]])))
+                          np.array([Diso, 0, Diso, 0, 0, Diso, tensor[6]])))
 
     # Compute the residuals
     residuals = data - y
@@ -570,14 +568,13 @@ def nlls_fit_tensor(design_matrix, data, fw_params=None, Diso=3e-3,
         provided here. According to [Chang2005]_, a good value to use is
         1.5267 * std(background_noise), where background_noise is estimated
         from some part of the image known to contain no signal (only noise).
-
-    cholesky : bool, optional 
+    cholesky : bool, optional
         If true it uses cholesky decomposition to insure that diffusion tensor
         is positive define.
         Default: False
 
     f_transform : bool, optional
-        If true, the water volume fractions is converted during the convergence 
+        If true, the water volume fractions is converted during the convergence
         procedure to ft = arcsin(2*f - 1) + pi/2, insuring f estimates between
         0 and 1.
         Default: True
@@ -591,7 +588,7 @@ def nlls_fit_tensor(design_matrix, data, fw_params=None, Diso=3e-3,
     flat_data = data.reshape((-1, data.shape[-1]))
 
     # Use the WLS method parameters as the starting point if fw_params is None:
-    if fw_params==None:
+    if fw_params is None:
         fw_params = wls_fit_tensor(design_matrix, flat_data,  Diso=Diso)
 
     fw_params = fw_params.reshape((-1, fw_params.shape[-1]))
@@ -599,7 +596,7 @@ def nlls_fit_tensor(design_matrix, data, fw_params=None, Diso=3e-3,
     # if bounds==None:
     #     bounds = ([0., -Diso, 0., -Diso, -Diso, 0., -10., 0.],
     #               [Diso, Diso, Diso, Diso, Diso, Diso, 10., 1.])
-    
+
     for vox in range(flat_data.shape[0]):
         if np.all(flat_data[vox] == 0):
             raise ValueError("The data in this voxel contains only zeros")
@@ -650,7 +647,7 @@ def nlls_fit_tensor(design_matrix, data, fw_params=None, Diso=3e-3,
 
         # Invert f transformation if this was requested
         if f_transform:
-            this_tensor[7] =  0.5 * (1 + np.sin(this_tensor[7] - np.pi/2))
+            this_tensor[7] = 0.5 * (1 + np.sin(this_tensor[7] - np.pi/2))
 
         # The parameters are the evals and the evecs:
         fw_params[vox, 12] = this_tensor[7]
@@ -664,8 +661,8 @@ def nlls_fit_tensor(design_matrix, data, fw_params=None, Diso=3e-3,
 
 
 def lower_triangular_to_cholesky(tensor_elements):
-    """ Perfoms Cholesky decompostion of the diffusion tensor 
-    
+    """ Perfoms Cholesky decompostion of the diffusion tensor
+
     Parameters
     ----------
     tensor_elements : array (6,)
@@ -697,12 +694,12 @@ def lower_triangular_to_cholesky(tensor_elements):
 
 def cholesky_to_lower_triangular(R):
     """ Convert Cholesky decompostion elements to the diffusion tensor elements
-    
+
     Parameters
     ----------
     R : array (6,)
         Array containing the six Cholesky's decomposition elements
-        (R0, R1, R2, R3, R4, R5) [1]_.    
+        (R0, R1, R2, R3, R4, R5) [1]_.
 
     Returns
     -------
--- a/dipy/reconst/fwdti.py
+++ b/dipy/reconst/fwdti.py
@@ -252,8 +252,8 @@ class FreeWaterTensorFit(TensorFit):
         return predict.reshape(shape + (gtab.bvals.shape[0], ))
 
 
-def _wls_iter(design_matrix, inv_design, sig, min_diffusivity, Diso=3e-3,
-              piterations=3, S0=None):
+def _wls_iter(design_matrix, sig, min_diffusivity, Diso=3e-3, piterations=3,
+              S0=None):
     """ Helper function used by wls_fit_tensor - Applies WLS fit of the
     water free elimination model to single voxel signals.
 
@@ -439,7 +439,7 @@ def wls_fit_tensor(design_matrix, data, S0=None, Diso=3e-3, piterations=3,
         S0 = np.zeros(len(data_flat))
         S0_p = np.zeros(len(data_flat_p))
         for vox in range(len(data_flat_p)):
-            fw_params_p[vox], S0_p[vox] = _wls_iter(design_matrix, inv_design,
+            fw_params_p[vox], S0_p[vox] = _wls_iter(design_matrix,
                                                     data_flat_p[vox],
                                                     min_diffusivity,
                                                     Diso=Diso,
@@ -448,7 +448,7 @@ def wls_fit_tensor(design_matrix, data, S0=None, Diso=3e-3, piterations=3,
         S0 = S0.ravel()
         S0_p = S0[~cond]
         for vox in range(len(data_flat_p)):
-            fw_params_p[vox], S0_p[vox] = _wls_iter(design_matrix, inv_design,
+            fw_params_p[vox], S0_p[vox] = _wls_iter(design_matrix,
                                                     data_flat_p[vox],
                                                     min_diffusivity,
                                                     Diso=Diso,
--- a/dipy/reconst/fwdti.py
+++ b/dipy/reconst/fwdti.py
@@ -594,16 +594,17 @@ def nlls_fit_tensor(design_matrix, data, fw_params=None, Diso=3e-3,
         evals = params[:3]
         evecs = params[3:12].reshape((3, 3))
         dt = lower_triangular(np.dot(np.dot(evecs, np.diag(evals)), evecs.T))
-        f = params[12]
         s0 = params[13]
 
         # Cholesky decomposition if requested
         if cholesky:
             dt = lower_triangular_to_cholesky(dt)
-        
+
         # f transformation if requested
         if f_transform:
-            f = np.arcsin(2*f - 1) + np.pi/2
+            f = np.arcsin(2*params[12] - 1) + np.pi/2
+        else:
+            f = params[12]
 
         # Use the Levenberg-Marquardt algorithm wrapped in opt.leastsq
         start_params = np.concatenate((dt, [-np.log(s0), f]), axis=0)
@@ -625,11 +626,11 @@ def nlls_fit_tensor(design_matrix, data, fw_params=None, Diso=3e-3,
             this_tensor[7] =  0.5 * (1 + np.sin(this_tensor[7] - np.pi/2))
 
         # The parameters are the evals and the evecs:
+        fw_params[vox, 12] = this_tensor[7]
+        fw_params[vox, 13] = np.exp(-this_tensor[6])
         evals, evecs = decompose_tensor(from_lower_triangular(this_tensor[:6]))
         fw_params[vox, :3] = evals
         fw_params[vox, 3:12] = evecs.ravel()
-        fw_params[vox, 12] = this_tensor[7]
-        fw_params[vox, 13] = np.exp(-this_tensor[6])
 
     fw_params.shape = data.shape[:-1] + (14,)
     return fw_params
--- a/dipy/reconst/fwdti.py
+++ b/dipy/reconst/fwdti.py
@@ -458,7 +458,10 @@ def _nlls_err_func(tensor_elements, design_matrix, data, Diso=3e-3,
 
     data : array
         The voxel signal in all gradient directions
-
+    Diso : float, optional
+        Value of the free water isotropic diffusion. Default is set to 3e-3
+        $mm^{2}.s^{-1}$. Please ajust this value if you are assuming different
+        units of diffusion.
     weighting : str (optional).
          Whether to use the Geman McClure weighting criterion (see [1]_
          for details)
@@ -528,9 +531,54 @@ def _nlls_err_func(tensor_elements, design_matrix, data, Diso=3e-3,
         return np.sqrt(w * se)
 
 
+def _nlls_jacobian_func(tensor_elements, design_matrix, data, Diso=3e-3,
+                        weighting=None, sigma=None, cholesky=False,
+                        f_transform=False):
+    """The Jacobian is the first derivative of the least squares error
+    function.
+
+    Parameters
+    ----------
+    tensor_elements : array (8, )
+        The six independent elements of the diffusion tensor followed by
+        -log(S0) and the volume fraction f of the water elimination
+        compartment. Note that if f_transform is true, volume fraction f is
+        converted to ft = arcsin(2*f - 1) + pi/2
+    design_matrix : array
+        The design matrix
+    Diso : float, optional
+        Value of the free water isotropic diffusion. Default is set to 3e-3
+        $mm^{2}.s^{-1}$. Please ajust this value if you are assuming different
+        units of diffusion.
+    f_transform : bool, optional
+        If true, the water volume fraction was converted to
+        ft = arcsin(2*f - 1) + pi/2, insuring f estimates between 0 and 1.
+        See fwdti.nlls_fit_tensor
+        Default: True
+    """
+    tensor = np.copy(tensor_elements)
+    if f_transform:
+        f = 0.5 * (1 + np.sin(tensor[7] - np.pi/2))
+    else:
+        f = tensor[7]
+
+    t = np.exp(np.dot(design_matrix, tensor[:7]))
+    s = np.exp(np.dot(design_matrix,
+                      np.array([Diso, 0, Diso, 0, 0, Diso, tensor[6]])))
+    T = (f-1.0) * t[:, None] * design_matrix
+    S = np.zeros(design_matrix.shape)
+    S[:, 6] = f * s
+
+    if f_transform:
+        df = (t-s) * (0.5*np.cos(tensor[7]-np.pi/2))
+    else:
+        df = (t-s)
+    return np.concatenate((T - S, df[:, None]), axis=1)
+
+
 def nlls_fit_tensor(design_matrix, data, fw_params=None, Diso=3e-3,
                     weighting=None, sigma=None, cholesky=False,
-                    f_transform=True):
+                    f_transform=True, jac=False):
     """
     Fit the water elimination tensor model using the non-linear least-squares.
 
@@ -578,6 +626,8 @@ def nlls_fit_tensor(design_matrix, data, fw_params=None, Diso=3e-3,
         procedure to ft = arcsin(2*f - 1) + pi/2, insuring f estimates between
         0 and 1.
         Default: True
+    jac : bool
+        Use the Jacobian? Default: True
 
     Returns
     -------
@@ -594,9 +644,10 @@ def nlls_fit_tensor(design_matrix, data, fw_params=None, Diso=3e-3,
         fw_params = fw_params.reshape((-1, fw_params.shape[-1]))
 
     # if bounds==None:
-    #     bounds = ([0., -Diso, 0., -Diso, -Diso, 0., -10., 0.],
-    #               [Diso, Diso, Diso, Diso, Diso, Diso, 10., 1.])
-
+    """
+    bounds = ([0., -Diso, 0., -Diso, -Diso, 0., -10., 0],
+              [Diso, Diso, Diso, Diso, Diso, Diso, 10., 1])
+    """
     for vox in range(flat_data.shape[0]):
         if np.all(flat_data[vox] == 0):
             raise ValueError("The data in this voxel contains only zeros")
@@ -621,25 +672,46 @@ def nlls_fit_tensor(design_matrix, data, fw_params=None, Diso=3e-3,
 
         # Use the Levenberg-Marquardt algorithm wrapped in opt.leastsq
         start_params = np.concatenate((dt, [-np.log(s0), f]), axis=0)
-        # lb = np.array(bounds[0])
-        # ub = np.array(bounds[1])
-        # start_params[start_params<lb] = lb[start_params<lb]
-        # start_params[start_params>ub] = ub[start_params>ub]
-        # out = opt.least_squares(_nlls_err_func, start_params[:8],
-        #                         args=(design_matrix, flat_data[vox],
-        #                              Diso, weighting, sigma, cholesky,
-        #                              f_transform),
-        #                         bounds=bounds)
-        # this_tensor = out.x
+        """
+        lb = np.array(bounds[0])
+        ub = np.array(bounds[1])
+        start_params[start_params<lb] = lb[start_params<lb]
+        start_params[start_params>ub] = ub[start_params>ub]       
+        if jac:
+            out = opt.least_squares(_nlls_err_func, start_params[:8],
+                                    args=(design_matrix, flat_data[vox],
+                                          Diso, weighting, sigma, cholesky,
+                                          f_transform),
+                                    jac=_nlls_jacobian_func,
+                                    bounds=bounds)
+        else:
+            out = opt.least_squares(_nlls_err_func, start_params[:8],
+                                    args=(design_matrix, flat_data[vox],
+                                          Diso, weighting, sigma, cholesky,
+                                          f_transform),
+                                    bounds=bounds)
+        this_tensor = out.x
+        """
+        if jac:
+            this_tensor, status = opt.leastsq(_nlls_err_func, start_params[:8],
+                                              args=(design_matrix,
+                                                    flat_data[vox],
+                                                    Diso,
+                                                    weighting,
+                                                    sigma,
+                                                    cholesky,
+                                                    f_transform),
+                                              Dfun=_nlls_jacobian_func)
+        else:
+            this_tensor, status = opt.leastsq(_nlls_err_func, start_params[:8],
+                                              args=(design_matrix,
+                                                    flat_data[vox],
+                                                    Diso,
+                                                    weighting,
+                                                    sigma,
+                                                    cholesky,
+                                                    f_transform))
 
-        this_tensor, status = opt.leastsq(_nlls_err_func, start_params[:8],
-                                          args=(design_matrix,
-                                                flat_data[vox],
-                                                Diso,
-                                                weighting,
-                                                sigma,
-                                                cholesky,
-                                                f_transform))
 
         # Invert the cholesky decomposition if this was requested
         if cholesky:
--- a/dipy/reconst/fwdti.py
+++ b/dipy/reconst/fwdti.py
@@ -426,7 +426,8 @@ def wls_fit_tensor(design_matrix, data, Diso=3e-3, piterations=3,
 
 
 def _nlls_err_func(tensor_elements, design_matrix, data, Diso=3e-3,
-                   weighting=None, sigma=None, cholesky=False):
+                   weighting=None, sigma=None, cholesky=False,
+                   f_transform=False):
     """ Error function for the non-linear least-squares fit of the tensor water
     elimination model.
 
@@ -434,7 +435,11 @@ def _nlls_err_func(tensor_elements, design_matrix, data, Diso=3e-3,
     ----------
     tensor_elements : array (8, )
         The six independent elements of the diffusion tensor followed by
-        -log(S0) and the volume fraction f of the water elimination compartment
+        -log(S0) and the volume fraction f of the water elimination
+        compartment. Note that if cholesky is set to true, tensor elements are
+        assumed to be written as Cholesky's decomposition elements. If
+        f_transform is true, volume fraction f has to be converted to
+        ft = arcsin(2*f - 1) + pi/2 
 
     design_matrix : array
         The design matrix
@@ -455,16 +460,24 @@ def _nlls_err_func(tensor_elements, design_matrix, data, Diso=3e-3,
         weighting.
     
     cholesky : bool, optional 
-        If true it uses cholesky decomposition to insure that diffusion tensor
-        is positive define.
+        If true, the diffusion tensor elements were decomposed using cholesky
+        decomposition. See fwdti.nlls_fit_tensor
+        Default: False
+
+    f_transform : bool, optional
+        If true, the water volume fraction was converted to
+        ft = arcsin(2*f - 1) + pi/2, insuring f estimates between 0 and 1.
+        See fwdti.nlls_fit_tensor
         Default: True
     """
     tensor = np.copy(tensor_elements)
     if cholesky:
         tensor[:6] = cholesky_to_lower_triangular(tensor[:6])
+
+    if f_transform:
         f = 0.5 * (1 + np.sin(tensor[7] - np.pi/2))
     else:
-        f = np.copy(tensor[7])
+        f = tensor[7]
 
     # This is the predicted signal given the params:
     y = (1-f) * np.exp(np.dot(design_matrix, tensor[:7])) + \
@@ -506,7 +519,8 @@ def _nlls_err_func(tensor_elements, design_matrix, data, Diso=3e-3,
 
 
 def nlls_fit_tensor(design_matrix, data, fw_params=None, Diso=3e-3,
-                    weighting=None, sigma=None, cholesky=False):
+                    weighting=None, sigma=None, cholesky=False,
+                    f_transform=True):
     """
     Fit the water elimination tensor model using the non-linear least-squares.
 
@@ -548,6 +562,12 @@ def nlls_fit_tensor(design_matrix, data, fw_params=None, Diso=3e-3,
     cholesky : bool, optional 
         If true it uses cholesky decomposition to insure that diffusion tensor
         is positive define.
+        Default: False
+
+    f_transform : bool, optional
+        If true, the water volume fractions is converted during the convergence 
+        procedure to ft = arcsin(2*f - 1) + pi/2, insuring f estimates between
+        0 and 1.
         Default: True
 
     Returns
@@ -577,24 +597,31 @@ def nlls_fit_tensor(design_matrix, data, fw_params=None, Diso=3e-3,
         f = params[12]
         s0 = params[13]
 
+        # Cholesky decomposition if requested
         if cholesky:
-            start_params = np.concatenate((lower_triangular_to_cholesky(dt),
-                                           [-np.log(s0),
-                                            np.arcsin(2*f - 1) + np.pi/2]),
-                                          axis=0)
-        else:
-            start_params = np.concatenate((dt, [-np.log(s0), f]), axis=0)
+            dt = lower_triangular_to_cholesky(dt)
+        
+        # f transformation if requested
+        if f_transform:
+            f = np.arcsin(2*f - 1) + np.pi/2
 
+        # Use the Levenberg-Marquardt algorithm wrapped in opt.leastsq
+        start_params = np.concatenate((dt, [-np.log(s0), f]), axis=0)
         this_tensor, status = opt.leastsq(_nlls_err_func, start_params[:8],
                                           args=(design_matrix,
                                                 flat_data[vox],
                                                 Diso,
                                                 weighting,
                                                 sigma,
-                                                cholesky))
+                                                cholesky,
+                                                f_transform))
                                                 
+        # Invert the cholesky decomposition if this was requested
         if cholesky:
             this_tensor[:6] = cholesky_to_lower_triangular(this_tensor[:6])
+
+        # Invert f transformation if this was requested
+        if f_transform:
             this_tensor[7] =  0.5 * (1 + np.sin(this_tensor[7] - np.pi/2))
 
         # The parameters are the evals and the evecs:
--- a/dipy/reconst/fwdti.py
+++ b/dipy/reconst/fwdti.py
@@ -426,7 +426,7 @@ def _wls_iter(design_matrix, inv_design, sig, min_diffusivity, Diso=3e-3,
 
 
 def _nlls_err_func(tensor_elements, design_matrix, data, Diso=3e-3,
-                   weighting=None, sigma=None):
+                   weighting=None, sigma=None, cholesky=False):
     """ Error function for the non-linear least-squares fit of the tensor water
     elimination model.
 
@@ -453,13 +453,24 @@ def _nlls_err_func(tensor_elements, design_matrix, data, Diso=3e-3,
         noise in each diffusion-weighting direction (if an array is
         provided). If 'gmm', the Geman-Mclure M-estimator is used for
         weighting.
+    
+    cholesky : bool, optional 
+        If true it uses cholesky decomposition to insure that diffusion tensor
+        is positive define.
+        Default: True
     """
-    f = tensor_elements[7]
+    tensor = np.copy(tensor_elements)
+    if cholesky:
+        tensor[:6] = cholesky_to_lower_triangular(tensor[:6])
+        f = 0.5 * (1 + np.sin(tensor[7] - np.pi/2))
+    else:
+        f = np.copy(tensor[7])
+
     # This is the predicted signal given the params:
-    y = (1-f) * np.exp(np.dot(design_matrix, tensor_elements[:7])) + \
+    y = (1-f) * np.exp(np.dot(design_matrix, tensor[:7])) + \
         f * np.exp(np.dot(design_matrix,
                           np.array([Diso, 0, Diso, 0, 0, Diso, 
-                                    tensor_elements[6]])))
+                                    tensor[6]])))
 
     # Compute the residuals
     residuals = data - y
@@ -495,7 +506,7 @@ def _nlls_err_func(tensor_elements, design_matrix, data, Diso=3e-3,
 
 
 def nlls_fit_tensor(design_matrix, data, fw_params=None, Diso=3e-3,
-                    weighting=None, sigma=None, jac=False):
+                    weighting=None, sigma=None, cholesky=False):
     """
     Fit the water elimination tensor model using the non-linear least-squares.
 
@@ -523,19 +534,21 @@ def nlls_fit_tensor(design_matrix, data, fw_params=None, Diso=3e-3,
         $mm^{2}.s^{-1}$. Please ajust this value if you are assuming different
         units of diffusion.
 
-    weighting: str
+    weighting: str, optional
            the weighting scheme to use in considering the
            squared-error. Default behavior is to use uniform weighting. Other
            options: 'sigma' 'gmm'
 
-    sigma: float
+    sigma: float, optional
         If the 'sigma' weighting scheme is used, a value of sigma needs to be
         provided here. According to [Chang2005]_, a good value to use is
         1.5267 * std(background_noise), where background_noise is estimated
         from some part of the image known to contain no signal (only noise).
 
-    jac : bool
-        Use the Jacobian? Default: False
+    cholesky : bool, optional 
+        If true it uses cholesky decomposition to insure that diffusion tensor
+        is positive define.
+        Default: True
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


etpeterson:
Looks like you're following the structure from DKI, right? That seems like a good way to start to me. I'll keep checking back to see if I can help out.

etpeterson:
Fortunately or unfortunately I'm getting some pressure to get working on this. @RafaelNH if you're otherwise occupied I'll forge on ahead and try to create my own fitting routine to merge into here.

RafaelNH:
Hi @etpeterson! I have to confess that currently I cannot work on this full time, so feel free to carry with the work done here. Actually, github is a perfect platform for developers to work simultaneously in the same scripts. Just keep me updated with your progress to avoid duplicated work. We also have the advantage of being in different time zones, thus if the end of the day you upload your work, I will start my day reviewing your changes and carrying on from what you did.

RafaelNH:
PS: I will be working on this in the following 2 hours.

arokem:
I'll just add for the benefit of both of you that you can also use our
gitter channel to coordinate: https://gitter.im/nipy/dipy

I've found it useful for coordinating work with others when working
simultaneously on the same thing. And I'd love to stay in the loop on this
work, so it will also allow me to see what's going on :-)

RafaelNH:
I am getting close to a final version of the WLLS solution of the water elimination model :). It runs without problem, however its outputs are not what I would expect (the new added test is causing Travis to fail).

Anyway, please let me know if you have suggestions on what was done so far or if you detect the problem that is making the test fail. Also coding contributions are welcome. Above, are the parts of the module that still have to be done:

- [] Find bug in WLS that is causing fwdti to fail
- [] Define the properties of the class FreeWaterTensorFit (that is easy)
- [] Add the models prediction functions (analogous to the dti and dki modules)
- [] Implementing the NLS part of the articles procedure
- [] Finding the Jacobian for the NLS fit

etpeterson:
Nice work. I copied yours and marked it up with changes and added the prediction function. Until I figure out how to contribute like a normal person I put it in my fork (https://github.com/etpeterson/dipy/blob/fwdti/dipy/reconst/fwdti.py).

I'm not sure I totally followed the steps so some comments may be off base, but I do think the piterations loop is zooming in to the wrong range of f-values.

RafaelNH:
Hi @etpeterson, thanks for your input :). Give a look to Dipy's documentation on how you can contribute with code http://nipy.org/dipy/devel/gitwash/index.html (@arokem, @Garyfallidis - do you have some recommendations of further reading for Eric?). You even can create a pull request of my branch. In this way the revision of your changes will be much more easy. Also it gives me the possibility of commenting code lines using tools of the github website (the comments directly to python should be only the info to stay with the final version of the files for guiding future users). Learning the basics of github is also important because is the only way for formally adding your contributions to Dipy's developing history.

Anyway, I give a look on that you did. I didn't include the function apparent_diffusion_coef - it is a duplicate of function in dti (users can call it from there). The prediction function seems to be correct, however we save to test in using nosetest (see test_fwdti,py for what I tested so far). Regarding to your in script comments, I will answer them here using Github website tools.

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,412 @@
...
+
+            # Estimate tissue's tensor from inv(A.T*S2*A)*A.T*S2*y
+            S2 = np.diag(np.square(np.dot(W, params)))
+            WS2 = np.dot(W.T, S2)
+            invWS2W = np.linalg.pinv(np.dot(WS2, W))
+            all_new_params = np.dot(np.dot(invWS2W, WS2), y)
+
+            # compute F2
+            S0r = np.exp(-np.array([all_new_params[6],]*nvol))
+            SIpred = (1-FS)*np.exp(np.dot(W, all_new_params)) + FS*S0r*SFW.T
```
RafaelNH: @etpeterson - this is already correct. First right side equation term already is multiplying by S0 because last element of all_new_params is log(S0), i.e:

np.exp(np.dot(W, all_new_params)) = S0*np.exp(np.dot(W[:, :6], all_new_params[:6])) 

Second term FS*S0r*SFW.T needs S0 since the free water contribution computed in line 363 do not take into account S0.

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,412 @@
...
+
+    # DTI weighted linear least square solution
+    WTS2 = np.dot(W.T, S2)
+    inv_WT_S2_W = np.linalg.pinv(np.dot(WTS2, W))
+    WT_S2_LS = np.dot(WTS2, log_s)
+    params = np.dot(inv_WT_S2_W, WT_S2_LS)
+    
+    # General free-water signal contribution
+    fwsig = np.exp(np.dot(design_matrix, 
+                          np.array([Diso, 0, Diso, 0, 0, Diso, -np.log(1.)])))
```
RafaelNH: negative in np.log(1) since design matrix is defined as Bm = [Bxx, Bxy, Byy, Bxz, Byz, Bzz, -1] instead of Bm = [Bxx, Bxy, Byy, Bxz, Byz, Bzz, 1].

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,412 @@
...
+
+            # Select params for lower F2
+            Mind = np.argmin(F2)
+            params = all_new_params[:, Mind]
+        # Updated f
+        f = fs[Mind]
+        # refining precision
+        flow = f - df
+        fhig = f + df
+        ns = 19
```
RafaelNH: @etpeterson - yes, you are right, this part is different to the article. However, I did this change intentionally to insure that we cover the full f space. For example, assume that the ground truth f is 0.54 however f=0.6 have a local F2 value lower value than 0.5. If you did what the article describes, the second step of the procedure will be searching in the range of 0.55 to 0.65 which doesn't include 0.54.

etpeterson: Agreed, that is an issue with their method and your implementation is very nice. I'm wondering though if it's a practical issue to never evaluate the end points. According to the article, 18% of the measurements are <0.05 and >0.95 so it seems that it would be good to evaluate those to try to avoid getting stuck in a local minimum.

RafaelNH: Lets assume that the ground truth is f=0. In the first step f=0.1 will have the lower F2 value. In the next step the algorithm will sample f from 0.01 to 0.19 with lower F2 value in f=0.01. Following this, the algorithm in theory can converge to 0 with the defined precision. But do not worry, we can add these cases in the testing module. If in practice the algorithm doesn't converge we can add the extreme values, which is computing F2 for the DTI model (f=0) and for a model only having a the free isotropic term.

etpeterson:
Looks great @RafaelNH! I just got it running on some data I have and at first glance the results look reasonable and it runs just fine. I'll test it some more later and let you know.

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,435 @@
...
+        self.args = args
+        self.kwargs = kwargs
+        self.min_signal = self.kwargs.pop('min_signal', None)
+        if self.min_signal is not None and self.min_signal <= 0:
+            e_s = "The `min_signal` key-word argument needs to be strictly"
+            e_s += " positive."
+            raise ValueError(e_s)
+
+    def fit(self, data, mask=None):
+        """ Fit method of the free water elinimation DTI model class
```
samuelstjean: elinimation -> elimination

dipy/reconst/fwdti.py
samuelstjean: log(1) = 0, probable not what you want.

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,435 @@
...
+    fhig = 1  # higher f evaluated
+    ns = 9  # initial number of samples per iteration
+    nvol = len(sig)
+    for p in range(piterations):
+        df = df * 0.1
+        fs = np.linspace(flow+df, fhig-df, num=ns)  # sampling f
+        # repeat fw contribution for all the samples
+        SFW = np.array([fwsig,]*ns)
+        FS, SI = np.meshgrid(fs, sig)
+        for r in range(riterations):
```
samuelstjean: you could add break statements when stuff converges, no point in doing iterations that bring little precision

dipy/reconst/fwdti.py
RafaelNH: thanks!

dipy/reconst/fwdti.py
RafaelNH: @samuelstjean - Well, for this to converge to a given precision it will require always the same amount of steps, that is why I didn't use break statements. Users can just decrease the number of steps it they want to decrease the precision - see the function's documentation.

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,435 @@
...
+    Diso : float, optional
+        Value of the free water isotropic diffusion. Default is set to 3e-3
+        $mm^{2}.s^{-1}$. Please ajust this value if you are assuming different
+        units of diffusion.
+    piterations : inter, optional
+        Number of iterations used to refine the precision of f. Default is set
+        to 3 corresponding to a precision of 0.01.
+    riterations : inter, optional
+        Number of iteration repetitions with adapted S0. To insure that S0 is
+        taken as a model free parameter Each precision iteration is repeated
```
samuelstjean: missing a .

dipy/reconst/fwdti.py
samuelstjean: I don't really get it, but if you say so, it's only 2-3 iterations anyway after all.

Garyfallidis:
Hi @RafaelNH, this looks like going towards the correct direction. For merging this soon you need to increase your coverage and report some timings and images with full datasets. After this is done you will need to go ahead with a tutorial. This can be in a second PR.   

```
dipy/reconst/fwdti.py                                183     38     36      8    75%   123-127, 134-136, 153-156, 161, 171-172, 462-485, 544, 581-587, 120->128, 133->134, 148->153, 158->161, 167->171, 459->462, 539->542, 543->544
```

Garyfallidis:
Also, you may want to ask from @etpeterson to write or extend the tutorial about this and you can mentor him a bit. In that way he can learn how to contribute to Dipy. Sounds like a good plan @etpeterson ?

etpeterson:
That works for me @Garyfallidis

RafaelNH:
@Garyfallidis, @etpeterson - Ok! Added some more tests! The only lines that I am not covering is when lsq non-linear procedure does not converge. Anyone have any idea how to test this? Anyway, even if this lines are not working property, it should not be problematic. Since fw_params is already set to the initial guess parameters at the beginning of the function, these lines of code are redundant (perhaps we can even remove the non-covered lines, I was keeping them only to be more consistent to the restore function in dti.py).

I think the next step for me will be running this function in some real data and time its speed. After that I will be more than happy to help @etpeterson with a new PR for the tutorial =). How that sounds?

Garyfallidis:
Sounds awesome! Fly high! :)

RafaelNH:
I used the fwdti model to process a slice of the CENIR multi b-values dataset. It took around 90 secs to be processed. Thus, to process the full dataset without a mask it will take around 1.75 hours. Not bad for a first non-optimized version of the procedures and considering that the free water elimination requires a non-linear convergence step.

The [free water diffusion tensor derived measures](http://gsoc2015dipydki.blogspot.it/) are looking very good. As expected, near to the ventricles and parenchyma, FA values are larger and Diffusion is smaller (Fig1 upper panels) when compared to the values of standard DTI (Fig1 lower panels). There are some voxels with overestimated FA in the ventricles, however these are expected according to [Koay et al., 2014](http://www.ncbi.nlm.nih.gov/pubmed/16828568). Users can easily remove these overestimated values of FA by excluding voxels with high free water diffusion volume fractions (Fig.2, idea also mentioned by [Koay et al., 2014](http://www.ncbi.nlm.nih.gov/pubmed/16828568). 

I added the lines of codes that I used to time the free water elimination model and plot the diffusion measures into doc/examples/reconst_fwdti.py which follows a similar structure of the DKI usage example.  @etpeterson - I think you can create the example script in the next PR from this script.

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,608 @@
+#!/usr/bin/python
+""" Classes and functions for fitting tensors """
```
arokem: This top-level docstring seems to be a copy-paste residual

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,608 @@
...
+    $S(\theta, b) = S_0 * [(1-f) * e^{-b ADC} + f * e^{-b D_{iso}]$, where
+    $ADC = \theta Q \theta^T$, $\theta$ is a unit vector pointing at any
+    direction on the sphere for which a signal is to be predicted, $b$ is the b
+    value provided in the GradientTable input for that direction, $Q$ is the
+    quadratic form of the tensor determined by the input parameters, $f$ is the
+    free water diffusion compartment, $D_{iso}$ is the free water diffusivity
+    which is equal to $3 * 10^{-3} mm^{2}s^{-1} [1]_.
+    References
+    ----------
+    .. [1] Pasternak, O., Sochen, N., Gur, Y., Intrator, N., Assaf, Y., 2009.
```
arokem: I don't think this is the right reference here, considering this is not an implementation of the Pasternak algorithm.

dipy/reconst/tests/test_fwdti.py
```diff
@@ -0,0 +1,209 @@
...
+    # Test non-linear fit
+    fwdm = fwdti.FreeWaterTensorModel(gtab_2s, 'NLS')
+    fwefit = fwdm.fit(S_conta)
+    FAfwe = fwefit.fa
+    Ffwe = fwefit.f
+
+    assert_almost_equal(FAdti, FAfwe)
+    assert_almost_equal(Ffwe, gtf)
+    
+    # Test non-linear fit, when no first quess is given
```
arokem: Typo: "quess" => "guess"

dipy/reconst/tests/test_fwdti.py
```diff
@@ -0,0 +1,209 @@
...
+    S0fwe = fwefit.S0
+
+    assert_almost_equal(FAdti, FAfwe)
+    assert_almost_equal(Ffwe, gtf)
+    assert_almost_equal(S0fwe, 100)
+
+
+def test_fwdti_precision():
+    # Simulation when water contamination is added
+    gtf = 0.63416  #ground truth volume fraction
```
arokem: PEP8 (space between `#` and comment text)

dipy/reconst/tests/test_fwdti.py
```diff
@@ -0,0 +1,209 @@
...
+    fwefit = fwdm.fit(DWI)
+    Ffwe = fwefit.f
+
+    assert_array_almost_equal(Ffwe, GTF, decimal=2)
+    
+
+def test_fwdti_predictions():
+    # single voxel case
+    # test funtion
+    gtf = 0.50  #ground truth volume fraction
```
arokem: Same here (PEP8 comment)

dipy/reconst/tests/test_fwdti.py
```diff
@@ -0,0 +1,209 @@
...
+    fwdtiM = fwdti.FreeWaterTensorModel(gtab_2s, min_signal=1)
+    correct_mask = np.zeros((2, 2, 2))
+    correct_mask[0, :, :] = 1;
+    correct_mask = correct_mask > 0
+    fwdtiF = fwdtiM.fit(DWI, mask=correct_mask)
+    assert_array_almost_equal(fwdtiF.fa, FAref)
+    assert_array_almost_equal(fwdtiF.f, GTF)
+
+    # 4th error - if a sigma is selected by no value of sigma is given for
+    # in the non-linear approach to performe restore
```
arokem: Typo: "performe" => "perform"

In general, maybe: "... if weighting is set to 'sigma', but no value of sigma is given (in case RESTORE is used)."

dipy/reconst/tests/test_fwdti.py
```diff
@@ -0,0 +1,209 @@
...
+    fwdm = fwdti.FreeWaterTensorModel(gtab_2s, 'NLS', weighting='sigma')
+    assert_raises(ValueError, fwdm.fit, DWI)
+
+
+def test_fwdti_restore():
+    # Restore have to work well even in nonproblematic cases
```
arokem: => "RESTORE has to work well"

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,608 @@
...
+from dipy.core.ndindex import ndindex
+
+
+def fwdti_prediction(params, gtab, Diso=3.0e-3):
+    """
+    Predict a signal given the parameters of the free water diffusion tensor
+    model.
+    Parameters
+    ----------
+    Params : ndarray
```
arokem: => "params" (lower case)

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,608 @@
...
+from dipy.reconst.dki import _positive_evals
+
+from dipy.core.sphere import Sphere
+from .vec_val_sum import vec_val_vect
+from dipy.core.ndindex import ndindex
+
+
+def fwdti_prediction(params, gtab, Diso=3.0e-3):
+    """
+    Predict a signal given the parameters of the free water diffusion tensor
```
arokem: Description on the first line of the docstring should preferably be just one line.

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,608 @@
...
+    Params : ndarray
+        Model parameters. The last dimension should have the 12 tensor
+        parameters (3 eigenvalues, followed by the 3 corresponding
+        eigenvectors) the volume fraction of the free water compartment, and
+        the non diffusion-weighted signal S0
+    gtab : a GradientTable class instance
+        The gradient table for this prediction
+    Diso : float, optional
+        Value of the free water isotropic diffusion. Default is set to 3e-3
+        $mm^{2}.s^{-1}$. Please ajust this value if you are assuming different
```
arokem: Typo: "ajust" => "adjust"

dipy/reconst/fwdti.py
arokem: I see - is this supposed to be a reference for the fact that water diffusivity is set to 3e-3? I am sure there are other, earlier references for that.

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,608 @@
...
+    ----------
+    .. [1] Pasternak, O., Sochen, N., Gur, Y., Intrator, N., Assaf, Y., 2009.
+           Free water elimination and mapping from diffusion MRI. Magn. Reson.
+           Med. 62, 717-739. http://dx.doi.org/10.1002/mrm.22055.
+    """
+    evals = params[..., :3]
+    evecs = params[..., 3:-2].reshape(params.shape[:-1] + (3, 3))
+    f = params[..., 12]
+    S0 = params[..., 13]
+    qform = vec_val_vect(evecs, evals)
```
arokem: Unless a `step` parameter is defined, this will create similar memory problems to the ones discussed in #840 and resolved in #857.

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,608 @@
...
+        References
+        ----------
+        .. [1] Hoy, A.R., Koay, C.G., Kecskemeti, S.R., Alexander, A.L., 2014.
+               Optimization of a free water elimination two-compartmental model
+               for diffusion tensor imaging. NeuroImage 103, 323-333.
+               doi: 10.1016/j.neuroimage.2014.09.053
+        """
+        ReconstModel.__init__(self, gtab)
+
+        # Code and comments have still to be addapted
```
arokem: Is this comment still needed?

dipy/reconst/fwdti.py
arokem: Preferably, use a similar logic to the one used in #857, where step is defined at the level of the Model object, at the level that calls out to this function, not in this function.

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,608 @@
...
+            min_signal = _min_positive_signal(data)
+        else:
+            min_signal = self.min_signal
+
+        data_in_mask = np.maximum(data_in_mask, min_signal)
+        params_in_mask = self.fit_method(self.design_matrix, data_in_mask,
+                                         *self.args, **self.kwargs)
+
+        if mask is None:
+            out_shape = data.shape[:-1] + (-1, )
```
arokem: Is this the same as  `(data.shape[:-1], -1)`? I think that's easier to read.

arokem: Oh - strike that - I see why that's not it. Carry on.

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,608 @@
...
+        fwdti_params : ndarray
+            The last dimension should have 14 parameters: the 12 tensor
+            parameters (3 eigenvalues, followed by the 3 corresponding
+            eigenvectors), the volume fraction of the free water compartment
+            and the estimate of the non diffusion-weighted signal S0.
+        S0 : float or ndarray
+            The non diffusion-weighted signal in every voxel, or across all
+            voxels. Default: 1
+        """
+        return fwdti_prediction(fwdti_params, self.gtab)
```
arokem: Here's where I would use `step`

arokem:
OK - that's all I have time to read through thoroughly for now. One general comment: I would prefer it if S0 was not a model fit parameter. This is because in some situtations (e.g. in cross-validation between data sets), you would want to input S0 as an input to the Fit object. 

In DTI, we don't have S0 set as a model parameter. Can we follow a similar pattern here?

RafaelNH:
Hi @arokem! I tried to follow a similar pattern. I only decided to generalize the code when I noticed that the non-linear procedures described by Hoy et al., 2014 always assumed S0 as a free parameter. Also, I want to keep this general since there are some groups that are starting to suggest not using the non-diffusion signal S0 in diffusion modeling fitting to eliminate perfusion pseudo diffusion influences. 

As an alternative, we can set S0 as an optional variable.Iif S0 is given FreeWaterDiffusionModel will use it an assume this parameter as fixed. However, as default we could assume S0 as a models parameter. How that sounds?

This alternative has the disadvantage of requiring some refactoring, but if this is required I can start doing it.

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,713 @@
...
+    Returns
+    -------
+    nlls_params: for each voxel the eigen-values and eigen-vectors of the
+        tissue tensor and the volume fraction of the free water compartment.
+    """
+    # Flatten for the iteration over voxels:
+    flat_data = data.reshape((-1, data.shape[-1]))
+
+    # Use the WLS method parameters as the starting point if fw_params is None:
+    if fw_params==None:
```
etpeterson: I'm making some suggestions here. Mostly related to inputting an initial guess from a WLS fit if you run it separately.
```
if np.any(fw_params) is None:
    fw_params = wls_fit_tensor(design_matrix, flat_data,  Diso=Diso)
else: #ETP
    fw_params = fw_params.reshape((-1, 14)) 
```

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,713 @@
...
+            dt = lower_triangular_to_cholesky(dt)
+
+        # f transformation if requested
+        if f_transform:
+            f = np.arcsin(2*params[12] - 1) + np.pi/2
+        else:
+            f = params[12]
+
+        # Use the Levenberg-Marquardt algorithm wrapped in opt.leastsq
+        start_params = np.concatenate((dt, [-np.log(s0), f]), axis=0)
```
etpeterson: Copy the parameters so we have them later without needing to re-create the vector.
```
input_params = params  # ETP copy it in case we need it later
```

etpeterson:
@RafaelNH apologies, just got done with a big lab deadline. What exactly were you asking for my help with? Happy to help, I just wanted to make sure before moving forward. The way I understand it is I should create a PR of this for an example tutorial, is that right?

RafaelNH:
Hi @etpeterson! That is right, however it would be great to have first this pull request merged in the way that you can continue the example from reconst_fwdti.py (well another solution is to fetch this pull request, and submit a pull request to this pull request =P). I will work on the small suggestions that I haven't address yet ASAP. I noticed that no one give any comments regarding how I am constraining the f values and forcing DT to be positively defined - feedback on this will be great!

samuelstjean:
I have little time to check stuff because of classes, and my laptop 
screen refuse to work at more than a 90 degree angles. I can have a look 
next week maybe. I also feel ike I am the only one to check math 
sometimes, I leave coding style and efficiency to the others btw.

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,694 @@
...
+            # For completeness, we predict the mean S0 for the non-diffusion
+            # weighted measurements, which is our best guess:
+            pred_s[..., gtab.b0s_mask] = S0[v]
+            pred_sig[v] = pred_s
+
+    return pred_sig
+
+
+class FreeWaterTensorModel(ReconstModel):
+    """ Class for the Free Water Elimitation Diffusion Tensor Model """
```
samuelstjean: elimination

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,694 @@
...
+        4) The estimate of the non diffusion-weighted signal S0
+    References
+    ----------
+    .. [1] Chung, SW., Lu, Y., Henry, R.G., 2006. Comparison of bootstrap
+       approaches for estimation of uncertainties of DTI parameters.
+       NeuroImage 33, 531-541.
+    """
+    tol = 1e-6
+
+    # preparing data and initializing parametres
```
samuelstjean: parameters

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,694 @@
...
+    data = np.asarray(data)
+    data_flat = data.reshape((-1, data.shape[-1]))
+    fw_params = np.empty((len(data_flat), 14))
+
+    # inverting design matrix and defining minimun diffusion aloud
+    min_diffusivity = tol / -design_matrix.min()
+    inv_design = np.linalg.pinv(design_matrix)
+
+    # lopping WLS solution on all data voxels
```
samuelstjean: looping

doc/examples/reconst_fwdti.py
```diff
@@ -0,0 +1,174 @@
...
+bvals = [200, 400, 1000, 2000]
+
+img, gtab = read_cenir_multib(bvals)
+
+data = img.get_data()
+
+affine = img.get_affine()
+
+"""
+Before fitting the data, we preform some data pre-processing. We first compute
```
samuelstjean: perform

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,694 @@
...
+def fwdti_prediction(params, gtab, Diso=3.0e-3):
+    """ Signal prediction given the free water DTI model parameters.
+    Parameters
+    ----------
+    params : ndarray
+        Model parameters. The last dimension should have the 12 tensor
+        parameters (3 eigenvalues, followed by the 3 corresponding
+        eigenvectors) the volume fraction of the free water compartment, and
+        the non diffusion-weighted signal S0
+    gtab : a GradientTable class instance
```
samuelstjean: You should also check for at least 2 bvalues and just raise an error if the acquisition does not support FWE.

doc/examples/reconst_fwdti.py
```diff
@@ -0,0 +1,174 @@
...
+.. figure:: Diffusion_measures_from_free_water_DTI_and_standard_DTI.png
+   :align: center
+   **Diffusion tensor measures obtain from the diffusion tensor estimated from
+   the free water DTI (upper panels) and standard DTI (lower panels).**.
+
+From the figure, we can see that ...
+
+In addition to the standard diffusion statistics, the FreeWaterDiffusionFit
+instance can be used to display the volume fraction of the free water diffusion
+componet
```
samuelstjean: component, and missing a dot

samuelstjean:
So why the by hand cholesky decomposition? If you have negative eigen values because of badly behaved voxels, you will still have the negativity problem and it still won't work because of the square root.

etpeterson:
@RafaelNH correct me if I'm wrong here but as I understand it the cholesky decomposition will keep the eigenvalues positive because you're working in cholesky coordinates so to speak, and these will always result in positive eigenvalues. You do need to check the initial input though and change all negatives to zeros perhaps? From there though you're only doing the cholesky->lower triangular which won't have issues.

RafaelNH:
@etpeterson exactly! Regarding to the initial guess we don't have negative values, because this was avoided previously when decomposing diffusion elements to eigenvalues (note that the first guess input format requires the diffusion tensor in eigenvalue and eigenvector decomposition format)!

etpeterson:
Yeah, that seems right to me. The f_transform looks good to me too though I think that needs a bounds check at the beginning otherwise I think you risk getting nans. I'm wondering if this transform makes the fitting more difficult by transforming a linear mixing into a wave?

arokem:
Re: negative eigenvalues. To be sure, it might be good to throw at it in testing something really funky that should bring up some negative eigenvalues, and see that it's unaffected.

dipy/reconst/fwdti.py
RafaelNH: I think you got the wrong line - step should be used in the function of FreeWaterTensorFit class and not in the function of FreeWaterTensorModel, right?

samuelstjean:
So you recast the whole thing between 0 and 1 with the arcsin transform to ensure that f stays between 0 and 1? It does seem fishy honestly, I guess the bounded optimization was too slow with scipy? And for the negative eigen values, well, it does not enforce them currently, so if you start with a ols tensor fit, you might end up with negatives eigen values, which would break the thing.

At this stage, might be easier to ensure they are positive and just exit if not for this particular voxel, or not let the user use ols as a first fit method. Nothing in the wls also ensures that (though it might be a lower occurance rate because of said weighting), but clipping to zero will definitely kill your metrics (FA,MD,RD, etc.) and be avoided I'd say.

arokem:
On Sun, Mar 6, 2016 at 3:41 AM, Samuel St-Jean <notifications@github.com>
wrote:

> So you recast the whole thing between 0 and 1 with the arcsin transform to
> ensure that f stays between 0 and 1? It does seem fishy honestly, I guess
> the bounded optimization was too slow with scipy?
>
This is actually not too different from the way bounds were applied in this
software: https://github.com/jjhelmus/leastsqbound-scipy, but I'm not sure
whether that's what ended up in scipy.

> And for the negative eigen values, well, it does not enforce them
> currently, so if you start with a ols tensor fit, you might end up with
> negatives eigen values, which would break the thing.
>
> I believe non-negativity is enforced here:
https://github.com/nipy/dipy/blob/master/dipy/reconst/dti.py#L1813

samuelstjean:
They also say themselves what they do is bad and to use scipy 0.17 now, might be worth a look if it is simple. 

For the non-negativity, it seems to be used after the cholesky decomposition, but it would blow up before that with negative eigen values. This is a bit hard to read (for me at least) though with all the object jumping around, but if you say that feeding it explicitly negative eigen values works and does not destroy tensor derived metrics (which are probably unreliable in the first case actually), then good.

dipy/reconst/fwdti.py
arokem: Yes. That's correct.

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,718 @@
...
+            df = df * 0.1
+            fs = np.linspace(flow+df, fhig-df, num=ns)  # sampling f
+            SFW = np.array([fwsig, ]*ns)  # repeat contributions for all values
+            FS, SI = np.meshgrid(fs, sig)
+            SA = SI - FS*S0*SFW.T
+            # SA < 0 means that the signal components from the free water
+            # component is larger than the total fiber. This cases are present
+            # for inapropriate large volume fractions (given the current S0
+            # value estimated). To avoid the log of negative values:
+            SA[SA <= 0] = 0.0001  # same min signal assumed in dti.py
```
arokem: Might be worth setting this as `min_signal` somewhere higher up, so that it's easy to find.

RafaelNH:
Hi @arokem! Thanks for the reference - I didn't know for example that I could edit commits using a text editor as VIM - sometimes I am still so naive on github. I will be more careful with my comments in future.

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,871 @@
...
+class FreeWaterTensorModel(ReconstModel):
+    """ Class for the Free Water Elimination Diffusion Tensor Model """
+    def __init__(self, gtab, fit_method="NLS", *args, **kwargs):
+        """ Free Water Diffusion Tensor Model [1]_.
+
+        Parameters
+        ----------
+        gtab : GradientTable class instance
+        fit_method : str or callable
+            str can be one of the following:
```
Garyfallidis: Hi @RafaelNH have you checked if this docstring is rendered correctly in the website?

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,871 @@
...
+                              lower_triangular, mean_diffusivity)
+from dipy.reconst.dti import wls_fit_tensor as dti_wls_fit
+from dipy.reconst.dki import _positive_evals
+
+from dipy.reconst.vec_val_sum import vec_val_vect
+from dipy.core.ndindex import ndindex
+
+
+def fwdti_prediction(params, gtab, S0=1, Diso=3.0e-3):
+    """ Signal prediction given the free water DTI model parameters.
```
Garyfallidis: If you are planning to have math in the docstring add r infront of the docstring. 
```
r"""
```

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,871 @@
...
+        Parameters
+        ----------
+        data : array
+            The measured signal from one voxel.
+        mask : array
+            A boolean array used to mark the coordinates in the data that
+            should be analyzed that has the shape data.shape[:-1]
+        """
+        if mask is None:
+            # Flatten it to 2D either way:
```
Garyfallidis: I would remove ```either way```.

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,871 @@
...
+        size = np.prod(shape)
+        if step is None:
+            step = self.model.kwargs.get('step', size)
+        if step >= size:
+            return fwdti_prediction(self.model_params, gtab, S0=S0)
+        params = np.reshape(self.model_params,
+                            (-1, self.model_params.shape[-1]))
+        predict = np.empty((size, gtab.bvals.shape[0]))
+        for i in range(0, size, step):
+            predict[i:i+step] = fwdti_prediction(params[i:i+step], gtab, S0=S0)
```
Garyfallidis: pep8

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,871 @@
...
+        return predict.reshape(shape + (gtab.bvals.shape[0], ))
```
Garyfallidis: remove space after comma

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,871 @@
...
+            # component is larger than the total fiber. This cases are present
+            # for inapropriate large volume fractions (given the current S0
+            # value estimated). To overcome this issue negative SA are replaced
+            # by data's min positive signal. 
+            SA[SA <= 0] = min_signal
+            y = np.log(SA / (1-FS))
+            all_new_params = np.dot(invWTS2W_WTS2, y)
+
+            # Select params for lower F2
+            SIpred = (1-FS)*np.exp(np.dot(W, all_new_params)) + FS*S0*SFW.T
```
Garyfallidis: Pep8!!! correct everywhere please. Use an editor that shows you at which lines the pep errors appear. E.g. Spyder and atom have this possibility.

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,871 @@
...
+        np.mean(data_flat[cond, :] / \
+                np.exp(np.dot(design_matrix[..., :6],
+                              np.array([Diso, 0, Diso, 0, 0, Diso]))),
+                -1)  # Only free water
+    S0out = S0out.reshape(data.shape[:-1])
+    return params_out, S0out
+
+
+def lower_triangular_to_cholesky(tensor_elements):
+    """ Perfoms Cholesky decompostion of the diffusion tensor
```
Garyfallidis: typo

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,871 @@
...
+
+    Parameters
+    ----------
+    tensor_elements : array (6,)
+        Array containing the six elements of diffusion tensor's lower
+        triangular.
```
Garyfallidis: space

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,871 @@
...
+    Returns
+    -------
+    cholesky_elements : array (6,)
+        Array containing the six Cholesky's decomposition elements
+        (R0, R1, R2, R3, R4, R5) [1]_.
+    References
```
Garyfallidis: space

doc/examples/reconst_fwdti.py
```diff
@@ -0,0 +1,174 @@
+"""
+=====================================================================
+Title
```
Garyfallidis: ?

arokem: Yeah. This should probably be labeled as "WIP", huh? :-)

It still needs to be rebased on top of https://github.com/nipy/dipy/pull/1036, which still needs to be merged, so I imagine this will change quite a bit.

Garyfallidis: We also suggested that @etpeterson would help @RafaelNH in writing this tutorial so he can be accustomed to the DIPY development processes. I hope this is still on the horizon. Let me know otherwise.

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,871 @@
...
+    S0out[cond] = np.mean(data_flat[cond, :] / 
+                          np.exp(np.dot(design_matrix[..., :6],
+                                        np.array([Diso, 0, Diso, 0, 0, Diso]))),
+                          -1)  # Only free water
+    S0out = S0out.reshape(data.shape[:-1])
+    return params_out, S0out
+
+
+def _nls_err_func(tensor_elements, design_matrix, data, Diso=3e-3,
+                   weighting=None, sigma=None, cholesky=False,
```
arokem: Indentation doesn't match previous line (and in next line as well)

dipy/reconst/fwdti.py
RafaelNH: Hi @Garyfallidis! I think this is already in Pep8 - you can and should remove spaces between operators so that computation priorities are easier to read. For example if you have e * (a + b) + (c + d) you can write e*(a+b) + (c+b). I am already using spyder pep error warnings :)

dipy/reconst/fwdti.py
```diff
@@ -140,7 +140,7 @@ def fit(self, data, mask=None):
             should be analyzed that has the shape data.shape[:-1]
         """
         if mask is None:
-            # Flatten it to 2D either way:
+            # Flatten mask to 2D:
```
arokem: I believe you are flattening the data, not the mask

dipy/reconst/fwdti.py
arokem: It's very similar to the docstring of the `TensorModel.__init__`, which *almost* renders fine. One more line-break here should do it, I believe.

arokem: https://github.com/nipy/dipy/pull/1080

doc/examples/reconst_fwdti.py
etpeterson: Sorry, this got lost down my inbox and discovered this getting back to thinking about this project again. Yes it's still on the horizon, hopefully in a week or two.

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,876 @@
...
+        self.design_matrix = design_matrix(self.gtab)
+        self.args = args
+        self.kwargs = kwargs
+        self.min_signal = self.kwargs.pop('min_signal', None)
+        if self.min_signal is not None and self.min_signal <= 0:
+            e_s = "The `min_signal` key-word argument needs to be strictly"
+            e_s += " positive."
+            raise ValueError(e_s)
+
+    def fit(self, data, mask=None):
```
arokem: Use the multivoxel decorator! See the IVIM PR.

arokem: Using the multivoxel decorator is probably still a good idea. It will significantly simplify this code.

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,876 @@
...
+
+        Parameters
+        ----------
+        fwdti_params : ndarray
+            The last dimension should have 13 parameters: the 12 tensor
+            parameters (3 eigenvalues, followed by the 3 corresponding
+            eigenvectors) and the free water volume fraction.
+        S0 : float or ndarray
+            The non diffusion-weighted signal in every voxel, or across all
+            voxels. Default: 1
```
arokem: A "Returns" section?

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,876 @@
...
+            return fwdti_prediction(self.model_params, gtab, S0=S0)
+        params = np.reshape(self.model_params,
+                            (-1, self.model_params.shape[-1]))
+        predict = np.empty((size, gtab.bvals.shape[0]))
+        for i in range(0, size, step):
+            predict[i:i+step] = fwdti_prediction(params[i:i+step], gtab, S0=S0)
+        return predict.reshape(shape + (gtab.bvals.shape[0],))
+
+
+def _wls_iter(design_matrix, sig, min_diffusivity, min_signal, Diso=3e-3,
```
arokem: I wonder whether this is the correct name for this function. Is it really doing a WLS iteration?

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,876 @@
...
+            # Select params for lower F2
+            S0r = np.exp(-np.array([all_new_params[6], ]*nvol))
+            SIpred = (1-FS)*np.exp(np.dot(W, all_new_params)) + FS*S0r*SFW.T
+            F2 = np.sum(np.square(SI - SIpred), axis=0)
+            Mind = np.argmin(F2)
+            params = all_new_params[:, Mind]
+            f = fs[Mind]  # Updated f
+            flow = f - df  # refining precision
+            fhig = f + df
+            ns = 19
```
arokem: Looks like there's a lot of code duplicated here (because of the `if`.... `else`). Maybe implement the internals as a helper function?

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,876 @@
...
+    md = mean_diffusivity(dtiparams[..., :3])
+    cond = md > mdreg  # removal condition
+    data_cond = data_flat[~cond, :]
+
+    # Initializing fw_params according to selected initial guess
+    if np.any(fw_params) is None:
+        if np.any(S0) is None:
+            # If parameters and S0 are not given they both are estimated here
+            params_out, S0 = wls_fit_tensor(design_matrix, data_flat,
+                                            Diso=Diso)
```
arokem: Test coverage is rather impressive already (at 98%), but this line is not being tested for some reason.

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,876 @@
...
+            number. Default: minimal signal in the data provided to `fit`.
+
+        References
+        ----------
+        .. [1] Hoy, A.R., Koay, C.G., Kecskemeti, S.R., Alexander, A.L., 2014.
+               Optimization of a free water elimination two-compartmental model
+               for diffusion tensor imaging. NeuroImage 103, 323-333.
+               doi: 10.1016/j.neuroimage.2014.09.053
+        """
+        ReconstModel.__init__(self, gtab)
```
arokem: Strangely, `coverage` is reporting that the lines 113-121 are not being tested. That must be an error with `coverage` right?

arokem: I think I'm going to stop commenting on test coverage now...

doc/examples/reconst_fwdti.py
```diff
@@ -0,0 +1,185 @@
+"""
+==========================================================================
+Using the free water elimination model to remove free water contamination
+==========================================================================
+
+As shown previously (see :ref:`example_reconst_dti`), the diffusion tensor
+model is a simple way to characterize the diffusion anisotropy. However, this
```
arokem: "characterize the diffusion anisotropy" => "characterize diffusion anisotropy" or "characterize the anisotropy of diffusion"

doc/examples/reconst_fwdti.py
```diff
@@ -0,0 +1,185 @@
...
+model is not specific to particular type of tissue. For example, diffusion
```
arokem: "to particular type of tissue" => "to a particular type of tissue" or "to particular types of tissue". 

Also, this fragment is not so clear, so I think that you can jump straight to the next clause: 

    However, in regions near the cerebral ventricles and parenchyma ...

doc/examples/reconst_fwdti.py
```diff
@@ -0,0 +1,185 @@
...
+A way to remove this free water influences is to expand the DTI model to take
+into account an extra compartment representing the contributions of free water
+diffusion. The expression of the expanded DTI model is shown below:
+
+.. math::
+
+    S(\mathbf{g}, b) = S_0(1-f)e^{-b\mathbf{g}^T \mathbf{D}
+    \mathbf{g}}+S_0fe^{-b D_{iso}}
+
+where $\mathbf{g}$ and $b$ are diffusion grandient direction and weighted
```
arokem: "grandient" => "gradient"

arokem: "weighted" => "diffusion weighting"

doc/examples/reconst_fwdti.py
```diff
@@ -0,0 +1,185 @@
...
+(more information see :ref:`example_reconst_dti`), $S(\mathbf{g}, b)$ is the
+diffusion-weighted signal measured, $S_0$ is the signal conducted in a
```
arokem: I would omit the word "conducted"

doc/examples/reconst_fwdti.py
```diff
@@ -0,0 +1,185 @@
...
+measurement with no diffusion weighting, $\mathbf{D}$ is the diffusion tensor,
+$f$ the volume fraction of the free water component, and $D_iso$ is the
+isotropic value of the free water diffusion (normally set to $3.0 \times
+10^{-3} mm^{2}s^{-1}$).
+
+In this example, we show how to process a diffusion weighted dataset using the
+free water elimantion.
```
arokem: "free water elimantion" => "free water elimination model"

doc/examples/reconst_fwdti.py
```diff
@@ -0,0 +1,185 @@
...
+Without spatial constrains the free water elimination model cannot be solved
+in data acquired from one non-zero b-value _[Hoy2014]. Therefore, here we
+download a dataset that was required from multi b-values.
+"""
+
+fetch_cenir_multib(with_raw=False)
+
+"""
+From the downloaded data, we read only the data acquired with b-values up to
+2000 $s.mm^{-2} to decrease the influence of non-Gaussain _[Hoy2014].
```
arokem: "non-Gaussain" => "non-Gaussian"

Also, non-Gaussian what? This sentence seems to cut off mid-way

doc/examples/reconst_fwdti.py
```diff
@@ -0,0 +1,185 @@
...
+data.
+"""
+
+axial_slice = 40
+
+mask_roi = np.zeros(data.shape[:-1], dtype=bool)
+mask_roi[:, :, axial_slice] = mask[:, :, axial_slice]
+
+"""
+The free water elimantion model fit can then be initialized by instantiating
```
arokem: Typo: "elimantion"

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,876 @@
...
+            str can be one of the following:
+
+            'WLS' for weighted linear least square fit according to [1]_
+                :func:`fwdti.wls_fit_tensor`
+            'NLS' for non-linear least square fit according to [1]_
+                :func:`fwdti.nls_fit_tensor`
+
+            callable has to have the signature:
+              fit_method(design_matrix, data, *args, **kwargs)
+        args, kwargs : arguments and key-word arguments passed to the
```
arokem: Maybe we can set a default value for the `cholesky` input?

doc/examples/reconst_fwdti.py
```diff
@@ -0,0 +1,185 @@
...
+a FreeWaterTensorModel class object:
+"""
+
+fwdtimodel = fwdti.FreeWaterTensorModel(gtab, 'NLS', cholesky=False)
```
arokem: Maybe you want to explain what "NLS" stands for here and what `cholesky=False` means? Also, if these are the default values (I see that "NLS" is), we might not want to set them in the example, and let users explore these options on their own.

doc/examples/reconst_fwdti.py
```diff
@@ -0,0 +1,185 @@
...
+dtimodel = dti.TensorModel(gtab)
+
+dtifit = dtimodel.fit(data, mask=mask_roi)
+
+dti_FA = dtifit.fa
+dti_MD = dtifit.md
+
+"""
+Below the FA values for both free water elimnantion DTI model and standard DTI
+model are ploted in panels A and B, while the repective MD values are ploted in
```
arokem: Typo: "ploted" => "plotted"

arokem:
Just ran the example. It's nice and quick. 

I think that it's almost ready to merge, but would love to have feedback from others.

doc/examples/reconst_fwdti.py
samuelstjean: I just gave a quick run to the example, and since you pass all the keywords, changing it to wls unfortunately crashed the function as cholesky is not an option for this one.

Creating the object works, but fitting gives 
~~~python
In [19]: fwdtifit = fwdtimodel.fit(data, mask=mask_roi)
TypeError                                 Traceback (most recent call last)
<ipython-input-19-3dbd1ad0d42c> in <module>()
----> 1 fwdtifit = fwdtimodel.fit(data, mask=mask_roi)

/home/samuel/python/dipy/reconst/fwdti.py in fit(self, data, mask)
    169         params_in_mask, S0_im = self.fit_method(self.design_matrix,
    170                                                 data_in_mask,
--> 171                                                 *self.args, **self.kwargs)
    172 
    173         if mask is None:

TypeError: wls_fit_tensor() got an unexpected keyword argument 'cholesky'
~~~

samuelstjean:
While at it, there seems to be some problems near partial volume voxels were the FA is degenerated/super high. 

figure_1

Near the ventricules for example at the top,  some voxels now have higher FA inside the ventricules. So, if you do a tracking mask on based FA threshold, you might end up tracking csf by mistake. 

Do you think that these voxels are actually boundary voxels which are now valid to use since the free water contamination was removed or something might have gone the wrong way in the optimisation, due in part to the probably high water level fraction? I am mostly asking as it could be intended for voxels near the csf as we seems to see some structure, but it could also be due to numerical issues, there was a ismrm toronto oral presentation about that and similar related issues in nls solving for example.

RafaelNH:
Hi @samuelstjean! Unfortunately, this FA overestimation in the boundary regions are an expected issue of this model (and nothing due to its implementation). You can see the same artifact on the images of the original article (see the articles Fig. 7). This is also expected from simulations since the fwDTI model was shown to produce overestimated FA values for simulated voxels with high water volume fraction (see original article Fig. 4). As the original article suggest, a way to deal with this problem is "... dismissing all diffusion metrics from voxels with an f-value set above a reasonable threshold such as 0.75." Parhaps, it will be a good idea to do this also in the usage example! What do you think?

samuelstjean:
Sounds a great idea to show limitation and where it is not applicable (so people don't do it/ask questions...), as if you did not play with these things directly like myself you would not really know about that. 

Reminds me of the tons of emails I had in the early days of piesno since I completely forgot to mention that it is made to be used on 2D slices and not 3D volume, or that it won't work on motion corrected data. Seemed obvious to me at the time indeed, but nobody else read the paper, so nobody else had the idea it would not be applicable in these situations (but it does work well if you follow the assumptions at least). 

Well anyway, if you would add a line or two about this effect and to discard highly water contaminated voxels, might save you some explanation time down in a few months.

RafaelNH:
Indeed, thanks for pointing this out :+1:

etpeterson:
@RafaelNH Thanks for holding my feet to the fire. It's on my list but probably won't happen until next week some time.

RafaelNH:
@arokem! I just review the multi-voxel decorator code. Are you sure that we should use this for the case of the free water DTI? I am not sure if this will simplify the code because here I am allowing users to select different free water DTI elimination fit procedures, each with different inputs. By using the multi-voxel decorator, I will have to specify the different inputs of all the different fits as self properties, which will make the code messier. Also we will have indeed problems if we expand this work by incorporating spatial regularization constrains.

Please let me know what do you think. If we decide to not use the multi-voxel decorator, I just need to do the last edits on the usage example and this will be ready to be merged.

arokem:
I think that these properties are already properties of self. The `fit`
function only accepts mask as an input (which is exactly as it should be!).
Everything seems to be set in the model's `__init__` (again, as it should
be!).

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,863 @@
...
+        shape = self.model_params.shape[:-1]
+        size = np.prod(shape)
+        if step is None:
+            step = self.model.kwargs.get('step', size)
+        if step >= size:
+            return fwdti_prediction(self.model_params, gtab, S0=S0)
+        params = np.reshape(self.model_params,
+                            (-1, self.model_params.shape[-1]))
+        predict = np.empty((size, gtab.bvals.shape[0]))
+        for i in range(0, size, step):
```
arokem: I don't think that using "step" makes sense here. You're going to do this voxel by voxel any way, right?

dipy/reconst/fwdti.py
```diff
@@ -0,0 +1,889 @@
...
+    lower_dt = lower_triangular(qform, S0)
+    lower_diso = lower_dt.copy()
+    lower_diso[..., 0] = lower_diso[..., 2] = lower_diso[..., 5] = Diso
+    lower_diso[..., 1] = lower_diso[..., 3] = lower_diso[..., 4] = 0
+    D = design_matrix(gtab)
+
+    pred_sig = np.zeros(f.shape + (gtab.bvals.shape[0],))
+    mask = _positive_evals(evals[..., 0], evals[..., 1], evals[..., 2])
+    index = ndindex(f.shape)
+    for v in index:
```
arokem: Would it be faster to loop over the indices in `np.where(mask)` instead?

doc/examples/reconst_fwdti.py
```diff
@@ -0,0 +1,228 @@
...
+import dipy.reconst.dti as dti
+import matplotlib.pyplot as plt
+from dipy.data import fetch_cenir_multib
+from dipy.data import read_cenir_multib
+from dipy.segment.mask import median_otsu
+
+"""
+Without spatial constrains the free water elimination model cannot be solved
+in data acquired from one non-zero b-value _[Hoy2014]. Therefore, here we
+download a dataset that was required from multi b-values.
```
Garyfallidis: from many b-values

Garyfallidis:
This looks very close to be merged. I want to try it with a different dataset too and then will merge  (see also comments above). Any other final comments from the rest?

arokem:
Yup: +1 from me - ready to go as soon as you've had a chance to check things out. @RafaelNH : is there anything we are forgetting, or is this ready to go in from your point-of-view?

RafaelNH:
Hey guys! Yes for my point of view this is ready to be merge! Please let me know if there are any comments that I forgot to address!

Please note that I did some refactoring to use the multi voxel decorator that force me to change some of the features. For example, now when using the non-linear approach this will always use the wls as the first parameters guess. Also S0 is always computed from the mean of the b0 - before I was allowing users to give their own b0 estimate or take it as a model parameter. Because of the multi voxel decorator this is not possible anymore, however I think this is not an issue because the S0 from the mean of the b0 images showed to provide the more robust fits. In general the code is more simplified and easier to read.

In addition, I add the f value correction step in the example (the aspect that I discussed above with @samuelstjean).

@arokem and @Garyfallidis after testing this please feel free to merge this =).

Garyfallidis:
Good job!

pppangpang:
Dear,
I am using the "dipy.reconst.fwdti" for the calculation of free-water elimination, I want to confirm that the dMRI should be multi-shell data (at least two non-zero b-values), or single-shell (e.g., b=0 and b=1000) is fine as from the paper https://pubmed.ncbi.nlm.nih.gov/19623619/ 

Thanks!

arokem:
Hello @pppangpang : the implementation in DIPY currently only works for multi-shell data. We do have an implementation of FWDTI with single shell data, based on Marc Golub's work, here: https://github.com/nrdg/fwe. I am currently working to port this to DIPY as well, so it should also be in DIPY soon.

RafaelNH:
To add to Ariel’s comment: there is a reason why I have not been pushing the single-shell Free-Water DTI implementation into DIPY. As shown in Marc Golub’s work ([https://pubmed.ncbi.nlm.nih.gov/33270935/](https://pubmed.ncbi.nlm.nih.gov/33270935/)), the single-shell fwDTI model is inherently degenerate. It appears to produce plausible results only because it fixes the tissue mean diffusivity to a constant value. Any deviation from that fixed MD is then interpreted as a change in free water, which can be misleading.

We also demonstrated this in a follow-up study ([https://pmc.ncbi.nlm.nih.gov/articles/PMC12272235/](https://pmc.ncbi.nlm.nih.gov/articles/PMC12272235/), where we showed that this constraint led to incorrect conclusions in ageing data, including apparent increases in tissue FA driven solely by the fixed-MD assumption.

Despite this, I am glad that @arokem is working on porting the single-shell code to DIPY. However, I would ask that these limitations be clearly stated in the documentation. Free-water changes estimated with single-shell fwDTI can only be considered genuine if there is no reason to suspect alterations in tissue mean diffusivity. Users should be encouraged to explicitly acknowledge this limitation to avoid misleading interpretations.

arokem:
Yes, I will include these caveats. As you can see in [our recent paper](https://direct.mit.edu/imag/article/doi/10.1162/IMAG.a.991/133658/Free-water-elimination-tractometry-for-aging), we did find that this approach is helpful in some aspects of tractometry (though to a much lesser degree than the non-degenerate multi-shell approach, if that is possible!), so there are legitimate uses of this model that do not fall into this particular pitfall. Nevertheless, I agree that a stern warning in the documentation about interpreting the FWF as though it is different from MD (e.g., from DTI) would be very appropriate. Your work on this has been very helpful, and I think that the field could use more empirical evaluations of these issues to hammer that message home. Hopefully the availability of the code will help more people make these evaluations. But I am not fooling myself and people will probably also rush to use this in a naive and misguided way.

EDIT: fixed a typo.

PR #849, opened 2016-01-29
nlmeans use num threads option in 3d

samuelstjean:
So it was always using everything instead of the specified number, so now it's on par with the 4D version.


arokem:
Is there any way to test this, so that that this bug doesn't come creeping back?

arokem:
Speaking of which, any thoughts on #832? Did you just abandon that one?

samuelstjean:
Half, fixed it on my side, thought we should get the float 32 or 64
tracking discussion going before. Not sure another test can be done, like
with the for loop in the recursive response, this is just a keyword that
got forgotten, can we even test this stuff? It's not like it was throwing
errors or wonky values.

arokem:
Yeah. I am not sure how to test this either. I think we can probably just
merge it without further ado. Anyone have any suggestions on this?

Garyfallidis:
All good! Thanks!

Issue #831, opened 2016-01-11
ACT tracking example gives weird streamlines


samuelstjean:
So I was happily running the example from http://nipy.org/dipy/examples_built/tracking_tissue_classifier.html#example-tracking-tissue-classifier. 
The last part about the ACT should only give streamlines going from GM to GM, but if you zoom on the results you get small nonsense streamlines. I also saw the same thing in another viewer, so it's probably not a bug of visualisation. So did I understand wrong what the ACT is doing, is the given segmentation in the example just too rough or is there a weird underlying bug?

![screenshot from 2016-01-11 12 21 37](https://cloud.githubusercontent.com/assets/3030760/12232465/19403fe0-b85f-11e5-8c53-1ea82244984b.png)

Just run block 1 from the example, then the last block about ACT with only valid streamlines and zoom in the window, you should be able to spot these fibers hopefully.

samuelstjean:
I used Gab's tracking script with the same maps and tracking parameters, and it does not produce these streamlines. If I ask it to return all the streamlines, there are some which do not make sense like in the picture above, so there may be a problem with the return_all=True option of the tracking.

samuelstjean:
More pictures :

From the example 

![screenshot](https://cloud.githubusercontent.com/assets/3030760/12649670/de7b4150-c5df-11e5-9cc4-9f6c5613626f.png)

From the other ACT implementation, which does not have the small fibers

![screenshot_gab](https://cloud.githubusercontent.com/assets/3030760/12649675/e76ec296-c5df-11e5-9d4b-0c102a8f1272.png)

arokem:
@gabknight - any ideas?

gabknight:
hey guys,

I just look thought it and, as far as I can see, everything behave as expected. The problem with those short streamlines is that the seed position is right in voxels with partial volume fraction of gm > 0.5 (see figure). So what we see is a line formed with the seed position and one step in forward and backward directions, both stopped in valid ending tissue (gm). 
I will make a PR updating the example's seed mask, taking into account pves.

![image](https://cloud.githubusercontent.com/assets/702455/12682402/3763379a-c681-11e5-958a-5624a1bfef69.png)

samuelstjean:
If I use your script with act, no pft, same masks, it doesn't produce those
small streamlines. Also tried 10 seeds/voxels just in case, still works. I
did not check your act implementation and assumed it does the same as the
dipy one though.

hey guys,

[...]

PR #851, opened 2016-01-29
Tissue Classifier tracking example - changed seeding mask to wm only voxels

gabknight:
Following discussion in issue #831, this PR changes the seeding mask to voxel with 100% wm partial volume fraction. Seeds placed in voxel with a partial volume fraction of gm > 0.5 created unwanted short streamlines.

Thanks @samuelstjean, @arokem

### DIFF
--- a/doc/examples/tracking_tissue_classifier.py
+++ b/doc/examples/tracking_tissue_classifier.py
@@ -18,7 +18,7 @@ or only the 'valid' ones.
 
 This example is an extension of the
 :ref:`example_deterministic_fiber_tracking` example. We begin by loading the
-data, creating a seeding mask of the white matter voxels of the corpus callosum,
+data, creating a seeding mask from white matter voxels of the corpus callosum,
 fitting a Constrained Spherical Deconvolution (CSD) reconstruction
 model and creating the maximum deterministic direction getter.
 """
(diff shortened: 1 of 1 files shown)


arokem:
Thanks @gabknight - for the quick turnaround! @samuelstjean - please confirm that this resolves the issue you reported. And thanks for noticing that!

samuelstjean:
There is some other weird thing I wanted to look at, and completely forgot what it was. Will come back to me in a few days hopefully.

arokem:
In the meanwhile, could you please confirm that this does resolve your previous complaint?

samuelstjean:
Seems good regarding the small fibers, lots of green fibers are gone though. 

screenshot from 2016-02-03 14 33 25

All fibers
screenshot from 2016-02-03 14 36 20


As gab suggested in an internal discussion, adding a length threshold to the tracking api would also help, someday. So it does fix the immediate problem, but as some stuff is gone, anyone who knows anatomy wants to pitch in?

gabknight:
The partial volume fraction maps are not perfect (e.g. there is false gm between the ventricles and the CC), which explains the stopped green streamlines on the first figure. 

Streamlines can easily be filtered by their length a posteriori. It won't be hard to included this filtering in the tracking as well..

Garyfallidis:
This looks ready to me. Anyone against merging it?

Garyfallidis:
Nobody replied so I am assuming this is okay. Thx Gab and Sam reporting it. Any improvements in another PR please.

PR #855, opened 2016-02-01
0.9.3rc

arokem:
As a stop-gap, we might want to release a 0.9.3, so that we can provide our users on debian platform with something that works through apt-get.

### DIFF
--- a/dipy/denoise/tests/test_noise_estimate.py
+++ b/dipy/denoise/tests/test_noise_estimate.py
@@ -3,8 +3,11 @@ from __future__ import division, print_function
 import numpy as np
 import nibabel as nib
 
-from numpy.testing import assert_almost_equal, assert_equal, assert_array_almost_equal
-from dipy.denoise.noise_estimate import _inv_nchi_cdf, piesno, estimate_sigma
+from numpy.testing import (assert_almost_equal, assert_equal, assert_,
+                           assert_array_almost_equal)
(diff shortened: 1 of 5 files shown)


yarikoptic:
not that travis is really unhappy

dipy/denoise/tests/test_noise_estimate.py
```diff
@@ -29,6 +32,26 @@ def test_piesno():
...
+    # Test using the median as the initial estimation
+    initial_estimation = (np.median(sigma) /
+                          np.sqrt(2 * _inv_nchi_cdf(1, 1, 0.5)))
+    sigma, mask = _piesno_3D(rician_noise, N=1, alpha=0.01, l=1, eps=1e-10,
+                             return_mask=True,
+                             initial_estimation=initial_estimation)
+
+    assert_(np.abs(sigma - 50) / sigma < 0.03)
+
+>>>>>>> 1868e71... BF: Cannot coerce into bool (?).
```
samuelstjean: poor tests had a merge problem

arokem: Oh bugger

arokem:
I think that given the Travis issues, we'll have to ditch this plan. I just don't have time to look into this in the next few days, so it seems like going with 0.10.1 is the best way forward. @yarikoptic - I will create the documentation folder for you from that -- hopefully some time today.

PR #856, opened 2016-02-02
Add reference to gitter chat room in the README

OrkoHunter:
As a new contributor, I believe it is helpful if the chat room is mentioned in the README.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -21,6 +21,8 @@ https://mail.python.org/mailman/listinfo/neuroimaging
 Please see the users' forum at
 https://neurostars.org
 
+Please join the gitter chatroom `here <https://gitter.im/nipy/dipy>`_.
+
 Code
 ====
 


arokem:
I'm +1 for this.

Garyfallidis:
+1

Garyfallidis:
But I am not sure if it should be at the top of the Readme file. Maybe the website link should be first. Just an idea no strong feelings here.

OrkoHunter:
@Garyfallidis What is your opinion of the current implementation?

I think we should expand the README a little bit more, with atleast a basic installation guide. Maybe something like [SymPy](https://github.com/sympy/sympy/blob/master/README.rst) or [NetworkX](https://github.com/networkx/networkx/blob/master/README.rst). If this sounds good to everyone, I may work upon it in this PR itself.

arokem:
It could do with a small rewrite, but I would avoid too much instructions
here. Rather, you would want to direct people to the documentation.
Otherwise, we need to maintain these instructions in two different
places... Thanks!

OrkoHunter:
@arokem Sounds fair !

PR #857, opened 2016-02-02
DTI memory: use the same step in prediction as you use in fitting.

arokem:
This should help us with the issues that are currently being discussed over on #840

I have not memory profiled this yet, though. This is what I intend to do next, so stay tuned. Would be great if others could try this out as well.

### DIFF
--- a/dipy/reconst/dti.py
+++ b/dipy/reconst/dti.py
@@ -1133,7 +1133,7 @@ class TensorFit(object):
         """
         return apparent_diffusion_coef(self.quadratic_form, sphere)
 
-    def predict(self, gtab, S0=1, step=1e4):
+    def predict(self, gtab, S0=1, step=None):
         r"""
         Given a model fit, predict the signal on the vertices of a sphere
 
@@ -1164,9 +1164,8 @@ class TensorFit(object):
         """
         shape = self.model_params.shape[:-1]
         size = np.prod(shape)
-        step = self.model.kwargs.get('step', size)
-        #if step == 2:
-        #    1/0.
+        if step is None:
+            step = self.model.kwargs.get('step', size)
         if step >= size:
             return tensor_prediction(self.model_params[..., 0:12], gtab, S0=S0)
         params = np.reshape(self.model_params,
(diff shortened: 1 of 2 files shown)


MrBago:
@arokem could you do me a favor, while you're at this could you profile [this](https://github.com/mrbago/dipy/tree/streamline_dti_predict)  for me too. It should also improve both the run time and the memory usage of tensor_predict.

arokem:
Will do.

I can also merge that commit into this branch, if that's OK with you.

arokem:
This is memory profiling with bdb2030: https://gist.github.com/arokem/36c41c91e8a4aa1b0bd3

arokem:
This is memory profiling including @MrBago's additions (ad567a1): https://gist.github.com/arokem/a62377a861ca4b02db15

arokem:
I believe this is ready for a review.

samuelstjean:
It actually works now
~~~python
In [35]: %memit data_p = tenfit.predict(gtab, S0)
/home/samuel/python/dipy/reconst/dti.py:1759: RuntimeWarning: divide by zero encountered in log
  D[..., 6] = -np.log(b0)

peak memory: 6689.07 MiB, increment: 5655.26 MiB
~~~
Might want to check that log(b0), just in case it threw off the profiler. A quick guess is that it does log(S0)? At least it does not blow up like before, so that's an improvement.

arokem:
Good to hear that it works for you. 

I am not sure what you are asking about b0/S0. Looks like you have some voxels with the non diffusion-weighted signal at 0 (outside the brain, possibly). I am not sure how that would throw off the profiler.

samuelstjean:
You know, it ran fairly quickly as opposed to not at all, so I was really
surprised. Good if it works, but indeed the hcp data is face/background
masked. Should the predict step skip zero valued voxels to not throw of
errors like this or it's left to the unaware at first user?

2016-02-04 16:29 GMT+01:00 Ariel Rokem <notifications@github.com>:

arokem:
"fairly quickly as opposed to not at all" should be our motto :-)

If you don't mask these voxels yourself, it will try to predict them for
you. We could add some magic to avoid that, but that's probably not as
urgent as this fix, so I will defer that to future work (maybe a GSoC on
handling missing data?).

PR #859, opened 2016-02-04
[WIP] Suppress warnings in tests

### LINKED ISSUE #641: Tests raise a deprecation warning
`UserWarning: This function is deprecated please use dipy.align.reslice.reslice instead.`

OrkoHunter:
Closes #641 

I think this should be WIP until there are no warnings on the travis log ! =D

### DIFF
--- a/dipy/core/tests/test_geometry.py
+++ b/dipy/core/tests/test_geometry.py
@@ -6,6 +6,8 @@ import numpy as np
 
 import random
 
+import warnings
+
 from dipy.core.geometry import (sphere2cart, cart2sphere,
                                 nearest_pos_semi_def,
                                 sphere_distance,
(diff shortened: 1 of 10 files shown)


OrkoHunter:
In my last commit, I've commented out the tests which produce other warnings, most of them being "divided by zero" warning. We can either, suppress that warning by setting
```python
np.seterr(divide='ignore', invalid='ignore')
```
Or, rewrite the test with different data sets.

I think there should not be any more non-genuine warnings on the travis. I'll seek out if any.

samuelstjean:
The test might be there to check if the functions can handle input which are zeros and warn, as expected, in those cases instead of throwing nonsense values or nans down the line.

arokem:
Yes -and in these cases, the warning needs to be caught and verified.

See for example:
https://github.com/nipy/dipy/blob/master/dipy/reconst/tests/test_csdeconv.py#L178

OrkoHunter:
> The test might be there to check if the functions can handle input which are zeros and warn

This sounds really cool. Sorry for interfering with them !

@arokem Okay, I'll use the given example to catch those warnings.

OrkoHunter:
Looking at the Travis log, it looks like 2 of the warnings are only raised in Python 3.4 and 3.5 (and 2.7 on my computer).

arokem:
This seems to be abandoned, so I am closing it, but feel free to reopen if you still want to work on this.

PR #902, opened 2016-02-10
fix typo

samuelstjean:


arokem:
While you're here, maybe a more informative docstring?

samuelstjean:
I have no idea why that docstring is there or what it should say, seems like really old stuff.

Issue #909, opened 2016-02-12
Inconsistent output for values_from_volume


MarcCote:
@arokem thanks for that function, really useful. I played a bit more with it and found some corner cases, that haven't been tested. For instance,

``` python
from dipy.tracking.streamline import values_from_volume
data3D = np.ones((2,2,2))
streamlines = np.ones((10, 2, 3))
values_from_volume(data3D, streamlines).shape
>>> (10, 2)  # So far so good.
streamlines = np.ones((10, 1, 3))
values_from_volume(data3D, streamlines).shape
>>> (10,)  # I would expect (10, 1)

data4D = np.ones((2,2,2,2))
streamlines = np.ones((10, 1, 3))
values_from_volume(data4D, streamlines).shape  # Even worst
>>> ValueError: bad axis1 argument to swapaxes
```

```
/data/research/src/dipy/dipy/tracking/streamline.py in values_from_volume(data, streamlines, affine)
    449 
    450         if isinstance(vals[-1], np.ndarray):
--> 451             return np.swapaxes(np.array(vals), 2, 1).T
    452         else:
    453             new_vals = []

/usr/lib/python2.7/dist-packages/numpy/core/fromnumeric.pyc in swapaxes(a, axis1, axis2)
    487     except AttributeError:
    488         return _wrapit(a, 'swapaxes', axis1, axis2)
--> 489     return swapaxes(axis1, axis2)
    490 
    491 

ValueError: bad axis1 argument to swapaxes
```

I think I found the source of the problem; the `.squeeze()` here:
https://github.com/nipy/dipy/blob/master/dipy/tracking/streamline.py#L394

Is the squeeze really necessary?

arokem:
Thanks for reporting this. As you can see I've started a PR to work on this. 

Just to understand how this arises: is this an actual use-case? Do you actually have a situation where you have a bunch of streamlines and each one has just one node? That might be a tractography problem :-)

arokem:
And to answer your question: that squeeze seems to be necessary for some other use-cases. You'll see the test failure appear soon for this commit: https://github.com/arokem/dipy/commit/a0eb5be64f6550021d061751bd61332caed92fef

arokem:
Is there something obviously wrong about the fix in the most recent commit on #910?

MarcCote:
[...]

PR #910, opened 2016-02-12
Singleton sl vals

arokem:
The goal: value_from_volume to deal with streamlines with only one node each. 

See: #909

### DIFF
--- a/dipy/tracking/streamline.py
+++ b/dipy/tracking/streamline.py
@@ -391,7 +391,7 @@ def _extract_vals(data, streamlines, affine=None, threedvec=False):
             vals = np.array(vfu.interpolate_vector_3d(data, sl_cat)[0])
         else:
             vals = np.array(vfu.interpolate_scalar_3d(data, sl_cat)[0])
-        vals = np.reshape(vals, (sl_shape[0], sl_shape[1], -1)).squeeze()
+        vals = np.reshape(vals, (sl_shape[0], sl_shape[1], -1))
 
     else:
         raise RuntimeError("Extracting values from a volume ",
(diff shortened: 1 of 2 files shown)


arokem:
I've removed the "WIP" from the title. I think this is ready for a review.

MarcCote:
Just tested it, it is working great. :+1: I'll wait tomorrow morning to merge it.

samuelstjean:
How can a single point even happen? Can the tracking return a single point?

samuelstjean:
Oh guess I should read the issues before the PR next time.

arokem:
Cool @MarcCote - thanks for the quick turnaround! Keep in mind that this is doing multiple checks in every call, so if you need more speed for your use-case (sounds interesting!), we can potentially pull out the core of the algorithm to run in an "unsafe" mode (that will do crazy stuff if not used as intended...). Just a thought. Either way, this now seems to cover that use-case as well...

Garyfallidis:
Cool, thx Ariel and Marco for reporting.

MarcCote:
@arokem thanks for the suggestion... oh, wait I just remember there is already a function in Dipy doing a similar thing. Am I missing something or `values_from_coordinates` is the same as @Garyfallidis's function `dipy.viz.util.map_coordinates` (see comments https://github.com/nipy/dipy/pull/587/files#diff-353e2ae8dea5b15c5d1e10a4e611e78eR91) ?

arokem:
Yes. I agree with you, and with the comment you made there. That definitely doesn't belong in viz.

Garyfallidis:
There is a previous discussion about moving all interpolation functions in dipy.core.interpolation or at least call them from there until the old ones get deprecated. I think it is a good plan. We should put it rolling a bit after the release.

PR #915, opened 2016-02-12
BF: correct handling of output paths in dipy_quickbundles.

jchoude:
Now using a splitext call to correctly get the full path of the output file.

Previously, setting the output_file as something in another directory (by using a path like "../other_dir/output.trk") would still output all files in the current directory, with a broken name.

EDIT: I also checked that the other scripts do not have a similar behavior. They don't.

### DIFF
--- a/dipy/tests/test_scripts.py
+++ b/dipy/tests/test_scripts.py
@@ -6,6 +6,7 @@ Run scripts and check outputs
 """
 from __future__ import division, print_function, absolute_import
 
+import glob
 import os
 import shutil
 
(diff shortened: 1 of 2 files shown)


arokem:
Any chance to get a test of this, so that it doesn't creep back in?

jchoude:
Yes, incoming.

arokem:
:sweet_potato:

jchoude:
Just to add, as I posted in an edit to the PR description: I also checked that the other scripts do not have a similar behavior. They don't.

dipy/tests/test_scripts.py
```diff
@@ -121,3 +122,25 @@ def test_qb_commandline():
...
+        os.mkdir('work')
+        os.mkdir('output')
+
+        os.chdir('work')
+        tracks_file = get_data('fornix')
+
+        # Need to specify an output directory with a "../" style path
+        # to trigger old bug.
+        cmd = ["dipy_quickbundles", tracks_file, '--pkl_file', 'mypickle.pkl',
+               '--out_file', '../output/tracks300.trk']
```
arokem: I am not sure that this will work well when testing on a Windows platform (because the path separator is hard coded here). I am running a test of that over here: https://ci.appveyor.com/project/arokem/dipy/build/1.0.649, but it might take a couple of hours to finish running.

jchoude:
Ok I pushed a new version with better handling for the path separators.

arokem:
Great. Thanks!

Issue #889, opened 2016-02-06
PEP8 in utils


arokem:
dipy/utils/_importlib.py:6:1: E302 expected 2 blank lines, found 1
dipy/utils/_importlib.py:16:31: E127 continuation line over-indented for visual indent
dipy/utils/optpkg.py:24:1: E302 expected 2 blank lines, found 1
dipy/utils/optpkg.py:65:80: E501 line too long (120 > 79 characters)
dipy/utils/optpkg.py:83:10: E261 at least two spaces before inline comment
dipy/utils/optpkg.py:85:33: E203 whitespace before ':'
dipy/utils/optpkg.py:91:5: E301 expected 1 blank line, found 0
dipy/utils/optpkg.py:96:1: W391 blank line at end of file
dipy/utils/six.py:12:80: E501 line too long (80 > 79 characters)
dipy/utils/six.py:134:1: E303 too many blank lines (3)
dipy/utils/six.py:158:80: E501 line too long (87 > 79 characters)
dipy/utils/six.py:171:80: E501 line too long (80 > 79 characters)
dipy/utils/six.py:172:80: E501 line too long (80 > 79 characters)
dipy/utils/six.py:290:1: E302 expected 2 blank lines, found 1
dipy/utils/six.py:294:1: E302 expected 2 blank lines, found 1
dipy/utils/six.py:298:1: E302 expected 2 blank lines, found 1
dipy/utils/six.py:306:5: E301 expected 1 blank line, found 0
dipy/utils/six.py:324:5: E301 expected 1 blank line, found 0
dipy/utils/six.py:328:5: E301 expected 1 blank line, found 0
dipy/utils/six.py:330:5: E301 expected 1 blank line, found 0
dipy/utils/six.py:332:5: E301 expected 1 blank line, found 0
dipy/utils/six.py:345:5: E303 too many blank lines (2)
dipy/utils/six.py:351:5: E303 too many blank lines (2)
dipy/utils/six.py:368:5: E303 too many blank lines (2)
dipy/utils/six.py:373:5: E303 too many blank lines (2)
dipy/utils/six.py:378:9: E301 expected 1 blank line, found 0
dipy/utils/tripwire.py:4:1: E302 expected 2 blank lines, found 1
dipy/utils/tripwire.py:38:80: E501 line too long (96 > 79 characters)
dipy/utils/tests/test_arrfuncs.py:18:1: E302 expected 2 blank lines, found 1
[...]

Issue #885, opened 2016-02-06
PEP8 in testing


arokem:
dipy/testing/__init__.py:8:1: E402 module level import not at top of file
dipy/testing/__init__.py:9:1: E402 module level import not at top of file
dipy/testing/__init__.py:10:1: E402 module level import not at top of file
dipy/testing/spherepoints.py:5:1: E302 expected 2 blank lines, found 1
dipy/testing/spherepoints.py:7:25: E231 missing whitespace after ','
dipy/testing/spherepoints.py:9:20: E231 missing whitespace after ','
dipy/testing/spherepoints.py:9:22: E231 missing whitespace after ','
dipy/testing/spherepoints.py:10:20: E231 missing whitespace after ','
dipy/testing/spherepoints.py:10:22: E231 missing whitespace after ','
dipy/testing/spherepoints.py:17:29: E231 missing whitespace after ','
dipy/testing/spherepoints.py:17:31: E231 missing whitespace after ','
dipy/testing/spherepoints.py:22:1: W391 blank line at end of file
dipy/testing/tests/test_decorators.py:34:9: E128 continuation line under-indented for visual indent
dipy/testing/tests/test_decorators.py:47:9: E128 continuation line under-indented for visual indent

ghoshbishakh:
@arokem closed through #958

Issue #875, opened 2016-02-06
PEP8 in denoise


arokem:
dipy/denoise/__init__.py:1:1: E265 block comment should start with '# '
dipy/denoise/noise_estimate.py:21:16: E128 continuation line under-indented for visual indent
dipy/denoise/noise_estimate.py:22:16: E128 continuation line under-indented for visual indent
dipy/denoise/noise_estimate.py:23:16: E128 continuation line under-indented for visual indent
dipy/denoise/noise_estimate.py:24:15: E128 continuation line under-indented for visual indent
dipy/denoise/noise_estimate.py:27:80: E501 line too long (81 > 79 characters)
dipy/denoise/noise_estimate.py:39:80: E501 line too long (81 > 79 characters)
dipy/denoise/noise_estimate.py:74:80: E501 line too long (80 > 79 characters)
dipy/denoise/noise_estimate.py:106:80: E501 line too long (91 > 79 characters)
dipy/denoise/noise_estimate.py:120:80: E501 line too long (96 > 79 characters)
dipy/denoise/noise_estimate.py:188:80: E501 line too long (80 > 79 characters)
dipy/denoise/noise_estimate.py:235:5: E303 too many blank lines (2)
dipy/denoise/noise_estimate.py:238:80: E501 line too long (95 > 79 characters)
dipy/denoise/noise_estimate.py:282:80: E501 line too long (80 > 79 characters)
dipy/denoise/noise_estimate.py:283:80: E501 line too long (90 > 79 characters)
dipy/denoise/noise_estimate.py:298:80: E501 line too long (83 > 79 characters)
dipy/denoise/noise_estimate.py:299:80: E501 line too long (81 > 79 characters)
dipy/denoise/noise_estimate.py:308:80: E501 line too long (89 > 79 characters)
dipy/denoise/noise_estimate.py:321:80: E501 line too long (82 > 79 characters)
dipy/denoise/noise_estimate.py:322:31: E261 at least two spaces before inline comment
dipy/denoise/noise_estimate.py:327:25: E128 continuation line under-indented for visual indent
dipy/denoise/noise_estimate.py:328:25: E128 continuation line under-indented for visual indent
dipy/denoise/noise_estimate.py:329:25: E128 continuation line under-indented for visual indent
dipy/denoise/noise_estimate.py:330:25: E128 continuation line under-indented for visual indent
dipy/denoise/noise_estimate.py:331:25: E128 continuation line under-indented for visual indent
[...]

PR #938, opened 2016-02-23
PEP8 issues solved in utils, testing and denoise

manu-tej:
#885 #875 #889 solved


Garyfallidis:
Thank you for this. But just a quick word to say that the commit messages that you wrote here are misleading. You say for example "Delete decorators.py" when you actually mean is "Corrected pep8 issues in decorators.py". You need to be as precise as possible in your commit messages. Here you are giving the impression that you are deleting files.

manu-tej:
Thanks. It will not happen again.

PR #942, opened 2016-02-28
NF: Added support to colorize each line points indivdually

gauvinalexandre:


MarcCote:
LGTM

gauvinalexandre:
Half a day to find those 5 lines. Damn.

arokem:
I am not sure what's going on here, but it looks like @MarcCote gave this the thumbs-up months ago. Unless someone objects, I can merge this early next week.

gauvinalexandre:
Thank you!

MarcCote:
@gauvinalexandre if you could add some information about this new feature in the docstring that would be nice. Here is a suggestion: "If an array (K, 3) is given, where K is the number of points of all lines then every point is colored with a different RGB color."

I remembered testing it and it was working fine. I should have merged it a while back, so early next week sounds good.

gauvinalexandre:
Thanks @MarcCote ! I would add some info, but unfortunately I don't have much time to go back in the code and explain the use cases and show examples... If this is necessary before the merge, maybe we should wait... I'll add this to my short term todo list.

arokem:
No need for examples here, in my opinion (though we would really welcome
that, in a separate PR). Just the addition of that sentence into the
docstring.

arokem:
I second the suggestion to document this! I'll wait for that to come in and
then merge.

MarcCote:
@gauvinalexandre just merge this PR that I've made on top of your branch: https://github.com/gauvinalexandre/dipy/pull/1

gauvinalexandre:
Thanks @MarcCote !

arokem:
I'm going to ignore the codecov weirdness here, and go ahead with the merge

arokem:
Thanks @gauvinalexandre !

MarcCote:
Probably a rebase was needed since we drop the coverage of benchmark files.

arokem:
That's probably it. Looks like it merged cleanly anyway...

PR #944, opened 2016-02-28
Slicer fix

gauvinalexandre:
I discovered a major contradiction and bug in the value scaling and lookup table making of the slicer feature. I found the way to make it work. Therefore, the behavior of the slicer changed and the test_fvtk_actors fails, probably with reason. It has to be adapted. I'll need help, since I don't know how to interpret the error. Also, there were some mistakes in the doc-strings. I took the occasion to clean the code. I've tested the Slicer with VTK 5 and 6 the best I could and ran the fvtk_actor test. Everything looks good.

### DIFF
--- a/dipy/viz/actor.py
+++ b/dipy/viz/actor.py
@@ -38,7 +38,8 @@ def slicer(data, affine=None, value_range=None, opacity=1.,
     opacity : float
         Opacity of 0 means completely transparent and 1 completely visible.
     lookup_colormap : vtkLookupTable
-        If None (default) then a grayscale map is created.
+        If None (default) then a grayscale map is created. It doesn't apply
+        to rgb volumes where the lookup colormap is ignored.
     interpolation : string
         If 'linear' (default) then linear interpolation is used on the final
         texture mapping. If 'nearest' then nearest neighbor interpolation is
@@ -53,6 +54,8 @@ def slicer(data, affine=None, value_range=None, opacity=1.,
         coordinates as calculated by the affine parameter.
 
     """
+
+    # Verify number of components
     if data.ndim != 3:
         if data.ndim == 4:
             if data.shape[3] != 3:
@@ -64,12 +67,14 @@ def slicer(data, affine=None, value_range=None, opacity=1.,
     else:
         nb_components = 1
 
+    # Rescale data
     if value_range is None:
         data = np.interp(data, xp=[data.min(), data.max()], fp=[0, 255])
     else:
         data = np.interp(data, xp=[value_range[0], value_range[1]], fp=[0, 255])
     data = data.astype('uint8')
 
+    # Allocate VTK Image Data
     im = vtk.vtkImageData()
     if major_version <= 5:
         im.SetScalarTypeToUnsignedChar()
@@ -80,29 +85,19 @@ def slicer(data, affine=None, value_range=None, opacity=1.,
     I, J, K = data.shape[:3]
     im.SetDimensions(I, J, K)
     voxsz = (1., 1., 1.)
-    # im.SetOrigin(0,0,0)
     im.SetSpacing(voxsz[2], voxsz[0], voxsz[1])
 
-    # copy data
-    # what I do below is the same as what is commented here but much faster
-    # for index in ndindex(data.shape):
-    #     i, j, k = index
-    #     im.SetScalarComponentFromFloat(i, j, k, 0, data[i, j, k])
-    data = np.swapaxes(data, 0, 2)
-    data = np.ascontiguousarray(data)
-
+    # Set numpy data to VTK Image Data
+    data = np.ascontiguousarray(np.swapaxes(data, 0, 2))
     if nb_components == 1:
         data = data.ravel()
     else:
         data = np.reshape(data, [np.prod(data.shape[:3]), data.shape[3]])
+    im.GetPointData().SetScalars(numpy_support.numpy_to_vtk(data, deep=0))
 
-    uchar_array = numpy_support.numpy_to_vtk(data, deep=0)
-    im.GetPointData().SetScalars(uchar_array)
-
+    # Set the transform
     if affine is None:
         affine = np.eye(4)
-
-    # Set the transform (identity if none given)
     transform = vtk.vtkTransform()
     transform_matrix = vtk.vtkMatrix4x4()
     transform_matrix.DeepCopy((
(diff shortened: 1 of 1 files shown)


dipy/viz/actor.py
```diff
@@ -38,7 +38,8 @@ def slicer(data, affine=None, value_range=None, opacity=1.,
     opacity : float
         Opacity of 0 means completely transparent and 1 completely visible.
     lookup_colormap : vtkLookupTable
-        If None (default) then a grayscale map is created.
+        If None (default) then a grayscale map is created. It doesn't apply
```
MarcCote: What about: "This parameter is ignored if \`data\` is an RGB volume."

dipy/viz/actor.py
```diff
@@ -64,46 +67,39 @@ def slicer(data, affine=None, value_range=None, opacity=1.,
...
+    # Allocate VTK Image Data
     im = vtk.vtkImageData()
-    if major_version <= 5:
-        im.SetScalarTypeToUnsignedChar()
-    I, J, K = vol.shape[:3]
+    if nb_components == 1:
+        if value_range is None:
+            value_range = (np.min(data), np.max(data))
+        else:
+            # Rescale data
```
MarcCote: Isn't it more "clamping" than rescaling?

dipy/viz/actor.py
```diff
@@ -64,46 +67,39 @@ def slicer(data, affine=None, value_range=None, opacity=1.,
...
+            data[data < value_range[0]] = value_range[0]
```
MarcCote: `data` is going to be modified in-place. This is an unexpected behaviour. If we are going to clamp the data, we should do it on the copy hold by the `vtkImageActor`.

dipy/viz/actor.py
gauvinalexandre: Very nice ! Very better.

dipy/viz/actor.py
gauvinalexandre: - [ ] Independent copy
Yes, you'Re right

dipy/viz/actor.py
gauvinalexandre: -[] Change for clamping

Absolutely.

dipy/viz/actor.py
```diff
@@ -64,46 +68,40 @@ def slicer(data, affine=None, value_range=None, opacity=1.,
...
     im = vtk.vtkImageData()
-    if major_version <= 5:
-        im.SetScalarTypeToUnsignedChar()
-    I, J, K = vol.shape[:3]
+    if nb_components == 1:
+        if value_range is None:
+            value_range = (np.min(data), np.max(data))
+        else:
+            # Clamping data to specified value range
+            data[data < value_range[0]] = value_range[0]
```
Garyfallidis: Hi @gauvinalexandre and sorry for taking long time to get back to you on this.

dipy/viz/actor.py
```diff
@@ -64,46 +68,40 @@ def slicer(data, affine=None, value_range=None, opacity=1.,
...
         nb_components = 1
 
-    if value_range is None:
-        vol = np.interp(data, xp=[data.min(), data.max()], fp=[0, 255])
-    else:
-        vol = np.interp(data, xp=[value_range[0], value_range[1]], fp=[0, 255])
-    vol = vol.astype('uint8')
-
+    # Allocate VTK Image Data
+    data = np.copy(data)
```
Garyfallidis: I used vol before so that we don't change the initial data. Or give the impression that we are doing so.

dipy/viz/actor.py
Garyfallidis: I think here we have a different understanding of what this function should be doing. I wanted this dataset to interpolate  from value_range to (0, 255). I think you want to clamp the data. That is a different goal.

dipy/viz/actor.py
```diff
@@ -64,46 +68,40 @@ def slicer(data, affine=None, value_range=None, opacity=1.,
...
-    #     i, j, k = index
-    #     im.SetScalarComponentFromFloat(i, j, k, 0, vol[i, j, k])
-    vol = np.swapaxes(vol, 0, 2)
-    vol = np.ascontiguousarray(vol)
 
+    # Set numpy data to VTK Image Data
+    data = np.ascontiguousarray(np.swapaxes(data, 0, 2))
     if nb_components == 1:
-        vol = vol.ravel()
+        data = data.ravel()
```
Garyfallidis: I think vol was fine.

dipy/viz/actor.py
```diff
@@ -119,31 +117,31 @@ def slicer(data, affine=None, value_range=None, opacity=1.,
...
-
     image_resliced.SetInterpolationModeToLinear()
     image_resliced.Update()
 
+    # Set the lookup table
     if nb_components == 1:
         if lookup_colormap is None:
             # Create a black/white lookup table.
-            lut = colormap_lookup_table((0, 255), (0, 0), (0, 0), (0, 1))
+            lut = colormap_lookup_table((value_range[0], value_range[1]),
```
Garyfallidis: Ah here you change the range to be anything. Okay this again is a different approach. So, I wonder when you said a bug did really something failed or by reading the docstring you didn't see the purpose of the function? We could add to parameters one for data_range (what I was calling value_range) and one for interp_range what here you use as input to the colormap.

Garyfallidis:
Here is a first review @gauvinalexandre Let me know what you think. Clamping and interpolation are two different things. We probably should try to support both. 

We should also call (optionally) this function for removing outliers. Often most images have only a few outliers with high intensity values and the interpolation generates images with less contrast.
```
from dipy.segment.threshold import upper_bound_by_percent
```

gauvinalexandre:
Hey @Garyfallidis ! This is from a long time ago. I'll get back in my use cases and discuss it with you soon. Thanks for the feedback!

thechargedneutron:
Is this issue still relevant?  Ping @dmreagan  @ranveeraggarwal

gauvinalexandre:
Hi @thechargedneutron ! This issue completely got out of my head a long time ago. What's going on with this repo. I'm not aware of the last year's updates about DIPY. This issue was real. but I didn't took enough time to implement tests and answer to @Garyfallidis review.

skoudoro:
A new visualization tool has been created: [FURY - click here to go to the repo](https://github.com/fury-gl/fury/pulls).

This PR should moved to the new package. Feel free to reopen it on FURY.

As a reminder, the issue fury-gl/fury#34 has been created.

PR #958, opened 2016-03-01
FIX: PEP8 in testing

manu-tej:
issue #885 is fixed


dipy/testing/tests/test_decorators.py
```diff
@@ -31,7 +31,7 @@ def f():
     f2 = doctest_skip_parser(f)
     assert_true(f is f2)
     assert_equal(f2.__doc__,
-        """ Header
+                 """ Header
```
arokem: I think that the indentation here (and below) was intentionally set that way. Could you please change these back?

manu-tej:
@arokem reverted the changes

arokem:
Thanks. LGTM - will let others take a look before I merge.

Issue #880, opened 2016-02-06
PEP8 in reconst


arokem:
dipy/reconst/**init**.py:1:1: E265 block comment should start with '# '
dipy/reconst/**init**.py:12:1: W391 blank line at end of file
dipy/reconst/base.py:25:18: E225 missing whitespace around operator
dipy/reconst/base.py:27:34: E231 missing whitespace after ','
dipy/reconst/base.py:30:1: E302 expected 2 blank lines, found 1
dipy/reconst/cache.py:3:1: E302 expected 2 blank lines, found 1
dipy/reconst/csdeconv.py:94:80: E501 line too long (81 > 79 characters)
dipy/reconst/csdeconv.py:126:45: E712 comparison to False should be 'if cond is False:' or 'if not cond:'
dipy/reconst/csdeconv.py:142:80: E501 line too long (80 > 79 characters)
dipy/reconst/csdeconv.py:167:27: E221 multiple spaces before operator
dipy/reconst/csdeconv.py:183:5: E303 too many blank lines (2)
dipy/reconst/csdeconv.py:257:80: E501 line too long (81 > 79 characters)
dipy/reconst/csdeconv.py:278:45: E712 comparison to False should be 'if cond is False:' or 'if not cond:'
dipy/reconst/csdeconv.py:294:80: E501 line too long (80 > 79 characters)
dipy/reconst/csdeconv.py:399:1: E302 expected 2 blank lines, found 1
dipy/reconst/csdeconv.py:464:80: E501 line too long (80 > 79 characters)
dipy/reconst/csdeconv.py:572:13: E129 visually indented line with same indent as next logical line
dipy/reconst/csdeconv.py:590:80: E501 line too long (88 > 79 characters)
dipy/reconst/csdeconv.py:592:80: E501 line too long (92 > 79 characters)
dipy/reconst/csdeconv.py:628:80: E501 line too long (81 > 79 characters)
dipy/reconst/csdeconv.py:642:16: E203 whitespace before ':'
dipy/reconst/csdeconv.py:648:5: E265 block comment should start with '# '
dipy/reconst/csdeconv.py:658:13: E122 continuation line missing indentation or outdented
dipy/reconst/csdeconv.py:705:80: E501 line too long (80 > 79 characters)
dipy/reconst/csdeconv.py:818:80: E501 line too long (92 > 79 characters)
dipy/reconst/dsi.py:29:80: E501 line too long (116 > 79 characters)
[...]

Issue #881, opened 2016-02-06
PEP8 in reconst


arokem:
dipy/reconst/**init**.py:1:1: E265 block comment should start with '# '
dipy/reconst/**init**.py:12:1: W391 blank line at end of file
dipy/reconst/base.py:25:18: E225 missing whitespace around operator
dipy/reconst/base.py:27:34: E231 missing whitespace after ','
dipy/reconst/base.py:30:1: E302 expected 2 blank lines, found 1
dipy/reconst/cache.py:3:1: E302 expected 2 blank lines, found 1
dipy/reconst/csdeconv.py:94:80: E501 line too long (81 > 79 characters)
dipy/reconst/csdeconv.py:126:45: E712 comparison to False should be 'if cond is False:' or 'if not cond:'
dipy/reconst/csdeconv.py:142:80: E501 line too long (80 > 79 characters)
dipy/reconst/csdeconv.py:167:27: E221 multiple spaces before operator
dipy/reconst/csdeconv.py:183:5: E303 too many blank lines (2)
dipy/reconst/csdeconv.py:257:80: E501 line too long (81 > 79 characters)
dipy/reconst/csdeconv.py:278:45: E712 comparison to False should be 'if cond is False:' or 'if not cond:'
dipy/reconst/csdeconv.py:294:80: E501 line too long (80 > 79 characters)
dipy/reconst/csdeconv.py:399:1: E302 expected 2 blank lines, found 1
dipy/reconst/csdeconv.py:464:80: E501 line too long (80 > 79 characters)
dipy/reconst/csdeconv.py:572:13: E129 visually indented line with same indent as next logical line
dipy/reconst/csdeconv.py:590:80: E501 line too long (88 > 79 characters)
dipy/reconst/csdeconv.py:592:80: E501 line too long (92 > 79 characters)
dipy/reconst/csdeconv.py:628:80: E501 line too long (81 > 79 characters)
dipy/reconst/csdeconv.py:642:16: E203 whitespace before ':'
dipy/reconst/csdeconv.py:648:5: E265 block comment should start with '# '
dipy/reconst/csdeconv.py:658:13: E122 continuation line missing indentation or outdented
dipy/reconst/csdeconv.py:705:80: E501 line too long (80 > 79 characters)
dipy/reconst/csdeconv.py:818:80: E501 line too long (92 > 79 characters)
dipy/reconst/dsi.py:29:80: E501 line too long (116 > 79 characters)
[...]

PR #961, opened 2016-03-02
PEP8 in reconst #880 

theaverageguy:
Fixes https://github.com/nipy/dipy/issues/880


arokem:
This seems to have some replication with #962. Is that my fault in how I set up the issues #880 and #881?

arokem:
Closing for now.

PR #966, opened 2016-03-11
fix typo

samuelstjean:


arokem:
Is this actually a typo? "iff" is sometimes used to denote "if and only if". 

Maybe @matthew-brett can help clarify (and merge or close as needed)?

matthew-brett:
No, it wasn't a typo, did mean 'if and only if'.  I think that's correct - Cython is only a dependency if we don't have the stamped c and pyx files.

samuelstjean:
So much for typo hunting, I had no idea on this one.

matthew-brett:
Typo hunting is good - we can survive a few false positives...

samuelstjean:
Actually I was just looking on how to bundle scripts according to the 
setup.py (still won't work, but whatever), I just fix what I see, can't 
always be right I guess.

PR #967, opened 2016-03-13
Update index.rst correcting the date of release 0.11

Garyfallidis:
I cherry picked the change from maint/0.11.x and updated the 0.11 website.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -41,7 +41,7 @@ See :ref:`older highlights <old_highlights>`.
 Announcements
 *************
 
-- :ref:`Dipy 0.11 <release0.11>` released February 21, 2015.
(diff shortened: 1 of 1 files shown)


arokem:
Pinging @matthew-brett, because I am not sure: is this the way we are supposed to add things that also go into maint/0.11.x?

matthew-brett:
The general idea is to do bug-fixes on maint/0.11.x and maybe doc fixes like this one, and then merge the main brain into master.   But - to do that, you have to be disciplined first to make sure you've correctly done the original merge that ignores the changes you don't want from maint/0.11.x - such as the version number changes, and second, not to cherry-pick back or forward from master.

In the end, it seems to usually happen that someone does this, and so you're back to old-school, cherry-picking commits back and forth.

arokem:
OK. Thanks.

PR #971, opened 2016-03-15
Add progress bar feature to dipy.data.fetcher

### LINKED ISSUE #970: Downloading data with dipy.data.fetcher does not show any progress bar
A progress bar like this might be helpful?
screenshot from 2016-03-15 12-27-13

ghoshbishakh:
fixes #970

### DIFF
--- a/dipy/data/fetcher.py
+++ b/dipy/data/fetcher.py
@@ -36,7 +36,7 @@ def _log(msg):
     print(msg)
 
 
-def update_progressbar(progress):
+def update_progressbar(progress, total_length):
     """Show progressbar
 
     Takes a number between 0 and 1 to indicate progress from 0 to 100%.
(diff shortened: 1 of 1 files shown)


arokem:
Nice one. +1 from me.

Not to bike-shed too much, but it would be nice to have the eventual file-size reported as part of the progress bar (we currently report file size "by hand"). That way, users know what "10%" of the file-size means.

ghoshbishakh:
@arokem sure so should I remove
`_log('Data size is approximately %s' % data_size)`   ?

> eventual file-size reported as part of the progress bar

by as a part of progress bar you mean beside it? or above it?

arokem:
Yes - if we can figure out data_size from the file (apparently you can!), we can get rid of all the places where `data_size` is specified by hand, and replace it by the report alongside the progress bar. 

Maybe put the size of the file right next to the percentage already downloaded?

ghoshbishakh:
screenshot from 2016-03-15 20-58-49
Things are looking a bit cluttered I guess? @arokem 
Should I remove the "Download Progress:" string?

arokem:
This actually looks fine to me.

arokem:
We'd want to get rid of all the other places that have the data_size. 

In the definition of `fetch_data`, and `make_fetcher`:

https://github.com/nipy/dipy/blob/master/dipy/data/fetcher.py#L88
https://github.com/nipy/dipy/blob/master/dipy/data/fetcher.py#L133

But also in the definition of individual fetchers. For example:
https://github.com/nipy/dipy/blob/master/dipy/data/fetcher.py#L284

ghoshbishakh:
@arokem Made the changes! I hope that does not break anything anywhere else.

arokem:
Yes. That looks right to me. If anyone else wants to take a look, please do so in the next few days. Otherwise, I will go ahead and merge this.

arokem:
Closes #970

PR #972, opened 2016-03-17
Fixes #901 : Added documentation for "step" in dti

### LINKED ISSUE #901: DTI `step` parameter not documented.
Need to document what this does at the TensorModel level.

sahmed95:
Fixes #901

### DIFF
--- a/dipy/reconst/dti.py
+++ b/dipy/reconst/dti.py
@@ -1133,7 +1133,13 @@ class TensorFit(object):
            all voxels.
 
         step : int
-            The chunk size as a number of voxels.
+            The chunk size as a number of voxels. Optional parameter with default value 10,000.
+
+            In order to increase speed of processing, tensor fitting is done simultaneously
+            over many voxels. This parameter sets the number of voxels that will be fit at 
+            once in each iteration. A larger step value should speed things up, but it will 
+            also take up more memory. It is advisable to keep an eye on memory consumption 
+            as this value is increased.
 
         Notes
         -----
@@ -1185,7 +1191,13 @@ def iter_fit_tensor(step=1e4):
     Parameters
     ----------
     step : int
-        The chunk size as a number of voxels.
+        The chunk size as a number of voxels. Optional parameter with default value 10,000.
+
+        In order to increase speed of processing, tensor fitting is done simultaneously
+        over many voxels. This parameter sets the number of voxels that will be fit at 
+        once in each iteration. A larger step value should speed things up, but it will 
+        also take up more memory. It is advisable to keep an eye on memory consumption 
+        as this value is increased.
     """
 
     def iter_decorator(fit_tensor):
--- a/dipy/reconst/dti.py
+++ b/dipy/reconst/dti.py
@@ -731,7 +731,7 @@ class TensorModel(ReconstModel):
         take up more memory. It is advisable to keep an eye on memory consumption as 
         this value is increased.
 
-        Example : In iter_fit_tensor() we have a default step value of 1e4            
+        Example : In :func:iter_fit_tensor we have a default step value of 1e4            
 
         References
         ----------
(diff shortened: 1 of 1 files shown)


arokem:
Do you want to sort out those PEP8 issues, while we're here?

dipy/reconst/dti.py
```diff
@@ -1132,6 +1132,9 @@ def predict(self, gtab, S0=1, step=None):
            The mean non-diffusion weighted signal in each voxel. Default: 1 in
            all voxels.
 
+        step : int
+            The chunk size as a number of voxels.
```
arokem: We might want to add a bit more explanation here (after the first sentence). Something like: "to increase speed of processing, tensor fitting is done simultaneously over many voxels. This parameter sets the number of voxels that will be fit at once in each iteration. Note that a larger number here should speed things up, but should also take up more memory, so keep an eye on your computers memory consumption as you increase this number.

Also, we should mention that this is an optional parameter, and that it has a default value (10,000).

We should also add all this to the documentation of the `TensorModel` object.

sahmed95: Ok. I will add this and also update the documentaion for TensorModel object. For the pep8 issues, I can do it in this commit itself. But, if making a new branch just for pep8 isn't a big deal, I guess its better to create a seperate issue and fix pep8 there.

sahmed95: Should the explanation come directly under the definition of step or I should include it under Notes. Also, the line - the increase in step value "should" speed things up but it ~~should~~ will also take up more memory - Is this correct? Regarding the memory 
```python
    """
            The chunk size as a number of voxels. Optional parameter with default value 10,000.

            In order to increase speed of processing, tensor fitting is done simultaneously
            over many voxels. This parameter sets the number of voxels that will be fit at 
            once in each iteration. A larger step value should speed things up, but will 
            also take up more memory. It is advisable to keep an eye on memory consumption 
            as this value is increased.
    """
```

sahmed95:
@arokem Is this fine?

arokem:
I was hoping to also document this at the level of the TensorModel object. Maybe in the `__init__`? It's specific to only some fitting methods (ols, wls), but these are rather important, and we want to give even the users who don't want to dig some information about this, I think.

sahmed95:
So all the functions [here](https://github.com/nipy/dipy/blob/master/dipy/reconst/dti.py#L1122)

arokem:
Rather, somewhere here: https://github.com/nipy/dipy/blob/master/dipy/reconst/dti.py#L696

sahmed95:
Done :smile:

dipy/reconst/dti.py
```diff
@@ -721,6 +722,17 @@ def __init__(self, gtab, fit_method="WLS", *args, **kwargs):
...
+        Note
+        -----
+        In order to increase speed of processing, tensor fitting is done simultaneously
+        over many voxels. Many fit_methods use the 'step' parameter to set the number of
+        voxels that will be fit at once in each iteration. This is the chunk size as a 
+        number of voxels. A larger step value should speed things up, but it will also 
+        take up more memory. It is advisable to keep an eye on memory consumption as 
+        this value is increased.
+
+        Example : In iter_fit_tensor() we have a default step value of 1e4
```
arokem: Please change 
`iter_fit_tensor()`=> `:func:`iter_fit_tensor``

arokem:
Cool. One more small change, so that this renders fine in sphinx (and add's a link in the html to the right function). Thanks for picking up those PEP8 bits as well!

sahmed95:
Done.

arokem:
One more small thing - in the sphinx markup, the name of the function has
to be surrounded by backticks. See here:
http://stackoverflow.com/questions/21289806/link-to-class-method-in-python-docstring

sahmed95:
Updated the backticks

arokem:
This now looks fine to me. Anyone else want to give this a look?

Issue #973, opened 2016-03-18
Minor change in `tools/github_stats.py`


waseem18:
When there are no commits during a particular duration, the python file outputs  'The following 0 authors contributed 0 commits'. 

Wouldn't it be better if it outputs something like "No author has contributed to any commits during this period." ?

arokem:
Yeah. Feel free to make a PR with that.

ghoshbishakh:
closed through #974

PR #974, opened 2016-03-19
Minor change in `tools/github_stats.py`

manu-tej:
Fixed the issue #973

### DIFF
--- a/tools/github_stats.py
+++ b/tools/github_stats.py
@@ -181,17 +181,17 @@ if __name__ == "__main__":
             print("No commits during this period.")
         else: 
             print("The following %i authors contributed %i commits." % (len(unique_authors), ncommits))
- +          print()
+            print()
             print('\n'.join(unique_authors))
             print()
 
-    print()
-    print("We closed a total of %d issues, %d pull requests and %d regular issues;\n"
-          "this is the full list (generated with the script \n"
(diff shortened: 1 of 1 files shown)


tools/github_stats.py
```diff
@@ -177,7 +177,10 @@ def report(issues, show_urls=False):
         all_authors = check_output(author_cmd).splitlines()
         unique_authors = sorted(set(all_authors))
 
-        print("The following %i authors contributed %i commits." % (len(unique_authors), ncommits))
+        if (len(unique_authors) != 0):
```
arokem: How about the following?

   if len(unique_authors) == 0:
       print("No commits during this period.")
   else: 
      ... do all the rest

tools/github_stats.py
```diff
@@ -177,10 +177,13 @@ def report(issues, show_urls=False):
...
+        if len(unique_authors) == 0:
+            print("No commits during this period.")
+        else: 
+            print("The following %i authors contributed %i commits." % (len(unique_authors), ncommits))
+ +          print()
+            print('\n'.join(unique_authors))
+            print()
 
     print()
     print("We closed a total of %d issues, %d pull requests and %d regular issues;\n"
```
arokem: I'd include this block in the `else` block

arokem:
LGTM. I'll let others take a look, before merging.

Issue #975, opened 2016-03-19
Use absolute import in imaffine, imwarp, metrics


sahmed95:
Refer to issue #969 
Change relative imports to absolute imports in .py files inside dipy/dipy/
```python
/dipy/dipy/align/imaffine.py:
   46  import numpy.linalg as npl
   47  import scipy.ndimage as ndimage
   48: from ..core.optimize import Optimizer
   49: from ..core.optimize import SCIPY_LESS_0_12
   50: from . import vector_fields as vf
   51: from . import VerbosityLevels
   52: from .parzenhist import (ParzenJointHistogram,
   53                           sample_domain_regular,
   54                           compute_parzen_mi)
   55: from .imwarp import (get_direction_and_spacings, ScaleSpace)
   56: from .scalespace import IsotropicScaleSpace
   57  from warnings import warn
   58  
/dipy/dipy/align/imwarp.py:
    8  import scipy as sp
    9  import nibabel as nib
   10: from . import vector_fields as vfu
   11: from . import floating
   12: from . import VerbosityLevels
   13: from . import Bunch
   14: from .scalespace import ScaleSpace

/dipy/dipy/align/metrics.py:
    6  import scipy as sp
    7  from scipy import gradient, ndimage
    8: from ..utils.six import with_metaclass
    9: from . import vector_fields as vfu
   10: from . import sumsqdiff as ssd
   11: from . import crosscorr as cc
   12: from . import expectmax as em
   13: from . import floating

```

riddhishb:
I would like to work on this issue

riddhishb:
The corresponding test codes also have the relative import issues
`/dipy/dipy/align/tests/test_imwarp.py`
and
`/dipy/dipy/align/tests/test_metrics.py`

fixing those too

sahmed95:
Hi @riddhishb Thanks for taking a look. You can find a complete list here: http://paste.ubuntu.com/15386849/

I am discussing with dipy developers and will open separate issues for all the cases.
Note: @Garyfallidis pointed out that only the ".py" files in dipy/dipy have to be changed. Leave the .pyx files unchanged.

riddhishb:
Yeah sure thing :) I saw the gitter chats. 
Thats quite a long list, if you want help in creating the separate cases for them, let me know.
here have a look before I make a PR
https://github.com/nipy/dipy/compare/master...riddhishb:relative_import_fix

sahmed95:
[...]

PR #977, opened 2016-03-19
Relative import fix in dipy/align

riddhishb:
FIX: issue #975 plus some codes in dipy/align/tests/ having the same problem.


arokem:
:+1:  from me

omarocegueda:
+1
Thanks @riddhishb!

Issue #978, opened 2016-03-19
Use absolute import in align


sahmed95:
Refer to #969 
Change relative imports to absolute

``` python
/dipy/dipy/align/scalespace.py:
    1: from . import floating
    2  import numpy as np
    3  import numpy.linalg as npl

/dipy/dipy/align/sumsqdiff.pyx:
    4  cimport cython
    5  cimport numpy as cnp
    6: from .fused_types cimport floating, number
    7  cdef extern from "dpy_math.h" nogil:
    8      int dpy_isinf(double)

/dipy/dipy/align/tests/test_crosscorr.py:
    1  import numpy as np
    2  from numpy.testing import assert_array_almost_equal
    3: from .. import floating
    4: from .. import crosscorr as cc
    5  
    6  def test_cc_factors_2d():

/dipy/dipy/align/tests/test_expectmax.py:
    4                             assert_array_almost_equal,
    5                             assert_raises)
    6: from .. import floating
    7: from .. import expectmax as em
    8  
    9  

/dipy/dipy/align/tests/test_imwarp.py:
    7                             assert_array_almost_equal,
    8                             assert_raises)
    9: from ...data import get_data
   10: from .. import floating
   11: from .. import imwarp as imwarp
   12: from .. import metrics as metrics
   13: from .. import vector_fields as vfu
   14: from .. import VerbosityLevels
   15: from ..imwarp import DiffeomorphicMap
   16  
   17  

/dipy/dipy/align/tests/test_metrics.py:
    1  import numpy as np
    2  from scipy import ndimage
    3: from .. import floating
    4: from ..metrics import SSDMetric, CCMetric, EMMetric
    5  from numpy.testing import (assert_array_equal,
    6                             assert_array_almost_equal,

/dipy/dipy/align/tests/test_scalespace.py:
    6                             assert_equal,
    7                             assert_raises)
    8: from .. import floating
    9: from ..imwarp import get_direction_and_spacings
   10: from ..scalespace import (ScaleSpace,
   11                            IsotropicScaleSpace)
   12: from .test_imwarp import get_synthetic_warped_circle
   13  
   14  def test_scale_space():

/dipy/dipy/align/tests/test_sumsqdiff.py:
    1  import numpy as np
    2: from .. import floating
    3: from .. import sumsqdiff as ssd
    4  from numpy.testing import (assert_equal,
    5                             assert_almost_equal,

/dipy/dipy/align/tests/test_transforms.py:
[...]

PR #982, opened 2016-03-20
FIX: relative imports in dipy/align

riddhishb:
#978 fix, except the .pyx files


riddhishb:
This is on top of the changes made in #977. I guess this cleans up atleast dipy/align/

arokem:
LGTM. No big deal here, but for future reference, more informative commit messages would be good :-)

riddhishb:
Yup would keep that in mind!

PR #981, opened 2016-03-20
Fixes #979 : No figures in DKI example - commented savefig()

sahmed95:
Commenting out savefig() in the example should render the plots in sphinx.

PR #990, opened 2016-03-21
No figures in DKI example

theaverageguy:
Fixes https://github.com/nipy/dipy/issues/979


arokem:
I believe this is already being handled over in #981. Let's close this one for now.

arokem:
Thanks for taking a look!

PR #992, opened 2016-03-22
FIX: Update the import statements to use absolute import in core

souravsingh:
The patch fixes the import statement to make use of absolute import in the scripts at core module.
Fixes a small part of https://github.com/nipy/dipy/issues/969


arokem:
LGTM. Let's see that the tests pass.

PR #1003, opened 2016-03-22
FIX: Change the import statements in direction

souravsingh:
Use absolute import in the direction scripts.
Fixes https://github.com/nipy/dipy/issues/996


arokem:
No comments from me. Let's merge this in a couple of days, if no one else pipes up.

PR #1004, opened 2016-03-22
FIX: Use absolute import in pkg_info

souravsingh:
Changed the import statement to use absolute import
Fixes https://github.com/nipy/dipy/issues/998


arokem:
Sure - looks good.

PR #1008, opened 2016-03-22
[WIP] NF: Implementation of CHARMED model

manu-tej:
Created preliminary functions for fitting the data at both low b values and high b values. Documentation will be added soon.


arokem:
Nice start. Don't forget to stick to PEP8 (you have a lot of really long lines).

manu-tej:
@arokem Found the bug that is stopping the fitting functions

Issue #1014, opened 2016-03-23
Progress bar works only for some data


arokem:
Moving the conversation that started over on #981 to here.

arokem:
@ghoshbishakh : I think that if the server doesn't automatically provide data size, then we need to fall back on whether or not the data-size was provided as metadata for this data-set (this is information we removed in #971)

ghoshbishakh:
yes so bringing back the data set metadata and autometically flling back is the way to go. I am working on it.

arokem:
Cool -- :+1: -- thanks for taking a look!

sahmed95:
Following.

arokem:
This is resolved through #1015

PR #1015, opened 2016-03-23
Fix progressbar of fetcher

ghoshbishakh:
Fallback to metadata download size when content-length is not defined in http headers

### DIFF
--- a/dipy/data/fetcher.py
+++ b/dipy/data/fetcher.py
@@ -45,7 +45,7 @@ def update_progressbar(progress, total_length):
     """
     # Try to set the bar_length according to the console size
     try:
-        rows, columns = os.popen('stty size', 'r').read().split()
+        columns = os.popen('tput cols', 'r').read()
         bar_length = int(columns) - 46
         if(not (bar_length > 1)):
             bar_length = 20
--- a/dipy/data/fetcher.py
+++ b/dipy/data/fetcher.py
@@ -118,16 +118,22 @@ def check_md5(filename, stored_md5=None):
 def _get_file_data(fname, url):
     with contextlib.closing(urlopen(url)) as opener:
         if sys.version_info[0] < 3:
-            response_size = opener.headers['content-length']
+            try:
+                response_size = opener.headers['content-length']
+            except KeyError:
+                response_size = None
         else:
             # python3.x
+            # returns none if header not found
             response_size = opener.getheader("Content-Length")
-
         with open(fname, 'wb') as data:
-            copyfileobj_withprogress(opener, data, response_size)
+            if(response_size is None):
+                copyfileobj(opener, data)
+            else:
+                copyfileobj_withprogress(opener, data, response_size)
 
 
-def fetch_data(files, folder):
+def fetch_data(files, folder, data_size=None):
     """Downloads files to folder and checks their md5 checksums
 
     Parameters
@@ -139,7 +145,9 @@ def fetch_data(files, folder):
     folder : str
         The directory where to save the file, the directory will be created if
         it does not already exist.
-
+    data_size : str, optional
+        A string describing the size of the data (e.g. "91 MB") to be logged to
+        the screen. Default does not produce any information about data size.
     Raises
     ------
     FetcherError
@@ -150,6 +158,9 @@ def fetch_data(files, folder):
     if not os.path.exists(folder):
         _log("Creating new folder %s" % (folder))
         os.makedirs(folder)
+
+    if data_size is not None:
+        _log('Data size is approximately %s' % data_size)
 
     all_skip = True
     for f in files:
@@ -168,7 +179,7 @@ def fetch_data(files, folder):
 
 
 def _make_fetcher(name, folder, baseurl, remote_fnames, local_fnames,
-                  md5_list=None, doc="", msg=None,
+                  md5_list=None, doc="", data_size=None, msg=None,
                   unzip=False):
     """ Create a new fetcher
 
@@ -190,6 +201,9 @@ def _make_fetcher(name, folder, baseurl, remote_fnames, local_fnames,
         files. Default: None, skipping checking md5.
     doc : str, optional.
         Documentation of the fetcher.
+    data_size : str, optional.
+        If provided, is sent as a message to the user before downloading
+        starts.
     msg : str, optional.
         A message to print to screen when fetching takes place. Default (None)
         is to print nothing
@@ -207,7 +221,7 @@ def _make_fetcher(name, folder, baseurl, remote_fnames, local_fnames,
         for i, (f, n), in enumerate(zip(remote_fnames, local_fnames)):
             files[n] = (baseurl + f, md5_list[i] if
                         md5_list is not None else None)
-        fetch_data(files, folder)
+        fetch_data(files, folder, data_size)
 
         if msg is not None:
             print(msg)
@@ -246,7 +260,8 @@ fetch_isbi2013_2shell = _make_fetcher(
     ['42911a70f232321cf246315192d69c42',
      '90e8cf66e0f4d9737a3b3c0da24df5ea',
      '4b7aa2757a1ccab140667b76e8075cb1'],
-    doc="Download a 2-shell software phantom dataset")
+    doc="Download a 2-shell software phantom dataset",
+    data_size="")
 
 fetch_stanford_labels = _make_fetcher(
     "fetch_stanford_labels",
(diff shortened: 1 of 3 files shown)


ghoshbishakh:
stops from crashing and falls back to hardcoded download size #1014

dipy/data/fetcher.py
```diff
@@ -117,16 +117,24 @@ def check_md5(filename, stored_md5=None):
 def _get_file_data(fname, url):
     with contextlib.closing(urlopen(url)) as opener:
         if sys.version_info[0] < 3:
-            response_size = opener.headers['content-length']
+            try:
+                response_size = opener.headers['content-length']
+            except:
```
arokem: Please specify the exception type. For example: 

    except KeyError:

ghoshbishakh:
@arokem I hope this is better :sweat_smile:

ghoshbishakh:
okey!

dipy/data/fetcher.py
```diff
@@ -117,16 +117,24 @@ def check_md5(filename, stored_md5=None):
...
+                response_size = None
         else:
             # python3.x
-            response_size = opener.getheader("Content-Length")
-
+            try:
```
arokem: Here as well.

dipy/data/fetcher.py
```diff
@@ -117,16 +117,24 @@ def check_md5(filename, stored_md5=None):
...
+                response_size = opener.getheader("Content-Length")
+            except:
+                response_size = None
         with open(fname, 'wb') as data:
-            copyfileobj_withprogress(opener, data, response_size)
+            if(response_size is None):
+                copyfileobj(opener, data)
+            else:
+                copyfileobj_withprogress(opener, data, response_size)
```
arokem: Is there some way to rewrite this to pass the data size from the metadata into here, so that you still get a progress bar?

ghoshbishakh: yes but that may lead to some weird appearence as the size is not accurate.
I will try and test that maybe tomorrow.

ghoshbishakh:
@arokem made the changes.

arokem:
Great. I don't have any more time to look today, but hopefully I will get
back to this tomorrow. In the meanwhile, we do also need to think about
this failure, if you get a chance:
https://travis-ci.org/arokem/dipy/jobs/118056298#L2279

ghoshbishakh:
Is it that tests are passing but build is failing?
It occured after the progressbar addition?

ghoshbishakh:
@arokem can you give me some idea when the error thrown as all the tests are successful in this PR

arokem:
Yes - the tests are all passing here, but that's because we are not running the `dipy.data` tests on Travis. 

That's what #1012 is supposed to fix: when we install the `data` module's tests, we get test-failures happening (but only on Python 3, you can see that here: https://travis-ci.org/arokem/dipy/builds/118056296)

To debug this, you will have to run the tests locally on your machine, on Python 3.

ghoshbishakh:
@arokem looks like this test provides a url that points to a file. 

`
testfile_url = pathname2url(op.split(symmetric362)[0] + op.sep)
`

And that is not a HTTP url. So it cannot have HTTP headers. That is why we are getting this error on 

`
response_size = opener.getheader("Content-Length")
`

arokem:
Yes. The test is written to mock up a network connection, even when we
don't have a network connection. As you can see, this works fine on Python
2. We might need to rethink how we mock up the network connection, though,
for Python 3.

ghoshbishakh:
I was studying urllib.request.urlopen and it actually returns different things for HTTP url and file url:

`
>>> from urllib.request import urlopen
>>> b = urlopen("http://www.google.com")
>>> a = urlopen("file:///home/bishakh/projects/dipy/dipy/data/files/evenly_distributed_sphere_362.npz")
>>> b
<http.client.HTTPResponse object at 0x7f9478d2e4a8>
>>> a
<addinfourl at 140275661532128 whose fp = <_io.BufferedReader name='/home/bishakh/projects/dipy/dipy/data/files/evenly_distributed_sphere_362.npz'>>
`

ghoshbishakh:
@arokem The most straight forward solution is to fallback on error. I am trying to figure out something more elegant.

ghoshbishakh:
@arokem should we enclose it in a try except to fallback then?

arokem:
It returns different things only on Python 3. One option is to start an http server locally (https://docs.python.org/3/library/http.server.html)and serve that file up (locally!) through that http server so that it looks just like it was served from a remote location.

ghoshbishakh:
@arokem aah! So provided fetcher will never have to fetch a file with file uri we can change the test case that way! So should I proceed to change the test case then?

arokem:
Yes! The fetchers are always supposed to operate against HTTP URLs. The file URL was just a (too naive, apparently) attempt to simulate the real situation. If you can serve the file up through and HTTP server, that's better. Please do ahead and change the test case.

ghoshbishakh:
@arokem Looks better?
fetcher_tests are passing locally.

arokem:
Yes - this looks right to me. Nice work! 

Let me rebase #1012 on top of this, so that we can see that it fixes that one.

arokem:
The rebase might get a bit complicated. I have a better idea: would you mind adding the following one-line change in the `setup.py` to this PR?

https://github.com/nipy/dipy/pull/1012/commits/7ce3a6f8589435f5e603af97b4a72f0b4b96ff75

This addition will make sure that the tests also run on all of our bots.

ghoshbishakh:
@arokem done! hoping for the best :smile:

ghoshbishakh:
@arokem waaah! tests passed! :smile:

arokem:
Travis seems to approve https://travis-ci.org/nipy/dipy/jobs/119295652#L1878

dipy/data/tests/test_fetcher.py
```diff
@@ -67,6 +101,11 @@ def test_fetch_data():
...
         # Test that an error is raised when the md5 checksum of the download
         # file does not match the expected value
-        files = {"testfile.txt": (testfile_url, bad_md5)}
+        files = {"testfile.txt": (test_server_url, bad_md5)}
         npt.assert_raises(fetcher.FetcherError,
                           fetcher.fetch_data, files, tmpdir)
+
+        # stop local HTTP Server
+        server.shutdown()
+        # change to original working directory
```
arokem: Do you want to stop the `server_thread` as well? I actually don't know whether that's needed, which is why I am asking.

dipy/data/fetcher.py
arokem: Just to be sure: we now have a progress bar wherever that's possible ('content-length' is defined), and otherwise it just prints a "data-size is approximately X MB"?

ghoshbishakh: not exactly.
"data-size is approximately X MB" is printed whenever the size is defined even if 'content-length' is defined and progressbar is displayed.

I noticed some datasets in which 'content-length' and data-size both are not available.

dipy/data/tests/test_fetcher.py
ghoshbishakh: no actually force closing the thread is not necessary. server.shutdown() ends the server and the thread target function returns safely and the thread dies peacefully (based on what I understood about threads)

arokem: Cool. Thanks!

arokem:
I've confirmed that this works well. One more tiny comment: could you put in a line-break after the progress bar? See here for example, how the "Files downloaded successfully" message comes in straight after the progress bar:

screen shot 2016-03-29 at 2 27 41 pm

arokem:
Also, is there a way to suppress these messages that appear in the terminal while running this in the jupyter notebook? 

screen shot 2016-03-29 at 2 30 02 pm

ghoshbishakh:
I'll add the line break.
I could not understand themessages in thejupyter notebook. Can you tell me the process to test the issue?

arokem:
Try running this notebook: https://gist.github.com/arokem/fce1873952a0b329c1007b44ec2d0b4e

As the notebook runs, these messages appear in the terminal (in the background). It might not be a big deal, but if you can find an easy fix, that would be good.

ghoshbishakh:
@arokem using `tput cols` to determine size of the terminal fixed the issue for me!

arokem:
Yep. That fixed it for me.

I am going to go ahead and merge this now, as it fixes an existing bug on master. If others have further things, please make a PR :-)

arokem:
Just waiting for Travis to be finished...

ghoshbishakh:
@arokem thanks :blush:

arokem:
Nice work!

ghoshbishakh:
My first proper patch to any proper open source project! Feeling awesome! :smile: 

Though the bug was introduced by me in the first place! :stuck_out_tongue:

PR #1017, opened 2016-03-23
Fixes #1016 : Raises VTK not installed

### LINKED ISSUE #1016: viz.fvtk has no attribute 'ren'
From the examples, in [reconst_dti.py](https://github.com/nipy/dipy/blob/master/doc/examples/reconst_dti.py#L218) we get an AttributeError : fvtk has no attribute 'ren'. Ran this in python2.7 and got the same error. 

```python
>>> from dipy.viz import fvtk
>>> ren = fvtk.ren()
AttributeError: 'module' object has no attribute 'ren'
```
Refer to #981 

Update : Same in Python3

sahmed95:
Informs the user that VTK is not installed.

### DIFF
--- a/dipy/segment/mask.py
+++ b/dipy/segment/mask.py
@@ -8,7 +8,7 @@ from dipy.reconst.dti import fractional_anisotropy, color_fa
 
 from scipy.ndimage.filters import median_filter
 try:
-    from skimage.filter import threshold_otsu as otsu
+    from skimage.filters import threshold_otsu as otsu
 except:
     from .threshold import otsu
 


sahmed95:
Hi, @Garyfallidis @arokem I closed this issue and reverted back to the previous state as it appears that when packages are not installed the optional_pakage function from utils.optpkg takes care of the import error. Instead of raising an error it returns a TripWire message as the import. Calling any attribute of this TripWire object should raise the ImportError and print the message. [Example](https://github.com/sahmed95/dipy/blob/issue-1016/dipy/utils/optpkg.py#L56). 
Thus if VTK is not installed. 
```python
vtk, have_vtk, setup_module = optional_package('vtk')
ren = vtk.ren()
```
Should show the TripWire error. But it doesn't.

fvtk documentation says "# Allow import, but disable doctests if we don't have vtk" I tried with other optional packages like tables in [io.dpy](https://github.com/sahmed95/dipy/blob/issue-1016/dipy/io/dpy.py#L44) and it was giving me the expected TripWire error. This needs more thinking from my side and I will dig into it. Any comments from your side would be helpful.

Issue #1011, opened 2016-03-23
Discrepancy with output of nlmeans.py


riddhishb:
I am working on adaptive denoising patch for dipy, and I noticed an interesting catch in the current implementation of nlmeans.py. I have discussed this with @Garyfallidis 

Following is the output with p=1,b=1 (3x3 block_size and 3x3 patch_size)
![1](https://cloud.githubusercontent.com/assets/7941975/13995710/5af19ef0-f150-11e5-8a73-952523f1b95c.png)

Following is the output with p=3,b=3 (7x7 block_size and 7x7 patch_size)
![2](https://cloud.githubusercontent.com/assets/7941975/13995751/8d5859c4-f150-11e5-8527-86c35d000c03.png)

and the error map between the above two images is 
![3](https://cloud.githubusercontent.com/assets/7941975/13995829/f8f8d96a-f150-11e5-9adb-ee9e83440ebd.png)

Ideally as per the equations (given below) the output should get more smooth on choosing the larger block/patch sizes (if we refer to [this](http://scil.dinf.usherbrooke.ca/wp-content/papers/descoteaux-etal-miccai08.pdf) or [[coupe2008]](http://www.ncbi.nlm.nih.gov/pubmed/18390341)), but that is not the case. (in fact there is no blockwise approach of averaging in denspeed.pyx)

equation 1 (here B is the block size and V is the Patch size)
<img width="298" alt="screen shot 2016-03-24 at 12 41 15 am" src="https://cloud.githubusercontent.com/assets/7941975/13997677/31829db8-f159-11e5-92c6-f509ea72e9df.png">

equation 2
<img width="248" alt="screen shot 2016-03-24 at 12 41 25 am" src="https://cloud.githubusercontent.com/assets/7941975/13997697/4dd70e5e-f159-11e5-8207-68e42adb2b2c.png">

In fact for all cases where patch_size = block_size we get the same output!

I have been working to solve this bug and will make a PR soon with some changes in denspeed.pyx file. I would like to know your comments on the same.

arokem:
Just to be clear: is the output identical in these two cases?

grlee77:
@riddhishb

One might reasonably expect that perhaps the "block size" means that the the central pixel of the patch would traverse a region equal to the block size, but that is not what the code actually does.

[...]

PR #1018, opened 2016-03-24
Fix: denspeed.pyx to give correct output for nlmeans

riddhishb:
The bugfix in the implementation of denspeed in reference to #1011 
Now we can use any patch_radius and block_radius for denoising purpose

- The implementation follows the block weighing approach which is then averaged for each voxel.
- Does not exactly follow the [coupe2008](http://www.ncbi.nlm.nih.gov/pubmed/18390341) block      averaging approach
- To follow the [coupe2008] approach, the existing code by @omarocegueda [here](https://github.com/omarocegueda/denoise/blob/master/ornlm/ornlm.pyx) is accurate, but it needs to be optimized (I am working on that currently!)

### DIFF
--- a/dipy/denoise/denspeed.pyx
+++ b/dipy/denoise/denspeed.pyx
@@ -58,14 +58,14 @@ def nlmeans_3d(arr, mask=None, sigma=None, patch_radius=1,
         raise ValueError('sigma needs to be a 3D ndarray', sigma.shape)
 
     arr = np.ascontiguousarray(arr, dtype='f8')
-    arr = add_padding_reflection(arr, block_radius)
-    mask = add_padding_reflection(mask.astype('f8'), block_radius)
+    arr = add_padding_reflection(arr, block_radius + patch_radius)
+    mask = add_padding_reflection(mask.astype('f8'), block_radius + patch_radius)
     sigma = np.ascontiguousarray(sigma, dtype='f8')
-    sigma = add_padding_reflection(sigma.astype('f8'), block_radius)
+    sigma = add_padding_reflection(sigma.astype('f8'), block_radius + patch_radius)
     arrnlm = _nlmeans_3d(arr, mask, sigma, patch_radius, block_radius,
                          rician, num_threads)
 
-    return remove_padding(arrnlm, block_radius)
+    return remove_padding(arrnlm, block_radius+patch_radius)
 
 
 @cython.wraparound(False)
@@ -103,10 +103,9 @@ def _nlmeans_3d(double [:, :, ::1] arr, double [:, :, ::1] mask,
 
     # move the block
     with nogil, parallel():
-
-        for i in prange(B, I - B):
-            for j in range(B, J - B):
-                for k in range(B, K - B):
+        for i in range(B+P, I - B-P):
+            for j in range(B+P, J - B-P):
+                for k in range(B+P, K - B-P):
 
                     if mask[i, j, k] == 0:
                         continue
@@ -153,49 +152,53 @@ cdef double process_block(double [:, :, ::1] arr,
 
     cdef:
         cnp.npy_intp m, n, o, M, N, O, a, b, c, cnt, step
-        double patch_vol_size
+        double patch_vol_size,block_vol_size
         double summ, d, w, sumw, sum_out, x, sigm
         double * W
         double * cache
         double * sigma_block
         double denom
         cnp.npy_intp BS = B * 2 + 1
+        cnp.npy_intp PS = P * 2 + 1
+        cnp.npy_intp TS = PS + BS - 1
 
     cnt = 0
     sumw = 0
     patch_vol_size = (P + P + 1) * (P + P + 1) * (P + P + 1)
+    block_vol_size = (B + B + 1) * (B + B + 1) * (B + B + 1)
+    
 
-    W = <double *> malloc(BS * BS * BS * sizeof(double))
-    cache = <double *> malloc(BS * BS * BS * sizeof(double))
-    sigma_block = <double *> malloc(BS * BS * BS * sizeof(double))
+    W = <double *> malloc(PS * PS * PS * sizeof(double))
+    cache = <double *> malloc(TS * TS * TS * sizeof(double))
+    sigma_block = <double *> malloc(TS * TS * TS * sizeof(double))
 
     # (i, j, k) coordinates are the center of the static patch
     # copy block in cache
-    copy_block_3d(cache, BS, BS, BS, arr, i - B, j - B, k - B)
-    copy_block_3d(sigma_block, BS, BS, BS, sigma, i - B, j - B, k - B)
+    copy_block_3d(cache, PS + BS- 1, PS + BS - 1, PS + BS - 1, arr, i - P - B, j - P - B, k - P - B)
+    copy_block_3d(sigma_block, PS + BS- 1, PS + BS - 1, PS + BS - 1, arr, i - P - B, j - P - B, k - P - B)
 
     # calculate weights between the central patch and the moving patch in block
     # (m, n, o) coordinates are the center of the moving patch
     # (a, b, c) run inside both patches
-    for m in range(P, BS - P):
-        for n in range(P, BS - P):
-            for o in range(P, BS - P):
+    for m in range(-P, P + 1):
+        for n in range(-P, P + 1):
+            for o in range(-P, P + 1):
 
                 summ = 0
                 sigm = 0
 
                 # calculate square distance
-                for a in range(-P, P + 1):
-                    for b in range(-P, P + 1):
-                        for c in range(-P, P + 1):
+                for a in range(-B, B + 1):
+                    for b in range(-B, B + 1):
+                        for c in range(-B, B + 1):
 
                             # this line takes most of the time! mem access
-                            d = cache[(B + a) * BS * BS + (B + b) * BS + (B + c)] - cache[(m + a) * BS * BS + (n + b) * BS + (o + c)]
+                            d = cache[(P + B + a) * TS * TS + (P + B + b) * TS + (P + B + c)] - cache[(P + B + m + a) * TS * TS + (P + B + n + b) * TS + (P + B + o + c)]
                             summ += d * d
-                            sigm += sigma_block[(m + a) * BS * BS + (n + b) * BS + (o + c)]
+                            sigm += sigma_block[(P + B + m + a) * TS * TS + (P + B + n + b) * TS + (P + B + o + c)]
 
-                denom = sqrt(2) * (sigm / patch_vol_size)**2
-                w = exp(-(summ / patch_vol_size) / denom)
+                denom = (sigm / block_vol_size)**2
+                w = exp(-(summ / block_vol_size) / denom)
                 sumw += w
                 W[cnt] = w
                 cnt += 1
@@ -205,16 +208,16 @@ cdef double process_block(double [:, :, ::1] arr,
 
     # calculate normalized weights and sums of the weights with the positions
     # of the patches
-    for m in range(P, BS - P):
-        for n in range(P, BS - P):
-            for o in range(P, BS - P):
+    for m in range(-P, P + 1):
+        for n in range(-P, P + 1):
+            for o in range(-P, P + 1):
 
                 if sumw > 0:
                     w = W[cnt] / sumw
                 else:
                     w = 0
 
-                x = cache[m * BS * BS + n * BS + o]
+                x = cache[(P + B + m) * TS * TS + (P + B + n) * TS + (P + B + o)]
                 sum_out += w * x * x
                 cnt += 1
 
--- a/dipy/denoise/denspeed.pyx
+++ b/dipy/denoise/denspeed.pyx
@@ -197,7 +197,7 @@ cdef double process_block(double [:, :, ::1] arr,
                             summ += d * d
                             sigm += sigma_block[(P + B + m + a) * TS * TS + (P + B + n + b) * TS + (P + B + o + c)]
 
-                denom = (sigm / block_vol_size)**2
+                denom = sqrt(2) * (sigm / block_vol_size)**2
                 w = exp(-(summ / block_vol_size) / denom)
                 sumw += w
                 W[cnt] = w
--- a/dipy/denoise/denspeed.pyx
+++ b/dipy/denoise/denspeed.pyx
@@ -19,6 +19,7 @@ def nlmeans_3d(arr, mask=None, sigma=None, patch_radius=1,
                block_radius=5, rician=True, num_threads=None):
     """ Non-local means for denoising 3D images
 
+
     Parameters
     ----------
     arr : 3D ndarray
@@ -59,19 +60,23 @@ def nlmeans_3d(arr, mask=None, sigma=None, patch_radius=1,
 
     arr = np.ascontiguousarray(arr, dtype='f8')
     arr = add_padding_reflection(arr, block_radius + patch_radius)
-    mask = add_padding_reflection(mask.astype('f8'), block_radius + patch_radius)
+    mask = add_padding_reflection(
+        mask.astype('f8'),
+        block_radius + patch_radius)
     sigma = np.ascontiguousarray(sigma, dtype='f8')
-    sigma = add_padding_reflection(sigma.astype('f8'), block_radius + patch_radius)
+    sigma = add_padding_reflection(
+        sigma.astype('f8'),
+        block_radius + patch_radius)
     arrnlm = _nlmeans_3d(arr, mask, sigma, patch_radius, block_radius,
                          rician, num_threads)
 
-    return remove_padding(arrnlm, block_radius+patch_radius)
+    return remove_padding(arrnlm, block_radius + patch_radius)
 
 
 @cython.wraparound(False)
 @cython.boundscheck(False)
-def _nlmeans_3d(double [:, :, ::1] arr, double [:, :, ::1] mask,
-                double [:, :, ::1] sigma, patch_radius=1, block_radius=5,
+def _nlmeans_3d(double[:, :, ::1] arr, double[:, :, ::1] mask,
+                double[:, :, ::1] sigma, patch_radius=1, block_radius=5,
                 rician=True, num_threads=None):
     """ This algorithm denoises the value of every voxel (i, j, k) by
     calculating a weight between a moving 3D patch and a static 3D patch
@@ -81,7 +86,7 @@ def _nlmeans_3d(double [:, :, ::1] arr, double [:, :, ::1] mask,
 
     cdef:
         cnp.npy_intp i, j, k, I, J, K
-        double [:, :, ::1] out = np.zeros_like(arr)
+        double[:, :, ::1] out = np.zeros_like(arr)
         double summ = 0
         cnp.npy_intp P = patch_radius
         cnp.npy_intp B = block_radius
@@ -103,9 +108,9 @@ def _nlmeans_3d(double [:, :, ::1] arr, double [:, :, ::1] mask,
 
     # move the block
     with nogil, parallel():
-        for i in range(B+P, I - B-P):
-            for j in range(B+P, J - B-P):
-                for k in range(B+P, K - B-P):
+        for i in prange(B + P, I - B - P):
+            for j in range(B + P, J - B - P):
+                for k in range(B + P, K - B - P):
 
                     if mask[i, j, k] == 0:
                         continue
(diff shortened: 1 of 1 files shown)


samuelstjean:
The implementation here is actually only for the voxelwise average, and not the blockwise average.

The thing we are actually missing is the weighting check based on the mean and variance of eq. 11, never got around to implement it. Not sure why the sqrt(2) factor in the denominator went away though.

riddhishb:
@samuelstjean if it's voxelwise average than the patch submitted fixes it, right? (except for the enhancement by adding equation 11)

Also sorry about the sqrt(2) removal, will change that :D 

One more thing, isn't the blockwise approach more helpful in denoising?  The results by [ornlm](https://github.com/omarocegueda/denoise/blob/master/ornlm/ornlm.pyx) are much better than the voxelwise implementation, the only drawback is the speed which can be improved if optimized.

samuelstjean:
That's the thing, might be hard to sell the it's better but slower angle, I tried to add the option to use a 3D volume, which had a performance cost of around 10-15 secs over a full run of the algorithm, and I pretty much had to outline a bunch of timing tests to show that.

If you go for splitting it in a second function, there will be a ton of duplication, so I don't know how to would go either. Adding a keyword between voxelwise and blockwise could be a compromise maybe.

The implementation we went for is actually this one http://scil.dinf.usherbrooke.ca/wp-content/papers/descoteaux-etal-miccai08.pdf, so that is why it's the voxelwise and not the blockwise version.

riddhishb:
@samuelstjean Fair point.

Will change the sqrt(2) and try to include the equation 11 thing. 
Yeah the patch submitted follows the http://scil.dinf.usherbrooke.ca/wp-content/papers/descoteaux-etal-miccai08.pdf approach, we need to update the docstrings as well (it mentions the [coupe2008] paper). Can you run this thing to verify the correctness?

A keyword may work, I will try to code it up.

Also I am not sure why is the travis cl check failing, it shows a fail in test_nlmeans.py for python 2.7
but when I do
`python2.7 -m nose test_nlmeans.py` 
in my local machine it passes all the tests.

samuelstjean:
It is not wrong per say for the docstring, the paper version outlines both the voxelwise and the blockwise version, but we had a binary executable of the voxelwise version, so we went with it for convenience.

These seems to be some weird test checking that the parallel version is faster than the multithreaded version, but travis might use some weird openmp stuff, explaining why some test fails. I know I had problem with the default not running fullspeed, but I did not write the part messing with openmp or it's tests, might be worth asking the original perosn who added that (default was to run full speed, but because of reasons, mac osx did not have openmp support by default in apple's stock compiler configuration or something, and it behaved weirdly).

riddhishb:
@samuelstjean yeah apple and openmp is a bad combo. 
So what are your suggestions for the patch itself (correctness in regards to bugfix for #1011 )? 
Should I try the keyword based approach, or let the voxelwise approach remain as it is?

If we were to code for some adaptive techniques for denoising then the blockwise approach gives a much better result, so maybe a keyword does make sense

samuelstjean:
Actually, I modified nlmeans a while back so it could support a whole 3D volume for the noise estimation, so it is easier for anyone to roll out his own adaptive version transparently thanks to that. For the blockwise approach, it does work better for structural images (which is what nlmeans what made for in the first place), but since this library does diffusion mri, many people use it for processing the whole bunch of 3D volumes indepently.

While this works on pape,r I fear that in practice the blockwise approach will smear out your edged near regions of signal drop between adjacent structure, hence why we went for the voxelwise version, but feel fre to try it. Now that I saw the issue (did not the first time), I am startint to wonder if the bug was in my original version of the cython translation, go figure. At the very least it should flag an error for patchsize >= blocksize I'd say.

riddhishb:
I will try it.
For the bugfix does anything else needs to be changed apart from the cython file?

dipy/denoise/denspeed.pyx
```diff
@@ -103,10 +103,9 @@ def _nlmeans_3d(double [:, :, ::1] arr, double [:, :, ::1] mask,
 
     # move the block
     with nogil, parallel():
-
-        for i in prange(B, I - B):
-            for j in range(B, J - B):
-                for k in range(B, K - B):
+        for i in range(B+P, I - B-P):
```
Garyfallidis: space between operators
All pep8 should be supported in Cython files except from the line width limitation.

dipy/denoise/denspeed.pyx
```diff
@@ -153,49 +152,53 @@ cdef double process_block(double [:, :, ::1] arr,
 
     cdef:
         cnp.npy_intp m, n, o, M, N, O, a, b, c, cnt, step
-        double patch_vol_size
+        double patch_vol_size,block_vol_size
```
Garyfallidis: same space after comma

Garyfallidis:
Hi @riddhishb  I am a bit worried about the error that you are currently getting. 

    assert_equal(duration_2core < duration_1core, True)
False

Basically the error says that using two cpus is slower than one. Have you looked into how the threading is used in this code? Please make sure that threading is used optimally in this code after your added changes. Let me know if you need some help figuring this out.

dipy/denoise/denspeed.pyx
Garyfallidis: space between operators

riddhishb:
@Garyfallidis I am trying to fix that bug, but for some reason I am not able to annotate the cython code for denspeed

I do `cython -a denspeed.pyx`

but I get this 


Error compiling Cython file

> import numpy as np
> cimport numpy as cnp
> 
> cimport cython
> cimport safe_openmp as openmp

denspeed.pyx:7:8: 'safe_openmp.pxd' not found

Error compiling Cython file

> import numpy as np
> cimport numpy as cnp
> 
> cimport cython
> cimport safe_openmp as openmp
> from safe_openmp cimport have_openmp

denspeed.pyx:8:0: 'safe_openmp/have_openmp.pxd' not found

I cannot figure out as to why os this happening, any ideas ?

Garyfallidis:
@riddhishb go to the main directory of dipy and do 
make clean
make ext

Garyfallidis:
Okay when I tried cython -a dipy/alignt/bundlemin.pyx I also get some errors. For example
For example bundlemin.pyx:224:36: Converting to Python object not allowed without gil

Garyfallidis:
Bundlmin uses also threads (openmp). What this is telling us is that the cython command doesn't know where to find openmp. And that is correct.

Garyfallidis:
In setup.py we especially express how to find the headers etc before compiling the cython files. However, the cython command doesn't have that information.

riddhishb:
oh fair point, so any workaround to that?, yeah bundlemin is have the same king of openmp imports.

Garyfallidis:
Maybe you can provide extra information to the cython command. I haven't tried it yet.

riddhishb:
Ok will try that!

Garyfallidis:
The cython command is like gcc you can add include files etc. I hope this tip helps ;).

riddhishb:
Yeah solved :)

Garyfallidis:
Cool, can you provide here the command that worked for you? It is good to keep track of things as we move ahead. It may be interesting for other people too.

riddhishb:
`cython -I src/ -a dipy/denoise/denspeed.pyx` did the trick, the cython files for openmp are in that folder

dipy/denoise/denspeed.pyx
```diff
@@ -153,49 +157,70 @@ cdef double process_block(double [:, :, ::1] arr,
...
 
     cnt = 0
     sumw = 0
     patch_vol_size = (P + P + 1) * (P + P + 1) * (P + P + 1)
+    block_vol_size = (B + B + 1) * (B + B + 1) * (B + B + 1)
 
-    W = <double *> malloc(BS * BS * BS * sizeof(double))
-    cache = <double *> malloc(BS * BS * BS * sizeof(double))
-    sigma_block = <double *> malloc(BS * BS * BS * sizeof(double))
+    W = <double * > malloc(PS * PS * PS * sizeof(double))
```
Garyfallidis: No need for space after * in this case because now * is not an operator but a pointer.

dipy/denoise/denspeed.pyx
```diff
@@ -153,49 +157,70 @@ cdef double process_block(double [:, :, ::1] arr,
...
+    cache = <double * > malloc(TS * TS * TS * sizeof(double))
```
Garyfallidis: same

dipy/denoise/denspeed.pyx
```diff
@@ -153,49 +157,70 @@ cdef double process_block(double [:, :, ::1] arr,
...
+    sigma_block = <double * > malloc(TS * TS * TS * sizeof(double))
```
Garyfallidis: same

dipy/denoise/denspeed.pyx
```diff
@@ -273,7 +307,7 @@ cdef cnp.npy_intp copy_block_3d(double * dest,
 
     for i in range(I):
         for j in range(J):
-            memcpy(&dest[i * J * K  + j * K], &source[i + min_i, j + min_j, min_k], K * sizeof(double))
+            memcpy(& dest[i * J * K + j * K], & source[i + min_i, j + min_j, min_k], K * sizeof(double))
```
Garyfallidis: No need for space here.

Garyfallidis:
Hi @riddhishb, did you figure out why the code was slower with 2 cpus? Can you also report to me the timings between the old version and the new version of the code? Any changes in quality? I also pointed out some simple stylistic changes. I think we are very close to merge this. Let us know what you think.

dipy/denoise/denspeed.pyx
```diff
@@ -103,10 +108,9 @@ def _nlmeans_3d(double [:, :, ::1] arr, double [:, :, ::1] mask,
 
     # move the block
     with nogil, parallel():
-
-        for i in prange(B, I - B):
-            for j in range(B, J - B):
-                for k in range(B, K - B):
+        for i in prange(B + P, I - B - P):
```
riddhishb: @Garyfallidis The bug for the cpu speed mismatch was pretty simple, I didn't use `prange` here which was giving that error. 
The times are, for the denoise_nlmeans example from doc/examples/ 
Original = 4.70s
Updated = 4.65s
so it's not much different. Also in the quality the updated one gives a much better denoised output (maybe too smooth at times), it also solves the bug reported in #1011 
I will commit those simple changes, I also think that this can be merged with the main branch, it's nearly done.

Garyfallidis:
Okay, I am merging this. Let's try to monitor it during the next months using different datasets and see if it is all good. For example a dataset where the brain is touching the edge of the volume would be good to try. Some times with patch methods someone needs to be careful with what happens near the boundaries. So far the tests are happy :) Congrats @riddhishb !

riddhishb:
@Garyfallidis Thank You :)
I will run different tests in several limiting conditions on this patch and report the outputs in some kind of documentation in next few days.

dipy/denoise/denspeed.pyx
```diff
@@ -153,49 +157,70 @@ cdef double process_block(double [:, :, ::1] arr,
...
+        i - P - B,
+        j - P - B,
+        k - P - B)
 
     # calculate weights between the central patch and the moving patch in block
     # (m, n, o) coordinates are the center of the moving patch
     # (a, b, c) run inside both patches
-    for m in range(P, BS - P):
-        for n in range(P, BS - P):
-            for o in range(P, BS - P):
```
riddhishb: @Garyfallidis @omarocegueda 
The essential algorithm following [this](http://link.springer.com/chapter/10.1007%2F978-3-540-85990-1_15) is that we have a patch(V) around a voxel(x) of radius (P), and then we do the following to get the denoised output
screen shot 2016-06-22 at 11 19 56 pm

and the weights are computed by computing distance between a patch (size S) with a different radius (B), centered at x and x_i, for all i in V. Mathematically 

screen shot 2016-06-22 at 11 28 25 pm

So in the current code there is some correction in normalization of the weights (will correct that, but it does not affect the oversmoothing problem). 

The changed code exactly follows the algorithm. The problem with the previous code lies primarily in these lines as it did something entirely different, and if you keep P=B  or P>B you can see these loops will cause a problem. Hence I restructured it totally and followed the paper's algorithm step by step. I rechecked it, I do not find any problem with the implementation apart from the normalization of sigma (lines 222, 223). Please do have look at this.

dipy/denoise/denspeed.pyx
```diff
@@ -153,49 +157,70 @@ cdef double process_block(double [:, :, ::1] arr,
...
                             summ += d * d
-                            sigm += sigma_block[(m + a) * BS * BS + (n + b) * BS + (o + c)]
+                            sigm += sigma_block[(P + B + m + a) * TS *
+                                                TS + (P + B + n + b) * TS + (P + B + o + c)]
 
-                denom = sqrt(2) * (sigm / patch_vol_size)**2
-                w = exp(-(summ / patch_vol_size) / denom)
+                denom = sqrt(2) * (sigm / block_vol_size)**2
+                w = exp(-(summ / block_vol_size) / denom)
                 sumw += w
```
riddhishb: @Garyfallidis these will be changed to
`denom = sigm / block_vol_size`
`w = exp(-sqrt(summ/ block_vol_size)/denom)`

dipy/denoise/denspeed.pyx
```diff
@@ -153,49 +157,70 @@ cdef double process_block(double [:, :, ::1] arr,
...
 
     cnt = 0
     sumw = 0
     patch_vol_size = (P + P + 1) * (P + P + 1) * (P + P + 1)
+    block_vol_size = (B + B + 1) * (B + B + 1) * (B + B + 1)
 
-    W = <double *> malloc(BS * BS * BS * sizeof(double))
-    cache = <double *> malloc(BS * BS * BS * sizeof(double))
-    sigma_block = <double *> malloc(BS * BS * BS * sizeof(double))
+    W = <double *> malloc(PS * PS * PS * sizeof(double))
```
grlee77: pretty sure this should still be BS * BS * BS.  You want to exam the distances for all patches falling within the block.

dipy/denoise/denspeed.pyx
```diff
@@ -153,49 +157,70 @@ cdef double process_block(double [:, :, ::1] arr,
...
 
     # (i, j, k) coordinates are the center of the static patch
     # copy block in cache
-    copy_block_3d(cache, BS, BS, BS, arr, i - B, j - B, k - B)
-    copy_block_3d(sigma_block, BS, BS, BS, sigma, i - B, j - B, k - B)
+    copy_block_3d(
+        cache,
+        PS + BS - 1,
+        PS + BS - 1,
+        PS + BS - 1,
```
grlee77: not important, but you have already defined TS = PS + BS - 1 above

dipy/denoise/denspeed.pyx
```diff
@@ -153,49 +157,70 @@ cdef double process_block(double [:, :, ::1] arr,
...
 
     # calculate weights between the central patch and the moving patch in block
     # (m, n, o) coordinates are the center of the moving patch
     # (a, b, c) run inside both patches
-    for m in range(P, BS - P):
-        for n in range(P, BS - P):
-            for o in range(P, BS - P):
+    for m in range(-P, P + 1):
+        for n in range(-P, P + 1):
+            for o in range(-P, P + 1):
```
grlee77: I think the outer m, n, o loops should be over blocks as it was before

dipy/denoise/denspeed.pyx
```diff
@@ -153,49 +157,70 @@ cdef double process_block(double [:, :, ::1] arr,
...
                 summ = 0
                 sigm = 0
 
                 # calculate square distance
-                for a in range(-P, P + 1):
-                    for b in range(-P, P + 1):
-                        for c in range(-P, P + 1):
+                for a in range(-B, B + 1):
+                    for b in range(-B, B + 1):
+                        for c in range(-B, B + 1):
```
grlee77: The inner a, b, c loops should be over the patches as it was before

dipy/denoise/denspeed.pyx
```diff
@@ -153,49 +157,70 @@ cdef double process_block(double [:, :, ::1] arr,
...
+                                (P + B + m + a) * TS * TS + (P + B + n + b) * TS + (P + B + o + c)]
                             summ += d * d
-                            sigm += sigma_block[(m + a) * BS * BS + (n + b) * BS + (o + c)]
+                            sigm += sigma_block[(P + B + m + a) * TS *
+                                                TS + (P + B + n + b) * TS + (P + B + o + c)]
 
-                denom = sqrt(2) * (sigm / patch_vol_size)**2
-                w = exp(-(summ / patch_vol_size) / denom)
+                denom = sqrt(2) * (sigm / block_vol_size)**2
+                w = exp(-(summ / block_vol_size) / denom)
```
grlee77: patch_sizes, not block sizes should be in these denominators

dipy/denoise/denspeed.pyx
```diff
@@ -205,16 +230,17 @@ cdef double process_block(double [:, :, ::1] arr,
 
     # calculate normalized weights and sums of the weights with the positions
     # of the patches
-    for m in range(P, BS - P):
-        for n in range(P, BS - P):
-            for o in range(P, BS - P):
+    for m in range(-P, P + 1):
+        for n in range(-P, P + 1):
+            for o in range(-P, P + 1):
```
grlee77: The m, n, o loops here should be over -B, B + 1

dipy/denoise/denspeed.pyx
riddhishb: Ok, maybe some naming inconsistency, I take the central patch of the size PS and then compute the distance between the blocks BS between the block around central voxel and the block around the voxel in the patch.

dipy/denoise/denspeed.pyx
riddhishb: I guess these two comments will also be taken care if we consider that naming inconsistency right.

dipy/denoise/denspeed.pyx
grlee77: The m, n, o loops here should also be over -B, B + 1 as before

dipy/denoise/denspeed.pyx
grlee77: But what you want are differences between patches of size PS for all **patches** within BS. There should be no comparing over distances between **blocks**.  This is what the code was doing before.  I think the bug in the old code was just that the block radius searched over was smaller than specified by one patch radius, not in the actual implementation.

riddhishb: Ok @grlee77 if you can just wait for few mins then I will try the changes you suggested and give you the output

samuelstjean:
Did realize this was back in march, I have not looked at it since april (ismrm + summer stuff), and now I started receiving tons of emails. 

So actually my original concern about 'there is probably no bug' is that if you select patch_size = block_size, your search region is of the same size as what you are trying to denoise, hence the mean of one element is one element, and nothing is processed. Of course this is still valid no matter the size you select, as you have basically no effective search region. Yes, it's a strange corner case, but perfectly valid mathematically.

So in that matter, apparently there was a problem, and I now see a bunch of TS variables, so what exactly is the logic we need to double check for problems? Pretty sure if anyone runs the example on the latest master with the default everywhere you will see the huge effect reported earlier.

samuelstjean:
Ah yes in any case the original (closed source) matlab code is available to double check in case you have discrepancies. The original version pretty much mimics the original C version (missing the mean and variable dropout check, if it has it), which we do not have the sourse anymore, as @mdesco mentionned earlier.

https://sites.google.com/site/pierrickcoupe/softwares/denoising-for-medical-imaging/mri-denoising/mri-denoising-software
http://personales.upv.es/jmanjon/denoising/index.htm

riddhishb:
@samuelstjean The TS variables are because the cache variable should be of the Patch size + block size, and then the big block TS takes care of all this, and even the initial buggy case of blocksize > patchsize .

Also on the coupe's website the ornlm implementation which incorporates the same algorithm as that of denspeed but with blockwise averaging is being incorporated with a keyword (like you initially suggested) through #1066,

samuelstjean:
Yes saw that go by, still busy doign stuff for end of july, might look at it when finished.
By the way, if you want ot roll out an adaptative version based on nlmeans, this version accepts a 3D array especially for that purpose, so might want to re-use all those parts.

Makes it easier to roll out your own version by computing the sigma volume externally with the various noise profile estimation functions around.

dipy/denoise/denspeed.pyx
grlee77: I see that the term in the exponential as defined here (and in the reference you linked) differ from that defined in Eq. 2 of the paper by Coupe et. al.  (http://www.ncbi.nlm.nih.gov/pubmed/18390341).  Namely, both the numerator and denominator of the `exp` exponent are squared in Eq.2 relative to the above.
I think the original Buades paper definition matches the Coupe one.  

This will affect the degree of smoothing as `(a / b) != (sqrt(a) / sqrt(b))` in most cases.  When the numerator of the `exp` term is large relative to `h`, the weights from Coupe should fall off faster than the ones as defined above giving a sharper image.

PR #1019, opened 2016-03-24
MRG: fix heavy dependency check; no numpy for setup

matthew-brett:
Fix a bug in dependency checking for heavy dependencies.

Make numpy an install dependency using extension builder that pulls in
include directories as late as possible.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,10 +56,12 @@ matrix:
     - python: 2.7
       env:
         - INSTALL_TYPE=wheel
+        # Dependency checking should get all needed dependencies
+        - DEPENDS=""
     - python: 2.7
       env:
         - INSTALL_TYPE=requirements
-        - DEPENDS=
+        - DEPENDS=""
 before_install:
     - source tools/travis_tools.sh
     - virtualenv $VENV_ARGS venv
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,9 +55,12 @@ matrix:
         - INSTALL_TYPE=sdist
     - python: 2.7
       env:
-        - INSTALL_TYPE=wheel
+        - INSTALL_TYPE=pip
         # Dependency checking should get all needed dependencies
         - DEPENDS=""
+    - python: 2.7
+      env:
+        - INSTALL_TYPE=wheel
     - python: 2.7
       env:
         - INSTALL_TYPE=requirements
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -56,6 +56,8 @@ matrix:
     - python: 2.7
       env:
         - INSTALL_TYPE=wheel
+        # Dependency checking should get all needed dependencies
+        - DEPENDS=
```
arokem: Is this supposed to undo the setting on L14? I am not sure that actually takes on Travis. It seems to still get the global setting somehow: https://travis-ci.org/nipy/dipy/jobs/118363119#L132, and I couldn't find any evidence that it gets unset.

matthew-brett: You're right - that's annoying.   I'm maxed out on travis-ci at the moment - would you mind trying a branch where the lines are `DEPENDS=""` to see if that takes?

matthew-brett:
I copied the small nibabel function into dipy to make it easier to install.   Works now I think.

arokem:
We had a little hiatus in CI. Would you mind doing a `git commit --amend` => `git push -f <remote> <branch>`, so that Travis can give this a go?

matthew-brett:
Done - thanks for the reminder.

setup.py
```diff
@@ -117,35 +103,32 @@
...
+build_ext, need_cython = cyproc_exts(EXTS,
+                                     info.CYTHON_MIN_VERSION,
+                                     'pyx-stamps')
+# Add openmp flags if they work
+simple_test_c = """int main(int argc, char** argv) { return(0); }"""
+omp_test_c = """#include <omp.h>
+int main(int argc, char** argv) { return(0); }"""
+extbuilder = add_flag_checking(
+    build_ext, [
+        [['/arch:SSE2'], [], simple_test_c, 'USING_VC_SSE2'],
```
arokem: This should rid us of all the SSE2 related warnings?

version_helpers.py
```diff
@@ -0,0 +1,72 @@
+''' Distutils / setuptools helpers for versioning
+
+This code started life in the nibabel package as nisexts/sexts.py
```
arokem: What's the reason not to get it from nibabel? We need to have that anyway, no?

matthew-brett: Honestly I would recommend getting rid of this code now, and preferring `versioneer` which gives version numbers from the git tag.

But - copying this function allows dipy to install without doing a previous and temporary install both of nibabel and numpy (because nibabel depends on numpy).   With the copy we can let pip sort out the nibabel and numpy dependencies by a standard install.

Garyfallidis:
There is a pending question to @matthew-brett from @arokem  about SSE and warnings. Please do answer this as soon as possible and I will happily merge this PR.

setup.py
matthew-brett: I'm afraid I didn't notice this one.  No, the flag checking logic hasn't changed, this is just an indentation difference.

Garyfallidis:
Thanks. This is what I thought too.

Garyfallidis:
Let's figure out the warnings in another PR. But good point @arokem 
I am moving forward with the merge.

PR #1021, opened 2016-03-26
Added warning for VTK not installed

sahmed95:
Gives a warning for VTK not installed

### DIFF
--- a/dipy/viz/fvtk.py
+++ b/dipy/viz/fvtk.py
@@ -18,7 +18,7 @@ For more information on VTK there many neat examples in
 http://www.vtk.org/Wiki/VTK/Tutorials/External_Tutorials
 '''
 from __future__ import division, print_function, absolute_import
-from warnings import warn, simplefilter
+from warnings import warn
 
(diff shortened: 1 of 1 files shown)


sahmed95:
Shouldn't raise an error like in #1017. This fixes #1016 and raises a Warning in VTK is not installed. Otherwise and fvtk.ren() gives no "ren" object found error which doesn't tell the user that VTK is not installed. 

@Garyfallidis @arokem Initially, I used an ImportWarning but since it is by default suppressed, I had to used filterwarning which was failing a build case in Python3. Hence I have just used a UserWarning here. I hope this is the correct way to go.

arokem:
This looks fine to me. You might want to add a bit more information here, such as: 

"Optional package VTK is not installed. Some dipy.viz functions might not work"

Garyfallidis:
Hi @sahmed95 and @arokem now with vtk 7 supporting python 3, vtk will not be optional for much longer for dipy. Please replace the message with a simple "Python VTK is not installed". Thank you.

PR #1027, opened 2016-04-05
(DO NOT MERGE THIS PR) NF: DKI microstructural model

RafaelNH:
Hello world! In this pull request, I am implementing the DKI based microstructural model which was suggested by ([Fieremans et al., 2011](http://www.ncbi.nlm.nih.gov/pubmed/21699989)).

I am started this fresh PR from the work done on the last weeks of my participation at last years Google summer of code ([GSoC2015 blog](http://gsoc2015dipydki.blogspot.co.uk/)). At the moment, the axonal water fraction is already implemented. The next step is to estimate the intra and extra-axonal diffusion tensors.

Hope you enjoy!


dipy/reconst/dki.py
```diff
@@ -10,8 +11,9 @@
 
 from dipy.reconst.utils import dki_design_matrix as design_matrix
 from dipy.utils.six.moves import range
-from ..core.onetime import auto_attr
 from .base import ReconstModel
+from dipy.core.geometry import (sphere2cart, cart2sphere)
+from .recspeed import local_maxima
```
arokem: Use absolute imports, please.

dipy/reconst/dki.py
```diff
@@ -10,8 +11,9 @@
 
 from dipy.reconst.utils import dki_design_matrix as design_matrix
 from dipy.utils.six.moves import range
-from ..core.onetime import auto_attr
 from .base import ReconstModel
```
arokem: While you are at it, you can also change this one into an absolute import

dipy/reconst/dki.py
```diff
@@ -359,6 +361,75 @@ def _F2m(a, b, c):
     return F2
 
 
+def _directional_kurtosis(dt, MD, kt, V, min_diffusivity=0, min_kurtosis=-1):
```
arokem: Might this be useful as a public function, rather than a private helper function?

dipy/reconst/dki.py
```diff
@@ -444,75 +515,6 @@ def apparent_kurtosis_coef(dki_params, sphere, min_diffusivity=0,
     return AKC.reshape((outshape + (len(V),)))
 
 
-def _directional_kurtosis(dt, MD, kt, V, min_diffusivity=0, min_kurtosis=-1):
```
arokem: Oh - sorry - just realized this function was just moved up :-)

dipy/reconst/dki.py
```diff
@@ -889,6 +891,169 @@ def axial_kurtosis(dki_params, min_kurtosis=0, max_kurtosis=3):
...
+            k_dir = np.array([sphere2cart(1., ang[0], ang[1])])
+            k_val = _directional_kurtosis(dt, MD, kt, k_dir)
+            if k_val > max_value:
+                max_value = k_val
+                max_direction = k_dir
+
+    return max_value, max_direction
+
+
+def single_fiber_model(dki_params, sphere, mask=None, gtol=1e-5):
```
arokem: I think `axonal_water_fraction` is a more appropriate name, given the output.

dipy/reconst/dki.py
```diff
@@ -889,6 +891,169 @@ def axial_kurtosis(dki_params, min_kurtosis=0, max_kurtosis=3):
...
+    AWF : ndarray (x, y, z, 27) or (n, 27)
+        Axonal Water Fraction
+    Da : ndarray (x, y, z, 27) or (n, 27)
+        Axonal Diffusivity
+    ADe : ndarray (x, y, z, 27) or (n, 27)
+        Axial Diffusivity of extra-cellular compartment
+    RDe : ndarray (x, y, z, 27) or (n, 27)
+        Radial Diffusivity of extra-cellular compartment
+    tort : ndarray (x, y, z, 27) or (n, 27)
+        Tortuosity
```
arokem: I see. You want this to eventually return all the statistics?

dipy/reconst/dki.py
```diff
@@ -889,6 +891,169 @@ def axial_kurtosis(dki_params, min_kurtosis=0, max_kurtosis=3):
...
+    # select voxels where to find fiber directions
+    if mask is None:
+        mask = np.ones(shape, dtype='bool')
+    else:
+        if mask.shape != shape:
+            raise ValueError("Mask is not the same shape as data.")
+
+    evals, evecs, kt = split_dki_param(dki_params)
+    # select non-zero voxels
+    # (when #677 is merged replace this code lines by function _positive_evals)
```
arokem: Is this still relevant? It seems that #677 was merged a while back

dipy/reconst/tests/test_dki.py
```diff
@@ -561,3 +562,109 @@ def test_dki_errors():
...
+                                      fractions=frac, snr=None)
+    # DKI fit
+    dkiM = dki.DiffusionKurtosisModel(gtab_2s, fit_method="WLS")
+    dkiF = dkiM.fit(signal)
+
+    # Single fiber model
+    sphere = get_sphere('symmetric724')
+    AWF = dki.single_fiber_model(dkiF.model_params, sphere, mask=None,
+                                 gtol=1e-5)
+    assert_almost_equal(AWF, fie)
```
arokem: Does this work well with values of fie far from 0.5?

dipy/reconst/dki.py
```diff
@@ -359,6 +361,75 @@ def _F2m(a, b, c):
     return F2
 
 
+def _directional_kurtosis(dt, MD, kt, V, min_diffusivity=0, min_kurtosis=-1):
+    r""" Helper function that calculate the apparent kurtosis coefficient (AKC)
```
arokem: "calculate" => "calculates"

dipy/reconst/dki.py
```diff
@@ -359,6 +361,75 @@ def _F2m(a, b, c):
...
+    V : array (g, 3)
+        g directions of a Sphere in Cartesian coordinates
+    min_diffusivity : float (optional)
+        Because negative eigenvalues are not physical and small eigenvalues
+        cause quite a lot of noise in diffusion based metrics, diffusivity
+        values smaller than `min_diffusivity` are replaced with
+        `min_diffusivity`. defaut = 0
+    min_kurtosis : float (optional)
+        Because high amplitude negative values of kurtosis are not physicaly
+        and biologicaly pluasible, and these causes huge artefacts in kurtosis
```
arokem: Typo in "plausible". 

"causes" => "cause". 

"high amplitude" => "high-amplitude"

I would also drop the word "huge" (https://twitter.com/search?q=%23yuuge :-D)

dipy/reconst/dki.py
```diff
@@ -359,6 +361,75 @@ def _F2m(a, b, c):
...
+        based measures, directional kurtosis values than `min_kurtosis` are
```
arokem: "kurtosis based" => "kurtosis-based"

arokem: There's a word missing here? 
"...directional kurtosis values *smaller* than..." ?

dipy/reconst/dki.py
```diff
@@ -889,6 +891,169 @@ def axial_kurtosis(dki_params, min_kurtosis=0, max_kurtosis=3):
     return AK.reshape(outshape)
 
 
+
+
+def _kt_maxima_converge(ang, dt, MD, kt):
+    """ Helper function that computes the negate of the directional kurtosis
```
arokem: With help from a English native speaking mathematician: I think the word "negate" should be "inverse". I know, that's not how it works with matrices, but apparently that's how it works with numbers.

dipy/reconst/dki.py
```diff
@@ -889,6 +891,169 @@ def axial_kurtosis(dki_params, min_kurtosis=0, max_kurtosis=3):
...
+        elements of the diffusion tensor of the voxel.
+    MD : float
+        mean diffusivity of the voxel
+    kt : array (15,)
+        elements of the kurtosis tensor of the voxel.
+
+    Returns
+    -------
+    neg_kt : float
+        The negated values of the directional kurtosis for the given direction.
```
arokem: Same here

arokem: Apparently "additive inverse", "opposite" and "negation" will also do: https://en.wikipedia.org/wiki/Additive_inverse

dipy/reconst/dki.py
```diff
@@ -889,6 +891,169 @@ def axial_kurtosis(dki_params, min_kurtosis=0, max_kurtosis=3):
...
+    Notes
+    -----
+    This function is used to refine the kurtosis maxima estimate
+    
+    See also
+    --------
+    dipy.reconst.dki.kurtosis_maxima
+    """
+    n = np.array([sphere2cart(1, ang[0], ang[1])])
+    return - _directional_kurtosis(dt, MD, kt, n)
```
arokem: Could you write this out as `-1 * ...`? That negative sign is small and easy to miss.

dipy/reconst/dki.py
```diff
@@ -889,6 +891,169 @@ def axial_kurtosis(dki_params, min_kurtosis=0, max_kurtosis=3):
...
+
+
+def kurtosis_maxima(dt, MD, kt, sphere, gtol=1e-5):
```
arokem: This is singular, right? Maybe `kurtosis_maximum`?

dipy/reconst/dki.py
```diff
@@ -889,6 +891,169 @@ def axial_kurtosis(dki_params, min_kurtosis=0, max_kurtosis=3):
...
+        mean diffusivity of the voxel
+    kt : array (15,)
+        elements of the kurtosis tensor of the voxel.
+    sphere : Sphere class instance, optional
+        The sphere providing sample directions for the initial search of the
+        maxima value of kurtosis.
+    gtol : float, optional
+        This input is to refine kurtosis maxima under the precision of the
+        directions sampled on the sphere class instance. The gradient of the
+        convergence procedure must be less than gtol before succesful
```
arokem: "succesful" => "successful"

dipy/reconst/dki.py
```diff
@@ -889,6 +891,169 @@ def axial_kurtosis(dki_params, min_kurtosis=0, max_kurtosis=3):
...
+        termination. If gtol is None, fiber direction is directly taken from 
+        the initial sampled directions of the given sphere object
+    
+    Returns
+    --------
+    max_value : float
+        kurtosis tensor maxima value
+    max_dir : array (3,)
+        Cartesian coordinates of the direction of the maxima kurtosis value
```
arokem: "maxima" => "maximal"

dipy/reconst/dki.py
```diff
@@ -889,6 +891,169 @@ def axial_kurtosis(dki_params, min_kurtosis=0, max_kurtosis=3):
...
+    dki_params : ndarray (x, y, z, 27) or (n, 27)
+        All parameters estimated from the diffusion kurtosis model.
+        Parameters are ordered as follow:
+            1) Three diffusion tensor's eingenvalues
+            2) Three lines of the eigenvector matrix each containing the first,
+               second and third coordinates of the eigenvector
+            3) Fifteen elements of the kurtosis tensor
+    sphere : Sphere class instance, optional
+        The sphere providing sample directions for the initial search of the
+        maxima value of kurtosis.
```
arokem: "maxima" => maximal

arokem:
This is great. We need to think what API we eventually want here. One idea I had is that we implement a `Model` and `Fit` object that inherit from DKI, the model `fit` method does this additional fitting step, and then the `Fit` object holds a few more parameters. What do you think about that?

arokem:
Question: I just noticed that in some places, min_kurtosis is set per-default to 0 (e.g. here: https://github.com/nipy/dipy/blob/master/dipy/reconst/dki.py#L827) and in some places it is set to -1 (e.g. here: https://github.com/nipy/dipy/blob/master/dipy/reconst/dki.py#L363). Would it make sense to have the same default everywhere? What's a more reasonable default for these? Would kurtosis ever go below that of free water (0?)?

dipy/reconst/dki.py
RafaelNH: Is it real necessary to convert _direction_kurtosis to a public function? Users can just use the more generalized function version apparent_kurtosis_coef which deals with both single and multi-voxel data.

RafaelNH:
Hi @arokem! According to  [Barmpoutis and Zhuo (2011)](http://ieeexplore.ieee.org/xpl/articleDetails.jsp?arnumber=5872402) there is a theoretical kurtosis of -3/7 minima for the case of water in spherical pores.

dipy/reconst/dki.py
```diff
@@ -455,7 +457,8 @@ def apparent_kurtosis_coef(dki_params, sphere, min_diffusivity=0,
         Because high amplitude negative values of kurtosis are not physicaly
         and biologicaly pluasible, and these causes huge artefacts in kurtosis
         based measures, directional kurtosis values than `min_kurtosis` are
-        replaced with `min_kurtosis`. defaut = -1
+        replaced with `min_kurtosis`. defaut = -3./7 (theoretical kurtosis
+        limit for regions that consist of water confined to spherical pores)
```
arokem: Would you mind adding the Barmpoutis reference here as well?

arokem:
I think you can handle the remaining test failures by reducing test sensitivity (use `assert_almost_equal` with the `decimal` key-word argument set to 5 or so).

RafaelNH:
Regarding the comment of what API we should use - I have been thinking and exploring ideas and at the moment this is that I am thinking to do:

Add a function named single_fiber_model in DiffusionKurtosisFit Class that will call an KurtosisSingleFiberModel class that will add the following parameters:

- axonal_water_fraction (according to Fieremans article)
- intracellular_params (same format of DTI model_params)
- extracellular_params (same format of DTI model_params)
- intracellular_diffusivity (according to Fieremans article)
- extracellular_rd (according to Fieremans article)
- extracellular_ad (according to Fieremans article)
- tortuosity (according to Fieremans article)

This will be the workflow to use kurtosis single fiber model:

1) dkiM = dki.DiffusionKurtosisModel(gtab)
2) dkiF = dkiM.fit(signal)
3) dkiWM = dkiF.single_fiber_model(sphere)

arokem:
It's a bit of added complexity in terms of the user, but is sounds like a good way to go to me, as long as we document properly and provide a good example.

dipy/reconst/dki.py
```diff
@@ -359,9 +361,80 @@ def _F2m(a, b, c):
...
+        mean diffusivity of the voxel
+    kt : array (15,)
+        elements of the kurtosis tensor of the voxel.
+    V : array (g, 3)
+        g directions of a Sphere in Cartesian coordinates
+    min_diffusivity : float (optional)
+        Because negative eigenvalues are not physical and small eigenvalues
+        cause quite a lot of noise in diffusion based metrics, diffusivity
+        values smaller than `min_diffusivity` are replaced with
+        `min_diffusivity`. defaut = 0
```
jhlegarreta: **defaut** should be changed by **default**

dipy/reconst/dki.py
```diff
@@ -889,6 +922,243 @@ def axial_kurtosis(dki_params, min_kurtosis=0, max_kurtosis=3):
...
+
+
+def axonal_water_fraction(dki_params, sphere, mask=None, gtol=1e-5):
+    """ Computes the DKI based axonal water fraction [1]_.
+
+    Parameters
+    ----------
+    dki_params : ndarray (x, y, z, 27) or (n, 27)
+        All parameters estimated from the diffusion kurtosis model.
+        Parameters are ordered as follow:
```
jhlegarreta: **as follow** should be changed to **as follows**, right?

jhlegarreta:
What do you mean by "defining minimun diffusion **aloud**"?
I'd use capital letters when referring to Fractional Anisotropy or its acronym **FA**. I think you refer to FA in the documentation of the methods **_ols_iter_** and **_wls_iter_**

dipy/reconst/dki.py
```diff
@@ -889,6 +922,243 @@ def axial_kurtosis(dki_params, min_kurtosis=0, max_kurtosis=3):
...
+            k_dir = np.array([sphere2cart(1., ang[0], ang[1])])
+            k_val = _directional_kurtosis(dt, MD, kt, k_dir)
+            if k_val > max_value:
+                max_value = k_val
+                max_direction = k_dir
+
+    return max_value, max_direction
+
+
+def axonal_water_fraction(dki_params, sphere, mask=None, gtol=1e-5):
```
jhlegarreta: Shall we extend/improve the computation for the case where the _Da,i=0_ assumption is not met?

Shall we create first another method like **exact_axonal_water_fraction**, then see if it works as expected, and then extend the existing with another parameter that will choose between the approximate expression and the exact depending on it?

jhlegarreta: @Rafel, @Ariel,
how can I get the axis directions of the reference frame to get the diffusion coefficientes and kurtoses?
I guess that the kurtoses are obtained from dki_params, right?
Concerning Da,min should we create another kurtosis_minimum function to compute it?

arokem: Hi @jhlegarreta - apologies that it's take so long to answer this. Have you taken a look at the DKI documentation example? http://nipy.org/dipy/examples_built/reconst_dki.html#example-reconst-dki. I might be misunderstanding your question, but I believe the answers are in there.

RafaelNH: Hi @jhlegarreta - sorry for my late reply - I have been a little bit tied with work.

Yup, the objective is really that, improve the axonal water fraction when the Da,i=0 is not met. I agree that this should be an optional parameter of the axonal_water_fraction, and also agree that you can start implementing this in a helper function that could be called by the other functions. Call it however something like axonal_water_fraction_correction - I am afraid that no diffusion MRI technique is able to give exact axonal water fractions =P.

The diffusion tensor evals are already the diffusion coefficients along axis directions of the reference frame. To get the kurtosis one, you can just use Wrotate_element(kt, indi, indj, indk, indl, B), where kt is the elements of the kurtosis tensor, B is the eigenvalues of the diffusion tensor, indi=indj=indk=indl=1 to obtain K1, indi=indj=indk=indl=2 to obtain K2 and indi=indj=indk=indl=3 to obtain K3.

The original kurtosis elements are indeed saved in dki_params together with the diffusion tensor parameters. To split diffusion and kurtois parameters use split_dki_param(dki_params).

I am afraid that to compute Da,min you have to follow a different approach than kurtosis_maxima since you need Da,min from a specified ROI while kurtosis_mixima finds the lower direction inside the voxel.

dipy/reconst/dki.py
```diff
@@ -966,7 +1236,7 @@ def __init__(self, gtab, fit_method="OLS", *args, **kwargs):
 
         References
         ----------
```
jhlegarreta: Watch the typo in line 1200:
# lopping for all voxels
# looping for all voxels
or as stated in other parts of the code
# loop over all voxels

dipy/reconst/dki.py
```diff
@@ -966,7 +1236,7 @@ def __init__(self, gtab, fit_method="OLS", *args, **kwargs):
```
jhlegarreta: Watch the typo in line 1200:
`# lopping for all voxels`
`# looping for all voxels`
or as stated in other parts of the code
`# loop over all voxels`

dipy/reconst/dki.py
jhlegarreta: @ariel, @rafaelnh,
sorry for the late reply and thanks for the comments.

I guess the kurtoses should be computed over all evecs. Is that right?
The same applies to the mean diffusivity (\bar{D}). Should we compute it over all evals?
The diffusivity coefficients (D_1, D2, and D_3), are they the ADCs along the axis directions (D_1 = ADC[1], etc.)?

dipy/reconst/dki.py
```diff
@@ -893,6 +950,158 @@ def axial_kurtosis(dki_params, min_kurtosis=0, max_kurtosis=3):
...
+        elements of the diffusion tensor of the voxel.
+    MD : float
+        mean diffusivity of the voxel
+    kt : array (15,)
+        elements of the kurtosis tensor of the voxel.
+
+    Returns
+    -------
+    neg_kt : float
+        The inverse values of the directional kurtosis for the given direction.
```
arokem: "values" => "value" ? 

(is there more than one?)

dipy/reconst/dki.py
```diff
@@ -893,6 +950,158 @@ def axial_kurtosis(dki_params, min_kurtosis=0, max_kurtosis=3):
...
+
+    Notes
+    -----
+    This function is used to refine the kurtosis maxima estimate
+
+    See also
+    --------
+    dipy.reconst.dki.kurtosis_maximum
```
arokem: To aid sphinx in cross-referencing this 

=> 

    :func:`kurtosis_maximum`

dipy/reconst/dki.py
```diff
@@ -893,6 +950,158 @@ def axial_kurtosis(dki_params, min_kurtosis=0, max_kurtosis=3):
...
+    """
+    n = np.array([sphere2cart(1, ang[0], ang[1])])
+    return -1. * _directional_kurtosis(dt, MD, kt, n)
+
+
+def _voxel_kurtosis_maximum(dt, MD, kt, sphere, gtol=1e-5):
+    """ Computes the maxima value of a single voxel kurtosis tensor
```
arokem: "the maxima" => "the maximum"

dipy/reconst/dki.py
```diff
@@ -893,6 +950,158 @@ def axial_kurtosis(dki_params, min_kurtosis=0, max_kurtosis=3):
...
+
+    # refine maxima direction
+    if gtol is not None:
+        for p in range(n):
+            r, theta, phi = cart2sphere(max_dir[p, 0], max_dir[p, 1],
+                                        max_dir[p, 2])
+            ang = np.array([theta, phi])
+            ang[:] = opt.fmin_bfgs(_kt_maxima_converge, ang, args=(dt, MD, kt),
+                                   gtol=gtol, disp=False, retall=False)
+            k_dir = np.array([sphere2cart(1., ang[0], ang[1])])
```
arokem: Can we somehow avoid the conversion 

cart => sphere => cart 

that is happening here?

for example, unless I am missing something it looks like `_kt_maxima_converge` would be even simpler, if it took the cartesian coordinates as input.

dipy/reconst/dki.py
```diff
@@ -893,6 +950,158 @@ def axial_kurtosis(dki_params, min_kurtosis=0, max_kurtosis=3):
...
+            k_val = _directional_kurtosis(dt, MD, kt, k_dir)
+            if k_val > max_value:
+                max_value = k_val
+                max_direction = k_dir
+
+    return max_value, max_direction
+
+
+def kurtosis_maximum(dki_params, sphere='symmetric362', gtol=1e-5,
```
arokem: It occurs to me that there are places where this is spelled `maxima` and others where this is spelled `maximum`. How about simplifying and unifying, by spelling all of these as `max` ? For example, this function would be renamed `kurtosis_max`.

dipy/reconst/dki.py
arokem: Second this comment

dipy/reconst/dki.py
```diff
@@ -893,6 +950,158 @@ def axial_kurtosis(dki_params, min_kurtosis=0, max_kurtosis=3):
...
+    evals, evecs, kt = split_dki_param(dki_params)
+
+    # select non-zero voxels
+    pos_evals = _positive_evals(evals[..., 0], evals[..., 1], evals[..., 2])
+    mask = np.logical_and(mask, pos_evals)
+
+    kt_max = np.zeros(mask.shape)
+
+    for idx in ndindex(shape):
+        if not mask[idx]:
```
arokem: This could be simplified as: 

    for idx in np.where(mask):

That way, you don't have to check whether the mask has a value in each iteration (and potentially loop over plenty of voxels you don't care about, doing that).

dipy/reconst/dki.py
```diff
@@ -1003,6 +1212,15 @@ def __init__(self, gtab, fit_method="OLS", *args, **kwargs):
             e_s += " positive."
             raise ValueError(e_s)
 
+        # Check if at least three b-values are given
```
arokem: This section should be superseded by #1221

dipy/reconst/dki.py
```diff
@@ -1266,23 +1503,48 @@ def rk(self, min_kurtosis=0, max_kurtosis=3):
...
+        .. [2] Barmpoutis, A., & Zhuo, J., 2011. Diffusion kurtosis imaging:
+               Robust estimation from DW-MRI using homogeneous polynomials.
+               Proceedings of the 8th {IEEE} International Symposium on
+               Biomedical Imaging: From Nano to Macro, ISBI 2011, 262-265.
+               doi: 10.1109/ISBI.2011.5872402
         """
         return radial_kurtosis(self.model_params, min_kurtosis, max_kurtosis)
 
+    def kmax(self, sphere='symmetric362', gtol=1e-5, mask=None):
+        r""" Computes the maxima value of a single voxel kurtosis tensor
```
arokem: This is not for a single voxel, is it? 

Also, there's got to be a way to propagate the mask from the fitting procedure, right? For example, you can look for voxels that have non-zero values of the parameters?

dipy/reconst/dki_micro.py
```diff
@@ -0,0 +1,489 @@
...
+    ----------
+    params : ndarray (x, y, z, 40) or (n, 40)
+    All parameters estimated from the diffusion kurtosis microstructural model.
+        Parameters are ordered as follows:
+            1) Three diffusion tensor's eigenvalues
+            2) Three lines of the eigenvector matrix each containing the
+               first, second and third coordinates of the eigenvector
+            3) Fifteen elements of the kurtosis tensor
+            4) Six elements of the hindered diffusion tensor
+            5) Six elements of the restricted diffusion tensor
```
arokem: The function `diffusion_components` seems to return 12 parameters for each of these (eigenvalues + eigenvectors). Are those inconsistent with each other? Can we just store 3 eigenvalues for each of these instead of the six tensor elements?

dipy/reconst/dki_micro.py
```diff
@@ -0,0 +1,489 @@
...
+
+    # Prepare parameters
+    f = params[..., 39]
+    adce = params[..., 27:33]
+    adce = params[..., 33:39]
+
+    # Process pred_sig for all data voxels
+    index = ndindex(evals.shape[:-1])
+    for v in index:
+        if mask[v]:
```
arokem: Same comment as above: use `for v in np.where(mask)`

dipy/reconst/dki_micro.py
```diff
@@ -0,0 +1,489 @@
...
+           fit_method. See dki.ols_fit_dki, dki.wls_fit_dki for details
+
+        Notes
+        -----
+        1) Since this model is an extension of DKI, class instance is defined
+           as subclass DiffusionKurtosisModel from dki.py
+        2) The first step of the DKI based microstructural model requires
+           diffusion tensor and kurtosis tensor fit. This fit is performed
+           using the DKI model solution specified by users by input parameter
+           fit_method.
```
arokem: I don't think that we need these notes in the documentation. They both seem like implementation details, rather than an explanation of functionality or API.

dipy/reconst/dki_micro.py
```diff
@@ -0,0 +1,489 @@
...
+        References
+        ----------
+        .. [1] Fieremans, E., Jensen, J.H., Helpern, J.A., 2011. White Matter
+               Characterization with Diffusion Kurtosis Imaging. Neuroimage
+               58(1): 177-188. doi:10.1016/j.neuroimage.2011.06.006
+        """
+        DiffusionKurtosisModel.__init__(self, gtab, fit_method="WLS", *args,
+                                        **kwargs)
+
+        ReconstModel.__init__(self, gtab)
```
arokem: Is this necessary? I think that this gets initialized through the inheritance hierarchy through DKI => DTI => ReconstModel

dipy/reconst/dki_micro.py
```diff
@@ -0,0 +1,489 @@
...
+
+    def fit(self, data, mask=None, sphere, gtol=1e-5, awf_only=False):
+        """ Fit method of the Diffusion Kurtosis Microstructural Model
```
arokem: "Fit method of the Diffusion ..." => "Fit the Diffusion..."

dipy/reconst/dki_micro.py
```diff
@@ -0,0 +1,489 @@
...
+
+        Parameters
+        ----------
+        data : array
+            The measured signal from one voxel.
```
arokem: I think this is incorrect. Looks like it generalizes to more than one voxel!

dipy/reconst/dki_micro.py
```diff
@@ -0,0 +1,489 @@
...
+        model_params : ndarray (x, y, z, 40) or (n, 40)
+            All parameters estimated from the diffusion kurtosis
+            microstructural model.
+            Parameters are ordered as follows:
+                1) Three diffusion tensor's eigenvalues
+                2) Three lines of the eigenvector matrix each containing the
+                   first, second and third coordinates of the eigenvector
+                3) Fifteen elements of the kurtosis tensor
+                4) Six elements of the hindered diffusion tensor
+                5) Six elements of the restricted diffusion tensor
```
arokem: Again - do  we really need all six of these?

dipy/reconst/tests/test_dki_micro.py
```diff
@@ -0,0 +1,217 @@
...
+    ADI = np.array([[[1e-3, 1.3e-3], [0.8e-3, 1e-3]],
+                    [[0.9e-3, 0.99e-3], [0.89e-3, 1.1e-3]]])
+    ADE = np.array([[[2.2e-3, 2.3e-3], [2.8e-3, 2.1e-3]],
+                    [[1.9e-3, 2.5e-3], [1.89e-3, 2.1e-3]]])
+    Tor = np.array([[[2.6, 2.4], [2.8, 2.1]],
+                    [[2.9, 2.5], [2.7, 2.3]]])
+    RDE = ADE / Tor
+
+    # prepare simulation:
+    DWIsim = np.zeros((2., 2., 2., gtab_2s.bvals.size))
```
arokem: These all need to be integers! 

    DWIsim = np.zeros((2, 2, 2, gtab_2s.bvals.size))

dipy/reconst/dki.py
RafaelNH: Humm! This didn't work :/. 
For the test example in which data is a 2x2x2 matrix, np.where gives 

np.where(mask)
(array([0, 0, 0, 0, 1, 1, 1, 1], dtype=int64),
 array([0, 0, 1, 1, 0, 0, 1, 1], dtype=int64),
 array([0, 1, 0, 1, 0, 1, 0, 1], dtype=int64) 

But what we want is:

(array[0, 0, 0], dtype=int64),
(array[1, 0, 0], dtype=int64),
...
(array[1, 1, 1], dtype=int64)

arokem:
How about `np.array(np.where(mask)).T`?

I hope that doesn't blow up too much in memory and doesn't take too long to
create

dipy/reconst/dki.py
RafaelNH: Hi @arokem! Yup, this needs to be converted to spherical coordinates because the convergence algorithm has to search in the angular coordinates space. If cartesian vectors are given in function _kt_maxima_convergence the search will not be limited to the unit sphere domain.

RafaelNH:
The problem with np.array(np.where(mask)).T is that it produces array indexes, but what we want is tuple indexes.

for idx in np.array(np.where(mask)).T:
    print(idx)
[0 0 0]
[0 0 1]
[0 1 0]
[0 1 1]
[1 0 0]
[1 0 1]
[1 1 0]
[1 1 1]

We want:

for idx in ndindex(mask.shape):
    print(idx)
(0, 0, 0)
(0, 0, 1)
(0, 1, 0)
(0, 1, 1)
(1, 0, 0)
(1, 0, 1)
(1, 1, 0)
(1, 1, 1)

dipy/reconst/dki_micro.py
```diff
@@ -0,0 +1,489 @@
...
+        Parameters of the extra-cellular diffusion tensor.
+    IDT : ndarray (x, y, z, 12) or (n, 12)
+        Parameters of the intra-cellular diffusion tensor.
+
+    Notes
+    -----
+    The parameters of both extra-cellular and intra-cellular diffusion tensors
+    are order as follow:
+        1) Three diffusion tensor's eingenvalues
+        2) Three lines of the eigenvector matrix each containing the first,
```
RafaelNH: Exactly - we only need the eigenvalues of the diffusion tensor to compute the standard diffusion and kurtosis measures that we imported from class inheritance. That is also why I am not decomposing the hindered and restricted diffusion tensors.

dipy/reconst/dki_micro.py
RafaelNH: I just noticed that some documentation was still describing an earlier version of the code. In the early version of the code, I was returning the decomposition of the hindered and restricted diffusions tensors. However, now the six independent lower elements of the diffusion tensor are always returned (I am about to update the PR with the updated documentation).

I didn't store only the 3 eigenvalues of the restricted and hindered diffusion tensors because we need the six elements for the model prediction functions.

dipy/reconst/dki_micro.py
RafaelNH: Oh yes! Sorry, this line is here because I start implementing this through inheritance of  ReconstModel before going through DKI inheritance! I complete forgot to remove this line!

dipy/reconst/dki_micro.py
RafaelNH: see comments above

RafaelNH:
DKI based microstructural model was already introduced in its own class object! All code in dki_micro.py seem to be working. Test covering all its lines seem to pass and everything was implemented in PEP8's standards! The only thing that is missing the example of usage (which will be added into reconst_dki.py). Please let me know if anyone as further comments that I should address!

dipy/reconst/dki.py
```diff
@@ -1360,11 +1702,11 @@ def ols_fit_dki(design_matrix, data):
     data_flat = data.reshape((-1, data.shape[-1]))
     dki_params = np.empty((len(data_flat), 27))
 
-    # inverting design matrix and defining minimun diffusion aloud
+    # inverting design matrix and defining minimun diffusion
```
arokem: "minimun" => "minimum"

dipy/reconst/dki.py
```diff
@@ -1465,11 +1807,11 @@ def wls_fit_dki(design_matrix, data):
...
```
arokem: "minimun" => "minimum"

dipy/reconst/tests/test_dki.py
```diff
@@ -1,10 +1,11 @@
-""" Testing DKI """
+""" Testing DKI microstructure """
```
arokem: This seems like the wrong file to be making this change.

dipy/reconst/dki_micro.py
```diff
@@ -0,0 +1,489 @@
...
+        convergence procedure must be less than gtol before successful
+        termination. If gtol is None, fiber direction is directly taken from
+        the initial sampled directions of the given sphere object
+    mask : ndarray
+        A boolean array used to mark the coordinates in the data that should be
+        analyzed that has the shape dki_params.shape[:-1]
+
+    Returns
+    --------
+    AWF : ndarray (x, y, z) or (n)
```
arokem: This was changed everywhere except right here in the docstring. Please change docstring accordingly.

dipy/reconst/dki_micro.py
```diff
@@ -0,0 +1,507 @@
...
+    Returns
+    --------
+    edt : ndarray (x, y, z, 6) or (n, 6)
+        Parameters of the hindered diffusion tensor.
+    idt : ndarray (x, y, z, 6) or (n, 6)
+        Parameters of the restricted diffusion tensor.
+
+    Notes
+    -----
+    The parameters of both hindered and restricted diffusion tensors are order
```
arokem: I think this note is inconsistent with the rest of the docstring. It seems to suggest that we return these tensors as 12 parameters each (three eigenvalues + three eigenvectors).

dipy/reconst/dki_micro.py
```diff
@@ -0,0 +1,507 @@
...
+    --------
+    S : (..., N) ndarray
+        Simulated signal based on the DKI microstructure model
+
+    Notes
+    -----
+    The predicted signal is given by:
+    $S(\theta, b) = S_0 * [f * e^{-b ADC_{r}} + (1-f) * e^{-b ADC_{h}]$, where
+    $ ADC_{r} and ADC_{h} are the apparent diffusion coefficients of the
+    diffusion hindered and restricted compartment for a given direction
```
arokem: Maybe this is the place to mention that these are referred to as "intracellular" and "extracellular" in the Fieremans paper.

dipy/reconst/dki_micro.py
```diff
@@ -0,0 +1,507 @@
...
+        """
+        if mask is None:
+            # Flatten it to 2D either way:
+            data_in_mask = np.reshape(data, (-1, data.shape[-1]))
+        else:
+            # Check for valid shape of the mask
+            if mask.shape != data.shape[:-1]:
+                raise ValueError("Mask is not the same shape as data.")
+            mask = np.array(mask, dtype=bool, copy=False)
+            data_in_mask = np.reshape(data[mask], (-1, data.shape[-1]))
```
arokem: It just occurred to me that `arr[None]` indexes into the entire array, so we can get rid of this whole `if...else` construct. I will make a PR with that (and another couple of small things here).

dipy/reconst/dki_micro.py
```diff
@@ -0,0 +1,507 @@
...
+    @property
+    def awf(self):
+        """ Returns the volume fraction of the restricted diffusion compartment
+        also known as axonal water fraction.
+        """
+        return self.model_params[..., 27]
+
+    @property
+    def restricted_evals(self):
+        """ Returns the eigenvalues of the restricted diffusion compartment.
```
arokem: Put a note here mentioning that this is called "intracellular" (or was it "extracellular"?) in the Fieremans paper.

dipy/reconst/dki_micro.py
```diff
@@ -0,0 +1,507 @@
...
+        """
+        self._is_awfonly()
+        rdt = self.model_params[..., 34:40]
+        evals, evecs = decompose_tensor(from_lower_triangular(rdt))
+        return evals
+
+    @property
+    def hindered_evals(self):
+        """ Returns the eigenvalues of the restricted diffusion compartment.
```
arokem: Put a note here mentioning that this is called "extracellular" (or was it "intracellular"?) in the Fieremans paper.

dipy/reconst/dki_micro.py
```diff
@@ -0,0 +1,507 @@
...
+    def hindered_rd(self):
+        """ Returns the radial diffusivity of the hindered compartment.
+        """
+        return radial_diffusivity(self.hindered_evals)
+
+    @property
+    def tortuosity(self):
+        """ Returns the tortuosity of the hindered diffusion which is defined
+        by ADe / RDe, where ADe and RDe are the axial and radial diffusivities
+        of the hindered compartment [1]_.
```
arokem: "also referred to as the 'extracellular compartment'"

dipy/reconst/tests/test_dki_micro.py
```diff
@@ -0,0 +1,241 @@
...
+    # Check predict of KurtosisMicrostruturalModel
+    pred = dkiM.predict(dkiF.model_params)
+    assert_array_almost_equal(pred, DWIsim_all_taylor)
+
+    pred = dkiM.predict(dkiF.model_params, S0=100)
+    assert_array_almost_equal(pred, DWIsim_all_taylor * 100)
+
+    # Check predict of KurtosisMicrostruturalFit
+    pred = dkiF.predict(gtab_2s, S0=100)
+    assert_array_almost_equal(pred, DWIsim_all_taylor * 100)
```
arokem: Can we add a prediction with a single voxel's signal (this is multi-voxel, correct?)

dipy/reconst/tests/test_dki_micro.py
```diff
@@ -0,0 +1,241 @@
...
+
+
+def test_dki_micro_awf_only():
+    dkiM = dki_micro.KurtosisMicrostructureModel(gtab_2s)
+    dkiF = dkiM.fit(DWIsim, awf_only=True)
+    awf = dki_micro.axonal_water_fraction(dkiF.model_params)
+    assert_almost_equal(awf, FIE)
```
arokem: Let's test that referring to other properties of the Fit object raises the appropriate error

arokem:
Looks like this needs a rebase on master (so that we incorporate the `check_multi_b` function).

dipy/reconst/tests/test_dki_micro.py
```diff
@@ -0,0 +1,241 @@
...
+    DWIsimc = DWIsim.copy()
+
+    FIEc[0, 0, 0] = 0
+    RDIc[0, 0, 0] = 0
+    ADIc[0, 0, 0] = 0
+    ADEc[0, 0, 0] = 0
+    Torc[0, 0, 0] = 0
+    RDEc[0, 0, 0] = 0
+    DWIsimc[0, 0, 0, :] = 0
+    mask = np.ones((2., 2., 2.))
```
arokem: These need to be integers: 

    np.ones((2, 2, 2))

dipy/reconst/tests/test_dki.py
RafaelNH: ups

dipy/reconst/dki_micro.py
RafaelNH: Cool thanks!

dipy/reconst/tests/test_dki_micro.py
RafaelNH: Hi @arokem! How is the best way to test errors in fit object properties?

I already tried the bellow lines of code, but it didn't work since errors are raised before the function evaluated the property call.

from nose.tools import assert_raises
dkiM = dki_micro.KurtosisMicrostructureModel(gtab_2s)
dkiF = dkiM.fit(DWIsim, awf_only=True)
assert_raises(ValueError, dkiF.hindered_rd)

RafaelNH:
Hello! Here is an update of the things missing in this PR:
- [ ]  Test the errors when the hindered and restricted diffusion propreties are called but awf_only was set to True. @arokem can you help me on this? See my comments above.
- [ ]  Merge @arokem 's PR with code change suggestions (e.g. removing else if conditions in mask)
- [ ] Add example of usage -> I will try to do this tomorrow!

arokem:
This commit implements the removal of redundant `if... else` clauses: https://github.com/RafaelNH/dipy/pull/3/commits/07717e58e2171adb9b20e2c2c147146cca9cfb42

arokem:
Superseded by #1226

PR #1028, opened 2016-04-14
BF: Predict DKI with a volume of S0

arokem:
The first commit introduces a test that demonstrates the bug: if the `S0` input to DKI prediction is provided as a ndarray (which should be possible!), this raises an error. 

The second commit fixes this bug, by separating into two cases: one when `S0` is an ndarray (assumed to have the right shape!), and one when it is a float.


arokem:
@RafaelNH : could you please take a look? Thanks!

RafaelNH:
@arokem - excellent, well spotted! In my opinion this can be merged!

arokem:
Does anyone have any comments here? It's just a small bugfix!

RafaelNH:
As mention above everything looks fine here. If there isn't any objections, I will merge this tomorrow!

PR #1031, opened 2016-04-15
improvements for denoise/denspeed.pyx

sdorkenw:
Speedup for multithreaded implementation
New implementation of nlmeans that outperforms the current implementation especially when no mask is given by removing redundant computations. Currently only as single threaded version -> use this for num_threads == 1 and fall back to old implementation elsewise

### DIFF
--- a/dipy/denoise/denspeed.pyx
+++ b/dipy/denoise/denspeed.pyx
@@ -452,6 +452,26 @@ cdef cnp.npy_intp copy_block_3d(double * dest,
     return 1
 
 
+@cython.wraparound(False)
+@cython.boundscheck(False)
+cdef cnp.npy_intp copy_block_3d_back(double [:, :, ::1] dest,
+                                     cnp.npy_intp min_i,
+                                     cnp.npy_intp min_j,
+                                     cnp.npy_intp min_k,
+                                     double * source,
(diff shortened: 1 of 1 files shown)


Garyfallidis:
Hi @sdorkenw thank you for this PR. What is the actual speedup achieved here? Does the code duplication worth the effort if this works only for 1 thread? Also there are no tests.

sdorkenw:
Hi @Garyfallidis. I plotted the actual speed in dependency of the number of voxels that are masked (= denoised). I also wrote a multiprocessed version of the "new" implementation, but it was not always faster than the original one. I am happy to also share/include that code.

data size: 145, 174, 145
the mask was produced with median_otsu and dpg.gradient_table and then dilated / eroded

speedup
time_comp_conv

arokem:
It looks like a lot of the benefit is probably coming from the first commit here, and not necessarily from the reimplementation of the internals convolution  

Maybe we can just take that (it's a one-line change, that changes the scheduling of `prange`)?

arokem:
Hey @sdorkenw - could we have a separate PR just for that first commit? Seems like that would already give us a substantial speed up!

arokem:
Looks like this was superseded by #1065. Closing.

PR #1039, opened 2016-04-29
Doc - fix definition of real_sph_harm functions

MrBago:
Can someone double check me, but I think this is right now.

### DIFF
--- a/dipy/reconst/shm.py
+++ b/dipy/reconst/shm.py
@@ -203,9 +203,9 @@ def real_sph_harm(m, n, theta, phi):
 
     Where the real harmonic $Y^m_n$ is defined to be:
 
-        Real($Y^m_n$) * sqrt(2) if m > 0
-        $Y^m_n$                 if m == 0
-        Imag($Y^m_n$) * sqrt(2) if m < 0
+        Imag($Y^m_n$) * sqrt(2)     if m > 0
+        $Y^0_n$                     if m = 0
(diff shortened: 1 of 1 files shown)


dipy/reconst/shm.py
```diff
@@ -203,9 +203,9 @@ def real_sph_harm(m, n, theta, phi):
 
     Where the real harmonic $Y^m_n$ is defined to be:
 
-        Real($Y^m_n$) * sqrt(2) if m > 0
-        $Y^m_n$                 if m == 0
-        Imag($Y^m_n$) * sqrt(2) if m < 0
+        Imag($Y^m_n$) * sqrt(2)     if m > 0
+        $Y^m_n$                     if m == 0
```
arokem: I think that's correct. I believe that we are following the same convention as the one outlined in @mdesco et al.'s 2007 paper (http://onlinelibrary.wiley.com/doi/10.1002/mrm.21277/full). See equation 3. I believe you can write this central line as: 

    $Y^0_n$     if m=0 

I don't think that we need to collapse the index here, as they do in that paper.

arokem:
Any more comments here? I'll merge this in a couple of days, if I don't hear from anyone.

dipy/reconst/shm.py
MrBago: I changed the middle line, does this look better?

arokem:
LGTM. Merging.

PR #1041, opened 2016-05-03
NF - Add PMF Threshold to Tractography

gabknight:
This PR adds pmf_threshold parameter to ProbabilisticDirectionGetter, which set to 0 all directions with probability mass function value < pmf_threshold.. It is common to remove low CSD-ODF value (e.g. mrtrix)  from the probability mass function used in probabilistic tractography. Increasing pmf_threshold tends to reduce erroneous streamlines by following 'less noisy' directions of the CSD-ODF.

I set the default to 0.1 as suggested by [Tournier et al. 2012, Girard et al. 2014]. However, this will change probabilistic streamline output of current scripts.

### DIFF
--- a/dipy/direction/probabilistic_direction_getter.py
+++ b/dipy/direction/probabilistic_direction_getter.py
@@ -81,7 +81,7 @@ class ProbabilisticDirectionGetter(PeakDirectionGetter):
 
     """
     @classmethod
-    def from_pmf(klass, pmf, max_angle, sphere, **kwargs):
+    def from_pmf(klass, pmf, max_angle, sphere, pmf_threshold=0, **kwargs):
         """Constructor for making a DirectionGetter from an array of Pmfs
 
         Parameters
@@ -93,6 +93,9 @@ class ProbabilisticDirectionGetter(PeakDirectionGetter):
             direction.
         sphere : Sphere
             The set of directions to be used for tracking.
+        pmf_threshold : float [0., 1.]
+            Used to remove direction from the probability mass function for
+            selecting the tracking direction.
         relative_peak_threshold : float in [0., 1.]
             Used for extracting initial tracking directions. Passed to
             peak_directions.
@@ -114,11 +117,11 @@ class ProbabilisticDirectionGetter(PeakDirectionGetter):
                    "points in sphere.")
             raise ValueError(msg)
         pmf_gen = SimplePmfGen(pmf)
-        return klass(pmf_gen, max_angle, sphere, **kwargs)
+        return klass(pmf_gen, max_angle, sphere, pmf_threshold, **kwargs)
 
     @classmethod
-    def from_shcoeff(klass, shcoeff, max_angle, sphere, basis_type=None,
-                     **kwargs):
+    def from_shcoeff(klass, shcoeff, max_angle, sphere, pmf_threshold=0,
+                     basis_type=None, **kwargs):
         """Probabilistic direction getter from a distribution of directions
         on the sphere.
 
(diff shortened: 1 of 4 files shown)


dipy/direction/probabilistic_direction_getter.py
```diff
@@ -152,21 +158,28 @@ def from_shcoeff(klass, shcoeff, max_angle, sphere, basis_type=None,
...
         ----------
         pmf_gen : PmfGen
-            Used to get probability mass function for choosing tracking
+            Used to get probability mass function for selecting tracking
             directions.
         max_angle : float, [0, 90]
             The maximum allowed angle between incoming direction and new
             direction.
+        mpmf_threshold : float, [0, 1.]
+            The maximum allowed angle between incoming direction and new
```
arokem: This is not the documentation for this parameter, is it?

arokem: Also, I think this shouldn't be here at all. There's no such parameter that I can see.

dipy/tracking/local/tests/test_local_tracking.py
```diff
@@ -260,12 +260,20 @@ def allclose(x, y):
...
     # The first path is not possible if 90 degree turns are excluded
-    dg = ProbabilisticDirectionGetter.from_pmf(pmf, 80, sphere)
+    dg = ProbabilisticDirectionGetter.from_pmf(pmf, 80, sphere,
+                                               pmf_threshold=0.1)
     streamlines = LocalTracking(dg, tc, seeds, np.eye(4), 1.)
 
     for sl in streamlines:
         npt.assert_(np.allclose(sl, expected[1]))
 
+    # The first path is not possible if pmf_threshold>0.4 degree turns are excluded
```
arokem: I think this comment has a few too many words :-)

dipy/tracking/local/tests/test_local_tracking.py
```diff
@@ -318,11 +330,23 @@ def allclose(x, y):
...
-    dg = DeterministicMaximumDirectionGetter.from_pmf(pmf, 80, sphere)
+    dg = DeterministicMaximumDirectionGetter.from_pmf(pmf, 80, sphere,
+                                                      pmf_threshold=0.1)
     streamlines = LocalTracking(dg, tc, seeds, np.eye(4), 1.)
 
     for sl in streamlines:
         npt.assert_(np.allclose(sl, expected[1]))
 
+    # Both path are not possible if 90 degree turns are exclude and
+    # if pmf_threhold is superior to 0.4. Streamlines should stop at
```
arokem: Small nit: "superior to" is perhaps not the right word. How about "larger than"?

arokem:
This looks great to me. What are your thoughts on absorbing the inevitable changes to previous behavior? I think it'll probably be alright (that's what happens when you upgrade software...), but if someone has any ideas on how to not surprise users with this, that would be good.

Either way, a note in the API changes doc needs to be made.

gabknight:
Thanks for your comments.

With pmf_threshold = 0.1, changes in the reconstruction are small, but I would prefer having it at 0.1 by default that 0.0. My only concern is reproducibility of previous results, with previous code. 

How should this note be made? In the documentation of the class `ProbabilisticDirectionGetter`?

arokem:
In here: https://github.com/nipy/dipy/blob/master/doc/api_changes.rst, under a new 0.12 heading.

gabknight:
done, thx @arokem

arokem:
Anyone have any more comments here, or anxious to give this a look? If not, I will merge it by the end of the week.

PR #1042, opened 2016-05-04
RF - move direction getters to dipy/direction/

gabknight:
This PR moved directionGetter classes to dipy/direction/. Let me know what you think. @MrBago 
- Moved direction getters from `dipy/tracking/local/` and from `dipy/reconst/` to `dipy/direction/`
- Rename `probabilistic_direction_getter.py` to `pmf_direction_getter.py`. The file contains more than probabilistic getters, e.g. `DeterministicMaximumDirectionGetter`. All of them need discrete probability mass functions (pmf).

Best,
Gab


MrBago:
So figuring out where exactly to put all these files was a bit tricky. The issue is with import order. In python imports must form an DAG (directed a-cyclic graph). We obviously have a DAG in modules but as is, it's all a bit fragile. For example if you try adding "import .csdeconv" to "reconst.__init__.py" you'll break dipy.direction.peaks. We can avoid this situation by forcing the parent directories to also form a DAG. An example ("->" means imports from):

dipy.reconst.csdeconv -> dipy.direction.peaks -> dipy.reconst.shm. This is a DAG, but if we just look at the top level we have dipy.reconst -> dipy.direction -> dipy.reconst (a cycle). 

I suggest we establish an import hierarchy, for example reconst -> direction -> tracking -> data -> core ...

A package can import from all packages lower on the hierarchy. For example direction can import from data  and core. To be clear, I'm not suggesting this specific hierarchy, I need to look closer at to code to see the best options. Having this kind of hierarchy has two advantages:

* It allows us to safely "promote" classes and functions to higher levels of the API, like we do with dipy.direction.DirectionGetter
* It makes the import DAG more clear for future development. Otherwise it's only a matter of time until someone tries to import the wrong module and trips all over this.

A few exceptions, we can have "dummy modules" like dipy.reconst.peaks (mainly for backwards compatibility). These modules can break the rules, but nothing should import from them. Also tests don't need to follow these rules, but no "normal" modules should ever import from tests.

I hope all this is clear. If you guys think having a hierarchy is a good idea, we can look more at this. If you'd prefer to punt this thing, I understand :).

PR #1043, opened 2016-05-04
DO NOT MERGE: Add a test of local tracking, using data from dipy.data.

arokem:
I put this here, just to demonstrate something that I have encountered. The localtracking module seems to raise a spurious error, even for data that is regularly sampled. Unless I am misunderstanding something, the script that I have added as `test_local_data.py` should run with no error, but I think that this demonstrates that there is a bug (?) in the way that localtracking.py ascertains the voxel size. According to the nifti header, this data is acquired on a regular 2.5 x 2.5 x 2.5 grid. 

Shouldn't this all run smoothly?

### DIFF
--- a/dipy/tracking/local/tests/test_local_data.py
+++ b/dipy/tracking/local/tests/test_local_data.py
@@ -15,28 +15,29 @@ import dipy.reconst.dti as dti
 from dipy.reconst.dti import fractional_anisotropy
 
 
-fdata, fbval, fbvec = dpd.get_data('small_101D')
-img = nib.load(fdata)
-gtab = dpg.gradient_table(fbval, fbvec)
-data = img.get_data()
-affine = img.get_affine()
+def test_local_tracking_with_data():
+    fdata, fbval, fbvec = dpd.get_data('small_101D')
+    img = nib.load(fdata)
+    gtab = dpg.gradient_table(fbval, fbvec)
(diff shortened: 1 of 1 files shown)


gabknight:
In this image, the affine is:
```
affine = [[ -2.49969149e+00   0.00000000e+00  -3.92675400e-02]
          [ -6.74999756e-05   2.49999619e+00   4.36446397e-03]
          [ -3.92674841e-02  -4.36500227e-03   2.49968767e+00]]
```
It raises a flag because, somehow, the affine is not symetric 'enough'. affine[0,1] != affine[1,0]. Any idea why this is the case? I'm not very familiar with this...

If you would reduce the error tolerance, e.g. by changing line 28 of `localTracking.py` with: `if not np.allclose(np.triu(dotlin, 1), 0., atol=1e-5):`, the test would pass. 

For the test, add `list(streamlines)` at the end, otherwise the tracts won't be generated (`localTracking(.)` returns a generator).

arokem:
Thanks for the tip @gabknight! Indeed, adding tolerance here did make this work. 

I am not sure why this particular affine has this characteristic. I am not even sure where the `small_101D` data comes from (maybe @Garyfallidis knows?). But I also don't think that it's that unusual, so might be worth accommodating. I'll make another PR aimed at that, and close this one

PR #1046, opened 2016-05-09
BF - peaks_from_model with nbr_processes <= 0

gabknight:
Fixed behavior when nbr_processes parameter of peaks_from_model is less or equal to 0.

- returns peaks_from_model(..., parallel=False) when nbr_processes <= 0
- added tests


Garyfallidis:
Good catch @gabknight ! Thx! :)

PR #1050, opened 2016-05-13
ENH: expand OpenMP utilities and move from denspeed.pyx to dipy.utils

grlee77:
I was recently working on adding OpenMP support to some functions in dipy.align and needed some utilities similar to the ones defined within `denoise/denspeed.pyx`.

I propose moving these to a common location `dipy.utils.omp` where they can be more easily found and reused.

The only change in functionality to `nlmeans` would be the default number of threads set when `num_threads` is not specified,but the user has also defined the `OMP_NUM_THREADS` environement variable.  Previously the default number of threads was always `openmp.omp_get_num_procs()`, but now it will try to default to the `OMP_NUM_THREADS` instead.  At least on my system, `omp_get_num_procs` returns the total number of possible threads including hyperthreading which actually makes things slower than just using the number of physical cores.

### DIFF
--- a/dipy/denoise/denspeed.pyx
+++ b/dipy/denoise/denspeed.pyx
@@ -98,8 +98,7 @@ def _nlmeans_3d(double[:, :, ::1] arr, double[:, :, ::1] mask,
     J = arr.shape[1]
     K = arr.shape[2]
 
-    if have_openmp:
-        set_num_threads(num_threads)
+    set_num_threads(num_threads)
 
     # move the block
     with nogil, parallel():
@@ -112,7 +111,7 @@ def _nlmeans_3d(double[:, :, ::1] arr, double[:, :, ::1] mask,
 
                     out[i, j, k] = process_block(arr, i, j, k, B, P, sigma)
 
-    if have_openmp and num_threads is not None:
+    if num_threads is not None:
         restore_default_num_threads()
 
     new = np.asarray(out)
--- a/dipy/denoise/denspeed.pyx
+++ b/dipy/denoise/denspeed.pyx
@@ -8,7 +8,9 @@ cimport safe_openmp as openmp
 from safe_openmp cimport have_openmp
 
 from cython.parallel import parallel, prange
-from multiprocessing import cpu_count
+# import cpu_count for backwards compatibility
+from dipy.utils.omp import cpu_count
+from dipy.utils.omp cimport set_num_threads, restore_default_num_threads
 
 from libc.math cimport sqrt, exp
 from libc.stdlib cimport malloc, free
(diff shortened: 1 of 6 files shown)


riddhishb:
@grlee77 this is a good idea, just a quick question. 
Does this mean that when `nlmeans` is run without specifying num_threads, it would use all the cores, I believe thats what `OMP_NUM_THREADS` does right?
I am not clear on that part

grlee77:
By default OMP_NUM_THREADS is probably not set on most systems unless the user has explicitly done so.  In that case the max number of threads returned by omp_get_num_procs would be used.

OMP_NUM_THREADS is an environment variable that can be set by the user (e.g. via `export OMP_NUM_THREADS=4` in a bash terminal on linux or osx prior to starting python).  For the laptop I am on now there are 4 physical CPU cores, but 8 maximum threads.  In practice I find using 4 threads is faster than 8 so I set OMP_NUM_THREADS=4 and would like dipy to honor that choice rather than always forcing to 8.

However, looking more closely at these docs (https://gcc.gnu.org/onlinedocs/libgomp/OMP_005fNUM_005fTHREADS.html), I see that technically OMP_NUM_THREADS can also be set to a comma separated list.  I have not supported that case here.  The code as implemented here would fail to convert the list to an integer and fallback to the maximum number of threads as returned by omp_get_num_procs.

riddhishb:
Got it! :) and this is only when the num_threads keyword is not specified, right?

grlee77:
Right.

arokem:
@omarocegueda : would you mind giving this one a look as well? Is it still relevant?

dipy/denoise/denspeed.pyx
```diff
@@ -117,8 +111,8 @@ def _nlmeans_3d(double[:, :, ::1] arr, double[:, :, ::1] mask,
 
                     out[i, j, k] = process_block(arr, i, j, k, B, P, sigma)
 
-    if have_openmp and num_threads is not None:
-        openmp.omp_set_num_threads(all_cores)
+    if num_threads is not None:
+        restore_default_num_threads()
```
omarocegueda: Just out of curiosity, is it possible that this logic has unexpected side effects? What I understand is that we are "restoring" the number of threads after finishing the parallel block. But what if the number of threads was not the default when we called this function?

omarocegueda: Yes, I think that would have been better. Resetting the number of threads to the default seems to me as unexpected as not restoring it at all. Anyway, that issue is unrelated to this PR =).

arokem:
Other than your previous comment, @omarocegueda : what do you think here?

arokem:
OK - this seems ready to go, so unless someone pipes up, I will merge it in a couple of days.

PR #1053, opened 2016-05-20
WIP: Local pca and noise estimation

riddhishb:
This is the rough PR for the LocalPCA denoising which I am working on. This not at all the final version, and is for reviewing. 
@Garyfallidis @RafaelNH @omarocegueda @villalonreina do have a look at this.


arokem:
Exciting to see the beginning of the GSoC project! I realize this is early days. For now, just a naming comment: since the `localPCA_denoise` module is a submodule of the `denoise` module, I think that it would be fine to call it `localPCA`.

Garyfallidis:
Hi @riddhishb I changed the title to be WIP (work in progress). That allows to the other developers to see that this not ready to be merged but that people are allowed to look and give suggestions/comments.

dipy/denoise/localPCA_denoise.py
```diff
@@ -0,0 +1,116 @@
+import numpy as np
+from dipy.denoise.rician_adaptation import rician_adaptation
+
+def localPCA_denoise(arr, sigma, patch_radius=1, tou=0, rician=True):
```
Garyfallidis: Please change the name of the file to localpca.py
And the name of the function to localpca. We do not use capital letters in function names.

riddhishb:
@Garyfallidis @arokem will do :)

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,114 @@
+import numpy as np
+
+def localpca(arr, sigma, patch_radius=1, tou=0, rician=True):
+    '''
+    Local PCA Based Denoising of Diffusion Datasets
+
+    References
```
RafaelNH: References should be after Parameters and Returns

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,114 @@
...
+    tou : float or 3D array
+        threshold parameter for diagonal matrix thresholding,
+        default value = (2.3 * sigma * sigma)
+    rician : boolean
+        If True the noise is estimated as Rician, otherwise Gaussian noise
+        is assumed.
+
+    Returns
+    -------
+
```
RafaelNH: Remove unnecessary white space.

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,114 @@
...
+    denoised_arr : 4D array
+        this is the denoised array of the same size as that of
+        arr (input data) 
+
```
RafaelNH: Same as above

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,124 @@
...
+def localpca(arr, sigma, patch_radius=1, tou=0, rician=True):
+    '''
+    Local PCA Based Denoising of Diffusion Datasets
+
+    References
+    ----------
+    Diffusion Weighted Image Denoising Using Overcomplete Local PCA
+    Manjon JV, Coupe P, Concha L, Buades A, Collins DL
+
```
RafaelNH: Simplify first line description of function - add a reference instead of list of names: "Diffusion Weighted Image Denoising Using Overcomplete Local PCA [1]_"

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,124 @@
+import numpy as np
+import scipy as sp
+
+def localpca(arr, sigma, patch_radius=1, tou=0, rician=True):
+    '''
+    Local PCA Based Denoising of Diffusion Datasets
```
RafaelNH: Add the reference in first line description of function: "Local PCA Based Denoising of Diffusion MRI Datasets [1]_"

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,124 @@
...
+        is assumed.
+
+    Returns
+    -------
+
+    denoised_arr : 4D array
+        this is the denoised array of the same size as that of
+        arr (input data) 
+
+    '''
```
RafaelNH: Add a reference field:

References
--------------
.. [1] Manjon JV, Coupe P, Concha L, Buades A, Collins DL, 2013. Diffusion Weighted Image Denoising Using Overcomplete Local PCA. PLOS one http://dx.doi.org/10.1371/journal.pone.0073021

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,124 @@
+import numpy as np
+import scipy as sp
+
+def localpca(arr, sigma, patch_radius=1, tou=0, rician=True):
+    '''
```
RafaelNH: default tou should be None. Below when checking tou you should use:
if tou is None:

riddhishb:
@RafaelNH will do :)

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,137 @@
...
+                    sigma[i,j,k] = beta * beta * np.median(d[np.sqrt(d[d > alpha]) < 2 * median_sqrt])
+                    if(np.isnan(sigma[i,j,k])):
+                        sigma[i,j,k] = 0
+
+                    d[d < sigma[i,j,k]] = 0
+                    D_hat = np.diag(d)
+                    Y = X.dot(W)
+                    # When the block covers each pixel identify it into the label matrix theta
+                    X_est = Y.dot(np.transpose(W))
+                    X_est = X_est.dot(D_hat)
```
RafaelNH: Perhaps a good test will be comparing these lines of code with: 
X_est = X.dot(D_hat)  # your previous way to compute X_est

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,137 @@
...
+                for i in range(patch_radius, arr.shape[0] - patch_radius , 1):
+                    
+                    X = np.zeros((patch_size * patch_size * patch_size, arr.shape[3]))
+                    M = np.zeros(arr.shape[3])
+
+                    temp = arr[i - patch_radius : i + patch_radius + 1,j - patch_radius : j + patch_radius + 1,
+                         k - patch_radius : k + patch_radius + 1,:]
+                    X = temp.reshape(patch_size * patch_size * patch_size, arr.shape[3])
+                    # compute the mean and normalize
+                    M = np.mean(X,axis=1)
```
RafaelNH: Use PEP8 style - after ´,´ use space 
M = np.mean(X, axis=1)

dipy/denoise/localpca.py
```diff
@@ -78,7 +79,19 @@ def localpca(arr, sigma, patch_radius=1, tou=0, rician=True):
                     # compute EVD of the covariance matrix of X get the matrices W and D, hence get matrix Y = XW
                     # Threshold matrix D and then compute X_est = YW_transpose D_est
                     [d,W] = np.linalg.eigh(C)
-                    d[d < tou[i][j][k][:]] = 0
+                    # for sigma estimate we perform the median estimation
+
+                    # find the median of the standard deviation of the eigenvalues
+                    median_sqrt = np.median(np.sqrt(d[d > alpha]))
```
omarocegueda: I think `alpha` needs to be a parameter of the function, otherwise you are assuming this was already set (as in your example)

dipy/denoise/localpca.py
```diff
@@ -78,7 +79,19 @@ def localpca(arr, sigma, patch_radius=1, tou=0, rician=True):
...
+                    if(np.isnan(median_sqrt)):
+                        median_sqrt = 0
+                    # Chop of the positive eigenvalues whose standard deviation is more that 2 times that of the above quantity
+                    # Take the remaining eigenvalues and estimate sigma
+                    sigma[i,j,k] = beta * beta * np.median(d[np.sqrt(d[d > alpha]) < 2 * median_sqrt])
```
omarocegueda: Same here, `beta` needs to be an input parameter as well

dipy/denoise/localpca.py
```diff
@@ -78,7 +79,19 @@ def localpca(arr, sigma, patch_radius=1, tou=0, rician=True):
                     # compute EVD of the covariance matrix of X get the matrices W and D, hence get matrix Y = XW
                     # Threshold matrix D and then compute X_est = YW_transpose D_est
                     [d,W] = np.linalg.eigh(C)
-                    d[d < tou[i][j][k][:]] = 0
+                    # for sigma estimate we perform the median estimation
+
+                    # find the median of the standard deviation of the eigenvalues
```
omarocegueda: Actually `... median of the square root of positive eigenvalues...`, right?

dipy/denoise/localpca.py
```diff
@@ -78,7 +79,19 @@ def localpca(arr, sigma, patch_radius=1, tou=0, rician=True):
...
+                    median_sqrt = np.median(np.sqrt(d[d > alpha]))
+                    if(np.isnan(median_sqrt)):
+                        median_sqrt = 0
+                    # Chop of the positive eigenvalues whose standard deviation is more that 2 times that of the above quantity
```
omarocegueda: Same here, `whose square root is ...` instead of `standard deviation...`, right?

riddhishb:
@omarocegueda I did the median of eigenvalues noise estimation, they initially trim the eigenvalues to be taken for median as
screen shot 2016-05-29 at 12 23 53 pm
hence the code.
Also I was testing with the example mainly so might have left to update the localpca.py I did that in this commit. However if you follow the hangout chat the sigma estimation by MUBE / SIBE seems to work better than the median based denoising (at least for the gaussian case with the retained eigenvalue plot).

@RafaelNH I will make the codes PEP8 the next time I commit :)

dipy/denoise/noise_estimate_localpca.py
```diff
@@ -0,0 +1,111 @@
...
+        # X = NxK
+        # Now to subtract the mean
+        M = np.mean(X,axis = 0)
+        X = X - M
+        C = np.transpose(X).dot(X)
+        # C = C/data0.shape[3]
+        # Do PCA and find the lowest principal component
+        [d,W] = np.linalg.eigh(C)
+        d[d < 0] = 0;
+        d_new = d[d != 0]
```
RafaelNH: I didn't get these two last code lines! I thought you just need to take the eigenvector of the lower eigenvalue, compute the Y for the this eigenvector and from this compute local std using a sliding window [3 x 3 x 3]

dipy/denoise/noise_estimate_localpca.py
```diff
@@ -0,0 +1,111 @@
...
+        # As the W is sorted in increasing eignevalue order, so is V
+        # choose the smallest positive principle component
+        I = V[:,d.shape[0] - d_new.shape[0]].reshape(data.shape[0], data.shape[1], data.shape[2])
+        
+
+    else:
+        print("SIBE Estimate")
+        # if only one b0 value then
+        # SIBE Noise Estimate
+        data0 = data[...,~gtab.b0s_mask]
```
RafaelNH: We will have memory issues here if number of gradient directions is large.

dipy/denoise/noise_estimate_localpca.py
```diff
@@ -0,0 +1,111 @@
...
+        [d,W] = np.linalg.eigh(C)
+        d[d < 0] = 0;
+        d_new = d[d != 0]
+        # we want eigen vectors of XX^T so we do
+        V = X.dot(W)
+        # unit normalize the clolumns of V
+        # V = V / np.linalg.norm(V, ord=2, axis=0)
+        # As the W is sorted in increasing eignevalue order, so is V
+        # choose the smallest principle component
+        I = V[:,d.shape[0] - d_new.shape[0]].reshape(data.shape[0], data.shape[1], data.shape[2])
```
RafaelNH: These last lines of code are the same for if(K >1), so I will suggest you to take these out of the if/else statement. The only think that have to be in the statement is 

if(K > 1):
    data0 = data[..., gtab.b0s_mask]
else:
    data0 = data[..., ~gtab.b0s_mask]

right?

riddhishb: Oh yeah will make the changes

dipy/denoise/noise_estimate_localpca.py
riddhishb: We have to take the eigenvector for the lowest **positive** value and hence the code

dipy/denoise/noise_estimate_localpca.py
```diff
@@ -0,0 +1,111 @@
...
+        print(i)
+        for j in range(1, I.shape[1] - 1):
+            for k in range(1, I.shape[2] - 1):
+                    
+                temp = I[i-1:i+2, j-1:j+2, k-1:k+2]
+                temp = (temp - np.mean(temp)) * (temp - np.mean(temp))
+                sigma[i-1:i+2, j-1:j+2, k-1:k+2] += temp  
+                temp = data[i-1:i+2, j-1:j+2, k-1:k+2, :]
+                mean[i-1:i+2, j-1:j+2, k-1:k+2, :] +=  np.mean(np.mean(np.mean(temp,axis = 0),axis=0),axis=0)
+                count[i-1:i+2, j-1:j+2, k-1:k+2,:] += 1
```
RafaelNH: You should use the information of the neighbours to compute the local variance and mean of the central voxel only. I know that this is not the case from the PCA denoising results (localpca.py lines 107-117 since doing the denoising in "an overcomplete manner"). However, here there is no justification why the variance should be estimated in this manner. How about using the following lines?

sigma = np.var(I[i-1:i+2, j-1:j+2, k-1:k+2], ddof=1)

Also as we discussed by e-mail the mean should be computed from the original data and not from the lower PCA component which only contains information of noise and not the signal.

riddhishb: @RafaelNH Fair point, Will just take the local variance and not in the over complete manner. And the mean is from the original data as I have redefined temp as `temp = data[i-1:i+2, j-1:j+2, k-1:k+2, :]` and data is the original data, but yeah that too not in an overcomplete manner.

dipy/denoise/noise_estimate_localpca.py
RafaelNH: I know that we discuss this in the meeting, however I do not get it - if the number of voxels per volume is larger than the number of volumes (i.e. X has a dimention of Nvoxels x Nvolumes), you will not have zero eigenvalues (assuming that all volumes contain different information). So you basically have to use the lower eigenvalue to have the PCA lower component as mentioned in the article (a suggestion - plot the PCA components to check if this only contains noise as I am assuming).

dipy/denoise/noise_estimate_localpca.py
RafaelNH: yes! You are right, you are already computing the mean from the original data. I take that out =P.

However, my concern regarding how you compute the variance and local mean from the average of overlapping voxels of slided patches still maintain.

RafaelNH: From lines 77-92, I still think that the right way to do this is only the following:

sigma = np.zeros(I.shape, dtype = np.float64)
mean = np.zeros_like(data).astype(np.float64)
# Compute the local noise variance of the I image
     for i in range(1,I.shape[0] - 1):
              for j in range(1, I.shape[1] - 1):
                           for k in range(1, I.shape[2] - 1):
                sigma[i, j, k] = np.var(I[i-1:i+2, j-1:j+2, k-1:k+2], ddof=1)
                 mean[i, j, k] +=  np.mean(data[i-1:i+2, j-1:j+2, k-1:k+2])

riddhishb: Yeah cool, only a minor point the mean will be a 4 dimensional matrix so `mean[i, j, k] = np.mean(data[i-1:i+2, j-1:j+2, k-1:k+2])` should be replaced by `mean[i, j, k] = np.mean(np.mean(np.mean(data[i-1:i+2, j-1:j+2, k-1:k+2], axis =0),axis =0), axis =0)`

RafaelNH: But you want one estimate of sigma for each voxel, right?

RafaelNH: Oh yes! I have a nice idea, you can have the mean signal for different gradient directions. Then correct the sigma for each direction independently. The final sigma estimate is then the average of the corrected sigmas for each direction (theoretically this will be the most accurate and unbiased sigma estimate).

riddhishb: Thats a good idea indeed, previously I was doing something, and had as many estimates of sigma as the number of gradient directions, this taking the mean of all the sigma estimates makes more sense

RafaelNH: Yes, because what you want is to have a sigma estimate per voxel (3D variable), right?

riddhishb: Yes correct!, cant believe it escaped my notice :( . Will try to incorporate these changes once I am done with the adaptive denoising PR. Thanks :)

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,171 @@
...
+                        j -
+                        patch_radius: j +
+                        patch_radius +
+                        1,
+                        k -
+                        patch_radius: k +
+                        patch_radius +
+                        1,
+                        :]
+                    X = temp.reshape(
```
omarocegueda: Sorry, this maybe a dumb question but I always get confused with this kind of reshapes: are we sure that after this reshape, each **row** of X has one of the diffusion signals within the local window?

dipy/denoise/noise_estimate_localpca.py
```diff
@@ -0,0 +1,110 @@
...
+
+    # find the SNR and make the correction
+    # SNR Correction
+
+    for l in range(data.shape[3]):
+        print(l)
+        snr = mean[..., l] / np.sqrt(sigma)
+        eta = 2 + snr**2 - (np.pi / 8) * np.exp(-0.5 * (snr**2)) * ((2 + snr**2) * sp.special.iv(
+            0, 0.25 * (snr**2)) + (snr**2) * sp.special.iv(1, 0.25 * (snr**2)))**2
+        sigma_corr[..., l] = sigma / eta
```
RafaelNH: Hi @riddhishb! This is indeed frustrating - at the weekend I mentioned that the most correct sigma estimate will be averaging the corrected sigmas for each directions. However, I was forgetting one thing, we only have one sigma estimate for each voxel and not sigma estimates for each direction of the voxel. I will suggest you to try correcting the sigma using the overall mean instead of the looping the means for each direction.

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,171 @@
...
+                        patch_size * patch_size * patch_size, arr.shape[3])
+                    # compute the mean and normalize
+                    M = np.mean(X, axis=1)
```
RafaelNH: Humn - Can you double check this line? You have to take the mean of each column - in this way you are taking the mean of each line, right?

riddhishb: Oh I missed that. I corrected it and the results improved a great deal by that

dipy/denoise/fast_local_covariance.pyx
```diff
@@ -0,0 +1,200 @@
...
+        for i in range(n):
+            for j in range(n):
+                out[i,j] -= cov[i, j]
+
+
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+@cython.cdivision(True)
+def fast_local_covariance(double[:,:,:,:] I, int radius, double[:,:,:,:,:] out, double[:,:,:,:] meanout):
```
omarocegueda: This algorithm was not intended to be used this way. Notice that the amount of memory required to store all local covariances (and local means) is prohibitively large. We need to process each covariance "on the fly" so that we only need to save the local covariances within two slices of the volume.

riddhishb: oh so at the end of the where the cython ends I will have to continue with the rest of the code, is it ?

riddhishb: oh so at the end of the where the cython code ends I will have to continue with the rest of the code, is it ?

dipy/denoise/fast_lpca.pyx
```diff
@@ -0,0 +1,255 @@
...
+                            for jj in range(ndiff):
+                                cov[ii,jj] /= nsamples
+                        # Subtract self-outer product of mu from cov
+                        update_outer_prod(cov, mu, -1)
+
+                        #Local covariance ready
+                        # update_matrix(out[i-radius, j-radius, k-radius], cov, 0)
+                        # update_vector(meanout[i-radius, j-radius, k-radius], mu, 0)
+
+                    with gil:
```
omarocegueda: This seems to be part where you process the local covariance matrix, it should be part of the `if` condition

dipy/denoise/fast_lpca.pyx
```diff
@@ -0,0 +1,255 @@
...
+                        # eigen value decomposition
+                        [d,W] = np.linalg.eigh(cov)
```
omarocegueda: I think this is the only part of the code that needs to be part of the "with gil" block.

dipy/denoise/fast_lpca.pyx
```diff
@@ -0,0 +1,255 @@
...
+                        
+                        if((i >= radius and i < n0-radius) and (j >= radius and j < n1-radius) and (k >= radius and k < n2-radius)):
```
omarocegueda: Here, `(i, j, k)` is the last corner of the window. The center is `(i-radius, j-radius, k-radius)`. Since all this processing should be part of the above `if`, we know that the center is valid.

dipy/denoise/fast_lpca.pyx
```diff
@@ -0,0 +1,255 @@
...
+                            X = X - np.array([mu, ] * X.shape[0],
+                                             dtype=np.float64)
+
+                            d[d < tou[i, j, k, :]] = 0
+                            # Y = X.dot(W)
+                            W_hat = W * 0
+                            W_hat[:, d > 0] = W[:, d > 0]
+
+                            Y = X.dot(W_hat)
+                            X_est = Y.dot(np.transpose(W_hat))
```
omarocegueda: This is what we need to do signal by signal: first precompute `P = W_hat.dot(np.transpose(W_hat))`, and then iterate over all voxels within the window and multiply by P

dipy/denoise/fast_lpca.pyx
```diff
@@ -0,0 +1,255 @@
...
+
+                            # add the mean
+                            temp = X_est + \
+                                np.array([mu, ] * X_est.shape[0], dtype=np.float64)
```
omarocegueda: This can be part of the projection function too: 1.Subtract the mean, 2.Project, 3.Add the mean.

riddhishb:
@omarocegueda Will work on your comments :)

dipy/denoise/fast_lpca.pyx
```diff
@@ -0,0 +1,307 @@
...
+
+                        i0 = i - radius
+                        j0 = j - radius
+                        k0 = k - radius
+
+                        with gil:
+                            # eigen value decomposition
+                            d, W_hat = np.linalg.eigh(cov)
+
+                        l0norm = 0
```
omarocegueda: Instead of first collecting the eigenvectors whose eigenvalue is not less than `sigma*2,3^2` and then multiply the resulting matrix with its transpose, you can simply start with a zero `P`and just add the outer product of the eigenvectors: 
```
                        l0norm = 0
                        for p in range(ndiff):
                            if d[p] >= sigma[i0, j0, k0, p] * 2.3 * 2.3:
                                l0norm += 1
                                update_outer_prod(P, W_hat[:,p], 1)

                        l0norm = 1.0 / (1 + l0norm)
```
Before doing this change, please convince yourself that this is equivalent and verify!.

riddhishb: @omarocegueda yeah that makes sense :), it improved the performance a bit! I verified this and it works!

dipy/denoise/fast_lpca.pyx
```diff
@@ -0,0 +1,270 @@
...
+    elif utype==-1:
+        for i in range(n):
+            for j in range(n):
+                out[i,j] -= cov[i, j]
+
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+@cython.cdivision(True)
+def fast_lpca(double[:, :, :, :] I, int radius, double[:, :, :, :] sigma):
```
omarocegueda: Are we considering a different sigma field for each gradient?

riddhishb: no we are not, the function is having 4D for convenience, but the sigma at all directions are same

dipy/denoise/fast_lpca.pyx
```diff
@@ -0,0 +1,270 @@
...
+                        k0 = k - radius
+
+                        with gil:
+                            # eigen value decomposition
+                            P[...] = 0
+                            d, W_hat = np.linalg.eigh(cov)     
+                            
+                        l0norm = 0
+                        for p in range(ndiff):
+                            if d[p] >= sigma[i0, j0, k0, p] * 2.3 * 2.3:
```
omarocegueda: I don't think having duplicated sigmas is more convenient, it is actually decreasing performance because this product can be precomputed out of the for loop here

omarocegueda: Oh! and it may be using a lot of unnecessary memory =s

riddhishb: Yeah :) great point will change asap.

omarocegueda: Let's squeeze up to the last bit of memory =P

dipy/denoise/fast_lpca.pyx
```diff
@@ -0,0 +1,270 @@
...
+        cnp.npy_intp nsamples = m * m * m
+        cnp.npy_intp cur_i, prev_i, i, j, k, ii
+        cnp.npy_intp i0, j0, k0, p, q, r, s
+        double l0norm
+        double[:, :, :, :, :] T = np.zeros((2, n1, n2, ndiff, ndiff))
+        double[:, :, :, :] S = np.zeros((2, n1, n2, ndiff))
+        double[:, :] cov = np.empty((ndiff, ndiff))
+        double[:] mu = np.empty(ndiff)
+        double[:] x
+        double[:, :, :, :] theta = np.zeros((n0, n1, n2, ndiff))
```
omarocegueda: We also have several copies of theta here, one `(n0, n1, n2)` volume would be enough

riddhishb: @omarocegueda I changed that in the code which I will commit soon :), memory saving mode on!

dipy/denoise/fast_lpca.pyx
```diff
@@ -0,0 +1,286 @@
+import numpy as np
+cimport cython
+cimport numpy as cnp
+
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+@cython.cdivision(True)
+cdef void fast_vecmat_mul(double[:] x, double[:, :] A, double[:] out) nogil:
```
omarocegueda: Is it possible to pass raw pointers instead of memory views here?:
`fast_vecmat_mul(double *x, double ** A, double* out)`

dipy/denoise/fast_lpca.pyx
```diff
@@ -0,0 +1,286 @@
...
+
+                        for p in range(i0 - radius, i0 + radius + 1):
+                            for q in range(j0 - radius, j0 + radius + 1):
+                                for r in range(k0 - radius, k0 + radius + 1):
+
+                                    for s in range(ndiff):
+                                        temp[s] = I[p, q, r, s] - mu[s]
+
+                                    fast_vecmat_mul(temp, P, temp1)
+                                    update_vector(temp1, mu, 1)
```
omarocegueda: If we refactor `fast_vecmat_mult` then `temp` and `temp1` will be plain arrays, so we may need another version of `update_vector` receiving `double*` too

dipy/denoise/fast_lpca.pyx
```diff
@@ -0,0 +1,286 @@
...
+                        with gil:
+                            # eigen value decomposition
+                            P[...] = 0
+                            d, W_hat = np.linalg.eigh(cov)
+
+                        l0norm = 0
+                        for p in range(ndiff):
+                            if d[p] >= sigma[i0, j0, k0] * 2.3 * 2.3:
+                                l0norm += 1
+                                update_outer_prod(P, W_hat[:, p], 1)
```
omarocegueda: If we refactor `fast_vecmat_mul` then `P` is now `double **` so we may need a new `update_outer_prod` as well

dipy/denoise/fast_lpca.pyx
```diff
@@ -0,0 +1,286 @@
...
+        cnp.npy_intp n0 = I.shape[0]
+        cnp.npy_intp n1 = I.shape[1]
+        cnp.npy_intp n2 = I.shape[2]
+        cnp.npy_intp ndiff = I.shape[3]
+        cnp.npy_intp m = 1 + (2 * radius)
+        cnp.npy_intp nsamples = m * m * m
+        cnp.npy_intp cur_i, prev_i, i, j, k, ii
+        cnp.npy_intp i0, j0, k0, p, q, r, s
+        double l0norm
+        double[:, :, :, :, :] T = np.zeros((2, n1, n2, ndiff, ndiff))
```
omarocegueda: It may be tempting to refactor everything to plain arrays. I just tried but allocating memory resulted in an extreme overhead, so unless we reserve the temporary memory as one single memory block it won't improve performance.

doc/examples/denoise_localpca.py
```diff
@@ -0,0 +1,117 @@
+"""
+===============================
+Denoise images using Local PCA
+===============================
+
+Using the local PCA based denoising for diffusion
```
RafaelNH: Can you rewrite this paragraph? A bit confusing - perhaps it will be nice to give some more information of the PCA denoising techniques.

riddhishb: @RafaelNH On it!

doc/examples/denoise_localpca.py
```diff
@@ -0,0 +1,120 @@
...
+images [Manjon2013]_ we can denoise diffusion images significantly. This algorithm is
+popular as it takes into account the directional information in diffusion data.
```
RafaelNH: Last sentence: 'For diffusion-weighted data this algorithm is advantageous for taking into account the directional information in diffusion data.'

dipy/denoise/localpca_slow.py
```diff
@@ -0,0 +1,118 @@
...
+    References
+    ----------
+    .. [Manjon13] Diffusion Weighted Image Denoising Using Overcomplete Local PCA
+                  Manjon JV, Coupe P, Concha L, Buades A, Collins DL
+
+    """
+
+    if arr.ndim == 4:
+
+        tou = 2.3 * 2.3 * sigma
```
samuelstjean: This should be the variance, not the standard deviation, so your docstring is misnommed or you missed the square. See see http://journals.plos.org/plosone/article?id=10.1371/journal.pone.0073021#s2

~~~
After an exhaustive search for the optimum value of τ parameter, it was set to (2.3σ)2, where σ2 is the estimated local noise variance
~~~

riddhishb: I checked it's not named correctly, and I will change that.

PR #1038, opened 2016-04-25
BF: Should fix #1037

arokem:

PR #1054, opened 2016-05-21
Remove tracking interfaces

arokem:
This should supersede #1038, and removes the `tracking.interfaces` module altogether. It's been in limbo since as far back as 2012: https://github.com/nipy/dipy/commit/4ccf0bb


Garyfallidis:
Thx @arokem This was an orphan for many years. And there is no plan to support traits. It would be good to have a thumbs up from @MrBago who wrote this code. So we can move ahead with the removal.
@arokem please also remove the dipy_tracking_gui command https://github.com/nipy/dipy/blob/master/bin/dipy_tracking_gui

arokem:
Merged a commit from @MrBago removing a couple more outdated modules, and removed the tracking gui script.

MrBago:
See #1056.

PR #740, opened 2015-10-19
Improved mapmri implementation with laplacian regularization and new …

rutgerfick:
…q-space scalar indices: q-space inverse variance and mean squared displacement. Also includes generalized cross-validation to find optimal regularization weights. Other convenience functions have been added.

PR #1055, opened 2016-05-21
WIP: Ivim implementation

sahmed95:
I will be working on this over the summer as a GSoC project mentored by @arokem, @RafaelNH and Eric Peterson. Inside the folder ivim_dev, I have a working code built on Eric's initial implementation (https://github.com/etpeterson) which has been modified to follow the dki code. A [Jupyter notebook](https://github.com/sahmed95/dipy/blob/ivim/dipy/ivim/ivim_dev/ivim_example.ipynb) demonstrates the earlier work using simulated data. It closely follows the dki code. 

We will go part by part with necessary tests at each stage. ivim.py is the main file and in my opinion reviewing ivim_dev shouldn't be a priority for now as most of it will be changed

The model fits the equation S = S0(f*e^(-bD_star) + (1-f)e^(-bD)) which gives a measure of the diffusion and perfusion parameters. You can find the complete proposal [here](https://drive.google.com/file/d/0B2sGDIE9AVckUVZZYnZ5U0h2NTQ/view?usp=sharing)


sahmed95:
Ah, I think I should have broken down the commits into smaller parts !

arokem:
Yes. This is probably too large to be merged as one piece. We usually prefer PRs with less LOC: there's empirical evidence that reviewing more than 200 LOC becomes unwieldy. We've seen that here. For example, #740 is languishing among other reasons because it's so large. But let's leave this here for reference, so that we can see what you are thinking about. We can break it apart for review, after you explain all the moving pieces to us (on our hang out this Monday).

Garyfallidis:
Hello @sahmed95. Thank you for this PR. This is the reconstruction model so the main  modules (ivim.py) should be under dipy/reconst also we need to break this as Ariel says in many PRs and decide which tutorials will be available and which not. I see also that you use ipython notebooks which is something which is the first time to save directly in the codebase. Thanks again and best of luck with GSoC!

etpeterson:
Hi @sahmed95, good to see this. I'll take a look and I'm looking forward to talking it though on Monday.

sahmed95:
Ok, I am getting the hang of it now. I will keep pushing smaller commits. In ["clean slate"](https://github.com/nipy/dipy/pull/1055/commits/1d22bb13744ed50e6bd383b5e9273c3444bdeb1a) I removed all previous code. This can be ignored and reviewing can be started from the next commit. 

@Garyfallidis I was using the Ipython notebooks for my own reference and for trying out code. I have gotten rid of it and will only upload the final codes from now on. As you suggested, I have moved ivim.py into reconst. Next up is writing tests. Will keep updating in smaller commits. 

Thanks.

etpeterson:
Here's an IVIM dataset: https://figshare.com/s/4733b6fc92d4977c2ee1

Let me know if you can't get to it for some reason, I'm still new to figshare so I may not have gotten everything right.

RafaelNH:
Hi @sahmed95! How are things going? Any advances on the simulations/IVIM-fitting (based on [Bihan et al., 1988](http://www.ncbi.nlm.nih.gov/pubmed/3393671))

sahmed95:
Hi @RafaelNH, I forgot to post here. Please find the initial code and simulations here - https://github.com/nipy/dipy/pull/1058

I will work on implementing leastsq and carry out some more simulations today.

RafaelNH:
Cool @sahmed95! I will start reviewing what you did =)

arokem:
Closing in favor of #1058

PR #1056, opened 2016-05-23
Remove tracking interfaces

MrBago:
We missed one file in #1054, re-based to make the commits more clear.


Garyfallidis:
Hi @MrBago There must be a mistake here. You are removing also markov.py and tests/test_markov.py? These are critical parts of the tracking API. Correct me if wrong. Please explain.

MrBago:
markov.py was my first attempt at the tracking API. There isn't anything in
markov.py that didn't make it into either dipy.tracking.local or
dipy.direction. Plus the new stuff is in cython so markov.py isn't
compatible with it.

Garyfallidis:
Okay thanks. Please also give a thumbs up to https://github.com/nipy/dipy/pull/1042

PR #1059, opened 2016-05-25
Documentation to discourage misuse of GradientTable

kesshijordan:
This was misused in the nipype implementation such that the bvals were being set as attributes after the creation of the GradientTable object (so the tensor fit didn't use the bvalues). mrbago suggested I add to the documentation to prevent this in the future.


arokem:
Yeah - I saw your PR over on https://github.com/nipy/nipype/pull/1484. Do you think that clarification of the inputs would be sufficient? That is, instead of what we currently have, rewrite the documentation of the `gradients` input to be: 
  

    gradients : array_like (N, 3)
         Diffusion gradients. The direction of each of these vectors corresponds to the b-vector, and the 
         length corresponds to the b-value.

Or would people still be likely to miss the point?

MrBago:
@arokem Do you think a doc fix is enough here or do you think it's worth trying to "lock down" the instance a little bit? In principle this applies to any class that uses the `@atuo_attr` decorator right? I vote for doc fix, but I think it's worth asking the question.

arokem:
I think that a doc-fix should be enough.

arokem:
Hello @kesshijordan : any more thoughts here?

kesshijordan:
I think a doc fix is good.

On Jul 1, 2016, at 7:36 AM, Ariel Rokem <notifications@github.com<mailto:notifications@github.com>> wrote:


Hello @kesshijordan<https://github.com/kesshijordan> : any more thoughts here?

arokem:
Thanks. And do you think that it would be enough to document the `gradients` parameter (as I suggested here: https://github.com/nipy/dipy/pull/1059#issuecomment-221703759), or do you think we should add a note (as you are proposing here)? Either way, a "note" section should follow after the "parameters" and "returns", not be added in front (as you propose; see numpy docstring standards for details: https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt).

kesshijordan:
I think it's nice to include both, since we aren't "locking down" the instance... the misuse won't, necessarily, return an error that is easily noticed... the user has to know what scale the output image values should be on in order to catch it.  Re formatting: I'll rearrange and commit.

kesshijordan:
How does that look?

arokem:
👍 from me! 

Let's give others a chance to chime in, but if I don't hear anything I'll merge this in a few days.

arokem:
Thanks!

PR #1060, opened 2016-05-26
Fast computation of Cross Correlation metric

omarocegueda:
This PR implements a new algorithm for computing the Cross Correlation metric, which is substantially faster than the original algorithm implemented in ANTS. The computational complexity of the original algorithm for a volume of size N^3 using rectangular windows of size K^3 is Theta(N^3 * K^2). This means that the time increases quadratically with the window radius. 

The computational complexity of this new algorithm is Theta(N^3), it does not depend on the window size, so we observe a substantial speed up, especially for large rectangular windows. After a conversation with @grlee77, the plan would be to add this algorithm instead of #1051, and then after #1050 is in, we can go ahead and use that framework to parallelize this sequential algorithm to obtain even higher speed up.

These are some results on a 4-Core i5-4200H CPU @ 2.80GHz:
Old algorithm, parallel (using #1051):
1 thread
  pre: 1.23728 s
2 threads
  pre: 0.72080 s
4 threads (default)
  pre: 0.69974 s

New algorithm (this PR)
1 thread
  pre: 0.15835 s
2 threads
  pre: 0.15837 s
4 threads (default)
  pre: 0.15925 s
(of course it doesn't depend on the number of threads, it's not parallel right now)

Reference: http://www.sciencedirect.com/science/article/pii/S0167865516300861

### DIFF
--- a/dipy/align/crosscorr.pyx
+++ b/dipy/align/crosscorr.pyx
@@ -136,7 +136,7 @@ def precompute_cc_factors_3d(floating[:, :, :] static,
     Pre-computes the separate terms of the cross correlation metric and image
     norms at each voxel considering a neighborhood of the given radius to
     efficiently compute the gradient of the metric with respect to the
-    deformation field [Ocegueda16][Avants08][Avants11].
+    deformation field [Ocegueda2016]_[Avants2008]_[Avants2011]_.
 
     Parameters
     ----------
@@ -160,16 +160,16 @@ def precompute_cc_factors_3d(floating[:, :, :] static,
 
     References
     ----------
-    [Ocegueda16] Ocegueda, O., Dalmau, O., Garyfallidis, E., Descoteaux, M.,
-               & Rivera, M. (2016), On the computation of integrals over
-               fixed-size rectangles of arbitrary dimension, Pattern
-               Recognition Letters
-    [Avants08] Avants, B. B., Epstein, C. L., Grossman, M., & Gee, J. C. (2008)
-               Symmetric Diffeomorphic Image Registration with
-               Cross-Correlation: Evaluating Automated Labeling of Elderly and
-               Neurodegenerative Brain, Med Image Anal. 12(1), 26-41.
-    [Avants11] Avants, B. B., Tustison, N., & Song, G. (2011).
-               Advanced Normalization Tools ( ANTS ), 1-35.
+    .. [Ocegueda2016] Ocegueda, O., Dalmau, O., Garyfallidis, E., Descoteaux,
+        M., & Rivera, M. (2016). On the computation of integrals over
+        fixed-size rectangles of arbitrary dimension, Pattern Recognition
+        Letters. doi:10.1016/j.patrec.2016.05.008
+    .. [Avants2008] Avants, B. B., Epstein, C. L., Grossman, M., & Gee, J. C.
+        (2008). Symmetric Diffeomorphic Image Registration with
+        Cross-Correlation: Evaluating Automated Labeling of Elderly and
+        Neurodegenerative Brain, Med Image Anal. 12(1), 26-41.
+    .. [Avants2011] Avants, B. B., Tustison, N., & Song, G. (2011). Advanced
+        Normalization Tools ( ANTS ), 1-35.
     """
     cdef:
         cnp.npy_intp ns = static.shape[0]
@@ -354,8 +354,8 @@ def compute_cc_forward_step_3d(floating[:, :, :, :] grad_static,
     r"""Gradient of the CC Metric w.r.t. the forward transformation
 
     Computes the gradient of the Cross Correlation metric for symmetric
-    registration (SyN) [Avants08] w.r.t. the displacement associated to
-    the moving volume ('forward' step) as in [Avants11]
+    registration (SyN) [Avants2008] w.r.t. the displacement associated to
+    the moving volume ('forward' step) as in [Avants2011]
 
     Parameters
     ----------
@@ -378,12 +378,12 @@ def compute_cc_forward_step_3d(floating[:, :, :, :] grad_static,
 
     References
     ----------
-    [Avants08] Avants, B. B., Epstein, C. L., Grossman, M., & Gee, J. C. (2008)
-               Symmetric Diffeomorphic Image Registration with
-               Cross-Correlation: Evaluating Automated Labeling of Elderly and
-               Neurodegenerative Brain, Med Image Anal. 12(1), 26-41.
-    [Avants11] Avants, B. B., Tustison, N., & Song, G. (2011).
-               Advanced Normalization Tools ( ANTS ), 1-35.
+    .. [Avants2008] Avants, B. B., Epstein, C. L., Grossman, M., & Gee, J. C.
+        (2008). Symmetric Diffeomorphic Image Registration with
+        Cross-Correlation: Evaluating Automated Labeling of Elderly and
+        Neurodegenerative Brain, Med Image Anal. 12(1), 26-41.
+    .. [Avants2011] Avants, B. B., Tustison, N., & Song, G. (2011). Advanced
+        Normalization Tools ( ANTS ), 1-35.
     """
     cdef:
         cnp.npy_intp ns = grad_static.shape[0]
@@ -498,10 +498,10 @@ def precompute_cc_factors_2d(floating[:, :] static, floating[:, :] moving,
                              cnp.npy_intp radius):
     r"""Precomputations to quickly compute the gradient of the CC Metric
 
-    Pre-computes the separate terms of the cross correlation metric [Avants08]
-    and image norms at each voxel considering a neighborhood of the given
-    radius to efficiently [Avants11] compute the gradient of the metric with
-    respect to the deformation field.
+    Pre-computes the separate terms of the cross correlation metric
+    [Avants2008] and image norms at each voxel considering a neighborhood of
+    the given radius to efficiently [Avants2011] compute the gradient of the
+    metric with respect to the deformation field.
 
     Parameters
     ----------
@@ -525,12 +525,12 @@ def precompute_cc_factors_2d(floating[:, :] static, floating[:, :] moving,
 
     References
     ----------
-    [Avants08] Avants, B. B., Epstein, C. L., Grossman, M., & Gee, J. C. (2008)
-               Symmetric Diffeomorphic Image Registration with
-               Cross-Correlation: Evaluating Automated Labeling of Elderly and
-               Neurodegenerative Brain, Med Image Anal. 12(1), 26-41.
-    [Avants11] Avants, B. B., Tustison, N., & Song, G. (2011).
-               Advanced Normalization Tools ( ANTS ), 1-35.
+    .. [Avants2008] Avants, B. B., Epstein, C. L., Grossman, M., & Gee, J. C.
+        (2008). Symmetric Diffeomorphic Image Registration with
+        Cross-Correlation: Evaluating Automated Labeling of Elderly and
+        Neurodegenerative Brain, Med Image Anal. 12(1), 26-41.
+    .. [Avants2011] Avants, B. B., Tustison, N., & Song, G. (2011). Advanced
+        Normalization Tools ( ANTS ), 1-35.
     """
     ftype = np.asarray(static).dtype
     cdef:
@@ -665,8 +665,8 @@ def compute_cc_forward_step_2d(floating[:, :, :] grad_static,
     r"""Gradient of the CC Metric w.r.t. the forward transformation
 
     Computes the gradient of the Cross Correlation metric for symmetric
-    registration (SyN) [Avants08] w.r.t. the displacement associated to
-    the moving image ('backward' step) as in [Avants11]
+    registration (SyN) [Avants2008] w.r.t. the displacement associated to
+    the moving image ('backward' step) as in [Avants2011]
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


dipy/align/crosscorr.pyx
```diff
@@ -27,17 +28,115 @@ cdef enum:
...
+def precompute_cc_factors_3d(floating[:, :, :] static,
+                             floating[:, :, :] moving,
+                             cnp.npy_intp radius, num_threads=None):
     r"""Precomputations to quickly compute the gradient of the CC Metric
 
     Pre-computes the separate terms of the cross correlation metric and image
     norms at each voxel considering a neighborhood of the given radius to
     efficiently compute the gradient of the metric with respect to the
-    deformation field [Avants08][Avants11]
+    deformation field [Ocegueda16][Avants08][Avants11].
```
arokem: Can we rewrite these references in the sphinx rst format? 

See for example: https://github.com/nipy/dipy/blob/master/dipy/reconst/dti.py#L1063-L1093

omarocegueda: Thanks @arokem!, =)
I just changed the citations in `crosscorr.pyx`. Should I fix the format of all references in the `align` module on this PR, or make a separate one?

grlee77:
The code for the new algorithm looks good and the performance is great.  I also found the new algorithm proposed here to be faster than the multithreaded version of the old algorithm (in #1051). 

Unlike the previous algorithm, this one is not trivial to make parallel, though.  The main difficulty is that the computations for a given slice, row, or column also make use of previously computed results from the previous slice, row and/or column. If we split slices across threads for instance then we can no longer assume slices are computed sequentially which seems to be required here.  There may be some clever way around this, but it is no longer a trivial case of just using `prange` for the outer loop.

omarocegueda:
Thanks for testing the algorithm @grlee77 !
I just finished a prototype of how to implement this algorithm in parallel. We just assign consecutive slices to each thread. You are right, this algorithm references the previous sums but we can solve this issue by allowing the slice intervals to overlap: at the beginning, each thread must compute the result for the first slice using the direct algorithm. For a volume of NxNxN, computing integrals over windows of KxKxK using T threads, we will be making O(T*K*N^2) total extra work (i.e. T*K extra slices because of the overlap), which will be effective if the window size, K, is not too big and we use just a few cores T.

Here is the kind of performance I am observing:

**First verify that both approaches are equivalent:**
Maximum difference (direct vs. fast-sequential): 0.000000
Maximum difference (direct vs. fast-parallel[1 threads]): 0.000000
Maximum difference (direct vs. fast-parallel[2 threads]): 0.000000
Maximum difference (direct vs. fast-parallel[3 threads]): 0.000000
Maximum difference (direct vs. fast-parallel[4 threads]): 0.000000

**Now check performance (on a computer with 2 physical cores i5-4200H  @ 2.80GHz):**
Fast sequential. Elapsed: 0.352472
1 threads. Elapsed: 0.364174<<<< A bit slower than the sequential algorithm
2 threads. Elapsed: 0.196001<<<< We only have 2 physical cores
3 threads. Elapsed: 0.250694
4 threads. Elapsed: 0.200792

We can see that it is using the 2 cores very efficiently. You can find this proof of concept here:
https://github.com/omarocegueda/fast_integrals_parallel

We can implement this strategy here once your PR is merged =)

dipy/align/crosscorr.pyx
```diff
@@ -27,17 +28,115 @@ cdef enum:
...
     CNT = 5
 
+
 @cython.boundscheck(False)
 @cython.wraparound(False)
 @cython.cdivision(True)
-def precompute_cc_factors_3d(floating[:, :, :] static, floating[:, :, :] moving,
-                             cnp.npy_intp radius):
+cdef inline int _wrap(int x, int m)nogil:
+    r""" Auxiliary function to `wrap` an array around its low-end side
```
arokem: Period at the end of this sentence

dipy/align/crosscorr.pyx
```diff
@@ -27,17 +28,115 @@ cdef enum:
...
+    array's low-end boundary.
+
+    Parameters
+    ----------
+    x : int
+        the array position to be wrapped
+    m : int
+        array length
+    """
+    if x<0:
```
arokem: Whitespace around operator

dipy/align/crosscorr.pyx
```diff
@@ -27,17 +28,115 @@ cdef enum:
...
+        return x + m
+    return x
+
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+@cython.cdivision(True)
+cdef inline void _update_factors(double[:,:,:,:] factors,
```
arokem: Whitespace after commas

dipy/align/crosscorr.pyx
```diff
@@ -27,17 +28,115 @@ cdef enum:
...
+    at (`ss`, `rr`, `cc`) by adding the factors corresponding to voxel
+    (`s`, `r`, `c`) of input images `moving` and `static`.
+
+    Parameters
+    ----------
+    factors : array, shape (S, R, C, 5)
+        array containing the current precomputed factors to be updated
+    moving : array, shape (S, R, C)
+        the moving volume (notice that both images must already be in a common
+        reference domain, i.e. the same S, R, C)
```
arokem: Does this mean they have the same shape?

dipy/align/crosscorr.pyx
```diff
@@ -27,17 +28,115 @@ cdef enum:
...
+        (`operation`=0) the current factors for the rectangular window centered
+        at (`ss`, `rr`, `cc`).
+
+    """
+    cdef:
+        double sval
+        double mval
+    if s>=moving.shape[0] or r>=moving.shape[1] or c>=moving.shape[2]:
+        if operation == 0:
+            factors[ss, rr, cc, SI] = 0
```
arokem: I just noticed that SI, SI2, etc. are cdef'd above, but as far as I can tell, they're not typed. Wouldn't it be advantageous to cdef them as int types?

dipy/align/crosscorr.pyx
```diff
@@ -27,17 +28,115 @@ cdef enum:
...
+def precompute_cc_factors_3d(floating[:, :, :] static,
+                             floating[:, :, :] moving,
+                             cnp.npy_intp radius, num_threads=None):
     r"""Precomputations to quickly compute the gradient of the CC Metric
 
     Pre-computes the separate terms of the cross correlation metric and image
     norms at each voxel considering a neighborhood of the given radius to
     efficiently compute the gradient of the metric with respect to the
-    deformation field [Avants08][Avants11]
+    deformation field [Ocegueda2016]_[Avants2008]_[Avants2011]_.
```
arokem: I think you need spaces between these references.

dipy/align/crosscorr.pyx
```diff
@@ -61,91 +160,134 @@ def precompute_cc_factors_3d(floating[:, :, :] static, floating[:, :, :] moving,
...
+                lastr = _int_min(nr - 1, rr + radius)
+                sider = (lastr - firstr + 1)
+                for c in range(nc+radius):
+                    cc = _wrap(c - radius, nc)
+                    # New corner
+                    _update_factors(temp, moving, static,
+                                       sss, rr, cc, s, r, c, 0)
+
+                    # Add signed sub-volumes
+                    if s>0:
```
arokem: Space around operator

dipy/align/crosscorr.pyx
```diff
@@ -61,91 +160,134 @@ def precompute_cc_factors_3d(floating[:, :, :] static, floating[:, :, :] moving,
...
+                        prev_ss = 1 - sss
+                        for it in range(5):
+                            temp[sss, rr, cc, it] += temp[prev_ss, rr, cc, it]
+                        if r>0:
```
arokem: Ditto

dipy/align/crosscorr.pyx
```diff
@@ -61,91 +160,134 @@ def precompute_cc_factors_3d(floating[:, :, :] static, floating[:, :, :] moving,
...
+                            prev_rr = _wrap(rr-1, nr)
+                            for it in range(5):
+                                temp[sss, rr, cc, it] -= \
+                                    temp[prev_ss, prev_rr, cc, it]
+                            if c>0:
```
arokem: And here

dipy/align/crosscorr.pyx
```diff
@@ -61,91 +160,134 @@ def precompute_cc_factors_3d(floating[:, :, :] static, floating[:, :, :] moving,
...
+                                prev_cc = _wrap(cc-1, nc)
+                                for it in range(5):
+                                    temp[sss, rr, cc, it] += \
+                                        temp[prev_ss, prev_rr, prev_cc, it]
+                        if c>0:
```
arokem: Same

dipy/align/crosscorr.pyx
```diff
@@ -61,91 +160,134 @@ def precompute_cc_factors_3d(floating[:, :, :] static, floating[:, :, :] moving,
...
+                            prev_cc = _wrap(cc-1, nc)
+                            for it in range(5):
+                                temp[sss, rr, cc, it] -= \
+                                    temp[prev_ss, rr, prev_cc, it]
+                    if(r>0):
```
arokem: etc.

dipy/align/crosscorr.pyx
```diff
@@ -61,91 +160,134 @@ def precompute_cc_factors_3d(floating[:, :, :] static, floating[:, :, :] moving,
...
         cnp.npy_intp nc = static.shape[2]
-        cnp.npy_intp s, r, c, k, i, j, t, q, qq, firstc, lastc, firstr, lastr
-        double Imean, Jmean
+        cnp.npy_intp side = 2 * radius + 1
+        cnp.npy_intp firstc, lastc, firstr, lastr, firsts, lasts
+        cnp.npy_intp s, r, c, it, sides, sider, sidec
+        double cnt
+        cnp.npy_intp ssss, sss, ss, rr, cc, prev_ss, prev_rr, prev_cc
+        double Imean, Jmean, IJprods, Isq, Jsq
+        double[:, :, :, :] temp = np.zeros((2, nr, nc, 5), dtype=np.float64)
```
arokem: Should `5` here be `CNT`? That's what it corresponds to, right?

dipy/align/crosscorr.pyx
```diff
@@ -61,91 +160,134 @@ def precompute_cc_factors_3d(floating[:, :, :] static, floating[:, :, :] moving,
...
         floating[:, :, :, :] factors = np.zeros((ns, nr, nc, 5),
```
arokem: Same here: `5` => `CNT`?

dipy/align/crosscorr.pyx
```diff
@@ -212,8 +354,8 @@ def compute_cc_forward_step_3d(floating[:, :, :, :] grad_static,
     r"""Gradient of the CC Metric w.r.t. the forward transformation
 
     Computes the gradient of the Cross Correlation metric for symmetric
-    registration (SyN) [Avants08] w.r.t. the displacement associated to
-    the moving volume ('forward' step) as in [Avants11]
+    registration (SyN) [Avants2008] w.r.t. the displacement associated to
```
arokem: => `[Avants2008]_`

dipy/align/crosscorr.pyx
```diff
@@ -212,8 +354,8 @@ def compute_cc_forward_step_3d(floating[:, :, :, :] grad_static,
...
+    the moving volume ('forward' step) as in [Avants2011]
```
arokem: => `[Avants2011]_`

dipy/align/crosscorr.pyx
```diff
@@ -356,10 +498,10 @@ def precompute_cc_factors_2d(floating[:, :] static, floating[:, :] moving,
...
     r"""Precomputations to quickly compute the gradient of the CC Metric
 
-    Pre-computes the separate terms of the cross correlation metric [Avants08]
-    and image norms at each voxel considering a neighborhood of the given
-    radius to efficiently [Avants11] compute the gradient of the metric with
-    respect to the deformation field.
+    Pre-computes the separate terms of the cross correlation metric
+    [Avants2008] and image norms at each voxel considering a neighborhood of
+    the given radius to efficiently [Avants2011] compute the gradient of the
+    metric with respect to the deformation field.
```
arokem: Same comment here (add underscore, so that references get rendered).

dipy/align/crosscorr.pyx
omarocegueda: Added.

dipy/align/crosscorr.pyx
omarocegueda: Yes, I just edited the docstring to explicitly say that

dipy/align/crosscorr.pyx
omarocegueda: As far as I know, cython enum elements are always integer, like in C. I did a quick search to see if there exists a more general enum implementation using different element types (who knows! people are very creative! =) ), but I couldn't find any.

dipy/align/crosscorr.pyx
omarocegueda: Fixed.

dipy/align/crosscorr.pyx
omarocegueda: `CNT` refers to the index of an array element containing the `count` of voxels within a window. For example when we compute the mean of the `I` image values within the window:
`Imean = sums[SI] / sums[CNT]`
It happens to be equal to 5, but I think it's better to be explicit here.

dipy/align/crosscorr.pyx
omarocegueda: Fixed.

dipy/align/crosscorr.pyx
omarocegueda: Fixed.

dipy/align/crosscorr.pyx
omarocegueda: Fixed.

omarocegueda:
Thank you very much for your thorough review Ariel!, =D

arokem:
Thank you for the quick turnaround! I will merge this early next week, unless some else pipes up.

PR #1062, opened 2016-05-27
BF - fix CSD.predict to work with nd inputs.

MrBago:
This small change should allow the csd.predict method to accept nd-arrays of coefficients.


arokem:
Thanks - looks good. Unless someone objects (including travis -- so far everything seems to be coming up good over there), I will merge this in a couple of days.

PR #1063, opened 2016-05-27
Fixes #1061 : Changed all S0 to 1.0

### LINKED ISSUE #1061: Inconsistency in specifying S0 values in multi_tensor and single_tensor
As pointed out by @etpeterson in https://github.com/nipy/dipy/pull/1058#discussion_r64775504

> Looks like dipy may be a little conflicted also. Single_tensor uses S0=1 and multi_tensor uses S0=100, though others seem to prefer the 100 area.

In dipy/sims/voxel.py multi_tensor specifies S0=100 while single_tensor specifies S0=1

Should it be consistent over both functions ?

sahmed95:
There were specific places where S0=123. which I changed to 100.

### DIFF
--- a/dipy/reconst/mapmri.py
+++ b/dipy/reconst/mapmri.py
@@ -326,7 +326,7 @@ class MapmriFit(ReconstFit):
                     i, 2]) / 2.0) * self._mapmri_coef[i] * Bm[i]
         return const * rtop
 
-    def predict(self, gtab, S0=100.0):
+    def predict(self, gtab, S0=1.):
         """
         Predict a signal for this MapmriModel class instance given a gradient
         table.
--- a/dipy/reconst/mapmri.py
+++ b/dipy/reconst/mapmri.py
@@ -108,7 +108,7 @@ class MapmriModel(ReconstModel):
         >>> data, gtab = dsi_voxels()
         >>> sphere = get_sphere('symmetric724')
         >>> from dipy.sims.voxel import SticksAndBall
-        >>> data, golden_directions = SticksAndBall(gtab, d=0.0015, S0=1, angles=[(0, 0), (90, 0)], fractions=[50, 50], snr=None)
+        >>> data, golden_directions = SticksAndBall(gtab, d=0.0015, S0=100, angles=[(0, 0), (90, 0)], fractions=[50, 50], snr=None)
         >>> from dipy.reconst.mapmri import MapmriModel
         >>> radial_order = 4
         >>> map_model = MapmriModel(gtab, radial_order=radial_order)
@@ -326,7 +326,7 @@ class MapmriFit(ReconstFit):
                     i, 2]) / 2.0) * self._mapmri_coef[i] * Bm[i]
         return const * rtop
 
-    def predict(self, gtab, S0=1.0):
+    def predict(self, gtab, S0=100.0):
         """
         Predict a signal for this MapmriModel class instance given a gradient
         table.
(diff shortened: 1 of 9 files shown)


arokem:
I think we want to leave all the tests as they were. We want to make sure that things work for arbitrary inputs. For example, if a user decides they want to use S0=123 for some reason, it should still work. 

Does that make sense?

sahmed95:
Yes. I think so. That's why I pointed it out as I wasn't sure whether to change the tests. I will keep the tests as they were and update the PR.

Thanks

sahmed95:
@I guess I should break this down into multiple commits ? Makes it easier to review and pinpoint where the S0 values should be changed. 

Or multiple PR's ? @arokem

arokem:
No need for multiple PRs or more commits. This is fine. But there is a test failure that needs to be addressed.

sahmed95:
Ah, @arokem I made by pushing without commiting

arokem:
@sahmed95 : did you mean to close this PR? It was going in the right direction, and I think we should keep going here, unless you have an alternative plan.

arokem:
Oh - cool. I see you reopened this right as I sent my message. All good.

sahmed95:
Sorry, that was accidental. I had rebased with master and started fresh and forgot to commit the changes before a forced push. I have made the changes and tests pass locally. There is one place though which still has S0 = 1.

arokem:
Any chance to revisit this one, and complete it?

sahmed95:
Hi, seems like S0 = 1. is the preferred choice in most tests. So, I am setting S0 as 1.

sahmed95:
This is up for review.

dipy/reconst/mapmri.py
```diff
@@ -108,7 +108,7 @@ def __init__(self,
         >>> data, gtab = dsi_voxels()
         >>> sphere = get_sphere('symmetric724')
         >>> from dipy.sims.voxel import SticksAndBall
-        >>> data, golden_directions = SticksAndBall(gtab, d=0.0015, S0=1, angles=[(0, 0), (90, 0)], fractions=[50, 50], snr=None)
+        >>> data, golden_directions = SticksAndBall(gtab, d=0.0015, S0=100, angles=[(0, 0), (90, 0)], fractions=[50, 50], snr=None)
```
arokem: While we're at it, would you mind making this PEP8 compliant, by introducing a line-break? 

It should look like this: 

>>> data, golden_directions = SticksAndBall(gtab, d=0.0015, S0=100,
... angles=[(0, 0), (90, 0)], fractions=[50, 50], snr=None)

arokem: Trying again: 

    >>> data, golden_directions = SticksAndBall(gtab, d=0.0015, S0=100,
    ... angles=[(0, 0), (90, 0)], fractions=[50, 50], snr=None)

arokem: Yes - like my second attempt (hard to indent and align things in this GH textbox...).

arokem:
It needs 3 dots (`...`) at the beginning of every line in the docstring example. Just like you would get if you ran this in an interactive python session, and put in a line-break there.

arokem:
As in the example I gave in my comment above.

arokem:
You can run the docstring example tests by running
    
    nosetests --with-doctest

sahmed95:
This should do it. I picked up some more pep8's on the way. I hope that's okay.

arokem:
Thanks!

PR #1067, opened 2016-06-02
Generate b vectors using disperse_charges

sahmed95:
A function to generate arbitary b vectors using the disperse function

### DIFF
--- a/dipy/core/gradients.py
+++ b/dipy/core/gradients.py
@@ -12,6 +12,7 @@ from scipy.linalg import inv
 from dipy.io import gradients as io
 from dipy.core.onetime import auto_attr
 from dipy.core.geometry import vector_norm
+from dipy.core.sphere import disperse_charges, HemiSphere
 
 
 class GradientTable(object):
@@ -48,6 +49,7 @@ class GradientTable(object):
     gradient_table
 
     """
+
     def __init__(self, gradients, big_delta=None, small_delta=None,
                  b0_threshold=0):
         """Constructor for GradientTable class"""
@@ -308,3 +310,16 @@ def reorient_bvecs(gtab, affines):
     return_bvecs = np.zeros(gtab.bvecs.shape)
     return_bvecs[~gtab.b0s_mask] = new_bvecs
     return gradient_table(gtab.bvals, return_bvecs)
+
+
+def generate_b_vectors(N):
+    """Function to generate arbitary b-vectors of length N
+    This may be used for testing or various simulations.
+    It uses the disperse_charges function from dipy.core.sphere
+    Refer to the example gradients_spheres for details"""
+    theta = np.pi * np.random.rand(N)
+    phi = 2 * np.pi * np.random.rand(N)
+    hsph_initial = HemiSphere(theta=theta, phi=phi)
+    hsph_updated, potential = disperse_charges(hsph_initial, 5000)
+    vertices = hsph_updated.vertices
+    return vertices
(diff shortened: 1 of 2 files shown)


arokem:
Thanks! Would be good to have a test for this.

dipy/core/gradients.py
```diff
@@ -308,3 +310,16 @@ def reorient_bvecs(gtab, affines):
     return_bvecs = np.zeros(gtab.bvecs.shape)
     return_bvecs[~gtab.b0s_mask] = new_bvecs
     return gradient_table(gtab.bvals, return_bvecs)
+
+
+def generate_b_vectors(N):
+    """Function to generate arbitary b-vectors of length N
+    This may be used for testing or various simulations.
+    It uses the disperse_charges function from dipy.core.sphere
+    Refer to the example gradients_spheres for details"""
```
arokem: Docstring needs to be reformatted to the standard format

dipy/core/gradients.py
```diff
@@ -308,3 +310,16 @@ def reorient_bvecs(gtab, affines):
...
+    theta = np.pi * np.random.rand(N)
+    phi = 2 * np.pi * np.random.rand(N)
+    hsph_initial = HemiSphere(theta=theta, phi=phi)
+    hsph_updated, potential = disperse_charges(hsph_initial, 5000)
```
arokem: This number (5000) should be a key-word argument to the function.

sahmed95:
I am wondering how can we test this ? Won't the b vectors be in arbitrary directions ?

arokem:
They will be, but there are still properties of the results that we can check. For example, for `N=2`, the vectors should be just about 90 degrees from each other, so you can check the angle between the resulting vectors.

grlee77:
another test idea:
upcoming scipy 0.18 will have [spherical Voronoi diagram support](http://scipy.github.io/devdocs/generated/scipy.spatial.SphericalVoronoi.html).  That would allow you to verify that each point represents an approximately equal area on the spherical surface.

sahmed95:
@arokem I added two tests, one to verify we are getting unit vectors and the other to check orthogonality when we generate two vectors.

@grlee77 Thanks. I will keep an eye on the latest update and add tests accordingly.

RafaelNH:
Simple but useful! Thanks @sahmed95! For my side, this can be merged!

RafaelNH:
Well It can be merged after resolving the issue with the compatibility with numpy 1.7.1 =P

sahmed95:
:-( I am using numpy 1.9.1 and its passing on my system since linalg.norm accepts the axis parameter. I see that Travis (current dipy?) uses version 1.7.1. I can simply calculate the norm without using numpy.

Should I do that or let this wait till dipy updates dependency to 1.9.1 ?

arokem:
The reason we test with older versions of numpy is that we want scientists
to be able to use Dipy, even if they are reluctant to upgrade numpy to
newer versions, because their other software is not up to date. If there is
a way to rewrite the code so that it works on numpy 1.7 that would be best.

etpeterson:
Looks like it may be fixed now but a 3D vector norm isn't hard to do yourself either `(bvecs[:,0]**2+bvecs[:,1]**2+bvecs[:,2]**2)**0.5`

sahmed95:
@etpeterson Thanks ! I wanted to use linalg.norm as I thought that it might be faster. The current commit uses a list comprehension [np.linalg.norm(vec) for vec in bvecs]

arokem:
This looks good to me. @Garyfallidis: is this the right place for this function? Do you have any opinions? If we don't hear back, we'll assume you're fine with this, and I will go ahead with the merge.

PR #1068, opened 2016-06-03
Renamed test_shore for consistency

sahmed95:
Renamed the test for shore in dipy/reconst/tests/ for consistency.


arokem:
Yep. That makes sense.

PR #1075, opened 2016-06-09
Drop26

arokem:
Addresses https://github.com/nipy/dipy/issues/940

No hurry to do this (no problems there, as far as I can tell), but might be good to do this at some point.


arokem:
This might be worth reconsidering. Any thoughts here @Garyfallidis ? @matthew-brett : you've already dropped 2.6 support on nibabel, is that correct?

arokem:
At the very least it might shorten our Travis runs a little bit.

PR #1058, opened 2016-05-25
IVIM

sahmed95:
Initial simulations of IVIM

Notebook with examples
https://github.com/sahmed95/notebooks/blob/ivim/ivim_dev.ipynb

PR #1078, opened 2016-06-12
Fetcher for ivim data, needs md5

sahmed95:
Needs the hashes, but otherwise good to go. This is for the upcoming ivim module currently being developed in https://github.com/nipy/dipy/pull/1058.


arokem:
This looks good to me, but would you mind merging this into #1055 instead of having us merge this directly into master at this point? That way, we can prepare everything up to and including the IVIM example on that branch, and then merge it all together.

sahmed95:
@arokem I think you meant #1058

arokem:
I did indeed!

sahmed95:
This is included in #1058

PR #1079, opened 2016-06-15
Return a generator from `orient_by_roi`

arokem:
This is more consistent with other API for handling streamlines, and also makes more sense.

### DIFF
--- a/dipy/tracking/streamline.py
+++ b/dipy/tracking/streamline.py
@@ -260,26 +260,68 @@ def select_by_rois(streamlines, rois, include, mode=None, affine=None,
             yield sl
 
 
-def orient_by_rois(streamlines, roi1, roi2, affine=None, copy=True):
+def _orient_generator(out, roi1, roi2):
+    """
+    Helper function to `orient_by_rois`
+
+    Performs the inner loop separately. This is needed, because functions with
+    `yield` always return a generator
+    """
+    for idx, sl in enumerate(out):
+        dist1 = cdist(sl, roi1, 'euclidean')
+        dist2 = cdist(sl, roi2, 'euclidean')
+        min1 = np.argmin(dist1, 0)
+        min2 = np.argmin(dist2, 0)
+        if min1[0] > min2[0]:
+            yield sl[::-1]
+        else:
+            yield sl
+
+
+def _orient_list(out, roi1, roi2):
+    """
+    Helper function to `orient_by_rois`
+
+    Performs the inner loop separately. This is needed, because functions with
+    `yield` always return a generator.
+    """
+    for idx, sl in enumerate(out):
+        dist1 = cdist(sl, roi1, 'euclidean')
+        dist2 = cdist(sl, roi2, 'euclidean')
+        min1 = np.argmin(dist1, 0)
+        min2 = np.argmin(dist2, 0)
+        if min1[0] > min2[0]:
+            out[idx] = sl[::-1]
+    return out
+
+
+def orient_by_rois(streamlines, roi1, roi2, out=None, as_generator=False,
+                   affine=None):
     """Orient a set of streamlines according to a pair of ROIs
 
     Parameters
     ----------
-    streamlines : list
-        List of 3d arrays. Each array contains the xyz coordinates of a single
-        streamline.
+    streamlines : list or generator
+        List or generator of 2d arrays of 3d coordinates. Each array contains
+        the xyz coordinates of a single streamline.
     roi1, roi2 : ndarray
         Binary masks designating the location of the regions of interest, or
         coordinate arrays (n-by-3 array with ROI coordinate in each row).
+    out : list
+        A list to assign the output into (for example, this could be the
+        input `streamlines`, in which case the change will be made in-place).
+        If None (default), a copy of the original input is made and that copy
+        is returned, so that the input is retained
+    as_generator : bool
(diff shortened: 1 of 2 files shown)


MarcCote:
So you cannot orient the streamlines in-place anymore? If so, do you still need the `copy` parameter?

arokem:
Good point. What do you think? Should I implement a separate logic for in-place reorienting, in case the input is a list? We can make this function a little bit more flexible, I think. Currently, if you pass this function a generator, it chokes because you can't copy a generator. I stumbled on that during my work with this function, which is why I implemented this change.

MarcCote:
Maybe it could have an `out` parameter (like [Numpy's dot](http://docs.scipy.org/doc/numpy/reference/generated/numpy.dot.html)) instead of `copy`. In addition, it could have a `as_generator` parameter.

```python
orient_by_rois(streamlines, roi1, roi2)  # returns a copy
orient_by_rois(streamlines, roi1, roi2, out=streamlines)  # in-place
orient_by_rois(streamlines, roi1, roi2, as_generator=True)  # returns a generator
orient_by_rois(gen_streamlines, roi1, roi2)  # returns a generator?
orient_by_rois(gen_streamlines, roi1, roi2, as_generator=True)  # returns a generator
orient_by_rois(gen_streamlines, roi1, roi2, out=gen_streamlines)  # will raise
```

arokem:
Yes, I like that. I'll ping again (or it seems that GH will send you an
email), when I have that implemented.

On Thu, Jun 16, 2016 at 11:45 AM, Marc-Alexandre Côté <

> Maybe it could have an out parameter (like Numpy's dot
> <http://docs.scipy.org/doc/numpy/reference/generated/numpy.dot.html>)
> instead of copy. In addition, it could have a as_generator parameter.
>
> orient_by_rois(streamlines, roi1, roi2)  # returns a copy
> orient_by_rois(streamlines, roi1, roi2, out=streamlines)  # in-place
> orient_by_rois(streamlines, roi1, roi2, as_generator=True)  # returns a generator
> orient_by_rois(gen_streamlines, roi1, roi2)  # returns a generator?
>
> I'd say "yes" -- return a generator in this case. That is, keep the output
consistent with the input.

arokem:
OK - I think I got all the cases you outlined. Does the documentation seem sufficient?

dipy/tracking/streamline.py
```diff
@@ -260,26 +260,68 @@ def select_by_rois(streamlines, rois, include, mode=None, affine=None,
...
+        else:
+            yield sl
+
+
+def _orient_list(out, roi1, roi2):
+    """
+    Helper function to `orient_by_rois`
+
+    Performs the inner loop separately. This is needed, because functions with
+    `yield` always return a generator.
```
MarcCote: Maybe mention this function flips streamlines in-place and returns a reference to the updated list?

dipy/tracking/streamline.py
```diff
@@ -312,21 +353,23 @@ def orient_by_rois(streamlines, roi1, roi2, affine=None, copy=True):
...
+        return _orient_generator(streamlines, roi1, roi2)
 
-    for idx, sl in enumerate(streamlines):
-        dist1 = cdist(sl, roi1, 'euclidean')
-        dist2 = cdist(sl, roi2, 'euclidean')
-        min1 = np.argmin(dist1, 0)
-        min2 = np.argmin(dist2, 0)
-        if min1[0] > min2[0]:
-            yield sl[::-1]
+    else:
```
MarcCote: This `else` statement could be removed and its content could be dedented.

dipy/tracking/streamline.py
```diff
@@ -312,21 +353,23 @@ def orient_by_rois(streamlines, roi1, roi2, affine=None, copy=True):
...
+        # If it's a generator on input, we may as well generate it
+        # here and now:
+        if isinstance(streamlines, types.GeneratorType):
+            out = list(streamlines)
+        elif out is None:
+            # Make a copy, so you don't change the output in place:
+            out = deepcopy(streamlines)
+        else:
+            out = streamlines
```
MarcCote: From what I see, a user can provide `out` but it is always overwritten. Am I missing something?

dipy/tracking/tests/test_streamline.py
```diff
@@ -791,25 +805,103 @@ def test_orient_by_rois():
...
+                                     mask2_vol,
+                                     out=None,
+                                     affine=None,
+                                     as_generator=True)
+
+    npt.assert_(isinstance(new_streamlines, types.GeneratorType))
+    ll = list(new_streamlines)
+    npt.assert_equal(ll, flipped_sl)
+
+    # Generator input with an `out` set to something raises an error:
```
MarcCote: Generator input with an `out` set to something should **not** raise an error. However, `as_generator=True` and `out` set to something should raise an error. I think that is what you are already doing in the code anyway.

dipy/tracking/streamline.py
arokem: On further thought, I think that an `out` kwarg doesn't make sense. It would have to be a list of exactly the same length as the original streamlines. May as well have an `inplace` kwarg that would just do that.

arokem:
Thanks for all the comments! I made another pass through, implementing the `in_place` kwarg.

MarcCote:
LGTM

PR #1082, opened 2016-06-22
Add documentation uploading script

ghoshbishakh:
upload_docs.py script builds documentation for the current version and uploads it to gh-pages branche of the dipy_web repository

### DIFF
--- a/doc/upload_docs.py
+++ b/doc/upload_docs.py
@@ -3,6 +3,7 @@
 # automatically detected by the dipy website.
 import os
 import re
+import sys
 from os import chdir as cd
 from subprocess import check_call
 
@@ -45,7 +46,25 @@ if __name__ == '__main__':
         sh("git clone %s %s" % (docs_repo_url, docs_repo_path))
     cd(docs_repo_path)
     print("Moved to " + os.getcwd())
-    sh("git checkout gh-pages")
+    try:
+        sh("git checkout gh-pages")
+    except:
+        while(1):
+            print("\nLooks like gh-pages branch does not exist!")
+            print("Do you want to create a new one? (y/n)")
+            choice = str(input()).lower()
+            if choice == 'y':
+                sh("git checkout -b gh-pages")
+                sh("rm -rf *")
+                sh("git add .")
+                sh("git commit -m 'cleaning gh-pages branch'")
+                sh("git push origin gh-pages")
+                break
+            if choice == 'n':
+                print("Please manually create a new gh-pages branch and try again.")
+                sys.exit(0)
+            else:
+                print("Please enter valid choice ..")
     sh("git pull origin gh-pages")
 
     # check if docs for current version exists
--- a/doc/upload_docs.py
+++ b/doc/upload_docs.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
-
+# Script to upload docs to gh-pages branch of dipy_web that will be
+# automatically detected by the dipy website.
 import os
 import re
 from os import chdir as cd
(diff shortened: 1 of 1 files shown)


sahmed95:
@ghoshbishakh 
Hi, looks like you're working on documentation. Can you please have a look at #1022. 

Running "make ipynbexamples" from the /doc folder generates all the "valid_examples" as .ipynb files in the folder "ipython_notebooks". The notebooks aren't executed though and all that this script does is convert the .py files in the "examples" directory to Ipython notebooks. This is pertaining to #720

ghoshbishakh:
@sahmed95 So we have to include it in the documentation somehow?

sahmed95:
As @arokem mentioned in #720 

>> And vice versa: enable writing examples in IPython notebooks, which then get converted into the compliant python + rst to get built into the documentation.
 :+1: 1  

>> However, I would prefer that we automate this, such that the notebooks get automatically generated when we build the documentation. That is, rather than committing all these notebooks to the repository, we would have a script that converts the python examples to jupyter notebooks, and uploads them together with the python example scripts, so they can be downloaded. Even better, a script that would automatically create/update a gist on github and create the nbviewer page would be fantastic.

>> Would it be possible to do this as a separate instruction in the doc Makefile, rather than part of running the examples?

>> For now, all I am suggesting is to automate things so that when we run make html it not only produces the .rst versions of the examples, but also the .ipynb versions. 

So I wrote the script such that running ```make ipynbexamples``` generates the ipython notebooks.

sahmed95:
I guess, the target is to build the ipython notebooks as one runs ```make html``` and produce the .rst versions of the examples and .ipynb notebooks and upload them. 

Currently ipython notebooks are converted by running ```make ipynbexamples```. So I guess you can help out by making this a part of ```make html``` and uploading the ipynb files as well after converting to .rst perhaps.

ghoshbishakh:
Okey.. understood. I will start working on it.

sahmed95:
Thanks for taking this up.

ghoshbishakh:
@Garyfallidis @jchoude In Scikit image and Django website the docs are listed according to release versions and there is one version called "dev".
So in our case should I check if the version number has ".dev" at the end and put it as the latest dev version?

jchoude:
Yes I think that would be the bvest. However, I would default to showing the doc for te latest release as the default, since that is what most users will install.

jchoude:
Hi @ghoshbishakh this is still set as WIP. What's the status?

ghoshbishakh:
@jchoude It is actually complete. May be we can merge it and make separate PRs later for bug fixes?

ghoshbishakh:
@Garyfallidis can this be merged? If further changes are required in future then I will make separate PRs.

doc/upload_docs.py
```diff
@@ -0,0 +1,80 @@
...
+def sh(cmd):
+    """Execute command in a subshell, return status code."""
+    print("--------------------------------------------------")
+    print("Executing: %s" % (cmd, ))
+    print("--------------------------------------------------")
+    return check_call(cmd, shell=True)
+
+# paths
+docs_repo_path = "_build/docs_repo"
+docs_repo_url = "git@github.com:ghoshbishakh/dipy_web.git"
```
jchoude: Should probably point to nipy/dipy_web.git

jchoude:
Except for my last comment, for me this is +1

ghoshbishakh:
@jchoude I have changed the repository url

jchoude:
Seems good to me. @ghoshbishakh Thanks.

@arokem @Garyfallidis anyone wants to have a quick look? Else I'll merge by the end of the week.

jchoude:
@ghoshbishakh Two things

1) Can you rebase on master, please?

2) Upon final review and testing, I realize that changing the repo_url to point to the normal dipy_web repo causes an issue: the gh-pages branch doesn't currently exist there. Can you please fix?

ghoshbishakh:
@jchoude I have rebased it on master. And added some error handeling.

Garyfallidis:
This looks good. Thanks for addressing the comments!

PR #1086, opened 2016-06-22
ENH: avoid calling log() on zero-valued elements in anisotropic_power

jchoude:
In reconst.shm.anisotropic_power, there is a log() normalization, to bring back values to a reasonable range.

However, for some voxels (for example, when outside of the brain), the pre-normalized AP values are already worth 0. In this case, applying the log() call would generate a numpy warning, that could be misleading for end users.

In this patch, we simply apply the log only to elements that are > 0. Guaranteed to be >= 0, since it's a sum of absolute values.

Also, if the non_negative kw arg is set to False, those 0-valued voxels get "normalized" to -inf, which is not a really useful value. They can also create other issues if those maps are further post-processed.

Finally, a few quick typos were fixed in the doc for this function.

### DIFF
--- a/dipy/reconst/shm.py
+++ b/dipy/reconst/shm.py
@@ -1037,7 +1037,8 @@ def anisotropic_power(sh_coeffs, norm_factor=0.00001, power=2,
     A l=2 SH coefficient matrix will then be composed of a IxJxKx6 volume.
     The power, $n$ is usually set to $n=2$.
 
-    The final AP image is then shifted by -log(norm_factor), to be strictly non-negative. Remaining values < 0 are discarded (set to 0), per default,
+    The final AP image is then shifted by -log(norm_factor), to be strictly
+    non-negative. Remaining values < 0 are discarded (set to 0), per default,
     and this option is controlled through the `non_negative` keyword argument.
 
     References
--- a/dipy/reconst/shm.py
+++ b/dipy/reconst/shm.py
@@ -1010,7 +1010,7 @@ def anisotropic_power(sh_coeffs, norm_factor=0.00001, power=2,
     ----------
     sh_coeffs : ndarray
         A ndarray where the last dimension is the
-        SH coeff estimates for that voxel.
+        SH coefficients estimates for that voxel.
     norm_factor: float, optional
         The value to normalize the ap values. Default is 10^-5.
     power : int, optional
@@ -1026,7 +1026,7 @@ def anisotropic_power(sh_coeffs, norm_factor=0.00001, power=2,
 
     Notes
     ----------
-    Calculate AP image based on a IxJxKxC SH coeffecient matrix based on the
+    Calculate AP image based on a IxJxKxC SH coefficient matrix based on the
     equation:
     .. math::
         AP = \sum_{l=2,4,6,...}{\frac{1}{2l+1} \sum_{m=-l}^l{|a_{l,m}|^n}}
@@ -1034,11 +1034,11 @@ def anisotropic_power(sh_coeffs, norm_factor=0.00001, power=2,
     Where the last dimension, C, is made of a flattened array of $l$x$m$
     coefficients, where $l$ are the SH orders, and $m = 2l+1$,
     So l=1 has 1 coeffecient, l=2 has 5, ... l=8 has 17 and so on.
-    A l=2 SH coeffecient matrix will then be composed of a IxJxKx6 volume.
+    A l=2 SH coefficient matrix will then be composed of a IxJxKx6 volume.
     The power, $n$ is usually set to $n=2$.
 
-    The final AP image is then shifted by -log(normal_factor), to be strictly non-negative. Remaining values < 0 are discarded (set to 0), per default,
-    and this option is controlled throug the `non_negative` key word argument.
+    The final AP image is then shifted by -log(norm_factor), to be strictly non-negative. Remaining values < 0 are discarded (set to 0), per default,
+    and this option is controlled through the `non_negative` keyword argument.
 
     References
     ----------
--- a/dipy/reconst/shm.py
+++ b/dipy/reconst/shm.py
@@ -1062,8 +1062,9 @@ def anisotropic_power(sh_coeffs, norm_factor=0.00001, power=2,
         n_start = n_stop
 
     # Shift the map to be mostly non-negative,
-    # only applying the log operation to non-zero elements
+    # only applying the log operation to positive elements
     # to avoid getting numpy warnings on log(0).
+    # It is impossible to get ap values smaller than 0.
     # Also avoids getting voxels with -inf when non_negative=False.
     log_ap = np.zeros_like(ap)
     log_ap[ap > 0] = np.log(ap[ap > 0]) - np.log(norm_factor)
--- a/dipy/reconst/shm.py
+++ b/dipy/reconst/shm.py
@@ -1060,8 +1060,12 @@ def anisotropic_power(sh_coeffs, norm_factor=0.00001, power=2,
         ap += ap_i
         n_start = n_stop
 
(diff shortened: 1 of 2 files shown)


dipy/reconst/shm.py
```diff
@@ -1026,19 +1026,19 @@ def anisotropic_power(sh_coeffs, norm_factor=0.00001, power=2,
...
     Where the last dimension, C, is made of a flattened array of $l$x$m$
     coefficients, where $l$ are the SH orders, and $m = 2l+1$,
     So l=1 has 1 coeffecient, l=2 has 5, ... l=8 has 17 and so on.
-    A l=2 SH coeffecient matrix will then be composed of a IxJxKx6 volume.
+    A l=2 SH coefficient matrix will then be composed of a IxJxKx6 volume.
     The power, $n$ is usually set to $n=2$.
 
-    The final AP image is then shifted by -log(normal_factor), to be strictly non-negative. Remaining values < 0 are discarded (set to 0), per default,
-    and this option is controlled throug the `non_negative` key word argument.
+    The final AP image is then shifted by -log(norm_factor), to be strictly non-negative. Remaining values < 0 are discarded (set to 0), per default,
```
arokem: While we're already here, would you mind wrapping this to 80 characters?

dipy/reconst/shm.py
```diff
@@ -1060,8 +1060,12 @@ def anisotropic_power(sh_coeffs, norm_factor=0.00001, power=2,
         ap += ap_i
         n_start = n_stop
 
-    # Shift the map to be mostly non-negative:
-    log_ap = np.log(ap) - np.log(norm_factor)
+    # Shift the map to be mostly non-negative,
+    # only applying the log operation to non-zero elements
```
arokem: `non-negative`. There might still be negative values here, right?

dipy/reconst/shm.py
```diff
@@ -1060,8 +1060,12 @@ def anisotropic_power(sh_coeffs, norm_factor=0.00001, power=2,
...
+    # to avoid getting numpy warnings on log(0).
+    # Also avoids getting voxels with -inf when non_negative=False.
+    log_ap = np.zeros_like(ap)
+    log_ap[ap > 0] = np.log(ap[ap > 0]) - np.log(norm_factor)
```
MarcCote: This is more for my intellectual curiosity, does anyone knows if there is any substantial speedup of caching this boolean indexing (i.e. `ap > 0`) or does Python will optimize it anyway?

arokem: I don't know whether Python has some optimization in place, but there seems to be only a small advantage to caching: 
https://gist.github.com/arokem/fba319fedf2b541240504f56f97af45e

dipy/reconst/shm.py
jchoude: Shouldn't be. ap is defined as sum of absolute values of the SH coefficients. Since we have an abs in the sum, it should never be negative.

While we are there, do we want a check on the value of the power argument, or we allow anything?

dipy/reconst/shm.py
jchoude: I had made quick tests to see if it was worth it, and I also didn't see a significant improvement. However, I'm no expert on the internals of Numpy or python, so there might be some magic caching behind the scenes, or an optimized accessor.

MarcCote: Ok, good to know, thanks.

dipy/reconst/shm.py
arokem: What about `ap` values between 0 and 1? These would result in negative `log(ap)`

jchoude: Negative log(ap) is not really a problem, for 2 reasons:

1. The ```-log(norm_factor)``` will shift most of them back to positive values, as long as the norm_factor stays reasonable. For example, the default value is 0.00001, for which the ```-log(norm_factor)``` will result in ```+11.5129```, shifting all values by that value.

2. If the ```non_negative``` arg is set to True (which is the default), there is the additional check later in the function that can handle all remaining negavtive values.

arokem: Yes - I agree with everything you say:I just think that the comment is inaccurate, considering the code below. 

OTOH, that's not crucial. Apart from the line-wrapping (which is really my bad...), I think this is ready to go.

arokem:
Hmm. Travis is not happy: https://travis-ci.org/nipy/dipy/jobs/139731117#L2267

This looks suspiciously similar to https://github.com/nipy/dipy/issues/1074, but that might just be a red herring. Seems to be happening only with older numpy.

dipy/reconst/shm.py
jchoude: I pushed the line-wrapping fix, except if you,re talking about some other line...

Also, maybe I misunderstood... What part of the comment is inaccurate? That we apply only to non_zero voxels? Sorry I'm just not able to pinpoint directly which part is inaccurate.

arokem: Yes - somehow the changes I was looking at didn't have your PEP8 commit. Thanks for doing that.

Non-zero would be:

    log_ap[ap != 0] = np.log(ap[ap != 0]) - np.log(norm_factor)

As it is, the code selects for non-negative. But really - this is not crucial - as you pointed up above.

jchoude:
Yes, that's weird for Travis. I manually ran the test, didn't have this issue. I'm not sure how to see the version of numpy that is used... I see that, at the beginning of the log, it says that it depends on Numpy 1.7.1. Does it mean it is that version that is used?

dipy/reconst/shm.py
jchoude: Ah! I get it. Well, it is true that the condition is non_zero. It is impossible to have values of ```ap``` that are negative, and we only want to avoid 0s, since those are the ones that throw the log warning.

I agree that with small ap values between 0 and 1, the log will yield a negative value. As I said, it will be handled by the rest of the statement and later code.

I'll push a quick fix to the comment.

arokem:
Yes, numpy 1.7.1

As you can see here: https://travis-ci.org/nipy/dipy/builds/139731111, only that machine fails. This is a machine set up to support the minimal versions of dependencies that we support. It looks to me that the 0d array arises, because we are dereferencing the array in the test: 

https://github.com/nipy/dipy/blob/master/dipy/reconst/tests/test_shm.py#L440

arokem:
Any more thoughts here? Looks like we're still stuck on that test failure.

arokem:
Yo @jchoude : have you had a chance to look at this one again?

jchoude:
@arokem I hadn't until now. I pushed a potential fix, waiting on travis to do its magic.

Then, I'll rebase.

jchoude:
@arokem  so, after more investigation, even using a ```np.asarray()``` in the test doesn't work for this. The thing is that we want to test if it works in the single voxel case, and therefore send a 1D array to the ```anisotropic_power()``` function. In there, we derive the shape of the AP array, and then index in there.

The issue is that, at least with numpy 1.10.1, indexing using code like ```ap[ap > 0]``` works even if ```ap``` is a single scalar. There is probably an internal numpy cast as an array. For numpy 1.7, this doesn't work, hence the error that we see.

In this case, what would be your preferred course of action?

- One basic choice, but probably will be considered not possible, is to support numpy 1.9.2 and up only, since the issue is fixed in 1.9.2.
- Add a special code path / in anisotropic_power to correctly behave in the face of a single scalar value.

What would you prefer? Any other idea?

arokem:
Hi @jchoude : apologies for the long gap here. I'd prefer to special-case it in the code. It's a bit of a weird edge-case, but since we're testing it, I think that's probably fine.

jchoude:
Closing since it was superseded by pull request #1198

Issue #1083, opened 2016-06-22
nlmeans problem


mdesco:
The nlmeans version of 0.12dev is considerably different from previous releases. As a result, our in-house scripts in my lab and at the hospital using the current nlmeans of the 0.12dev gives "broken" and unsatisfactory results. The nlmeans is over-blurring heavily at the moment. 

Attached is a figure comparing the 0.11 nlmeans result verus 0.12dev result with the same sigma value. 
![nlmeans_broken](https://cloud.githubusercontent.com/assets/2170887/16261411/b8f5f002-386b-11e6-9ee9-88c997db6d2f.png)

I have a very hard time to read the code (https://github.com/nipy/dipy/blob/master/dipy/denoise/denspeed.pyx) but here are some potential explanations:
- there is a new search region size but the denominator where the sigma value may not have been correctly updated
- there is something fishy going on with the sigma

I see this comes from PR https://github.com/nipy/dipy/pull/1018 and this issue: https://github.com/nipy/dipy/issues/1011

One has to be very careful changing the default behaviour of patch size and block size. The previous nlmeans version was really mimicking the behaviour of the C++ and Matlab implementation of the its inventor Pierrick Coupé, Nicolas Wiest-Daesslé that goes ways back to our miccai papers from 2007 to 2008.

Garyfallidis:
Hi @mdesco and thank you for the feedback. I think that most likely there is nothing broken. I think what happened is that Riddhish (@riddhishb) improved our current implementation and now we can have adaptive nlmeans too which is even better than standard nlmeans in many ways. Riddhish is coming from a lab with expertise in denoising and his supervisor was part of the MNI denoising team back in the days when these methods were created (before 2008). 

Probably this is again an issue of normalization/sigma estimation but that is a problem in the recent papers that didn't follow the standards for previous non-local denoising papers. Probably not a problem that Riddhish introduced. 

I will assign this PR to Riddhish as he is the main denoising expert right  now and hopefully you can both decide what is the correct normalization/sigma or any other issue. 

I think the addition of the blockwise version does not affect the patch-based version.

Finally, it makes sense that you cannot immediately figure out the code as it has been severely optimized  for speed. I hope this helps. @riddhishb because this affects the behavior of what is in master we will need to prioritize fixing or explain it to @mdesco the change and of course document it.

riddhishb:
[...]

PR #1088, opened 2016-06-24
Restored the older implementation of nlmeans

riddhishb:
The changes proposed in #1018 for the nlmeans function are now omitted, after the discussion in #1083. This results in still having the corner case mentioned in #1011 
@Garyfallidis Do have a look.

### DIFF
--- a/dipy/denoise/denspeed.pyx
+++ b/dipy/denoise/denspeed.pyx
@@ -104,7 +104,7 @@ def _nlmeans_3d(double[:, :, ::1] arr, double[:, :, ::1] mask,
     # move the block
     with nogil, parallel():
 
-        for i in prange(B, I - B):
+        for i in prange(B, I - B, schedule="dynamic"):
             for j in range(B, J - B):
                 for k in range(B, K - B):
 
--- a/dipy/denoise/denspeed.pyx
+++ b/dipy/denoise/denspeed.pyx
@@ -19,7 +19,6 @@ def nlmeans_3d(arr, mask=None, sigma=None, patch_radius=1,
                block_radius=5, rician=True, num_threads=None):
     """ Non-local means for denoising 3D images
 
-
     Parameters
     ----------
     arr : 3D ndarray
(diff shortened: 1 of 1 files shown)


dipy/denoise/denspeed.pyx
```diff
@@ -108,9 +103,10 @@ def _nlmeans_3d(double[:, :, ::1] arr, double[:, :, ::1] mask,
 
     # move the block
     with nogil, parallel():
-        for i in prange(B + P, I - B - P, schedule="dynamic"):
```
arokem: I believe that you can retain this change (introduced here: https://github.com/nipy/dipy/commit/4ee6d49a03474b512a6c1c912a9e38c717f3d1f7) without issue. As I understand it, this only (substantially!) speeds up performance without change in the results. Worth making sure, though.

riddhishb: Yes looking at it. compared speeds, the `schedule = "dynamic"` makes a difference!

arokem: And results are identical?

riddhishb: Yeah identical, the speed improvements are from 4.4 seconds to 3.7 seconds with 4 threads

dipy/denoise/denspeed.pyx
```diff
@@ -157,70 +153,51 @@ cdef double process_block(double[:, :, ::1] arr,
...
 
     cnt = 0
     sumw = 0
     patch_vol_size = (P + P + 1) * (P + P + 1) * (P + P + 1)
-    block_vol_size = (B + B + 1) * (B + B + 1) * (B + B + 1)
 
-    W = <double *> malloc(PS * PS * PS * sizeof(double))
-    cache = <double *> malloc(TS * TS * TS * sizeof(double))
-    sigma_block = <double *> malloc(TS * TS * TS * sizeof(double))
+    W = <double * > malloc(BS * BS * BS * sizeof(double))
```
Garyfallidis: No need for space after *

dipy/denoise/denspeed.pyx
```diff
@@ -307,7 +283,7 @@ cdef cnp.npy_intp copy_block_3d(double * dest,
 
     for i in range(I):
         for j in range(J):
-            memcpy(&dest[i * J * K + j * K], &source[i + min_i, j + min_j, min_k], K * sizeof(double))
+            memcpy(& dest[i * J * K  + j * K], & source[i + min_i, j + min_j, min_k], K * sizeof(double))
```
Garyfallidis: No need for space after &. I know this is a bit puzzling. Because those operators are from C in a way you can consider them as part of the variable name.

Garyfallidis:
Thank you @riddhishb. 

@mdesco this PR restores the older version of nlmeans. To make things clearer @riddhishb will create a new function named ```non_local_means``` that will have use the blockwise implementation which from what we see it is better than the voxelwise implementation. The new function will be added with the adaptive non local means PR. When that happens (which is very soon) a warning will be added to the initial nlmeans to say that this function has now a better alternative (i.e. the blockwise version) so you should start updating your scripts.

Thanks again to @riddhishb for all the excellent work. Keep it up!

Garyfallidis:
@mdesco please check if what is in master works for you now. Thanks again for reporting this quickly.

mdesco:
Things seem back to normal according to my simple tests and scripts. Thank you!

PR #1089, opened 2016-06-25
BF: fix test error on Python 3

matthew-brett:
Numpy does not implement the `__round__` method, causing the following
error when trying to use builtin `round`:

    >>> round(np.array([1]))

For the test error in dipy, see:
https://travis-ci.org/MacPython/dipy-wheels/jobs/140257955#L357

Use numpy's `assert_almost_equal` instead of the version from nose, to
avoid this problem.

### DIFF
--- a/dipy/segment/tests/test_metric.py
+++ b/dipy/segment/tests/test_metric.py
@@ -2,9 +2,9 @@ import numpy as np
 import dipy.segment.metric as dipymetric
 import itertools
 
-from nose.tools import (assert_true, assert_false,
-                        assert_equal, assert_almost_equal)
-from numpy.testing import assert_array_equal, assert_raises, run_module_suite
+from nose.tools import (assert_true, assert_false, assert_equal)
+from numpy.testing import (assert_array_equal, assert_raises, run_module_suite,
+                           assert_almost_equal)
 
 
 def norm(x, ord=None, axis=None):


MarcCote:
This PR seems to only change a few imports in `test_metric.py`. In the current Dipy's master `round` is not used.

arokem:
On a slightly related topic: how hard would it be to set up testing with a mac on Travis? It's fairly easy if you use Anaconda, but I don't know what it would entail here. As it is, errors like this will only get caught when building the wheels, which seems precarious.

matthew-brett:
Notice that the use of `round` is in standard library version of the `assert_almost_equal` function - so the change in the import makes sure we are using the numpy version of that function instead.

Ariel - the test failure is on 32-bit Linux.  It's a little involved to test this, because you need to spin up a 32-bit docker container to run the tests.   It would certainly be a good idea to run the wheel tests fairly often, for example, make sure we have several beta or RC tags before release, and run on those.

MarcCote:
@matthew-brett oh, I see. My bad, I used the wrong import.

arokem:
Oh, gotcha. I must have missed that in the complexity of the setup of that thing.

arokem:
This seems good to me. Unless anyone has any objections, we can move ahead with this in a couple of days.

PR #1087, opened 2016-06-23
Base workflow enhancements + tests

matthieudumont:
Redux version of my previous enhanced_workflows PR. 
Cuts the additions in half but it is still pretty big. The way it is made, this is close to the smallest I can get while keeping something coherent enough to be merged by itself. Let me know what you think of the size. 

Content:
Enhanced workflows:
   -Better command line interface
   -Automatic command line parameters for all workflows like logging params and force overwrite.
   -Easier to create new workflows.
   -Simpler and more consistent ouput management
Combined workflow (pipelines)
    -Instantly exposes sub workflows parameters to command line. 
Tests for all the workflow/combined workflow mecanics, argparser, flow_runner
Examples on how to create a workflow and a combined workflow.

PR #1090, opened 2016-06-27
Documentation for command line interfaces

ghoshbishakh:


doc/tools/docgen_cmd.py
```diff
@@ -0,0 +1,117 @@
...
+import re
+from os.path import join as pjoin
+from os import listdir
+from subprocess import Popen, PIPE, CalledProcessError
+
+# version comparison
+from distutils.version import LooseVersion as V
+
+
+def sh3(cmd):
```
MarcCote: What does `sh3` stands for?

ghoshbishakh: I was looking at how other projects generate documentation. This is taken from a script which I really liked. I did not bother to change the name actually. :yum:
I can change it if you suggest doing so.
https://github.com/scikit-image/scikit-image/blob/master/doc/gh-pages.py

MarcCote: Oh, I see. It's the third version of a function that runs a command in a shell. We can keep that name and add the link you provided in the Notes section of the docstring.

jchoude:
Next time, please add at least a quick description of the goal of the PR. For @Garyfallidis and I, it's easy to guess since it's part of your project, but for other reviewers, it might not be as easy.

doc/Makefile
```diff
@@ -37,6 +37,8 @@ api-clean:
 api:
 	@mkdir -p reference
 	$(PYTHON) tools/build_modref_templates.py dipy reference
+	@mkdir -p reference_cmd
+	$(PYTHON) tools/docgen_cmd.py dipy reference_cmd
```
jchoude: No strong opinion on this, but do others think this might be better moved in a separate section, called something like ```command-line-api```.

doc/tools/docgen_cmd.py
```diff
@@ -0,0 +1,117 @@
+#!/usr/bin/env python
+"""
+Script to generate documentation for command line utilities
+"""
+import sys
```
jchoude: Import ordering.

doc/tools/docgen_cmd.py
```diff
@@ -0,0 +1,117 @@
...
+    out, err = p.communicate()
+    retcode = p.returncode
+    if retcode:
+        raise CalledProcessError(retcode, cmd)
+    else:
+        return out.rstrip(), err.rstrip()
+
+
+def abort(error):
+    print('*WARNING* API documentation not generated: %s' % error)
```
jchoude: Command line API ....

doc/tools/docgen_cmd.py
```diff
@@ -0,0 +1,117 @@
...
+if __name__ == '__main__':
+    # package name: Eg: dipy
+    package = sys.argv[1]
+    # directory in which the generated rst files will be saved
+    outdir = sys.argv[2]
+
+    try:
+        __import__(package)
+    except ImportError as e:
+        abort("Can not import " + package)
```
jchoude: Can not -> Cannot

jchoude:
Also, general question: I'm trying to build the API documentation, to test this PR. However, I work in a virtualenv. When it tries to call the bin/* scripts, there only is the dipy_median_otsu that works. All other 4 scripts have a bad interpreter error.

This happens because of the shebang line, which is 

```
#!python
```
in the failing scripts, while it is
```
#! /usr/bin/env python
```

Do we have a standard for this. StackOverflow seems to prefer the second version, since it honors the environment (and hence, why it works in my venv), but in some cases people advise against it.

Thoughts?

ghoshbishakh:
It is confusing. For me only dipy_median_otsu does not work (while not using virtual environment).

Garyfallidis:
As far as I know the ``#!python`` is the correct shebang line. Is that correct @matthew-brett ?

matthew-brett:
Yes, that's correct.  With the "!python" shebang line, the scripts should install correctly via pip, at least.

ghoshbishakh:
@Garyfallidis @matthew-brett yes using `#!python` , the script becomes executable as a command. But in PR #1087 there is only one workflow (for now) and that uses `#! /usr/bin/env python`
So does that need to be changed?

jchoude:
Also, do we want to be able to call the doc/tools/docgen_cmd.py script directly, or only through the makefile? If we want to be able to call it directly, might be useful to have a help message about the expected arguments, to help users.

ghoshbishakh:
@jchoude It will be called from the makefile. Just like build_examples.py works. But I will give a help text for arguments if any.

MarcCote:
@jchoude: I had a lot of problems when messing with my `PYTHONPATH`. Now, what I do is `pip install -e .` which simply adds a link in the `site-packages/` to your project folder. It works fine within virtualenv too.

doc/tools/docgen_cmd.py
arokem: Please add this link to the docstring. 

And the scikit image license.

doc/tools/docgen_cmd.py
```diff
@@ -0,0 +1,117 @@
...
+        abort("Installed version does not match source version")
+
+    # generate docs
+    bin_folder = pjoin('..', 'bin')
+
+    command_list = []
+    for f in listdir(bin_folder):
+        if f.startswith("dipy_"):
+            try:
+                help_string, err = sh3("%s -h" % (f,))
```
arokem: This is where we need to suppress warnings

ghoshbishakh:
@arokem 
The warning is marked as:
`warnings.simplefilter('always', DeprecationWarning)`
Is it possible to ignore that warning?

https://github.com/nipy/dipy/blob/master/dipy/denoise/nlmeans.py#L8

arokem:
I think so: https://docs.python.org/2/library/warnings.html#
temporarily-suppressing-warnings

It would be ironic to suppress this one, considering we put it there, but
there you go.

ghoshbishakh:
@arokem No what i wanted to say is that should I change the nlmeans file to change the warning filter?
As with `warnings.simplefilter("always")`, I could not find any way to ignore the warning through command line. However after removing the filter line we should be able to ignore it with `python -W ignore filename.py`

arokem:
No - don't change the warning in `nlmeans`! It's there for a reason -- that function is deprecated and we will want to eventually remove it altogether. So users that use this function should be warned that they should start changing their code to use the other implementation of this function.

ghoshbishakh:
So `DeprecationWarning is ignored by default.`

That is why `warnings.simplefilter('always', DeprecationWarning)` is used.

But this makes it difficult to ignore it this way.

With `python -W file.py` I get the output:
```usage: python [option] ... [-c cmd | -m mod | file | -] [arg] ...
Options and arguments (and corresponding environment variables):
-b     : issue warnings about str(bytes_instance), str(bytearray_instance)
         and comparing bytes/bytearray with str. (-bb: issue errors)
-B     : don't write .py[co] files on import; also PYTHONDONTWRITEBYTECODE=x
-c cmd : program passed in as string (terminates option list)
-d     : debug output from parser; also PYTHONDEBUG=x
-E     : ignore PYTHON* environment variables (such as PYTHONPATH)
-h     : print this help message and exit (also --help)
-i     : inspect interactively after running script; forces a prompt even
         if stdin does not appear to be a terminal; also PYTHONINSPECT=x
-I     : isolate Python from the user's environment (implies -E and -s)
-m mod : run library module as a script (terminates option list)
-O     : optimize generated bytecode slightly; also PYTHONOPTIMIZE=x
-OO    : remove doc-strings in addition to the -O optimizations
-q     : don't print version and copyright messages on interactive startup
-s     : don't add user site directory to sys.path; also PYTHONNOUSERSITE
-S     : don't imply 'import site' on initialization
-u     : unbuffered binary stdout and stderr, stdin always buffered;
         also PYTHONUNBUFFERED=x
         see man page for details on internal buffering relating to '-u'
-v     : verbose (trace import statements); also PYTHONVERBOSE=x
         can be supplied multiple times to increase verbosity
-V     : print the Python version number and exit (also --version)
         when given twice, print more information about the build
-W arg : warning control; arg is action:message:category:module:lineno
         also PYTHONWARNINGS=arg
-x     : skip first line of source, allowing use of non-Unix forms of #!cmd
-X opt : set implementation-specific option
file   : program read from script file
-      : program read from stdin (default; interactive mode if a tty)
arg ...: arguments passed to program in sys.argv[1:]

Other environment variables:
PYTHONSTARTUP: file executed on interactive startup (no default)
PYTHONPATH   : ':'-separated list of directories prefixed to the
               default module search path.  The result is sys.path.
PYTHONHOME   : alternate <prefix> directory (or <prefix>:<exec_prefix>).
               The default module search path uses <prefix>/pythonX.X.
PYTHONCASEOK : ignore case in 'import' statements (Windows).
PYTHONIOENCODING: Encoding[:errors] used for stdin/stdout/stderr.
PYTHONFAULTHANDLER: dump the Python traceback on fatal errors.
PYTHONHASHSEED: if this variable is set to 'random', a random value is used
   to seed the hashes of str, bytes and datetime objects.  It can also be
   set to an integer in the range [0,4294967295] to get hash values with a
   predictable seed.
PYTHONMALLOC: set the Python memory allocators and/or install debug hooks
   on Python memory allocators. Use PYTHONMALLOC=debug to install debug
   hooks.
```

So this adds the unnecessary section below.

ghoshbishakh:
Warnings have been removed

doc/tools/docgen_cmd.py
arokem: Please don't forget to add the scikit-image license.

ghoshbishakh: Should I include it in this docstring? @arokem

arokem:
Either in this docstring, or in a top-level docstring for the module.

ghoshbishakh:
Done

Garyfallidis:
Hey @ghoshbishakh this PR needs a rebase to get rid of the failing test.  Then I believe we are good to merge if you agree.

ghoshbishakh:
Yes Ill update it by tonight

PR #1092, opened 2016-06-30
Line-based target()

nilgoyyou:
A line-based version of `tracking/target()` because we will see more and more compressed fibers.

I tried to mimick as much as possible `target()` and its test. The logic being: if `target()` was good enough for dipy, I guess my version will be too :)
The only difference is that I don't raise a `ValueError`. I could do it though, for a loss of speed.

Here are some metrics that I compiled while I was testing. I used a small mask. I tested with non-compressed tck and their compressed version.
```
------------------ Actual results ------------------
      target  target_comp  target_lb  target_lb_comp
50k     137        135         138         138
100k    292        281         299         299
250k    685        674         695         695
500k   1432       1400        1460        1459
1000k  2845       2791        2893        2896
```
We can see a small drop (~1%) from `target` to `target_comp`, which is normal because the standard `target()` function doesn't handle compressed fibers, so it may miss some voxels.

`target_lb` always finds more tract than `target`, which may be suprising at first sight. The problem with `target` is that it misses some voxels where the line passes near the edge of a voxel. `target_lb` catches those voxels.

`target_lb_comp` finds more tract than `target_lb` on large datasets. This also seems surprising, but the compression algorithm may change a little the position of some points, which may make some lines pass by different voxels.
```
----------- SPEED (s) -----------
        target  target_lb  speedup
50k      0.945    0.638      1.5
100k     2.060    1.363      1.5
250k     5.117    3.161      1.6
500k    10.461    6.594      1.6
1000k   20.663   12.722      1.6
```
We can see that `target_lb` is about 1.6x faster. This version is heavily cythonized. This makes the code less readable, but it's faster than `target()`.


jchoude:
Just a quick message: I'll review this, since I know the algorithm (@nilgoyyou started from some of my code), but I'd still like another reviewer to go through it also.

dipy/tracking/vox2track.pyx
```diff
@@ -162,6 +161,172 @@ def streamline_mapping(streamlines, voxel_size=None, affine=None,
...
+                               current_edge)
+
+    # Check last point
+    x = <int>next_pt[0]
+    y = <int>next_pt[1]
+    z = <int>next_pt[2]
+    return mask[x, y, z]
+
+
+@cython.boundscheck(False)
```
jchoude: You added those 2 decorators for this function that you didn't touch. Was it intended? Did you make sure that it didn't change the behavior and that the function code was hnadling those cases itself?

dipy/tracking/vox2track.pyx
```diff
@@ -162,6 +161,172 @@ def streamline_mapping(streamlines, voxel_size=None, affine=None,
...
+     edge[2] = floor(p_z + eps) if d_z >= 0.0 else ceil(p_z - eps)
+
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+@cython.cdivision(True)
+def streamlines_in_mask(list streamlines,
+                        cnp.uint8_t[:,:,:] mask,
+                        lin, offset):
+    """Filters streamlines based on whether or not they pass through an ROI,
```
jchoude: through an ROI -> through a ROI

dipy/tracking/vox2track.pyx
```diff
@@ -162,6 +161,172 @@ def streamline_mapping(streamlines, voxel_size=None, affine=None,
...
+    using a line-based algorithm for compressed tract.
```
jchoude: tract -> streamlines.

dipy/tracking/vox2track.pyx
```diff
@@ -162,6 +161,172 @@ def streamline_mapping(streamlines, voxel_size=None, affine=None,
...
+                                    double eps=1.) nogil:
+     edge[0] = floor(p_x + eps) if d_x >= 0.0 else ceil(p_x - eps)
+     edge[1] = floor(p_y + eps) if d_y >= 0.0 else ceil(p_y - eps)
+     edge[2] = floor(p_z + eps) if d_z >= 0.0 else ceil(p_z - eps)
+
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+@cython.cdivision(True)
+def streamlines_in_mask(list streamlines,
```
jchoude: Would you say this function should never be called directly, going through dipy.tracking.utils.target_line_based instead? Because if you consider it could be called directly, what you expect for the ```lin``` parameter is not in line with what most dipy functions expect (i.e. the basic affine transform)

dipy/tracking/vox2track.pyx
```diff
@@ -162,6 +161,172 @@ def streamline_mapping(streamlines, voxel_size=None, affine=None,
...
+    ----------
+    streamlines : sequence
+        A sequence of streamlines.
+    mask : array-like (uint8)
+        A mask used as a target. Non-zero values are considered to be within
+        the target region.
+    lin : array (3, 3)
+        Transpose of the linear part of the mapping to voxel space, (ie
+        ``inv(affine)[:3, :3].T``)
+    offset : array or scaler
```
jchoude: scaler -> scalar

dipy/tracking/vox2track.pyx
```diff
@@ -162,6 +161,172 @@ def streamline_mapping(streamlines, voxel_size=None, affine=None,
...
+        Offset part of the mapping (ie, ``inv(affine)[:3, 3]``) + ``.5``. The
```
jchoude: Which mapping?

dipy/tracking/vox2track.pyx
```diff
@@ -162,6 +161,172 @@ def streamline_mapping(streamlines, voxel_size=None, affine=None,
...
+        half voxel shift is so that truncating the result of this mapping
+        will give the correct integer voxel coordinate.
+
+    Returns
+    -------
+    in_mask : array of bool
```
jchoude: You should say that you have one bool per streamline, telling if each streamline is going to through the mask. It's not currently clear if it's based on the mask shape or the number of streamlines.

dipy/tracking/vox2track.pyx
```diff
@@ -162,6 +161,172 @@ def streamline_mapping(streamlines, voxel_size=None, affine=None,
...
+        0 if passing in mask, 1 otherwise
```
jchoude: passing in -> passing through

dipy/tracking/vox2track.pyx
```diff
@@ -162,6 +161,172 @@ def streamline_mapping(streamlines, voxel_size=None, affine=None,
...
+
+    cdef int nb_streamlines = len(streamlines)
+    cdef cnp.uint8_t[:] in_mask = np.zeros(nb_streamlines, dtype=np.uint8)
+    cdef int track_idx
+
+    for track_idx in range(nb_streamlines):
+        # Can't call _to_voxel_coordinates because it casts to int
+        voxel_indices = np.dot(streamlines[track_idx], lin) + offset
+
+        if streamline_in_mask(voxel_indices, mask,
```
jchoude: This is really for my personnal knowledge: did you try putting the mapping code (which is in the streamline_in_mask function) directly here with a ```with nogil``` ? If so, was there any performance hits / increases? I don't think there should be anything different, but I sometimes saw weird things with the function calls.

dipy/tracking/vox2track.pyx
```diff
@@ -162,6 +161,172 @@ def streamline_mapping(streamlines, voxel_size=None, affine=None,
...
+                              current_pt, next_pt, direction, current_edge):
+            in_mask[track_idx] = 1
+
+    return np.asarray(in_mask)
+
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+@cython.cdivision(True)
+cdef int streamline_in_mask(cnp.double_t[:,:] streamline,
```
jchoude: I would add the ```_``` prefix to the function name, to show it should be "private", and to make it more easily distinguishable from the calling function (which only has an additional "s" in streamlines).

dipy/tracking/vox2track.pyx
```diff
@@ -162,6 +161,172 @@ def streamline_mapping(streamlines, voxel_size=None, affine=None,
...
+                            cnp.uint8_t[:,:,:] mask,
+                            cnp.double_t[:] current_pt,
+                            cnp.double_t[:] next_pt,
+                            cnp.double_t[:] direction,
+                            cnp.double_t[:] current_edge) nogil:
+    """
+    Check if a single streamline is passing by a mask. This ia a utility
```
jchoude: by -> through

jchoude: This ia a utility -> This is an utility

dipy/tracking/vox2track.pyx
```diff
@@ -162,6 +161,172 @@ def streamline_mapping(streamlines, voxel_size=None, affine=None,
...
+    function of streamlines_in_mask() to be more readable.
```
jchoude: function to make streamlines_in_mask more readable.

dipy/tracking/vox2track.pyx
jchoude: Good job on the "no yellow highlight" in ```cython -a``` for this function :)

jchoude: Also, I think it would be useful to add the reference to the Bresenham algorithm (which this function is an implementation of) here.

The reference is 
```
Bresenham,  Jack Elton. "Algorithm for computer control of a digital plotter", IBM Systems Journal, vol 4, no. 1, 1965.
```
Not sure the format of the ref is the correct one for Dipy.

jchoude: When we discussed live, we talked about an error case when there was only 1 point in the streamline. Is it handled anywhere, or documented that this is not supported?

dipy/tracking/utils.py
```diff
@@ -571,7 +572,7 @@ def target(streamlines, target_mask, affine, include=True):
 
     Raises
     ------
-    IndexError
+    ValueError
```
jchoude: Why would this be a ValueError instead of an IndexError?

dipy/tracking/utils.py
```diff
@@ -595,6 +596,45 @@ def target(streamlines, target_mask, affine, include=True):
             yield sl
 
 
+@_with_initialize
+def target_line_based(streamlines, target_mask, affine=None, include=True):
+    """Filters streamlines based on whether or not they pass through an ROI,
```
jchoude: an ROI -> a ROI

dipy/tracking/utils.py
```diff
@@ -595,6 +596,45 @@ def target(streamlines, target_mask, affine, include=True):
...
+    using a line-based algorithm for compressed tract.
```
jchoude: I know that this is for now the most frequent use case, but it's useful for any tractogram. For example, the step size might be "big" (eg: 0.5mm). In this case, this function would still give more robust results, since we get even the corner of voxels traversed by a streamline, even if no point is inside the voxel.

Will also be useful for tractograms where the step size is varying, eg: global tractography results.

dipy/tracking/utils.py
```diff
@@ -595,6 +596,45 @@ def target(streamlines, target_mask, affine, include=True):
             yield sl
 
 
+@_with_initialize
+def target_line_based(streamlines, target_mask, affine=None, include=True):
```
jchoude: Might be useful in the doc to refer to Bresenham's algorithm and my paper 

```
Houde et al. How to avoid biased streamlines-based metrics for streamlines with variable step sizes, ISMRM 2015.
```

dipy/tracking/utils.py
```diff
@@ -595,6 +596,45 @@ def target(streamlines, target_mask, affine, include=True):
...
+    streamlines : generator
+        A sequence of streamlines that pass through `target_mask`.
+
+    See Also
+    --------
+    density_map
+    """
+    target_mask = np.array(target_mask, dtype=np.uint8, copy=True)
+    lin_T, offset = _mapping_to_voxel(affine, voxel_size=None)
+    yield
```
jchoude: Since you use the ```@_with_initialize``` decorator, it will be initialized up to here. However, on the first real iteration on this generator, the whole filtering code will be executed. I guess that's a matter of philosophy, but wouldn't it make more sense to call ```streamlines_in_mask()``` directly in the initialization, and then yield in constant time for all the iterations?

dipy/tracking/tests/test_utils.py
```diff
@@ -4,16 +4,15 @@
 
 import numpy as np
 import nose
-import nibabel as nib
```
jchoude: Good catch.

dipy/tracking/tests/test_utils.py
```diff
@@ -224,48 +223,62 @@ def test_target():
...
     assert_true(new[0] is streamlines[0])
-    new = list(target(streamlines, mask, affine=affine, include=False))
+    new = list(target_f(streamlines, mask, affine=affine, include=False))
     assert_equal(len(new), 1)
     assert_true(new[0] is streamlines[1])
 
-    # Test that changing mask and affine do not break target
-    include = target(streamlines, mask, affine=affine)
-    exclude = target(streamlines, mask, affine=affine, include=False)
+    # Test that changing mask and affine do not break target_line_based
```
jchoude: In fact, this tests that it doesn,t break either target function, since the current function is called for both target versions.

dipy/tracking/vox2track.pyx
nilgoyyou: It seems those 2 lines were added on master in the meantime so I'll ignore this comment.

dipy/tracking/vox2track.pyx
nilgoyyou: Yes, I tried. It did it this way in my first version. I thought it would be clearer in 2 functions but wondered about the cost of the function calls. I do not remember if there was any significant difference but I'm gonna guess there wasn't or I wouldn't have created another function.

dipy/tracking/utils.py
nilgoyyou: This seems strange because I modified the exception of another function that I haven't modified, but I simply fixed the error. The fonction actually raises an `ValueError`, not a `IndexError`.

dipy/tracking/utils.py
```diff
@@ -595,6 +596,56 @@ def target(streamlines, target_mask, affine, include=True):
...
+    References
+    ----------
+    [Bresenham5] Bresenham, Jack Elton. "Algorithm for computer control of a
+                 digital plotter", IBM Systems Journal, vol 4, no. 1, 1965.
+    [Houde15] Houde et al. How to avoid biased streamlines-based metrics for
+              streamlines with variable step sizes, ISMRM 2015.
+
+    See Also
+    --------
+    density_map
```
Garyfallidis: Hi @nilgoyyou and @jchoude , great job! 
And apologies for delaying to review this PR. Bad timing. 
Can I ask to add the compression function to the see also? You may want to use the full module path when writing in see also to make it easier for the user to know where to find things. For example I would write
dipy.tracking.utils.density_map
dipy.tracking.streamline.compress_streamlines

dipy/tracking/vox2track.pyx
```diff
@@ -162,6 +161,176 @@ def streamline_mapping(streamlines, voxel_size=None, affine=None,
...
 @cython.wraparound(False)
+cdef inline double norm(double x, double y, double z) nogil:
+    cdef double val = sqrt(x*x + y*y + z*z)
+    return val
+
+
+# Changing this to a memview was slower.
+@cython.boundscheck(False)
+@cython.wraparound(False)
+cdef inline void c_get_closest_edge(double p_x, double p_y, double p_z,
```
Garyfallidis: Nice :)

dipy/tracking/vox2track.pyx
Garyfallidis: Sure, no worries this looks good.

dipy/tracking/vox2track.pyx
```diff
@@ -162,6 +161,176 @@ def streamline_mapping(streamlines, voxel_size=None, affine=None,
...
+                             cnp.double_t[:] next_pt,
+                             cnp.double_t[:] direction,
+                             cnp.double_t[:] current_edge) nogil:
+    """
+    Check if a single streamline is passing through a mask. This ia an utility
+    function to make streamlines_in_mask() more readable.
+    """
+    cdef cnp.double_t direction_norm, remaining_distance
+    cdef cnp.double_t length_ratio, half_ratio
+    cdef int point_idx, dim_idx
```
Garyfallidis: Rather than int use ```cnp.npy_intp``` variable that makes sure you will be using the correct integer data types in different platforms. Change everywhere. It's good practice.

dipy/tracking/vox2track.pyx
```diff
@@ -162,6 +161,176 @@ def streamline_mapping(streamlines, voxel_size=None, affine=None,
...
+        if floor(current_edge[0]) != current_edge[0] and \
+           floor(current_edge[1]) != current_edge[1] and \
+           floor(current_edge[2]) != current_edge[2]:
+            # All coordinates are not "integers", and therefore, not on the
+            # edge. Fetch the closest edge.
+            c_get_closest_edge(current_pt[0], current_pt[1], current_pt[2],
+                               direction[0], direction[1], direction[2],
+                               current_edge)
+
+        # TODO Could condition be optimized?
```
Garyfallidis: Yes, I think it can. But you would need to statically define a variable. Have you tried it?
```
cdef cnp.npy_intp a
a = 1
while a==1:
```
Try it it may give you an extra speedup.

dipy/tracking/vox2track.pyx
nilgoyyou: Do you have the same rule for 'double'?

nilgoyyou: In fact, I tested changing all my `cnp.double_t` for `double` and I think it's slower so I'll keep the `cnp.double_t` even when I don't use numpy. Imo, it doesn't make any sense though.

dipy/tracking/vox2track.pyx
nilgoyyou: I tested and, as far as I can tell, it wasn't faster and it forced me to 1) still use a `break` or 2) put everything in an `else`, which I don't want to do because it's already indented enough and we have a 80c maximum.

Garyfallidis:
Great job @nilgoyyou and @jchoude! Thanks for all that speed! And apologies for delaying to review this. Busy times. @jchoude and @nilgoyyou I would suggest to create a tutorial showcasing this new functionality. We have nothing right now (as a tutorial) for compression which is quite a shame. Let's bring it!!! :dancing_women:

PR #1097, opened 2016-07-16
Added to quantize_evecs: multiprocessing and v

xjhc:
Added a multiprocessing option and the ability to choose which eigenvector to use as primary vector for fiber tracking.

### DIFF
--- a/dipy/reconst/dti.py
+++ b/dipy/reconst/dti.py
@@ -1906,21 +1906,21 @@ def _quantize_evecs_parallel(data, odf_vertices, v, nbr_processes):
         
         pool = Pool(nbr_processes)
 
-        pam_res = pool.map(_quantize_evecs_parallel_sub,
+        peaks_res = pool.map(_quantize_evecs_parallel_sub,
                            zip(repeat(data_file_name),
                                indices,
-                               repeat(v),
-                               repeat(odf_vertices)))
+                               repeat(odf_vertices),
+                               repeat(v)))
         pool.close()
         
         
         peak_indices = np.memmap(path.join(tmpdir, 'peak_indices.npy'),
-                                     dtype=pam_res[0].dtype,
+                                     dtype=peaks_res[0].dtype,
                                      mode='w+',
                                      shape=data.shape[0])
         
         for i, (start_pos, end_pos) in enumerate(indices):
-            peak_indices[start_pos: end_pos] = pam_res[i]
+            peak_indices[start_pos: end_pos] = peaks_res[i]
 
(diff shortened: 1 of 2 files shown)


arokem:
Interesting. Do you have any idea whether this actually improves performance in practice?

xjhc:
Paralleling over N cores gives roughly a N-fold speedup for a test dataset of size 1.5 GB. Because the original implementation also reshapes the array (for vectorization), there is very little additional overhead.

This allows processing of 210 GB array of eigenvectors (i.e. 70 GB array of the primary eigenvector) within ~2 hours on Intel Xeon E5-2667 v2 with 20 x SSD on RAID 10.

Also, the ability to choose the primary eigenvector is useful for studies like doi: 10.1016/j.neuroimage.2015.01.061, doi:10.1016/j.cell.2016.05.010 and doi: 10.1117/1.JBO.20.3.036003, where the eigenvector associated with the smallest eigenvalue is actually the primary eigenvector for tracking.

arokem:
OK - sounds like there is a clear use case for this. I do have a couple of comments about the code. The first is that we should add some tests for this functionality. The other is a suggestion: instead of having two different variables for `nbr_processes` and `parallel`, would it make sense to have just one variable `nbr_processes`? Set it to 1 per default, which means no parallelism, -1 means as many processes as possible (the return value of `cpu_count`), etc. Does that make sense, or am I missing something?

arokem:
@xjhc : any more thoughts here? Have you had a chance to write tests for this functionality?

xjhc:
Thanks for the mention - I hadn't checked in on Github lately. 

And thanks for the insightful suggestions. 
- I'll add the tests (in accordance with other tests in this project); I really should've done this before.
- I'll make the update for the nbr_processes.

xjhc:
Just an update - I finished the tasks, but haven't had the opportunity to test it out yet. I'll have access to our workstation in two days, and I hope to get it done then. Thanks for your patience!

arokem:
OK - great. Let us know when you have a chance to try this out, and to push the changes here for review.

xjhc:
Everything's good to go.

arokem:
Sorry for the long gap here. Could you please run a PEP8 checker on the tests? There are a lot of missing white-space in that file. We've generally been following a convention where tests of a module are all in one file, in this case in `test_dti`. Would you mind moving the tests into that file? Thanks again.

arokem:
Pinging @xjhc : Have you had a chance to run a PEP8 checker on this? Thanks!

arokem:
This is superseded by #1164

PR #1098, opened 2016-07-23
Adding custom interactor for vizualisation

MarcCote:
This PR adds a new `vtkInteractorStyle` that allows us to propagate events (mouse, keyboard, etc.) to objects/actors that have been selected.

This is needed for the coming PRs about UI framework that @ranveeraggarwal is working on.

*This PR is still missing some tests.


dipy/viz/window.py
```diff
@@ -486,6 +469,52 @@ def start(self):
         del self.iren
         del self.window
 
+    def record(self, filename="record.log"):
```
Garyfallidis: Let's change the name to record_events (because we are already using this name in another ) and the next one play_recorded_events. Or we could add parameters to start(mode=None or 'record' or 'play')

MarcCote: After thinking about it, I will simply rename the methods `record` to `record_events` and `play` to `play_events`. However, I think that adding a parameter `mode` to the `start` method won't be that useful and it would also require adding a `filename` parameter (which is mandatory for both `record_events` and `play_events`).

MarcCote:
@arokem which Travis build is used for codecov? It seems it doesn't have VTK. See https://codecov.io/gh/nipy/dipy/compare/c09a194e48431325e071bf9d3607805bab75c318...2f297fdca1044548eeae1cce035f4d6248da716a#646970792F76697A2F696E7465726163746F722E7079-8

arokem:
No idea. Looks like @yarikoptic added it (
https://github.com/nipy/dipy/blame/d0bee8c811daf00c5f9c153168ccbc82fa3b5557/.travis.yml#L85),
so he might know.

Garyfallidis:
Hi @MarcCote the coverage jumped from 82 to 88% and then back to 82? We should probably calculate the coverage with VTK enabled. Do you agree?

Garyfallidis:
I get this with VTK 5. 
tests (enh_viz_renderer)]$ python test_fvtk_widgets.py
X Error of failed request:  BadValue (integer parameter out of range for operation)
  Major opcode of failed request:  154 (GLX)
  Minor opcode of failed request:  3 (X_GLXCreateContext)
  Value in failed request:  0x0
  Serial number of failed request:  24
  Current serial number in output stream:  29
Will try also with VTK 6 and see what happens.

Garyfallidis:
This error was in VTK 5.10.1 but when I upgraded to VTK 6.2 the errors disappeared completely.

dipy/viz/interactor.py
```diff
@@ -0,0 +1,269 @@
...
+# Allow import, but disable doctests if we don't have vtk
+vtk, have_vtk, setup_module = optional_package('vtk')
+
+if have_vtk:
+    vtkInteractorStyleUser = vtk.vtkInteractorStyleUser
+    # version = vtk.vtkVersion.GetVTKSourceVersion().split(' ')[-1]
+    # major_version = vtk.vtkVersion.GetVTKMajorVersion()
+else:
+    vtkInteractorStyleUser = object
+
```
Garyfallidis: one more line

dipy/viz/tests/test_interactor.py
```diff
@@ -0,0 +1,105 @@
...
+if use_xvfb == 'skip':
+    skip_it = True
+else:
+    skip_it = False
+
+
+@npt.dec.skipif(not actor.have_vtk or not actor.have_vtk_colors or skip_it)
+@xvfb_it
+def test_custom_interactor_style_events(recording=False):
+    recording_filename = pjoin(DATA_DIR, "test_custom_interactor_style_events.log.gz")
```
Garyfallidis: line too long (pep8)

dipy/viz/tests/test_interactor.py
```diff
@@ -0,0 +1,105 @@
...
+    renderer = window.Renderer()
+
+    # the show manager allows to break the rendering process
+    # in steps so that the widgets can be added properly
+    interactor_style = interactor.CustomInteractorStyle()
+    show_manager = window.ShowManager(renderer, size=(800, 800), interactor_style=interactor_style)
```
Garyfallidis: same here

dipy/viz/tests/test_interactor.py
```diff
@@ -0,0 +1,105 @@
...
+             np.array([[-1, 1, 0.], [1, 1, 0.]])]
+    colors = np.array([[1., 0., 0.], [0.3, 0.7, 0.]])
+    tube1 = actor.streamtube([lines[0]], colors[0])
+    tube2 = actor.streamtube([lines[1]], colors[1])
+    # renderer.add(stream_actor)
+    renderer.add(tube1)
+    renderer.add(tube2)
+
+    # Define some counter callback.
+    states = defaultdict(lambda: 0)
```
Garyfallidis: empty line

dipy/viz/window.py
```diff
@@ -478,14 +464,111 @@ def start(self):
...
+        ----------
+        filename : str
+            Name of the file containing the recorded events (.log|.log.gz).
+        """
+        # Uncompress file if needed.
+        if filename.endswith(".gz"):
+            events = gzip.open(filename, 'r').read()
+        else:
+            events = open(filename).read()
+
```
Garyfallidis: No need to close the file after reading it?

.travis.yml
```diff
@@ -37,6 +37,7 @@ matrix:
     - python: 2.7
       sudo: true   # This is set to true for apt-get
       env:
+        - COVERAGE=1
```
MarcCote: Should I delete the other ones: lines 26 and 30? That way, we will stop receiving multiple coveralls posts.

dipy/viz/window.py
MarcCote: Good point.

dipy/viz/tests/test_interactor.py
MarcCote: Done

.travis.yml
MarcCote: See PR #1132

dipy/viz/window.py
```diff
@@ -417,46 +427,22 @@ def __init__(self, ren, title='DIPY', size=(300, 300),
             self.style = vtk.vtkInteractorStyleImage()
         elif self.interactor_style == 'trackball':
             self.style = vtk.vtkInteractorStyleTrackballCamera()
+        elif self.interactor_style == 'custom':
+            self.style = CustomInteractorStyle()
         else:
             self.style = interactor_style
 
         self.iren = vtk.vtkRenderWindowInteractor()
-        self.iren.SetRenderWindow(self.window)
```
Garyfallidis: Okay this here will have to have a internal dialog. But it should be called from a button not from a keyword. In that way you have more freedom to type and edit labels on the fly.

dipy/viz/window.py
```diff
@@ -478,14 +464,113 @@ def start(self):
...
+
+    def play_events(self, events):
+        """ Plays recorded events of a past interaction.
+
+        The VTK events that happened during the recorded interaction will be
+        played back.
+
+        Parameters
+        ----------
+        events : str
```
Garyfallidis: Is this a string or a list of strings?

dipy/viz/window.py
MarcCote: Agreed

dipy/viz/interactor.py
```diff
@@ -0,0 +1,289 @@
...
+    - Left mouse button: rotates the camera
+    - Right mouse button: dollys the camera
+    - Mouse wheel: dollys the camera
+    - Middle mouse button: pans the camera
+    """
+    def __init__(self):
+        # Default interactor is responsible for moving the camera.
+        self.default_interactor = vtk.vtkInteractorStyleTrackballCamera()
+        # The picker allows us to know which object/actor is under the mouse.
+        self.picker = vtk.vtkPropPicker()
```
MarcCote: The interactor shouldn't be the one that picks individual voxels. The interactor is responsible for picking actor2D/3D. The way I see it, picking voxels should be handled in a callback function associated to the event LeftButtonPress. So, when the user clicks somewhere on a 3D slicer, the interactor sees that as picking the "3D volume" actor then calls the associated callback function in which we have a vtkcellpicker that uses the event position to retrieve the voxel that was picked..

Also, if I remember vtkcellpicker wasn't supported on some computers maybe because of wrong vtk version or not having a dedicated graphic.

Garyfallidis:
Okay this PR looks good on my side. I will wait for a couple of days for others to comment and merge it. In the  meantime @ranveeraggarwal and @MarcCote start breaking up #1111 into smaller PRs. GGT!!! :)

Garyfallidis:
Okay thanks @MarcCote :)

PR #1096, opened 2016-07-11
Robust Brain Extraction

riddhishb:
This PR will mark the start of some work in template based brain extraction as a part google summer of code 2016.
Currently only two changes
1) Added a new function `brain_extraction` in dipy.segment.mask
2) Added a tutorial in doc/examples

All the tests and refinement will be coming along in this month

PR #1106, opened 2016-08-02
Including MNI Template 2009c in Fetcher

riddhishb:
This PR readjusts the MNI template data in fetcher.py and also includes MNI Template T1 data 2009c and it's brain mask in the data along with the 2009a T1 and T2 files. The 2009c template is useful in doing template based brain extraction #1096.

The changes made are the following
* All the flies are stored in figshare
* A version argument = "a" or "c" in read_mni_template is given for choice of data selection.

@arokem @Garyfallidis Do have a look at this


dipy/data/fetcher.py
```diff
@@ -592,16 +595,19 @@ def read_syn_data():
 """
 
 
-def read_mni_template(contrast="T2"):
+def read_mni_template(version, contrast="T1"):
```
RafaelNH: should we add a default?

dipy/data/fetcher.py
```diff
@@ -610,20 +616,42 @@ def read_mni_template(contrast="T2"):
...
-        return nib.load(file_dict[contrast])
+    file_dict_a = {"T1": pjoin(folder, 'mni_icbm152_t1_tal_nlin_asym_09a.nii'),
+                   "T2": pjoin(folder, 'mni_icbm152_t2_tal_nlin_asym_09a.nii')}
+
+    file_dict_c = {
+        "T1": pjoin(
+            folder, 'mni_icbm152_t1_tal_nlin_asym_09c.nii'), "mask": pjoin(
+            folder, 'mni_icbm152_t1_tal_nlin_asym_09c_mask.nii')}
+
+    if contrast == "T2" and version == "c":
```
RafaelNH: You are not covering however the case:
T1_nifti, T2_nifti = read_mni_template("c", contrast = ["T1", "T2"])

dipy/data/fetcher.py
```diff
@@ -610,20 +616,47 @@ def read_mni_template(contrast="T2"):
...
+        raise ValueError("No T2 image for MNI template 2009c")
+
+    if contrast == "mask" and version == "a":
+        raise ValueError("No template mask available for MNI 2009a")
+
+    if not(isinstance(contrast,str)) and version == "c":
```
RafaelNH: PEP8 - missing a space - isinstance(contrast, str)

RafaelNH:
For my side this is ready to be merged! Does anyone else have any comments?

dipy/data/fetcher.py
```diff
@@ -592,16 +595,19 @@ def read_syn_data():
 """
 
 
-def read_mni_template(contrast="T2"):
+def read_mni_template(version="a", contrast="T1"):
```
jchoude: I know this PR is already merged, but isn't it a bad idea to silently change the default value of some parameter like this? Someone calling this from a script without any parameter might suddently encounter bizarre results...

riddhishb: A very valid point, the default value was changed because the version c
does not have a T2 image, so to rectify this we can either rechange the
default to T2 and it will throw a ValueError when version=c.
Or we can also include T2 for version c in the fetcher that way the default
can remain same.

On 10-Aug-2016 7:44 pm, "Jean-Christophe Houde" <notifications@github.com>
wrote:

In dipy/data/fetcher.py
<https://github.com/nipy/dipy/pull/1106#discussion_r74251872>:

> @@ -592,16 +595,19 @@ def read_syn_data():
>  """
>
>
> -def read_mni_template(contrast="
T2"):
> +def read_mni_template(version="a", contrast="T1"):

I know this PR is already merged, but isn't it a bad idea to silently
change the default value of some parameter like this? Someone calling this
from a script without any parameter might suddently encounter bizarre
results...

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/nipy/dipy/pull/1106/files/8499605f20b135f1ae80cf1020bcd75e55eafbd5#r74251872>,
or mute the thread

jchoude: Well, I personally think it's better for backwards compatibility to keep the old value.

Also, which template was fetched before the 2 versions were added? Asking because I just want to know if setting ```version="a"``` will keep the same behavior as before for default cases.

riddhishb: @jchoude It was version "a" which was being fetched when there was only a single version, so that would not cause any issues.

jchoude: Good :+1:

PR #1108, opened 2016-08-09
Local PCA Slow Version

riddhishb:
Only the python implementation of the localPCA algorithm which is without bugs and can be merged. A separate PR will be made very soon to incorporate the cythonized version of this algorithm.

@Garyfallidis Do have a look at this

### DIFF
--- a/dipy/denoise/fast_noise_estimate.pyx
+++ b/dipy/denoise/fast_noise_estimate.pyx
@@ -1,0 +1,134 @@
+"""
+===========================================
+Estimate Noise Levels for LocalPCA Datasets
+===========================================
+
+"""
+
+import numpy as np
+import nibabel as nib
+import scipy as sp
+from scipy import ndimage
+cimport cython
+cimport numpy as cnp
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+@cython.cdivision(True)
+def fast_noise_estimate(data, gtab):
+    """ Noise estimation for local PCA denoising
+
+    Parameters
+    ----------
+    data: 4D array
+        the input dMR data
+
+    gtab: gradient table object
+        gradient information for the data gives us the
+        bvals and bvecs of diffusion data, which is needed
+        here to select between the noise estimation
+        methods
+
+    Returns
+    -------
+    sigma_corr: 3D array
+        The local noise variance estimate
+
+    References
+    ----------
+    [1]: Manjon JV, Coupe P, Concha L, Buades A, Collins DL
+        "Diffusion Weighted Image Denoising Using Overcomplete
+         Local PCA"
+         
+
+    """
+    
+    # first identify the number of the b0 images
+    K = gtab.b0s_mask[gtab.b0s_mask].size
+
+    if(K > 1):
+        # If multiple b0 values then
+        data0 = data[..., gtab.b0s_mask]
+
+    else:
+        # if only one b0 value then
+        # SIBE Noise Estimate
+        data0 = data[..., ~gtab.b0s_mask]
+        # MUBE Noise Estimate
+
+    cdef:
+        cnp.npy_intp n0 = data0.shape[0]
+        cnp.npy_intp n1 = data0.shape[1]
+        cnp.npy_intp n2 = data0.shape[2]
+        cnp.npy_intp n3 = data0.shape[3]
+        cnp.npy_intp nsamples = n0 * n1 * n2
+        cnp.npy_intp i, j, k, i0, j0, k0, l0
+        double sum_reg, temp1
+        double[:, :, :] I = np.zeros((n0, n1, n2))
+        double[:, :, :] count = np.zeros((n0, n1, n2))
+        double[:, :, :] mean = np.zeros((n0, n1, n2))
+        double[:, :, :] sigma = np.zeros((n0, n1, n2))
+        double[:, :, :, :] data0temp = data0
+
+    X = data0.reshape(
+        data0.shape[0] *
+        data0.shape[1] *
+        data0.shape[2],
+        data0.shape[3])
+    # X = NxK
+    # Now to subtract the mean
+    M = np.mean(X, axis=0)
+    X = X - np.array([M, ] * X.shape[0],
+                                      dtype=np.float64)
+    C = np.transpose(X).dot(X)
+    C = C / X.shape[0]
+    # Do PCA and find the lowest principal component
+    [d, W] = np.linalg.eigh(C)
+    d[d < 0] = 0
+    d_new = d[d != 0]
+    # we want eigen vectors of XX^T so we do
+    V = X.dot(W)
+
+    I = V[:, d.shape[0] - d_new.shape[0]].reshape(
+        data0.shape[0],
+        data0.shape[1],
+        data0.shape[2])
+                
+    with nogil:
+        for i in range(1, n0 - 1):
+            for j in range(1, n1 - 1):
+                for k in range(1, n2 - 1):
+                    sum_reg = 0
+                    temp1 = 0
+                    for i0 in range(-1,2):
+                        for j0 in range(-1,2):
+                            for k0 in range(-1,2):
+                                sum_reg += I[i + i0, j + j0, k + k0] / 27.0
+                                for l0 in range(n3):
+                                    temp1 += (data0temp[i + i0, j+ j0, k + k0, l0]) / (27.0 * n3)
+
+                    for i0 in range(-1,2):
+                        for j0 in range(-1,2):
+                            for k0 in range(-1,2):
+                                sigma[i + i0, j +j0, k + k0] += (I[i + i0, j + j0, k + k0] - sum_reg)**2
+                                mean[i + i0, j + j0, k + k0] += temp1
+                                count[i + i0, j +j0, k + k0] += 1             
+
+    sigma = np.divide(sigma, count)
+    
+    # Compute the local mean of for the data
+    mean = np.divide(mean, count)
+    snr = np.zeros_like(I).astype(np.float64)
+    sigma_corr = np.zeros_like(data).astype(np.float64)
+
+    # find the SNR and make the correction
+    # SNR Correction
+
+    snr = np.divide(mean, np.sqrt(sigma))
+    eta = 2 + snr**2 - (np.pi / 8) * np.exp(-0.5 * (snr**2)) * ((2 + snr**2) * sp.special.iv(
+        0, 0.25 * (snr**2)) + (snr**2) * sp.special.iv(1, 0.25 * (snr**2)))**2
+    sigma_corr = sigma / eta
+    # smoothing by lpf
+    sigma_corr[np.isnan(sigma_corr)] = 0
+    sigma_corrr = ndimage.gaussian_filter(sigma_corr, 3)
+    return sigma_corrr
--- a/dipy/denoise/localpca.py
+++ b/dipy/denoise/localpca.py
@@ -1,5 +1,6 @@
 import numpy as np
 import scipy as sp
+from warnings import warn
 
 
 def localpca(arr, sigma, patch_radius=1):
@@ -28,7 +29,9 @@ def localpca(arr, sigma, patch_radius=1):
 
     if arr.ndim == 4:
 
-        tou = 2.3 * 2.3 * sigma
+        patch_size = 2 * patch_radius + 1
+
+        tou = 2.3 * 2.3 * sigma * sigma
 
         if isinstance(sigma, np.ndarray) and sigma.ndim == 3:
 
@@ -44,8 +47,6 @@ def localpca(arr, sigma, patch_radius=1):
         # declare arrays for theta and thetax
         theta = np.zeros(arr.shape, dtype=np.float64)
         thetax = np.zeros(arr.shape, dtype=np.float64)
-
-        patch_size = 2 * patch_radius + 1
 
         for k in range(patch_radius, arr.shape[2] - patch_radius, 1):
             for j in range(patch_radius, arr.shape[1] - patch_radius, 1):
--- a/dipy/denoise/localpca.py
+++ b/dipy/denoise/localpca.py
@@ -2,7 +2,7 @@ import numpy as np
 import scipy as sp
 
 
-def localpca_slow(arr, sigma, patch_radius=1, rician=True):
+def localpca(arr, sigma, patch_radius=1, rician=True):
     r"""Local PCA Based Denoising of Diffusion Datasets
 
     Parameters
--- a/dipy/denoise/localpca.py
+++ b/dipy/denoise/localpca.py
@@ -2,7 +2,7 @@ import numpy as np
 import scipy as sp
 
 
-def localpca(arr, sigma, patch_radius=1, rician=True):
+def localpca(arr, sigma, patch_radius=1):
     r"""Local PCA Based Denoising of Diffusion Datasets
 
     Parameters
@@ -12,9 +12,6 @@ def localpca(arr, sigma, patch_radius=1, rician=True):
         standard deviation of the noise estimated from the data
     patch_radius : The radius of the local patch to
                 be taken around each voxel
-    rician : boolean
-        If True the noise is estimated as Rician, otherwise Gaussian noise
-        is assumed.
 
     Returns
     -------
--- a/dipy/denoise/localpca.py
+++ b/dipy/denoise/localpca.py
@@ -15,7 +15,6 @@ def localpca(arr, sigma, patch_radius=1):
 
     Returns
     -------
-
     denoised_arr : 4D array
         this is the denoised array of the same size as that of
         arr (input data)
@@ -25,7 +24,6 @@ def localpca(arr, sigma, patch_radius=1):
     .. [Manjon13] Manjon JV, Coupe P, Concha L, Buades A, Collins DL
                   Diffusion Weighted Image Denoising Using Overcomplete Local
                   PCA, PLOS 2013
-
     """
 
     if arr.ndim == 4:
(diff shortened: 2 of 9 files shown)


RafaelNH:
Hi @riddhishb! Can you rename the localpca_slow.py file to localpca.py. I know there will be a faster version in a future PR, however calling this slow is misleading because the current version is not that slow right?

dipy/denoise/fast_noise_estimate.pyx
```diff
@@ -0,0 +1,134 @@
...
+    -------
+    sigma_corr: 3D array
+        The local noise variance estimate
+
+    References
+    ----------
+    [1]: Manjon JV, Coupe P, Concha L, Buades A, Collins DL
+        "Diffusion Weighted Image Denoising Using Overcomplete
+         Local PCA"
+
```
RafaelNH: Remove the extra blank space

dipy/denoise/fast_noise_estimate.pyx
```diff
@@ -0,0 +1,134 @@
...
+import nibabel as nib
+import scipy as sp
+from scipy import ndimage
+cimport cython
+cimport numpy as cnp
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+@cython.cdivision(True)
+def fast_noise_estimate(data, gtab):
```
RafaelNH: We have to find a more suggestive name for this!

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,119 @@
...
+        standard deviation of the noise estimated from the data
+    patch_radius : The radius of the local patch to
+                be taken around each voxel
+    rician : boolean
+        If True the noise is estimated as Rician, otherwise Gaussian noise
+        is assumed.
+
+    Returns
+    -------
+
```
RafaelNH: Remove this blank space

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,119 @@
...
+        arr (input data)
+
+    References
+    ----------
+    .. [Manjon13] Manjon JV, Coupe P, Concha L, Buades A, Collins DL
+                  Diffusion Weighted Image Denoising Using Overcomplete Local
+                  PCA, PLOS 2013
+
+    """
+
```
RafaelNH: Remove this Blank space

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,119 @@
...
+    Parameters
+    ----------
+    arr : A 4D array which is to be denoised
+    sigma : float or 3D array
+        standard deviation of the noise estimated from the data
+    patch_radius : The radius of the local patch to
+                be taken around each voxel
+    rician : boolean
+        If True the noise is estimated as Rician, otherwise Gaussian noise
+        is assumed.
```
RafaelNH: @riddhishb - I think the code for this is missing, right?

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,114 @@
...
+                    thetax[i - patch_radius: i + patch_radius + 1,
+                           j - patch_radius: j + patch_radius + 1,
+                           k - patch_radius: k + patch_radius + 1,
+                           :] = thetax[i - patch_radius: i + patch_radius + 1,
+                                       j - patch_radius: j + patch_radius + 1,
+                                       k - patch_radius: k + patch_radius + 1,
+                                       :] + temp / (1 + np.linalg.norm(d,
+                                                                       ord=0))
+
+        denoised_arr = thetax / theta
```
samuelstjean: You seem to be missing the debias step from eq. 5 for rician noise http://journals.plos.org/plosone/article?id=10.1371%2Fjournal.pone.0073021

It would be useful to have it has a keyword like in nlmeans for using it on gaussian noise for example.

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,114 @@
...
+            sigma = np.ones(arr.shape, dtype=np.float64) * sigma
+            tou = np.ones(arr.shape, dtype=np.float64) * tou
+
+        # loop around and find the 3D patch for each direction at each pixel
+
+        # declare arrays for theta and thetax
+        theta = np.zeros(arr.shape, dtype=np.float64)
+        thetax = np.zeros(arr.shape, dtype=np.float64)
+
+        patch_size = 2 * patch_radius + 1
```
samuelstjean: The default implementation is using a patch size of 4 directly [1], so you might want to change the keyword to allow this directly as a patch size rather than patch center. You will probably need to play with border effects to get this though to fit exactly with any array.

http://journals.plos.org/plosone/article?id=10.1371%2Fjournal.pone.0073021 section Local PCA Denoising (LPCA)

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,114 @@
...
+
+    if arr.ndim == 4:
+
+        tou = 2.3 * 2.3 * sigma
+
+        if isinstance(sigma, np.ndarray) and sigma.ndim == 3:
+
+            sigma = (np.ones(arr.shape, dtype=np.float64)
+                     * sigma[..., np.newaxis])
+            tou = (np.ones(arr.shape, dtype=np.float64) * tou[..., np.newaxis])
```
samuelstjean: For the threshold, they use the greek letter tau (unless tou is a valid way of writing it though, never saw it before).

riddhishb: @samuelstjean oh I will change the name

dipy/denoise/localpca.py
riddhishb: What I actually found is that the adaptation step (eq 5 and 6) worsens the results and it doesn't mat at all with the Manjon's MATLAB implementation. We compared our results withe their output with the current algorithm and it is really really close. Plus even though the MATLAB implementation has an option to choose between the Gaussian and Rician Noise, the error map between the two output is exactly zero, which makes us believe that they might have used is minus the debias step.

samuelstjean: That is, hmm, unexpected, I had a look at the code and they do 

lpca denoising

if rician

samuelstjean: Ah yes, also this correction is most noticeable at low SNR, as with a high enough SNR a gaussian approximation is enough, so it could be why you don't see too much difference if you have a high SNR. Here is a graph form the original paper that proposed this correction
capture d ecran de 2016-08-16 19-55-14

It is equation 4 from the lpca paper, and closely related to eq. 5. As you can see, above SNR 4 the correction factor is almost 1, and thus does not produce that much of a difference in the Rician noise case. It is more pronounced for other cases, but at low SNR it should give you different results though.

Figure 1 (or equation 8 in this ref) is from Koay, C. G., & Basser, P. J. (2006). Analytically exact correction scheme for signal extraction from noisy magnitude MR signals. Journal of Magnetic Resonance, 179(2), 317–322. http://doi.org/10.1016/j.jmr.2006.01.016

riddhishb: @samuelstjean This makes sense!, I had the rician correction implemented exactly how you pointed out, but I dont know if it was wrong or something else, it just didnt work for me. I got very decent result compared to the MATLAB implementation. However If you can look at how eq 5 / 6 are implemented in matlab and give some pointers as to how it is used then it would be great!

RafaelNH: Hi guys! This rician bias correction is indeed important! However, my suggestion is to make the implementation of this in a separate PR and after the GSoC submission. The rician bias correction is a related but, since its is applied after the denoising procedure, we can deal with it separately! We can even make it a general function compatible with other denoising techniques.

samuelstjean: So, while my tracking computes, it seems they recompute the final noise variance on the residuals maps + correction factor, similar to the way it is computed on the smallest pca components at first. Since you now have a 4D local map of noise variance, they take the median at the end to get a robust 3D estimation.

For the debias step itself, it gets a tad more complicated as you need to compute a numerical inverse for the mapping function. So, eq. 5 works with the local SNR once again, this time using the computed denoised value and the new value from the residual noise map at each voxel.

This value of E(x)/sigma is then used in eq. 6 to build a table of values (or it can just be applied to each voxel, they do that to save some computation time here it seems).
------
And a new comment appeared between me starting and finishing this message, so here goes : 

It would be easier to get it included now as it is part of the algorithm. Since apparently when looking a difference of FA you see stuff along 1-2% difference, the value you can get out with and without bias correction is in the same order/higher if you move to high b-value/finer voxels (and thus less signal). 

Well anyway, let's see how it goes but all the parts seem to be in place already except for the one liner of eq 6. While it is true that you could use it for other functions, nlmeans work in the square domain and already include such a correction, so I don't know how applicable to other stuff it would be.

dipy/denoise/fast_noise_estimate.pyx
Garyfallidis: Yeah, @riddhishb you need a better name. Is this noise estimation useful only for LPCA?

riddhishb: @Garyfallidis Yes this is only useful for the localPCA, what do you suggest should a good name be ?
Also I noticed there are some build errors and merge conflicts let me correct them

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,114 @@
...
+
+                    temp = arr[i - patch_radius: i + patch_radius + 1,
+                               j - patch_radius: j + patch_radius + 1,
+                               k - patch_radius: k + patch_radius + 1, :]
+                    X = temp.reshape(
+                        patch_size *
+                        patch_size *
+                        patch_size,
+                        arr.shape[3])
+                    # compute the mean and normalize
```
RafaelNH: Hi @riddhishb! I was reviewing the math begin this and when patch_size ** 3 < arr.shape[3] you will have zero eigenvalues! I will suggest adding at the beginning of this function an error message if patch_size ** 3 is smaller than the number of gradient directions informing user to give a larger value of patch_size.

riddhishb:
@RafaelNH I think you are right, I will get some time in the weekend, I think this PR has taken long enough and I will see to all the changes are done. Fingers crossed :)

arokem:
Don't forget to rebase on master!

On Thu, Sep 29, 2016 at 8:21 AM, Riddhish Bhalodia <notifications@github.com

RafaelNH:
Cool - During the summer I was also able to implement the marchenko pastur based pca! After having your PR merged I am planing to extent your work with the selection of signal component based on the Marchenko-Pastur distribution! So I am looking so forward for your revision @riddhishb!

riddhishb:
@RafaelNH Thats great!! Lets all work, may be have a chat on hangouts as well.

RafaelNH:
@riddhishb! Sounds good! But let's do this step by step. First milestone is to have this PR merged!

dipy/denoise/fast_noise_estimate.pyx
samuelstjean: So before this gets close to merging, this function cna be used of anything else needing noise estimation. If not, it would be faster to embed it in the lpca code directly as you end up doing two pcas instead, which is probably the most expensive operation here.

dipy/denoise/fast_noise_estimate.pyx
```diff
@@ -0,0 +1,134 @@
...
+    with nogil:
+        for i in range(1, n0 - 1):
+            for j in range(1, n1 - 1):
+                for k in range(1, n2 - 1):
+                    sum_reg = 0
+                    temp1 = 0
+                    for i0 in range(-1,2):
+                        for j0 in range(-1,2):
+                            for k0 in range(-1,2):
+                                sum_reg += I[i + i0, j + j0, k + k0] / 27.0
```
samuelstjean: Might want to use the shape of the 3D block instead of hard coding 3x3x3 = 27 in case someone changes it later.

dipy/denoise/fast_noise_estimate.pyx
```diff
@@ -0,0 +1,134 @@
...
+                                    temp1 += (data0temp[i + i0, j+ j0, k + k0, l0]) / (27.0 * n3)
+
+                    for i0 in range(-1,2):
+                        for j0 in range(-1,2):
+                            for k0 in range(-1,2):
+                                sigma[i + i0, j +j0, k + k0] += (I[i + i0, j + j0, k + k0] - sum_reg)**2
+                                mean[i + i0, j + j0, k + k0] += temp1
+                                count[i + i0, j +j0, k + k0] += 1             
+
+    sigma = np.divide(sigma, count)
```
samuelstjean: The (unbiased for samples) standard deviation needs a N-1 denominator, you seem to use sum(X - mean)**2 / N instead.

dipy/denoise/localpca.py
samuelstjean: This also means you don't need padding nor playing with negatives indices, you can juts feed the blocks directly to the pca.

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,114 @@
...
+                    M = np.mean(X, axis=0)
+                    X = X - np.array([M, ] * X.shape[0],
+                                     dtype=np.float64)
+
+                    # Compute the covariance matrix C = X_transpose X
+                    C = np.transpose(X).dot(X)
```
samuelstjean: This is more of a performance and stability issue, but you can also use the SVD for computing PCA which is more stable https://en.wikipedia.org/wiki/Principal_component_analysis#Singular_value_decomposition. It kind of falls in the same category as don't invert matrices with the Gauss-Jordan method, don't compute likelihood directly but use the log likelihood, etc.

Unfortunately, back in the days, the SVD of scipy was terribly slow, so I gave up on it. Nowadays, you can hook up directly into the lapack version instead, so it should be pretty fast https://docs.scipy.org/doc/scipy/reference/linalg.lapack.html That's also what the original is using, in a multithreaded loop for efficiency.

dipy/denoise/fast_noise_estimate.pyx
RafaelNH: Well in theory this function can be used for other techniques needing noise estimation. Right @riddhishb ?
Also, even if we embed this code in the lpca code directly, we will always need to do two PCAs - the first pca is done only at part of the data.

samuelstjean: Well, you can do PCA, take smallest eigen value, compute local std noise field, then hard threshold and rebuild in one step. The second part for the noise bias (which is not included here) is separate and is using the difference image as the new noise field for bias correction.

samuelstjean: Oh wait, that is the version where the noise field is estimated from multiple b0s or from the lowest component of the data? In the first case you would need two pca yes (and thus the function could be accessible from outside, invalidating the previous remark.)

samuelstjean:
I gave it a run on some real data, and the noise estimation throws some warning
~~~python
In [22]: sigma = np.array(fast_noise_estimate(data.astype(np.float64), gtab))
/home/samuel/anaconda3/bin/ipython:1: RuntimeWarning: overflow encountered in multiply
  #!/home/samuel/anaconda3/bin/python

/home/samuel/anaconda3/bin/ipython:1: RuntimeWarning: overflow encountered in square
  #!/home/samuel/anaconda3/bin/python
/home/samuel/anaconda3/bin/ipython:1: RuntimeWarning: invalid value encountered in multiply
  #!/home/samuel/anaconda3/bin/python
~~~

I checked the local standard deviation, seemed a bit high (this is real data, so who knows the right value, might want to test a synthetic dataset at a lower SNR).

capture d ecran de 2016-10-01 14-51-26

It also seems to overblur compared to the original lpca, could be the default settings from the example. 

lpca

lpca_out

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,114 @@
...
+    References
+    ----------
+    .. [Manjon13] Manjon JV, Coupe P, Concha L, Buades A, Collins DL
+                  Diffusion Weighted Image Denoising Using Overcomplete Local
+                  PCA, PLOS 2013
+    """
+
+    if arr.ndim == 4:
+
+        tou = 2.3 * 2.3 * sigma
```
riddhishb: The sigma from the function is actually sigma squared but yeah I will change the naming so that so that there is no confusion.

dipy/denoise/localpca.py
riddhishb: I do not get this. Why is this true, Every time I have experimented the patch_size**3 has been smaller than the gradient directions.

doc/examples/denoise_localpca.py
```diff
@@ -0,0 +1,116 @@
...
+
+* First we estimate the local noise variance at each voxel
+
+* Then we apply PCA at local patches around each voxels over the gradient
+  directions
+
+* Threshold the eigenvalues based on the local estimate of sigma and then do
+  the PCA reconstruction
+
+Let's load the necessary modules
```
riddhishb: All changes to the headi incorporated

dipy/denoise/localpca.py
RafaelNH: What patch_size do you normally use?

When patch_size ** 3 < arr.shape[3] the covariance matrix is not full rank. In this cases you will have a number of zero eigenvalues: 
N = arr.shape[3]-patch_size ** 3.

RafaelNH: Perhaps instead of adding the error message, you could just remove the N lower eigenvalues when patch_size ** 3 is smaller than arr.shape. In tests we have to make sure that we cover the two cases: patch_size ** 3 < arr.shape and  patch_size ** 3 > arr.shape.

dipy/denoise/fast_noise_estimate.pyx
riddhishb: What can be a better name ? PCA noise estimate ?

dipy/denoise/localpca.py
riddhishb: I think I use patch size 3 and 5 for testing in normal cases. Oh I think thats right, removing N lower eigenvalues makes sense

Garyfallidis:
@riddhishb can you give us an update with this PR?

Garyfallidis:
Ping @riddhishb :)

arokem:
I believe this was superseded by #1223. Closing.

PR #1109, opened 2016-08-11
Changed default value of mni template

riddhishb:
In reference to @jchoude comment on #1106, I have changed the default value back to T2


RafaelNH:
If no one opposes, I will merge this at the end of this day!

jchoude:
Seems good to me.

PR #1110, opened 2016-08-15
WIP : Ivim linear fitting

sahmed95:
This is a continuation of https://github.com/nipy/dipy/pull/1058. Simple least squares fitting fails for some regions with the fit returning the 'f' values quite high (> 0.3) and converging to D/D_star = 0. Setting bounds for 'f' strangely sets it to the higher bound and again gives an erroneous result. 
A linear fitting strategy is used here which initially fits for parameters D and S0, then finds 'f' and finally uses least squares fitting for D_star as follows : 

    i. Linear fitting for D (bvals > split_b) and store S0_prime
    ii. Another linear fit for S0 (bvals < split_b).
    iii. Estimate f using 1 - S0_prime/S0.
    iv. Use least squares only to fit for D_star. 

This gives a sufficiently good prediction for S0 and D but not so good for 'f' and 'D_star'. The `two_stage` method can be used for a more accurate fit but that fails for noisy data or when 'f' values are very low. Two proposed solutions for that are : 

    Fit D and S0 linearly and fit f and D_star using least_squares
    Write a customized gradient descent for bi exponenetial fits for this particular case.

Regularization, increasing the error tol, changing the loss function for least_squares (SCIPY 0.17) or the method (trf, dogleg, LM) has been tried but eveytime it seems that the solution with unfeasible 'f' (higher value) gets selected.

@arokem @etpeterson can perhaps shed more light on regularization and why it might or might not work.


sahmed95:
By using the one stage fitting, this is the fitting failure. This is better than the last commit as the value of f is obtained by least squares and not 1 - S0_prime/S0 as before. D and S0 are guessed from linear fits and D_star and f are obtained by least_squares. 
```
FAIL: Test if the fit returns the correct S0 value using 'one_stage' fitting.
(mismatch 100.0%)
 x: array(1000.0)
 y: array(990.4734414875644)

FAIL: Test if the fit returns the correct f value using 'one_stage' fitting.
(mismatch 100.0%)
 x: array(0.132)
 y: array(0.10881928872785954)

FAIL: Test if the fit returns the correct D_star value using 'one_stage' fitting.
(mismatch 100.0%)
 x: array(0.00885)
 y: array(0.009072104361755918)

FAIL: Test if the fit returns the correct D value using 'one_stage' fitting.
(mismatch 100.0%)
 x: array(0.000921)
 y: array(0.0009546938098179949)
```

sahmed95:
@etpeterson Is it okay to use two different split_b values ? Here I am using bval >= 500 to get D and bval <= 200 to get the S0 and a D_prime_star. Then, using ```f_guess = 1 - S0_prime / S0``` and finally fitting both D_star and f using leastsq is giving me a better result for the linear fitting. 
```
FAIL: Test if the fit returns the correct S0 value using 'one_stage' fitting.

(mismatch 100.0%)
 x: array(1000.0)
 y: array(990.4734414875644)

FAIL: Test if the fit returns the correct f value using 'one_stage' fitting.
(mismatch 100.0%)
 x: array(0.132)
 y: array(0.12381789392632028)

FAIL: Test if the fit returns the correct D_star value using 'one_stage' fitting.
(mismatch 100.0%)
 x: array(0.00885)
 y: array(0.007959302301038805)

FAIL: Test if the fit returns the correct D value using 'one_stage' fitting.

(mismatch 100.0%)
 x: array(0.000921)
 y: array(0.0009261460229925849)

```

sahmed95:
I am starting to think that a direct leastsq will not work for this model whatever the initial conditions are in case of even slight noise in the data. Can we think of some kind of an SNR test to determine whether least squares should be used or simple linear fitting. I tried a lot of different things today like setting a scale or changing bounds etc. Nothing is working for leastsq once a slight nose kicks in... :-(

etpeterson:
I think using a different split is OK. Especially on the S0 and D*' (which you can probably just call D* because that's how you're calling the other estimates) because there you're fitting a curve with a line. The tradeoff there is that it's probably potentially more accurate but less robust with a lower threshold.

A question for you: how much is the `self.estimate_f_D_star(data, x0_guess)` improving things compared to the linear fit in `self.estimate_S0_D_star_prime(data)`?

And I know I've said it before but it seems like a good place to say it again, if the estimates are pretty good could a regularization be in order? If only on D and S0 which you know quite accurately. That final fit could be turned on and off with your two_stage flag easily as well as modifying the regularization parameters.

sahmed95:
```
actual_params - [1.00e+02, 1.32e-01, 8.85e-03, 9.21e-04]

x0_guess (estimate_S0_D_star_prime) -  [  9.90473441e+02   1.19496745e-01   1.50878636e-03   9.26146023e-04]

x0 (estimate_f_D_star) - [  9.90473441e+02   1.23817894e-01   7.95930231e-03   9.26146023e-04]
```

sahmed95:
>if the estimates are pretty good could a regularization be in order?

I will try regularization and check.

sahmed95:
@etpeterson I tried some sort of Regularization after normalizing the data. Doesn't give any significant advantage.

etpeterson:
Try `+ np.sum(((params - reg)/reg ** 2)*scaling)` so dividing by reg scales the parameters similarly and scaling is something like [1,1,1,1] for starters. Then if you increase scaling you can control the regularization of each parameter individually.

sahmed95:
I have tried scaling with minimize which didn't give anything better than the linear fitting. Surprisingly having two different splits for bvalues improves the results for noisy data. The other surprising thing is scaling before applying least squares and not scaling also leads to a difference in the results. Scaling keeps the f values low and hence acceptable.

I will run the current code on the slice of a full brain and plot the results soon. For now this probably can be reviewd as I have to complete my submission for GSoC. @arokem @etpeterson 

After that I will try some new things to take care of noisy data (writing gradient descent, preprocessing or using [basinhopping](http://docs.scipy.org/doc/scipy-0.16.0/reference/generated/scipy.optimize.basinhopping.html) )

sahmed95:
Some plots from the notebook : 
https://github.com/sahmed95/notebooks/blob/ivim/noisy_fit.ipynb

p1
p2
plot
plot2

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,494 @@
...
+from dipy.core.gradients import gradient_table
+from dipy.reconst.base import ReconstModel
+from dipy.reconst.multi_voxel import multi_voxel_fit
+
+SCIPY_LESS_0_17 = LooseVersion(scipy.version.short_version) < '0.17'
+
+if SCIPY_LESS_0_17:
+    leastsq = scipy.optimize.leastsq
+else:
+    least_squares = scipy.optimize.least_squares
```
arokem: I would change these lines to: 

    if SCIPY_LESS_0_17:
        from scipy.optimize import leastsq
    else:
        from scipy.optimize import least_squares

arokem:
I had several formatting suggestions, so I made a PR: 

https://github.com/sahmed95/dipy/pull/2

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,478 @@
...
+            from None. default : ([0., 0., 0., 0.], [np.inf, 1., 1., 1.])
+
+        tol : float, optional
+            Tolerance for convergence of minimization.
+            default : 1e-7
+
+        scaling : array, optional
+            Scaling for the parameters. This is passed to `_ivim_error` to
+            normalize the signal and scale the parameters appropriately.
+            default: [1e-4, 100., 1000., 10000.]
```
arokem: A bit more guidance would be good. For example, something like: "these scaling factors should bring the different parameters more or less to the same order of magnitude, so should be inversely proportional to the scale of S0, f, D D_star and D, respectively."

sahmed95:
@arokem Thank you. I will go through the code again and refactor. The current code imposes the condition that if 'f' values after fitting become more than 50% then the parameters obtained after scaling are selected, otherwise the parameters from a normal fit is returned. There is no one-stage or two-stage fit now. The issue is that I have to do two fits to find the f value and then reject one set of parameters. Is there a way to guess from just the voxel data whether the signal is a good bi exponenetial or not. That way I can do the final fitting just once and save time. I tried finding the SNR to decide but the difference in SNR is just around 1.2 for a good set of data against noisy data.

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,478 @@
...
+
+        The fitting takes place in the following steps: Linear fitting for D
+        (bvals > 200) and store S0_prime. Another linear fit for S0 (bvals <
+        200).Estimate f using 1 - S0_prime/S0. Use least squares to fit D_star
+        and f.
+
+        We do a final fitting of all four parameters after scaling the
+        parameters and select the set of parameters which make sense physically.
+        The criteria for selecting a particular set of parameters is checking
+        the perfusion fraction. If the fraction is more than 30%, we will reject
```
arokem: This says 30%, but the code below (L259) seems to have 0.5. Is that consistent?

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,478 @@
...
+                                           self.split_b_D]), 1)
+        S0_prime = np.exp(-neg_log_S0)
+        return S0_prime, D
+
+    def estimate_S0_D_star_prime(self, data):
+        """Estimate S0 and D_star_prime for bvals < split_b_S0
+        """
+        bvals_le_split = self.gtab.bvals[self.gtab.bvals < self.split_b_S0]
+        bvecs_le_split = self.gtab.bvecs[self.gtab.bvals < self.split_b_S0]
+        gtab_le_split = gradient_table(bvals_le_split, bvecs_le_split.T)
```
arokem: I don't think you need the gtab here. These three lines could be changed to:

    bvals_le_split = self.gtab.bvals[self.gtab.bvals < self.split_b_S0]

And then line 283 would change to: 

    D_star_prime, neg_log_S0 = np.polyfit(bvals_le_split,

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,478 @@
...
+        params_in_mask = np.where(params_no_scaling[1] <= .5,
+                                  params_no_scaling, params_with_scaling)
+        return IvimFit(self, params_in_mask)
+
+    def estimate_S0_prime_D(self, data):
+        """Estimate S0_prime and D for bvals > split_b_D
+        """
+        bvals_ge_split = self.gtab.bvals[self.gtab.bvals >= self.split_b_D]
+        bvecs_ge_split = self.gtab.bvecs[self.gtab.bvals >= self.split_b_D]
+        gtab_ge_split = gradient_table(bvals_ge_split, bvecs_ge_split.T)
```
arokem: Same here, you never use the gtab.

sahmed95:
:-( The map for f : 
index

The only thing that can be done not is perhaps smoothening the data by taking average or returning just the results from a linear fit without doing another least_squares. I am stuck !

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,478 @@
...
+                warningMsg = "x0 obtained from linear fitting is unfeasibile."
+                warningMsg += "Using parameters from the linear fitting"
+                warnings.warn(warningMsg, UserWarning)
+                f, D_star = x0[1], x0[2]
+                return f, D_star
+        else:
+            try:
+                res = least_squares(f_D_star_error,
+                                    [x0[2], 10 * x0[3]],
+                                    bounds=((0., 0.), (.5, 0.01)),
```
arokem: Should these bounds really be hard-coded? Also, should this be .5 or .3?

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,478 @@
...
+
+    return signal - f_D_star_prediction([f, D_star], gtab, x0)
+
+
+class IvimModel(ReconstModel):
+    """Ivim model
+    """
+
+    def __init__(self, gtab, split_b_D=400.0, split_b_S0=200., bounds=None,
+                 tol=1e-15, scaling=np.array([10e-4, 10e2, 1000., 10000.]),
```
etpeterson: `scaling=np.array([1e-3, 1e3, 1e3, 1e4]),`

Also, that seems like a very large scaling for f, is that intentional?

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,478 @@
...
+            warnings.warn(warningMsg, UserWarning)
+            f, D_star = f_guess, D_star_prime
+
+        x0 = np.array([S0, f, D_star, D])
+        # Fit parameters again with scaling
+        params_with_scaling = self._leastsq(data, x0, scaling=self.scaling)
+        # Fit parameters without scaling
+        params_no_scaling = self._leastsq(data, x0)
+
+        params_in_mask = np.where(params_no_scaling[1] <= .5,
```
etpeterson: Is this double fitting useful? Are there times when no scaling works better than scaling? It just seems like fitting twice would be good to avoid if possible.

etpeterson:
@sahmed95 In response to your question "Is there a way to guess from just the voxel data whether the signal is a good bi exponenetial or not." The easiest way that comes to mind now that I think about it a little is compare your D and D* estimates from the linear (single exponential) fitting. If the slopes (and maybe intercepts too?) are close then it's a single exponential, otherwise it isn't. That may be kind of noisy though so perhaps something like the sum of the residual, but the value of that doesn't have such a clear cutoff.

dipy/reconst/ivim.py
sahmed95: I think f should not be scaled at all and the factor should be 1. But if I do that I get test failures for all parameters. I think its better to handle scaling with least_squares built in method instead of this way.

dipy/reconst/ivim.py
sahmed95: Not scaling and scaling the parameters do not make any difference though as values are not significantly different from the linear fit. I am hence removing the manual scaling introduced and using `x_scale` with [least_square's](http://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.least_squares.html)

sahmed95:
I spent the day trying various combinations of bi-exponential curves with and without noise and fitting mono exponential and bi-exponential models. I think writing a custom gradient descent will not be of any use as imo this is not an issue of settling into one local minima. Higher values of f DO fit the data well and it just means that the signal can be better represented as a single exponential.

The current code is good enough for fits as it checks if the `f` value crosses the threshold(20%) and returns the results from the linear fit. 

Another approach which allows the user to select which fit he/she wants : 

- Use `two_stage=False` keyword to specify that only linear fitting should take place.
    This will give a quick first estimate.
- If `two_stage = True` store the results of linear fit as well as two_stage fit in the `fit`object.
- Based on the keyword argument `fit.model_params(two_stage=True)` or `fit.model_params(two_stage=False)` return the respective parameters.

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,487 @@
...
+    f, D_star = params
+    return signal - f_D_star_prediction([f, D_star], gtab, x0)
+
+
+class IvimModel(ReconstModel):
+    """Ivim model
+    """
+
+    def __init__(self, gtab, split_b_D=400.0, split_b_S0=200., bounds=None,
+                 tol=1e-15, f_threshold=0.25, x_scale=np.array([1e03, 1e-01, 1e-03, 1e-4]),
```
arokem: Would you mind writing the x_scale as : `[1000, 0.1, 0.001, 0.0001]` instead? It's easier to read, I think.

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,487 @@
...
+        f_threshold : float, optional
+            Threshold value to consider for f being erroneous after leastsq
+            fitting. If the value of f obtained crosses this threshold the
+            parameters will be taken from the linear fits.
+            default : .25
+
+        x_scale : array, optional
+            Scaling for the parameters. This is passed to `least_squares` which is
+            only available for Scipy version > 0.17.
+            default: [1e03, 1e-01, 1e-03, 1e-4]
```
arokem: Change here as well.

dipy/reconst/ivim.py
arokem: This is now 0.25? The docs are still wrong.

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,487 @@
...
+        x0 = np.array([S0, f, D_star, D])
+        # Fit parameters again with scaling
+        params = self._leastsq(data, x0)
+
+        if params[1] > self.f_threshold:
+            params = x0
+
+        return IvimFit(self, params)
+
+    def estimate_linear_fit(self, data, split_b, lesser=True):
```
arokem: I prefer `less_than` instead of `lesser`.

arokem:
The threshold is 25%, right? Make sure that is consistent with the documentation. I had just a couple of small comments, but assuming those are fixed and this passes the tests, I am +1 for the merge. @etpeterson : what are your thoughts?

sahmed95:
@arokem Yes, I have taken the threshold as 25% and updated the docs accordingly. This is based on the Federau paper which rejects f values greater than 20%. @etpeterson can perhaps suggest a better value.

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,540 @@
...
+
+        The fitting takes place in the following steps: Linear fitting for D
+        (bvals > 200) and store S0_prime. Another linear fit for S0 (bvals <
+        200).Estimate f using 1 - S0_prime/S0. Use least squares to fit D_star
+        and f.
+
+        We do a final fitting of all four parameters and select the set of
+        parameters which make sense physically. The criteria for selecting a
+        particular set of parameters is checking the perfusion fraction.
+        If the fraction is more than 25%, we will reject the solution obtained
```
arokem: Sorry for being a pain in the neck about this. Given that it is an argument it will be 25% *per default*, but will change based on user input. Can we write: 
   
    ... is more than `f_threshold` (default: 25%), we will reject...

sahmed95: Not a problem at all. :smile: And thank you for being so thorough with the reviews ! Also, I believe I should write more tests. 

> "Untested code is broken code"

:p

arokem: Yes. I just ran the tests with coverage. The following lines of code are not currently being tested: 

15, 199, 312-320, 383, 392-399, 413-418, 450, 454, 458, 462, 14->15, 382->383, 443->445

Some of these can't be tested (e.g., selection among different versions of Scipy), but we should definitely write tests to cover as much of this as possible.

etpeterson:
I like the `two_stage=False` thoughts. I think some way of being able to return the quick estimate is good, especially if it may be more robust in some cases. I'd go for that as long as it won't be too esoteric or confusing.

I think 25% is a legitimate value. It will probably have issues with CSF but the user can change it so that's fine for a default.

I'm going to read through the code a bit and probably make some suggestions but I'm also +1 on merging. Just give me some time to look it over and also I'm assuming the tests are passing.

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,541 @@
...
+
+def _ivim_error(params, gtab, signal):
+    """Error function to be used in fitting the IVIM model.
+
+    Parameters
+    ----------
+    params : array
+        An array of IVIM parameters - [S0, f, D_star, D]
+
+    gtab : GradientTable class instance
```
etpeterson: Minor thought for @arokem and @sahmed95, the other things in gtab aren't being used, only the bvals. Maybe the input should just be the bvals?

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,541 @@
...
+    Returns
+    -------
+    residual : array
+        An array containing the difference between actual and estimated signal.
+    """
+    residual = signal - ivim_prediction(params, gtab)
+    return residual
+
+
+def f_D_star_prediction(params, gtab, S0, D):
```
etpeterson: I guess the gtab thought applies here too.

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,541 @@
...
+    S : array
+        An array containing the IVIM signal estimated using given parameters.
+    """
+    f, D_star = params
+    b = gtab.bvals
+    S = S0 * (f * np.exp(-b * D_star) + (1 - f) * np.exp(-b * D))
+    return S
+
+
+def f_D_star_error(params, gtab, signal, S0, D):
```
etpeterson: I guess the gtab pops up a few times

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,541 @@
...
+    def __init__(self, gtab, split_b_D=400.0, split_b_S0=200., bounds=None,
+                 tol=1e-15, f_threshold=0.25, x_scale=[1000., 0.1, 0.001, 0.0001],
+                 options={'gtol': 1e-15, 'ftol': 1e-15,
+                          'eps': 1e-15, 'maxiter': 1000}):
+        """
+        Initialize an IVIM model.
+
+        The IVIM model assumes that biological tissue includes a volume
+        fraction 'f' of water flowing in perfused capillaries, with a
+        perfusion coefficient D* and a fraction (1-f) of static (diffusion
```
etpeterson: I'd change this to "... of water flowing, with a pseudo-perfusion coefficient D*..."

It's not truly the perfusion.

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,541 @@
...
+            supported for Scipy version > 0.17. When using a older Scipy
+            version, this function will raise an error if bounds are different
+            from None. default : ([0., 0., 0., 0.], [np.inf, .3, 1., 1.])
+
+        tol : float, optional
+            Tolerance for convergence of minimization.
+            default : 1e-15
+
+        f_threshold : float, optional
+            Threshold value to consider for f being erroneous after leastsq
```
etpeterson: change leastsq to non-linear

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,541 @@
...
+        else:
+            self.bounds = bounds
+
+    @multi_voxel_fit
+    def fit(self, data, mask=None):
+        """ Fit method of the Ivim model class.
+
+        The fitting takes place in the following steps: Linear fitting for D
+        (bvals > 200) and store S0_prime. Another linear fit for S0 (bvals <
+        200).Estimate f using 1 - S0_prime/S0. Use least squares to fit D_star
```
etpeterson: The 200s here should be split_b_D and split_b_S0. And mention the defaults like you do for f_threshold below.

etpeterson: Also, "Use non-linear least squares..."

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,541 @@
...
+        and f.
+
+        We do a final fitting of all four parameters and select the set of
```
etpeterson: "... final non-linear fitting..."

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,541 @@
...
+        parameters which make sense physically. The criteria for selecting a
+        particular set of parameters is checking the perfusion fraction.
+        If the fraction is more than `f_threshold` (default: 25%), we will
+        reject the solution obtained from least squares fitting and consider
+        only the linear fit.
```
etpeterson: Would it be valid to say here that with noiseless or very low noise bi-exponential data the f_threshold is best at 100? If so maybe that would be something good to say here to help the user and that may be a case where having f_threshold at 25% could be a truly bad assumption.

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,541 @@
...
+            should be analyzed that has the shape data.shape[:-1]
+
+        Returns
+        -------
+        IvimFit object
+        """
+        # Get S0_prime and D.
+        S0_prime, D = self.estimate_linear_fit(data, self.split_b_D, less_than=False)
+        # Get S0 and D_star_prime.
+        S0, D_star_prime = self.estimate_linear_fit(data, self.split_b_S0,
```
etpeterson: Do you define `S0_prime` or `D_star_prime` anywhere? It probably would be good to help explain what is going on here. If not I think here is fine because they're kind of intermediate variables not truly related to the model. But `S0_prime` is just the intercept of the diffusion signal, `D_star_prime` is the linear estimate of `D_star`.

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,541 @@
...
+            bvals_split = self.gtab.bvals[self.gtab.bvals >= split_b]
+            D, neg_log_S0 = np.polyfit(bvals_split,
+                                       -np.log(data[self.gtab.bvals >=
+                                                    split_b]), 1)
+
+        S0 = np.exp(-neg_log_S0)
+        return S0, D
+
+    def estimate_f_D_star(self, params_f_D, data, S0, D):
+        """Estimate D_star using the values of all the other parameters obtained before
```
etpeterson: "Estimate f and D_star..."

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,541 @@
...
+                              args=(self.gtab, data, S0, D),
+                              gtol=gtol,
+                              xtol=xtol,
+                              ftol=ftol,
+                              epsfcn=epsfcn,
+                              maxfev=maxfev)
+                f, D_star = res[0]
+                return f, D_star
+            except:
+                warningMsg = "x0 obtained from linear fitting is unfeasibile as "
```
etpeterson: I'm confused by this warningMsg. It's if leastsq fails but the warning is that x0 from the linear fitting is not feasible. Is that really true or is that just a common cause of leastsq failing? Maybe preface what you have with "Non-linear fitting using leastsq failed"?

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,541 @@
...
+                                    bounds=((0., 0.), (self.bounds[1][1], self.bounds[1][2])),
+                                    args=(self.gtab, data, S0, D),
+                                    ftol=ftol,
+                                    xtol=xtol,
+                                    gtol=gtol,
+                                    max_nfev=maxfev)
+                f, D_star = res.x
+                return f, D_star
+            except:
+                warningMsg = "x0 obtained from linear fitting is unfeasibile "
```
etpeterson: same as previous comment

dipy/reconst/ivim.py
sahmed95: This is for consistency I guess. In case later this is extended to tensors.

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,541 @@
...
+                warningMsg += "linear fit."
+                warnings.warn(warningMsg, UserWarning)
+                f, D_star = params_f_D
+                return f, D_star
+
+    def predict(self, ivim_params, gtab, S0=1.):
+        """
+        Predict a signal for this IvimModel class instance given parameters.
+
+        Parameters
```
etpeterson: Should also have gtab and S0 in here as parameters.

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,541 @@
...
+                              args=(self.gtab, data),
+                              gtol=gtol,
+                              xtol=xtol,
+                              ftol=ftol,
+                              epsfcn=epsfcn,
+                              maxfev=maxfev)
+                ivim_params = res[0]
+                return ivim_params
+            except:
+                warningMsg = "x0 obtained from linear fitting is unfeasibile as "
```
etpeterson: Same question as `estimate_f_D_star`

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,541 @@
...
+                                    ftol=ftol,
+                                    xtol=xtol,
+                                    gtol=gtol,
+                                    max_nfev=maxfev,
+                                    args=(self.gtab, data),
+                                    x_scale=self.x_scale)
+                ivim_params = res.x
+                return ivim_params
+            except:
+                warningMsg = "x0 obtained from linear fitting is unfeasibile as "
```
etpeterson: And again

etpeterson:
I suppose you could be really specific about your tests and test the `estimate_` functions individually. It may help narrow down errors in the future.

sahmed95:
@etpeterson The gtab instead of bvals is for consistency. With regards to the warning message, let me write some tests to raise the warning. This happens for x0 values being infeasible or out of bounds.

sahmed95:
I found out the issue which was giving an unfeasible x0 for estimating f and D_star using non-lienar fitting. Turns out that after a linear fitting the f value turns out to be -ve. The corresponding singnal can as well be fitted using a single exponential decay. What should we do for such cases ? 

Also, in some cases, the linear fit itself gives a f value of > .25 which is returned as the current code does not check results from a linear fit. Need to deal with such cases too.

Here is a plot for when f is -ve after a linear fitting.

unfeasible

etpeterson:
Maybe the linear fitting should check to see if the results are reasonable (e.g. >0) and if not set to those. As for what to do about mono-exponential decays, I don't know if there's a good way to detect overfitting. I suppose in those cases it would be good if it returned 0 for f and D* to reflect that but I'm not sure how easy to detect that is. What are your thoughts?

Higher f values can be totally fine. Maybe the `f_threshold` should be a change from linear to nonlinear rather than a hard cutoff.

sahmed95:
In my opinion, it is better to have a nan set instead of 0 whenever linear
fit gives an erroneous result. As @Ariel had once mentioned, we want to let
the user have control over what to do about the erroneous values. Both
extreme cases are parametrized by `f`. If `f` becomes very high then we are
moving to a mono exponential signal and if `f` or any parameter turns out
to be negative the data has a lot of noise in the bvals < 200 part. I think
in both cases, instead of forcing the result to 0,  we should let it be nan
or atleast results from the linear fit. I am particularly not very happy
about setting an upper threshold for `f` inside the code. I would rather
prefer whatever the linear fit gives and let users decide.

etpeterson:
So return Nan in f and D* to indicate a monoexponential or at least an out of bounds result?

sahmed95:
@etpeterson @arokem I added a flag for two stage fitting and a method in the IvimFit class `fill_na` which can be used to fill nan values for voxels where the f is not in bounds and returns a new fit object. This way a user has control over what to do for un physical parameters. The two stage flag can be switched off for a faster linear fit.

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,568 @@
...
+
+        bounds : tuple of arrays with 4 elements, optional
+            Bounds to constrain the fitted model parameters. This is only
+            supported for Scipy version > 0.17. When using a older Scipy
+            version, this function will raise an error if bounds are different
+            from None. This parameter is also used to fill nan values for out
+            of bounds parameters in the `IvimFit` class using the method fill_na.
+            default : ([0., 0., 0., 0.], [np.inf, .3, 1., 1.])
+		
+		two_stage : bool
```
arokem: This line is indented too far.

sahmed95:
@arokem I implemented a conditional filling of nan values in the function `fill_na()` in the `IvimFit` class. The user has the option of specifying what to fill using the keyword argument `fill`. For example `fill=[np.nan, np.nan, True, 0.]` fills NaN for S0 and f, does not replace D_star and fills 0. for D. Similarly, `fill=[True, True, True, True]` leaves the array unchanged. @etpeterson Is this okay ? 

I need some help converting the multivoxel fit array into a callable array. Currently, I am returning `IvimFit(model, params)` which makes it an ndarray with each element being a `IvimFit` object. This doesn't allow calls such as `fit.model_params[0,0,0]` and instead the call has to be `fit[0,0,0].model_params`. I think I have to use the `CallableArray` class from https://github.com/nipy/dipy/blob/master/dipy/reconst/multi_voxel.py#L102 but a straight forward `return CallableArray(IvimFit(model, params))` does not work and throws the error `TypeError: expected sequence object with len >= 0 or a single integer`

Any suggestions ?
Thank you.

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,585 @@
...
+            default : (np.nan, np.nan, np.nan, np.nan)
+
+        Returns
+        -------
+        IvimFit : IvimFit object
+        """
+        params = self.model_params
+        bounds_violated = ~(np.all(params[..., :] >= bounds[
+                            0]) and np.all(params[..., :] <= bounds[1]))
+        filled_params = np.multiply(np.array(fill), params)
```
etpeterson: This is just `np.array(fill) * params` right? I think that would be more clear than using np.multiply.

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,585 @@
...
+            `fill=(True, 0., 0., True)`. Using `False` for any fill will set
+            it to 0.
+            default : (np.nan, np.nan, np.nan, np.nan)
+
+        Returns
+        -------
+        IvimFit : IvimFit object
+        """
+        params = self.model_params
+        bounds_violated = ~(np.all(params[..., :] >= bounds[
```
etpeterson: I think `not `rather than `~` would be more consistent with `and` later on, and as far as I know it's the same operation with `True` and `False` and should be cleaner to read.

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,585 @@
...
+        signal : array
+            The signal values predicted for this model using its parameters.
+
+        """
+        return ivim_prediction(self.model_params, gtab)
+
+    def fill_na(self, bounds=((0., 0., 0., 0), (np.inf, 0.30, 0.01, 0.001)),
+                fill=(np.nan, np.nan, np.nan, np.nan)):
+        """
+        Function to fill nan values for parameters which are not within the bounds.
```
etpeterson: The documentation here is a little confusing. At first I thought it was looking for any parameter but later it reads like only `f` is checked. But then looking at the code it's checking them all. Just some rewording here is all I'm thinking of.

dipy/reconst/ivim.py
etpeterson: Also, is this working as you intend? I think the `np.all` call is collapsing the whole array down to a single `True` or `False`. It then would replace all the parameters in the array rather that each element individually. So unless I'm misunderstanding what you're trying to do with this function (or it's supposed to be single voxel only) I don't see how it would work. Am I wrong here? To me I think you want `np.any` with the correct axis specified.

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,585 @@
...
+
+        Returns
+        -------
+        signal : array
+            The signal values predicted for this model using its parameters.
+
+        """
+        return ivim_prediction(self.model_params, gtab)
+
+    def fill_na(self, bounds=((0., 0., 0., 0), (np.inf, 0.30, 0.01, 0.001)),
```
etpeterson: Where is this going to be used in the fitting, or is that still to come?

etpeterson: Oh, never mind, looks like it's on the fit once that's done. Please disregard!

etpeterson:
I added some comments about the function inline but regarding your question about the IvimFit object, can you convert it to an array using `np.asarray(IvimFit_object_array)`? If that works it would return an ndarray. Is that the kind of thing you're looking for? @arokem probably has a good thought on this one though.

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,595 @@
...
+                              params_f_D_star,
+                              args=(self.gtab, data, S0, D),
+                              gtol=gtol,
+                              xtol=xtol,
+                              ftol=ftol,
+                              epsfcn=epsfcn,
+                              maxfev=maxfev)
+                f, D_star = res[0]
+                return f, D_star
+            except:
```
arokem: Would be good to specify the exception class that you expect to get here.

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,595 @@
...
+                                    bounds=((0., 0.), (self.bounds[1][1],
+                                                       self.bounds[1][2])),
+                                    args=(self.gtab, data, S0, D),
+                                    ftol=ftol,
+                                    xtol=xtol,
+                                    gtol=gtol,
+                                    max_nfev=maxfev)
+                f, D_star = res.x
+                return f, D_star
+            except:
```
arokem: Same here -- what exception class would be acceptable to run the next block of code?

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,595 @@
...
+                              x0,
+                              args=(self.gtab, data),
+                              gtol=gtol,
+                              xtol=xtol,
+                              ftol=ftol,
+                              epsfcn=epsfcn,
+                              maxfev=maxfev)
+                ivim_params = res[0]
+                return ivim_params
+            except:
```
arokem: Same here -- specify the exception class.

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,595 @@
...
+                                    bounds=bounds,
+                                    ftol=ftol,
+                                    xtol=xtol,
+                                    gtol=gtol,
+                                    max_nfev=maxfev,
+                                    args=(self.gtab, data),
+                                    x_scale=self.x_scale)
+                ivim_params = res.x
+                return ivim_params
+            except:
```
arokem: And here - same comment.

dipy/reconst/ivim.py
```diff
@@ -0,0 +1,595 @@
...
+        """
+        self.model = model
+        self.model_params = model_params
+
+    def __getitem__(self, index):
+        model_params = self.model_params
+        N = model_params.ndim
+        if type(index) is not tuple:
+            index = (index,)
+        elif len(index) >= model_params.ndim:
```
arokem: These lines (L503-L505) are not currently being tested. Could you add a little test for error handling in indexing?

sahmed95:
Using scipy > 0.17 doesn't really give us any advantage right now. We are already doing a linear fitting and getting a good x0. Using 0.17 just allows us to pass bounds while fitting, but we are anyways doing a check for bounds after that and returning values from the linear fitting if bounds are violated. Having two different scipy versions makes the code complicated as well as the tests. If there is no other specific advantage, in my opinion we can get rid of the scipy 0.17 dependence completely.

dipy/reconst/tests/test_ivim.py
```diff
@@ -0,0 +1,423 @@
...
+    assert_array_almost_equal(fit.model_params, linear_fit_params)
+    assert_array_almost_equal(fit.predict(gtab), linear_fit_signal)
+
+
+def test_warning():
+    """
+    Test for warning when linear fit gives parameters which are not feasible.
+    """
+    # This signal gives negative values for the parameters
+    if SCIPY_VERSION > '0.17':
```
arokem: This is not a great way to test for the version of a package. Use `LooseVersion` instead

arokem:
I don't think that adding bounds upfront is exactly the same as pruning out-of-bounds results in post-hoc. Unless I am missing something, I think we should leave this as-is, including the added complexity of checking for Scipy 0.17. Eventually (a long time from now), we'll drop support for earlier versions of scipy, and we'll be able to remove all these checks.

In my opinion, once you've sorted out the current test failures on Travis (looks like checks for warnings are not working...), this is ready to go.

sahmed95:
@arokem It seems that the warning is only shown for Python versions < 3.3. I have removed the test for now as I am not sure whether a conditional test based on Python version is okay. Makes me rethink whether we need those warning messages. Instead we could mention it in the doc.

arokem:
This is still failing on older versions of scipy: https://travis-ci.org/nipy/dipy/jobs/159431132

arokem:
Looks like this is still failing with older versions of Scipy? https://travis-ci.org/nipy/dipy/jobs/159584745

arokem:
One potential solution is to raise the minimal scipy version required. I am testing that out here: 

https://github.com/arokem/dipy/pull/28

arokem:
Yeah - looks like that should do the trick: https://travis-ci.org/arokem/dipy/jobs/160846233

sahmed95:
Should I change the travis.yml file and commit ? 

Also, the test which fails is a test for noisy data. I am assuming that whenever fitting fails while using leastsq, or bounds are violated, the results for a simple linear fit is returned at least. I think that is a sensible thing to do rather than sending nan or clipping all values to 0.

arokem:
It's worth understanding this failure first. Does it always happen, or is it stochastic? If it is stochastic, under what conditions does it happen? 

I agree that it might be reasonable to return the results of the linear fit for really messy data, but then this distinction should somehow be made, and the test should also reflect that, right?

See also discussion about this here: https://mail.python.org/pipermail/neuroimaging/2016-September/001175.html

sahmed95:
> and the test should also reflect that, right?

[test_leastsq_failing()](https://github.com/nipy/dipy/pull/1110/files#diff-074ac59cfbb165f9b3220f2a797f3d8cR397) reflects this. In recent Scipy versions the parameter values returned for noisy data after a linear fit are : 
[4.35626890e+03  -1.97516012e-02   5.47370742e-04   6.93668428e-04] 

while in the older Scipy versions, the params are : 
[  4.35626890e+03,   1.35121509e+02,   6.93404658e-04, 6.93668428e-04]. 

The linear fitting still gives either too high or negative values of f. Right now, I am checking for bounds only once, after the leastsq fitting and if bounds are violated the linear fit result are returned. We could perhaps skip this test for values of 'f' like in test_leastsq_failing(). I want the results of linear fit to be returned rather than setting to 0 or nan. Maybe we can have a default keyword argument `fill_na = (False, False, False, False)` which can be set to True to fill NaN's for violated parameters and left as False to return parameters as they are. (I had tried this as a separate function using np.where before)

arokem:
Sorry for being slow here. This all seems fine to me. So - is this ready? 

Unless someone raises any issues that I am not considering, I can go ahead and merge this early next week.

arokem:
Congratulations @sahmed95 ! You're GSoC 2016 contribution is now in `master`!

etpeterson:
Congrats @sahmed95!

sahmed95:
Thank you very much. Feels awesome to get my first open source project merged. Hopefully, this is the first of many to come ! :-)

arokem:
@sahmed95 
@etpeterson 
@RafaelNH 

Are you interested in writing a short paper for JORS (http://openresearchsoftware.metajnl.com/) or JOSS (http://joss.theoj.org/). JOSS is probably a bit easier (short!), given that we've already done a lot of the documentation here, but they still have to resolve whether they allow papers on PRs (see https://github.com/openjournals/joss/issues/52). I'll ping them there again.

arokem:
I reread the discussion on the JOSS issue. It looks like they are OK with us submitting a PR as a paper. Should we start with that?

etpeterson:
That's a good idea. I'm in.

sahmed95:
@arokem @RafaelNH @etpeterson. Please have a look. I might need to update your affiliations and orcid's. I created a new ID for myself.

etpeterson:
I got through the waitlist on the GSoC mentor summit, so I'm planning to go to that in a little over 2 weeks! I've never been, so if anyone (@arokem?) has I'd be interested in hearing about it.

While I'm there is there anything I should mention/suggest/tout? There are lightning talks (~3 mins each) about projects, so @sahmed95 if you want I could probably present that for you if you want to create one. Here are the talks for last year for reference: https://drive.google.com/drive/folders/0B-c5jwGzSQyPOHZyUVdNV2hFRUk

Also, if I'm forgetting anything then let me know.

arokem:
Cool! I haven't been. Please do report back about your experience!

sahmed95:
@etpeterson That's great ! You could talk about the fitting method we implemented and the profiling of optimize vs leastsq for performance. Do you want me to prepare a small talk, introducing the IVIM model in a simple way and discussing the issues we faced while implementing the fitting ?

etpeterson:
I took another look at the lightning talks and it looks like it's typically a group reporting on a few projects as a whole. Maybe that would be something to do for Dipy as a group?

arokem:
I think that would be great. We had so many nice projects this summer, on a
whole range of stuff, so it would be good to show-case that.

@garyfallidis: where do the other projects stand? Would it be useful if
Eric reported on all the projects at the GSoC summit? Maybe we can ask each
one of the students to make a single slide summarizing what they did? We
can also share an overview slide with you, so that you can open with that.

etpeterson:
Just bringing this back to the top. Any interest in putting something together for me to present? Don't feel obliged, but I'm happy to do it if you want.

etpeterson:
The mentor summit was awesome. It was basically all weekend and was all about various aspects of FOSS and basically whatever people wanted to organize. It is all about discussing and talking with people, really a summit not a conference with set proceedings.

Also, if you get a chance, I'd definitely try to do a lightning talk. Just a really quick way to tout our work. Dipy could do a really cool one just by showing images, most of the talks are about adding features or support or something and are good but not showy. And a few talks mentioned projects from previous years so pretty much anything goes as long as you don't go over time :-)

sahmed95:
Hi, my poster - **Make your code count : Writing open source code for science and research(en)** got selected for PyCon Japan 2017 and I will be using what I learnt on this project to make a case for more open source contributions in scientific computing.

https://pycon.jp/2017/en/proposals/vote/166/

@etpeterson (responding very late perhaps) lightning talks seem like a good idea. I will try to find some events where I could do so. I was just so caught up with college. Finally I have the time to do all these stuff. :-)

etpeterson:
Congrats! That should be fun and interesting. Hopefully it's not all in Japanese :-)

sahmed95:
Nah, English. I got lucky they have English options. But I think you can't git in Japanese anyway :). Just joined a group here near Tokyo at Riken. Members are European / British / Chinese / Swiss. Japanese is very difficult. Good thing is people here are interested in learning about open source scientific computing and how to contribute.

arokem:
Congratulations Shahnawaz. That's great!

PR #1113, opened 2016-09-01
TST: Test for invariance of model_params to splitting of the data.

arokem:
This looks like a bug to me: it shouldn't matter whether you fit the
entire data array or just one.

This happens because min_signal currently depends on the entire data array
that is passed, and that varies here depending on whether only one voxel
or the entire array is passed.

For now, I only have the bug and no fix, hence the WIP. Suggestions most welcome!

### DIFF
--- a/dipy/reconst/dki.py
+++ b/dipy/reconst/dki.py
@@ -6,7 +6,7 @@ import numpy as np
 from dipy.reconst.dti import (TensorFit, mean_diffusivity, axial_diffusivity,
                               radial_diffusivity, from_lower_triangular,
                               lower_triangular, decompose_tensor,
-                              _min_positive_signal)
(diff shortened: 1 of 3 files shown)


arokem:
How about this for a solution? Removing the WIP tag.

arokem:
Anyone want to give this one a review? It's a rather simple bug that really shouldn't be here, with a rather simple fix (I think) -- less than 40 lines of changes.

MarcCote:
LGTM. It makes sense to use a constant value for the min signal value if none was provided.

MarcCote:
Oh, there is now a conflict because I merged your other PR.

arokem:
Rebased! 

It looks like all conflicts were in docstrings, but maybe let Travis run through this, just to be on the safe side.

Issue #1119, opened 2016-09-16
Buildbots failing with workflow merge


arokem:
I think the merge of #1087 causes the following buildbot failures (both the same): 

http://nipy.bic.berkeley.edu/builders/dipy-bdist-whl-3.3/builds/41/steps/shell_10/logs/stdio
http://nipy.bic.berkeley.edu/builders/dipy-bdist-whl-2.7/builds/43/steps/shell_10/logs/stdio

matthieudumont:
WIll look into it

matthieudumont:
Does the build only fail on osx?

It might have to do with os.path.getmtime not working the same on osx.

arokem:
Thanks for taking a look! Indeed, it seems to be an OSX thing. You can see the status of all the bots here: 

http://nipy.bic.berkeley.edu/waterfall

Also, it is failing locally on my laptop (a mac). I'll try to see if I can find a fix on my setup.

arokem:
https://github.com/nipy/dipy/pull/1120

arokem:
Looks like this was fixed in #1120

PR #1120, opened 2016-09-16
Fix 1119

arokem:
Address #1119

### DIFF
--- a/dipy/workflows/segment.py
+++ b/dipy/workflows/segment.py
@@ -14,11 +14,10 @@ class MedianOtsuFlow(Workflow):
     def get_short_name(cls):
         return 'medotsu'
 
-
     def run(self, input_files, save_masked=False, median_radius=2, numpass=5,
             autocrop=False, vol_idx=None, dilate=None, out_dir='',
             out_mask='brain_mask.nii.gz', out_masked='dwi_masked.nii.gz'):
-        """ Workflow wrapping the median_otsu segmentation method.
+        """Workflow wrapping the median_otsu segmentation method.
 
         Applies median_otsu segmentation on each file found by 'globing'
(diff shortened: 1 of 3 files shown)


arokem:
I've started a try_branch job on one of the failing bots, to make sure this helps: 

http://nipy.bic.berkeley.edu/builders/dipy-bdist-whl-2.7/builds/44

matthieudumont:
Tested it, works flawlessly.

For the XXX comments: 

-get_sub_runs does not needed to be overloaded in combined_workflow but thats it. If someone creates a new workflow or combined workflow, he does not need to overload it. 

-From what I can see, workflow.set_sub_flows_optionals has no business there. The real implementation is in combined_workflow. Probably made sense at one point in the dev and I forgot to remove it. I think it can be removed (along with the test dipy.workflows.tests.test_workflow.test_set_sub_flows_optionals).

arokem:
On Fri, Sep 16, 2016 at 10:11 AM, matthieudumont <notifications@github.com>
wrote:

> Tested it, works flawlessly.
>
> For the XXX comments:
>
> -get_sub_runs does not needed to be overloaded in combined_workflow but
> thats it. If someone creates a new workflow or combined workflow, he does
> not need to overload it.
>
> Thanks for clarifications. I am still not sure I understand. Why do we
need a class method that returns an empty list?

arokem:
This is failing coveralls because of the removal of the (unnecessary) test removed in the last commit.

arokem:
Is this GTG?

matthieudumont:
Yes GTG, do I need to merge it?

arokem:
If you have push rights here, please do. Otherwise, one of the core maintainers should come along and press that green button.

PR #1122, opened 2016-09-23
Add the reconst_dti workflow

matthieudumont:
DTI reconstruction workflow and some minor comments correction.

### DIFF
--- a/dipy/workflows/reconst.py
+++ b/dipy/workflows/reconst.py
@@ -181,8 +181,8 @@ class ReconstDtiRestoreFlow(ReconstDtiFlow):
     def get_short_name(cls):
         return 'dti_restore'
 
-    def run(self, input_files, bvalues, bvectors, mask_files, b0_threshold=0.0,
-            save_metrics=[], sigma=None, jacobian=True,
+    def run(self, input_files, bvalues, bvectors, mask_files, sigma,
+            b0_threshold=0.0, save_metrics=[], jacobian=True,
             out_dir='', out_tensor='tensors.nii.gz', out_fa='fa.nii.gz',
             out_ga='ga.nii.gz', out_rgb='rgb.nii.gz', out_md='md.nii.gz',
             out_ad='ad.nii.gz', out_rd='rd.nii.gz', out_mode='mode.nii.gz',
@@ -207,18 +207,18 @@ class ReconstDtiRestoreFlow(ReconstDtiFlow):
             mask_files : string
                 Path to the input masks. This path may contain wildcards to use
                 multiple masks at once. (default: No mask used)
+            sigma : float
+                An estimate of the variance.
             b0_threshold : float, optional
                 Threshold used to find b=0 directions (default 0.0)
             save_metrics : variable string, optional
                 List of metrics to save.
                 Possible values: fa, ga, rgb, md, ad, rd, mode, tensor, evec, eval
                 (default [] (all))
-            sigma : float, optional
-                An estimate of the variance. (default None)
             jacobian : bool, optional
                 Whether to use the Jacobian of the tensor to speed the
                 non-linear optimization procedure used to fit the tensor
-                paramters (default 0.0)
+                parameters (default True)
             out_dir : string, optional
                 Output directory (default input file directory)
             out_tensor : string, optional
--- a/dipy/workflows/reconst.py
+++ b/dipy/workflows/reconst.py
@@ -81,7 +81,6 @@ class ReconstDtiFlow(Workflow):
         out_eval : string, optional
             Name of the eigenvalues to be saved (default 'evals.nii.gz')
         """
-
         io_it = self.get_io_iterator()
 
         for dwi, bval, bvec, mask, otensor, ofa, oga, orgb, omd, oad, orad, \
@@ -98,7 +97,8 @@ class ReconstDtiFlow(Workflow):
             else:
                 mask = nib.load(mask).get_data().astype(np.bool)
 
-            tenfit, _ = get_fitted_tensor(data, mask, bval, bvec, b0_threshold)
+            tenfit, _ = self.get_fitted_tensor(data, mask, bval, bvec,
+                                               b0_threshold)
 
             if not save_metrics:
                 save_metrics = ['fa', 'md', 'rd', 'ad', 'ga', 'rgb', 'mode',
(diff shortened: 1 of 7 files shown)


dipy/workflows/reconst.py
```diff
@@ -0,0 +1,151 @@
...
+            Path to the input masks. This path may contain wildcards to use
+            multiple masks at once. (default: No mask used)
+        b0_threshold : float, optional
+            Threshold used to find b=0 directions (default 0.0)
+        out_dir : string, optional
+            Output directory (default input file directory)
+        out_tensor : string, optional
+            Name of the tensors volume to be saved (default 'tensors.nii.gz')
+        out_fa : string, optional
+            Name of the fractionnal anisotropy volume to be saved
```
arokem: Typo: "fractionnal" => "fractional"

dipy/workflows/reconst.py
```diff
@@ -0,0 +1,151 @@
...
+            (default 'fa.nii.gz')
+        out_ga : string, optional
```
arokem: Isn't GA the same as FA for DTI?

dipy/workflows/reconst.py
```diff
@@ -0,0 +1,151 @@
...
+        out_evec : string, optional
+            Name of the eigen vectors volume to be saved
+            (default 'evecs.nii.gz')
+        out_eval : string, optional
+            Name of the eigen vvalues to be saved (default 'evals.nii.gz')
+        """
+
+        io_it = self.get_io_iterator()
+
+        for dwi, bval, bvec, mask, otensor, ofa, oga, orgb, omd, oad, ord, \
```
arokem: `ord` is a builtin function in Python. Could you please use another name?

dipy/workflows/reconst.py
```diff
@@ -0,0 +1,151 @@
...
+    def get_short_name(cls):
+        return 'dti'
+
+    def run(self, input_files, bvalues, bvectors, mask_files, b0_threshold=0.0,
+            out_dir='', out_tensor='tensors.nii.gz', out_fa='fa.nii.gz',
+            out_ga='ga.nii.gz', out_rgb='rgb.nii.gz', out_md='md.nii.gz',
+            out_ad='ad.nii.gz', out_rd='rd.nii.gz', out_mode='mode.nii.gz',
+            out_evec='evecs.nii.gz', out_eval='evals.nii.gz'):
+
+        """ Workflow for tensor reconstruction and DTI metrics computing.
```
arokem: => "Workflow for tensor reconstruction and for computing DTI metrics"

dipy/workflows/reconst.py
```diff
@@ -0,0 +1,151 @@
...
+        It a tensor recontruction on the files by 'globing' ``input_files`` and
```
arokem: "recontruction" => "reconstruction"

arokem: Also, maybe: "Compute a tensor reconstruction...".

dipy/workflows/reconst.py
```diff
@@ -0,0 +1,151 @@
...
+        saves the dti metrics in a directory specified by ``out_dir``.
```
arokem: "dti" => "DTI"

dipy/workflows/reconst.py
```diff
@@ -0,0 +1,151 @@
...
+            (default 'md.nii.gz')
+        out_ad : string, optional
+            Name of the axial diffusivity volume to be saved
+            (default 'ad.nii.gz')
+        out_rd : string, optional
+            Name of the radial diffusivity volume to be saved
+            (default 'rd.nii.gz')
+        out_mode : string, optional
+            Name of the mode volume to be saved (default 'mode.nii.gz')
+        out_evec : string, optional
```
arokem: Is there any way to *not* save all these volumes? For example, if all you want is FA?

dipy/workflows/reconst.py
```diff
@@ -0,0 +1,151 @@
...
+            Name of the eigen vectors volume to be saved
+            (default 'evecs.nii.gz')
+        out_eval : string, optional
+            Name of the eigen vvalues to be saved (default 'evals.nii.gz')
```
arokem: "eigen vvalues" => "eigenvalues"

dipy/workflows/reconst.py
```diff
@@ -0,0 +1,151 @@
...
+        out_ad : string, optional
+            Name of the axial diffusivity volume to be saved
+            (default 'ad.nii.gz')
+        out_rd : string, optional
+            Name of the radial diffusivity volume to be saved
+            (default 'rd.nii.gz')
+        out_mode : string, optional
+            Name of the mode volume to be saved (default 'mode.nii.gz')
+        out_evec : string, optional
+            Name of the eigen vectors volume to be saved
```
arokem: "eigen vectors"=> "eigenvectors" (one word).

dipy/workflows/reconst.py
```diff
@@ -0,0 +1,151 @@
...
+        out_eval : string, optional
+            Name of the eigen vvalues to be saved (default 'evals.nii.gz')
+        """
+
+        io_it = self.get_io_iterator()
+
+        for dwi, bval, bvec, mask, otensor, ofa, oga, orgb, omd, oad, ord, \
+            omode, oevecs, oevals in io_it:
+
+            logging.info('Computing dti metrics for {0}'.format(dwi))
```
arokem: "dti" => "DTI"

dipy/workflows/reconst.py
```diff
@@ -0,0 +1,151 @@
...
+            logging.info('All dti metrics saved in {0}'.
+                         format(os.path.dirname(oevals)))
+
+
+def get_fitted_tensor(data, mask, bval, bvec, b0_threshold=0):
+    logging.info('Tensor estimation...')
+    bvals, bvecs = read_bvals_bvecs(bval, bvec)
+    gtab = gradient_table(bvals, bvecs, b0_threshold=b0_threshold)
+
+    tenmodel = TensorModel(gtab)
```
arokem: What about options into the `TensorModel` object `__init__`? For example, what if we want to use RESTORE as a fitting method?

dipy/workflows/reconst.py
```diff
@@ -0,0 +1,151 @@
...
+                mask = nib.load(mask).get_data().astype(np.bool)
+
+            tenfit, _ = get_fitted_tensor(data, mask, bval, bvec, b0_threshold)
+
+            FA = fractional_anisotropy(tenfit.evals)
+            FA[np.isnan(FA)] = 0
+            FA = np.clip(FA, 0, 1)
+
+            tensor_vals = lower_triangular(tenfit.quadratic_form)
+            correct_order = [0, 1, 3, 2, 4, 5]
```
arokem: Isn't there a module-level variable for that in the DTI module? Might be better to coordinate across with whatever is there.

matthieudumont: Couldnt find this in dti or anywhere else. Do you want me to add it in dti.py?

arokem: Yeah - that's probably where it belongs, if it's useful in other places. 

I see now that this not in dti.py. Instead there's these other reordering variables: 

https://github.com/nipy/dipy/blob/master/dipy/reconst/dti.py#L1734
https://github.com/nipy/dipy/blob/master/dipy/reconst/dti.py#L1759

But not this one.

BTW - why is this the "correct order"?

dipy/workflows/reconst.py
matthieudumont: If I give the option to choose the method, there is the 'problem' of method specific parameters. I might have to add sigma, jacobian and weighting parameters to this workflow and only use them when the  fitting method needs them. Does that sound good?

arokem: Is there any way to pass `**kwargs` in here?

matthieudumont: From what I can see there is no way to do it that is not very ugly and error prone. If you guys have an idea im really open to it.

Right now what I can see working is adding a string param and parsing it as a dict for kwargs but im not sure it is the kind of solution you were thinking about. 

Also, I might make generic dti_reconst workflow and make a child for each tensor model (with good params).  This would actually make a callabable script per tensor_model, probablu not what you wanted too. 

Or like I said just add the three params possibilities and use it when needed.

arokem: actually, I'd be fine with a few different tensor command lines. This one would be the vanilla `dipy_tensor` and (at least) another one for `dipy_tensor_restore`

dipy/workflows/reconst.py
matthieudumont: Added a system to do that, tell me what you think about it.

arokem:
Sorry - I just notice now that this has no testing. Please add tests.

matthieudumont:
About the tests, I was wondering, should testing the actual workflows be handled by dipy_qa or do we need to add all results groundthruths (fa, rgb, rd etc.) to the get_data method? This would be a lot of new data considering the upcoming workflows with tracts.
@arokem @Garyfallidis

arokem:
You can run a minimal test of this with small amounts of data without doing
a full end-to-end. It's a completely different kind of testing than what
dipy_qa would provide.

I see that you've commented out these tests:
https://github.com/nipy/dipy/blob/master/dipy/tests/test_scripts.py. You
can use these as an example for how to write such tests.

dipy/workflows/reconst.py
matthieudumont: Just added a subclass for restore

dipy/workflows/reconst.py
```diff
@@ -161,13 +161,101 @@ def run(self, input_files, bvalues, bvectors, mask_files, b0_threshold=0.0,
...
+            save_metrics : variable string, optional
+                List of metrics to save.
+                Possible values: fa, ga, rgb, md, ad, rd, mode, tensor, evec, eval
+                (default [] (all))
+            sigma : float, optional
+                An estimate of the variance. (default None)
+            jacobian : bool, optional
+                Whether to use the Jacobian of the tensor to speed the
+                non-linear optimization procedure used to fit the tensor
+                paramters (default 0.0)
```
arokem: typo: `paramters` => `parameters`. Also, the default value seems to be `True`.

dipy/workflows/reconst.py
```diff
@@ -161,13 +161,101 @@ def run(self, input_files, bvalues, bvectors, mask_files, b0_threshold=0.0,
...
+                Path to the input masks. This path may contain wildcards to use
+                multiple masks at once. (default: No mask used)
+            b0_threshold : float, optional
+                Threshold used to find b=0 directions (default 0.0)
+            save_metrics : variable string, optional
+                List of metrics to save.
+                Possible values: fa, ga, rgb, md, ad, rd, mode, tensor, evec, eval
+                (default [] (all))
+            sigma : float, optional
+                An estimate of the variance. (default None)
```
arokem: What does `None` do in practice? I believe you need a `weighting` kwarg in addition. See logic here: 

https://github.com/nipy/dipy/blob/master/dipy/reconst/dti.py#L1475

dipy/workflows/reconst.py
```diff
@@ -161,13 +161,101 @@ def run(self, input_files, bvalues, bvectors, mask_files, b0_threshold=0.0,
...
-    tenmodel = TensorModel(gtab)
-    tenfit = tenmodel.fit(data, mask)
+        super(ReconstDtiRestoreFlow, self).\
+            run(input_files, bvalues, bvectors, mask_files, b0_threshold,
+                save_metrics, out_dir, out_tensor, out_fa, out_ga, out_rgb,
+                out_md, out_ad, out_rd, out_mode, out_evec, out_eval)
 
-    return tenfit, gtab
+    def get_tensor_model(self, gtab):
+        return TensorModel(gtab, fit_method="RT", sigma=self.sigma,
```
arokem: And here as well - the `weighting` kwarg.

dipy/workflows/reconst.py
matthieudumont: None was the default in restore_fit_tensor but it actually raises an exception. I moved sigma out of optionals and is now requiered. 

For weighting, from what I can see, it is only passable to NLLS and not Restore.

matthieudumont:
I added tests for both workflows

arokem:
+1 for the merge on my end

PR #1123, opened 2016-09-23
Add the mask workflow

matthieudumont:


dipy/workflows/mask.py
```diff
@@ -0,0 +1,43 @@
...
+        out_mask : string, optional
+           Name of the masked file (default 'mask.nii.gz')
+        """
+        io_it = self.get_io_iterator()
+
+        for input_path, out_mask_path in io_it:
+            logging.info('Creating mask of {0}'
+                         .format(input_path))
+            data, affine = load_nifti(input_path)
+            mask = np.bitwise_and(data > greater_than, data < less_than)
```
arokem: Maybe error handling for the case where greater_than > less_than?

And testing for that case

dipy/workflows/mask.py
```diff
@@ -0,0 +1,43 @@
...
+            out_mask='mask.nii.gz'):
+
+        """ Workflow for creating a binary mask
+
+        Parameters
+        ----------
+        input_files : string
+           Path to image to be masked.
+        greater_than : float
+            Default is 0.2.
```
arokem: What's the 0.2 default all about?

Garyfallidis:
@matthieudumont what is the purpose of this PR?

matthieudumont:
@Garyfallidis : Add the mask workflow

Garyfallidis:
@matthieudumont I noticed something while playing with the current master. In setup.py the script dipy_reconst_dti_restore is not available. The others are. You may want to add it here together with dipy_mask.

matthieudumont:
Good catch, will add.

arokem:
And add a test that these things are installed and can run, please.

dipy/workflows/mask.py
matthieudumont: I guess this should not be an optional param after all, no idea where 0.2 came from when I originally did this.

dipy/workflows/mask.py
matthieudumont: Good point, will do.

matthieudumont:
About adding tests that all the available workflows are present and runable, I will make this another PR.

dipy/workflows/mask.py
```diff
@@ -14,7 +14,7 @@ class MaskFlow(Workflow):
     def get_short_name(cls):
         return 'mask'
 
-    def run(self, input_files, greater_than=0.2, less_than=np.inf, out_dir='',
+    def run(self, input_files, greater_than, less_than=np.inf, out_dir='',
```
arokem: How about `ub` and `lb` (for 'upper bound' and 'lower bound')?

setup.py
```diff
@@ -214,6 +214,8 @@ def main(**extra_args):
           data_files=[('share/doc/dipy/examples',
                        glob(pjoin('doc', 'examples','*.py')))],
           scripts      = [pjoin('bin', 'dipy_reconst_dti'),
+                          pjoin('bin', 'dipy_reconst_dti_restore'),
+                          pjoin('bin', 'dipy_mask'),
```
arokem: This all seems pretty error prone. Is there some way to automate this, instead of having to add a line in here every time?

matthieudumont: I can look into that in the upcoming script tests PR.

Garyfallidis:
This looks good. Will merge in 2 days.

matthieudumont:
bump! :)

Garyfallidis:
Sorry for the delay!

PR #1124, opened 2016-09-26
Small fix in free water DTI model

RafaelNH:
Procedure was crashing when negative S0 are present. In theory such values are implausible. However, in practice this values can be present in datasets! For example, I found this cases when processing the full CENIR dataset.

When such cases are detected all parameters are directly set to zero.

### DIFF
--- a/dipy/reconst/tests/test_fwdti.py
+++ b/dipy/reconst/tests/test_fwdti.py
@@ -295,3 +295,18 @@ def test_md_regularization():
     assert_array_almost_equal(fwefit.fa, FAref)
     assert_array_almost_equal(fwefit.md, MDref)
     assert_array_almost_equal(fwefit.f, GTF)
+
+
+def test_zero_s0():
+    # single voxel
+    gtf = 0.55
+    mevals = np.array([[0.0017, 0.0003, 0.0003], [0.003, 0.003, 0.003]])
+    S_conta, peaks = multi_tensor(gtab_2s, mevals, S0=100,
+                                  angles=[(90, 0), (90, 0)],
+                                  fractions=[(1-gtf) * 100, gtf*100], snr=None)
+    S_conta[gtab_2s.bvals == 0] = -100
+    fwdm = fwdti.FreeWaterTensorModel(gtab_2s, 'NLS')
+    fwefit = fwdm.fit(S_conta)
+    assert_array_almost_equal(fwefit.fa, 0.0)
+    assert_array_almost_equal(fwefit.md, 0.0)
+    assert_array_almost_equal(fwefit.f, 0.0)
--- a/dipy/reconst/tests/test_fwdti.py
+++ b/dipy/reconst/tests/test_fwdti.py
@@ -297,7 +297,7 @@ def test_md_regularization():
     assert_array_almost_equal(fwefit.f, GTF)
 
 
-def test_zero_s0():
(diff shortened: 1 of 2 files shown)


arokem:
Negative signal?! Must be some weird artifact of preprocessing. Was the error you got very hard to decipher? Otherwise I am thinking that maybe throwing an error might actually be the right thing to do.

RafaelNH:
Hi @arokem! I know that some eddy currents artifact correction algorithms can produce this negative values - it might be the case here. In dti and dki models we don't have this problem because all intensities smaller that min_signal are replaced by min_signal. My first attempt to correct this problem here was to follow the same idea. If we do that, FA and MD values for these cases tend to zero however f values can tend to any value on its range [0-1]. The correction that I am doing here insures that f is also set to zero for these cases. The only way that users can manually correct this problem is indeed replacing negative or zero values with a min signal intensity, thus I am not sure if adding an error message is a good idea. I don't want that the processing of the fwdti model stops for a problem that dti and dki automatically deals. Also these problematic voxels are normally only present in regions of the skull or background. What do you think? Let me know your coding suggestions I will address them asap.

arokem:
All clear - thanks for the explanation. I'll go ahead and review.

dipy/reconst/tests/test_fwdti.py
```diff
@@ -295,3 +295,18 @@ def test_md_regularization():
     assert_array_almost_equal(fwefit.fa, FAref)
     assert_array_almost_equal(fwefit.md, MDref)
     assert_array_almost_equal(fwefit.f, GTF)
+
+
+def test_zero_s0():
```
arokem: Should this be `test_negative_s0`?

dipy/reconst/tests/test_fwdti.py
```diff
@@ -295,3 +295,18 @@ def test_md_regularization():
...
+    mevals = np.array([[0.0017, 0.0003, 0.0003], [0.003, 0.003, 0.003]])
+    S_conta, peaks = multi_tensor(gtab_2s, mevals, S0=100,
+                                  angles=[(90, 0), (90, 0)],
+                                  fractions=[(1-gtf) * 100, gtf*100], snr=None)
+    S_conta[gtab_2s.bvals == 0] = -100
+    fwdm = fwdti.FreeWaterTensorModel(gtab_2s, 'NLS')
+    fwefit = fwdm.fit(S_conta)
+    assert_array_almost_equal(fwefit.fa, 0.0)
+    assert_array_almost_equal(fwefit.md, 0.0)
+    assert_array_almost_equal(fwefit.f, 0.0)
```
arokem: Would you mind adding a multi-voxel case where one voxel has negative S0 and others do not, just to make sure that we don't get cross-talk (that is - that the parameters in other voxels are unaffected)?

dipy/reconst/tests/test_fwdti.py
RafaelNH: Done =)

dipy/reconst/tests/test_fwdti.py
RafaelNH: Done =)

arokem:
This LGTM. Considering this is a small fix, I will go ahead and merge this in a couple of days, unless someone voices opposition.

Issue #759, opened 2015-11-02
How to resolve python-vtk6 link issues in Ubuntu 


Garyfallidis:
Moving back to python-vtk6 from vtk5(python-vtk) creates an error

```
<ipython-input-1-b7e11aadda62> in <module>()
----> 1 import vtk

/usr/lib/python2.7/dist-packages/vtk/__init__.py in <module>()
37 
38 # --------------------------------------
---> 39 from vtkCommonCore import *
40 from vtkCommonMath import *
41 from vtkCommonMisc import *

/usr/lib/python2.7/dist-packages/vtk/vtkCommonCore.py in <module>()
----> 1 from vtkCommonCorePython import *

ImportError: No module named vtkCommonCorePython
```

The resolution was to relink the broken so files

```
So I went to /usr/lib/x86_64-linux-gnu and ran for i in *-6.0.so.6.0; do sudo ln -s $i ${i/-6.0.so.6.0/-6.0.so}; done. 
```

(May need sudo first too)

The rest of the discussion is found here
http://stackoverflow.com/questions/13495285/importerror-no-module-named-vtkcommonpython

Garyfallidis:
@arokem is this the same error you got the other day in Travis?

arokem:
Yes. Looks the same. We should be able to run that in the vtk6 machine on
travis

MarcCote:
I saw that we have a Travis build that is supposed to use VTK6 but Ubuntu 12.04 "precise", the one currently used by Travis, doesn't offer the package `python-vtk6`. See
https://travis-ci.org/nipy/dipy/jobs/167835093#L399

This means tests related to VTK are skipped in that build. Maybe we should drop that specific Travis build (speed up the testing) or try the Travis's Trusty beta environment (https://docs.travis-ci.com/user/trusty-ci-environment)?

Also, I think Anaconda now comes with VTK 6.3 (https://anaconda.org/anaconda/vtk)

Garyfallidis:
A generic comment. Can you check again the PR for length limit? I believe you go beyond 79 characters quite a few times.

Garyfallidis:
I think what you suggest is a good idea. +1

MarcCote:
I'm not sure if we can use Trusty beta environment only for a specific build (i.e. the one testing VTK6). That might impact the other builds we have. The easiest solution would be to drop the build with Python-vtk6.

skoudoro:
This issue was moved to fury-gl/fury#5

PR #1132, opened 2016-10-15
Update .travis.yml and README.md

MarcCote:
This PR: 
- removes coveralls badge from README.md. It is currently displaying unknown and codecov seems better (see https://www.slant.co/versus/7928/7929/~coveralls_vs_codecov).
- makes sure we use COVERAGE=1 only in one Travis build. Having multiple builds with COVERAGE=1 generate several unnecessary coveralls posts in PRs. Also, currently no builds with COVERAGE=1 were running VTK tests, so the code coverage being reported wasn't accurate. 
- removes unnecessary Travis build which was supposedly testing with VTK6. However, Ubuntu 12.04 (the one use by Travis) doesn't offer the package python-vtk6 (see my comments in #759). So, that particular build was equivalent to another one that wasn't testing VTK code at all.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,12 +22,9 @@ python:
 matrix:
   include:
     - python: 2.7
-      env:
-        - COVERAGE=1
     # To test minimum dependencies
     - python: 2.7
       env:
-        - COVERAGE=1
         # Check these values against requirements.txt and dipy/info.py
         - DEPENDS="cython==0.18 numpy==1.7.1 scipy==0.9.0 nibabel==1.2.0"
     - python: 2.7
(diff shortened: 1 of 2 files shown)


arokem:
Yeah. Seems fine to me. I think that the reason we wanted to separately test for coverage with minimum dependency requirements was to cross-check between coverage reports for code where there are alternative paths depending on, for example, the scipy version (`core.optimize`). But I think that these cases need to be handled by testing in local virtual env anyhow.

@matthew-brett : was there any other reason for this that I don't remember?

arokem:
Anyone have any more thoughts here?

PR #1135, opened 2016-10-22
WIP : Multiprocessing - implemented a parallel_voxel_fit decorator 

sahmed95:
This is the original work of @MrBago from his [branch](https://github.com/MrBago/dipy/tree/nf_parallel_framework/dipy/parallel). I have refactored the code a little and added a new decorator `parallel_voxel_fit` similar to the multi_voxel_fit. Both can be merged into one later.

The current implementation adopts his code in dipy/multi. The decorator `parallel_voxel_fit` can be used directly in modules such as dipy/reconst/dsi.py in the following manner (this is also a test case):

```
class TestModel(ReconstModel):
    """
    Reconst model to test `parallel_voxel_fit` decorator.
    """
    @parallel_voxel_fit
    def fit(self, single_voxel_data):
        return ReconstFit(self, single_voxel_data.sum())
```

To do:
- [ ] Discuss a better way to activate or deactivate mutiprocessing rather than setting a global _dipy_num_cpu variable. (Is it okay ?)
- [ ] Improve documenation, refactor and optimize the current code.
- [ ] Profile the performance on multicore CPU's
- [ ] Better handling of masks and splitting of voxels for fitting.
- [ ] Simplify current multi voxel models into single voxel fitting by using the decorators `parallel_voxel_fit` or `multi_voxel_fit`. (For examole in reconst/dti.py, reconst/dki.py multi voxel fitting is carried out by iterating over voxels whereas in reconst/dsi.py the mult_voxel_fit decorator is used and the fit method is written for a single voxel.)

### DIFF
--- a/dipy/multi/__init__.py
+++ b/dipy/multi/__init__.py
@@ -1,0 +1,1 @@
+# init for the parallel framework
--- a/dipy/multi/config.py
+++ b/dipy/multi/config.py
@@ -10,7 +10,7 @@ _dipy_num_cpu = 1
 manager = None
 
 
-def activate_multiprocessing(num_cpu=None):
+def activate_multithreading(num_cpu=None):
     """
     Function to activate multiprocessing.
 
@@ -30,15 +30,15 @@ def activate_multiprocessing(num_cpu=None):
         _dipy_num_cpu = num_cpu
 
     if manager is not None:
-        # manager.shut_down()
-        raise NotImplemented()
+        manager.shut_down()
+        # raise NotImplemented()
     if _dipy_num_cpu > 1:
         manager = PoolMananger(_dipy_num_cpu)
     else:
         manager = None
 
 
-def deactivate_multiprocessing():
+def deactivate_multithreading():
     """
     Function to deactivate multiprocessing.
     """
--- a/dipy/multi/config.py
+++ b/dipy/multi/config.py
@@ -1,0 +1,80 @@
+"""
+Configuration for multiprocessing
+"""
+from multiprocessing import cpu_count, Pool
+from contextlib import contextmanager
+import atexit
+
+
+_dipy_num_cpu = 1
+manager = None
+
+
+def activate_multiprocessing(num_cpu=None):
+    """
+    Function to activate multiprocessing.
+
+    Parameters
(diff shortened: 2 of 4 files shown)


MrBago:
@sahmed95 thanks for pushing this along! One thing I ran into when writing this was that it might be a good idea to launch (either on startup or the first time a parallel function is called) and re-use the `Pool` on different parallel calls. If you do this, I believe you need to register a "pool exit" function with python or come up with some other way of quitting the pool.

The other thing that's closely related to this pool request is the implementation of a "peaks from fit" type function. This function should be like `peaks_from_model`, but take a `Fit` or `MultiVoxelFit` object as input.

I also have a few other ideas about where this kind of framework could go :), but I'll save those for later. Let me know if I can answer any questions or help in some other way.

Garyfallidis:
All good points. Currently the peaks_from_model uses a lot of memory when parallelized. We need to make sure that this does not happen in the new version where the parallelization has moved to the fitting part.

sahmed95:
Should implement pool manager from the latest [commit](https://github.com/nipy/dipy/pull/642/commits/78c9f609a9367103ba5b43db26660be0c22fa3b5)

Garyfallidis:
@sahmed95 what is the state of this PR? Can you bring us up to speed? It would be nice if we could make this work.

sahmed95:
Hi, I have updated the code implementing pool manager at launch by @MrBago. To use multiprocessing, we just need to replace the decorator `multi_voxel_fit` with `parallel_voxel_fit` at places like [this](https://github.com/nipy/dipy/blob/master/dipy/reconst/dsi.py#L128). Regarding the `peaks_from_model` type function, my understanding is that we need a way to specify if we want the fit to be implemented using multiprocessing or not. This is currently done by replacing the `multi_voxel_fit` decorators to `parallel_voxel_fit`. Instead of that, we wish to have something like reconst_model.fit(data, parallel=True) which activates the multiprocessing. Is that correct ?

> Currently the peaks_from_model uses a lot of memory when parallelized. 

Do we have an idea why that happens ?

skoudoro:
closing in favor of #1418

PR #1137, opened 2016-10-31
Small fix to insure that fwDTI non-linear procedure does not crash

RafaelNH:
Analogous to DTI non-linear code, I added some lines of code to insure that fwDTI does not crash when the non-linear procedure gives NaN.

### DIFF
--- a/dipy/reconst/tests/test_dti.py
+++ b/dipy/reconst/tests/test_dti.py
@@ -739,11 +739,11 @@ def test_decompose_tensor_nan():
     lref, vref = decompose_tensor(from_lower_triangular(D_fine))
     lfine, vfine = _decompose_tensor_nan(from_lower_triangular(D_fine),
                                          from_lower_triangular(D_alter))
-    assert_array_almost_equal(lfine, np.array(1.7e-3, 0.3e-3, 0.2e-3))
+    assert_array_almost_equal(lfine, np.array([1.7e-3, 0.3e-3, 0.2e-3]))
     assert_array_almost_equal(vfine, vref)
     
     lref, vref = decompose_tensor(from_lower_triangular(D_alter))
     lalter, valter = _decompose_tensor_nan(from_lower_triangular(D_nan),
-                                         from_lower_triangular(D_alter))
-    assert_array_almost_equal(lalter, np.array(1.6e-3, 0.4e-3, 0.3e-3))
+                                           from_lower_triangular(D_alter))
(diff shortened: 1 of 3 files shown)


arokem:
Could you please add a test? I assume you've encountered some data that raises this?

dipy/reconst/tests/test_dti.py
```diff
@@ -706,3 +707,20 @@ def test_eig_from_lo_tri():
 
     lo_tri = lower_triangular(dmfit.quadratic_form)
     assert_array_almost_equal(dti.eig_from_lo_tri(lo_tri), dmfit.model_params)
+
```
arokem: PEP8: one more line here (two lines between functions)

RafaelNH:
Hi @arokem - actually finding an concrete signal to reproduce this problem is not that trivial. I encountered this problem when processing more than 600 subjects in my units cluster. The problem only occurred in background voxels of 9 subjects. I tried to create a test based on these signals however I noticed that the previous codes were processing these signals fine with the previous code version. After looking to the source of the issue in the cluster nodes, I notice that this has happening due to an overflow of the exponential function of non-linear diffusion model fits which were given NaN tensor elements. Since in my home computer I was not getting this problem, I assume that testing this issue from signals depend on the system that you are running dipy. Therefore, I decided to test the problem from tensors with NaN elements. For this, I created a helper function with the lines of code that were dealing with this problem for the standard dti non-linear fit. With this I am reducing duplicate code and testing previously non covered lines of dti.py - the test was added in test_dti.py.

arokem:
That all makes sense - thank you for expanding on that.

RafaelNH:
@arokem - just noticed your comments! I will carry on the work tomorrow (getting late here in Cambridge). I think my last comment covers your 1st point, right? Regarding point 2 - since _decompose_tensor_nan is only a helper function, I didn't want to increase complexity and eventually decrease code speed performance. Moreover, if you see _decompose_tensor_nan is just a generalization of decompose_tensor. I would also prefer not giving a default value for tensor_alternative - since this should be used for non-linear approaches the best tensor alternative is always the parameters initial estimate.

RafaelNH:
Minor corrections done! @arokem let me know about the other 2 points -  I will resume work tomorrow.

arokem:
👍  for the merge on my end. If someone else wants to take a look, I will give you a few days before going ahead with it.

RafaelNH:
Hi @arokem - should we merge this?

arokem:
Sorry for the delay. In it goes!

PR #1141, opened 2016-11-03
Do not cover files related to benchmarks.

MarcCote:
Right now, benchmarks count towards our coverage but we are not running them on Travis. I suggest we exclude them.


arokem:
Yeah. Get it up in the 90s!

PR #1143, opened 2016-11-03
Slice through ODF fields 

Garyfallidis:
This PR introduces a new viz.actor (odf_slicer) that allows to slice ODF fields in native or world coordinates  matching the underlying anatomy. So, it works nicely with the viz.actor.slicer.

Some of the new features include applying a colormap using all the voxels in the slice or volume. 
jet_odfs

Also all the previous capabilities found in sphere_func are available too. Finally, the new function works with memory mapped ODFs and masking which helps reduce memory and video memory usage. I prefer to wait for the UI elements PR (https://github.com/nipy/dipy/pull/1140)  for creating a tutorial for this function.  With the new UI elements we can nicely upgrade some of our reconstruction and viz tutorials. But in the meantime you can review this PR with the tests.

### DIFF
--- a/dipy/viz/actor.py
+++ b/dipy/viz/actor.py
@@ -177,6 +177,12 @@ def slicer(data, affine=None, value_range=None, opacity=1.,
             im_actor.input_connection(self.output)
             im_actor.SetDisplayExtent(*self.GetDisplayExtent())
             im_actor.opacity(opacity)
+            if interpolation == 'nearest':
+                im_actor.SetInterpolate(False)
+            else:
+                im_actor.SetInterpolate(True)
+            if major_version >= 6:
+                im_actor.GetMapper().BorderOn()
             return im_actor
 
     image_actor = ImageActor()
@@ -545,7 +551,7 @@ def axes(scale=(1, 1, 1), colorx=(1, 0, 0), colory=(0, 1, 0), colorz=(0, 0, 1),
 def odf_slicer(odfs, affine=None, mask=None, sphere=None, scale=2.2,
                norm=True, radial_scale=True, opacity=1.,
                colormap='plasma', global_cm=False):
-    """ Slice spherical fields in native or wold coordinates
+    """ Slice spherical fields in native or world coordinates
 
     Parameters
     ----------
@@ -677,6 +683,8 @@ def _odf_slicer_mapper(odfs, affine=None, mask=None, sphere=None, scale=2.2,
         all_faces.append(faces + k * xyz.shape[0])
         all_ms.append(m)
 
+    # from ipdb import set_trace
+    # set_trace()
     all_xyz = np.ascontiguousarray(np.concatenate(all_xyz))
     all_xyz_vtk = numpy_support.numpy_to_vtk(all_xyz, deep=True)
 
--- a/dipy/viz/actor.py
+++ b/dipy/viz/actor.py
@@ -3,7 +3,7 @@ from __future__ import division, print_function, absolute_import
 import numpy as np
 from nibabel.affines import apply_affine
 
-from dipy.viz.colormap import colormap_lookup_table
+from dipy.viz.colormap import colormap_lookup_table, create_colormap
 from dipy.viz.utils import lines_to_vtk_polydata
 from dipy.viz.utils import set_input
 
@@ -544,7 +544,7 @@ def axes(scale=(1, 1, 1), colorx=(1, 0, 0), colory=(0, 1, 0), colorz=(0, 0, 1),
 
 def odf_slicer(odfs, affine=None, mask=None, sphere=None, scale=2.2,
                norm=True, radial_scale=True, opacity=1.,
-               colormap=None):
+               colormap=None, global_cm=False):
     """ Slice spherical fields in native or wold coordinates
 
     Parameters
@@ -567,6 +567,9 @@ def odf_slicer(odfs, affine=None, mask=None, sphere=None, scale=2.2,
         Takes values from 0 (fully transparent) to 1 (non-transparent)
     colormap : None or str
         If None then no color is used.
+    global_cm : bool
+        If True the colormap will be applied in all ODFs. If False
+        it will be applied individually at each voxel (default False).
     """
 
     if mask is None:
@@ -591,19 +594,21 @@ def odf_slicer(odfs, affine=None, mask=None, sphere=None, scale=2.2,
                                              norm=norm,
                                              radial_scale=radial_scale,
                                              opacity=opacity,
-                                             colormap=colormap)
+                                             colormap=colormap,
+                                             global_cm=global_cm)
             self.SetMapper(self.mapper)
 
     odf_actor = OdfSlicerActor()
     I, J, K = odfs.shape[:3]
-    odf_actor.display_extent(0, I, 0, J, K/2, K/2 + 1)
+    odf_actor.display_extent(0, I, 0, J,
+                             int(np.floor(K/2)), int(np.floor(K/2 + 1)))
 
     return odf_actor
 
 
 def _odf_slicer_mapper(odfs, affine=None, mask=None, sphere=None, scale=2.2,
                        norm=True, radial_scale=True, opacity=1.,
-                       colormap=None):
+                       colormap=None, global_cm=False):
 
     if mask is None:
         mask = np.ones(odfs.shape[:3])
@@ -620,7 +625,8 @@ def _odf_slicer_mapper(odfs, affine=None, mask=None, sphere=None, scale=2.2,
     all_faces = []
     all_ms = []
     for (k, center) in enumerate(ijk):
-        m = odfs[tuple(center)].copy()
+
+        m = odfs[tuple(center.astype(np.int))].copy()
 
         if norm:
             m /= abs(m).max()
@@ -645,8 +651,9 @@ def _odf_slicer_mapper(odfs, affine=None, mask=None, sphere=None, scale=2.2,
     all_faces = np.ascontiguousarray(all_faces.ravel(), dtype='i8')
     all_faces_vtk = numpy_support.numpy_to_vtkIdTypeArray(all_faces,
                                                           deep=True)
-
-    all_ms = np.ascontiguousarray(np.concatenate(all_ms), dtype='f4')
+    if global_cm:
+        all_ms = np.ascontiguousarray(
+            np.concatenate(all_ms), dtype='f4')
 
     points = vtk.vtkPoints()
     points.SetData(all_xyz_vtk)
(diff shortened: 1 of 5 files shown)


dipy/viz/actor.py
```diff
@@ -539,3 +540,186 @@ def axes(scale=(1, 1, 1), colorx=(1, 0, 0), colory=(0, 1, 0), colorz=(0, 0, 1),
     ass.AddPart(arrowz)
 
     return ass
+
+
+def odf_slicer(odfs, affine=None, mask=None, sphere=None, scale=2.2,
+               norm=True, radial_scale=True, opacity=1.,
+               colormap=None, global_cm=False):
+    """ Slice spherical fields in native or wold coordinates
```
MarcCote: wold => world

dipy/viz/actor.py
```diff
@@ -539,3 +540,186 @@ def axes(scale=(1, 1, 1), colorx=(1, 0, 0), colory=(0, 1, 0), colorz=(0, 0, 1),
...
+            cols = np.zeros((ijk.shape[0],) + sphere.vertices.shape,
+                            dtype='f4')
+            for k in range(ijk.shape[0]):
+                tmp = create_colormap(all_ms[k].ravel(), colormap)
+                cols[k] = tmp.copy()
+
+            cols = np.ascontiguousarray(
+                np.reshape(cols, (cols.shape[0] * cols.shape[1],
+                           cols.shape[2])), dtype='f4')
+        # cols = np.interp(cols, [0, 1], [0, 255]).astype('ubyte')
```
MarcCote: unused

dipy/viz/actor.py
```diff
@@ -539,3 +540,186 @@ def axes(scale=(1, 1, 1), colorx=(1, 0, 0), colory=(0, 1, 0), colorz=(0, 0, 1),
...
+        # vtk_colors = numpy_to_vtk_colors(255 * cols)
```
MarcCote: unused

Garyfallidis:
I believe we should have a PR soon to rename all test_fvtk** and a PR to update all tutorials to not be using fvtk. After we get in the UI elements we should do that. Maybe Ranveer can take the lead on that RF. But there is no need to do this RF in this PR. If you agree.

MarcCote:
Should also have a look at https://github.com/nipy/dipy/pull/944

dipy/viz/colormap.py
```diff
@@ -255,3 +261,52 @@ def line_colors(streamlines, cmap='rgb_standard'):
                     for streamline in streamlines]
 
     return np.vstack(col_list)
+
+
+lowercase_cm_name = {'blues': 'Blues', 'accent': 'Accent'}
+
+
+def create_colormap(v, name='jet', auto=True):
```
arokem: "jet" is a really bad colormap (see e.g., ). Can we use switch to using a different default (viridis??)? If we want to keep this as a default to support back-compatibility, can we at least add a warning that we will change this in the future?

dipy/viz/actor.py
```diff
@@ -539,3 +540,186 @@ def axes(scale=(1, 1, 1), colorx=(1, 0, 0), colory=(0, 1, 0), colorz=(0, 0, 1),
...
+    scale : float
+        Distance between spheres.
+    norm : bool
+        Normalize `sphere_values`.
+    radial_scale : bool
+        Scale sphere points according to odf values.
+    opacity : float
+        Takes values from 0 (fully transparent) to 1 (non-transparent)
+    colormap : None or str
+        If None then no color is used.
```
arokem: No color, or the default color ("jet"? See comment below)

dipy/viz/actor.py
```diff
@@ -539,3 +540,186 @@ def axes(scale=(1, 1, 1), colorx=(1, 0, 0), colory=(0, 1, 0), colorz=(0, 0, 1),
...
```
arokem: No color, or the default colormap?

dipy/viz/actor.py
```diff
@@ -539,3 +540,186 @@ def axes(scale=(1, 1, 1), colorx=(1, 0, 0), colory=(0, 1, 0), colorz=(0, 0, 1),
...
+        all_ms = np.ascontiguousarray(
+            np.concatenate(all_ms), dtype='f4')
+
+    points = vtk.vtkPoints()
+    points.SetData(all_xyz_vtk)
+
+    cells = vtk.vtkCellArray()
+    cells.SetCells(ncells, all_faces_vtk)
+
+    if colormap is not None:
```
arokem: I see. It really skips this block if there's no colormap argument. So what does it look like then? 

BTW - could you share the code that created the example that you pasted into the PR text? I'd like to try running this on my machine to see that it works well (on Python 3!!)

dipy/viz/colormap.py
```diff
@@ -255,3 +261,52 @@ def line_colors(streamlines, cmap='rgb_standard'):
...
+        'accent', 'bone' and matplotlib colormaps if you have matplotlib
+        installed.
+    auto : bool,
+        if auto is True then v is interpolated to [0, 10] from v.min()
+        to v.max()
+
+    Notes
+    -----
+    Dipy supports a few colormaps for those who do not use Matplotlib, for
+    more colormaps consider downloading Matplotlib.
```
arokem: Maybe a link to the MPL website?

dipy/viz/actor.py
Garyfallidis: No color for GL means all is white. Will add something.

dipy/viz/actor.py
Garyfallidis: Just take the reconst_csd.py example or even better your SFM example and replace sphere_funcs with odf_slicer. Then to change the view just select the ROI you want with the display_extent method.

dipy/viz/colormap.py
Garyfallidis: Nope there is no reason to change the default behavior of this function neither to add another warning. What we need to do is use different colormaps in our tutorials. I will add something in the documentation. To suggest using viridis or inferno. Sounds good? By the way I don't think there is one colormap that is good for all the different jobs.

dipy/viz/colormap.py
Garyfallidis: ok

dipy/viz/actor.py
```diff
@@ -539,3 +540,186 @@ def axes(scale=(1, 1, 1), colorx=(1, 0, 0), colory=(0, 1, 0), colorz=(0, 0, 1),
...
+    sphere : Sphere
+        a sphere
+    scale : float
+        Distance between spheres.
+    norm : bool
+        Normalize `sphere_values`.
+    radial_scale : bool
+        Scale sphere points according to odf values.
+    opacity : float
+        Takes values from 0 (fully transparent) to 1 (non-transparent)
```
Garyfallidis: ok

dipy/viz/colormap.py
arokem: I didn't say one colormap is good for all the different jobs you might want to do with it, just that this one is particularly bad for any task, and we shouldn't use it as the default (see: https://jakevdp.github.io/blog/2014/10/16/how-bad-is-your-colormap/). Some visualization software goes so far as to prevent users from using "jet": https://github.com/mwaskom/seaborn/commit/9f5a177639a7b5d5d8c7cce98def0e2e10c947b0

I don't think we need to go that far, but users will tend to stick to the default, so I think we should change it.

arokem: I respectfully disagree. The colormap is supposed to help derive the information from the data, but this one hides and distorts information.

Garyfallidis: Okay, so what should be the default? Ideas?

Garyfallidis: From top to bottom. Inferno, jet and viridis on a response function. Clearly jet shows the minima better. But the other have a smoother transition.

inferno
jet
viridis

arokem:
'viridis' or 'hot'

arokem:
I'll amend that: I would strongly favor 'viridis' as a default.

It was designed through a careful process that took into account the best
knowledge we have of HCI perceptual considerations.

Garyfallidis:
Did they consider spherical 3D functions? Or only images in their evaluation?

Garyfallidis:
Because they chose one that does not mean we need to chose the same. Here veridis is option 4 for example
https://bids.github.io/colormap/

So, here is magma.
magma

To me veridis looks really depressing. Too green.

Why don't we all investigate a bit which colormap serves dMRI data better? It's good we started this discussion. But we do not need to make a decision immediately. Let's play with different colormaps and she which one prevails.

Garyfallidis:
Plasma looks good too.

image

arokem:
Yes: viridis, inferno, plasma, and magma are all equally good from a
perceptual/HCI point of view.

I agree that plasma looks good, and has nice cheerful colors. I'd be happy
to replace jet with plasma as a default.

MarcCote:
Yes, plasma looks great, we can go with that. For fun, can you post one with viridis, I'm curious!

arokem:
@MarcCote : there's one with viridis here: https://github.com/nipy/dipy/pull/1143#discussion_r86457402

Not too bad, in my opinion, but definitely tending towards the green shades

MarcCote:
On further consideration, the default for the colormap should be set in the `create_colormap` function and avoid setting it in every function that accepts a colormap in input. Whether or not we should do it in this PR is up to you.

Garyfallidis:
Hello @arokem and @MarcCote I added plasma as the default for the odf_slicer and added a message in create_colormap suggesting to users to prefer other colormaps. I think that is better than raising a warning at this stage. This direction gives us time to look into different colormaps but also informs people to start looking and trying alternatives.

dipy/viz/colormap.py
```diff
@@ -255,3 +261,59 @@ def line_colors(streamlines, cmap='rgb_standard'):
...
+    -----
+    Dipy supports a few colormaps for those who do not use Matplotlib, for
+    more colormaps consider downloading Matplotlib (see matplotlib.org).
+    """
+
+    if name == 'jet':
+        msg = 'Jet is a popular colormap but can often be misleading and'
+        msg += 'we will remove it from being the default in the near future.'
+        msg += 'Try for example plasma, viridis, hot or inferno.'
+        print(msg)
```
arokem: I think the direction is right. We want to warn people of this now, and deprecate later. You probably want to raise a warning here, rather than just printing a message, though.

arokem: It's a bit strange that the default behavior raises a warning. Why not set the default to be plasma here as well? I guess we do want the default to be consistent across different functions, and you changed it to plasma in several other places.

MarcCote:
@Garyfallidis having a closer look it seems that the other test with slicers `test_slicer` is being skip (because of `@npt.dec.skipif(skip_slicer)`) on Travis since it is still using VTK5.8.
See https://github.com/nipy/dipy/blob/master/dipy/viz/tests/test_fvtk_actors.py#L25

Garyfallidis:
Closing on behalf of https://github.com/nipy/dipy/pull/1269

Issue #1142, opened 2016-11-03
pytables version and streamlines_format.py example


skoudoro:
I am still playing with dipy examples and i got a problem with streamlines_format.py.  (dipy has been installed on new environnement via conda install on OS X 10.11)

I got an import error after running this command (means pytables dependency miss on conda package ?) :
`from dipy.io.dpy import Dpy ; dpw = Dpy('fornix.dpy', 'w')` 

So I install pytables (last version - 3.3.0), and now i get the following error :
`AttributeError: 'module' object has no attribute 'openFile'`

Indeed, the function name is "open_file" with pytables 3.3.0.

So my question is which version of pytables should i use for dipy 0.11.0?

if i can continue with pytables 3.3.0, should i create a new branch for correcting this problem ?

Thanks in advance, 

Serge

Garyfallidis:
Yes, please do. Thank you @skoudoro

arokem:
We should add pytables as a conda dependency: https://github.com/conda-forge/dipy-feedstock/issues/4

PR #1147, opened 2016-11-06
add this fix for newer version of pytables. 

skoudoro:
This request resolve and refer to the issue #1142

### DIFF
--- a/dipy/io/dpy.py
+++ b/dipy/io/dpy.py
@@ -10,6 +10,8 @@
 
 import numpy as np
 
+from distutils.version import LooseVersion
+
 # Conditional import machinery for pytables
 from dipy.utils.optpkg import optional_package
 
@@ -17,9 +19,7 @@ from dipy.utils.optpkg import optional_package
 tables, have_tables, _ = optional_package('tables')
 
 # Useful variable for backward compatibility.
-if have_tables:
-    PY_TABLE2 = tables.__version__[0] == "2"
-    PY_TABLE3 = tables.__version__[0] == "3"
+TABLES_LESS_3_0 = LooseVersion(tables.__version__) < "3.0"
 
 # Make sure not to carry across setup module from * import
 __all__ = ['Dpy']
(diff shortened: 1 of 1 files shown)


Garyfallidis:
@skoudoro  please check the errors reported by Travis https://travis-ci.org/nipy/dipy/jobs/173770112

Garyfallidis:
Actually, I believe you don't even have to have two variables. One would be enough. Just check if tables is greater than 3 (using LooseVersion). One definition TABLES3 should be enough. No need for PY in front. Let me know if this suggestion is clear.  Thanks in advance @skoudoro.

skoudoro:
Thanks for the suggestion @Garyfallidis . Everything is clear, but i still think that my last commit will have a problem on travis test with `tables.__version__` when tables package is not available. what is the good way for optional package ?

Garyfallidis:
Yes, correct. You need to skip the relevant test when `has_tables` is False.

Garyfallidis:
The test is still trying to import `have_tables` but now you call it `has_tables`
```
from dipy.io.dpy import Dpy, have_tables
ImportError: cannot import name have_tables
```

skoudoro:
ok, sorry... I will configure my local environment correctly and run test locally instead of committing stupid thing...

dipy/io/dpy.py
```diff
@@ -39,50 +48,60 @@ def __init__(self, fname, mode='r', compression=0):
...
-        >>> dpw.write_track(B)
-        >>> dpw.write_track(C)
-        >>> dpw.close()
-        >>> dpr = Dpy(fname,'r')
-        >>> A=dpr.read_track()
-        >>> B=dpr.read_track()
-        >>> T=dpr.read_tracksi([0,1,2,0,0,2])
-        >>> dpr.close()
-        >>> os.remove(fname) #delete file from disk
+        >>> def dpy_example():
```
Garyfallidis: The indentation inside the function here seems incorrect. It should be 4 spaces or go back to previous version.

Garyfallidis:
Hi @skoudoro Travis looks happy. Check my last comment above on indentation and then I can merge this.

dipy/io/dpy.py
skoudoro: ok. previous one was wrong (5 spaces cause an indentationError) so a new commit below.

arokem:
OK. I just tested this (including doctest), both with pytables > 3 and pytables < 3. Seems to work fine. Merging.

skoudoro:
Thank you @arokem !  So I close issue #1142

arokem:
Thank you @skoudoro for your work on this!

Issue #1145, opened 2016-11-04
DKI_signal should be dki_signal in dipy.sims.voxel


Garyfallidis:
DKI_signal function should be dki_signal 
No caps in function names please. Thanks!

RafaelNH:
@Garyfallidis! Sorry for my late reply - I sort this out after the ISMRM deadline! Sounds good?

Garyfallidis:
thanks

PR #1149, opened 2016-11-13
Change DKI_signal to dki_signal

RafaelNH:
To address Issue #1145


RafaelNH:
@Garyfallidis - any comments?

arokem:
Do we need to do something for back-compatibility and deprecation

Garyfallidis:
Question. No tutorial or other test is using this function?

arokem:
Not as far as I can tell (`git grep DKI_signal`)

Garyfallidis:
Okay thanks. This is a minor change so I am going to merge...

Issue #1138, opened 2016-11-01
A few broken builds


arokem:
With errors of this kind: 

http://nipy.bic.berkeley.edu/builders/dipy-bdist32-27/builds/378/steps/shell_9/logs/stdio
http://nipy.bic.berkeley.edu/builders/dipy-bdist32-33/builds/283/steps/shell_9/logs/stdio
http://nipy.bic.berkeley.edu/builders/dipy-bdist32-34/builds/287/steps/shell_9/logs/stdio

I *think* that they originated from this PR: https://github.com/nipy/dipy/pull/1092. 

@nilgoyyou, @Garyfallidis : could you please take a look and confirm?

nilgoyyou:
I think it's related to the problem described [here](http://cython.readthedocs.io/en/latest/src/tutorial/external.html#dynamic-linking): "The libc math library is special in that it is not linked by default on some Unix-like systems, such as Linux."

I reproduced the problem in Windows.
- I tried adding the `libraries=["m"]` parameter but `m.lib` doesn't exist on Windows. Or `math.lib`, or `cmath.lib`, etc. I don't really know what to write here.
- I tried redeclaring `fmin` with `cdef extern from "math.h" nogil:` but it stays undefined. Anyway, we can [clearly see that it's defined in cython](https://github.com/cython/cython/blob/master/Cython/Includes/libc/math.pxd) so we shouldn't have to do that.

This being said, I was not able to fix it on my computer. Anyone known how?

matthew-brett:
This may be relevant : http://stackoverflow.com/questions/19234452/visual-c-express-fmin-and-fmax-identifier-not-found#19234517

```
fmin and fmax are a C99 feature. Visual Studio only implements the C89 standard, so it doesn't have that functions.
```

Maybe you could do something like [this](http://powdertoy.co.uk/Discussions/Thread/View.html?Thread=8877&PageNum=0#Message=140656) in `src/dpy_math.h`:

```
#if (defined(WIN32) || defined(_WIN64)) && !defined(__GNUC__)
#define fmin min
#endif
```

Then in the pyx file:

```
cdef extern from "dpy_math.h" nogil:
    double fmin(double x)
```

or similar.  @omarocegueda may better remember our struggles on this kind of thing back in the day.

arokem:
This is addressed in #1151

PR #1151, opened 2016-11-18
Define fmin() for Visual Studio

nilgoyyou:
Should fix issue #1138.

As @matthew-brett wrote, `fmin()` is not defined in the "old" versions of Visual Studio.
I defined it in `dpy_math.h`, used it in `vox2track.pyx` and ran `setup.py bdist_wheel` on Windows and Linux Mint to test the broken builds. Seem good.

### DIFF
--- a/dipy/tracking/vox2track.pyx
+++ b/dipy/tracking/vox2track.pyx
@@ -4,7 +4,9 @@ implemented in cython.
 """
 import cython
 
-from libc.math cimport ceil, fmin, floor, fabs, sqrt
+cdef extern from "dpy_math.h" nogil:
+    double fmin(double x, double y)
+from libc.math cimport ceil, floor, fabs, sqrt
 
 import numpy as np
 cimport numpy as cnp
--- a/src/dpy_math.h
+++ b/src/dpy_math.h
@@ -18,6 +18,10 @@ double dpy_log2(double x)
 #endif
 }
 
+#if (defined(WIN32) || defined(_WIN64)) && !defined(__GNUC__)
+#define fmin min
+#endif
+
 #define dpy_floor(x) floor((double)(x))
 
 double dpy_rint(double x)


arokem:
Thanks! I will launch a `try_branch` on one of the bots.

arokem:
http://nipy.bic.berkeley.edu/builders/dipy-bdist32-27/builds/383

arokem:
Harumph. It seems to be getting `master` instead of this branch. I'll keep trying

arokem:
I seem to be failing at this. I am quite confident that I am doing the same thing that I always do. Checkout the branch and then run: 

    try_branch.py dipy-bdist32-27

But it seems to still be getting the most recent commit on `master`. For example: http://nipy.bic.berkeley.edu/builders/dipy-bdist32-27/builds/387. I do notice that the relevant patch is somehow in there as well: http://nipy.bic.berkeley.edu/builders/dipy-bdist32-27/builds/387/steps/git/logs/patch, but it doesn't get used, and the hash points to current `master`. @matthew-brett : any ideas about what might be going on?

matthew-brett:
I think what you see is what's expected - try_branch sends the patch relative to master, and applies it.  I just looked on the machine, the code looks correct, with these patches applied.

matthew-brett:
I think I found the problem - the macro for detecting windows should be `_WIN32` not `WIN32`:

```
diff --git a/src/dpy_math.h b/src/dpy_math.h
index 3f9e227..cfadc34 100644
--- a/src/dpy_math.h
+++ b/src/dpy_math.h
@@ -18,7 +18,7 @@ double dpy_log2(double x)
 #endif
 }

-#if (defined(WIN32) || defined(_WIN64)) && !defined(__GNUC__)
+#if (defined(_WIN32) || defined(_WIN64)) && !defined(__GNUC__)
 #define fmin min
 #endif
```

nilgoyyou:
I wasn't aware there was a problem, but it seems you're right, `_WIN32` is the official way to detect Windows. I'll change it.

matthew-brett:
The problem was the error that Ariel was finding with `try_branch.py` - the compilation is still failing with Python 2.7, at least on Windows 32-bit.  It passes with the change to `_WIN32`.

arokem:
Yep. Seems to compile fine with this change: http://nipy.bic.berkeley.edu/builders/dipy-bdist32-33/builds/289

PR #1153, opened 2016-11-25
Athena mapmri 

arokem:
(a rebased version of #740 )

### DIFF
--- a/dipy/sims/voxel.py
+++ b/dipy/sims/voxel.py
@@ -135,7 +135,7 @@ def add_noise(signal, snr, S0, noise_type='rician'):
     return noise_adder[noise_type](signal, noise1, noise2)
 
 
-def sticks_and_ball(gtab, d=0.0015, S0=1., angles=[(0, 0), (90, 0)],
+def sticks_and_ball(gtab, d=0.0015, S0=100, angles=[(0, 0), (90, 0)],
                     fractions=[35, 35], snr=20):
     """ Simulate the signal for a Sticks & Ball model.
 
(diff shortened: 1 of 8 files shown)


arokem:
This is now rebased on top of master. I have also added test coverage in the `mapmri` module, so that it  is now up to 97% (on my laptop; 93% on Travis). @rutgerfick : when you have a chance, could you please take a look? Also, if you have data that generates all of the different error codes, it would be very useful to have data from these voxels, to add to the tests. @Garyfallidis : what else remains to do here, from your point of view?

Garyfallidis:
@arokem we need to try it and test it with multi-shell data. And also have another check in the code. Then merge.

arokem:
Did you try running the example (`reconst_mapmri.py`)? Seems to work fine
on my end (but I am not sure what these maps are supposed to look like...).

rutgerfick:
Hello!
I'll try to add the extra tests to the code tonight and see what are the
last tweeks to the code.
In my personal files I also have the code to add the Propagator Anisotropy
metric, but I suggest we leave that for another pull request since it
actually requires a significant extension.

Rutger.

arokem:
yes - best to do that as a follow-up after we merge this one. It's already a bit of a beast.

Garyfallidis:
Agreed. I am super excited about this PR. Trying in it now with some HCP data. What is crucial is find datasets that really show the capability of the method.

dipy/sims/voxel.py
```diff
@@ -188,7 +190,130 @@ def sticks_and_ball(gtab, d=0.0015, S0=1., angles=[(0, 0), (90, 0)],
...
+        signal attenuation
+    """
+    return np.exp(-(2 * np.pi * q) ** 2 * tau * D)
+
+
+def cylinders_and_ball_soderman(gtab, tau, radii=[5e-3, 5e-3], D=0.7e-3,
+                                S0=1., angles=[(0, 0), (90, 0)],
+                                fractions=[35, 35], snr=20):
+    r""" Calculates the three-dimensional signal attenuation E(q) originating
+    from within a cylinder of radius R using the Soderman approximation [1]_.
```
Garyfallidis: The underscore is fine here.

dipy/sims/voxel.py
```diff
@@ -188,7 +190,130 @@ def sticks_and_ball(gtab, d=0.0015, S0=1., angles=[(0, 0), (90, 0)],
...
+        noise is added.
+
+    Returns
+    -------
+    E : array, shape (N,)
+        signal attenuation
+
+    References
+    ----------
+    .. [1]_ Söderman, Olle, and Bengt Jönsson. "Restricted diffusion in
```
Garyfallidis: But not needed here. I believe you are doing the same in many areas. I am not sure if it valid and perhaps the tutorial will not render correctly.

Garyfallidis:
I am currently getting this error
```
/home/elef/Devel/dipy/dipy/reconst/mapmri.py:357: UserWarning: The MAPMRI positivity constraint depends on CVXOPT (http://cvxopt.org/). CVXOPT is licensed under the GPL (see: http://cvxopt.org/copyright.html) and you may be subject to this license when using the positivity constraint.
  warn(w_s)
  File "reconst_mapmri.py", line 143, in <module>
    mapfit_positivity_aniso = map_model_positivity_aniso.fit(data_small)
  File "/home/elef/Devel/dipy/dipy/reconst/multi_voxel.py", line 33, in new_fit
    fit_array[ijk] = single_voxel_fit(self, data[ijk])
  File "/home/elef/Devel/dipy/dipy/reconst/mapmri.py", line 395, in fit
    cvxopt.solvers.options['show_progress'] = False
AttributeError: 'module' object has no attribute 'solvers'
```
Do I need to upgrade cvxopt?

arokem:
Looks like it didn't properly import cvxopt.

What happens when you do `python -c"import cvxopt;
cvxopt.solvers.options['show_progress'] = False"`?

(I'm betting on an ImportError)

Garyfallidis:
No it seems the code is not compatible with the python-cvxopt version available in Ubuntu 16.04. When I downloaded  cvxopt with pip and compiled it then it did work.

arokem:
Out of curiosity: what are the different versions?

Garyfallidis:
1.1.4 (ubuntu 16.04) vs 1.1.9 (pip)

dipy/sims/voxel.py
rutgerfick: Hi Elef, so to clarify, I need to use the underscore only in the 'text' and not when i define the citation?

Garyfallidis: correct

rutgerfick: Just a question, but why do you automatically put S0 to an arbitrary value that is not 1?
I saw it sometimes as 150 or 100 for default. Why not by default set it to 1 and keep the signal attenuation convention instead of unnormalized signal?

rutgerfick:
As for the specific data for the errorcodes. In practice it generally happens when background or skull voxels are included in the fitting, causing the positivity constraint to not converge or resulting in some non-invertible observation matrix. Adding a test for this would probably work by trying to fit random noise and it will most likely break down.

rutgerfick:
As far as I can tell the implementation looks fine, as well as the tests.
I'm not sure how I can go into this new pull request and make changes to the underscore issue Elef brought up. Would it be something like this? http://stackoverflow.com/questions/14947789/github-clone-from-pull-request

Garyfallidis:
All you need to do is make a PR on top of Ariel's PR. This PR. You do not need to clone again.

arokem:
I made the underscore fixes on my branch.

arokem:
@rutgerfick : regarding data for the different error codes: it would be good to have specific examples of these failures. Do you have such data? Ideally, we would put data from a single voxel that corresponds to each error straight into our tests (like we do here, for example: https://github.com/nipy/dipy/blob/master/dipy/reconst/tests/test_ivim.py#L78) and verify that we always get the corresponding error to each of the error codes.

arokem:
@Garyfallidis : regarding CVXOPT versions: it looks like the `solvers` module was introduced in version 1.1.7 of the software (released in 2014). How should we handle this version requirement? I guess that when a user calls these functions, and cvxopt is triggered, we can check for that. For the time being, I have edited the documentation to reflect this minimal requirement.

arokem:
I've also added a test for the CVXOPT version. Unfortunately, it's rather brittle. This is because as far as I can tell, there is no way to infer the version of CVXOPT, except to check whether the software you have installed has the `solvers` sub-module. The software doesn't have anything like a version string stored anywhere that I could find it.

arokem:
@rutgerfick : do you have data you could provide (individual voxels) that raise the linear algebra errors corresponding to the error codes? @Garyfallidis : where does this stand? Is this otherwise ready to be merged from your point of view?

rutgerfick:
Hi Ariel. I don't have the voxels for the specific error codes right now,
and I'm afraid I don't have any time to spare on this project until I
submit my thesis this month (First draft is supposed to be done this
Friday...).

At this point, the errorcodes serve as a way to not make the fitting
procedure stop when it encounters an invalid voxel (nearly always the
skull).
For the rest, I think this piece of code is ready to be merged.
I would suggest it's a good idea to just merge it like this and we can make
a new pull request in a few weeks when my thesis is done.
I will then implement the last updates (propagator anisotropy) and any
other comments you might have or may come up.

Does that sound good to you?

arokem:
Sounds perfect. Good luck with your thesis!

arokem:
@Garyfallidis : any luck finding more people/data to try this out?

arokem:
Rebased. @Garyfallidis : any more thoughts here?

Garyfallidis:
I am having a meeting with a postdoc on Thursday morning to look more deeply into the scientific parts of this work. But codewise this is ready and working with different datasets!!! We should try to find some higher resolution datasets to showcase the power of this method. Thanks @rutgerfick and all for contributing. Merging!

PR #1155, opened 2016-11-25
csa and csd reconstruction workflow

matthieudumont:


dipy/io/peaks.py
```diff
@@ -0,0 +1,99 @@
...
+def save_peaks(fname, pam, compressed=True):
+    """ Save NPZ file with all important attributes of object PeaksAndMetrics
+    """
+
+    if compressed:
+        save_func = np.savez_compressed
+    else:
+        save_func = np.savez
+
+    save_func(fname,
```
Garyfallidis: Hi @matthieudumont I had one user strongly suggesting to use hdf5 to save the peaks rather than npy or npz. This will allow other non-pythonic projects to read our results. We already depend on pytables that allows to save to hdf5. So, it shouldn't be difficult to make the change. If you agree I can make a PR on top of yours with the required changes. Or if you know how to do it and agree with the suggestion you can go ahead and make the change yourself. See also here https://github.com/nipy/dipy/pull/1147/files
Let me know what you think.

matthieudumont: I think it is a great idea!

Ill let you make the PR and concentrate on the ones I have already running.

Thanks!

Garyfallidis: okay sure.

Garyfallidis:
Hi @matthieudumont, now that we have merged the mask workflow we have some conflicts here. Can you please resolve?

Garyfallidis:
Resolved here #1241

PR #1162, opened 2016-12-10
Return S0 value for DTI fits

etpeterson:
The goal of this PR is to optionally return the S0 (b=0) value for DTI fits without negatively impacting performance or memory unless the S0 is requested. It captures the fitted S0 values generated in the DTI fitting routines and allows the user to access them.

It's similar in intent to PR [1036](https://github.com/nipy/dipy/pull/1036) but rather than calculating the S0 values from the fits, it captures the values directly in each fit method.

My local tests look like the speed is only slightly slower but keeping the S0 values does increase memory usage because it's another image being saved. I'm timing it using time for the whole fit and finding the size by pickling the fit object so if there's a better way to do that let me know.

I think it's missing a few tests and probably some documentation but I think the major pieces are there.


RafaelNH:
Hi @etpeterson! I like what you did! We just have to understand why tests "dipy.viz.tests.test_fvtk_widgets.test_button_and_slider_widgets" and "dipy.viz.tests.test_fvtk_widgets.test_text_widget" are failing at 4387.8.

And perhaps thinking about a better input name for input_S0_hat

etpeterson:
This test failure is strange. The exact same thing worked in my fork. I'm tempted to just try the tests again because it looks like it failed trying to fetch some data in code that looks completely unrelated to what I changed.

I'm also open to suggestions for changing return_S0_hat to something else more descriptive, or even just dropping the _hat at the end.

arokem:
I just restarted that test on Travis.

arokem:
Yeah. Looks like it was a temporary glitch. Travis is now all-green.

arokem:
I would like to also review this PR, and will probably have some time to do that later this week.

etpeterson:
Sounds good all around. I probably won't have much time this week to work on it anyway.

dipy/reconst/dti.py
```diff
@@ -858,6 +877,16 @@ def quadratic_form(self):
         # np.einsum('...ij,...j,...kj->...ik', evecs, evals, evecs)
         return vec_val_vect(self.evecs, self.evals)
 
+    @property
+    def S0(self):
```
RafaelNH: Why do we need to have a self.S0 and a self.S0_hat?
Can we just have S0 (i.e. define S0_hat as S0 in line 834 and delete this property)?

dipy/reconst/dti.py
```diff
@@ -858,6 +877,16 @@ def quadratic_form(self):
...
+        """
+        Returns the S0_hat value of the fit
+        """
+        if self.S0_hat is None:
+            return None
+        else:
+            return self.S0_hat.reshape(self.shape)
```
RafaelNH: To have only S0, you need to reshape it also in line 834.

dipy/reconst/dti.py
```diff
@@ -1242,22 +1275,38 @@ def wrapped_fit_tensor(design_matrix, data, step=step,
             size = np.prod(shape)
             step = int(step) or size
             if step >= size:
-                return fit_tensor(design_matrix, data, *args, **kwargs)
+                if return_S0_hat:
+                    return fit_tensor(design_matrix, data,
+                                      return_S0_hat=return_S0_hat, *args, **kwargs)
```
RafaelNH: PEP8: too long line

dipy/reconst/dti.py
```diff
@@ -1242,22 +1275,38 @@ def wrapped_fit_tensor(design_matrix, data, step=step,
...
+                    fit_output = fit_tensor(design_matrix,
                                                    data[i:i + step],
+                                                   return_S0_hat=return_S0_hat,
                                                    *args, **kwargs)
-            return dtiparams.reshape(shape + (12, ))
+                    dtiparams[i:i + step] = fit_output[0]
+                    S0params[i:i + step] = fit_output[1]
+                else:
+                    dtiparams[i:i + step] = fit_tensor(design_matrix,
+                                                   data[i:i + step], *args, **kwargs)
```
RafaelNH: Also, ensure that you don't have a line with more than 79 characters

dipy/reconst/dti.py
```diff
@@ -1242,22 +1275,38 @@ def wrapped_fit_tensor(design_matrix, data, step=step,
...
+            if return_S0_hat:
+                return (dtiparams.reshape(shape + (12, )), S0params.reshape(shape + (1, )))
```
RafaelNH: PEP8 line too long

dipy/reconst/dti.py
```diff
@@ -1374,10 +1427,12 @@ def ols_fit_tensor(design_matrix, data):
     """
     tol = 1e-6
     data = np.asarray(data)
-    return eig_from_lo_tri(
-        np.einsum('...ij,...j', np.linalg.pinv(design_matrix), np.log(data)),
-        min_diffusivity=tol / -design_matrix.min(),
-    )
+    fit_result = np.einsum('...ij,...j', np.linalg.pinv(design_matrix), np.log(data))
+    if return_S0_hat:
+        return (eig_from_lo_tri(fit_result, min_diffusivity=tol / -design_matrix.min()),
```
RafaelNH: too long line

dipy/reconst/dti.py
```diff
@@ -1374,10 +1427,12 @@ def ols_fit_tensor(design_matrix, data):
...
+                np.exp(-fit_result[:, -1]))
+    else:
+        return eig_from_lo_tri(fit_result, min_diffusivity=tol / -design_matrix.min())
```
RafaelNH: PEP8 too long line

dipy/reconst/dti.py
```diff
@@ -1624,19 +1681,27 @@ def nlls_fit_tensor(design_matrix, data, weighting=None,
...
-    return dti_params
+    if return_S0_hat:
+        S0_hat.shape = data.shape[:-1] + (1,)
+        return (dti_params, S0_hat)
+    else:
+        return dti_params
 
 
-def restore_fit_tensor(design_matrix, data, sigma=None, jac=True):
+def restore_fit_tensor(design_matrix, data, sigma=None, jac=True, return_S0_hat=False,):
```
RafaelNH: PEP8

dipy/reconst/tests/test_dti.py
```diff
@@ -139,12 +139,13 @@ def test_tensor_model():
     # Test fitting with different methods:
     for fit_method in ['OLS', 'WLS', 'NLLS']:
         tensor_model = dti.TensorModel(gtab,
-                                       fit_method=fit_method)
+                                       fit_method=fit_method, return_S0_hat=True)
```
RafaelNH: PEP8

dipy/reconst/dti.py
```diff
@@ -1162,6 +1191,10 @@ def predict(self, gtab, S0=1., step=None):
         which a signal is to be predicted and $b$ is the b value provided in
         the GradientTable input for that direction
         """
+        if S0 is None:
+            S0 = self.S0  # note it's S0, not S0_hat because the shapes aren't the same
```
RafaelNH: If you remove S0_hat, you will have to remove this comment.

etpeterson:
Thanks @RafaelNH, I fixed the PEP-8 issues.

About S0_hat vs S0, in the original effort to add S0 output ([#1036](https://github.com/nipy/dipy/pull/1036)) it was named S0_hat, so that's what I adopted as well. Perhaps to not confuse the optional input S0 with the output S0? I'd prefer S0 because I think it's more intuitive but I'm open to suggestions.

arokem:
This looks generally good to me. Regarding `S0_hat`: the idea I have in mind is that there are two quantities that we want to distinguish: one is the intercept parameter from the model. That's what I called `S0_hat`. The other is the measurement that was done with no diffusion-weighting. That's what I called `S0`. I would still like to distinguish these. That is, I don't think that a model fit should have an `S0` attribute, only an `S0_hat` attribute. `S0` by itself can be an input to a prediction, though, because sometimes you want to enforce the measured value, instead of the estimated intercept. Does that make sense?

etpeterson:
That sounds reasonable to me. I made the changes so S0_hat is the model fit. I think it's working and doing everything it should.

I suspect I'm missing some test cases and documentation so I'll try to check that when I can but comments are welcome before then.

etpeterson:
I updated the tests and documentation. Let me know what you think. I'm also interested if you have thoughts about speed and memory testing beyond the basic script I'm using locally.

dipy/reconst/dti.py
```diff
@@ -1242,22 +1283,41 @@ def wrapped_fit_tensor(design_matrix, data, step=step,
             size = np.prod(shape)
             step = int(step) or size
             if step >= size:
-                return fit_tensor(design_matrix, data, *args, **kwargs)
+                if return_S0_hat:
```
arokem: Here as well, consider whether the logic could be simplified by passing along the default value in the case where `return_S0_hat` is set to `False`.

dipy/reconst/tests/test_dti.py
```diff
@@ -631,6 +657,11 @@ def test_restore():
     tensor_model = dti.TensorModel(gtab, fit_method='restore', sigma=0.0001)
     tensor_model.fit(Y.copy())
 
+    # Test return_S0_hat
+    tensor_model = dti.TensorModel(gtab, fit_method='restore', sigma=0.0001, return_S0_hat=True)
```
arokem: PEP8: Line's too long

dipy/reconst/dti.py
```diff
@@ -777,17 +782,35 @@ def fit(self, data, mask=None):
             min_signal = self.min_signal
 
         data_in_mask = np.maximum(data_in_mask, min_signal)
-        params_in_mask = self.fit_method(self.design_matrix, data_in_mask,
-                                         *self.args, **self.kwargs)
+
+        if self.return_S0_hat:
```
etpeterson: I agree on the input side that's fine but the output of fit_method is a little more tricky because they either return 1 or 2 arguments. So I think the alternative below still requires an if statement but is perhaps more clean?

```python
params_in_mask = self.fit_method(
                self.design_matrix,
                data_in_mask,
                return_S0_hat=self.return_S0_hat,
                *self.args,
                **self.kwargs)
if self.return_S0_hat:
    params_in_mask, model_S0 = params_in_mask
```

dipy/reconst/dti.py
```diff
@@ -777,17 +782,35 @@ def fit(self, data, mask=None):
...
+                **self.kwargs)
+        else:
+            params_in_mask = self.fit_method(self.design_matrix,
+                                             data_in_mask,
+                                             *self.args, **self.kwargs)
 
         if mask is None:
             out_shape = data.shape[:-1] + (-1, )
             dti_params = params_in_mask.reshape(out_shape)
+            if self.return_S0_hat:
```
etpeterson: Could you explain how to simplify the if statements in the mask logic? I don't see how that would work. If `model_S0` is `None` then it would fail trying to reshape or it would return a large `None` array. 

I do agree that the ifs with `TensorFit` can be removed though to leave `return TensorFit(self, dti_params, model_S0=S0_params)` as long as `S0_params` is `None`.

dipy/reconst/dti.py
```diff
@@ -808,21 +831,32 @@ def predict(self, dti_params, S0=1.):
...
         model_params = self.model_params
+        model_S0 = self.model_S0
         N = model_params.ndim
         if type(index) is not tuple:
             index = (index,)
         elif len(index) >= model_params.ndim:
             raise IndexError("IndexError: invalid index")
         index = index + (slice(None),) * (N - len(index))
-        return type(self)(self.model, model_params[index])
+        if model_S0 is None:
```
etpeterson: But `model_S0=model_S0[index_S0]` would fail if `model_S0` is `None`. Or are you suggesting just pass it along if it's `None` anyway on line 848?

dipy/reconst/dti.py
etpeterson: You're right, this can be simplified. Though not anywhere else in this function as far as I can see.

etpeterson:
I'll give profiling a shot. I'll try to do this PR with and without returning S0 and master. I'll also see if I can track memory in a meaningful way.

Regarding testing different structured inputs, is there a dataset that's already in dipy you recommend? Do you think it would be good to test on data without any b=0 at all?

etpeterson:
Here are some times, timed with the version that's up right now. It's not taking much more time at all, but the memory size does increase because we need to hold on to the S0 data rather than throw it away.

Master for comparison because it has no S0 logic at all:
timeit repetition time =  2.9841374158859253 s
structure size using pickle = 1.5746278100000002 GB

Current version, S0 OFF:
timeit repetition time = 3.0333045959472655 s
structure size using pickle = 1.574627854 GB

Current version, S0 ON:
timeit repetition time = 3.0728065013885497 s
structure size using pickle = 1.6998900060000002 GB

dipy/reconst/dti.py
arokem: Yes - that's cleaner. In general, I prefer it when there is only one call to the function. That way, if the API changes for reasons unrelated to `S0_hat`, we only have to change the call in one place, instead of two places.

dipy/reconst/dti.py
arokem: Yes - I would make `S0_params` an optional input to `fit` with a default of None. Does that make sense? But you're right that I don't see a way to simplify more.

dipy/reconst/dti.py
arokem: Something like: 

```
if model_S0 is not None: 
   model_S0 = model_S0[index[:-1]] 

return type(self)(self.model, model_params[index], model_S0=model_S0)

```

etpeterson:
@arokem your comments sound good to me so I made and pushed the changes. Also, do you want more details about timing and profiling or is what I have above using timeit satisfactory?

arokem:
Code looks good to me. Regarding profiling: what data did you use to run the profiling? Assuming this is the difference on reasonably large whole-brain data, it seems to be that there wouldn't be harm in adding this code -- the difference seems minor. Any thoughts from anyone else?

etpeterson:
Good to hear.

I used some 3T human brain data I have here that has multiple b-values so the S0 fitting is meaningful. It's 256x256x64 and 3 directions with 7 b-values. I did mask it but it's still fitting 905,768 voxels.

etpeterson:
A somewhat related question: is Dipy participating in Google Summer of Code this year? I'm thinking about this PR and integrating it into IVIM and other IVIM improvements that could work as a project for GSoC. I think the first deadline is Feb 9th so it's already looming on the horizon.

arokem:
OK - I am happy with this. If no one else has any comments here, I will merge this in a couple of days.

Re: GSoC -- I will not be able to be involved this year. I am going to be on family leave during the summer 🍼  😄 !

etpeterson:
Congrats on the 👶 ! That's a legit reason to skip it. We're expecting our first in February so I was figuring I'd at least be partly back in action by April in time for GSoC. But maybe less is more this year 😄

RafaelNH:
I've just did a last look to the code - for me this is ready to go! Thanks @etpeterson for the really nice work!

PR #1163, opened 2016-12-11
WIP: Combined contour function with slicer to use affine

kesshijordan:
I took the code from the contour function from fvtk.py (which doesn't use an affine) and combined it with code from the slicer in actor.py to make a function contour_actor in actor.py that uses an affine to place contours. I'm using this to display a roi with background image (slicer) and streamlines. I haven't written tests, yet.

### DIFF
--- a/dipy/viz/actor.py
+++ b/dipy/viz/actor.py
@@ -201,6 +201,139 @@ def slicer(data, affine=None, value_range=None, opacity=1.,
     return image_actor
 
 
+def contour_actor(data, affine=None, levels=[50],
+            colors=[np.array([1.0, 0.0, 0.0])], opacities=[1]):
+    """Take a volume and draw surface contours for any any number of
+    thresholds (levels) where every contour has its own color and opacity
+
+    Parameters
+    ----------
+    data : array, shape (X, Y, Z) or (X, Y, Z, 3)
+        A grayscale or rgb 4D volume as a numpy array.
+    affine : array, shape (4, 4)
+        Grid to space (usually RAS 1mm) transformation matrix. Default is None.
+        If None then the identity matrix is used.
+    levels : array_like
+        Sequence of thresholds for the contours taken from image values needs
+        to be same datatype as `vol`.
+    colors : (N, 3) ndarray
+        RGB values in [0,1].
+    opacities : array_like
+        Opacities of contours.
+
+    Returns
+    -------
+    contour_assembly : vtkAssembly
+        An object that is capable of displaying an roi as contours in space
+        coordinates as calculated by the affine parameter.
+
+    """
+    if data.ndim != 3:
+        if data.ndim == 4:
+            if data.shape[3] != 3:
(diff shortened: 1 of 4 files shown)


dipy/viz/actor.py
```diff
@@ -201,6 +201,139 @@ def copy(self):
     return image_actor
 
 
+def contour_actor(data, affine=None, levels=[50],
+            colors=[np.array([1.0, 0.0, 0.0])], opacities=[1]):
```
arokem: PEP8: indentation level here doesn't match the line above (`colors` needs to be flush with `data`)

dipy/viz/actor.py
```diff
@@ -201,6 +201,139 @@ def copy(self):
...
+        skin_actor.GetProperty().SetOpacity(opacities[i])
+
+        skin_actor.GetProperty().SetColor(colors[i][0], colors[i][1], colors[i][2])
+        skin_assembly.AddPart(skin_actor)
+        del skin_actor
+        del skinMapper
+        del skinExtractor
+
+    return skin_assembly
+
```
arokem: One more line of white-space between functions please.

dipy/viz/actor.py
kesshijordan: Thanks, ariel! I just added PEP8 comments to my pycharm so this should happen less. It doesn't like the default values I set; is this fine for dipy or is there another way I should do this? (It hilights levels, colors, and opacities with this comment: "This inspection detects when a mutable value as list or dictionary is detected in a default value for an argument. Default argument values are evaluated only once at function definition time, which means that modifying the default value of the argument will affect all subsequent calls of the function. ") Thanks! -Kesshi

kesshijordan: (Fixed this)

dipy/viz/actor.py
kesshijordan: Fixed this

kesshijordan:
Thanks, Ariel!

Yes; you can add it as an actor. I used the same code as was already implemented in the fvtk.py contour function, but I combined it with the slicer code so that it takes an affine (when I used the contour function originally, it rendered successfully, but was placed in the wrong part of the brain). The previous version took an affine, but didn't use it. See, below, how it's used for a binary roi.

-Kesshi

roidata = roidata.astype('uint8')
roi_actor = actor.contour_actor(roidata, levels=[1], affine=affine)
fvtk.add(ren,roi_actor)

arokem:
Hey @kesshijordan - did you mean to push the bootstrap_dg stuff into this same PR?

kesshijordan:
I just messed this up with git and merged with something that isn't ready.

arokem:
(it should be a separate PR, in my opinion)

arokem:
Gotcha. Let me know if you need any help rebasing this out of here.

kesshijordan:
Yes; I'm trying to fix it. Didn't mean to merge that.

kesshijordan:
Thanks, @arokem. I'm just going to make it a new PR and close this one.

PR #1164, opened 2016-12-16
Parallell quantize_evecs + PEP8

arokem:
This supersedes #1097, fixing up a few PEP8 issues that existed in that PR.


arokem:
Anyone want to take another look here?

arokem:
Any thoughts from anyone?

dipy/reconst/dti.py
```diff
@@ -1905,7 +1914,65 @@ def design_matrix(gtab, dtype=None):
...
+        data_file_name = path.join(tmpdir, 'data.npy')
+        np.save(data_file_name, data)
+
+        pool = Pool(nbr_processes)
+
+        peaks_res = pool.map(_quantize_evecs_parallel_sub,
+                             zip(repeat(data_file_name),
+                                 indices,
+                                 repeat(odf_vertices),
+                                 repeat(v)))
```
skoudoro: Instead of using temporary file, What do you think about using the technique explained [here](http://thousandfold.net/cz/2014/05/01/sharing-numpy-arrays-between-processes-using-multiprocessing-and-ctypes/).

dipy/reconst/dti.py
```diff
@@ -1905,7 +1914,65 @@ def design_matrix(gtab, dtype=None):
     return -B
 
 
-def quantize_evecs(evecs, odf_vertices=None):
+def _quantize_evecs_parallel(data, odf_vertices, v, nbr_processes):
+    if nbr_processes is -1:
+        try:
+            nbr_processes = cpu_count()
+        except NotImplementedError:
+            warn("Cannot determine number of cpus. \
```
arokem: Thanks for noticing. I fixed this, and extended this based on http://stackoverflow.com/a/13544896/3532933, but I am not sure how to test all this, because it seems that we would end up here only in some rare circumstances.

dipy/reconst/dti.py
```diff
@@ -1905,7 +1914,65 @@ def design_matrix(gtab, dtype=None):
...
+def _quantize_evecs_parallel_sub(args):
+    data_file_name = args[0]
+    (start_pos, end_pos) = args[1]
+    odf_vertices = args[2]
+    v = args[3]
+
+    data = np.load(data_file_name, mmap_mode='r')[start_pos:end_pos]
+
+    return quantize_evecs(data, odf_vertices, v, nbr_processes=1)
+
```
arokem: Good catch.

dipy/reconst/dti.py
```diff
@@ -1914,18 +1981,33 @@ def quantize_evecs(evecs, odf_vertices=None):
...
+    if len(evecs.shape) == 2:
+        IN = np.array([np.argmin(np.dot(odf_vertices, m)) for m in evecs])
+    else:
+        max_evecs = evecs[..., :, v]
+        tup = max_evecs.shape[:-1]
+        mec = max_evecs.reshape(np.prod(np.array(tup)), 3)
+        IN = np.array([np.argmin(np.dot(odf_vertices, m)) for m in mec])
+        IN = IN.reshape(tup)
+
     return IN
```
arokem: Renamed this to `closest`

dipy/reconst/dti.py
```diff
@@ -1993,7 +2002,65 @@ def design_matrix(gtab, dtype=None):
     return -B
 
 
-def quantize_evecs(evecs, odf_vertices=None):
+def _quantize_evecs_parallel(data, odf_vertices, v, nbr_processes):
+    if nbr_processes is -1:
```
matthew-brett: Should be `if nbr_processes == -1:` - although CPython does keep an interned list of singleton small integers, so this check (equivalent to `id(-1) == id(-1)`) will generally work, it's an implementation detail - see http://stackoverflow.com/questions/2988017/string-comparison-in-python-is-vs#2988117 for discussion.

dipy/reconst/dti.py
arokem: Still thinking about this one...

arokem:
The Travis test failures are interesting. Seem completely unrelated :-/

arokem:
I am getting these errors intermittently, which only makes them harder to understand...

dipy/reconst/dti.py
```diff
@@ -2038,6 +2023,132 @@ def eig_from_lo_tri(data, min_diffusivity=0):
...
+            nbr_processes = cpu_count()
+        except NotImplementedError:
+            try:
+                import psutil
+                psutil.cpu_count()
+            except:
+                warnings.warn("Cannot determine number of cpus.",
+                              "returns quantize_evecs(...,",
+                              "nbr_processes=1).")
+                return quantize_evecs(data, odf_vertices, v, nbr_processes=1)
```
arokem: Good catch! This would have to be tested on the Windows build-bots

arokem:
I have not been able to figure out the in-memory multiprocessing shared array. I keep getting what seems like randomly-generated garbage that appears in the array. I am going to back off and go back to the in-file solution here. Then, if someone else (@skoudoro ?) wants to take that on again, we can do this on a subsequent PR.

arokem:
Alright. This now passes the tests on my laptop. Could you please take another look, @skoudoro ?

skoudoro:
The shared array is added to my todolist. I want to understand what is going on with this memory problem. 

Anyway, everything is ok from my side too with the in-file solution, I think you can merge it.

arokem:
Once #1223 is merged it will be easier to see what lines are not covered by the tests (https://github.com/nipy/dipy/pull/1223/commits/49fa4e533fca5ca008b137243d7cd0c36f16f06a)

skoudoro:
What is the status of this PR @arokem?

arokem:
It's still broken :-/

arokem:
OK. Sorry that it's taken me so long to get back to this one. I think that it might be good to go now!

arokem:
But let's wait and see what the CI says.

dipy/reconst/dti.py
arokem: OK - I implemented this in https://github.com/nipy/dipy/pull/1164/commits/bf858f9cd018e447f328e85b20cad0135f891a2b

dipy/reconst/dti.py
```diff
@@ -2050,6 +2036,117 @@ def eig_from_lo_tri(data, min_diffusivity=0):
...
+        If 0, uses the largest eigenvalue.
+    nbr_processes : int
+        Number of processes to use. If > 1, divides data into nbr_processes for
+        multiprocessing. If -1, uses as many processes as possible.
+
+    Returns
+    -------
+    closest : ndarray
+    """
+    if nbr_processes != 1:
```
skoudoro: This will fail if `nb_processes=0` because `nbr_chunks` will be equal to 0. This is a strange case so should we manage it as 1 (not set) or -1(all cpu's).
Can you add a test for this case?

dipy/reconst/dti.py
```diff
@@ -2050,6 +2036,117 @@ def eig_from_lo_tri(data, min_diffusivity=0):
     return dti_params.reshape(data.shape[:-1] + (12, ))
 
 
+def quantize_evecs(evecs, odf_vertices=None, v=0, nbr_processes=1):
```
skoudoro: Can you replace the variable name `v` by something else like `vect` or `v_pos`?

dipy/reconst/dti.py
```diff
@@ -2050,6 +2036,117 @@ def eig_from_lo_tri(data, min_diffusivity=0):
...
+
+def _quantize_evecs_parallel_sub(args):
+    (start_pos, end_pos) = args[1]
+    odf_vertices = args[2]
+    v = args[3]
+    evecs = args[0][start_pos:end_pos]
+    return quantize_evecs(evecs, odf_vertices, v, nbr_processes=1)
+
+
+def _quantize_evecs_parallel(evecs, odf_vertices, v, nbr_processes):
```
skoudoro: same comments here for the variable name `v`

arokem:
OK. Finally got back around to address these comments.

skoudoro:
superseded by #3605, closing

Issue #1131, opened 2016-10-14
new non_local_means module seems to disregard 3D array of noise standard deviation


samuelstjean:
So @mdesco seems to have some problem with the denoising module (we are still investigating what's happening ), but I noticed so far the new nlmean replacement seems to discard the 3D information. This code

~~~python
    if arr.ndim == 3:
        sigma = np.ones(arr.shape, dtype=np.float64) * sigma
        return np.array(nlmeans_block(
            np.double(arr),
            mask,
            patch_radius,
            block_radius,
            sigma[0, 0, 0],
            np.int(rician))).astype(arr.dtype)
    elif arr.ndim == 4:
        denoised_arr = np.zeros_like(arr)

        if isinstance(sigma, np.ndarray) and sigma.ndim == 3:
            sigma = (np.ones(arr.shape, dtype=np.float64) *
                     sigma[..., np.newaxis])
        else:
            sigma = np.ones(arr.shape, dtype=np.float64) * sigma

        for i in range(arr.shape[-1]):
            denoised_arr[..., i] = np.array(nlmeans_block(np.double(
                arr[..., i]), mask, patch_radius, block_radius,
                sigma[0, 0, 0, 0], np.int(rician))).astype(arr.dtype)

        return denoised_arr
~~~

will strip out the 3D array passed and extract a single value from voxel (0,0,0,0) in case a 3D array is passed. If, say, the estimation was done on a datasets with a masked background, the passed down value would thus be 0, which disregard the original 3D volume of standard deviation passed to the function.

arokem:
Doesn't this test address this use-case? https://github.com/nipy/dipy/blob/master/dipy/denoise/tests/test_non_local_means.py#L17-L28

samuelstjean:
No, actually this line in 4D

https://github.com/nipy/dipy/blob/master/dipy/denoise/non_local_means.py#L75

and this line in 3D

https://github.com/nipy/dipy/blob/master/dipy/denoise/non_local_means.py#L61

trash out completely the 3D array and pick the voxel 0,0,0,0 no matter what it seems, which is pretty bad if it really does this indiscriminately. It would mean that for any 3D array, it wholly gets disregarded, and if you have a masked volume, it would pass variance = 0 everywhere and segfault.

It would be better to

A. Discard the 3D behavior and return to a pure original nlmeans like this implementation is (actually, I think the code IS the original 2008 nlmeans, and it contain some more hardcoded values to this regard)
[...]

PR #1166, opened 2016-12-17
RF: Use the average sigma in the mask.

arokem:
Instead of the value of sigma in the corner of the image.

Addresses #1131

### DIFF
--- a/dipy/denoise/tests/test_non_local_means.py
+++ b/dipy/denoise/tests/test_non_local_means.py
@@ -34,6 +34,16 @@ def test_nlmeans_random_noise():
         assert_equal(np.round(S0nb[mask].mean()), 100)
 
 
+def test_scalar_sigma():
+    S0 = 100 + np.zeros((20, 20, 20))
+    noise = 2 * np.random.standard_normal((20, 20, 20))
+    S0 += noise
+    S0[:10, :10, :10] = 300 + noise[:10, :10, :10]
+
+    S0n = assert_raises(
+        ValueError, non_local_means, S0, sigma=noise, rician=False)
+
+
 def test_nlmeans_boundary():
     # nlmeans preserves boundaries
 
--- a/dipy/denoise/tests/test_non_local_means.py
+++ b/dipy/denoise/tests/test_non_local_means.py
@@ -20,8 +20,7 @@ def test_nlmeans_random_noise():
     masker = np.zeros(S0.shape[:3]).astype(bool)
     masker[8:15, 8:15, 8:15] = 1
     for mask in [None, masker]:
-        S0nb = non_local_means(S0, sigma=np.ones((22, 23, 30)) *
-                               np.std(S0), rician=False, mask=mask)
+        S0nb = non_local_means(S0, sigma=np.std(S0), rician=False, mask=mask)
 
         assert_(S0nb[mask].min() > S0[mask].min())
         assert_(S0nb[mask].max() < S0[mask].max())
@@ -46,7 +45,7 @@ def test_nlmeans_boundary():
 
     S0[:10, :10, :10] = 300 + noise[:10, :10, :10]
 
-    S0n = non_local_means(S0, sigma=np.ones((20, 20, 20)) * np.std(noise),
+    S0n = non_local_means(S0, sigma=np.std(noise),
                           rician=False)
 
     print(S0[9, 9, 9])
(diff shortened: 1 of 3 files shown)


arokem:
Urgh. What's the right way to handle `nanmean`? It's not in `scipy.stats` in newer versions of scipy, and it's not in `numpy` in older versions of numpy. Do I really have to check the version of both numpy and scipy to be sure that I can compute a nanmean?

MarcCote:
Maybe compute it explicitly instead of relying on nanmean? See http://stackoverflow.com/questions/5480694/numpy-calculate-averages-with-nans-removed for some variants.

NB: Maybe a bit slower compared to np.nanmean, though.

arokem:
Thanks! That looks like a good solution to me. We can check whether numpy has a nanmean and use that one if possible, and otherwise use our own. What do you think?

MarcCote:
I liked it. :+1:

arokem:
What do you think about all the rest of this? Does it properly address #1131 in your opinion?

arokem:
Any more thoughts here? Seems like a bug we need to address before we can make a release.

arokem:
Anyone want to take a look at this?

MarcCote:
This looks good to me. @samuelstjean last chance to have a look at this PR. Does it address your issue #1131 correctly?

samuelstjean:
It kinds of circumvent the problem. It is a less brute approach than picking the top left corner voxel, but it does cause problem if you have a large variation of the noise profile in your image. In that case, I would prefer if it just threw me an error that 3D arrays are not supported, and let me choose how to deal with it. 

For example, say I collected a noise measurement maps, extract local standard deviation from that, and feed it to the function. It will pick out the mean, and if the map is first masked (say by the scanner), then the median would probably be more appropriate to deal with those 3/4 of zeros values. As of now, I'll probably get a useless value, where the previous nlmeans version would happily work with my 3D map at each spatial location. Since that is now the case anymore, this would all happen silently and give me a less optimal result than I expected (true story, ask JC).

arokem:
I am back to this. I agree with @samuelstjean. It doesn't make sense to accept 3D sigma inputs to this function, because it will create a scalar under the hood using less than obvious rules. If you want to extract your sigma from a 3D volume, please do so before you feed it to this function. Simplified accordingly.

arokem:
Now rebased on master

MarcCote:
@arokem @samuelstjean I'll be merging this PR on Monday.

samuelstjean:
+ the nanmean function serves no purpose
+ not sure what error message it will throw you if you pass a 3D array, and it might be hard to decipher

not sure it is fully finished in this current form, but the logic seems there

arokem:
Good comments both -- thanks for taking a look. I've removed the nanmean
cruft and added a check up-front for scalar sigma.

arokem:
I added the option of passing in an array, if that array is a singleton. This fixes the test failures from yesterday (at least on my machine), and actually seems reasonable to me. I don't think that we need to adjust the documentation. The canonical input is still a single float, it's just that if you happen to deliver that single float in an array, we'll still accept that. 

Now also rebased on master

dipy/denoise/tests/test_non_local_means.py
```diff
@@ -81,7 +96,7 @@ def test_nlmeans_dtype():
     S0 = 200 * np.ones((20, 20, 20), dtype=np.uint16)
     mask = np.zeros((20, 20, 20))
     mask[10:14, 10:14, 10:14] = 1
-    S0n = non_local_means(S0, sigma=np.ones((20, 20, 20)), mask=mask,
+    S0n = non_local_means(S0, sigma=1, mask=mask,
```
MarcCote: That probably fits on one line now.

dipy/denoise/tests/test_non_local_means.py
```diff
@@ -39,7 +55,7 @@ def test_nlmeans_boundary():
 
     S0[:10, :10, :10] = 300 + noise[:10, :10, :10]
 
-    S0n = non_local_means(S0, sigma=np.ones((20, 20, 20)) * np.std(noise),
+    S0n = non_local_means(S0, sigma=np.std(noise),
```
MarcCote: That probably fits on one line now.

arokem:
Indeed. I've removed some other un-needed whitespace, while I was at it.

PR #1171, opened 2017-01-24
fix:dev: added numpy.int64 for my triangle array

vibhatha:
I added the data type of my_triangle array to fix the error caused in running with Anaconda2.


arokem:
Looks fine to me. Maybe you could add a comment mentioning that VTK requires this type, and sometimes that's not the type you automatically get from numpy.

While we're here, could you please change all the places where it says `vetices` to be `vertices`? Seems like a typo to me

vibhatha:
Yes I can do that.


With Regards,
Vibhatha Abeykoon

vibhatha:
I added the comments and also did the spelling check for vertices word.

vibhatha:
Hey, did I do something wrong. It's not clear to me, the coverage decrease
matter.

With Regards,
Vibhatha Abeykoon

Garyfallidis:
@vibhatha this is fine. The final fix should go into the function itself. But we are going to make an ``actor.surface`` function for that give a more thorough inspection into the code. So, this fix is fine. The coverage errors are just not informative in this case. So we can merge with no issue.

vibhatha:
@Garyfallidis I can understand. Happy to help to improve things.

PR #1174, opened 2017-01-31
NF: added the repulsion 200 sphere.

jchoude:
This sphere was generated using the electrostatic repulsion algorithm
to correctly space 200 points on the unit sphere.

This sphere is needed to reproduce results for the 2012 Raffelt paper
on Apparent Fiber Density (doi:10.1016/j.neuroimage.2011.10.045). They used a 200 points sphere.


arokem:
LGTM. Tell me that you are also planning to upload an example where you reproduce that paper! ;-)

RafaelNH:
Any plans of incorporating the electrostatic repulsion algorithm code in dipy? I am guessing that you cannot use function disperse_charges from dipy.core.sphere because this assumes bipolar repulsion to space directions on the hemisphere, right?

jchoude:
@arokem Well, for that, I would need to add the code to compute said AFD :) Currently working on it, I'll try to merge it here.

@RafaelNH well, in fact, I used the ```disperse_charges``` function to tune the points. Maybe I didn't use the correct name for the algorithm?

arokem:
@RafaelNH : does that answer your question? I think this is all in order, but want to make sure that this makes sense to you, before merging.

arokem:
Since I haven't heard back, I will assume that this is fine. Here goes. Can't wait for the follow-up PR, @jchoude ;-)

PR #1177, opened 2017-02-06
BF: fix interpolation call with Numpy 1.12

jchoude:
This fixed the issue described in https://github.com/nipy/dipy/issues/1176

Numpy 1.12 now Raises when indexing an with a double instead of an int. We cast to int.


arokem:
LGTM. I can merge as soon as Travis comes back green.

Issue #1167, opened 2016-12-17
Potential replacement for CVXOPT?


arokem:
https://github.com/biosustain/optlang

Released under Apache license.

arokem:
Another one, much more similar to cvxopt (I think), also licensed under Apache: 

http://www.cvxpy.org/en/latest/license/index.html

Garyfallidis:
So many possibilities! How nice!

skoudoro:
I think, it is done via #1285, we close this issue ?

arokem:
indeed!

PR #1180, opened 2017-02-20
WIP: replacing cvxopt with cvxpy.

arokem:
Following up on #1167

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ addons:
 
 env:
     global:
-        - DEPENDS="cython numpy scipy matplotlib h5py nibabel cvxopt"
+        - DEPENDS="cython numpy scipy matplotlib h5py nibabel cvxpy"
         - VENV_ARGS="--python=python"
         - INSTALL_TYPE="setup"
         - EXTRA_WHEELS="https://5cf40426d9f06eb7461d-6fe47d9331aba7cd62fc36c7196769e4.ssl.cf2.rackcdn.com"
@@ -39,10 +39,10 @@ matrix:
         - DEPENDS="cython==0.18 numpy==1.7.1 scipy==0.9.0 nibabel==2.1.0"
     - python: 2.7
       env:
-        - DEPENDS="$DEPENDS scikit_learn tables"
+        - DEPENDS="cython numpy scipy matplotlib h5py nibabel cvxpy scikit_learn tables"
     - python: 3.5
       env:
-        - DEPENDS="$DEPENDS scikit_learn tables"
+        - DEPENDS="cython numpy scipy matplotlib h5py nibabel cvxpy scikit_learn tables"
     # To test vtk functionality
     - python: 2.7
       sudo: true   # This is set to true for apt-get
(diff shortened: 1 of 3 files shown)


arokem:
So far based on this conversation: https://groups.google.com/forum/#!searchin/cvxpy/quad_form$20cvxopt%7Csort:relevance/cvxpy/LiPoOGWmiIY/i2TOloCuDQAJ

grlee77:
I assume the motivation for switching to CVXPY is for the Apache license terms (as opposed to the GPL-licensed CVXOPT)?  In that case, does it really help if the underlying GPL-licensed ECOS solver is still being used by CVXPY here?

It looks like CVXPY can also use the MIT-licensed SCS solver instead of ECOS, but I am not familiar enough with the solvers to know whether it is a suitable substitute for ECOS in this case.

arokem:
Jokes on me, huh? Well, FWIW, SCS is inaccurate in an identical way to ECOS.

arokem:
Hey @matthew-brett : what do we need to do in order to get cvxpy into the wheel-house? Thanks!

arokem:
Ugh, ecos: https://github.com/embotech/ecos-python/issues/8

samuelstjean:
While people are on the subject of switching solvers, please make sure that the replacement would at least be as accurate as cvxopt, these thing are always super hard to debug when you have a midway loss of precision due to who knows what.

Also, if you really, really just want to avoid it, scipy.optimize.nnls, but the speed is abysmal (like, unusable in practice when called a million times) because it is an all around purpose solver. But other implementation of the same algorithm might exist elsewhere (although a well done convex formulation will always beat it, this one does the old fashion way we learned of solving the KKT conditions for problems on paper.)

Does not look like sklearn ridge solvers does constraints or positivity though, but they might have some stuff in that area.

arokem:
Thanks for your input. Help figuring out specifically how to make ECOS work here would be very welcome.

samuelstjean:
In what sense? I thought it was amongst the few solvers that came out of the box with cvxpy, unless they changed stuff when they changed the license also (it used to be gpl not too long ago, well I checked it a few years back already though).

arokem:
Oh yes. I meant SCS, which is in fact MIT-licensed (
https://github.com/cvxgrp/scs/blob/master/LICENSE.txt), and what the
current code in this PR uses (unsuccessfully, so far).

samuelstjean:
Not sure what you mean, cvxpy is designed as a easy to use frontend to many solvers. You write stuff in it's syntax, and it does the job of using the best available option for you, as not all solvers are equally good at everything.

And as you do here you force it to use scs in particular, but it seems to not be able to solve the problem in some cases (which is expected somewhat, because if not there would be only the need for a single solver to do everything, the cool part about cvxpy is that you don't have to bother with those details).

Anyway, that's what I got from your problem. If not, well it builds (unlike ecos, were the instructions on windows is basically give up https://www.embotech.com/ECOS/Install).

Does it work if you do not force it to use a solver? Might want to check what it does use under the hood and if it's the same in all cases. I also have stuff to prepare for ismrm, so I won't likely be able to help out for stuff if it was a pressing matter though.

arokem:
Thanks anyway.

skoudoro:
I think, it is done via #1285, we close this PR?

arokem:
yes. this was superseded by #1285

PR #1186, opened 2017-03-06
MRG: use newer wheelhouse for installs

matthew-brett:
We were using a custom wheelhouse, but now manylinux wheels are pretty
well supported; switch to using the standard pypi wheels plus some extra
wheels for older distributions.

This should address the need for cvxpy wheels - see #1180.


arokem:
Time to also drop Python 3.3?

arokem:
Looks like numpy is no longer supporting it? https://travis-ci.org/nipy/dipy/jobs/208408058

matthew-brett:
I think this PR is correct now, but it has exposed a couple of test errors with latest numpy / matplotlib:

* #1188
* #1187

matthew-brett:
I also dropped Python 3.3.

arokem:
See #1189, which addresses the negative integer powers. When that's merged you can rebase here.

arokem:
Also made a PR directly into this branch (https://github.com/matthew-brett/dipy/pull/5), so that we can test this on Travis with the right version of numpy, before moving ahead with that.

matthew-brett:
A couple more issues arising:

* Python 3.6 failure : #1190;
* Failure for current numpy master : #1191.

Otherwise, we're just waiting on #1188 .

matthew-brett:
From my point of view, this is ready to merge when #1188 and #1190 are fixed.

arokem:
This should be GTG after a rebase!

matthew-brett:
Done rebase - but is #1190 fixed yet?

arokem:
Oh - you're right 😞 

Looks nasty.

arokem:
Are we ready for this one yet?

matthew-brett:
Rebased.

arokem:
So how should we handle the failures in the "pre" machine? 

https://travis-ci.org/nipy/dipy/jobs/214060347

Should we go ahead and merge this and fix those on subsequent PRs? Given how long we go between releases, we should definitely fix these before making a release. But do we need to fix these before merging this?

matthew-brett:
If it was up to me, I'd make sure the pre-release numpy checks were working, before the release, to give us some confidence that dipy won't be broken by the next release of numpy.

matthew-brett:
Having said that, I think this one's OK to merge, but with a plan to fix #1191 before the next release.

arokem:
Yup. Makes sense. Merging.

arokem:
Interesting. Looks like this also sped up Travis runs quite substantially. My impression is that it went down from ~3 hrs down to ~1 hr. It might also be something else.

Issue #1188, opened 2017-03-07
Colormap test errors for new matplotlib


matthew-brett:
Test error when testing against latest matplotlib:

```
======================================================================
FAIL: dipy.viz.tests.test_fvtk.test_colormaps_matplotlib
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/matthew-brett/dipy/venv/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/build/matthew-brett/dipy/venv/lib/python2.7/site-packages/numpy/testing/decorators.py", line 147, in skipper_func
    return f(*args, **kwargs)
  File "/home/travis/build/matthew-brett/dipy/venv/lib/python2.7/site-packages/dipy/viz/tests/test_fvtk.py", line 118, in test_colormaps_matplotlib
    npt.assert_array_almost_equal(rgba1, rgba2, 1)
  File "/home/travis/build/matthew-brett/dipy/venv/lib/python2.7/site-packages/numpy/testing/utils.py", line 979, in assert_array_almost_equal
    precision=decimal)
  File "/home/travis/build/matthew-brett/dipy/venv/lib/python2.7/site-packages/numpy/testing/utils.py", line 796, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Arrays are not almost equal to 1 decimals
(mismatch 16.675%)
 x: array([[ 1. ,  1. ,  0.6,  1. ],
       [ 0.2,  0.4,  0.7,  1. ],
       [ 0.7,  0.7,  0.8,  1. ],...
 y: array([[ 0.8,  0.8,  0.6,  1. ],
       [ 0.6,  0.7,  0.6,  1. ],
       [ 0.8,  0.7,  0.7,  1. ],...
```

https://travis-ci.org/matthew-brett/dipy/jobs/208424940

I'm guessing this is because of the new default 'viridis' colormap in matplotlib >= 2.0.

matthew-brett:
Cross ref new testing regime at #1186 .

arokem:
Strangely enough, this is happening with the "Accent" color map. Looks like that color-map might have changed in Matplotlib. We might need to change the values we have in: 

https://github.com/nipy/dipy/blob/master/dipy/data/files/dipy_colormaps.json

@MrBago : how were those values generated? IIRC, you created that file.

matthew-brett:
[...]

PR #1193, opened 2017-03-13
Deprecate the Accent colormap, in anticipation of changes in MPL 2.0

arokem:
Addresses #1188

### DIFF
--- a/dipy/data/__init__.py
+++ b/dipy/data/__init__.py
@@ -362,6 +362,9 @@ dipy_cmaps = None
 
 def get_cmap(name):
     """Makes a callable, similar to maptlotlib.pyplot.get_cmap"""
+    if name.lower() == "accent":
+        raise ValueError("The `Accent` colormap is deprecated as of version",
+                         " 0.12 of Dipy. Please use another colormap")
     global dipy_cmaps
     if dipy_cmaps is None:
         filename = pjoin(DATA_DIR, "dipy_colormaps.json")
--- a/dipy/data/__init__.py
+++ b/dipy/data/__init__.py
@@ -365,7 +365,8 @@ def get_cmap(name):
     if name.lower() == "accent":
         warnings.warn("The `Accent` colormap is deprecated as of version" +
                       " 0.12 of Dipy and will be removed in a future " +
-                      "version. Please use another colormap")
+                      "version. Please use another colormap",
+                      DeprecationWarning)
     global dipy_cmaps
     if dipy_cmaps is None:
         filename = pjoin(DATA_DIR, "dipy_colormaps.json")
--- a/dipy/data/__init__.py
+++ b/dipy/data/__init__.py
@@ -1,11 +1,11 @@
 """
 Read test or example data
 """
-
 from __future__ import division, print_function, absolute_import
 
 import sys
 import json
+import warnings
 
 from nibabel import load
 from os.path import join as pjoin, dirname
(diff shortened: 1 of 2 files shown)


dipy/data/__init__.py
```diff
@@ -362,6 +362,9 @@ def mrtrix_spherical_functions():
 
 def get_cmap(name):
     """Makes a callable, similar to maptlotlib.pyplot.get_cmap"""
+    if name.lower() == "accent":
+        raise ValueError("The `Accent` colormap is deprecated as of version",
```
matthew-brett: Not deprecating though - seems a bit much to go straight to error.

dipy/viz/tests/test_fvtk.py
```diff
@@ -109,14 +109,17 @@ def test_colormap():
 @npt.dec.skipif(not fvtk.have_matplotlib)
 def test_colormaps_matplotlib():
     v = np.random.random(1000)
-    for name in 'jet', 'Blues', 'Accent', 'bone':
+    for name in 'jet', 'Blues', 'bone':
```
matthew-brett: I guess you add make the `accent` test conditional on mpl < 2.

arokem: On further thought, I think we should deprecate this colormap, regardless of the version of MPL that the user happens to have installed. The reasons they have for changing it into a linear segmented color map apply to us as well, and we shouldn't be using it as a continuous color map at all.

matthew-brett: Sure, I agree about deprecation regardless, just suggesting testing the colormap to make sure it's correct, even if deprecated.

dipy/data/__init__.py
```diff
@@ -362,6 +362,10 @@ def mrtrix_spherical_functions():
 
 def get_cmap(name):
     """Makes a callable, similar to maptlotlib.pyplot.get_cmap"""
+    if name.lower() == "accent":
+        warnings.warn("The `Accent` colormap is deprecated as of version" +
```
matthew-brett: FutureWarning ?   DeprecationWarning?

arokem: Oh yes - just what I needed! 

Also adding a test for the color-map for MPL < 2!

dipy/viz/tests/test_fvtk.py
```diff
@@ -109,7 +118,18 @@ def test_colormap():
...
+        accent_cm = data.get_cmap("Accent")
+        # Test that the deprecation warning was raised:
+        npt.assert_equal(len(w) > 0, True)
+
+    names = ['jet', 'Blues', 'bone']
+
+    if have_matplotlib and mpl_version < "2":
+        names.append('Accent')
+
+    for name in names:
```
matthew-brett: Won't we need to wrap the mpl >= 2 / accent test in a catch_warnings block?

arokem: Well, the warning is raised regardless of if you are using MPL > 2 or not, because we are deprecating this cmap for all users. But now I also implemented a test that checks if the warning is raised in the loop below as well.

dipy/viz/tests/test_fvtk.py
```diff
@@ -109,13 +118,31 @@ def test_colormap():
...
+        with warnings.catch_warnings(record=True) as w:
+            # Matplotlib version of get_cmap
+            rgba1 = fvtk.get_cmap(name)(v)
+            # Dipy version of get_cmap
+            rgba2 = data.get_cmap(name)(v)
+            # dipy's colormaps are close to matplotlibs colormaps, but not
+            # perfect:
+            npt.assert_array_almost_equal(rgba1, rgba2, 1)
+            if name == "Accent":
```
matthew-brett: How about:

```
npt.assert_(len(w) == (1 if name == 'Accent' else 0))
```

arokem: Sure. That's neat.

matthew-brett:
Great - thanks a lot.

Issue #1190, opened 2017-03-07
Pickle error for Python 3.6 and test_peaksFromModelParallel


matthew-brett:
```
======================================================================
ERROR: dipy.direction.tests.test_peaks.test_peaksFromModelParallel
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/matthew-brett/dipy/venv/lib/python3.6/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/build/matthew-brett/dipy/venv/lib/python3.6/site-packages/dipy/direction/tests/test_peaks.py", line 509, in test_peaksFromModelParallel
    return_sh=True, parallel=True)
  File "/home/travis/build/matthew-brett/dipy/venv/lib/python3.6/site-packages/dipy/direction/peaks.py", line 419, in peaks_from_model
    nbr_processes)
  File "/home/travis/build/matthew-brett/dipy/venv/lib/python3.6/site-packages/dipy/direction/peaks.py", line 227, in _peaks_from_model_parallel
    repeat(invB)))
  File "/opt/python/3.6.0/lib/python3.6/multiprocessing/pool.py", line 260, in map
    return self._map_async(func, iterable, mapstar, chunksize).get()
  File "/opt/python/3.6.0/lib/python3.6/multiprocessing/pool.py", line 608, in get
    raise self._value
multiprocessing.pool.MaybeEncodingError: Error sending result: '[<dipy.direction.peaks.PeaksAndMetrics object at 0x2974a18>]'. Reason: 'TypeError("can't pickle PeaksAndMetrics objects",)'
```

https://travis-ci.org/matthew-brett/dipy/jobs/208734316#L3259

See #1186 for PR extending testing to Python 3.6.

arokem:
Strangely, I can't replicate this on my laptop. Will try to do this on a linux box.

matthew-brett:
Can you pickle the `PeaksAndMetrics` instances on Python 3.6, Numpy 1.12 ?

arokem:
OK - just tested specifically that. 

No, I can't pickle a PAM object under these conditions. Will continue to investigate.

arokem:
This seems to help: #1195 (but I will admit that I don't entirely understand it)

arokem:
closed through #1195

PR #1195, opened 2017-03-20
Make PeaksAndMetrics pickle-able

arokem:
Might help with #1190 

See: http://stackoverflow.com/a/2050357/3532933

### DIFF
--- a/dipy/direction/peaks.py
+++ b/dipy/direction/peaks.py
@@ -547,8 +547,15 @@ def peaks_from_model(model, data, sphere, relative_peak_threshold,
     if not return_odf:
         odf_array = None
 
-    return _pam_from_attrs(PeaksAndMetrics, sphere, peak_indices, peak_values,
-                           peak_dirs, gfa_array, qa_array, shm_coeff, B,
+    return _pam_from_attrs(PeaksAndMetrics,
+                           sphere,
+                           peak_indices,
+                           peak_values,
+                           peak_dirs,
+                           gfa_array,
+                           qa_array,
+                           shm_coeff,
+                           B,
                            odf_array)
 
 
--- a/dipy/direction/peaks.py
+++ b/dipy/direction/peaks.py
@@ -157,8 +157,20 @@ def peak_directions(odf, sphere, relative_peak_threshold=.5,
     return directions, values, indices
 
 
+class PeaksAndMetrics(PeaksAndMetricsDirectionGetter):
+    def __reduce__(self): return _pam_from_attrs, (self.sphere,
+                                                   self.peak_indices,
+                                                   self.peak_values,
+                                                   self.peak_dirs,
+                                                   self.gfa,
+                                                   self.qa,
+                                                   self.shm_coeff,
+                                                   self.B,
+                                                   self.odf)
+
+
 def _pam_from_attrs(sphere, peak_indices, peak_values, peak_dirs,
-                    gfa, qa, shm_coeff, B, odf):
+                    gfa, qa, shm_coeff, B, odf, klass=PeaksAndMetrics):
     """
     Construct a PeaksAndMetrics object from its attributes.
 
@@ -187,12 +199,14 @@ def _pam_from_attrs(sphere, peak_indices, peak_values, peak_dirs,
         coefficients.
     odf : ndarray
         The orientation distribution function on the sphere in each voxel.
+    klass : class, optional
+        The class of object to be created. Default: PeaksAndMetrics
 
     Returns
     -------
-    PeaksAndMetrics class instance.
+    pam : Instance of the class `klass`.
     """
-    this_pam = PeaksAndMetrics()
+    this_pam = klass()
     this_pam.sphere = sphere
     this_pam.peak_dirs = peak_dirs
     this_pam.peak_values = peak_values
@@ -203,18 +217,6 @@ def _pam_from_attrs(sphere, peak_indices, peak_values, peak_dirs,
     this_pam.B = B
     this_pam.odf = odf
     return this_pam
-
-
-class PeaksAndMetrics(PeaksAndMetricsDirectionGetter):
-    def __reduce__(self): return _pam_from_attrs, (self.sphere,
-                                                   self.peak_indices,
-                                                   self.peak_values,
-                                                   self.peak_dirs,
-                                                   self.gfa,
-                                                   self.qa,
-                                                   self.shm_coeff,
-                                                   self.B,
-                                                   self.odf)
 
 
 def _peaks_from_model_parallel(model, data, sphere, relative_peak_threshold,
--- a/dipy/direction/peaks.py
+++ b/dipy/direction/peaks.py
@@ -548,7 +548,8 @@ def peaks_from_model(model, data, sphere, relative_peak_threshold,
         odf_array = None
 
     return _pam_from_attrs(PeaksAndMetrics, sphere, peak_indices, peak_values,
-                           peak_dirs, gfa_array, qa_array, shm_coeff, B, odf_array)
+                           peak_dirs, gfa_array, qa_array, shm_coeff, B,
+                           odf_array)
 
 
 def gfa(samples):
--- a/dipy/direction/peaks.py
+++ b/dipy/direction/peaks.py
@@ -157,8 +157,31 @@ def peak_directions(odf, sphere, relative_peak_threshold=.5,
     return directions, values, indices
 
 
+def _rebuild_pam(sphere, peak_indices, peak_values, peak_dirs,
+                 gfa, qa, shm_coeff, B, odf):
+    this_pam = PeaksAndMetrics()
+    this_pam.sphere = sphere
+    this_pam.peak_dirs = peak_dirs
+    this_pam.peak_values = peak_values
+    this_pam.peak_indices = peak_indices
+    this_pam.gfa = gfa
+    this_pam.qa = qa
+    this_pam.shm_coeff = shm_coeff
+    this_pam.B = B
+    this_pam.odf = odf
+    return this_pam
+
+
 class PeaksAndMetrics(PeaksAndMetricsDirectionGetter):
-    pass
+    def __reduce__(self): return _rebuild_pam, (self.sphere,
+                                                self.peak_indices,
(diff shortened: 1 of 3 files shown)


arokem:
Travis failure unrelated to this PR, and only on Python 3.3

matthew-brett:
Further explanation : https://snorfalorpagus.net/blog/2016/04/16/pickling-cython-classes

dipy/reconst/peak_direction_getter.pyx
```diff
@@ -38,6 +38,10 @@ cdef class PeaksAndMetricsDirectionGetter(DirectionGetter):
         self.ang_thr = 60
         self.total_weight = .5
 
+    def __getstate__(self): return self.__dict__
+
+    def __setstate__(self, d): self.__dict__.update(d)
```
matthew-brett: The `update` will leave any previous members in place, that are not defined in `d`.  Is that the right thing to do?

arokem:
Thanks for finding that post!

Do you understand why we don't see this in other versions of Python? Seems
like we should have been getting this as far back as at least 3.4.

matthew-brett:
No, that is puzzling.  I wonder whether it's something random about when classes get thrown around by the multiprocessing ?

matthew-brett:
Sorry - I hate to say - but it would be good to have a test of pickling and unpickling these guys in fresh and used state.

arokem:
To answer the other thing: From reading the documentation, it seems that
maybe we don't even need to implement `__setstate__`, so I have removed
that one.

arokem:
Yes. Fair enough. Since I just did that in a notebook, to be sure that I am
not messing anything up, may as well add this to the tests... Coming!

arokem:
OK - here it is. Let's see if I managed to write something that works on
both Python 2 and Python 3...

arokem:
I'm learning so many things about pickles over here :smile:

Previous attempt seemed to be failing on Python 2. Here's a new approach that might work on both.

arokem:
Changing the title of the PR accordingly 😉

dipy/direction/peaks.py
```diff
@@ -157,8 +157,63 @@ def peak_directions(odf, sphere, relative_peak_threshold=.5,
     return directions, values, indices
 
 
+def _rebuild_pam(sphere, peak_indices, peak_values, peak_dirs,
```
matthew-brett: Why not make this a classmethod called `from_params` or similar?

arokem: It would be a bit weird as a class method, considering that it returns another instance of that class. Or would it? 

There does seem to be an opportunity to reduce boiler-plate here though, so I am going to try out a couple more things. Stand by.

arokem:
What do you think about this? I renamed it to `_pam_from_attrs` and I am now also using it to initialize the PAM instance in the `peaks_from_model` function below (which had a similar pattern).  I prefer `attrs` to `params`, to not confuse this with model parameters.

arokem:
And I think this looks better than: 


   pam = PeaksAndMetrics()
   pam = pam._from_attrs(...)

which is what we would do if we had this implemented as a class method (I think).

matthew-brett:
The floating external function looks funny to me, especially for a class that is otherwise empty.

Here's the alternative classmethod I was thinking of:

```
pam = PeaksAndMetrics()
pam = PeaksAndMetrics.from_attrs(...)
```

To me it looks like a normal alternative class constructor pattern...

arokem:
Gotcha. How's this?

arokem:
Sadly, this approach does not work on Python 2: https://travis-ci.org/arokem/dipy/jobs/213491090#L2580

Any ideas?

arokem:
Looks like the class method is unpickle-able in Python 2, which defeats the original purpose... 
What do you think? Move it back to be a weird external function?

matthew-brett:
Does it work as an instance method?

On 21 Mar 2017 12:55, "Ariel Rokem" <notifications@github.com> wrote:

Looks like the class method is unpickle-able in Python 2, which defeats the
original purpose...
What do you think? Move it back to be a weird external function?

—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub
<https://github.com/nipy/dipy/pull/1195#issuecomment-288198921>, or mute
the thread

arokem:
Nope. On both Python 2 and Python 3, I get `TypeError: can't pickle
instancemethod objects`

arokem:
Oh - sorry. On Python3 that seems to lead to a recursion error. But still.

arokem:
For now, I have reverted to a funny free-standing function, until another
solution comes up.

arokem:
OK - this is now again only failing on Python 3.3

dipy/direction/peaks.py
```diff
@@ -157,8 +157,64 @@ def peak_directions(odf, sphere, relative_peak_threshold=.5,
     return directions, values, indices
 
 
+def _pam_from_attrs(sphere, peak_indices, peak_values, peak_dirs,
```
matthew-brett: Consider adding class as first argument?   Or as optional argument at end?  In case we need to sub-class?

Sorry to ask, but why do we need this otherwise empty sub-class? 

```
class PeaksAndMetrics(PeaksAndMetricsDirectionGetter): pass
```

Why not use `PeaksAndMetricsDirectionGetter` directly?

dipy/direction/peaks.py
```diff
@@ -157,8 +157,64 @@ def peak_directions(odf, sphere, relative_peak_threshold=.5,
...
+        each voxel.
+    B : ndarray
+        The spherical harmonic matrix, for multiplication with the
+        coefficients.
+    odf : ndarray
+        The orientation distribution function on the sphere in each voxel.
+
+    Returns
+    -------
+    PeaksAndMetrics class instance.
```
matthew-brett: I think docstring standard says something like:

```
Returns
-------
pam : PeaksAndMetrics class instance
```

or similar (give name to output arg).

arokem:
On Wed, Mar 22, 2017 at 11:50 AM, Matthew Brett <notifications@github.com>
wrote:

> *@matthew-brett* commented on this pull request.
> ------------------------------
>
> In dipy/direction/peaks.py
> <https://github.com/nipy/dipy/pull/1195#discussion_r107501051>:
>
> > @@ -157,8 +157,64 @@ def peak_directions(odf, sphere, relative_peak_threshold=.5,
>      return directions, values, indices
>
>
> +def _pam_from_attrs(sphere, peak_indices, peak_values, peak_dirs,
>
> Consider adding class as first argument? Or as optional argument at end?
> In case we need to sub-class?
>
> Sorry to ask, but why do we need this otherwise empty sub-class?
>
> class PeaksAndMetrics(PeaksAndMetricsDirectionGetter): pass
>
> Why not use PeaksAndMetricsDirectionGetter directly?
>
>
I'm not sure why we are subclassing here.

I added another argument for sub-classing purposes, with PeaksAndMetrics as
the default value.

dipy/direction/peaks.py
```diff
@@ -157,8 +157,20 @@ def peak_directions(odf, sphere, relative_peak_threshold=.5,
...
+class PeaksAndMetrics(PeaksAndMetricsDirectionGetter):
+    def __reduce__(self): return _pam_from_attrs, (self.sphere,
+                                                   self.peak_indices,
+                                                   self.peak_values,
+                                                   self.peak_dirs,
+                                                   self.gfa,
+                                                   self.qa,
+                                                   self.shm_coeff,
+                                                   self.B,
+                                                   self.odf)
```
matthew-brett: Add `self.__class__` as argument to allow sub-classing?   Maybe, it would be better as first argument.

arokem:
Yup. Moved to first argument.

dipy/direction/peaks.py
```diff
@@ -157,8 +157,68 @@ def peak_directions(odf, sphere, relative_peak_threshold=.5,
...
+    """
+    Construct a PeaksAndMetrics object (or object of a subclass) from its
+    attributes.
+
+    This is also useful for pickling/unpickling of these objects (see also
+    :func:`__reduce__` below).
+
+    Parameters
+    ----------
+    klass : class, optional
```
matthew-brett: No longer optional !

matthew-brett:
All OK for me, after fix of `optional` in docstring.

arokem:
Good catch. Thanks!

dipy/direction/peaks.py
```diff
@@ -480,27 +540,16 @@ def peaks_from_model(model, data, sphere, relative_peak_threshold,
...
 
-    if return_odf:
-        pam.odf = odf_array
-    else:
-        pam.odf = None
+    if not return_odf:
+        odf_array = None
 
-    return pam
+    return _pam_from_attrs(PeaksAndMetrics, sphere, peak_indices, peak_values,
```
matthew-brett: I guess, if you can be bothered:

```
return _pam_from_attrs(PeaksAndMetrics,
                       sphere,
                       peak_indices,
                       peak_values,
                       peak_dirs,
                       gfa_array,
                       qa_array,
                       shm_coeff if return_sh else None,
                       B if return_sh else None,
                       odf_array if return_odf else None)
```

arokem:
I can always be bothered

matthew-brett:
Ah sorry - I mean particularly cleaning up the slightly ugly if blocks with the conditionals:

```
...
shm_coeff if return_sh else None,
B if return_sh else None,
odf_array if return_odf else None)
```

arokem:
Oops. I didn't really bother reading all the way down :-)

Fixed now.

matthew-brett:
Thanks for your patience.  One of us merge when the tests pass (apart from the 3.3 failure)?

arokem:
The patience is all yours. Last one to the merge button is a rotten egg.

arokem:
Beat you to it!

matthew-brett:
Good job - but - aren't the tests still running?

arokem:
I think not: https://travis-ci.org/arokem/dipy/builds/214021129

arokem:
Addmitedly, I cheated a little bit, because I have Travis looking over my
own fork as well :-)

matthew-brett:
Well played sir.

PR #1198, opened 2017-03-23
ENH: avoid log zero

arokem:
This continues #1086


arokem:
Let's see how our minimal numpy does with this...

arokem:
Looks like it did OK. @jchoude : what do you think about this solution?

jchoude:
Seems all good to me! I can merge if you want. Do we mind if the coveralls check did not pass (-0.002% difference)

arokem:
I think that coveralls is complaining because we added four lines of code, but didn't add tests. So the difference is negative, if tiny. I wouldn't worry about it. That said, I now added a test with zeros as input, to make sure that the warning is properly suppressed, and that the result is correct.

jchoude:
Ok! I'll wait until Travis reports back, and then I'll merge this and close my original PR.

jchoude:
Seems all good!

jchoude:
Thanks @arokem !

PR #1183, opened 2017-03-02
A lightweight UI for medical visualizations #2: TextBox

ranveeraggarwal:
After introducing the base UI class, the button, and the panel in #1140 this is the next update in the series of PRs meant to incorporate changes of #1111.  
This PR includes a `TextActor2D` element that subclasses `vtkTextActor` and a `TextBox2D` element that gives you an editable multi-line text area.

PR #1111, opened 2016-08-19
WIP: A lightweight UI for medical visualizations based on VTK-Python

ranveeraggarwal:
This PR introduces a VTK-based UI. All the UI elements are within the same window as the other elements. We do not use Qt/GTK or any other UI library apart from VTK. More information about the project here: http://ranveeraggarwal.com/subblogs/gsoc-16/

PR #1199, opened 2017-03-27
A lightweight UI for medical visualizations #3: Changes to Event Handling

ranveeraggarwal:
This PR introduces a new way of handling events. Earlier, the `handle_events` method was a part of the base `UI` class and an `add_callback` method had to be written within each subclass to make `handle_events` work. This was kind of confusing because, 

1. It restricted all subclasses to have a compulsory override of `add_callback`
2. The event handling is not the same for all kinds of UI elements, some elements might have multiple actors, in which case, this gets very confusing

Now, there is just one `add_callback` function that is a part of the base UI class. `handle_events` needs to be explicitly called by each subclass. If a subclass has multiple actors, `handle events` needs to be overridden.

This PR also breaks down the tests into multiple functions so that the events of the previous elements needn't be re-recorded.

This is the third PR to complete #1111 and continues from #1183.


MarcCote:
@ranveeraggarwal: on to the line slider :)

PR #1201, opened 2017-03-28
Fix out of bounds point not being classified OUTSIDEIMAGE (binary cla…

matthieudumont:
Binary Classifier: Fix bad classification for a point 1 voxel outside the mask shape in any direction.


arokem:
Would you mind adding a test that demonstrates why this fix is needed?

arokem:
Yup. LGTM. Thanks!

Issue #1107, opened 2016-08-08
Dipy.align.reslice: either swallow the scipy warning or rework to avoid it


jchoude:
When calling dipy.align.reslice and use scipy 0.17.1, we get the following UserWarning:

```
scipy/ndimage/interpolation.py:435: UserWarning: The behaviour of affine_transform with a one-dimensional array supplied for the matrix parameter will change in scipy 0.18.0.
  "The behaviour of affine_transform with a one-dimensional "
```

This is caused by a change in behavior from Scipy, between 0.17.1 and 0.18, as explained [in Scipy's doc](http://docs.scipy.org/doc/scipy-0.18.0/reference/generated/scipy.ndimage.affine_transform.html).

For us, there won't be any consequence, since the behavior won't change, because we do not use the `offset` parameter. However, the warning might be misleading or simply distracting for end users, and it will still be present in scipy 0.18. That means that it will pollute our logs for a long time.

Two ideas:
1- Swallow the warning by using something like [this](https://github.com/scikit-image/scikit-image/blob/master/skimage/_shared/_warnings.py#L77) (as was suggested in issue #641 )
2- Simply reformat the matrix we send as a diagonal matrix, which will provide exactly the same behavior, without the warning.

I would go for option 2, but would like to have some other advice before proceeding @Garyfallidis @arokem

arokem:
Yes: I like option 2 better too. Presumably changing this line?

https://github.com/nipy/dipy/blob/master/dipy/align/reslice.py#L71

to the following?

    R = np.diag(new_zooms / zooms)

sahmed95:
Changing to ```R = np.diag(new_zooms / zooms)``` fails the test
```
FAIL: dipy.align.tests.test_reslice.test_resample
----------------------------------------------------------------------
Traceback (most recent call last):
  File "dipy/dipy/align/tests/test_reslice.py", line 25, in test_resample
    assert_almost_equal(new_zooms, new_zooms_confirmed)

AssertionError: 
Arrays are not almost equal to 7 decimals

(mismatch 100.0%)
 x: array([ 1. ,  1.2,  2.1])
 y: array([ 1.7320508,  2.0784609,  3.6373067], dtype=float32)
```

sahmed95:
@arokem Any comments on this ? Is changing test parameters okay or is ```R = np.diag(new_zooms / zooms) ``` not the correct way to go. @jchoude

arokem:
[...]

PR #1206, opened 2017-03-30
Revert #1204, and add a filter to suppress warnings.

arokem:
See also #1107


MarcCote:
I agree with what you said. Thank you @dimrozakis for the clarification.

Issue #1213, opened 2017-04-04
dipy.align.vector_fields.simplify_warp_function_3d: Wrong equation in docstring


quakberto:
Hi,

the docstring of the function simplify_warp_function_3d says that its result will be a deformation field satisfying:
(1) T[i] = W * d[U * i] + V * i

However, I think this equation should be:
(1) T[i] = W * d[U * i] + V * i - i or
(1) T[i] = W * d[U * i] + (V - I)* i,
since in lines 1585, 1587 and 1589 the coordinate is subtracted after the applying the affine transform V:
_apply_affine_3d_x0(k, i, j, 1, affine_idx_out) - k

Hope this is the right place for this comment and thanks for the great tool!
Best,
quakberto

omarocegueda:
Hi @quakberto!,
thank you very much for reporting this!, we are taking a closer look to make sure it's just a documentation error.

omarocegueda:
Hi @quakberto,
The documentation is correct. I think the confusion is that the user may expect `out = T`, but that would be inconsistent with the rest of the `align` module, since all non-linear *transforms* are represented using *deformation fields* instead of actual mapped values for accuracy reasons. 

More precisely, if the output of `simplify_warp_function_3d ` is `out` then, in order to compute `T` in (1) we need to do:
```
T[i] = i + out[i] 
```
I'll make a PR to state this more precisely.
Thanks!

quakberto:
Hi @omarocegueda,
that makes sense! Indeed I thought that T was the intended output.
I'm only starting to get familiar with nipy/dipy's conventions, so thanks a lot for the clarification!
Best

arokem:
(hopefully) Closed through #1214

PR #1214, opened 2017-04-04
DOC: add a clarification note to simplify_warp_funcion_3d

omarocegueda:
Adds a clarification note to prevent confusion about the return value of `simplify_warp_funcion_3d` reported in #1213.

### DIFF
--- a/dipy/align/vector_fields.pyx
+++ b/dipy/align/vector_fields.pyx
@@ -1369,6 +1369,28 @@ def simplify_warp_function_2d(floating[:, :, :] d,
     (1) T[i] = W * d[U * i] + V * i
 
     Where U = affine_idx_in, V = affine_idx_out, W = affine_disp.
+
+    Parameters
+    ----------
+    d : array, shape (R', C', 2)
+        the non-linear part of the transformation (displacement field)
+    affine_idx_in : array, shape (3, 3)
+        the matrix U in eq. (1) above
+    affine_idx_out : array, shape (3, 3)
+        the matrix V in eq. (1) above
+    affine_disp : array, shape (3, 3)
+        the matrix W in eq. (1) above
+    out_shape : array, shape (2,)
+        the number of rows and columns of the sampling grid
+
+    Returns
+    -------
+    out : array, shape = out_shape
+        the deformation field `out` associated with `T` in eq. (1) such that:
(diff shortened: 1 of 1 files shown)


arokem:
@omarocegueda : maybe it would help if you reordered the docstring to conform with the numpy docstring standard? That would entail moving this long note to a "Notes" section at the bottom of the, and clarifying the "Returns" section a bit.

omarocegueda:
Sounds good @arokem ! I'll do that =)

dipy/align/vector_fields.pyx
```diff
@@ -1475,16 +1479,39 @@ def simplify_warp_function_3d(floating[:, :, :, :] d,
...
+    affine_disp : array, shape (4, 4)
+        the matrix W in eq. (1) above
+    out_shape : array, shape (3,)
+        the number of slices, rows and columns of the sampling grid
+
+    Returns
+    -------
+    out : array, shape = out_shape
+        the deformation field `out` associated with `T` in eq. (1) such that:
+        T[i] = i + out[i]
```
omarocegueda: This is the actual updated note. The rest is just moving lengthy comments to the `Notes` section.

dipy/align/vector_fields.pyx
```diff
@@ -1475,16 +1479,39 @@ def simplify_warp_function_3d(floating[:, :, :, :] d,
...
-    an affine transformation and voxel-to-space transforms associated to
+    Modifies the given deformation field by incorporating into it
+    an affine transformation and voxel-to-space transforms associated with
     the discretization of its domain and codomain.
+
     The resulting transformation may be regarded as operating on the
     image spaces given by the domain and codomain discretization.
     More precisely, the resulting transform is of the form:
 
     (1) T[i] = W * d[U * i] + V * i
```
omarocegueda: I think this explanation should not be moved to `Notes` because it is useful to explain the `Parameters` below.

omarocegueda:
I just moved the lengthy comments to `Notes` and clarified the `output`. I think the original explanation up to eq. (1) should remain there and not moved to the `Notes` because it makes it easier to explain the `Parameters`

arokem:
Looks good to me. Hopefully this will help unwary users :-)

PR #1220, opened 2017-04-13
RF: Replaces 1997 definitions of tensor geometric params with 1999 definitions.

arokem:
@RafaelNH just pointed out to me that Westin et al. changed their definition of planarity, linearity and sphericity in a paper that came out in MICCAI in 1999, and that some of the subsequent literature used these definitions, rather than the ones we have been using. 

Compare this paper:

http://lmi.bwh.harvard.edu/papers/pdfs/1999/westinMICCAI99.pdf

Instead of this paper:

http://lmi.bwh.harvard.edu/papers/papers/westinISMRM97.html

The main feature of the new definitions is that the three measures sum to 1 in every voxel. 

It seems that this is the version that people should use, but this is a change from the previous definitions in our software, and users will get *different results* with this change. I am not sure how to handle that. Any thoughts?

### DIFF
--- a/dipy/reconst/dti.py
+++ b/dipy/reconst/dti.py
@@ -517,17 +517,19 @@ def linearity(evals, axis=-1):
 
     .. math::
 
-        Linearity = \frac{\lambda_1-\lambda_2}{\lambda_1+\lambda_2+\lambda_3}
+        Linearity = \frac{\lambda_1-\lambda_2}{\lambda_1}
 
     Notes
     -----
-    [1] Westin C.-F., Peled S., Gubjartsson H., Kikinis R., Jolesz F.,
-        "Geometrical diffusion measures for MRI from tensor basis analysis" in
-        Proc. 5th Annual ISMRM, 1997.
+    .. [1] Westin CF, Maier SE, Khidhir B, Everett P, Jolesz FA, Kikinis R
+           (1999). Image Processing for Diffusion Tensor Magnetic Resonance
+           Imaging. In Proceedings of Second Int. Conf. on Medical Image
+           Computing and Computer-assisted Interventions (MICCAI’99),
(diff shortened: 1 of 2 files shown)


etpeterson:
Nothing revolutionary in this suggestion, but I'd say support both and default to the more common one, I guess the 99 version. Probably do the typical slow rollout (implement as non-default, warn it will change, change to default and warn it changed, remove warning) over a year or so.

Also, it looks like he claims both versions sum to 1, though I haven't thought that through.

arokem:
Thanks for the suggestion @etpeterson! I have now implemented a `version` key-word argument. The default behavior entails no change from the current `master`, but users can now select to use the "1999" version. 

And to address this issue: as I discovered by running this test on both: no, the older version of these does not sum to 1. :shrug:

Garyfallidis:
Can we e-mail Westin about this?

arokem:
My guess is that their position would be that the latter paper supersedes the earlier paper, but asking is a good idea. Even if that is the answer, we need to consider how we transition to using the newer metrics, or possibly continue allowing use of both formulations of this.

Garyfallidis:
I talked with Westin about this in ISMRM and he told me that he will get back to me. So, even for him it was not a direct choice for 97 or 99.

arokem:
Thanks for checking. I also emailed him about this, but no response yet.

skoudoro:
Hi @arokem, did you get a feedback from Westin? 

What is the way to go for this PR?

Garyfallidis:
When I spoke with Westin some time ago he told me that he is fine with what we have. We can close this PR.

arokem:
+1 as long as we’re clear about what we’re doing ( I think we are)

skoudoro:
ok, closing

PR #1221, opened 2017-04-13
NF: Check multi b-value

arokem:
Some models require that you have more than one (non-zero) b-value in your gradient table. 

This new feature implements a simple function that checks a gradient table, to see that it has enough different b-values.


RafaelNH:
This looks good! I will merge this - it is only one added function and it is already well tested! I will add this into dki and fwdti!

Issue #1169, opened 2017-01-13
Time for a new release - scipy 0.18?


matthew-brett:
I just set of a wheel build, and noticed that the current release of dipy errors out with the current release of scipy:

```
======================================================================
ERROR: Failure: ImportError (cannot import name nanmean)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/venv/local/lib/python2.7/site-packages/nose/loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "/venv/local/lib/python2.7/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/venv/local/lib/python2.7/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/venv/local/lib/python2.7/site-packages/dipy/reconst/sfm.py", line 25, in <module>
    from scipy.stats import nanmean
ImportError: cannot import name nanmean
```

https://travis-ci.org/MacPython/dipy-wheels/jobs/191741388#L413

Time for a new release?

arokem:
Past time for a release, for sure.

If we want to handle different versions, we could (should?) do what we did here: 

https://github.com/nipy/dipy/pull/1166

jchoude:
Also, the interpolation code in dipy.tracking.interpolation errors out when using Numpy 1.12, because the Deprecation Warning was finally changed to an Error (see https://docs.scipy.org/doc/numpy/release.html#deprecationwarning-to-error)

Do we have a buildbot using the 1.12 release yet?

jchoude:
Also, my previous message could also be linked with https://github.com/nipy/dipy/issues/728

arokem:
Could this be related? http://nipy.bic.berkeley.edu/builders/dipy-bdist32-27/builds/401/steps/shell_12/logs/stdio

jchoude:
No, you would get an error like:

```
File "dipy/tracking/local/interpolation.pyx", line 116, in dipy.tracking.local.interpolation.nearestneighbor_interpolate (dipy/tracking/local/interpolation.c:2847)
    return data[index]

IndexError: only integers, slices (`:`), ellipsis (`...`), numpy.newaxis (`None`) and integer or boolean arrays are valid indices
```

Casting the index to an int fixes the issue. I'm testing the code right now.

arokem:
[...]

PR #1223, opened 2017-04-14
local PCA using SVD

arokem:
This extends #1108 and supersedes it, addressing some of the issue raised in the review of that previous PR.

1. Following a comment by @samuelstjean (https://github.com/nipy/dipy/pull/1108/files#r81179260), I implemented the PCA using the SVD, instead of decomposing the covariance matrix. This is slower by about 50%, when done using scipy.linalg.svd, but as suggested there, using the direct lapack API performs as fast as the covariance matrix decomposition approach. I am worried that this will not work with older versions of scipy, but let's see what Travis says.

2. I fixed a bunch of other things, typos, structure of the code overall (the whole thing was wrapped in a bit `if...else` structure for error handling, which I moved to the top).


arokem:
As predicted, older versions of Scipy don't have the lapack API (https://travis-ci.org/nipy/dipy/jobs/222247543#L2671), so we might want to abandon this, or do this conditional on newer versions of scipy. Anyone have any strong opinions about that? 

I still have a couple more things to clean up here. I'd like to go over the example once over, and we still need to deal with the issue described in this comment: https://github.com/nipy/dipy/pull/1108/files#r81159634

samuelstjean:
You definitely want to use the svd based version, as the covariance based version will see its condition number grow by the square or the svd based one (or rather, the covariance one will be unstable by a factor O(n2) while the svd one is only O(n), roughly). And that will most likely lead to weird corner issues where the eigenvalues are badly estimated and the whole thing will get cut off.

You might want to do a thin svd wrapper which returns either the lapack svd or the scipy svd depending on the version found for that.

samuelstjean:
Don't have time to check in more details, but I'll be back around mid may, so that leaves time to figure stuff out. Did not recheck the whole logic as the old version was, well,, not really working, but it could be due to the covariance based thing and it just did not kick in on simple phantoms I guess.

Anyway, it's also missing the signal bias inversion formula (which is 3-4 lines, so it should be doable), but if you want more realistic phantoms and data to throw at it just ask. Better fix it now than 3-4 months later because Maxime and I find problems in it, much easier to fix it at the source directly than revert it afterwards after all.

arokem:
I believe this PR is almost done. The one thing I still want to do is to add a check for ill-conditioned patch-size to number-of-directions inputs, when the number of voxels in the patch is smaller than the number of directions in the data (see also discussion here: https://github.com/nipy/dipy/pull/1108#discussion_r81159634). 

But if someone wants to start taking a look at this, please feel free to do so. You can also see the changes relative to #1108 here: https://github.com/riddhishb/dipy/pull/1/files (but it's a bit of a tohuwabohu, because @riddhishb hasn't rebased in a few months...)

Garyfallidis:
@villalonreina and all can you test this brunch and tell us if it works for you?

arokem:
OK - error handling added. This one is ready to go from my point of view. Please review!

riddhishb:
@arokem I am going to give this a try, I dont see any apparent problem with the implementation,

RafaelNH:
@riddhishb - I think you close this by mistake! I will reopen it!

riddhishb:
@RafaelNH Oops sorry

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,137 @@
...
+                # Also update the estimate matrix which is X_est * theta
+
+                theta[i - patch_radius: i + patch_radius + 1,
+                      j - patch_radius: j + patch_radius + 1,
+                      k - patch_radius: k + patch_radius + 1,
+                      :] = theta[i - patch_radius: i + patch_radius + 1,
+                                 j - patch_radius: j + patch_radius + 1,
+                                 k - patch_radius: k + patch_radius + 1,
+                                 :] + 1.0 / (1.0 + np.linalg.norm(d,
+                                                                  ord=0))
```
RafaelNH: Not sure if this is correct!
weights have to be equal to 1/(1 + nd) where nd is the number of non-zero eigenvalues.
np.linalg.norm is doing the sum of non-zero eigenvalues instead. In the original article the used the norm of matrix d, however matrix d is previously converted to zeros and ones.

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,137 @@
...
+
+                thetax[i - patch_radius: i + patch_radius + 1,
+                       j - patch_radius: j + patch_radius + 1,
+                       k - patch_radius: k + patch_radius + 1,
+                       :] = thetax[i - patch_radius: i + patch_radius + 1,
+                                   j - patch_radius: j + patch_radius + 1,
+                                   k - patch_radius: k + patch_radius + 1,
+                                   :] + temp / (1 + np.linalg.norm(d,
+                                                                   ord=0))
```
RafaelNH: Same here

dipy/denoise/pca_noise_estimate.pyx
```diff
@@ -0,0 +1,150 @@
...
+    ----------
+    .. [1] Manjon JV, Coupe P, Concha L, Buades A, Collins DL "Diffusion
+           Weighted Image Denoising Using Overcomplete Local PCA". PLoS ONE
+           8(9): e73021. doi:10.1371/journal.pone.0073021.
+    """
+    # first identify the number of the b0 images
+    K = gtab.b0s_mask[gtab.b0s_mask].size
+
+    if(K > 1):
+        # If multiple b0 values then
```
RafaelNH: Add inline comment #MUBE Noise estimate

dipy/denoise/pca_noise_estimate.pyx
```diff
@@ -0,0 +1,150 @@
...
+        data0 = data[..., gtab.b0s_mask]
+
+    else:
+        # if only one b0 value then
+        # SIBE Noise Estimate
+        data0 = data[..., ~gtab.b0s_mask]
+        # MUBE Noise Estimate
```
RafaelNH: remove this inline comment

dipy/denoise/pca_noise_estimate.pyx
```diff
@@ -0,0 +1,150 @@
...
+        cnp.npy_intp  patch_size = 2 * pr + 1
+        double norm = patch_size ** 3
+        double sum_reg, temp1
+        double[:, :, :] I = np.zeros((n0, n1, n2))
+        double[:, :, :] count = np.zeros((n0, n1, n2))
+        double[:, :, :] mean = np.zeros((n0, n1, n2))
+        double[:, :, :] sigma_sq = np.zeros((n0, n1, n2))
+        double[:, :, :, :] data0temp = data0
+
+    X = data0.reshape(nsamples, n3)
```
RafaelNH: I think this is incorrect!
For Mube, X should have as many columns as image voxels and as many rows as images.
For Sibe, X should be reconstructed from the patches. Note: number of voxels should always the larger than number of images.

dipy/denoise/pca_noise_estimate.pyx
```diff
@@ -0,0 +1,150 @@
...
+    # Project into the data space
+    V = X.dot(W)
+
+    # Grab the column corresponding to the smallest eigen-vector/-value:
+    I = V[:, -1].reshape(data0.shape[0], data0.shape[1], data0.shape[2])
+
+    with nogil:
+        for i in range(pr, n0 - pr):
+            for j in range(pr, n1 - pr):
+                for k in range(pr, n2 - pr):
```
RafaelNH: This loop should only be done for SIBE

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,137 @@
...
+from warnings import warn
+
+# Try to get the SVD through direct API to lapack:
+try:
+    from scipy.linalg.lapack import sgesvd as svd
+    svd_args = [1, 0]
+# If you have an older version of scipy, we fall back
+# on the standard scipy SVD API:
+except ImportError:
+    from scipy.linalg import svd
```
arokem: That's correct. The direct interface to Lapack was introduced in Scipy 0.12, released in 2013.

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,137 @@
...
+    svd_args = [False]
+
+
+def localpca(arr, sigma, patch_radius=1):
+    r"""Local PCA-based denoising of diffusion datasets.
+
+    Parameters
+    ----------
+    arr : 4D array
+        Array of data to be denoised.
```
arokem: Do they really have to be *different*?

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,137 @@
...
+                # 2. Alternatively, calculate the eigenvalues and
+                #  eigenvectors of the covariance matrix through an SVD:
+                U, S, Vt = svd(X, *svd_args)[:3]
+                # Items in S are the eigenvalues, but in ascending order
+                # We invert the order (=> descending), square and normalize
+                # \lambda_i = s_i^2 / n
+                d = S[::-1] ** 2 / X.shape[0]
+                # Rows of Vt are eigenvectors, but also in ascending eigenvalue
+                # order:
+                W = Vt[::-1].T
```
arokem: Yes. For a nice explanation on the relationship, see also: https://stats.stackexchange.com/a/134283

arokem:
Thanks for the review! I believe that I have addressed all your comments. Please do take a look again. In particular, the noise estimation code changed quite a bit, especially for MUBE. Now, the noise estimate for MUBE is simply the PC of the entire data that is associated with the lowest eigenvalue. Does that make sense?

arokem:
I should say: the PC associated with the smallest eigenvalue, *corrected for Rician noise*

dipy/denoise/localpca.py
RafaelNH: You are right! They don't need to be different, what I meant was that the last dimension should contain the diffusion-weighted signals.

RafaelNH:
Looks good! This is ready for my side!

samuelstjean:
Lost my login again, but anyway, does it work on real low snr data? Other
version had an overblurring problem by a large margin. Can maybe give it a
try tonight.

2017-05-15 10:02 GMT+02:00 Rafael Neto Henriques <notifications@github.com>:

arokem:
I believe that a student in @Garyfallidis's lab was going to take a look at this with a few data-sets (sorry, I don't know Saber's Github user-name to tag him here), so we should wait to see what he finds, before we merge this.

samuelstjean:
Well, I sent an email and it did not pop ove,r or this will be a duplicate message. Anyway,

---------------

Well, it still seems to overblur stuff again unfortunately, if the attached picture works you'll see what I mean.
capture d ecran de 2017-05-15 20-12-16


Also, the noise estimation is a bit memory hungry, it was using a full 6.5 gb of ram while the lpca itself was using 4 gb of ram (up to 4.6 near the end, might be leaking stuff). And it jumped over my 8 gb of ram at the end until it swapped, so some operation could reuse arrays to prevent that, as everything is working on double.

While on the subject, the array is not cast to double for the noise estimation part and will just crash in cython instead, so converting it in the upper function would fix that. This might also be the case elsewhere in case you want to double check.
~~~python
In [13]:
    ...: t = time()
    ...: sigma = pca_noise_estimate(data, gtab, correct_bias=True, smooth=3)
    ...: print("Sigma estimation time", time() - t)
    ...:
/home/samuel/anaconda3/bin/ipython:3: VisibleDeprecationWarning: boolean index did not match indexed array along dimension 3; dimension is 65 but corresponding boolean dimension is 64
  import sys
<ipython-input-13-c25ce03dd25d> in <module>()
      1
      2 t = time()
----> 3 sigma = pca_noise_estimate(data, gtab, correct_bias=True, smooth=3)
      4 print("Sigma estimation time", time() - t)

/home/samuel/Bureau/dipy/dipy/denoise/pca_noise_estimate.pyx in dipy.denoise.pca_noise_estimate.pca_noise_estimate (dipy/denoise/pca_noise_estimate.c:3054)()
     89         double[:, :, :] mean = np.zeros((n0, n1, n2))
     90         double[:, :, :] sigma_sq = np.zeros((n0, n1, n2))
---> 91         double[:, :, :, :] data0temp = data0
     92
     93     X = data0.reshape(nsamples, n3)

ValueError: Buffer dtype mismatch, expected 'double' but got 'short'
~~~
You can also somehow get negative values after the thresholding and reconstruction, so it might be unstable somewhere or at the very least you'll want to clip those out.

In [33]: denoised_arr.min()
Out[33]: -37.144656821091232

It also misses eq. 5 for the bias inversion, as this takes care of the variance problem, you also have this part to account for (although they use a numerical inverse for rician noise, I wonder if we can track the analytic version of that somehow). Anyway, that's just a quick recheck before someone goes ahead and merges it, still some small stuff left and right to take care of.

arokem:
Could you share the data you used for that? I am not sure what I am supposed to look at to diagnose "overblurring". Is there some way to quantify it? Or is it one of these subjective "it doesn't look right to me" things?

samuelstjean:
Kind of both, that's the data (the 1.2 mm one)

https://github.com/samuelstjean/nlsam_data/tree/master/invivo

Figure 6D is what it should look like with the original (which I might still have somewhere leftover, not sure, it takes like 20 mins on my old university desktop to run the original lpca)

https://arxiv.org/pdf/1606.07239.pdf

I also have fancier phantoms, but they looked 'ok', although the noise estimation was a bit lower than the real value I used, could just be a limitation of the estimator as the original lpca does not have a separate function for it.

samuelstjean:
Turns out I found it, filenames are the one from the example + the original lpca implementation.

https://www.dropbox.com/sh/dhskq6kki7m3wgk/AAA3jDjeqiAHlyqBkv50q5hPa?dl=0

samuelstjean:
Looking at residuals = abs(original - denoised) should not show structure in principle if it works as intended. I just added them to the folder, seems like the last volume in the reimplementation is off by one or something, the image is way off in term of apparent structure when compared to the other ones.

The residuals will show you visually the variance part that got removed, but your eyes/viewer will trick you for the noise bias part. For that looking at difference maps from diffusion metrics is more apparent.

I'll probably be off for the night also after that (I noticed the notification while looking at something else), so hopefully stuff uploads and it has everything needed.

arokem:
Thanks for the data, and for the tips on debugging. I will poke around when
I get a chance.

I'm disinclined to compare the results of our implementation to another
implementation (especially one I cannot inspect). I'd rather focus on
getting it right.

Garyfallidis:
@ssheybani any comments?

samuelstjean:
They have it available https://drive.google.com/file/d/0B9aYHyqVxr04aEpobURaZFhNTmM/edit

It's just that the results are clearly off, all the data is washed out in this one, while the original has much sharper boundary. It also has constant patches of same valued data while this one has values which are constantly changing, so it might just be missing something on the final aggregation of patches. Worst case you can maybe ask for the code.

arokem:
Are you sharing that code with the original author's permission? Either
way, it's p files and compiled mex files, so it's not much help.

It's published work - we should be able to replicate it from the paper...

Since I might be taking leave any minute now, I suggest that we postpone
merging this until after @ssheybani has had a chance to take a close look.
Either way, I think that we should cut a release before that.

See #1169. Any person trying to install Dipy right now with our
instructions would have a complicated experience, because the released
version is not supporting newer versions of Numpy/Scipy. So, we should
focus on bugs that need to be fixed, rather than new features.

samuelstjean:
It's the link from this page at the top https://sites.google.com/site/pierrickcoupe/softwares/denoising-for-medical-imaging/dwi-denoising/dwi-denoising-software

It was mostly so you can double check on other datasets what is happening, but there seems to be a off by one index somewhere as the residuals from the last volume are way off when compared to all of the others. The m-code version is private though, but asking nicely might get you to it. I was asked to not share it, and as I helped test the pre-release version it might not be up to date anymore anyway.

ssheybani:
Hi,
I applied the code to some Ferret data that we have received from another group, and here's the result:
denoisingsvd-r3

arokem:
Thanks for running that. I think this looks good. 

Could you please post an image of the residuals (i.e., the difference between your top two images)? Do you see anything unusual about the last image (see comments above about that)? About the residuals of the last image?

RafaelNH:
@ssheybani  - what is the acquisition scheme that you are using? Can you also show us some images of other gradient directions or b-values? If the algorithm is excluding too many PCA components, it might visually produce nice figure but remove the important diffusion-weighted information (i.e. increasing the similarity of images with different gradient directions)! Indeed plotting the residuals for different diffusion-weighted images will be a nice way to see if there is any loss of diffusion-weighted anatomical contrast.  

Apart from that, the FA maps seem to be quite improved! I don't get why you are saying that the bottom of the denoised image looks weird. Do you think that we have some information lost in that region? We can use FA residual maps to inspect that :)

ssheybani:
@arokem @RafaelNH 
The acquisition is single-shell DtiEpi with 200 gradient directions. Here are the residuals ( abs(difference) ):
residuals_abs

And here are the images for multiple directions. Note that the images are in z-plane, so only consider the first two elements of the gradient vector.
multipledirs_abs

I think the images for the different directions look fairly different. What do you think?

About the last image, I think Samuel is right and it does look over smoothed if the intensity range is correct. However, it definitely looks very different from the one in the paper (p. 13, fig. 6d). I'll try to replicate it as soon as possible.

ssheybani:
I ran the LPCA on this file as well:
https://github.com/samuelstjean/nlsam_data/blob/master/invivo/dwi_1_2.nii.gz

The result is even better than the Ferret's:

denoise_human

FA (The mask is not perfect!):
fa_human

riddhishb:
@ssheybani I think this was the data I wanted to test on, and the results looks pretty good, and as far as I remember this is similar to what I got when I made the comparison in my GSOC.
here are some of the results https://riddhishbgsoc2016.wordpress.com/2016/07/07/tying-up-loose-ends/

samuelstjean:
small fyi, there is some ghosting in your topmost ferret image in the horizontal direction somehow. could be just appearing due to the intensity you used to display stuff (there is always some small ghosting anyway) or you might need some acquisition tweaking for the next time.

Anyway,
+ These is no option for accepting a mask. That would dramatically speed up the whole thing if say anything inside at least 1 voxel of the mask was computed, then remasked at the reconstruction step at the end. The negative values where also in the back ground in my example.
+ Anyone has access to super high b-value data? Or we can just make a phantom. Seems like there is a small discrepancy, and it starts to really kick in a low SNR. The data I put over on dropbox was just 40 directions though, so it could be also on a lower number of direction (although the original paper is using a 21 directions dataset, so this case should also work in the reimplementation).

RafaelNH:
Hello guys! From the figures that you are sending me, the algorithm seems to be already working quite well. From the residual of the Ferret's data (particularly for the different diffusion gradient directions), it seems that some structural information is being lost (and that is why denoising data can also look a bit smooth). However, this should be due to the noise sigma estimate. Since here we are using Manjon's empirical signal principal component classification, it is normal that this will work better for some datasets. I think the only way to overcome this is using more robust component classification as proposed by [Veraart et al., (2016)](http://www.sciencedirect.com/science/article/pii/S1053811916303949). This will adapt better for different datasets. I have already implemented this approach in the past, I was just waiting to have this merged so that I can add this in the top of the work done here. 

This PR seems to contain all what [Manjón et al., 2013](http://journals.plos.org/plosone/article?id=10.1371/journal.pone.0073021) proposed. In my point of view, this is ready to be merged!

samuelstjean:
On quite the contrary, I'd say it definitely does not work. Can't attach
pictures to email and still can't remember my password, but anyway, for the
data I mentioned, there is a discrepancy with the original. See the
attached png screenshot in the dropbox link I posted earlier, it's the
residuals for a random volume and the b0 and a comparison of the first
volume between both implementation.

It still should be easier to find the problem and fix it now instead of
having people complaining two years down the line about different behavior
between the original and the reimplementation (true story, that's about how
long it took to find out why the CSD was having ringing artefact, turns out
the positivity constraints were left out by mistake but the cholesky
decomposition enforced them on the side).

2017-05-19 10:36 GMT+02:00 Rafael Neto Henriques <notifications@github.com>:

arokem:
On Fri, May 19, 2017 at 4:55 AM, Samuel St-Jean <notifications@github.com>
wrote:

> On quite the contrary, I'd say it definitely does not work. Can't attach
> pictures to email and still can't remember my password, but anyway, for the
> data I mentioned, there is a discrepancy with the original. See the
> attached png screenshot in the dropbox link I posted earlier, it's the
> residuals for a random volume and the b0 and a comparison of the first
> volume between both implementation.
>
>
Sigh. That's not a lot your giving us to go on there. I tend to believe
@ssheybani's examples are pretty informative -- and I don't see anything
clearly wrong there, except what Rafael mentioned. You'll have to be a bit
more clear about what you think is wrong, if you want us to understand your
objections. You seem to think something is wrong, and you have a hunch that
it's an off-by-one error, but that's all I got (oh yeah, and you've lost
your password, that too ;-D).

@ssheybani: could you try running this on the data you used, with a lower
estimated sigma? That is, multiply the sigma estimate by something like 0.7
and run it through? Does that change the residuals you posted?


> It still should be easier to find the problem and fix it now instead of
>

I disagree about that. I think that having more people to use this would
make us more likely to find errors, not less likely. That is to say that
Dipy, *like any other software* probably does contain errors. Our advantage
over much of the scientific software out there is that we have clear
mechanisms to self-correct: our code is open to inspection, and we welcome
corrections to our code from anyone. We have a test suite and a development
process that tries to make sure that once we discover an error in the code,
that error will not reoccur.

That's not to say that I think that we should be oblivious to obvious
errors in our implementation, but I don't see anything like this here. The
PR has been read and commented on by multiple people, including yourself,
and we seem to have reached the point where we mostly think that the
implementation represents the published paper well.

If after discussing this more, we are still worried that this doesn't work
as expected, we might want to hold off and merge this into master after we
cut a release. We can keep banging on this after it's merged into master,
and like I said, more people using this will actually help us track down
errors, not vice versa.

having people complaining two years down the line about different behavior

samuelstjean:
I'll have to disagree once again on every point, but hey at least my laptop has my login saved so I can put pictures. People keep suggesting me to just write the passwords on post-its instead, but whatever.

Anyway, on the left is the absolute difference of the original and this reimplementation, and on the right the ratio abs(original-reimplementation)/original.

difference

For the region I selected, there is a rather large difference, which is around 5 times higher. It may just starts to kick in on rather low SNR dataset, so it might still be there but less apparent on the ferret data. So yeah, I am pretty sure there is a problem somewhere, it could also be in the noise estimation part. In any case, lpca is a data driven push button algorithm, there should be no need for variance value tweaking if the estimation works correctly.

I also think it's far easier to fix it now than later (just recall the problems that @mdesco  reported about nlmeans and the new block version, I was fairly certain there was an issue, took us 3 days to get a solid case and everything about it, and it still took about one month to get stuff running because nobody seemed to believe us. If it has reached a release, people would have started using it and either found it not working, or worse not realized there was na issue as it was working before, therefore breaking everything else they do afterwards.) As such, it is easier to fix it now than have it reach mainstream, especially that a few number of persons can catch things like that, rather than have everyone use it and not realize it does not fully replicate the original.

Unfortunately, more people using it does not mean more people finding bugs, as you need to understand how it works and what is it supposed to produce in the first place. Painful example of this for mainstream software would be say heartbleed, shellshock or even the afni problems and the PNAS paper which revealed that an important bug screwed everybody for a decade.

So, yes it is nice and all that people want to reimplement stuff, but since we are lucky enough to have access to the original version, might as well make sure it checks against it to get it right the first time before someone else starts using it and finds a mistake to minimize backlash and tears.

For those wondering what I was talking about : 
http://heartbleed.com/
https://en.wikipedia.org/wiki/Shellshock_(software_bug)
http://www.pnas.org/content/113/28/7900.full

Yes, you can not have perfect stuff everytime, but we can at least find and fix the problems we see. Oh and it is still missing eq. 5 (I think, the one that inverts the rician bias for the voxel values, not for the variance), so hey, that could be just that also. 

I'd rather prefer an analytic version of it than the numerical one they suggest, but that would require tracking down a bunch of formulas as I did not find a version specifically for the rician case anywhere so far, if nobody understood that part I guess I could have a shot at it (but ahem, the guy who did it in the first place probably understands the code more than I do).

Oh yeah for anyone wondering, the files are here along with a notebook to save the volumes I posted screenshots of https://www.dropbox.com/sh/dhskq6kki7m3wgk/AAA3jDjeqiAHlyqBkv50q5hPa?dl=0

Peace out

RafaelNH:
@samuelstjean - the issues that you are pointing at your data is something that I was not expecting! Just to double check, are you using the latest version of the code?

Anyway, I would love to further help you all on this PR! However, I will be on leave for some weeks to finish my thesis writing up! I will come back to this PR, if this is still open!

samuelstjean:
There have been 3 commits afterwards, but they don't affect this as they mostly add some tests, it was mostly workarounds for other small bugs I noted down.

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,123 @@
...
+                    patch_size, patch_size, patch_size, arr.shape[3])
+                # Also update the estimate matrix which is X_est * theta
+
+                theta[i - patch_radius: i + patch_radius + 1,
+                      j - patch_radius: j + patch_radius + 1,
+                      k - patch_radius: k + patch_radius + 1,
+                      :] = theta[i - patch_radius: i + patch_radius + 1,
+                                 j - patch_radius: j + patch_radius + 1,
+                                 k - patch_radius: k + patch_radius + 1,
+                                 :] + 1.0 / (1.0 + np.linalg.norm(d,
```
arokem: Are you absolutely sure you are looking at the most recent version of the code? This line, for example, was changed as you suggest in  https://github.com/nipy/dipy/pull/1223/commits/9c5de321b297803f2a8702cf91cd5df82454b6fa

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,111 @@
...
+
+    if isinstance(sigma, np.ndarray):
+        if not sigma.shape == arr.shape[:-1]:
+            e_s = "You provided a sigma array with a shape {0} for data with "
+            e_s = "shape {1}. Please provide a sigma array that matches the "
+            e_s = "spatial dimensions of the data."
+            e_s.format(sigma.shape, arr.shape)
+            raise ValueError(e_s)
+
+    tau = np.ones(arr.shape) * 2.3 * 2.3 * sigma * sigma
```
samuelstjean: This line give me a crash. For starter,s the error message is not appended, but replaced at each line. The result of using the example as-is is to give you a 3D standard deviation array, but the function accepts a 4D one. Therefore, it is missing a broadcasting (and noise correction) along the last axis.

~~~python

In [8]: 
   ...: t = time()
   ...: sigma = pca_noise_estimate(data, gtab, correct_bias=True, smooth=3)
   ...: print("Sigma estimation time", time() - t)
   ...: 

Sigma estimation time 34.688045024871826

In [9]: 

In [9]: 
   ...: 
   ...: t = time()
   ...: 
   ...: denoised_arr = localpca(data, sigma=sigma, patch_radius=2)
   ...: 
   ...: print("Time taken for local PCA (slow)", -t + time())
   ...: 

<ipython-input-9-9945762433a7> in <module>()
      3 t = time()
      4 
----> 5 denoised_arr = localpca(data, sigma=sigma, patch_radius=2)
      6 
      7 print("Time taken for local PCA (slow)", -t + time())

/home/samuel/Bureau/dipy/dipy/denoise/localpca.py in localpca(arr, sigma, patch_radius)
     62             raise ValueError(e_s)
     63 
---> 64     tau = np.ones(arr.shape) * 2.3 * 2.3 * sigma * sigma
     65 
     66     # declare arrays for theta and thetax

ValueError: operands could not be broadcast together with shapes (192,192,96,41) (192,192,96) 
~~~

I am currently making sure it still gives erroneous results with the latest changes.

arokem:
Yeah - I just ran into that one too. Mysterious that it is not triggering
any of the tests...

samuelstjean:
Seems like it checks for a 3D array, then tries to broadcast against a 4D one, and that won't ever work. I guess no tests feeds the result of the first into the second function, so ahem, it at least warrant another pass of the code, I just had a quick look at it before ismrm as I mentioned long ago, and now raphael is busy too, so yeah, I'm not even sure one person with sufficient knowledge in this subject did a full pass over the code after all. At the same time making sure it does work with the two other noise estimation function would be nice, as everything can be plugged into and replaced easily for more flexibility (including 4D noise array, I do have a use case for that personally).

Tests are good for breakage due to not your fault stuff, but passing tests are in no way a guarantee that it works, just that is it not plain broken (and god knows I did pull a few quick ones on the self test submission in my first computer science class). Anyway, corner cases are always hard to figure out and fall into the 90%/10% rule according to my data structure class teacher.

dipy/denoise/pca_noise_estimate.pyx
```diff
@@ -0,0 +1,144 @@
...
+
+    sigma_sq = np.divide(sigma_sq, count)
+
+    # find the SNR and make the correction for bias due to Rician noise:
+    if correct_bias:
+      mean = np.divide(mean, count)
+      snr = np.zeros_like(I).astype(np.float64)
+      snr = np.divide(mean, np.sqrt(sigma_sq))
+      snr_sq = snr ** 2
+      zeta = (2 + snr_sq - (np.pi / 8) * np.exp(-snr_sq / 2) *
```
arokem: This is equation 8 of the paper, which uses the greek letter zeta: 

screen shot 2017-05-21 at 11 13 15 am

samuelstjean: I always mix these two, bring back bad memories. Anyway, turns out it is xi, but that's not super major important https://en.wikipedia.org/wiki/Xi_(letter)

arokem:
I've started implementing the bias correction in the last couple of commits. 

Thanks @samuelstjean for the hints on avoiding overflow from your notebook. With a bit of log-taking, I can push us up to about SNR=55 (see the test for a comparison with the 'naive' approach). 

I propose that we move ahead with the release without this, because it might take me a while to finish the work I started here (I expect to go on parental leave any minute...).

Comments would be welcome, in the meanwhile. In particular, @samuelstjean : can you please take a look at the bias correction? Why is this not working as expected? I must be missing something simple.

arokem:
By the way, I ran our implementation over two different data-sets, and here are the volume-by-volume residuals: https://gist.github.com/arokem/890f001afa6d93c983949102ce4a6640

As you can see, there does indeed seem to be something different about the last volume of the low-snr data-set (the 1.2 mm data that @samuelstjean shared), but I don't think it's because it's the last volume. I think it's because that's the b=0 volume. The residuals in the first 10 volumes of the Stanford dataset (I know this data is not great for this, but I am just playing around here) are also slightly more "structured" than the . 

@samuelstjean : could you please post volume-by-volume residuals that you get when running the original LPCA? That file you posted only has one volume (what does that correspond to?). Thanks!

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,185 @@
...
+
+    Returns
+    -------
+    inv_eta : float
+        The inverse of the eta function
+    """
+    phi_sq = phi ** 2
+    inv_eta1 = np.log(np.sqrt(np.pi/2)) - phi_sq / 2
+    inv_eta2 = 2 * np.log((((1 + (phi_sq / 2)) * sps.iv(0, phi_sq/4)) +
+                          ((phi_sq / 2) * sps.iv(1, (phi_sq/4)))))
```
samuelstjean: I was also using ive, which scales the bessel function by an exponential so it does not overflow, but then of course the exponential scaling back does overflow. I was about to wrote down the logs on a piece of paper to get them right, but you might want to look into it as you already got then written down.

Worse case SNR 50 something is the same as no correction at this point, as over 30 was doing nothing, and in pratice people don't bother changing it from 10. From my experience, it is negligible around SNR 30 already though.

dipy/denoise/pca_noise_estimate.pyx
```diff
@@ -0,0 +1,157 @@
...
+
+    sigma_sq = np.divide(sigma_sq, count)
+
+    # find the SNR and make the correction for bias due to Rician noise:
+    if correct_bias:
+      mean = np.divide(mean, count)
+      snr = np.divide(mean, np.sqrt(sigma_sq))
+      snr_sq = (snr ** 2)
+      # xi is practically equal to 1 above 37.4, and we overflow, raising
+      # warnings and creating ot-a-numbers.
```
samuelstjean: missing an n to not a number

samuelstjean:
The original residuals were the sum of absolute difference I salvaged from a super old backup that somehow survived a few round of cleanups on my personal external drive. So I computed a new per volume residual which is under the terrible generic name residuals_original.nii.gz

That's the thing with the dataset I posted, very low SNR (on purpose, due to very fine voxels), 40 directions. I don't think it problems will show up with other regular everyday dataset as they have usually fairly high SNR in comparison. Same thing with the stanford one, at 10 b0s an 150 directions, it should be fairly easy to process and therefore show little if any difference. However, having a small difference like this will most certainly cause problems at high b-value, where the signal is already very low and where the advantage of denoising really start to shine with regards to being close to the noise floor and having your signal values all over the place.

As r the residuals, yeah I think I was definitely wrong, it is the b0, and the structure is due to the CSF, under our definition of SNR = signal/noise_std, it of course has a crazy high SNR and is never touched by the algorithm, hence stuff seems off to our eyes since the rest of the wm/gm will light up in the residuals.

I can have a look at the inversion formula next weekend, I'll try checking out if it behave similarly to the one I have been using.

arokem:
Thanks for posting those residuals. Just to clarify your comments from a couple of hours, it looks like there is quite a bit of structure in the residuals for the b=0 volume in this implementation as well: 

image

So, this seems to be a characteristic of these data (and the extreme differences in SNR between b=0 and diffusion-weighted volumes), rather than of our implementation. I wonder what will happen with this when @RafaelNH adds his extension of this to select eigenvalues/components in a more principled manner? Hopefully, we can make this additional structure go away, but it might be that this discrepancy in SNR will continue to bedevil us. 

How about we go ahead and merge the implementation we have, minus the bias correction? 

I will go ahead and move that to a separate branch, so we can keep banging on it after we merge this one.

arokem:
I've rebased here, to pull out the bias correction stuff, and started another PR for that work on #1249

samuelstjean:
I still need to rerun it on the newer version to check if it matches out with the original version. The difference was mostly appearing at low SNR (hence might be a noise estimation issue), not on the b0 image anyway. 

As for raphael suggestion, it just computes a more data driven threshold rather than 2.3**2, it bases itself on finding the statistical distribution of the noisy eigenvalues. However, it also requires a largish number of directions to do so (around 45 by experience) and lower than that (including this dataset), it can not properly identify the distribution. But at least it's available under mrtrix3 as dwidenoise to test out (and in C to read it). So well, if you can afford it, should work better in theory. The original lpca is a rather heuristic, but always work decently well approach in comparison.

Jelle also told me he has a sightly updated formula he is using now, taken from the companion paper below rather than his original formulation marchenko-pastur distribution fitting, supposedly faster.      

Veraart, J.; Novikov, D.S.; Christiaens, D.; Ades-aron, B.; Sijbers, J. &
     Fieremans, E. Denoising of diffusion MRI using random matrix theory.
     NeuroImage, 2016, in press, doi: 10.1016/j.neuroimage.2016.08.016

RafaelNH:
Yup @samuelstjean! That is the article that I want to incorporate in this code after merging this PR.

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,118 @@
...
+            e_s += "{0} for data with".format(sigma.shape)
+            e_s += "shape {0}. Please provide a sigma array".format(arr.shape)
+            e_s += " that matches the spatial dimensions of the data."
+            raise ValueError(e_s)
+
+    tau = np.ones(arr.shape[:-1]) * ((tau_factor * sigma) ** 2)
+    del sigma
+    # declare arrays for theta and thetax
+    theta = np.zeros(arr.shape, dtype=np.float64)
+    thetax = np.zeros(arr.shape, dtype=np.float64)
```
samuelstjean: To ease up on memory, since these arrays only store data and not intermediate results, they could be float32 and the computations kept up in float64. Currently running the 1.2mm and it is using 3.8 gb of ram, still manageable but that would help cutting on the usage.

samuelstjean:
Finished running it, still overblurs, but ahem, could be entirely related to the above mistake I noted.

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,118 @@
...
+
+    if isinstance(sigma, np.ndarray):
+        if not sigma.shape == arr.shape[:-1]:
+            e_s = "You provided a sigma array with a shape"
+            e_s += "{0} for data with".format(sigma.shape)
+            e_s += "shape {0}. Please provide a sigma array".format(arr.shape)
+            e_s += " that matches the spatial dimensions of the data."
+            raise ValueError(e_s)
+
+    tau = np.ones(arr.shape[:-1]) * ((tau_factor * sigma) ** 2)
```
samuelstjean: that one also can be at float32 since it just serves as a hard threshold value.

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,118 @@
...
+                W = Vt[::-1].T
+                # Threshold by tau:
+                W[:, d < tau[i, j, k]] = 0
+                # This is equations 1 and 2 in Manjon 2013:
+                Xest = X.dot(W).dot(W.T) + M
+                Xest = Xest.reshape(patch_size,
+                                    patch_size,
+                                    patch_size, arr.shape[-1])
+                # This is equation 3 in Manjon 2013:
+                theta[ix1:ix2, jx1:jx2, kx1:kx2] += 1.0 / (1.0 + np.sum(d > 0))
```
samuelstjean: Even better, you don't need temp arrays here which are the full 4D size of the dataset, you can just compute it at each location and store that directly to denoised_arr using the same indexing system.

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,118 @@
...
+                U, S, Vt = svd(X, *svd_args)[:3]
+                # Items in S are the eigenvalues, but in ascending order
+                # We invert the order (=> descending), square and normalize
+                # \lambda_i = s_i^2 / n
+                d = S[::-1] ** 2 / X.shape[0]
+                # Rows of Vt are eigenvectors, but also in ascending eigenvalue
+                # order:
+                W = Vt[::-1].T
+                # Threshold by tau:
+                W[:, d < tau[i, j, k]] = 0
```
samuelstjean: That tau value is just a pointwise estimate, it would be more robust to use the mean or median of the full 3D/4D tau array instead.

samuelstjean:
I think I'll have to go trough the whole thing once again after some small stuff gets fixed, it seems like there are just lots of small inconsistencies which makes it hard to find the problem when they are all stacked up together. So hmm, not ready for prime time yet I'd say.

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,118 @@
...
+                # This is equations 1 and 2 in Manjon 2013:
+                Xest = X.dot(W).dot(W.T) + M
+                Xest = Xest.reshape(patch_size,
+                                    patch_size,
+                                    patch_size, arr.shape[-1])
+                # This is equation 3 in Manjon 2013:
+                theta[ix1:ix2, jx1:jx2, kx1:kx2] += 1.0 / (1.0 + np.sum(d > 0))
+                thetax[ix1:ix2, jx1:jx2, kx1:kx2] += Xest / (1 + np.sum(d > 0))
+
+    denoised_arr = thetax / theta
```
arokem: No, I think that you are reading this incorrectly. 

1. The summation is implemented as `+=` in each of these (was that a problem?)

2. The multiplication with theta is implemented through the division with the denominator on line 115 (it's like multiplying by `(1/(1+sum(d>0)`). The division with sum(theta) is implemented in the division on line 117.

dipy/denoise/localpca.py
arokem: I just took the median over the entire thing, following your suggestion here https://github.com/nipy/dipy/pull/1223#pullrequestreview-41927096

dipy/denoise/localpca.py
arokem: I don't see how we would do this with the summation. Leaving this as is, for the time being. 

It shouldn't change the result anyway.

arokem:
Since this still overblurs, I'd love to see the residuals that you are getting with this. If the residuals are uniform and "white", then there is no overblurring, right? 

Or I might just not really understand what overblurring means. As far as I understand overblurring would arise when the residuals have some structure in them.

dipy/denoise/localpca.py
samuelstjean: Eq. 3 says (for each location)

theta = 1  / (1 + D_0)

output = sum(theta * x) / sum(theta)

but here it does 

output = sum ( x / theta) / theta due to line 115 

So line 115 is currently a division, not a multiplication. Eq. 3 says x = stuff, theta = stuff, so only two thetas come into play, one at the numerator and one at the denominator, but it currently does.

output = thetax / theta = x / theta**2 

Or I don't see what you mean with a multiplication on line 115.

dipy/denoise/localpca.py
samuelstjean: Indeed, but stashing only say D and recomputing it at the end would save you a full double 4D array of ram. It does a difference when you are playing with large datasets like the hcp.

arokem:
Thanks. I clarified the arithmetic a bit more, just to make sure.

I think it's still exactly mathematically identical to what we had here
before, but I think that now it resembles equation 3 in the article more
closely.

dipy/denoise/localpca.py
arokem: Honestly, I really just don't see how to do this. Since this doesn't change the functionality, I suggest that you do another PR with this improvement once we get this one merged (or you can do another PR against this branch right now, if you want).

Garyfallidis:
@arokem and @ssheybani is this ready for you to be merged? Or should we not included in this release?

arokem:
Ready to go from my point of view.

ssheybani:
I also think it's ready.

samuelstjean:
I've look at the residuals, a little bit off but it should mostly be numerical error/difference in noise estimation implementation I'd guess now. The per volume residuals is mostly higher on partial volume gm/wm voxels, no idea why.

 It is still missing some performance stuff (such as using a mask, not storing intermediate theta arrays but putting them directly in the result, etc.) but that can be added on the same time as the bias correction part. 

Did not have time to have a relook though, initial noise estimation could also probably be made more efficient in the next version of the followup pr.

picture taken form the notebook in that old dropbox link if anyone wanted to check out residuals stuff.

capture d ecran de 2017-06-14 08-05-29

And the absolute sum of them in picture, intensities may seem high but it's the sum over all volume after all.

capture d ecran de 2017-06-14 08-12-39

Oh yeah return type should be left as float32 as instead of original dtype, since sometime you might read a nifti file stored as say uint16, and there was some small negative values near the background. Those would loop around and that creates huge artifacts for later, especially that there is no masking here now.

Garyfallidis:
Going always to float64 maybe too much given that the input of this algorithm is usually large DWI datasets. Think of HCP data for example. If you go to 64 bit precision you are in trouble. Float32 maybe a safer bet here. But if the input data are float 64 then they should stay float 64. So basically, what we need is to move to float32 only with unsigned ints of 8, 16, 32 bits. 

**Or what would be easier is to add a new parameter suggesting output datatype. This is much better and give us control over the output of the algorithm.**

So, my suggestion is to add the parameter ```dtype=None`` which returns the same dtype as output and then the user can change that to 'f8' or anything else only when needed. Do we agree?

Apart from this. It seems this is ready and will merge after this small change is done.

samuelstjean:
All the intermediate arrays should be float64 for the computation of the svd, but the temporary ones and output can be float32, as it only changes the allowed precision. It's also cleaner to get out a float array than na integer one at the end, since there will be a small truncation induced. It's probably nothing, but it also costs nothing to keep it.

And the intermediary arrays theta/thetax can be declared as float 32 also instead. As for pushing the computation in the loop instead of storing them, you might need to do some denominator trickery so it's probably not worth the trouble.

Could also be useful to have a mask option, it takes 5x time longer than the original one (around 5 mins vs ~20 something here). Yes, the original one odes not use a mask, but it would help bring computation time down easily at least.

arokem:
Unless I am missing something (I might be -- I am rather sleep-deprived
:-D), since this really only applies to the last line of the function
(whether to cast the output or not), I think that we don't need a "dtype"
keyword argument. If users want to change the dtype of the output, they can
do that when they get their result back from the function.

How's this?

samuelstjean:
Well ding it inside saves you half the memory, whenever you exit the function a lot of stuff will get killed anyway, it's for when you are computing stuff that it stresses ram, hence why it is useful.

arokem:
By golly, looks like you'r right! 

https://gist.github.com/arokem/1475c223d9dcd43626dec5b22cf11783

One of you computer scientists will have to explain this to me one day... 

In the meanwhile, I will implement @Garyfallidis' proposal

samuelstjean:
Luckily it's something they teach you in your first year classes, so it makes it easy to explain with only having a few classes.

So I always mix bits and bytes first, but whatever, so many people around here who can actually correct me properly.

So, the dtype 16,32 and 64 refers to how many bytes you assign to store each number, with obviously 16 taking 16 bytes and 64 using 4x thta at 64 bytes.

In term of integer number, the largest one you can have is 2**(number_of_bytes-1), so for int32, that is 
In [3]: 2**31
Out[3]: 2147483648

One byte is for the sign, so that makes it for uint
In [2]: 2**32
Out[2]: 4294967296

And of course using more or less bytes lets you represent numbers, anything outside that range just wraps around (hence why uint is not a good choice fr numerical purpose, any negative and you end up wiht the largest possible number)

Float are different, they use instead the bytes to store number x 10**number_of_bytes (roughly), so you cna represent the same numbers, unline int dtypes, but their *precision* will change.

That's why you always check for absolute difference around 10**-7, you cannot go below that anyway. See https://en.wikipedia.org/wiki/Floating-point_arithmetic#Internal_representation for actually correct information

So in short, int -> bytes refer to the largest number you can represent
float -> precision of the those number, with a much greater range available

So to do svd, you definitely want float 64 to not run into weird numerical and accuracy bugs, but to store regular numbers like the signal output, float32 is perfectly fine. and you save half the memory doing so.

P.S. I always mix up bytes and bits (and it's even worse in french, they only change by capital and small letter), so reading the same text with those words switched might make more sense.

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,134 @@
...
+
+    References
+    ----------
+    .. [Manjon13] Manjon JV, Coupe P, Concha L, Buades A, Collins DL (2013)
+                  Diffusion Weighted Image Denoising Using Overcomplete Local
+                  PCA. PLoS ONE 8(9): e73021.
+                  https://doi.org/10.1371/journal.pone.0073021
+    """
+    if out_dtype is None:
+        out_dtype = arr.dtype
```
samuelstjean: you will still need to convert uint dtype to int dtype, to prevent negative wrap around

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,134 @@
+import numpy as np
+import scipy as sp
+import scipy.linalg as sla
+from warnings import warn
+
+# Try to get the SVD through direct API to lapack:
+try:
+    from scipy.linalg.lapack import sgesvd as svd
```
samuelstjean: actually it's useless to use float64 here, fortran only works in float32 (or at least returns that) but the older version does not (it might actually convert it without telling you, since they only bypass wrapper stuff)

~~~python
from scipy.linalg.lapack import sgesvd as svd
a=np.ones((10,100),np.float64)
u,s,v,info = svd(a)
print(u.dtype)
print(s.dtype)
print(v.dtype)

# older version
from scipy.linalg import svd
u,s,v = svd(a)
print(u.dtype)
print(s.dtype)
print(v.dtype)

## -- End pasted text --
float32
float32
float32
float64
float64
float64
~~~

arokem:
That was a great explanation. But not of the thing I was asking about 😉 

The thing I was wondering about is why casting to float32 and then returning seems to require more memory than returning the array and then casting to float32. Why would it matter whether the casting is done inside the function or out of it?

samuelstjean:
Because when you do it inside, you preassign the array to float32 and put stuff inside.
If not, it will default to float64 and use that much memory. Upon exiting the function, it will flushes out stuff (like all those big arrays of x, xestimated, theta and thetax)and return a float64. Yes, if you convert it yourself afterwards, it uses less memory, but only at the end.

While inside the function, more arrays are needed for storage, and that is when it is dangerous to hit the swap with large datasets. But yes the end result should be pretty much the same, it's the inbetween that had us worried. So far it always over around 7go on my 8go laptop (and everything else is closed), that's pretty much borderline on the 1.2 mm dataset, which is not that big even compared to say hcp datasets.

arokem:
I still don't understand how Python could know that the array would eventually be cast to a float32, before that happens. First the array is calculated in float64, and then it is cast into float32, no? I didn't think that Python knew how to look forward to anticipate this kind of type change, and optimize for it! 

More importantly: here's my proposal for what to do with unsigned integers: if you require an unsigned integer output, we set all the negative values in your array to 0, and give you back that array in the same unsigned int dtype. What do you think about that?

samuelstjean:
It doesn't only the small portion of the svd is made on float64, if it was the full 4D array we would be screwed. Because you move on small locations at a time, a small array is in float64, and that portion is stashed to float32 in a preallocated stockage array.

As for the negative, they are just small residual errors, might as well always clip it instead. Unless you know what you are doing, everyone is using magnitude images and it is physically impossible to have values. 

Even better, use array.clip(min=0, out=array) to not cast it to memory. A side project had me keeping around ~30 gb of data in memory, so I has to pretty much always resort to such trick to not kill the whole thing away.

arokem:
This is now coming back green (on my fork: https://travis-ci.org/arokem/dipy/builds/245181737). @Garyfallidis : could you please take one more look, and merge this if it's ready?

Garyfallidis:
Sure, waiting for Travis.

arokem:
You'll be waiting for a while. The build for this commit hasn't even
started yet for the nipy/dipy fork.

Garyfallidis:
Sure, okay its 12:15 am here. Will merge it tomorrow morning.

arokem:
It's all green now! 💚

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,136 @@
...
+
+    if isinstance(sigma, np.ndarray):
+        if not sigma.shape == arr.shape[:-1]:
+            e_s = "You provided a sigma array with a shape"
+            e_s += "{0} for data with".format(sigma.shape)
+            e_s += "shape {0}. Please provide a sigma array".format(arr.shape)
+            e_s += " that matches the spatial dimensions of the data."
+            raise ValueError(e_s)
+
+    tau = np.median(np.ones(arr.shape[:-1]) * ((tau_factor * sigma) ** 2))
```
ssheybani: @arokem Sorry, I don't understand why you are using the median of Sigma, instead of the whole array. I couldn't find it in the paper. Isn't it more accurate to use the whole array, which has a specific value for each voxel?

arokem: Good point. The paper does state that they are using "local estimates of the noise variance" for sigma/tau.

But following this comment: https://github.com/nipy/dipy/pull/1223#pullrequestreview-41927096

I made the change to a median instead: https://github.com/nipy/dipy/pull/1223#discussion_r121295629

Using one estimate for the entire array saves memory. I take the median, rather than the mean to avoid sensitivity to outliers in the sigma array (which may arise around edges and such).

PR #1225, opened 2017-04-17
Add one more multi b-value data-set

arokem:
This one is smaller than the CENIR data-set, and is not preprocessed.


RafaelNH:
Can you please add the reference of this data? I will be great that this reference is displayed when fetching the data!

RafaelNH:
Yup, this seem to work - I was able to fetch the data to my PC! I will merge this after the reference is added!

RafaelNH:
Cool! I will merge this right away!

PR #1227, opened 2017-04-19
BF: Use np.npy_intp instead of assuming long for ArraySequence attributes

MarcCote:
A small patch addressing the issue @arokem raised in comment https://github.com/nipy/dipy/pull/1076#issuecomment-274866713

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,7 @@ matrix:
     - python: 2.7
       env:
         # Check these values against requirements.txt and dipy/info.py
-        - DEPENDS="cython==0.18 numpy==1.7.1 scipy==0.9.0 nibabel==2.1.0"
+        - DEPENDS="cython==0.25.1 numpy==1.7.1 scipy==0.9.0 nibabel==2.1.0"
     - python: 2.7
       env:
         - DEPENDS="$DEPENDS scikit_learn tables"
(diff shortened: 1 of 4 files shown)


matthew-brett:
I think you'll need to bump up the minimum Cython version from 0.18 to 0.25.1 - see requirements.txt, .travis.yml, dipy/info.py ...

matthew-brett:
Unrelated test failure for pre-release wheels addressed by #1229.

matthew-brett:
I get the int / long testing error reliably when using cython 0.18, 0.21, 0.24, but not 0.25.1.  All of these compile OK, the error only appears at run-time.  But why not try Cython==0.21 as the minimum dependency in `.travis.yml` and see?  I think you'll find you need Cython==0.25.1 as the minimum dependency.

MarcCote:
@matthew-brett yes, you are right about needing Cython >= 0.25.1.

matthew-brett:
Don't forget `dipy/info.py` and `.travis.yml` for the Cython dependency.

MarcCote:
@matthew-brett thanks. @arokem this PR is ready.

arokem:
LGTM. Unless someone has more comments, I can merge this tomorrow.

matthew-brett:
Looks good to me too.

PR #1228, opened 2017-04-19
RF: Use cython imports instead of relying on extern

MarcCote:
Small refactoring as suggested by @matthew-brett (see comment https://github.com/nipy/dipy/pull/1076#issuecomment-294845381)

### DIFF
--- a/dipy/segment/clustering_algorithms.pyx
+++ b/dipy/segment/clustering_algorithms.pyx
@@ -9,13 +9,6 @@ from metricspeed cimport Metric
 from clusteringspeed cimport ClustersCentroid, Centroid, QuickBundles
 from dipy.segment.clustering import ClusterMapCentroid, ClusterCentroid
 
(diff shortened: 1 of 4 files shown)


matthew-brett:
Finishing up stdlib import removals at https://github.com/MarcCote/dipy/pull/10 .

MarcCote:
Thanks.

arokem:
Darn. Should've probably asked you to rebase on master before merging. Looks like this broke the master branch: https://travis-ci.org/nipy/dipy/jobs/224600749. Any idea how these things interact?

arokem:
You're right. This is also relatively new, so might just be the precision wasn't set appropriately: https://github.com/nipy/dipy/pull/1231

To answer your question: yes, there is some randomness in that test: https://github.com/nipy/dipy/blob/master/dipy/reconst/tests/test_dki_micro.py#L88-L89

PR #1229, opened 2017-04-19
RF: allow for scipy pre-release deprecations

matthew-brett:
factorial, factorial2 should come from scipy.special.

See https://s3.amazonaws.com/archive.travis-ci.org/jobs/223388982/log.txt for
current failure on --pre testing, where warnings overcome the travis-ci log
limit.


matthew-brett:
Coverage failure is because we only test on one scipy version - I think that's unavoidable.

PR #1233, opened 2017-04-25
Feature: Added environment override for dipy_home variable

tomwright01:
The fetcher module defines a path based from the users home directory. This can cause a problem when running in a cluster environment where the home directory can be machine specific.

This PR enables an environment variable DIPY_HOME to override the default setting.

### DIFF
--- a/dipy/data/fetcher.py
+++ b/dipy/data/fetcher.py
@@ -23,7 +23,10 @@ else:
 
 
 # Set a user-writeable file-system location to put files:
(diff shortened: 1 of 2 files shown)


arokem:
This looks fine to me. I've heard people complain that we are not using and XDG-compliant directory for the data, but as long as we are going to do this, I am fine with adding this additional twist. Anyone else have any comments about this?

MarcCote:
Good idea. LGTM

PR #1234, opened 2017-04-26
BUG: Fix non-ASCII characters in reconst_dki.py example.

jhlegarreta:
Non-ASCII characters were introduced in the reconst_dki.py example;
since no encoding was specified in that file, a syntax error was being
reported by Python.


arokem:
Whoops. Thanks!

PR #1231, opened 2017-04-23
TST: Reduce precision requirement for test of tortuosity estimation.

arokem:

PR #1235, opened 2017-04-30
New hdf5 file format for saving PeaksAndMetrics objects

Garyfallidis:
Hi @matthieudumont and all this PR is resolving the issue with saving the peaks object. Making it an HDF5 file allows to be read by other languages etc.

Mr. Dumont you will need to update your workflow PRs to use this one and then we can move to merging them asap.

Do you agree with the file extension .pam or .pam5 (as in hdf5)?

Also have in mind that the helper peaks_to_niftis is not in this PR so you will have to use it from your other PR. Let me know how it goes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ addons:
 
 env:
     global:
-        - DEPENDS="cython numpy scipy matplotlib h5py nibabel cvxopt tables"
+        - DEPENDS="cython numpy scipy matplotlib h5py nibabel cvxopt"
         - VENV_ARGS="--python=python"
         - INSTALL_TYPE="setup"
         - EXTRA_WHEELS="https://5cf40426d9f06eb7461d-6fe47d9331aba7cd62fc36c7196769e4.ssl.cf2.rackcdn.com"
@@ -42,6 +42,7 @@ matrix:
         - DEPENDS="$DEPENDS scikit_learn tables"
     - python: 3.5
       env:
+        - COVERAGE=1
         - DEPENDS="$DEPENDS scikit_learn tables"
     # To test vtk functionality
     - python: 2.7
@@ -53,6 +54,8 @@ matrix:
         - LIBGL_ALWAYS_INDIRECT=y
         - VENV_ARGS="--system-site-packages --python=/usr/bin/python2.7"
         - TEST_WITH_XVFB=true
+        - DEPENDS="$DEPENDS scikit_learn tables"
+
     - python: 2.7
       env:
         - INSTALL_TYPE=sdist
--- a/.travis.yml
+++ b/.travis.yml
@@ -53,6 +53,7 @@ matrix:
         - LIBGL_ALWAYS_INDIRECT=y
         - VENV_ARGS="--system-site-packages --python=/usr/bin/python2.7"
         - TEST_WITH_XVFB=true
+        - DEPENDS="$DEPENDS tables"
     - python: 2.7
       env:
         - INSTALL_TYPE=sdist
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ addons:
 
 env:
     global:
-        - DEPENDS="cython numpy scipy matplotlib h5py nibabel cvxopt"
+        - DEPENDS="cython numpy scipy matplotlib h5py nibabel cvxopt tables"
         - VENV_ARGS="--python=python"
         - INSTALL_TYPE="setup"
         - EXTRA_WHEELS="https://5cf40426d9f06eb7461d-6fe47d9331aba7cd62fc36c7196769e4.ssl.cf2.rackcdn.com"
@@ -53,7 +53,6 @@ matrix:
         - LIBGL_ALWAYS_INDIRECT=y
         - VENV_ARGS="--system-site-packages --python=/usr/bin/python2.7"
         - TEST_WITH_XVFB=true
-        - DEPENDS="$DEPENDS tables"
     - python: 2.7
       env:
         - INSTALL_TYPE=sdist
--- a/dipy/io/peaks.py
+++ b/dipy/io/peaks.py
@@ -217,16 +217,16 @@ def save_peaks(fname, pam, affine=None, verbose=False):
         print('PAM5 version')
         print(version_string)
         print('Affine')
-        print(pam.affine)
+        print(affine)
         print('Dirs shape')
         print(pam.peak_dirs.shape)
         print('SH shape')
-        if pam.shm_coeff is not None:
-            print(pam.shm_coeff.shape)
+        if shm_coeff is not None:
+            print(shm_coeff.shape)
         else:
             print('None')
         print('ODF shape')
-        if pam.odf is not None:
+        if odf is not None:
             print(pam.odf.shape)
         else:
             print('None')
--- a/dipy/io/peaks.py
+++ b/dipy/io/peaks.py
@@ -44,7 +44,7 @@ def _safe_save(f, group, array, name):
 
 
 def load_peaks(fname, verbose=False):
-    """ Load PeaksAndMetrics PAM5 file (HDF5)
+    """ Load a PeaksAndMetrics HDF5 file (PAM5)
 
     Parameters
     ----------
@@ -118,9 +118,15 @@ def load_peaks(fname, verbose=False):
         print('Dirs shape')
         print(pam.peak_dirs.shape)
         print('SH shape')
-        print(pam.shm_coeff.shape)
+        if pam.shm_coeff is not None:
+            print(pam.shm_coeff.shape)
+        else:
+            print('None')
         print('ODF shape')
-        print(pam.odf.shape)
+        if pam.odf is not None:
+            print(pam.odf.shape)
+        else:
+            print('None')
         print('Total weight')
         print(pam.total_weight)
         print('Angular threshold')
@@ -132,9 +138,8 @@ def load_peaks(fname, verbose=False):
 
 
 def save_peaks(fname, pam, affine=None, verbose=False):
-    """ Save PAM5 file (HDF5) with all important attributes of object
-
-    PeaksAndMetrics
+    """ Save all important attributes of object PeaksAndMetrics in a PAM5 file
+    (HDF5).
 
     Parameters
     ----------
@@ -176,42 +181,58 @@ def save_peaks(fname, pam, affine=None, verbose=False):
     group = func_create_group(f.root, 'pam')
     version = func_create_array(f.root, 'version',
                                 [b"0.0.1"], 'PAM5 version number')
+    version_string = f.root.version[0].decode()
 
     try:
         affine = pam.affine
     except AttributeError:
         pass
 
+    try:
+        shm_coeff = pam.shm_coeff
+    except AttributeError:
+        shm_coeff = None
+
+    try:
+        odf = pam.odf
+    except AttributeError:
+        odf = None
+
     _safe_save(f, group, affine, 'affine')
     _safe_save(f, group, pam.peak_dirs, 'peak_dirs')
     _safe_save(f, group, pam.peak_values, 'peak_values')
     _safe_save(f, group, pam.peak_indices, 'peak_indices')
-    _safe_save(f, group, pam.shm_coeff, 'shm_coeff')
+    _safe_save(f, group, shm_coeff, 'shm_coeff')
     _safe_save(f, group, pam.sphere.vertices, 'sphere_vertices')
     _safe_save(f, group, pam.B, 'B')
     _safe_save(f, group, np.array([pam.total_weight]), 'total_weight')
     _safe_save(f, group, np.array([pam.ang_thr]), 'ang_thr')
     _safe_save(f, group, pam.gfa, 'gfa')
     _safe_save(f, group, pam.qa, 'qa')
-    _safe_save(f, group, pam.odf, 'odf')
+    _safe_save(f, group, odf, 'odf')
 
     f.close()
 
     if verbose:
         print('PAM5 version')
-        print(version)
+        print(version_string)
         print('Affine')
         print(pam.affine)
         print('Dirs shape')
         print(pam.peak_dirs.shape)
         print('SH shape')
-        print(pam.shm_coeff.shape)
+        if pam.shm_coeff is not None:
+            print(pam.shm_coeff.shape)
+        else:
+            print('None')
         print('ODF shape')
-        print(pam.odf.shape)
+        if pam.odf is not None:
+            print(pam.odf.shape)
+        else:
+            print('None)
         print('Total weight')
         print(pam.total_weight)
         print('Angular threshold')
         print(pam.ang_thr)
         print('Sphere vertices shape')
         print(pam.sphere.vertices.shape)
-
--- a/dipy/io/peaks.py
+++ b/dipy/io/peaks.py
@@ -69,7 +69,6 @@ def load_peaks(fname, verbose=False):
     pamh = f.root.pam
 
     version = f.root.version[0].decode()
-    print(version[0].decode())
 
     try:
         affine = pamh.affine[:]
--- a/dipy/io/peaks.py
+++ b/dipy/io/peaks.py
@@ -47,8 +47,15 @@ def _safe_save(f, group, array, name):
         ds[:] = array
 
 
-def load_peaks(fname, verbose=True):
-    """ Load PeaksAndMetrics PAM file (HDF5)
+def load_peaks(fname, verbose=False):
+    """ Load PeaksAndMetrics PAM5 file (HDF5)
+
+    Parameters
+    ----------
+    fname : string
+        Filename of PAM5 file.
+    verbose : bool
+        Print summary information about the loaded file.
     """
 
     if TABLES_LESS_3_0:
@@ -133,7 +140,15 @@ def load_peaks(fname, verbose=True):
 
 
 def save_peaks(fname, pam):
-    """ Save NPZ file with all important attributes of object PeaksAndMetrics
+    """ Save PAM5 file (HDF5) with all important attributes of object
+    PeaksAndMetrics
+
+    Parameters
+    ----------
+    fname : string
+        Filenam of PAM5 file
+    pam : PeakAndMetrics
+        Object holding peak_dirs, shm_coeffs and other attributes
     """
 
     if TABLES_LESS_3_0:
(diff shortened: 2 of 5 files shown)


doc/examples/introduction_to_basic_tracking.py
```diff
@@ -66,6 +66,7 @@
 the ODF shows significant restricted diffusion by thresholding on
 the general fractional anisotropy (GFA).
 """
+1/0
```
arokem: This probably shouldn't be here anymore 😄

doc/examples/viz_widgets.py
```diff
@@ -121,9 +121,9 @@ def text_clicked(obj, event):
 Uncomment the following lines to start the interaction.
 """
 
-# show_manager.initialize()
-# show_manager.render()
-# show_manager.start()
+show_manager.initialize()
```
arokem: These should probably be recommented

dipy/io/peaks.py
```diff
@@ -0,0 +1,167 @@
...
+        func_create_carray = f.create_carray
+
+    if array is not None:
+        atom = tables.Atom.from_dtype(array.dtype)
+        ds = func_create_carray(group, name, atom, array.shape)
+        ds[:] = array
+
+
+def load_peaks(fname, verbose=True):
+    """ Load PeaksAndMetrics PAM file (HDF5)
```
arokem: Docstring?

dipy/io/peaks.py
```diff
@@ -0,0 +1,167 @@
...
+        print('Angular threshold')
+        print(pam.ang_thr)
+        print('Sphere vertices shape')
+        print(pam.sphere.vertices.shape)
+
+    return pam
+
+
+def save_peaks(fname, pam):
+    """ Save NPZ file with all important attributes of object PeaksAndMetrics
```
arokem: Docstring

arokem:
Why not use just use the `hdf5` extension?

dipy/io/tests/test_io_peaks.py
```diff
@@ -0,0 +1,63 @@
...
+        pam.B = np.zeros((45, sphere.vertices.shape[0]))
+        pam.total_weight = 0.5
+        pam.ang_thr = 60
+        pam.gfa = np.zeros((10, 10, 10))
+        pam.qa = np.zeros((10, 10, 10, 5))
+        pam.odf = np.zeros((10, 10, 10, sphere.vertices.shape[0]))
+
+        save_peaks(fname, pam)
+        pam2 = load_peaks(fname, verbose=True)
+        npt.assert_array_equal(pam.peak_dirs, pam2.peak_dirs)
```
arokem: Why not test that all the attributes survived the round-trip?

doc/examples/introduction_to_basic_tracking.py
Garyfallidis: Sure

doc/examples/viz_widgets.py
Garyfallidis: Sure.

arokem: Please don't reply to these comments, unless there is something to discuss. I get an email every time you make a comment like this one.

Garyfallidis:
Hi @arokem. The reason why I suggest not using the extension hdf5 is that hdf5 is too generic and then .pam_hdf5 is is too long. Also, we are using hdf5 to store streamlines too in the .dpy format. This is used by quite some people who were really unpleased with trk format (for a good reason). However, .pam is already used by netpbm as an image format. So, I suggest to use .pam5 instead. I think with the clue of the 5 and some info in the docstring it will be difficult for people to miss that this is actually an hdf5 file.
Agree? Also if we use the convention ``_pam.hdf5`` we violate the extension rules. I mean the file format extension definition should be after the last dot.

arokem:
As you say, hdf5 is generic. That is also an advantage. For example, imagine that someone created an application that is a tool for visualizing the content of an hdf5 file (just an example, it can be other things). This kind of application might not work on these files for prosaic reasons, because the application wouldn't 'see' these files as hdf5, even though that's what they are.

If you insist on creating a new file format, I suggest versioning it and checking for the version. It's possible that you will want to change this in the future and you might run into cryptic incompatibilities unless you check for the version of the file. 

But I really wouldn't hurry to proliferate a new standard -- seems like something that should be done only when there is  clear necessity for it, which I don't see here.

Garyfallidis:
There is a necessity for the following reasons:
 
a) If we do use a different extension name then it is becomes easier for our visualization software to read the data. Is it clear why? The viz software will check the file extension and then directly call load_peaks (from dipy.viz). Otherwise if the file is .hdf it will have first to open the file and then see if it has the expected attributes and then see if it has loaded other datasets of .hdf files that are of different type. This makes the visualization duplicate effort with IO which I think is a bad idea.

b) We really need a file to store results between the reconstruction and tracking this has been a bottleneck for some time. Recently we started pickling the pam objects which was not possible before. But hdf5 allows us to use memory maps too and it should be faster than pickling too.

c) External software can build a similar functionality as we could potentially do in dipy.viz
There is also a function that we have implemented with Mr. Dumont that allows to extract some important arrays from PeaksAndMetrics to niftis to help with visualization tools that only read nifti files. This functionality is still on Mr. Dumont's PR.

Nonetheless I think we can both be happy if I use the name .pam_hdf5 rather than .pam5?
Adding a version is good idea too. I am happy to do that.
Let me know what you think.

arokem:
On Mon, May 1, 2017 at 1:37 PM, Eleftherios Garyfallidis <

> There is a necessity for the following reasons:
>
> a) If we do use a different extension name then it is becomes easier for
> our visualization software to read the data. Is it clear why? The viz
> software will check the file extension and then directly call load_peaks
> (from dipy.viz). Otherwise if the file is .hdf it will have first to open
> the file and then see if it has the expected attributes and then see if it
> has loaded other datasets of .hdf files that are of different type. This
> makes the visualization duplicate effort with IO which I think is a bad
> idea.
>
OK - this seems like an argument in favor of a new file format. Of course,
if users always feed these hdf5 files in, that's probably not a problem.
But I agree that having visualization software read files in and check them
wouldn't be great. Is there some way to read an hdf5 file's header for some
pertinent meta-data?

> b) We really need a file to store results between the reconstruction and
> tracking this has been a bottleneck for some time. Recently we started
> pickling the pam objects which was not possible before. But hdf5 allows us
> to use memory maps too and it should be faster than pickling too.
>
This seems irrelevant to the discussion. When you pickle the pam objects,
do you store them in a different file-format (I'm guessing not)?

> c) External software can build a similar functionality as we could
> potentially do in dipy.viz
>
It's more likely that external software would build functionality for a
well-established file format, than for our newly designed file format.

> There is also a function that we have implemented with Mr. Dumont that
> allows to extract some important arrays from PeaksAndMetrics to niftis to
> help with visualization tools that only read nifti files. This
> functionality is still on Mr. Dumont's PR.
>
This actually seems like an argument for using existing file formats. You
wouldn't design a new `nipam` file format for these, right? ;-)

> Nonetheless I think we can both be happy if I use the name .pam_hdf5
> rather than .pam5?
>
How about a double extension: `.pam.hdf5`? That way, external software will
recognize this as an hdf5 file, and our software can look for the full
`.pam.hdf5` to make sure that this is the right file.

> Adding a version is good idea too. I am happy to do that.
> Let me know what you think.
>
:+1:

Garyfallidis:
Added version and improved testing. Waiting for travis. On the name I do prefer .pam5 than .pam.hdf5

arokem:
Why do you prefer `.pam5` over `.pam.hdf5`? The latter seems to cover all the use-cases described above, while the former will fail at least for the case of software written for hdf5 files by people unaware of the `pam5` format.

Garyfallidis:
We do use the two dots for example when we say that a file has two encodings for example .tar.gz and .nii.gz but those are actually two different processing levels in both cases. In our case there is only one processing level. This is one reason why I do not like the .pam.hdf5. Pam is an already available fileformat (for images). The second reason is that people will need to be checking for two last extensions rather than one. Which is not slick. 
Now, .pam_hdf5 is too long. So a shortened version of that is .pam5 which is short and unique.
Alternatively we can use .peaks but I would keep that for when the fit is separated from the peaks_and_models. But I am fine with .peaks too if you prefer that.

Garyfallidis:
About something else. The coverage I am getting in my machine for dipy/io/peaks.py is 96% but one of the bots is reporting 9%. Similarly it reports very low coverage for dipy/io/dpy.py 24% when in my machine I am getting 90%.

arokem:
That's because the coverage machine doesn't have pytables installed:
https://travis-ci.org/nipy/dipy/jobs/229548518#L2313

Garyfallidis:
Yes, it skips the test but still gives coverage for the file. Okay let me add pytables in.

arokem:
Yes. It gives coverage for everything that it can import from dipy.

Garyfallidis:
Travis and coveralls look happy. :)

Garyfallidis:
Now trying to reach maximum possible coverage.

arokem:
Regarding file format: to be clear, all the options that are new file
format extensions that we invent for this purpose seem almost equally bad
to me. I agree that `pam5` is the least offensive, but it's still a new
file format, and I just don't see the justification to create an entirely
new file format for what is essentially an hdf5 file.

The dual extension solution lets us eat our cake and have it too, but you
don't seem to like that. I am not sure why, though. Because in other cases
the two extensions represent different parts of the processing? Arguably
that's true here as well, one part reads the HDF5 file, and the other
parses it for PAM attributes.

Garyfallidis:
What I do not like is that one needs to parse for two extensions rather than one. Also the filename size is increasing. But more importantly .pam is already reserved for another popular format. So .pam.hdf is misleading. It's like we put a picture file (.pam) to an .hdf5 file. I hope this clarifies more.

arokem:
Yes - the latter is a good point. What picture file format uses `pam`? I
found this file format that uses pam: http://filext.com/file-extension/PAM.
Seems rather obscure.

arokem:
On the other hand, `dpy` also seems to be used by others:
http://www.liutilities.com/products/winbackup/filextlibrary/files/DPY/

Garyfallidis:
It's part of the pbm suite of file formats http://netpbm.sourceforge.net/doc/pam.html

I made .dpy back in 2009. There was nothing with that extension at the time but also we were at the very early stages of development. I am happy to change that extension in the near future. I think we should but this time promote the tracking format. We didn't promote it enough so others used the same extension name.

Garyfallidis:
If we can merge this then I can go and update the workflow PRs and move ahead with release. 
All is left now is the workflows, new website and local pca right? I don't remember something else that was crucial for this release. But maybe I am wrong.

arokem:
Also these: 

https://github.com/nipy/dipy/issues/1178
https://github.com/nipy/dipy/pull/1164

.travis.yml
```diff
@@ -53,6 +53,7 @@ matrix:
         - LIBGL_ALWAYS_INDIRECT=y
         - VENV_ARGS="--system-site-packages --python=/usr/bin/python2.7"
         - TEST_WITH_XVFB=true
+        - DEPENDS="$DEPENDS tables"
```
arokem: Add this on line 20, with the rest of the dependencies, instead of here.

dipy/io/peaks.py
```diff
@@ -0,0 +1,237 @@
...
+    pam.ang_thr = pamh.ang_thr[:][0]
+    pam.gfa = pamh.gfa[:]
+    pam.qa = pamh.qa[:]
+    pam.odf = odf
+
+    f.close()
+
+    if verbose:
+        print('PAM5 version')
+        print(version)
```
arokem: I think that we should test up-front for the version, and raise an error if a file with the wrong version(s) is provided.

dipy/io/peaks.py
```diff
@@ -0,0 +1,237 @@
...
+
+
+def save_peaks(fname, pam, affine=None, verbose=False):
+    """ Save all important attributes of object PeaksAndMetrics in a PAM5 file
+    (HDF5).
+
+    Parameters
+    ----------
+    fname : string
+        Filenam of PAM5 file
```
arokem: typo: "Filenam" => "Filename". You might also want to put periods at the end of the sentences that describe parameters (here and elsewhere in these docstrings).

dipy/io/peaks.py
```diff
@@ -0,0 +1,237 @@
...
+        Print summary information about the saved file.
+    """
+
+    if os.path.splitext(fname)[1] != '.pam5':
+        raise IOError('This function saves only PAM5 (HDF5) files')
+
+    if not (hasattr(pam, 'peak_dirs') and hasattr(pam, 'peak_values') and
+            hasattr(pam, 'peak_indices')):
+
+        raise ValueError('Cannot save object without peak_dirs/values/indices')
```
arokem: This error string might be confusing. How about: "Cannot save object without peak_dirs, peak_values and peak_indices" (to make it clear that all three are required).

dipy/io/tests/test_io_peaks.py
arokem: This comment is still relevant.

doc/examples/viz_widgets.py
```diff
@@ -118,11 +118,11 @@ def text_clicked(obj, event):
...
 
 """
-Uncomment the following lines to start the interaction.
+Uncomment the line with method start() to begin the interaction.
 """
 
-# show_manager.initialize()
-# show_manager.render()
+show_manager.initialize()
+show_manager.render()
```
arokem: Do these still needs to be recommented, or is it just the line with `start` that will block?

arokem:
BTW - where does the new website stand? We still had this issue related to that: https://github.com/nipy/dipy/issues/1197, and it doesn't seem like there's much progress on that...

Garyfallidis:
All good comments. Will get back to you later tonight.

.travis.yml
Garyfallidis: sure

dipy/io/peaks.py
Garyfallidis: Agree!

dipy/io/peaks.py
Garyfallidis: Sure.

dipy/io/peaks.py
Garyfallidis: sure

arokem:
Did you know that I get emails when you make commits? That way you don't
have to say "sure" and "ok" every time you fix something. Github simply
sends me an email when you put in the commit. You can also put in a comment
in the thread when you are done addressing all the comments, asking me to
take another look. This is also helpful for me, because it minimizes the
amount of emails I get from Github. Thanks!

doc/examples/viz_widgets.py
Garyfallidis: No, it does need to be commented because it can stop the execution of some files.

Garyfallidis:
The website is moving ahead. I made an account in a server her for Bishakh to run the daily doc builds, this a temporary solution and working with future systems https://portal.futuresystems.org to have a permanent solution for building the docs.

Garyfallidis:
About the reply strategy. I am using inbox (by gmail) and all dipy commit mails are stacked together automatically. I like replying directly to a comment so that you can see that I have addressed it. That is for making the reviewers' life easier.  It's some kind of reminder in a way for keeping track of the discussion. I mean github put a reply button there for a reason. Right? It's also giving some credit to the reviewer. Now, I could start putting all the short answers to the overall comment but that would make it difficult to connect to which lines these are referring.

Garyfallidis:
Inbox (by gmail) has also some smart notification system that sends you notifications only for crucial e-mails and ignores github e-mails. Anyway, I can try to reduce my short answers and you may want to relax your notification system.

arokem:
OK -- if it helps you track what you did, that's fine. I can always have
more email (it seems...)

I just thought that you thought it helps me see what's done, so I thought I
would mention that.

Are you done with this PR? Ready for another review?

Garyfallidis:
Okay good. Now travis is complaining on dki which is irrelevant to this PR. Interesting ...

Garyfallidis:
Sure, you can continue with the review. Please do check that I correctly change the travis configuration file. Now that tables is on the top shouldn't we remove it from the DEPENDS sections?

arokem:
I think that bug should be fixed in #1231.

arokem:
Regarding pytables: the question is whether we want it installed on all the machines? I think that since it is an optional install, the answer is "no". But it seems that you want it installed at least on the machine that checks coverage. Maybe remove it from the global depends and add both scikit learn and tables to the DEPENDS variable on the one machine that checks coverage? I would make two coverage machines: one with Python 3.5, and the other with Python 2.7. The Python 3.5 machine would have pytables and scikit learn installed. Meanwhile, the Python 2.7 machine would have both of these, as well as VTK. How's that?

Garyfallidis:
Hi, I think after this release we should actually start having tables, sklearn, cvxpy and vtk as not optional.
For now, I updated .travis.yml and rebased. Waiting for travis to see what we are getting. Added coverage for 3.5 and 2.7 as suggested.

Garyfallidis:
Hi Ariel, can you look into the change I made in travis for some reason the coverage was decreased after rebasing. This seems to be unrelated to the focus of this PR. And we may want to put a not to increase coverage but move on with the merging this and updating the workflow PRs. Let me know what you think.

dipy/io/tests/test_io_peaks.py
arokem: I think this comment is still relevant?

arokem:
Yeah - I don't understand what coveralls wants. Maybe this has to do with
the addition of the Python 3 machines?

I had one tiny comment left: test all the attributes in the object for
round-tripping.  Other than that, this looks ready to go from my point of
view, so if anyone else wants to take a look, now is the time.

Garyfallidis:
I am happy to test the attributes more. But what do you mean with round-tripping? Do you mean rounding errors?

arokem:
Sorry - with "round trip", I mean from memory to file and back. That is,
create the object, save the file, read the file into a new object and then
iterate over all of its attributes, checking that they are all identical to
the ones that came from the file.

Something like this:
https://github.com/nipy/dipy/blob/master/dipy/direction/tests/test_peaks.py#L507-L512

I think you are already checking a few of the attributes, and I would like
to check all of them.

Garyfallidis:
Thanks for this tip. Good idea! Done and waiting for travis.

PR #1243, opened 2017-05-14
Better fvtk.viz error when no VTK installed

### LINKED ISSUE #1238: Cryptic fvtk.viz error when no VTK installed
I still get this error when running the tracking_quickstart example:

    AttributeError: module 'dipy.viz.fvtk' has no attribute 'ren'

Would be good if there was some more explanatory error handling fencing this.

ghoshbishakh:
fixes #1238

### DIFF
--- a/dipy/utils/tripwire.py
+++ b/dipy/utils/tripwire.py
@@ -41,9 +41,14 @@ class TripWire(object):
         ...
     TripWireError: We do not have silly_module_name
     """
+
     def __init__(self, msg):
         self._msg = msg
 
     def __getattr__(self, attr_name):
         ''' Raise informative error accessing attributes '''
+        raise TripWireError(self._msg)
+
+    def __call__(self, *args, **kwargs):
+        ''' Raise informative error while calling '''
         raise TripWireError(self._msg)
--- a/dipy/viz/fvtk.py
+++ b/dipy/viz/fvtk.py
@@ -79,6 +79,9 @@ else:
     msg = "Python VTK is not installed"
     warn(msg)
 
+    ren, have_ren, _ = optional_package('dipy.viz.window.ren',
+                                        'Python VTK is not installed')
+
 
 def dots(points, color=(1, 0, 0), opacity=1, dot_size=5):
     """ Create one or more 3d points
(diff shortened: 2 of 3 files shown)


dipy/utils/tripwire.py
```diff
@@ -41,9 +41,14 @@ class TripWire(object):
...
     def __init__(self, msg):
         self._msg = msg
 
     def __getattr__(self, attr_name):
         ''' Raise informative error accessing attributes '''
         raise TripWireError(self._msg)
+
+    def __call__(self, *args, **kwargs):
+        ''' Raise informative error while calling '''
+        raise TripWireError(self._msg)
```
arokem: Could you please write a test that exercises this code?

ghoshbishakh:
@arokem done!

dipy/viz/fvtk.py
```diff
@@ -79,6 +79,9 @@
     msg = "Python VTK is not installed"
     warn(msg)
 
+    ren, have_ren, _ = optional_package('dipy.viz.window.ren',
```
arokem: Reading this again -- do you understand why the line above this one throws a warning, instead of raising an error?

ghoshbishakh: @arokem 
```
ren, have_ren, _ = optional_package('dipy.viz.window.ren',
                                        'Python VTK is not installed')
```
returns a TripWire object into `ren`

So calling `ren` will only raise an exception not the line 82.

arokem: I mean: if we don't have VTK installed, why don't we just raise an exception in the previous line (line 80)? Why do we bother going ahead and trying to do anything?

ghoshbishakh: @arokem I am not sure about that design decision. But there must be a reason for example it says `# Allow import, but disable doctests if we don't have vtk`, although I am not sure what that means. @Garyfallidis

arokem: @garyfallidis - I don't understand this. Is there any reason line 80 doesn't raise an error? If have_vtk is False, we're bound to hit an error sooner or later, no?

arokem:
Merging. Will follow up with another PR if needed.

Issue #1178, opened 2017-02-16
Recheck indexing with nlmeans as some of it is indexing wrong positions


samuelstjean:
There are some inconsistencies, I guess they did not cause much problem with the default settings, but the indexing is incorrect for the patch comparison.

For example, this one indexes by j,i,k, and calls the corresponding function later on with i,j,k, making the indexes not match.

main loop 
https://github.com/nipy/dipy/blob/master/dipy/denoise/nlmeans_block.pyx#L414

inner calls
https://github.com/nipy/dipy/blob/master/dipy/denoise/nlmeans_block.pyx#L417
https://github.com/nipy/dipy/blob/master/dipy/denoise/nlmeans_block.pyx#L452
https://github.com/nipy/dipy/blob/master/dipy/denoise/nlmeans_block.pyx#L461

Same thing in this function, the indexing goes by j,i,k then reverts to i,j,k later in the loop
https://github.com/nipy/dipy/blob/master/dipy/denoise/nlmeans_block.pyx#L218
https://github.com/nipy/dipy/blob/master/dipy/denoise/nlmeans_block.pyx#L251

Also, the main summing loop hardcodes a value of 27, which is 3x3x3, therefore invalidating the result for a pathchsize different than the default

https://github.com/nipy/dipy/blob/master/dipy/denoise/nlmeans_block.pyx#L280

There were just small things I remember spotting quickly from last summer, but it would be worth it to do a thorough check that takes more than my 2 minutes one and fix remaining issues before it reaches mainstream.

riddhishb:
Thanks for bringing this up, in the main loop and the inner calls the indexing is based on the fact that the volume represents a cartesian grid, and in the point where the indexing goes to i,j,k are the nested loops for a local patch and hence the usual naming.

Also about the hardcoding the patch size of 27, that is only done in the local mean and local variance  functions which takes a single pass over the volume and are independent of the local PCA patch where the variable patch size is taken into account. However, I will make a PR with option to change the patch size in the local mean and variance estimate as well.

PR #1245, opened 2017-05-14
non_local_means : patch size argument for local mean and variance

riddhishb:
Added the argument which allows user to change the patch size used for computation of local mean and local variance in the non_local_means.
Addresses the issue brought up in #1178

### DIFF
--- a/dipy/denoise/nlmeans_block.pyx
+++ b/dipy/denoise/nlmeans_block.pyx
@@ -257,18 +257,19 @@ cdef double _distance(double[:, :, :] image, int x, int y, int z,
 @cython.boundscheck(False)
 @cython.wraparound(False)
 @cython.cdivision(True)
-cdef double _local_mean(double[:, :, :]ima, int x, int y, int z) nogil:
+cdef double _local_mean(double[:, :, :]ima, int x, int y, int z, int p) nogil:
     """
-    local mean of a 3x3x3 patch centered at x,y,z
+    local mean of a patch with patch radius = p centered at x,y,z
     """
     cdef int dims0 = ima.shape[0]
     cdef int dims1 = ima.shape[1]
     cdef int dims2 = ima.shape[2]
     cdef double ss = 0
+    cdef double vol = (2 * p + 1) * (2 * p + 1) * (2 * p + 1)
     cdef int px, py, pz, dx, dy, dz, nx, ny, nz
-    for px in range(x - 1, x + 2):
-        for py in range(y - 1, y + 2):
-            for pz in range(z - 1, z + 2):
+    for px in range(x - p, x + p + 1):
+        for py in range(y - p, y + p + 1):
+            for pz in range(z - p, z + p + 1):
                 px = (-px if px < 0 else (2 * dims0 -
                                           px - 1 if px >= dims0 else px))
                 py = (-py if py < 0 else (2 * dims1 -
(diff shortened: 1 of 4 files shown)


dipy/denoise/nlmeans_block.pyx
```diff
@@ -277,25 +278,25 @@ cdef double _local_mean(double[:, :, :]ima, int x, int y, int z) nogil:
...
 
-    return ss / 27.0
+    return ss / vol
 
 
 @cython.boundscheck(False)
 @cython.wraparound(False)
 @cython.cdivision(True)
-cdef double _local_variance(double[:, :, :] ima, double mean, int x, int y, int z) nogil:
+cdef double _local_variance(double[:, :, :] ima, double mean, int x, int y, int z, int p) nogil:
```
arokem: This is going over 80 characters. Please break into two lines, by adding a line-break somewhere.

dipy/denoise/nlmeans_block.pyx
```diff
@@ -277,25 +278,25 @@ cdef double _local_mean(double[:, :, :]ima, int x, int y, int z) nogil:
...
     cdef int dx, dy, dz, nx, ny, nz
-    for px in range(x - 1, x + 2):
-        for py in range(y - 1, y + 2):
-            for pz in range(z - 1, z + 2):
+    for px in range(x - p, x + p + 1):
+        for py in range(y - p, y + p + 1):
+            for pz in range(z - p, z + p + 1):
                 if ((px >= 0 and py >= 0 and pz > 0) and
                         (px < dims0 and py < dims1 and pz < dims2)):
                     ss += (ima[px, py, pz] - mean) * (ima[px, py, pz] - mean)
```
arokem: Doesn't variance have to be normalized by the volume?

dipy/denoise/nlmeans_block.pyx
riddhishb: cool!

dipy/denoise/nlmeans_block.pyx
riddhishb: Variance is normalized by the count variable which is essentially the volume, just a gimmick of the code

riddhishb:
@arokem Thats a good idea, let me add a test

dipy/denoise/nlmeans_block.pyx
```diff
@@ -346,7 +349,8 @@ cpdef upfir(double[:, :] image, double[:] h):
 @cython.boundscheck(False)
 @cython.wraparound(False)
 @cython.cdivision(True)
-def nlmeans_block(double[:, :, :]image, double[:, :, :] mask, int patch_radius, int block_radius, double h, int rician):
+def nlmeans_block(double[:, :, :]image, double[:, :, :] mask, int patch_radius,
```
Garyfallidis: indentation issue

ssheybani:
I'm comparing your nlmeans() and non_local_means(). But I'm not sure how we can use the sigma given by estimate_sigma function in noise_estimate.py to test your non_local_means(). Because non_local_means() expects a single float value for sigma, while estimate_sigma() gives an array (with as many elements as the gradient directions). 
What I've done so far is that I gave both nlmeans and non_local_means the average of the sigmas. Here's the result:
/home/saber/Courses/Semester_1/Garyfallidis_Lab/Denoising/NLMeans/Wu/nlm olv vs new.png

I also tried sigma=5 for both. Here is the result:
/home/saber/Courses/Semester_1/Garyfallidis_Lab/Denoising/NLMeans/Wu/nlm olv vs new sigma5.png

In both cases, it looks like non_local_means is oversmoothing.
By the way, for the execution time, what I observed was that for small datasets (smal portions of complete datasets), non_local_means is faster. but when the dataset is large enough, non_local_means is twice slower than nlmeans.

samuelstjean:
While we are there, the other half (and most important I'd say) part of the issue is that the outer loop does iterate over j,i,k, but it passes to the inner function the indexes in the i,j,k order, so that effectively swap in your local block the x and y coordinate. 

Since a small region is mostly homogeneous, I guess that's why you don't see the problem much, but it could explain the problem reported above as the weights will be inflated, and less stuff gets compared because of that.

arokem:
Closing. @skoudoro is going to follow up on this. We are looking to streamline things quite a bit before our 1.0.

PR #1246, opened 2017-05-18
WIP: Replace widget with ui components in example.

dmreagan:
@ranveeraggarwal, @MarcCote, and @Garyfallidis:

I'm trying to update the viz_advanced example to use the new UI, and I've run into two quick problems. 

First, after I interact with the slider, I can no longer rotate/zoom the data. If I move the panel, then I can interact with the data again. So it seems the panel behaves correctly, but the slider is stealing focus or something. 

Second, how do I tell the slider what to do when the value updates? I tried to add callbacks following the example in viz_ui, but the method used there for the button doesn't seem to work for my slider.


dmreagan:
I got the new UI slider working, but not in the way that was demonstrated in the viz_ui example. In viz_ui, a callback is added to a button using `on_left_mouse_button_drag`. That did not work for a `LineSlider2D`. Instead, I had to use `add_callback`. Why?

Garyfallidis:
@ranveeraggarwal have you seen this?

ranveeraggarwal:
@dmreagan did you run the viz_ui example? Does it give the same issue? If so, it might be a problem with VTK 7. I'll install it and confirm.

Garyfallidis:
Yes, the problem was also with viz_ui.py We used 7.1.0 not 7.0.0

ranveeraggarwal:
I didn't face such an issue with 6.3.0. We might have to look into API changes in 7.1.0. 😕

arokem:
I think this needs a rebase. There are a lot of commits here that are from PRs that have recently been merged. If you need help doing this, let us know. We can walk you through the process.

Garyfallidis:
Too much merging here. Closing this one and will reopen a new one. Coming soon.

PR #1249, opened 2017-05-23
Bias correction for local pca

arokem:
This extends #1223 to include bias correction at the end of the local PCA denoising.

Should be merged only after we are done with #1223.


Garyfallidis:
Sorry what is left to do here?

arokem:
This follows on the work in #1223. 

It adds the bias correction in equations 4,5,6 of the Manjon et al. 2013 article. As you can see, this does not currently pass the simple test of having smaller RMSE for denoised vs non-denoised data. 

But I have not had time to look into the reasons very deeply.

For the time being, I think that we should finish up and merge #1223 (is there anything left to do there? I'm frankly not sure following the recent comments from @samuelstjean -- I believe that we resolved all his concerns, but I am not entirely sure...). Once #1223 merged, I can rebase the current PR on top of master, to add just those last couple of commits. 

Does that make sense?

RafaelNH:
Yup! That is why I always argued that PCA denoising and noise bias correction should be implemented separately. Although one should argue that these two techniques are relevant to each other, noise bias correction has their own issues that should be dealt separately. I many be able to help with this PR after my thesis submission. Actually, I've already implemented Koay's analytical solution in the past, so we can use this instead of the LUT tables.

samuelstjean:
That's also the one I was talking about, from the 2009 follow up paper, but it break up for SNR > 30 in the general case. Could not find equations for the rician case, but it's probably just a matter of looking up the equation in a book about statistics, it should do the same thing as this one here according to the theory.

samuelstjean:
Oh yeah before it starts becoming overly confusing, there are two thing (well 4) things happening here :

This is the bias correction formula for the variance in the rician case (which is already part of the old pr under the name of the xi function):

capture d ecran de 2017-05-26 15-16-29

This is the same formula in the general case which raphael has written down for the mppca noise estimation part

capture d ecran de 2017-05-26 15-16-43

This is the same one, in a later 2009 paper, where instead of using the SNR=mean/std, it decouples the values in a more stabler form (than the rician counterpart at the very least).

capture d ecran de 2017-05-26 15-17-07

And this is the new formula we need to check if it works out
capture d ecran de 2017-05-26 15-17-54

Adn that in itself I suspect to be the same as this one 
capture d ecran de 2017-05-26 15-23-11

Except we would like P_rice (and therefore need to track it down in a book) instead of P_chi, which is as of now in the form of a lookup table.

That's pretty much it so far. And having all the functions neatly separated would be indeed much cleaner since you can mix and match those with the noise estimators already written down. Then it would also be possible to replace the ugly lookup table I put down long ago for the old nlmeans noise estimation with these formulas.

RafaelNH:
Hi @samuelstjean! Can you send me the link the 2009 paper? I might never have read that one!

samuelstjean:
It's a more general framework for bias correction, here we are just using step 3 of these formulas. A bit equations heavy, but it does mean it contains everything at least.

https://www.ncbi.nlm.nih.gov/pubmed/19138540

Oh yeah regarding equation 2 and 3, I said more stable, but it should have read more flexible, as it also comes into play into the aforementioned paper.

dipy/denoise/localpca.py
```diff
@@ -0,0 +1,185 @@
...
+    ----------
+    phi : float
+    """
+    if phi < 1.25:
+        return 0
+    else:
+        return f(phi)
+
+
+def localpca(arr, sigma, patch_radius=2, tau_factor=2.3, correct_bias=False):
```
samuelstjean: correct bias should be turned to correct_rician_bias, to go hand in hand with the rest of the text as that's what it fixes. It should also be true (as the original also uses it, and it would make little sense to have it true for the noise estimation but not the signal part). And it's missing a small explanation of what it does in the docstring.

samuelstjean:
I also get an indexing error while running it, looks like the steps are not plug & play. Pretty sure it's the latest version of branch arokem/localpca_slow-svd-bias-correction

~~~python
t = time()
sigma = pca_noise_estimate(data, gtab, correct_bias=True, smooth=3)
print("Sigma estimation time", time() - t)

t = time()

denoised_arr = localpca(data, sigma=sigma, correct_bias=True)

print("Time taken for local PCA (slow)", -t + time())

## -- End pasted text --

IndexError                                Traceback (most recent call last)
<ipython-input-3-9805c172b07d> in <module>()
      1 t = time()
      2 
----> 3 denoised_arr = localpca(data, sigma=sigma, correct_bias=True)
      4 
      5 print("Time taken for local PCA (slow)", -t + time())

/home/samuel/Bureau/dipy/dipy/denoise/localpca.py in localpca(arr, sigma, patch_radius, tau_factor, correct_bias)
    176         if isinstance(sigma, np.ndarray):
    177             for idx in index:
--> 178                 denoised_arr[idx] = (sigma[idx] /
    179                                      func(denoised_arr[idx]))
    180         else:

IndexError: too many indices for array
~~~

arokem:
its not done yet

arokem:
but feedback and suggesyions are welcome

arokem:
is #1223 ready to go?

samuelstjean:
It's the same thing without the signal bias correction part? I still need to re-run it to check for blurring then, I thought this was the new omnibus one.

arokem:
Sigh. We still need to come back to this.

Issue #322, opened 2014-02-04
Missing content in tracking.utils' documentation


jchoude:
When checking the page for dipy.tracking.utils documentation (http://nipy.org/dipy/reference/dipy.tracking.utils.html#module-dipy.tracking.utils), it seems that the content is broken, especially starting at the example of the affine transform.

skoudoro:
Can you close this one @jchoude  ? it seems to be ok :-)

jchoude:
@skoudoro I might not be looking in the correct place, but I still don't find the tracking.utils documentation when looking at the link in my initial issue. Is it somewhere else?

skoudoro:
My bad, you are right, it seems to be excluded here: https://github.com/nipy/dipy/blob/master/doc/tools/build_modref_templates.py#L71

all packages dipy.*.utils are not included. I will create a PR.

skoudoro:
![](https://github.trello.services/images/mini-trello-icon.png) [#322](https://trello.com/c/N9RPVik9/7-322)

skoudoro:
website updated : http://nipy.org/dipy/reference/dipy.tracking.html#module-dipy.tracking.utils

jchoude:
Thanks guys!

Issue #570, opened 2015-01-27
The documentation for `dipy.viz` is not in the API reference


arokem:
No-where to be found here: http://nipy.org/dipy/reference/index.html

skoudoro:
@arokem "dipy.viz" is defined as skip pattern. you can see on this link https://github.com/nipy/dipy/blob/master/doc/tools/build_modref_templates.py#L72

Any reason for that ? should I change it ?

arokem:
Unless someone has a compelling reason not to include this in the documentation, please do change it. You can add that in #1255

skoudoro:
![](https://github.trello.services/images/mini-trello-icon.png) [#570](https://trello.com/c/QXr0MNXu/11-570)

skoudoro:
website updated : http://nipy.org/dipy/reference/dipy.viz.html

PR #1255, opened 2017-06-13
Fix missing documentation content

skoudoro:
All dipy.*.utils (ex: dipy.tracking.utils) are missing on the documentation so this PR change the regular expression to make this files available on api reference. (see #322)

### DIFF
--- a/dipy/tracking/utils.py
+++ b/dipy/tracking/utils.py
@@ -21,7 +21,7 @@ As an example, lets take a 2d image where the affine is::
      [0., 2., 0.],
      [0., 0., 1.]]
 
-The pixels of an image with this affine would look something like:
+The pixels of an image with this affine would look something like::
 
     A------------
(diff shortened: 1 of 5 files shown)


skoudoro:
I should add a modification for resolving #570 at the same time

arokem:
Does this build the docs OK? Because full documentation building takes a while (the examples take a while to run), you can relatively quickly test this by removing most of the lines in this file: https://github.com/nipy/dipy/blob/master/doc/examples/valid_examples.txt

skoudoro:
I just tried to generate the API reference via "make api" command, and the *.rst generated was ok.

I will generate the full documentation a little bit later to be sure that everything is ok and let you know.

arokem:
This looks ready to go. I'll merge this, and if there are any follow-ups, we can take them on a separate PR.

PR #1257, opened 2017-06-16
revamp piesno docstring

samuelstjean:
I was looking at finding why it does throw some divide by zero warning, even though they do not seem to affect the outcome. Seems like maybe when a slice has a lot of high values, the median is not zero but there might be no noise also, so I need to further check it out if it is just safe to squelch it instead.

For now some rewording and I need to also re-read the example to check it still makes sense, but I had this one for a few weeks already, so there is.

### DIFF
--- a/dipy/denoise/noise_estimate.py
+++ b/dipy/denoise/noise_estimate.py
@@ -33,14 +33,16 @@ def piesno(data, N, alpha=0.01, l=100, itermax=100, eps=1e-5,
     -----------
     data : ndarray
         The magnitude signals to analyse. The last dimension must contain the
-        same realisation of the volume, such as dMRI or fMRI data.
+        same realisation of the volume under different noise condition,
+        e.g. volumes collected during dMRI or fMRI acquisitions.
 
     N : int
         The number of phase array coils of the MRI scanner.
-        If your scanner does a SENSE reconstruction, ALWAYS use N=1, as the
-        noise profile is always Rician.
+        If your scanner does a SENSE or adaptive combine reconstruction, N=1,
+        is preferred as they yield Rician noise profile.
         If your scanner does a GRAPPA reconstruction, set N as the number
-        of phase array coils.
+        of phase array coils or a slightly lower number depending
+        on the coil correlation.
 
     alpha : float
         Probabilistic estimation threshold for the gamma function.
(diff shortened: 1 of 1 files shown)


dipy/denoise/noise_estimate.py
```diff
@@ -149,10 +146,11 @@ def _piesno_3D(data, N, alpha=0.01, l=100, itermax=100, eps=1e-5,
     -----------
     data : ndarray
         The magnitude signals to analyse. The last dimension must contain the
-        same realisation of the volume, such as dMRI or fMRI data.
+        same realisation of the volume.
```
arokem: Maybe use the same wording as above? ("e.g. volumes collected during dMRI or fMRI acquisitions.")

samuelstjean:
I still want to figure out why it does stuff like that before though
~~~bash
/home/travis/miniconda/envs/test-environment/lib/python3.5/site-packages/dipy/denoise/noise_estimate.py:248: RuntimeWarning: divide by zero encountered in true_divide
  s = sum_m2 / (2 * K * sigma**2)
/home/travis/miniconda/envs/test-environment/lib/python3.5/site-packages/dipy/denoise/noise_estimate.py:248: RuntimeWarning: invalid value encountered in true_divide
  s = sum_m2 / (2 * K * sigma**2)
/home/travis/miniconda/envs/test-environment/lib/python3.5/site-packages/dipy/denoise/noise_estimate.py:249: RuntimeWarning: invalid value encountered in less_equal
  mask[...] = np.logical_and(lambda_minus <= s, s <= lambda_plus)
~~~
probably nothing as it still works, but at least somehow squelching that would be useful.

arokem:
The latter two might be because of nans or infs in the array?

samuelstjean:
That the symptom, I'm trying ot figure out why the std value goes to zero sometimes in the first place

Garyfallidis:
I am sorry but this or any other PRs from Sam will not be merged as decided and expressed clearly last year. Sam there is no reason for you making any more PRs in dipy. We discussed about this last year let's not get back to the same discussions. I am closing this one and will remove PIESNO from the documentation. You can copy PIESNO to your own github project and work on improvements there.

samuelstjean:
not sure why you would want to remove it from the doc, it works but it
could be better worded.

2017-06-19 16:52 GMT+02:00 Eleftherios Garyfallidis <
notifications@github.com>:

PR #1258, opened 2017-06-16
Fix nlmeans indexing

samuelstjean:
I though I had made an issue about it, must have been a mental note.

Anyway, the cython part of nlmeans had a few indexing error which inverted along the way the first and second dimension, so hopefully I fixed them here. That's why not all the code is changed, the main loop was using j,i,k, but passing i,j,k down, so now they should match.

I also revamped the calling file to remove parts I personally deemed unnecessary at the same time. From memory, the example also had some statements which were not truc, so I'll have a look afterwards if this here is fine.

### DIFF
--- a/dipy/denoise/nlmeans_block.pyx
+++ b/dipy/denoise/nlmeans_block.pyx
@@ -181,7 +181,6 @@ cdef void _value_block(double[:, :, :] estimate, double[:, :, :] Label, int x, i
 
 
 @cython.boundscheck(False)
-@cython.wraparound(False)
 @cython.cdivision(True)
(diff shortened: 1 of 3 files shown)


samuelstjean:
There it is after all #1178. It fixes half the stuff from the issue (as per title), the other half seems to be fixed around in #1245

samuelstjean:
erm are they fixing it in the other pr? it seemed like nobody understand what  I meant, half the indexing in the other loop works, and the one in the inner loop is different. It only has a small effect since it swaps local coordinate, but it is still wrong, that's why you only need to change half of them.

PR #1222, opened 2017-04-14
A lightweight UI for medical visualizations #5: 2D Circular Slider

ranveeraggarwal:
Continuing from #1205 and the fourth installment of #1111, this PR adds a Circular Slider element to the Viz module.

Complete blogpost here: http://ranveeraggarwal.com/blog/gsoc-2016-summary

**To Do**
- [x] Add tests.

PR #1140, opened 2016-11-02
A lightweight UI for medical visualizations #1: Button

ranveeraggarwal:
This is one of the several PRs that will be made over time to incorporate features of #1111 with tests.

PR #1262, opened 2017-06-17
VIZ: A lightweight UI for medical visualizations #6: 2D File Selector

ranveeraggarwal:
Continuing from #1222 and the sixth installment of #1111, this PR adds a File Selector element to the Viz module.

### ToDo

- [x] Fix events
- [x] Write tests
- [x] Find and fix PEP8 issues
- [x] Record new tests


ranveeraggarwal:
@MarcCote can you clone this and see if it works for you?

ranveeraggarwal:
Okay, this is really weird.

1. The `FileSelectMenu2D` panel isn't moving like it's supposed to (on click and drag) and the events on the `FileSelectMenu2D` (again, click and drag) are getting carried over to the cubes behind (except for the up and down buttons). 
2. The text actors for the file menu are unresponsive despite adding callbacks. 

I don't get why this is happening. 

@MarcCote @dmreagan any ideas here?

dmreagan:
@ranveeraggarwal I can confirm that the panel and text actors show the behavior you describe on my system as well (Windows 10, Python 3, VTK 7.1). I don't have any leads on a solution yet, though.

ranveeraggarwal:
@MarcCote @dmreagan the events have been fixed, do have a look at the file menu behavior.

MarcCote:
I've started looking at this PR. Really awesome :+1: . Here are some weird/missing behaviors I've found while playing with the example `viz_ui.py`. I'll do a more thorough review of the code later (hopefully by tomorrow).

The filedialog (this is also true for the UI Panel) needs to be clicked once before we are able to move it around. This happens with VTK 5.8 and VTK 6.0.0. With VTK 7.0.0, the clicked events are buggy as you mentioned. 

Would it be easy to support scrolling with the mouse wheel?

Is it easy to support long click on the "scroll bar" arrows?

I think in the example we should be able to see when a file is selected. Can we change the text color? Or put a rectangle of a different color underneath it? Maybe for the moment, we could just display it somewhere (console or use another Textbox). Edit: Ok, I see the coloring seems to only work with VTK > 6.2. I still think we should find a way of showing the selected file. Do you know if we can put a border around a text actor?   

It will be nice to be able to manually specify the current directory of the filedialog. Speaking of the current directory, I don't think we should modify the actual "current directory" of the program (which happens when using `os.chdir`). We should handle the current directory with an internal variable and fetch the files/folders using something similar to `glob.glob(self.current_directory)`.

ranveeraggarwal:
> The filedialog (this is also true for the UI Panel) needs to be clicked once before we are able to move it around. This happens with VTK 5.8 and VTK 6.0.0. With VTK 7.0.0, the clicked events are buggy as you mentioned.

This is indeed weird and shouldn't happen. I will look into this.

> Would it be easy to support scrolling with the mouse wheel?

Not sure. I guess we'll need to have an invisible(?) rectangle between the two arrows and implement a scroll listener on it?

> Is it easy to support long click on the "scroll bar" arrows?

Again, we'll need to look into this. We'll also need to have a variable to determine the speed of scrolling (or, increase the speed as the button is kept pressed).

> I think in the example we should be able to see when a file is selected. Can we change the text color? Or put a rectangle of a different color underneath it? Maybe for the moment, we could just display it somewhere (console or use another Textbox). Edit: Ok, I see the coloring seems to only work with VTK > 6.2. I still think we should find a way of showing the selected file. Do you know if we can put a border around a text actor?

I am currently storing it in a variable, so it should be easy to display it. Adding a border? Let me see if that can happen. Worst case, we can hide/unhide a rectangle actor under the text.

I'll also need to work a bit more on the tests, the default way won't work.

Garyfallidis:
Hi I am confirming here that there is no color change for the selected file. I am in vtk 7.1.0. Just to make sure @ranveeraggarwal does the selection works for you? I mean the selected file is returned but does the selected color change? Also, we need an X somewhere so we can exit the fileselector. Marco's idea about the scrolling sounds reasonable too. 

Can you remind me what was the difference between the file selector and the file dialogue that we created during gsoc. And why we had those separated? Was it for modularity?
http://ranveeraggarwal.com/blog/gsoc-2016-summary

ranveeraggarwal:
Yes, the selected file behavior hasn't been implemented. I'll do it in my upcoming commits. 
We separated the file selector in case we need it in something other than a file dialog. I can't think of a case off the top of my head right now, but it's good to have an API for that, IMO.

Garyfallidis:
Also I want to confirm that I can drag and move the FileSelector with no problem

ranveeraggarwal:
I think the last commit fixed it. Can you confirm the same?

Garyfallidis:
About testing this. @dmreagan has asked a question about the record/play scheme. And we need to address that issue.
https://github.com/nipy/dipy/issues/1282

Garyfallidis:
Yes, I agree that it is great to be as modular as possible.

Garyfallidis:
You said ``I think the last commit fixed it. Can you confirm the same?`` What exactly you want me to confirm?

ranveeraggarwal:
Oh I meant, confirm that the file selector moves around.

Garyfallidis:
The file selector moves around after double clicking and dragging with no problem.
Ah remember also to return the full path of a file. Now you are returning only the filename.
Also, we should think how to be able to select multiple files. But let's finish selecting one first.

ranveeraggarwal:
@MarcCote I am not sure why we need to double click to select files and to drag around a panel. Does this have something to do with the `CustomInteractorStyle` or am I using it wrong? Can you please confirm? Apparently this bug has existed since #1140.

ranveeraggarwal:
@MarcCote I have merged your PR and fixed the event recorder counts. One last fix to be done is to use `glob.glob` instead of `os.getcwd`. It's currently messing things up. I am on it.

ranveeraggarwal:
@MarcCote @dmreagan so the `FileSelectMenu2D` now uses an internal variable to keep track of the directory. This also fixes the tests. 

There is one small teeny tiny issue for testing this UI element. When I record the test on my machine, it might not necessarily work on Travis because I use MacOS and it has a different directory structure. What do I do about this?

dipy/viz/interactor.py
```diff
@@ -36,6 +36,13 @@ def abort(self):
         """ Aborts the event i.e. do not propagate it any further. """
         self._abort_flag = True
 
+    def done(self):
```
dmreagan: I suggest renaming this to something like `reset`. `Event`'s other methods are verbs.

dipy/viz/ui.py
```diff
@@ -1884,3 +1905,478 @@ def handle_events(self, actor):
...
+
+        Returns
+        -------
+        file_names: list(string)
+            List of all file names as string.
+
+        """
+        # A list of file names with extension in the current directory
+        file_names = []
+        for extension in self.extensions:
```
MarcCote: This should be tested to make sure it returns only files with the desired extensions. Thinking about it, maybe these methods (that one and `get_directory_names`) should be standalone functions in some utils module and tested separately.

dipy/viz/ui.py
```diff
@@ -1884,3 +1905,478 @@ def handle_events(self, actor):
...
+            The FileSelectMenu2D reference this text belongs to.
+
+        """
+        super(FileSelectMenuText2D, self).__init__()
+
+        self.file_name = ""
+        self.file_type = ""
+
+        self.file_select = file_select
+
```
MarcCote: Remove empty lines that don't help readability.

dipy/viz/ui.py
```diff
@@ -1884,3 +1905,478 @@ def handle_events(self, actor):
...
+        """ Builds a text actor.
+
+        Parameters
+        ----------
+        text: string
+            The initial text while building the actor.
+        position: (float, float)
+            The text position (x, y) in pixels.
+        color: (float, float, float)
+            Values must be between 0-1.
```
MarcCote: Mention it is a RGB color.

dipy/viz/ui.py
```diff
@@ -1884,3 +1905,478 @@ def handle_events(self, actor):
...
+        return text_actor
+
+    def get_actors(self):
+        """ Returns the actors that compose this UI component.
+
+        """
+        return [self.text_actor.get_actor()]
+
+    def set_attributes(self, file_name, file_type):
+        """ This function is for use by a FileSelectMenu2D to set the
```
MarcCote: PEP8: short title on the first line.

dipy/viz/tests/test_ui.py
```diff
@@ -319,6 +319,47 @@ def test_ui_disk_slider_2d(recording=False):
         event_counter.check_counts(expected)
 
 
+@npt.dec.skipif(not have_vtk or skip_it)
+@xvfb_it
+def test_ui_file_select_menu_2d(recording=False):
```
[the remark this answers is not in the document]
MarcCote: Also, we should try to increase test coverage:
https://codecov.io/gh/nipy/dipy/pull/1262/src/dipy/viz/ui.py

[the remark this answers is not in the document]
ranveeraggarwal: For this, I am thinking we could make another PR with all the tests re-recorded. Then we can have a description of all the tests. 
(We could also give this work away for the upcoming Hacktoberfest 😛)

dipy/viz/ui.py
ranveeraggarwal: Which file in the utils module should this go into? Or do I create another file?

dipy/viz/tests/test_ui.py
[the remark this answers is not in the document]
MarcCote: Sure, we can make another PR for that.

dipy/viz/ui.py
MarcCote: Hmm, I thought more about it and we don't have to test the behavior of `glob` :P. It should be fine like this.

dipy/viz/tests/test_ui.py
```diff
@@ -319,6 +319,47 @@ def test_ui_disk_slider_2d(recording=False):
...
+    filename = "test_ui_file_select_menu_2d"
+    recording_filename = pjoin(DATA_DIR, filename + ".log.gz")
+    expected_events_counts_filename = pjoin(DATA_DIR, filename + ".pkl")
+
+    file_select_menu = ui.FileSelectMenu2D(size=(500, 500),
```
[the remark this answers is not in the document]
MarcCote: Here's an example of how to use `InTemporaryDirectory`
https://github.com/nipy/nibabel/blob/master/nibabel/streamlines/tests/test_streamlines.py#L216

ranveeraggarwal:
@MarcCote I have addressed your comments.
We will fix the tests (re-record and note down) through a different PR. 
Do let me know if there's anything else to be done for this one.

dipy/viz/tests/test_ui.py
```diff
@@ -325,41 +326,47 @@ def test_ui_file_select_menu_2d(recording=False):
...
-                                      event, event_counter.count)
-        file_select_menu.add_callback(file_select_menu.buttons["down"].actor,
-                                      event, event_counter.count)
-        file_select_menu.menu.add_callback(file_select_menu.menu.panel.actor,
-                                           event, event_counter.count)
-        for text_ui in file_select_menu.text_item_list:
-            file_select_menu.add_callback(text_ui.text_actor.get_actors()[0],
+    with InTemporaryDirectory() as tmpdir:
+        for i in range(10):
+            _ = open("test" + str(i) + ".txt", 'wt').write('some text')
```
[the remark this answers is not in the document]
ranveeraggarwal: `t` is [text mode](https://docs.python.org/3/library/functions.html#open). We could do without it - we're not using the file contents to test anyway.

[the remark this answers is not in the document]
MarcCote: Oh, good to know. It is fine, then.

MarcCote:
@Garyfallidis, @dmreagan this is all good on my end. Let me know if we can go ahead with the merge.

dmreagan:
I see a few problems here:

1. If you click on a file name, then scroll up or down, the file no longer appears selected.
2. No scrollbar?
3. If you scroll down so that `../` is no longer visible, then click on a directory name to enter it, `../` will not be visible in that lower directory, so you cannot return to the parent.
4. I'd like to see an example of using the menu in `viz_ui.py`. Right now you can select a file in the menu, but it doesn't affect anything outside of itself. You should be able to select a file and assign it to some variable in the example, for instance. Then you could print the filename or show it in a text block. That would probably require `select` and `cancel` buttons like a proper dialog.

ranveeraggarwal:
@dmreagan some of these are bugs I need to work on before merging, other enhancements.

> If you click on a file name, then scroll up or down, the file no longer appears selected.

Yep, this is a persistence bug. Just got to know, working on it.

> No scrollbar?

In the future, as an enhancement maybe?

> If you scroll down so that ../ is no longer visible, then click on a directory name to enter it, ../ will not be visible in that lower directory, so you cannot return to the parent.

This is a serious bug that needs to be explored. On it.

> I'd like to see an example of using the menu in viz_ui.py. Right now you can select a file in the menu, but it doesn't affect anything outside of itself. You should be able to select a file and assign it to some variable in the example, for instance. Then you could print the filename or show it in a text block. That would probably require select and cancel buttons like a proper dialog.

This will be the next PR. Menus like save/open are very common, so we will add them in the default API.

ranveeraggarwal:
@dmreagan thanks for reporting the bugs. I have fixed 1, 3. 2 is an enhancement we can work on later and the PR for 4 is up next.

MarcCote:
Nice. While I was reviewing the changes I asked myself "why is the name of the class `FileSelectMenu2D`?" Are we planning on having 3D file dialog?

ranveeraggarwal:
@MarcCote I don't think that's happening soon. But this is just in line with all the other 2D elements --  a naming convention of sorts. This distinguishes it from the 3D elements and tells you that this is not to be used in a 3D sense.

Garyfallidis:
One of the bots is failing because the PR hasn't been rebased. Please do rebase.

ranveeraggarwal:
@Garyfallidis done!

PR #1263, opened 2017-06-18
Cast Streamline attrs to numpy ints, to avoid buffer mismatch.

arokem:
See: https://nipy.bic.berkeley.edu/builders/dipy-bdist64-27/builds/122/steps/shell_12/logs/stdio

### DIFF
--- a/dipy/tracking/streamlinespeed.pyx
+++ b/dipy/tracking/streamlinespeed.pyx
@@ -104,14 +104,14 @@ def length(streamlines):
         if streamlines.data.dtype == np.float32:
             c_arclengths_from_arraysequence[float2d](
                                     streamlines.data,
-                                    streamlines._offsets.astype(np.int),
-                                    streamlines._lengths.astype(np.int),
+                                    streamlines._offsets.astype(np.intp),
+                                    streamlines._lengths.astype(np.intp),
                                     arclengths)
         else:
             c_arclengths_from_arraysequence[double2d](
                                       streamlines.data,
-                                      streamlines._offsets.astype(np.int),
-                                      streamlines._lengths.astype(np.int),
+                                      streamlines._offsets.astype(np.intp),
+                                      streamlines._lengths.astype(np.intp),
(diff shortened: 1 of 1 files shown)


dipy/tracking/streamlinespeed.pyx
```diff
@@ -102,15 +102,18 @@ def length(streamlines):
         arclengths = np.zeros(len(streamlines), dtype=np.float64)
 
         if streamlines.data.dtype == np.float32:
-            c_arclengths_from_arraysequence[float2d](streamlines.data,
-                                                     streamlines._offsets,
```
matthew-brett: What type are the offsets, lengths normally?  I guess they should be `np.intp` type.  I guess this conversion could have a cost, most efficient might be guarantee their type on streamline object creation, then you can do `streamlines._offsets.astype(np.np_intp)` here assuming it will have a very small cost (just the type check).

arokem: Streamlines is a nibabel.streamlines.ArraySequence object! So, I guess this is a question to the nibabel developers :-)

More seriously, from: https://github.com/nipy/nibabel/blob/master/nibabel/streamlines/array_sequence.py#L77-L78 it does look like that's a np.intp! Changed that cast to intp instead.

arokem:
Seems to have resolved the original problem: https://nipy.bic.berkeley.edu/builders/dipy-bdist64-35/builds/123 (from a try_branch run with this branch).

dipy/tracking/streamlinespeed.pyx
matthew-brett: @MarcCote - any thoughts here?  I guess these streamlines are somehow of type long, but they should be of type np.intp.  Any way of making them be np.intp from the start, to avoid memory costs of the conversion?

matthew-brett:
OK - looks good to me - @MarcCote - I'm still interested whether there's a way of making sure the type is np.intp on creation.

PR #1270, opened 2017-06-20
Doc cleaning deprecation warning

skoudoro:
- Removes non_local_mean, ascm, piesno from documentation since there are not validated
- Remove deprecation warning from nlmeans


arokem:
LGTM. Merging.

Issue #10, opened 2011-06-17
data.get_data() should be consistent across datasets


matthew-brett:
`dipy.data.get_data('small_64')` returns npy files for the gradients and bvalues, but `dipy.data.get_data('small_101D')` returns text files, so they need to be loaded using different commands (`np.load` compared to `np.loadtxt`).   For consistency they should be the same, or (preferably) return data, rather than filenames.

theaverageguy:
Is this one still active @arokem ?

arokem:
I believe so.

theaverageguy:
Shall I work on the return types of both functions to fix this ? What would be the best way?

skoudoro:
fix by #1283, closing

PR #1283, opened 2017-06-28
get_data consistence

skoudoro:
The goal of this PR is to resolve an old issue #10 .

Can you have a look @matthew-brett @arokem ?


arokem:
This looks good to me.

 @matthew-brett made the comment that it would be preferable to return data, but I think that returning file-names makes sense. It's more transparent for users to see how they would adapt this to their own data, when the file-names are explicitly mentioned. 

But maybe @matthew-brett has reconsidered this since 2011? 😄

matthew-brett:
This looks better to me - thanks.  I think it's fine to return files, just as long as it's obvious from the docstrings / variable names that that is the case.  For example, how about `get_data_fnames` rather than `get_data` ?  Would that help with backwards compatibility?

skoudoro:
Explicit better than implicit (like you said @arokem ) so I agree with @matthew-brett. I prefer `get_data_fnames`than `get_data`. I made this mistake the first time because I was not expecting filenames.

@Garyfallidis, any thoughts ? 

If it is ok for all of you, I can do a new PR with this change.

arokem:
You can make that change on this PR in a follow-up commit

Garyfallidis:
``get_fnames`` or ``get_data_fnames`` is fine for me. When refactoring this function take your time it is used in many many places.

arokem:
Thinking about this a bit more, I prefer `get_fnames`. It's already in the
`dipy.data` namespace.

skoudoro:
Any thought concerning this PR?

skoudoro:
Thanks!

PR #1285, opened 2017-06-28
mapmri using cvxpy instead of cvxopt

rutgerfick:
Hi @arokem,

As you requested, this is the the mapmri model using cvxpy instead of cvxopt. It still passes all the tests with the new optimizer but I haven't tested it on anything else :-)
I've left the specific solver that it uses unspecified (so it picks one automatically) but you can set a specific optimizer in the mapmri model call if you want.

Best,
Rutger


dipy/reconst/mapmri.py
```diff
@@ -159,6 +160,9 @@ def __init__(self,
             the tissue diffusivity that is used when dti_scale_estimation is
             set to False. The default is that of typical white matter
             D=0.7e-3 _[5].
+        cvxpy_solver : cvxpy solver name
```
arokem: This is a string? 

I would change this to: 

    cvxpy_solver : str, optional
        cvxpy solver name. Optionally optimize the positivity constraint with a particular cvxpy solver. See http://www.cvxp for details. Default: ??? 

And then fill in the default choice (what is the default?).

arokem:
Thanks for doing this! Just one comment about the documentation, but otherwise looks good. 

One more thing, we need to also edit the travis configuration file, to add cvxpy to the installs. On this line: https://github.com/nipy/dipy/blob/master/.travis.yml#L20 change `cvxopt` => `cvxpy`, and I think that should do it.

arokem:
Any chance I could convince you to also take a look at the SHORE module (https://github.com/nipy/dipy/blob/master/dipy/reconst/shore.py), and give it a similar treatment? 

Thanks!

dipy/reconst/mapmri.py
rutgerfick: Adapted the documentation according to you suggestion. The default in cvxpy is simply None so it makes its own decision.

rutgerfick:
Hi @arokem : Why are you keeping the shore module? the mapmri module completely absorbs all its functionality and more.

rutgerfick:
@arokem : It looks like the travis checks are not passing the cvxpy loading part. Any ideas?

arokem:
Ah yes. This is because of this issue: https://github.com/embotech/ecos-python/issues/8

You can resolve this by first install numpy and then installing all the other dependencies. Like this: https://github.com/nipy/dipy/pull/1180/commits/ed6c0e6c3f856c3f592d2e8d47eb2c5af06d10aa

arokem:
Regarding SHORE: should we deprecate that module for now for 0.13, and remove it in the next release (0.14 or 1.0, as the case may be)? What do you think, @Garyfallidis, @maurozucchelli ?

Garyfallidis:
SHORE and MAPL are similar but not exactly the same. The basis in SHORE is spherical and in MAPL is cartesian. Also, MAPL is still new and we need more user feedback. So, I think we should keep SHORE for some time. @maurozucchelli can you help @rutgerfick with replacing cvxopt with cvxpy for SHORE?

rutgerfick:
Hi @Garyfallidis , if you feel that the SHORE module should remain for the users then I completely agree with you. However, you should know that the MAP-MRI basis using isotropic scaling (so the same scaling in x,y,z) and the SHORE basis are analytically equivalent. In fact, @maurozucchelli showed that explicitly in his PhD thesis.

Since it's no big deal I just replaced cvxpy for cvxopt for you in SHORE in my latest commit here.

Best,
Rutger

Garyfallidis:
Yes, I know there are equivalent mathematically. I worked with Gonzalo on this some years ago! :D 
Hey, super thanks for updating SHORE too. We can keep SHORE around for a bit. It feels bad to get rid all these nice formulations using the spherical basis and I know we have users that still use SHORE. But we should update our documentation to promote MAPL if that is better.

rutgerfick:
Well, if it makes you feel any better you can still get all the same spherical basis formulations but setting anisotropic_scaling=False in the mapmri module, it is exactly the same ;-)

You already know this but the only differences between the shore and mapmri modules are the improved Laplacian regularization, the availability (but not necessity) of the anisotropic basis, and the amount of features (q-space indices) that can be calculated. I think it's a good idea to update the documentation since there are only upsides from a data modeling point of view.

dipy/reconst/mapmri.py
```diff
@@ -230,12 +241,14 @@ def __init__(self,
...
+            if not have_cvxpy:
                 raise ValueError(
-                    'CVXOPT package needed to enforce constraints')
-            if not hasattr(cvxopt, 'solvers'):
-                raise ValueError("CVXOPT version 1.1.7 or higher required")
+                    'CVXPY package needed to enforce constraints')
             msg = "pos_radius must be 'adaptive' or a positive float"
+            if cvxpy_solver is not None:
+                if cvxpy_solver not in cvxpy.installed_solvers():
+                    msg = "cvxpy_solver is not installed in cvxpy."
```
arokem: It's worth being more explicit here. A useful error is of the form: "I was expecting x, but I got y instead, check z". 

For example, here: 

    msg = "The input `cvxpy_solver` was set to %s."%cvxpy_solver
    msg += "One of %s"% ' , '.join(cvxp.installed_solvers)
    msg += " was expected."

dipy/reconst/mapmri.py
```diff
@@ -354,12 +368,6 @@ def fit(self, data):
                                         self.ind_mat.shape[0]))
 
         if self.positivity_constraint:
-            w_s = "The MAPMRI positivity constraint depends on CVXOPT "
```
arokem: Hooray! No more GPL 🎉

dipy/reconst/shore.py
```diff
@@ -188,6 +189,18 @@ def __init__(self,
...
             raise ValueError(msg)
+
+        if positive_constraint or constrain_e0:
+            if not have_cvxpy:
+                msg = "cvxpy must be installed for positive_constraint or "
+                msg += "constraint_e0."
+                raise ValueError(msg)
+            if cvxpy_solver is not None:
+                if cvxpy_solver not in cvxpy.installed_solvers():
+                    msg = "cvxpy_solver is not installed in cvxpy."
```
arokem: Same comment about this error message

dipy/reconst/shore.py
```diff
@@ -750,7 +761,8 @@ def shore_indices(radial_order, index):
     m_i = 0
 
     if n_c < (index + 1):
-        msg = "The index is higher than the number of coefficients of the truncated basis."
+        msg = "The index is higher than the number of coefficients of the "
+        msg += "truncated basis."
```
arokem: Here again it would be helpful to say what it received as input, and how it differs from the expected value (I know you only changed the line-wrapping here, but if we can improve it, why not?).

dipy/reconst/shore.py
```diff
@@ -787,7 +799,8 @@ def shore_order(n, l, m):
 
     """
     if l % 2 == 1 or l > n or l < 0 or n < 0 or np.abs(m) > l:
-        msg = "The index l must be even and 0 <= l <= n, the index m must be -l <= m <= l."
+        msg = "The index l must be even and 0 <= l <= n, the index m must be "
```
arokem: And same here. What did we get, and what are we expecting?

dipy/reconst/mapmri.py
```diff
@@ -205,9 +213,12 @@ def __init__(self,
         >>> odf = mapfit.odf(sphere)
         """
 
-        self.bvals = gtab.bvals
-        self.bvecs = gtab.bvecs
+        if np.sum(gtab.b0s_mask) == 0:
```
rutgerfick: Fixed.

dipy/reconst/shore.py
rutgerfick: Fixed.

dipy/reconst/shore.py
rutgerfick: Fixed.

dipy/reconst/shore.py
rutgerfick: Fixed.

dipy/reconst/mapmri.py
```diff
@@ -159,6 +161,10 @@ def __init__(self,
             the tissue diffusivity that is used when dti_scale_estimation is
             set to False. The default is that of typical white matter
             D=0.7e-3 _[5].
+        cvxpy_solver : str, optional
+            cvxpy solver name. Optionally optimize the positivity constraint
+            with a particular cvxpy solver. See http://www.cvxp for details.
```
Garyfallidis: cvxpy.org

dipy/reconst/mapmri.py
Garyfallidis: Yeah!

Garyfallidis:
Apart from a minor doc issue. This PR looks ready. I am not sure why coveralls decreased. It looks like another red herring.

dipy/reconst/mapmri.py
rutgerfick: Fixed for both mapmri and shore modules.

rutgerfick:
Hi @Garyfallidis , 
I fixed the minor doc issue. I guess this is ready to be merged then :-)
Cheers,
Rutger

PR #1288, opened 2017-06-29
NF Add the parameter fa_operator in auto_response function

GuillaumeTh:
Add a new parameter (fa_operator) in auto_response to give an operator for the comparison with the FA threshold.

@jchoude, @arokem, @MrBago Can you review ? This PR will be useful when the MSMT PR will be merged.

### DIFF
--- a/dipy/reconst/csdeconv.py
+++ b/dipy/reconst/csdeconv.py
@@ -809,9 +809,10 @@ def auto_response(gtab, data, roi_center=None, roi_radius=10, fa_thr=0.7,
         radius of cubic ROI
     fa_thr : float
         FA threshold
-    fa_operator : function
-        Function that defines operator used to compare the FA with the fa_thr.
-        Must be defined as `def function_name(FA, fa_thr)` and return a bool array.
+    fa_operator : callable
+        A callable that defines an operation that compares FA with the fa_thr. The operator
+        should have two positional arguments (e.g., `fa_operator(FA, fa_thr)`) and it should
+        return a bool array.
     return_number_of_voxels : bool
         If True, returns the number of voxels used for estimating the response
         function.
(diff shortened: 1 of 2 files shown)


arokem:
Looks useful. I'd prefer to have the operator be a proper callable (defined as a function), rather than a lambda defined in the function signature (that looks a bit weird to me). I'd also want to see a test that defines this callable as something different, to get a sense for how this will be used.

dipy/reconst/csdeconv.py
```diff
@@ -775,8 +809,9 @@ def auto_response(gtab, data, roi_center=None, roi_radius=10, fa_thr=0.7,
         radius of cubic ROI
     fa_thr : float
         FA threshold
-    fa_operator : lambda
-        operator used to compare the FA with the fa_thr.
+    fa_operator : function
+        Function that defines operator used to compare the FA with the fa_thr.
+        Must be defined as `def function_name(FA, fa_thr)` and return a bool array.
```
MrBago: why? I believe lambdas and functions are the same in python (expect lambdas don't get a function name).

GuillaumeTh: So, I don't know if my last commit is the right thing or not

arokem: I think that the code is fine (I think it's easier to understand this way. The documentation can be changed to say:

    fa_operator : callable
        A callable that defines an operation that compares FA with the fa_thr. The operator 
        should have two positional arguments (e.g., `fa_operator(FA, fa_thr)`) and it should
        return a bool array.

arokem:
This looks good to me. Unless I hear any objections, I will merge this after the holiday (July 4th).

Garyfallidis:
I am not sure about the name. fa_operator? fa_callable is preferred.

jchoude:
@skoudoro I work with @GuillaumeTh on this. I might be lost, but I don't see the change you requested. Mind pointing it?

Also, you say you prefer to merge #1283 before this one. Is it because both of them touch the test_csdeconv file? I don't think this might create any conflict.

Issue #1181, opened 2017-02-24
peaks warning while CSD reconstructing


salma1601:
Hi Dipy experts,

I systemtically have a warning when I FIT `ConstrainedSphericalDeconvModel` to my subjects data.
Is it something expected ?

>/usr/lib/python2.7/dist-packages/dipy/direction/peaks.py:504: RuntimeWarning: invalid value encountered in double_scalars
  return np.sqrt(numer / denom)

Thanks !

Garyfallidis:
Yes, it is expected. We are considering suppressing those warnings as they seem create more panic than help.

arokem:
How about doing something like this? https://github.com/nipy/dipy/pull/1289

salma1601:
>We are considering suppressing those warnings as they seem create more panic than help
+1

skoudoro:
It seems to be done, can I close this issue?

salma1601:
yes, thanks

PR #1289, opened 2017-06-30
MRG: Suppress a divide-by-zero warning.

arokem:
Suppress the warning described in #1181? Still needs testing.

### DIFF
--- a/dipy/direction/peaks.py
+++ b/dipy/direction/peaks.py
@@ -559,7 +559,10 @@ def gfa(samples):
     n = samples.shape[-1]
     numer = n * (diff * diff).sum(-1)
     denom = (n - 1) * (samples * samples).sum(-1)
-    return np.sqrt(numer / denom)
+    if denom > 0:
+        return np.sqrt(numer / denom)
+    else:
+        return np.nan
 
 
 def reshape_peaks_for_visualization(peaks):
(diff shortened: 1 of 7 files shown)


dipy/direction/peaks.py
```diff
@@ -552,15 +551,6 @@ def peaks_from_model(model, data, sphere, relative_peak_threshold,
                            odf_array if return_odf else None)
 
 
-def gfa(samples):
-    """The general fractional anisotropy of a function evaluated
```
skoudoro: it might be good to warn users / devs that the function moved in another module.

arokem: This would not affect users (and I believe that it wouldn't even affect the aformentioned tests!), because we are now importing the name `gfa` in from `reconst.odf` into `direction.peaks`. So the API hasn't changed at all.  

The only ones who might be surprised are developers who expect to find this function here. What do you propose to do to warn them about this?

skoudoro: Oh ok, I see. Sorry and thanks  !

Even if it does not have any bad effect, I do not understand why you want to keep this "old" import `from dipy.direction.peaks import gfa` in the aformentioned test?

arokem:
OK - changed these imports

arokem:
Does anyone understand how coverage went down when I removed some code and added a test?! That doesn't make sense...

skoudoro:
At the end, coverage went up ;-). It is more logic now. If there is no objection, I will merge it this monday (July 10th)

PR #1294, opened 2017-07-05
Suppress a warning in geometry.

arokem:
Another one of these annoying ones.

### DIFF
--- a/dipy/core/geometry.py
+++ b/dipy/core/geometry.py
@@ -126,7 +126,7 @@ def cart2sphere(x, y, z):
        azimuth angle
     '''
     r = np.sqrt(x * x + y * y + z * z)
-    theta = np.arccos(z / r, where=r > 0)
+    theta = np.arccos(np.divide(z, r, where=r > 0))
     theta = np.where(r > 0, theta, 0.)
     phi = np.arctan2(y, x)
     r, theta, phi = np.broadcast_arrays(r, theta, phi)
--- a/dipy/core/geometry.py
+++ b/dipy/core/geometry.py
@@ -126,10 +126,6 @@ def cart2sphere(x, y, z):
        azimuth angle
     '''
     r = np.sqrt(x * x + y * y + z * z)
-    # r is strictly positive by construction and will be 0 iff all three are
-    # zero, so it's safe to convert it to a 1 in this case, and prevents
-    # warnings from being raised in the following statement (z / r):
-    r = np.where(r > 0, r, 1)
     theta = np.arccos(z / r, where=r > 0)
     theta = np.where(r > 0, theta, 0.)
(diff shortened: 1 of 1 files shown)


dipy/core/geometry.py
```diff
@@ -125,8 +125,11 @@ def cart2sphere(x, y, z):
        azimuth angle
     '''
     r = np.sqrt(x * x + y * y + z * z)
+    # r is strictly positive by construction and will be 0 iff all three are
+    # zero, so it's safe to convert it to a 1 in this case, and prevents
+    # warnings from being raised in the following statement (z / r):
+    r = np.where(r > 0, r, 1)
     theta = np.arccos(z / r)
-    theta = np.where(r > 0, theta, 0.)
```
skoudoro: Since Travis is not happy because of some round values, why not just put it all in one line?  

`theta = np.where(r > 0, np.arccos(z / r), 0.) `

skoudoro: or even better : `theta = np.arccos(z / r, where=r>0)`

cool feature from Numpy! look [here](https://stackoverflow.com/a/37977222)

dipy/core/geometry.py
```diff
@@ -126,7 +126,11 @@ def cart2sphere(x, y, z):
        azimuth angle
     '''
     r = np.sqrt(x * x + y * y + z * z)
-    theta = np.arccos(z / r)
+    # r is strictly positive by construction and will be 0 iff all three are
+    # zero, so it's safe to convert it to a 1 in this case, and prevents
+    # warnings from being raised in the following statement (z / r):
+    r = np.where(r > 0, r, 1)
```
skoudoro: This part is not anymore necessary after the last changes :-)

skoudoro: Any chance to see this PR in the future release?  You just need to remove the lines from 129 to 132 and keep your change line 133.

arokem: Oh - I didn't understand that was all that was left. Done and rebased!

dipy/core/geometry.py
```diff
@@ -126,7 +126,7 @@ def cart2sphere(x, y, z):
        azimuth angle
     '''
     r = np.sqrt(x * x + y * y + z * z)
-    theta = np.arccos(z / r)
+    theta = np.arccos(z / r, where=r > 0)
```
skoudoro: it seems we still have the warning. That's weird... So,  I just tested another option: `np.arccos(np.divide(z, r, 
 where=r > 0))`
This solution seems to totally remove the warning.

arokem: Let's see...

skoudoro:
Yeah! this warning disappeared! merging!

Thank you @arokem !

PR #1297, opened 2017-07-10
Added eigh version of localpca to svd version

ssheybani:
Denoising with LocalPCA:
The bottleneck of LocalPCA is decomposition of the data into the eigenvalues/singular values and the corresponding vectors (basis). For this purpose, Eigenvalue decomposition was used but was previously replaced with Singular Values Decomposition, for more accuracy. However, SVD is many times slower than EVD. In this pull request, the code is changed to include both of them, as an optional parameter, "pca_method".
Also for more speed up, the function accepts the brain mask as an optional parameter and the denoising will be only applied to the voxels inside the mask.

### DIFF
--- a/dipy/denoise/localpca.py
+++ b/dipy/denoise/localpca.py
@@ -1,4 +1,13 @@
 import numpy as np
+try:
+    from scipy.linalg.lapack import dgesvd as svd
+    svd_args = [1, 0]
+    # If you have an older version of scipy, we fall back
+    # on the standard scipy SVD API:
+except ImportError:
+    from scipy.linalg import svd
+    svd_args = [False]
+from scipy.linalg import eigh
 
 
 def localpca(arr, sigma, mask=None, pca_method='eig', patch_radius=2,
@@ -53,26 +62,12 @@ def localpca(arr, sigma, mask=None, pca_method='eig', patch_radius=2,
         # If mask is not specified, use the whole volume
         mask = np.ones_like(arr, dtype=bool)[..., 0]
 
-    if pca_method == 'svd':
-        # Try to get the SVD through direct API to lapack:
-        try:
-            from scipy.linalg.lapack import dgesvd as svd
-            svd_args = [1, 0]
-            # If you have an older version of scipy, we fall back
-            # on the standard scipy SVD API:
-        except ImportError:
-            from scipy.linalg import svd
-            svd_args = [False]
-    else:
-        from scipy.linalg import eigh
-
     if out_dtype is None:
         out_dtype = arr.dtype
 
     # We retain float64 precision, iff the input is in this precision:
     if arr.dtype == np.float64:
         calc_dtype = np.float64
-
     # Otherwise, we'll calculate things in float32 (saving memory)
     else:
         calc_dtype = np.float32
@@ -80,6 +75,13 @@ def localpca(arr, sigma, mask=None, pca_method='eig', patch_radius=2,
     if not arr.ndim == 4:
         raise ValueError("PCA denoising can only be performed on 4D arrays.",
                          arr.shape)
+
+    if pca_method.lower() == 'svd':
+        is_svd = True
+    elif pca_method.lower() == 'eig':
+        is_svd = False
+    else:
+        raise ValueError("pca_method should be either 'eig' or 'svd'")
 
     patch_size = 2 * patch_radius + 1
 
@@ -109,7 +111,7 @@ def localpca(arr, sigma, mask=None, pca_method='eig', patch_radius=2,
         for j in range(patch_radius, arr.shape[1] - patch_radius):
             for i in range(patch_radius, arr.shape[0] - patch_radius):
                 # Shorthand for indexing variables:
-                if mask[i, j, k] == False:
+                if not mask[i, j, k]:
                     continue
                 ix1 = i - patch_radius
                 ix2 = i + patch_radius + 1
@@ -125,7 +127,7 @@ def localpca(arr, sigma, mask=None, pca_method='eig', patch_radius=2,
                 # Upcast the dtype for precision in the SVD
                 X = X - M
 
-                if pca_method == 'svd':
+                if is_svd:
                     # PCA using an SVD
                     U, S, Vt = svd(X, *svd_args)[:3]
                     # Items in S are the eigenvalues, but in ascending order
--- a/dipy/denoise/localpca.py
+++ b/dipy/denoise/localpca.py
@@ -1,20 +1,8 @@
 import numpy as np
-import scipy as sp
-import scipy.linalg as sla
-from warnings import warn
-
-# Try to get the SVD through direct API to lapack:
-try:
-    from scipy.linalg.lapack import sgesvd as svd
-    svd_args = [1, 0]
-# If you have an older version of scipy, we fall back
-# on the standard scipy SVD API:
-except ImportError:
-    from scipy.linalg import svd
-    svd_args = [False]
 
 
-def localpca(arr, sigma, patch_radius=2, tau_factor=2.3, out_dtype=None):
+def localpca(arr, mask, sigma, pca_method='eig', patch_radius=2,
+             tau_factor=2.3, out_dtype=None):
     r"""Local PCA-based denoising of diffusion datasets.
 
     Parameters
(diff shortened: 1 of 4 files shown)


dipy/denoise/localpca.py
```diff
@@ -53,10 +49,23 @@ def localpca(arr, sigma, patch_radius=2, tau_factor=2.3, out_dtype=None):
...
+            from scipy.linalg import svd
+            svd_args = [False]
+    else:
+        from scipy.linalg import eigh
+
     if out_dtype is None:
         out_dtype = arr.dtype
 
-    # We retain float64 precision, iff the input is in this precision:
+    # We retain float64 precision, if the input is in this precision:
```
arokem: This was intentional! To denote that we do this "if and only if".

dipy/denoise/localpca.py
```diff
@@ -97,6 +105,8 @@ def localpca(arr, sigma, patch_radius=2, tau_factor=2.3, out_dtype=None):
         for j in range(patch_radius, arr.shape[1] - patch_radius):
             for i in range(patch_radius, arr.shape[0] - patch_radius):
                 # Shorthand for indexing variables:
+                if mask[i, j, k] == False:
```
arokem: Nice!

dipy/denoise/localpca.py
```diff
@@ -133,4 +152,5 @@ def localpca(arr, sigma, patch_radius=2, tau_factor=2.3, out_dtype=None):
 
     denoised_arr = thetax / theta
     denoised_arr.clip(min=0, out=denoised_arr)
+    denoised_arr[~mask] = 0
```
arokem: Maybe put back the original values?

Garyfallidis:
Please also @ssheybani include more information in the description and show how eigh compares with svd in some slices (not in tests - just in the description for reference and helping the reviewers understand). We may have discussed some things in person but that doesn't mean that the others here know what we have discussed.

dipy/denoise/localpca.py
Garyfallidis: Nope, this is an application of the mask. If we put back the original values we will have noise in the background which is undesired.

arokem: That makes sense. Might be worth being explicit about this in the docstring. Some users might be surprised to see non-brain parts of the image (e.g., skull) disappear.

ssheybani:
Here are the results of comparison between the two methods (eig vs svd). The test data is a small portion of a Ferret dataset, with the shape [160, 120, 7, 210] (28224000 voxels in total)

## A summary of line profiling:

### With svd:
```
Total time: 450.4 s
Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   126      7731    419366666  54244.8     93.1                      U, S, Vt = svd(X, *svd_args)[:3]
   144      7731     20806916   2691.4      4.6                  Xest = X.dot(W).dot(W.T) + M
```
### With eig:
```
Total time: 83.4777 s
Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
137      7731      7992914   1033.9      9.6                      C = np.transpose(X).dot(X)
139      7731     47239798   6110.4     56.6                      [d, W] = eigh(C, turbo=True)
144      7731     18560852   2400.8     22.2                  Xest = X.dot(W).dot(W.T) + M
```
The denoised matrices resulted from the two methods were exactly the same (according to numpy.testing.assert_equal).
Here's one of the slices:
eigh_svd_compare_ferret
eigh_svd_compare_ferret_res

dipy/denoise/localpca.py
```diff
@@ -53,6 +49,23 @@ def localpca(arr, sigma, patch_radius=2, tau_factor=2.3, out_dtype=None):
                   PCA. PLoS ONE 8(9): e73021.
                   https://doi.org/10.1371/journal.pone.0073021
     """
+    if mask is None:
+        # If mask is not specified, use the whole volume
+        mask = np.ones_like(arr, dtype=bool)[..., 0]
+
+    if pca_method == 'svd':
```
Garyfallidis: Please add the imports at the top of the file.

dipy/denoise/tests/test_lpca.py
```diff
@@ -223,13 +228,13 @@ def test_phantom():
     rmse_noisy_wrc = np.sum(np.abs(DWI_clean_wrc - DWI)) / \
         np.sum(np.abs(DWI_clean_wrc))
 
+
```
Garyfallidis: Add a test with the mask.

dipy/denoise/localpca.py
```diff
@@ -110,15 +124,24 @@ def localpca(arr, sigma, patch_radius=2, tau_factor=2.3, out_dtype=None):
...
-                U, S, Vt = svd(X, *svd_args)[:3]
-                # Items in S are the eigenvalues, but in ascending order
-                # We invert the order (=> descending), square and normalize
-                # \lambda_i = s_i^2 / n
-                d = S[::-1] ** 2 / X.shape[0]
-                # Rows of Vt are eigenvectors, but also in ascending eigenvalue
-                # order:
-                W = Vt[::-1].T
+
+                if pca_method == 'svd':
```
Garyfallidis: No need to check the string each time. Add a boolean variable.

arokem:
Where does this stand? Looks like we are still missing a test with the mask, but that this is otherwise ready to go?

dipy/denoise/tests/test_lpca.py
```diff
@@ -212,9 +213,14 @@ def test_phantom():
...
 
+    # Check if the results of different PCA methods (eig, svd) are similar
+    DWI_den_svd = localpca(DWI, sigma, pca_method='svd', patch_radius=3)
+    assert_array_almost_equal(DWI_den, DWI_den_svd)
+
     # Try this with a sigma volume, instead of a scalar
     sigma_vol = sigma * np.ones(DWI.shape[:-1])
-    DWI_den = localpca(DWI, sigma_vol, patch_radius=3)
+    mask = np.ones_like(DWI, dtype=bool)[..., 0]
+    DWI_den = localpca(DWI, sigma_vol, mask, patch_radius=3)
```
arokem: Right now, this doesn't really test that the mask does anything. Maybe make a mask with some ones and some zeros and check that the part that has ones in it gets denoised, while the part that has zeros returns as all zeros.

dipy/denoise/localpca.py
```diff
@@ -1,29 +1,34 @@
...
     r"""Local PCA-based denoising of diffusion datasets.
 
     Parameters
     ----------
     arr : 4D array
         Array of data to be denoised. The dimensions are (X, Y, Z, N), where N
         are the diffusion gradient directions.
+    mask : 3D boolean array
+        A mask with voxels that are true inside the brain and false outside of
+        it.
```
arokem: I'm still missing an explanation of what the mask does here: it denoises within the True part and returns *zeros* outside of those voxels, right?

Garyfallidis:
Ping @ssheybani there are some final requests here. It would be great to merge this asap. Thanks.

dipy/denoise/localpca.py
ssheybani: Yes.

dipy/denoise/tests/test_lpca.py
ssheybani: I did that. But assert_(rmse_den < rmse_noisy) gave me an error. I think that's most likely because setting the out-of-mask areas to zero has caused the denoised image to have more error than the original noisy one. I checked the DWI_clean matrix and it looks like that for a good number of slices, the background is not actually black (0) but white (1)! Do you have any suggestions on what is the preferred way to deal with it?

arokem: Sounds to me like the mask is doing something funky! I am not sure where that arises. Could you implement this and push it here? I might be able to fetch your branch and play around with it.

dipy/denoise/localpca.py
arokem: Please add that to the docstring.

ssheybani: Done

dipy/denoise/tests/test_lpca.py
ssheybani: I just added this for instance:
`mask[0, ...] = False`

arokem: I get this failure only for the second comparison: 

    assert_(rmse_den_wrc < rmse_noisy_wrc)
 		
not for the first one: 

    assert_(rmse_den < rmse_noisy)

ssheybani: I just corrected it with applying the mask on the original data as well when it comes to calculating rmse values.

arokem:
Yup. I think that was it. 

This is ready to merge from my point of view. @Garyfallidis : feel free to merge if all your comments have been addressed.

Garyfallidis:
Great! Thank you @ssheybani :)

PR #1298, opened 2017-07-11
Make TextActor2D extend UI instead of object

dmreagan:
Previously, `TextActor2D` was used by other classes like `TextBox2D` and `LineSlider2D`, but could not be added to the `ShowManager` nor to a `Panel2D` as a standalone UI element. However, `TextActor2D` can be useful as a text label which is not editable by users, so this PR promotes it to a full member of `viz.ui`.

### DIFF
--- a/dipy/viz/tests/test_ui.py
+++ b/dipy/viz/tests/test_ui.py
@@ -154,7 +154,6 @@ def test_ui_button_panel(recording=False):
     text_block_test = ui.TextBlock2D()
     text_block_test.message = 'TextBlock'
     text_block_test.color = (0, 0, 0)
-    # /TextBlock
 
     # Panel
     panel = ui.Panel2D(center=(440, 90), size=(300, 150),
@@ -254,7 +253,6 @@ def test_text_block_2d():
     npt.assert_equal((1, 0, 0), text_block.color)
     text_block.position = (2, 3)
     npt.assert_equal((2, 3), text_block.position)
-    # /TextBlock2D
 
 
 @npt.dec.skipif(not have_vtk or skip_it)
--- a/dipy/viz/tests/test_ui.py
+++ b/dipy/viz/tests/test_ui.py
@@ -150,11 +150,18 @@ def test_ui_button_panel(recording=False):
     button_test.color = button_color
     # /Button
 
+    # TextBlock
+    text_block_test = ui.TextBlock2D()
+    text_block_test.message = 'TextBlock'
+    text_block_test.color = (0, 0, 0)
+    # /TextBlock
+
     # Panel
     panel = ui.Panel2D(center=(440, 90), size=(300, 150),
                        color=(1, 1, 1), align="right")
     panel.add_element(rectangle_test, 'absolute', (580, 150))
     panel.add_element(button_test, 'relative', (0.2, 0.2))
+    panel.add_element(text_block_test, 'relative', (0.7, 0.7))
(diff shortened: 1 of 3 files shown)


dmreagan:
What do you think, @ranveeraggarwal , @Garyfallidis ? Do I need to update the UI tests?

ranveeraggarwal:
Actually, `TextActor2D` has had quite a history. We initially inherited it from the vtk text actor and then later chose composition over inheritance and made everything a property. After that, we never really thought about it. This sounds like a good idea to me. In fact, I had been meaning to do so in my current dangling PR on `FileSelect2D`.    
Let's call it TextBlock2D instead of TextActor2D then, because it's not just an actor anymore, it's a full fledged UI element (yep, this is inspired from WPF, where you have textboxes and textblocks).

dipy/viz/ui.py
```diff
@@ -924,6 +931,16 @@ def position(self, position):
...
 
+    def set_center(self, position):
+        """ Sets the text center to position.
+
+        Parameters
+        ----------
+        position : (float, float)
+
+        """
+        self.actor.SetPosition(*position)
```
ranveeraggarwal: Use the property instead. 

    self.position = position

ranveeraggarwal:
For the tests, you can add a *textblock* to an existing panel's test so that the `get_actor` and `set_center` methods are covered.

dipy/viz/tests/test_ui.py
```diff
@@ -150,11 +150,18 @@ def modify_button_callback(i_ren, obj, button):
     button_test.color = button_color
     # /Button
 
+    # TextBlock
+    text_block_test = ui.TextBlock2D()
+    text_block_test.message = 'TextBlock'
+    text_block_test.color = (0, 0, 0)
+    # /TextBlock
```
Garyfallidis: Are comments like this necessary?

dipy/viz/tests/test_ui.py
```diff
@@ -217,37 +224,37 @@ def test_ui_textbox(recording=False):
...
+    text_block.italic = False
+    npt.assert_equal(False, text_block.italic)
+    text_block.shadow = True
+    text_block.shadow = False
+    npt.assert_equal(False, text_block.shadow)
+    text_block.color = (1, 0, 0)
+    npt.assert_equal((1, 0, 0), text_block.color)
+    text_block.position = (2, 3)
+    npt.assert_equal((2, 3), text_block.position)
+    # /TextBlock2D
```
Garyfallidis: Same here.

Garyfallidis:
I tried the code in my computer. Seems working well. Apart from some minor comments, seems this is ready to be merged.

dipy/viz/tests/test_ui.py
dmreagan: No, I just copied the style I saw for other tests. I'll remove them.

ranveeraggarwal: I just added them for easy readability, since we're putting in all components in a single file. They can be removed, no problem.

ranveeraggarwal:
@dmreagan do verify that the other dependent UI components (TextBox, LineSlider and Circular Slider) work with these changes. PR #1262 will need fixes with these changes.

dmreagan:
@ranveeraggarwal Those components look good to me in the viz_ui.py example, so I think it is good to go. I'm pretty sure I just had to switch the name.

Garyfallidis:
Thank you @dmreagan

Issue #1299, opened 2017-07-11
test_rmi on 32 bit:  invalid dims: array size defined by dims is larger than the maximum possible size.


yarikoptic:
```
======================================================================
ERROR: dipy.tracking.tests.test_utils.test_rmi
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/lib/python2.7/dist-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/build/dipy-0.12.0/debian/tmp/usr/lib/python2.7/dist-packages/dipy/tracking/tests/test_utils.py", line 4
72, in test_rmi
    I2 = ravel_multi_index([A, B, C, D], dims=[1000] * 4)
ValueError: invalid dims: array size defined by dims is larger than the maximum possible size.
```
was not sure if 1000 was intentionally large to possibly catch some stats deviations thus shouldn't be generally decreased...

arokem:
See #1300, which potentially solves this issue, by eliminating the whole thing (test and implementation...)

arokem:
Closed through #1300

PR #1300, opened 2017-07-13
RF: Remove patch for older numpy ravel_multi_index.

arokem:
Attempting to address #1299. 

It looks like the test that is failing there is related to a patch of numpy's ravel_multi_index for older versions of numpy. I am putting in this PR to see whether we can do without that patch altogether at this point.

### DIFF
--- a/dipy/tracking/tests/test_utils.py
+++ b/dipy/tracking/tests/test_utils.py
@@ -11,7 +11,7 @@ from dipy.tracking.utils import (affine_for_trackvis, connectivity_matrix,
                                  ndbincount, reduce_labels,
                                  reorder_voxels_affine, seeds_from_mask,
                                  random_seeds_from_mask, target,
-                                 target_line_based, _rmi, unique_rows, near_roi,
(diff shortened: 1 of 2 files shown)


arokem:
Yep. Looks like we don't need this patch anymore. Might already be fixed on our minimal numpy requirement?

Garyfallidis:
Do we need to remove the test completely?

arokem:
Yes. I think that if we remove the implementation, we should also remove the test. 

The use of `numpy`'s `ravel_multi_index` is tested through it's downstream use in `ndbincount`.

yarikoptic:
Could decision be made on this PR since it is a blocker for #1299 !

arokem:
Hey @Garyfallidis : any more thoughts about this one?

Garyfallidis:
Sure, no issue on my side @arokem ! :)

arokem:
I will take that as approval to go ahead and merge this PR.

arokem:
Resolves #1299

PR #1303, opened 2017-07-17
DOC: Add missing label to reciprocal space eq.

jhlegarreta:
Add missing label to reciprocal space equation in b_and_q.rst file in
the doc/theory folder.

Although according to
http://www.sphinx-doc.org/en/stable/ext/math.html
cross-referencing equations via their label will not work across
different documents.

The equation at issue is referenced from doc/faq.rst.


arokem:
Yes. I suspect this might still not work, but let's give it a try...

PR #1308, opened 2017-07-26
Fix inversion in the dti mode doc

matthieudumont:
According to the article referenced in the docstring and some quick tests **-1=planar anisotropy** and **1=linear anisotropy**


arokem:
Good catch. Thanks!

Issue #366, opened 2014-05-13
Some doc missing


EmmaRenauld:
Hi,
the module tracking.distances (http://nipy.org/dipy/reference/dipy.tracking.distances.html) doesn't appear on the documentation page for tracking (http://nipy.org/dipy/reference/dipy.tracking.html).

skoudoro:
![](https://github.trello.services/images/mini-trello-icon.png) [#366](https://trello.com/c/4OgLE3DQ/2-366)

PR #1311, opened 2017-07-29
Adding cython file(*.pyx) in documentation

### LINKED ISSUE #1302: [DOC] cython (pyx) files are not parsed
After looking at #366, All functions in a pyx files doesn't appear on the documentation page. This extension seems to be excluded from the doc generator.

Does it come from [here](https://github.com/nipy/dipy/blob/master/doc/tools/apigen.py#L400) or there is something else ? 

is there any historical reason to skip cython files  ?

skoudoro:
This PR should resolve #1302 and #366. Indeed, cython files were excluded from documentation generation. 

I believe that these files were excluded because Sphinx crashed during doc generation.

Sphinx seems to be tricky (it crashes) when we use default arguments on the cython function so I had to modify the definitions of some functions.

### DIFF
--- a/dipy/align/bundlemin.pyx
+++ b/dipy/align/bundlemin.pyx
@@ -1,4 +1,7 @@
 #!python
+
+# cython: embedsignature=True
+
 #cython: boundscheck=False
 #cython: wraparound=False
 #cython: cdivision=True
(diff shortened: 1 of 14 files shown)


Garyfallidis:
Hi @omarocegueda this PR is changing some parts of your code. Do have a look. The None parameters are removed because sphinx does not accept them for building the docs with cython.

omarocegueda:
Sure! I'll take a look tonight =)

omarocegueda:
Hi @skoudoro! thank you very much for doing this! I didn't know about this limitation in Sphinx. 

The constraint of not using default values in Cython seems too strong, isn't it? Is this a known bug/limitation in Sphinx? I mean, are there any plans to fix it on their side?

PR #1320, opened 2017-08-15
WIP: Bias correction

Tingyiwanyan:
This code implements a purely statistical based Non-parametric learning on bias
field correction using an N3 stratergy but it is not identical to N3.


arokem:
This is stalled and unlikely to move forward. Closing.

PR #1335, opened 2017-09-18
Catch a more specific warning in test_csdeconv

### LINKED ISSUE #1323: Warnings raised in csdeconv for upcoming numpy 1.14
See: https://travis-ci.org/nipy/dipy/jobs/268292931, and conversation in #1322

skoudoro:
The goal of this PR is to fix #1323.

Like @arokem said, the test was catching too many warnings. In this case, the test was catching this : 

`FutureWarning: 'rcond' parameter will change to the default of machine precision times ``max(M, N)`` where M and N are the input matrix dimensions.
To use the future default and silence this warning we advise to pass 'rcond=None', to keep using the old, explicitly pass 'rcond=-1'. `


arokem:
Looks great. The coveralls failure is bogus, I believe. Merging!

PR #1336, opened 2017-09-18
Test modification for windows 10 / numpy 1.14

skoudoro:
This PR is a correction for test_dti.py which is failing on windows 10 / numpy 1.14

The arrays seems to be different on the 14nth decimal place. This is really strange. 

This problem does not appear on linux.


Garyfallidis:
Cool, thanks.

PR #1340, opened 2017-09-22
NF - Particle Filtering Tractography

gabknight:
This PR add a new tissue classifier (CMC) and a new tractography algorithm (PFT). Both methods are described in [Girard et al., NImg, 2014].

- CMC - new tissue Classifier. `CmcTissueClassifier` used probabilisties and the tissue partial volume fraction maps to determine when the tracking stops. As it is very similar to `ActTissueClassifier`, both classes now extend an a new abstract class `ConstrainedTissueClassifier`.

- PFT - new tractography method. `ParticleFilteringTracking` is inherited from the `LocalTracking` class. The particle filter is used to find a valid streamline path when the streamline being traced reach the state `INVALIDPOINT`.

  - An example is giving in `doc/examples/particle_filtering_fiber_tracking.py`

@nilgoyette found potential optimisation for the cumsum(...) function, called several times in `dipy.tracking.local.localtrack._pft(...)` and `dipy.direction.probabilistic_direction_getter.ProbabilisticDirectionGetter.get_direction(...)`. A PR should follow :).

@MrBago I would be very glad to get your feedback on the way I actually structured the implementation of PFT in dipy. PFT is the first tracking algorithm in dipy that is not a different `direction getter` as such. PFT uses a probabilistic direction getter, and needs to know the tracking state from the `tissueClassifier` at every steps and use it change the reconstruction of the streamline. I thus choose to extend the `LocalTracking` class with the `ParticleFilteringTracking` class and added the _tracker(...) function to the `LocalTracking` class. `ParticleFilteringTracking` returns the pft_tracker cython function, `LocalTracking` returns the local_tracker cyhton function. Both use the same `_generate_streamlines(...)` function from `LocalTracking`. 

@Garyfallidis, @arokem, @skoudoro, @jchoude, I would be very much interested to hear your thoughts on this implementation. On the CC exemple, PFT is ~4.5x slower than ProbabilisticDirectionGetter LocalTracking, but outputs 3x more streamlines connecting the GM (i.e. not stopping in the WM or CSF). PFT has a seed to streamlines ratio of ~93% while the same algorithm With localTracking has a ratio of ~29% (for CC seeding of the example).

References
----------
Girard, G., Whittingstall, K., Deriche, R., & Descoteaux, M. Towards quantitative connectivity analysis: reducing tractography biases. NeuroImage, 98, 266-278, 2014.


gabknight:
I look forward to your feedback. I will unresponsive until next week as I'm in travelling...

StongeEtienne:
@gabknight I will try to look at it, in the next weeks !

skoudoro:
Since #1342 is merged, can you rebase this one @gabknight? Thanks !

nilgoyette:
@gabknight I don't remember what we said... you rebase and I open another PR for some enhancements, or I rebase and enhance in this PR?

gabknight:
@nilgoyette @skoudoro I will rebase this. 

@nilgoyette I will also update the particle filtering part following #1342. It would be ideal if you can then have a look to the whole PR for further optimisation.

gabknight:
I screw thing up here.... I open a new PR #1384 with the good commits

PR #1343, opened 2017-09-29
ActiveAx model fitting using MIX framework

maryamafzali:
ActiveAx model fitting using MIX framework


arokem:
I have three high-level comments here, before I have even looked at the code much:

1. It's awesome to have this as part of Dipy! Thanks for your work on this! I am excited to take a look.

2. ~3k lines of code is a lot to review in one go! Consider whether there is any way to break this up into a few different related PRs, or a series of PRs to follow each other. I would say that this should probably be broken down into at least 3 different PRs, just based on the number of LOC.

3. I would recommend using slightly more informative commit messages, so that you communicate more clearly what happened and when. See this excellent blog-post for some guidelines on that: https://chris.beams.io/posts/git-commit/

PR #1348, opened 2017-10-05
ENH: Add support for ArraySequence in`set_number_of_points` function

MarcCote:
Second small PR that aims to support Nibabel's `ArraySequence`objects (the first one was #1076).

Specifically, this PR improves `dipy.tracking.streamlinespeed.set_number_of_points` to support `ArraySequence` objects. This gives pretty good speedup even over the current Cython version (all speedups are with respect to the Python implementation):

```bash
Timing set_number_of_points() in 100,000 streamlines.
Cython time: 0.590 sec
Python time: 61.76 sec
Speed up of 104.68x
Cython time (ArrSeq): 0.13 sec
Speed up of 475.08x
```

### DIFF
--- a/dipy/tracking/benchmarks/bench_streamline.py
+++ b/dipy/tracking/benchmarks/bench_streamline.py
@@ -99,8 +99,8 @@ def bench_length():
     print("Speed up of {0:.2f}x".format(python_time/cython_time))
 
     # Make sure it produces the same results.
-    assert_array_equal([length_python(s) for s in DATA["streamlines"]],
-                       length(DATA["streamlines"]))
+    assert_array_almost_equal([length_python(s) for s in DATA["streamlines"]],
+                              length(DATA["streamlines"]))
 
     streamlines = DATA['streamlines_arrseq']
     cython_time_arrseq = measure("length(streamlines)", repeat)
--- a/dipy/tracking/benchmarks/bench_streamline.py
+++ b/dipy/tracking/benchmarks/bench_streamline.py
@@ -15,7 +15,7 @@ Run this benchmark with:
 """
 import numpy as np
 from numpy.testing import measure
-from numpy.testing import assert_array_equal
+from numpy.testing import assert_array_equal, assert_array_almost_equal
 
 from dipy.data import get_data
 from nibabel import trackvis as tv
@@ -54,29 +54,34 @@ def generate_streamlines(nb_streamlines, min_nb_points, max_nb_points, rng):
 
 
 def bench_set_number_of_points():
-    repeat = 1
-    nb_points_per_streamline = 100
+    repeat = 5
     nb_points = 42
-    nb_streamlines = int(1e4)
-    streamlines = [np.random.rand(nb_points_per_streamline,
-                                  3).astype("float32")
-                   for i in range(nb_streamlines)]
+    nb_streamlines = DATA['nb_streamlines']
+    streamlines = DATA["streamlines"]  # Streamlines as a list of ndarrays.
 
-    print("Timing set_number_of_points() in Cython"
-          "({0} streamlines)".format(nb_streamlines))
+    msg = "Timing set_number_of_points() with {0:,} streamlines."
+    print(msg.format(nb_streamlines * repeat))
     cython_time = measure("set_number_of_points(streamlines, nb_points)",
                           repeat)
-    print("Cython time: {0:.3}sec".format(cython_time))
-    del streamlines
+    print("Cython time: {0:.3f} sec".format(cython_time))
 
-    streamlines = [np.random.rand(nb_points_per_streamline,
-                                  3).astype("float32")
-                   for i in range(nb_streamlines)]
     python_time = measure("[set_number_of_points_python(s, nb_points)"
                           " for s in streamlines]", repeat)
-    print("Python time: {0:.2}sec".format(python_time))
-    print("Speed up of {0}x".format(python_time/cython_time))
-    del streamlines
+    print("Python time: {0:.2f} sec".format(python_time))
+    print("Speed up of {0:.2f}x".format(python_time/cython_time))
+
+    # Make sure it produces the same results.
+    assert_array_almost_equal([set_number_of_points_python(s) for s in DATA["streamlines"]],
+                              set_number_of_points(DATA["streamlines"]))
+
+    streamlines = DATA['streamlines_arrseq']
+    cython_time_arrseq = measure("set_number_of_points(streamlines, nb_points)", repeat)
+    print("Cython time (ArrSeq): {0:.3f} sec".format(cython_time_arrseq))
+    print("Speed up of {0:.2f}x".format(python_time/cython_time_arrseq))
+
+    # Make sure it produces the same results.
+    assert_array_equal(set_number_of_points(DATA["streamlines"]),
+                       set_number_of_points(DATA["streamlines_arrseq"]))
 
 
 def bench_length():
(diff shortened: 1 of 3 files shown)


skoudoro:
Hi @MarcCote , 

I got this error when I run the tests on my environment : Windows 10, Python 3.6, anaconda, Cython 0.25.2, numpy 1.12.1

```
=================================================================
ERROR: dipy.tracking.tests.test_streamline.test_set_number_of_points
  File "C:\Users\skoudoro\AppData\Local\Continuum\Anaconda3\envs\dipy-dev-3\lib\site-packages\nose\case.py", line 197, in runTest
    self.test(*self.arg)
  File "C:\Users\skoudoro\Devel\dipy\dipy\tracking\tests\test_streamline.py", line 239, in test_set_number_of_points
    new_streamlines_as_seq_cython = set_number_of_points(arrseq, nb_points)
  File "dipy\tracking\streamlinespeed.pyx", line 335, in dipy.tracking.streamlinespeed.set_number_of_points
    streamlines._data, streamlines._offsets,
ValueError: Buffer dtype mismatch, expected 'npy_intp' but got 'long'
```

dipy/tracking/benchmarks/bench_streamline.py
```diff
@@ -54,42 +54,48 @@ def generate_streamlines(nb_streamlines, min_nb_points, max_nb_points, rng):
 
 
 def bench_set_number_of_points():
-    repeat = 1
-    nb_points_per_streamline = 100
+    repeat = 5
     nb_points = 42
```
skoudoro: this variable seems to be replaced by `repeat` so you can remove it I think

dipy/tracking/streamlinespeed.pyx
```diff
@@ -276,15 +305,44 @@ def set_number_of_points(streamlines, nb_points=3):
...
+        new_streamlines._data = np.zeros((nb_streamlines * nb_points, 3),
+                                         dtype=dtype)
+        new_streamlines._offsets = nb_points * np.arange(nb_streamlines,
+                                                         dtype=np.intp)
+        new_streamlines._lengths = nb_points * np.ones(nb_streamlines,
+                                                       dtype=np.intp)
+
+        if dtype == np.float32:
+            c_set_number_of_points_from_arraysequence[float2d](
+                streamlines._data, streamlines._offsets,
```
skoudoro: It seems you are using the protected variables in many places. So, I just wonder why there is not get/set property for these variables on Nibabel?

skoudoro:
Hi @MarcCote ,

I try to run the benchmarks, but it fails. Did I do something wrong ?

```
In [1]: import dipy.tracking as dt

In [2]: dt.bench()
Running benchmarks for dipy.tracking
NumPy version 1.13.1
NumPy relaxed strides checking option: True
NumPy is installed in C:\Users\skoudoro\AppData\Local\Continuum\Anaconda3\envs\dipy-dev-3\lib\site-packages\numpy
Python version 3.6.2 |Continuum Analytics, Inc.| (default, Jul 20 2017, 12:30:02) [MSC v.1900 64 bit (AMD64)]
nose version 1.3.7
Timing set_number_of_points() with 100,000 streamlines.
Cython time: 0.430 sec
ETiming length() with 200,000 streamlines.
Python time: 2.03 sec
Cython time: 0.280 sec
Speed up of 7.25x
FTiming compress_streamlines() in Cython (300 streamlines)
Cython time: 0.02sec
Python time: 0.0sec
Speed up of 0.0x
.
ERROR: dipy.tracking.benchmarks.bench_streamline.bench_set_number_of_points
  File "C:\Users\skoudoro\AppData\Local\Continuum\Anaconda3\envs\dipy-dev-3\lib\site-packages\nose\case.py", line 197, in runTest
    self.test(*self.arg)
  File "C:\Users\skoudoro\Devel\dipy\dipy\tracking\benchmarks\bench_streamline.py", line 69, in bench_set_number_of_points
    " for s in streamlines]", repeat)
  File "C:\Users\skoudoro\AppData\Local\Continuum\Anaconda3\envs\dipy-dev-3\lib\site-packages\numpy\testing\utils.py", line 1315, in measure
    exec(code, globs, locs)
  File "Test name: None ", line 1, in <module>
    """ Benchmarks for functions related to streamline
  File "Test name: None ", line 1, in <listcomp>
    """ Benchmarks for functions related to streamline
NameError: name 'nb_points' is not defined

FAIL: dipy.tracking.benchmarks.bench_streamline.bench_length
  File "C:\Users\skoudoro\AppData\Local\Continuum\Anaconda3\envs\dipy-dev-3\lib\site-packages\nose\case.py", line 197, in runTest
    self.test(*self.arg)
  File "C:\Users\skoudoro\Devel\dipy\dipy\tracking\benchmarks\bench_streamline.py", line 103, in bench_length
    length(DATA["streamlines"]))
  File "C:\Users\skoudoro\AppData\Local\Continuum\Anaconda3\envs\dipy-dev-3\lib\site-packages\numpy\testing\utils.py", line 854, in assert_array_equal
    verbose=verbose, header='Arrays are not equal')
  File "C:\Users\skoudoro\AppData\Local\Continuum\Anaconda3\envs\dipy-dev-3\lib\site-packages\numpy\testing\utils.py", line 778, in assert_array_compare
    raise AssertionError(msg)
AssertionError:
Arrays are not equal

(mismatch 57.825%)
 x: array([ 36.168137,  14.202025,  31.208102, ...,  30.387572,  26.99025 ,
        10.31078 ])
 y: array([ 36.168137,  14.202025,  31.208102, ...,  30.387572,  26.99025 ,
        10.31078 ])

Ran 3 tests in 3.124s

FAILED (errors=1, failures=1)
Out[2]: False
 ```

**Comment by MarcCote** [comment]
@skoudoro thanks, I forgot to cast the offsets and lengths to `np.intp`. Regarding accessing those protected variables, you are right. We should provide get/set methods for those.

Regarding the benchmark function, I'm a bit confused. I don't understand why it says `nb_points` no defined :(. I don't have this problem on my linux machine. I use the following command to run the benchmark: `nosetests -s --match '(?:^|[\\b_\\.//-])[Bb]ench' dipy/tracking/benchmarks/bench_streamline.py:bench_set_number_of_points`.

The array mismatch error when benchmarking the `length` function seems to be a precision error (maybe a Windows vs. Linux thing).

**Comment by MarcCote** [comment]
@matthew-brett any suggestion on the best way to expose `ArraySequence._data`, ArraySequence._offsets` and `ArraySequence._lengths` in NiBabel? Should we just make them public? The main reason why I made them private in the first place was to present a cleaner API to the user, but clearly it can be useful to play with those attributes directly (for advanced users let's say).

File: dipy/tracking/benchmarks/bench_streamline.py [hunk 3]
```diff
@@ -98,7 +104,7 @@ def bench_length():
```
skoudoro: Because of windows error, can you replace it by 
`assert_array_almost_equal([length_python(s) for s in DATA["length(DATA["streamlines"]), decimal=12)`

MarcCote: It should be fine now. I used `assert_array_almost_equal` without default value for `decimal` as we do in the tests.

skoudoro: perfect ! Thanks !

dipy/tracking/streamlinespeed.pyx
```diff
@@ -305,7 +363,7 @@ def set_number_of_points(streamlines, nb_points=3):
             raise ValueError("All streamlines must have at least 2 points.")
 
     # Allocate memory for each modified streamline
-    modified_streamlines = []
+    new_streamlines = []
```
Garyfallidis: @MarcCote shouldn't this be Streamlines rather than []?

Garyfallidis:
Hi @MarcCote! This looks ready to be merged apart from a minor issue. It seems there are still some lists rather than Streamlines (ArraySequence) in streamlinespeed.pyx. Can you check?

dipy/tracking/streamlinespeed.pyx
MarcCote: This part of the code handles the case where a list of ndarrays was provided. In other words, if the user calls `set_number_of_points` on a list of ndarrays, we return a list of modified ndarrays. If the user provides an `ArraySequence` (a.k.a. `dipy.tracking.Streamlines`) object, we return a modified `ArraySequence`. Let me know if that behavior is not what you were expecting.

Garyfallidis:
Okay thank you for the explanation. It makes sense to have a plan B. We may want to enforce later having only Streamlines as input. Doesn't need to happen now.

PR #1349, opened 2017-10-06
Doctest FIX : use legacy printing

### LINKED ISSUE #1346: Test broken in numpy 1.14
As we can see on recent PR like [here](https://travis-ci.org/nipy/dipy/jobs/282429810#L4407),  all basic test seems to be broken. 

All numpy arrays seem to be different after 10-12 decimal. I am surprised to see this problem for such basic function like `vec_norm`. This behavior is really weird 

Unfortunately, I can not reproduce the problem on window 10, python=3.6, numpy=1.14.0.dev0+2995e6a
When I will find a time, I will try with my mac.

Do someone have any suggestions for the good way to solve this problem? @Garyfallidis @arokem @matthew-brett  @anyoneelse

skoudoro:
This PR is inspired by this @effigies to resolve printing problem. Indeed, Numpy change printing strategies so we revert printing behavior to keep Doctests from breaking.

Thanks to @matthew-brett for his advise.

this should resolve #1346

### DIFF
--- a/dipy/__init__.py
+++ b/dipy/__init__.py
@@ -36,6 +36,7 @@ if sys.version[0:3] < '2.6':
     raise ImportError('Dipy needs Python version 2.6 or above')
 
 from .info import __version__
+from .testing import setup_test
 
 # Test callable
 from numpy.testing import Tester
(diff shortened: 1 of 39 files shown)


arokem:
Looks like you got a few other things convolved in with these changes. This PR should be just that last commit, right?

skoudoro:
Right, I will fix that

PR #1350, opened 2017-10-07
WIP: Add mapmri flow

aarya22:
Not sure how to create a workflow or if this is correct

@Garyfallidis could you check this out and see if this is what you meant?


Garyfallidis:
Hi @aarya, okay, you are moving towards the correct direction but not there yet. Will start commenting asap. Thanks!

aarya22:
@Garyfallidis anything I can change?

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,77 @@
 from dipy.reconst.shm import CsaOdfModel
 from dipy.workflows.workflow import Workflow
 
+from dipy.reconst import mapmri
+import matplotlib
+matplotlib.use('agg')
+import matplotlib.pyplot as plt
```
skoudoro: I think, you do not need `matplotlib` here

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,77 @@
...
+from mpl_toolkits.axes_grid1 import make_axes_locatable
+
+class ReconstMAPMRIFlow(Workflow):
+    @classmethod
+    def get_short_name(cls):
+        return 'mapmri'
+
+    def run(self, data_file, data_bvecs, data_bvals, small_delta, big_delta):
```
skoudoro: you should define output files init the `run` function. Look at the example on the following link : https://github.com/nipy/dipy/blob/master/dipy/workflows/reconst.py#L34

It will be good if you can add default value too. Indeed, it will be useful for beginners who do not know what to put.

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,77 @@
...
+        img = nib.load(data_file)
```
skoudoro: try to use `io_iterator`like the example [here](https://github.com/nipy/dipy/blob/master/dipy/workflows/reconst.py#L90). It permits you to read many input files.

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,77 @@
...
+        bvals,bvecs = read_bvals_bvecs(data_bval,data_bvec)
```
skoudoro: same as above

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,77 @@
...
+        ax3 = fig.add_subplot(1, 3, 3, title=r'RTOP - Both')
+        ax3.set_axis_off()
+        ind = ax3.imshow(mapfit_both_aniso.rtop()[:, 0, :].T,
+                        interpolation='nearest', origin='lower', cmap=plt.cm.gray,
+                        vmin=0, vmax=5e7)
+        divider = make_axes_locatable(ax3)
+        cax = divider.append_axes("right", size="5%", pad=0.05)
+        plt.colorbar(ind, cax=cax)
+
+        plt.savefig('MAPMRI_maps_regularization.png')
```
skoudoro: I suppose we do not want to save this data in this way but we prefer to return many output files that can be input files for another process (or workflow). Currently, we can not do that with your code. So you should remove the `matplolib` part and replace it. you can see an example in the link above

dipy/workflows/reconst.py
skoudoro: Documentation is really important here. You can look at the example [here](https://github.com/nipy/dipy/blob/master/dipy/workflows/reconst.py#L36) how to make a good documentation

aarya22:
Sorry for the long delay in replying to your comments @skoudoro I have been busy with school and midterms. Take a look at the updated workflow. Unfortunately, I cannot change the output of the workflow. Brain-Life is planning on using this workflow to power one of its applications (app-dipy-mapmri) and needs the png file as its output in order to work. I am also not sure what else to change it to so it can be compatible with other dipy workflows. Let me know if there are any problems here and hopefully I can change it promptly.

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,114 @@
...
+        `data_file` and saves the png file to an output directory specified by 
+        `out_dir`.
+
+        Parameters
+        ----------
+        data_file : string
+            Path to the input volume.
+        data_bvecs : string
+            Path to the bvec files.
+        data_bvals :
```
skoudoro: add the datatype -> `data_bvals : string`

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,114 @@
...
+
+        Parameters
+        ----------
+        data_file : string
+            Path to the input volume.
+        data_bvecs : string
+            Path to the bvec files.
+        data_bvals : 
+            Path to the bval files.
+        small_delta :
```
skoudoro: add the datatype -> `small_delta : float`

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,114 @@
...
+            Small delta value used in generation of gradient table of provided
+            bval and bvec. (default: 0.0129)
+        big_delta :
```
skoudoro: add the datatype -> `big_delta : float`

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,114 @@
...
+            Big delta value used in generation of gradient table of provided 
+            bval and bvec. (default: 0.0218)
+        save_metrics :
```
skoudoro: add the datatype -> `save_metric : string`

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,114 @@
...
+            data = img.get_data()
+            bvals,bvecs = read_bvals_bvecs(bval, bvec)
+
+        
+            gtab = gradient_table(bvals=bvals, bvecs=bvecs,
+                                small_delta=small_delta,
+                                big_delta=big_delta, b0_threshold=50)
+
+
+            data_small = data[60:85, 80:81, 60:85]
```
skoudoro: Why do you crop the data? if it is really necessary, you have to warn the user that only a small part of its data will be processed. Otherwise, use the whole data.

arokem:
I would not make the output of this a png. If brain-life requires a png, I
would do this in the definition of the app there.

aarya22:
The reason I cropped the data and had the output be a png file is because I wanted this to be a simple preliminary test for making a brain-life application from a dipy workflow. I was thinking that eventually brain-life applications can simply become wrappers for the dipy workflow.

I agree that png files should not be an output of the workflow, but I am not sure what else I should return from Mapmir. I do not know enough about the model, if you have any suggestions I will gladly take them.

arokem:
We can talk about that when we meet tomorrow. Ideally, this would produce
something similar to what the tensor workflow produces (
https://github.com/nipy/dipy/blob/master/dipy/workflows/reconst.py#L25) --
nifti files that store model parameters and derived maps.

aarya22:
@skoudoro @arokem  I've managed to get the Travis build test work for most of the environments, but there are a few environments that do not provide the CVXPY package which is necessary for the workflow to finish. I'm not sure how to access the config so I can include the CVXPY package in all the environments.

arokem:
Instead of installing cvxpy on all the machines, skip the tests that
require cvxpy, if that's not installed:

Like here:

https://github.com/nipy/dipy/blob/master/dipy/reconst/tests/test_mapmri.py#L624

aarya22:
@skoudoro @arokem Travis has passed, but I'm not sure what the other two errors are about or what I am supposed to be changing there.

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,330 @@
...
 from dipy.workflows.workflow import Workflow
 
+from dipy.reconst import mapmri
+
+class ReconstMAPMRIFlow(Workflow):
+    @classmethod
+    def get_short_name(cls):
+        return 'mapmri'
+
+    def run(self, data_file, data_bvecs, data_bvals, model_type='', out_rtop='rtop.nii.gz',
```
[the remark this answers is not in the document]
arokem: And a keyword argument for the laplacian weighting as well.

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,330 @@
...
+                mapfit_aniso = map_model_aniso.fit(data)
+
+            elif model_type is 'both':
+                map_model_aniso = mapmri.MapmriModel(gtab, radial_order=radial_order,
+                                                     laplacian_regularization=True,
+                                                     laplacian_weighting=.05,
+                                                     positivity_constraint=True)
+                mapfit_aniso = map_model_aniso.fit(data)
+
+            # Not sure where to get affine or metadata?
```
arokem: Is this still an issue? Getting the affine? You can get it from the `data_file`.

As an aside: if you have questions, it's better to ask them in the PR interface than put them as comments in your code!

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,330 @@
...
+            if 'rtpp' in save_metrics:
+                r = mapfit_aniso.rtpp()
+                rtpp = nib.nifti1.Nifti1Image(r.astype(np.float32), affine)
+                nib.save(rtpp, out_rtpp)
+
+            logging.info('MAPMRI saved in {0}'.
+                         format(os.path.dirname(out_rtpp)))
+            
+
+class ReconstMAPMRILaplacian(ReconstMAPMRIFlow):
```
arokem: @Garyfallidis : is this the right way to do things? You create an interface for every setting of options to the model? I thought this was supposed to create a CLI that allows you to input key-word arguments?

dipy/workflows/reconst.py
aarya22: Ah thanks for pointing that out. I added those notes in the code to ask you at our weekly sessions and forgot to delete them.

arokem:
Closed in favor of #1395

PR #1351, opened 2017-10-07
From tables to h5py

Garyfallidis:
Switching from pytables to h5py. H5py still works with older numpy versions. pytables needs newer versions. Therefore h5py is easier to depend on. Also it seems that h5py gives better access to low level functions. Finally, testing in travis should be much simpler without the pytables dependency as that api has been changing quite a bit.

### DIFF
--- a/doc/dependencies.rst
+++ b/doc/dependencies.rst
@@ -7,9 +7,9 @@ Dependencies
 Depends on a few standard libraries: python_ (the core language), numpy_ (for
 numerical computation), scipy_ (for more specific mathematical operations),
 cython_ (for extra speed) and nibabel_ (for file formats; we require version 2.1
-or higher). Optionally, it can use python-vtk_ (for visualisation), pytables_
+or higher). Optionally, it can use python-vtk_ (for visualisation), h5py_
 (for handling large datasets), matplotlib_ (for scientific plotting), and
-ipython_ (for interaction with the code and its results). cvxopt_ (version
-1.1.7 or later) is required for some modules.
+ipython_ (for interaction with the code and its results). cvxpy is required for
+some modules.
 
 .. include:: links_names.inc
--- a/doc/dependencies.rst
+++ b/doc/dependencies.rst
@@ -6,9 +6,9 @@ Dependencies
 
 Depends on a few standard libraries: python_ (the core language), numpy_ (for
 numerical computation), scipy_ (for more specific mathematical operations),
-cython_ (for extra speed) and nibabel_ (for file formats; we require version 2.1
-or higher). Optionally, it can use python-vtk_ (for visualisation), h5py_
-(for handling large datasets), matplotlib_ (for scientific plotting), and
+cython_ (for extra speed), nibabel_ (for file formats; we require version 2.1
+or higher) and h5py_ (for handling large datasets. Optionally, it can use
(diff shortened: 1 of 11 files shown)


Garyfallidis:
Still a bit more work to do. But there is progress.

Garyfallidis:
@arokem I have only one bot failing now. The one with export DEPENDS="". What happens when DEPENDS is empty?. It installs minimum requirements by looking at requirements.txt or info.py or nothing? Can you check my .travis.yml file and suggest solutions?

arokem:
I think you might need to add a `pip install -r requirements.txt` in the
`install` block, for the requirements to be read from that file.

skoudoro:
@Garyfallidis, In this bot, `INSTALL_TYPE="setup"` so I think it does not use `requirement.txt` but you should add a `SetupDependency` in setup.py like you can see in this link: https://github.com/nipy/dipy/blob/master/setup.py#L150.

Garyfallidis:
Oh!! Good catch @skoudoro ! :) Let me fix this.

Garyfallidis:
Done. Let's wait for travis. I am not sure if the parameter heavy should be False or True.  The default option is False (and used for nibabel too) so I kept that. If anyone has other suggestions. Let me know.

Garyfallidis:
👍 we are game! Merge if you think it's good.

setup.py
```diff
@@ -150,6 +150,9 @@
 SetupDependency('nibabel', info.NIBABEL_MIN_VERSION,
                 req_type='install_requires',
                 heavy=False).check_fill(extra_setuptools_args)
+SetupDependency('h5py', info.H5PY_MIN_VERSION,
```
arokem: I don't understand this. Why are we not using the requirements file to set up requirements that don't make crazy loops (such as numpy) ?

doc/dependencies.rst
```diff
@@ -7,9 +7,9 @@ Dependencies
 Depends on a few standard libraries: python_ (the core language), numpy_ (for
 numerical computation), scipy_ (for more specific mathematical operations),
 cython_ (for extra speed) and nibabel_ (for file formats; we require version 2.1
-or higher). Optionally, it can use python-vtk_ (for visualisation), pytables_
+or higher). Optionally, it can use python-vtk_ (for visualisation), h5py_
```
arokem: Not an optional dependency, is it?

doc/dependencies.rst
```diff
@@ -7,9 +7,9 @@ Dependencies
...
 (for handling large datasets), matplotlib_ (for scientific plotting), and
-ipython_ (for interaction with the code and its results). cvxopt_ (version
-1.1.7 or later) is required for some modules.
+ipython_ (for interaction with the code and its results). cvxpy is required for
```
arokem: Good catch!

doc/dependencies.rst
Garyfallidis: Nope, corrected.

setup.py
Garyfallidis: I just went with what we have done for nibabel here. You may want to chat with @matthew-brett about this.  He most likely used this for a good reason.

doc/dependencies.rst
Garyfallidis: :)

arokem:
OK - we can follow up on the dependencies setup later. Merging!

PR #1355, opened 2017-10-13
WIP: ENH: UI Listbox

MarcCote:
I've refactored @ranveeraggarwal's code for the `FileSelectMenu2D` and made a ListBox that should be more flexible.

This new component supports multiselection via ctrl+click and shift+click. Also, the example in `viz_ui.py` shows how to hook a callback function such that it is called whenever the list selection changes.

What's missing is making `FileSelectMenu2D` use `ListBox2D`.

Here's a small demo:
listbox
**EDIT:** Updated demo.


dipy/viz/ui.py
```diff
@@ -1907,6 +1895,317 @@ def handle_events(self, actor):
...
+        self.view_offset = 0
+        self._build()
+        self.set_center(position)
+        self.update()
+
+        # Define events captured by this UI component.
+        self.add_callback(self.up_button.actor, "LeftButtonPressEvent",
+                          self.up_button_callback)
+        self.add_callback(self.down_button.actor, "LeftButtonPressEvent",
+                          self.down_button_callback)
```
ranveeraggarwal: In other UI elements I have separated the event handling logic in a separate function. This is just to keep the constructor a little clean.

ranveeraggarwal:
This is awesome! We can now remove `FileSelectMenuText2D` and modify `FileSelectMenu2D` to use this - once this is merged.

ranveeraggarwal:
On a further thought, `FileSelectMenu2D` is just a very specific use case of the `ListBox2D`, so we can remove it entirely and build a `FileMenu` element (a composite element) that uses the `ListBox2D` element with functions of the existing `FileSelectMenu2D`.

MarcCote:
Waiting after PR #1385 and PR #1362

MarcCote:
Rebased on top of #1441.

MarcCote:
I've updated the demo gif (see the first post).

skoudoro:
Let me test it quickly on windows @ranveeraggarwal. I will do it right now

skoudoro:
and should we merge #1441 first?

ranveeraggarwal:
Yeah, that makes sense @skoudoro

skoudoro:
After playing with this UI, it works like a charm,  +1 for merging when it is possible

dmreagan:
I'm having some hardware problems, so I'm not sure when I'll be able to look at this. If you all approve, go ahead and merge without me.

dipy/viz/ui.py
MarcCote: Good point.

Garyfallidis:
Needs rebase and update @MarcCote !

Garyfallidis:
@MarcCote can you fix the pep8 issues?

karandeepSJ:
I have added a scroll bar to this. 
scroll

dmreagan:
thanks @MarcCote now let's put back the FileSelectMenu

PR #1332, opened 2017-09-11
Update tracking_quick_start.py

justbennet:
Insert missing blank line at 35

PR #1357, opened 2017-10-14
Update tracking_quick_start.py rebase of #1332

Garyfallidis:
Insert missing blank line at 35


Garyfallidis:
This PR only enters an empty line in an example. That does not effect travis or our testing suite. So, going ahead with merge. Peace.

PR #1358, opened 2017-10-15
Det track workflow  rebased (merge)

Garyfallidis:
This a new PR that tries to correct issues with previous workflow PRs and puts down the necessary frameworks for the RecoBundles interfaces.
The main focus here is to provide a simple to use tracking workflow (a work that @matthieudumont started) however I realized that there are other important areas that need polishing together with the tracking workflow. I am providing some fixes and updates here.

https://github.com/nipy/dipy/pull/1324


Garyfallidis:
I need at least 3 people to check the workflows done here with their datasets. 
After installing the PR you should do 
```
dipy_[TAB]
```
to see the available commands.

Make sure first that dipy/bin is in your path. 

@jchoude, @mdesco, @arokem, @paolopost, @maryam, @stephanmeesters, @skoudoro can you look at this PR and test the cmds?

dipy/workflows/tests/test_tracking.py
```diff
@@ -0,0 +1,83 @@
...
+from os.path import join
+
+import nibabel as nib
+from nibabel.tmpdirs import TemporaryDirectory
+
+from dipy.data import get_data
+from dipy.io.image import save_nifti
+from dipy.workflows.mask import MaskFlow
+from dipy.workflows.reconst import ReconstCSDFlow
+from dipy.workflows.tracking import DetTrackSHFlow, DetTrackPAMFlow, \
```
arokem: Looks like maybe `DetTrackSHFlow` is not implemented yet?

bin/dipy_det_track
```diff
@@ -0,0 +1,9 @@
+#!python
```
arokem: Looks like some of these have `/usr/bin/env python` in the shebang, and others have just `python`. Which one should it be?

dipy/workflows/tests/test_tracking.py
Garyfallidis: I removed them from this PR.

dipy/workflows/tracking.py
```diff
@@ -0,0 +1,108 @@
...
+           Path to the peaks and metrics files. This path may contain
+           wildcards to use multiple masks at once.
+        stopping_files : string
+            Path of FA or other images used for stopping criteria for tracking.
+        seeding_files : string
+            A binary image showing where we need to seed for tracking.
+        stopping_thr : float, optional
+            Threshold applied to stopping volume's data to identify where
+            tracking has to stop. (default 0.25)
+        seed_density : int, optional
```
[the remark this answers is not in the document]
Garyfallidis: This is the number of equidistant points in each dimension of the voxel.
For example seed_density of 2 means 2x2x2 = 8 points per voxel
This can be problematic for some applications because just moving from 1 to 2 will mean 8 times more streamlines. Bago used this because he was trying to make absolutely sure that the points are distributed in an equal way around the voxel.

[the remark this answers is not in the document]
Garyfallidis: Was this the best way to do this? Probably not. We need to figure out a way were the seed density will actually mean number of points in the voxel and not the current definition. But let's do this in another PR. This is not a stopper for here.

bin/dipy_det_track
Garyfallidis: It should be just python. Corrected.

dipy/workflows/tracking.py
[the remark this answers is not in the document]
arokem: That's fine -- I don't think we should change the way it works (definitely not on this PR). But it would be good to provide a bit more detail in the documentation about what it does and what the units are (I understand it's not exactly a unitful quantity). Maybe write the example you provided in your answer here in the docstring?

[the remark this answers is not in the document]
Garyfallidis: Sure.

Garyfallidis:
Okay Travis is passing.

I need at least 3 people to check the workflows done here with your own datasets. 
After installing the PR you should do 
```
dipy_[TAB]
```
to see the available commands.

Make sure first that dipy/bin is in your path. 

@frheault @matthieudumont   @jchoude, @mdesco, @arokem, @paolopost, @maryam, @stephanmeesters, @skoudoro , @gabknight, @BramshQamar can you look at this PR and test the cmds?

We shall move rapidly. More PRs in on the way! :)

jchoude:
Very quick preliminary comment: the permissions for `dipy_det_track`, `dipy_info`, `dipy_reslice` do not include the `x` bit. This means that even with the `dipy/bin` path included in `$PATH`, tab-completion doesn't work. In fact, they simply are not executable.

bin/dipy_det_track
jchoude: I know we discussed about that a long time ago, but using a straight `python` doesn't work when using a virtualenv. Using the `/usr/bin/env python` correctly works. Is there a case where `/usr/bin/env` doesn't select the correct python?

dipy/workflows/reconst.py
```diff
@@ -295,8 +343,23 @@ def run(self, input_files, bvalues, bvectors, mask_files,
...
+            radius of cubic ROI in voxels (default 10)
+        fa_thr : float, optional
+            FA threshold for calculating the response function (default 0.7)
         frf : tuple, optional
-            Fiber response function to me mutiplied by 10**-4 (default: 15,4,4)
+            Fiber response function 15, 4, 4 to be mutiplied by 10**-4 
+            (default: None)
+        sh_order : int, optional
+            Spherical harmonics order (default 6) used in the CSA fit.
+        odf_to_sh_order : int, optional
```
jchoude: I don't know if that's just my setup, but the help message for this appears for the `--out_dir` argument, and the `out_dir` help message appears for this argument...

skoudoro:
can you change the description from `string` to `int`  at the following line : https://github.com/Garyfallidis/dipy/blob/99f802ab9f5c6dae83e992406cab75ea5aac64e9/dipy/workflows/segment.py#L47

bin/dipy_det_track
Garyfallidis: Not sure what you mean.

dipy/workflows/io.py
```diff
@@ -0,0 +1,95 @@
...
+        """
+
+        np.set_printoptions(3, suppress=True)
+
+        for input_path in input_files:
+            logging.info('------------------------------------------')
+            logging.info('Looking at {0}'.format(input_path))
+            logging.info('------------------------------------------')
+
+            if input_path.endswith('.nii') or input_path.endswith('.nii.gz'):
```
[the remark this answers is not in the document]
Garyfallidis: Corrected.

dipy/workflows/reconst.py
Garyfallidis: There was no problem with setup. When you create the docstrings of workflows they have to be in the same order as the input parameters. Should be corrected now.

Garyfallidis:
Okay @jchoude, @arokem, @skoudoro and all. This is ready to be merged on my side. But do try it. I increase coverage and added your suggestions. Let me know if something else is needed.

dipy/workflows/segment.py
```diff
@@ -40,7 +40,7 @@ def run(self, input_files, save_masked=False, median_radius=2, numpass=5,
             images are of 1x1x1 (mm^3) or higher resolution auto-cropping could
             reduce their size in memory and speed up some of the analysis.
             (default False)
-        vol_idx : string, optional
+        vol_idx : variable string, optional
```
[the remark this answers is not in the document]
Garyfallidis: Yes, this allows for multiple indices. Perhaps, variable int would be better. I can investigate this in another PR.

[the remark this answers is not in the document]
Garyfallidis: But please do try to see if it works for you.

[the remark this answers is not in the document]
Garyfallidis: Actually, I changed it to variable int in this PR.

dipy/workflows/reconst.py
```diff
@@ -187,79 +212,98 @@ def get_short_name(cls):
...
+        Basser, P., Pierpaoli, C., 1996. Microstructural and
+           physiological features of tissues elucidated by quantitative
+           diffusion-tensor MRI.  Journal of Magnetic Resonance 111, 209-219.
+        
+        Lin-Ching C., Jones D.K., Pierpaoli, C. 2005. RESTORE: Robust
+           estimation of tensors by outlier rejection. MRM 53: 1088-1095
+
+        """
         self.sigma = sigma
         self.jacobian = jacobian
```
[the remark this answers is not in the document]
Garyfallidis: Good catch. It seems like they are currently not being used. Let me check.

Garyfallidis:
Ready to be merged! :)

bin/dipy_det_track
jchoude: I mean that if the shebang is only `#!python`, calling this script doesn't work correctly when working from inside a virtualenv. If the shebang is changed to `/usr/bin/env python`, it works correctly.

arokem:
This was previously discussed here: https://github.com/nipy/dipy/issues/1139

jchoude:
Yes, thanks for finding it. Nevermind my comment, then.

Garyfallidis:
Okay so are we good then? Or you still want me to switch back to the old shebang?
Also @jchoude are the commands executing now correctly for you?

Garyfallidis:
Will wait for 2 days more. If nobody has issues will merge.

bin/dipy_csa
```diff
@@ -1,4 +1,4 @@
-#! /usr/bin/env python
```
gabknight: @Garyfallidis I might have missed the discussion here. Why removing "/usr/bin/env"? 

`#! /usr/bin/env python` works for me while `#!python` doesn't.  I execute the scripts from `dipy/bin`

dipy/workflows/tracking.py
```diff
@@ -0,0 +1,108 @@
...
+            Number of seeds per dimension inside voxel (default 1).
+            For example, seed_density of 2 means 8 regularly distributed points
+            in the voxel. And seed density of 1 means 1 point at the center
+            of the voxel.
+        use_sh : bool, optional
+            Use spherical harmonics saved in peask to find the
```
gabknight: peaks

dipy/workflows/tracking.py
```diff
@@ -0,0 +1,108 @@
...
+            out_tractogram='tractogram.trk'):
+
+        """ Workflow for deterministic using a saved peaks and metrics (PAM)
+        file as input.
+
+        Parameters
+        ----------
+        pam_files : string
+           Path to the peaks and metrics files. This path may contain
+           wildcards to use multiple masks at once.
```
gabknight: In my bash terminal, there is no space between the last word of a line and the first of the next line. e.g. for line 69-70 I have `... may containwildcards to use ....`

bin/dipy_det_track
```diff
@@ -3,7 +3,7 @@
 from __future__ import division, print_function
```
gabknight: A design comment. Since there will be other tracking bin soon in dipy, It might be interesting to have all of them grouped together by naming them `dipy_track_*`, this one being `dipy_track_det`. 

You just removed the `_reconst_` part for dti, csd, cas, so I might have missed that discussion... :)

bin/dipy_csa
gabknight: ok, there is a thread up there about this...

> I mean that if the shebang is only #!python, calling this script doesn't work correctly when working from inside a virtualenv. If the shebang is changed to /usr/bin/env python, it works correctly.

Garyfallidis: Yeap. Make sure you switch into the ```pip install --user -e . ```.

Garyfallidis:
Hi @gabknight, thanks for the review. I made the corrections. I also changed dipy_dti to dipy_fit_dti  and similarly for csa and csd. I had initially reconst but it was too long. Tried recon and rec as well but it didn't feel right too. Rec reminds me of recording. Lol. So, I think fit will do. We can always revisit that naming later. Waiting for our friend Travis.

arokem:
+1 for `fit`!

Should I change that on #1393 as well?

Garyfallidis:
Fine with me :)

skoudoro:
Thank you all for your feedback. merging !

PR #1362, opened 2017-10-20
Make more use of TextBlock2D constructor

dmreagan:


dipy/viz/ui.py
```diff
@@ -1079,8 +1079,15 @@ def __init__(self, width, height, text="Enter Text", position=(100, 10),
         """
         super(TextBox2D, self).__init__()
         self.text = text
-        self.actor = self.build_actor(self.text, position, color, font_size,
-                                      font_family, justification, bold, italic, shadow)
+        self.actor = TextBlock2D(self.text, font_size, font_family,
+                                 justification, bold, italic, shadow, color,
+                                 position)
+        if vtk.vtkVersion.GetVTKSourceVersion().split(' ')[-1] <= "6.2.0":
```
ranveeraggarwal: Can you create a function `_build()` and place the background logic inside it? I know it's kind of redundant, but we have been doing things that way for every UI element. Also, in future, if we are to add some more modifications to the text block at the text box level, we might end up complicating the constructor, so a separate build function makes more sense.

dipy/viz/ui.py
```diff
@@ -2297,65 +2254,18 @@ def __init__(self, font_size, position, file_select):
...
-
+        self.text_actor = TextBlock2D(position=position, font_size=font_size,
+                                      color=(0, 0, 0))
+        self.text_actor.actor.GetTextProperty().SetLineSpacing(1)
         if vtk.vtkVersion.GetVTKSourceVersion().split(' ')[-1] <= "6.2.0":
             pass
         else:
-            text_actor.actor.GetTextProperty().SetBackgroundColor(1, 1, 1)
-            text_actor.actor.GetTextProperty().SetBackgroundOpacity(1.0)
+            self.text_actor.actor.GetTextProperty().SetBackgroundColor(1, 1, 1)
```
ranveeraggarwal: Same as above, can you have the `build_actor` or `_build` method do this?
(I guess we should stick to the `_build` name for the method as it's private)

ranveeraggarwal:
Lgtm apart from a couple of modifications.

MarcCote:
I agree with @ranveeraggarwal let's stick with a `_build()` method that is responsible for the creation of the "visual" components (vtkActors or UIComponents).

dipy/viz/ui.py
dmreagan: Do you have other changes in mind that are specific to `TextBox2D` or `FileSelect2D`? If not, I think we could clean this up better by moving background color, background opacity, and line spacing to the `TextBlock2D` constructor. Those are probably common enough settings to belong there.

ranveeraggarwal: As of this moment, no.

dmreagan:
The comments from @ranveeraggarwal and @MarcCote concerned code that sets `TextBlock2D` properties. 5550e4b cleans up this configuration by adding those properties to the `TextBlock2D` constructor.

MarcCote:
I think background color for text was added only in VTK 7 (maybe 6.2 but not sure). That's why I manually add a Rectangle2D behind a TextBlock2D when I detect we are not using VTK7.

I don't mind branching from your PR and adding my stuff. I guess that will depend on what we decide in term of which version of VTKs should we support?

ranveeraggarwal:
This looks good. I would still like to move everything to a `_build` function rather than the constructor, but I guess in this case it'd be an overkill.

ranveeraggarwal:
Vertical justification is awesome, we need to merge this with the other PR.

dmreagan:
I was just about to try to port your background hack to this PR. But I'm getting stuck trying to set up an environment on Windows with VTK5 right now. I can go ahead and copy that stuff over, but I won't be able to test until I get this environment worked out.

ranveeraggarwal:
@dmreagan if you are on Windows 10, you can try setting up the environment on _Ubuntu on Windows_. I have tried it out and it works well for me.

dmreagan:
@MarcCote I have a VTK 5.10 environment set up that shows the same results as the Travis bot, and I'm struggling to figure out why that test fails. If your tests are passing on #1385, I think you should just go with that one and I'll add the changes from this PR later. This PR is really just cosmetic, so it shouldn't continue to hold up progress for #1385 and #1355.

MarcCote:
@dmreagan I had to re-record the test_ui_text_block on mine. I think the background actually impact the number of clicks. You might want to pick the .npz from my branch and try it out. If that still doesn't work, go ahead and review/merge PR #1385 .

skoudoro:
I just merged #1385, should we close this one? let me know @dmreagan.

dmreagan:
Closing this PR. May return to the idea once #1355 and #1441 are merged.

PR #1368, opened 2017-10-26
DOC: Update developers' affiliations.

jhlegarreta:
Update the DIPY developers' affilitations to reflect their current
institution.

Take advantage of the topic to fix accents for MA-Côté.

### DIFF
--- a/doc/developers.rst
+++ b/doc/developers.rst
@@ -9,7 +9,7 @@ The core development team consists of the following individuals:
 - **Ariel Rokem**, University of Washington, WA, USA
 - **Matthew Brett**, Birmingham University, Birmingham, UK
 - **Bago Amirbekian**, Databricks, San Francisco, CA, USA
-- **Omar Ocegueda**, Center for Research in Mathematics, Guanajuato, Mexico
+- **Omar Ocegueda**, Center for Research in Mathematics, Guanajuato, MX
 - **Stefan Van der Walt**, University of California, Berkeley, CA, USA
 - **Marc-Alexandre Côté**,  Maluuba, Sherbrooke, QC, CA
 - **Ian Nimmo-Smith**, retired, formerly at MRC Cognition and Brain Sciences Unit, Cambridge, UK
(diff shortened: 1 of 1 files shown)


jhlegarreta:
Updated the affiliations of the DIPY developers.

For those developers whose affiliation has changed from the one listed in previous version and having two (or more) affiliations I only kept one that is apparently the main one (e.g. Christopher Nguyen, Emmanuel Caruyer).

I ignore the current state/affiliation of a few (e.g. Vatsala Swaroop) either because I found no related info or because their most recent indexed reference is not that recent.

In other cases, I ignore where they point to (e.g. Endolith) so I cannot update the related information.

Finally, the only apparently related reference to Andrew Lawrence was that belonging to Cambridge. But I am not sure whether this is correct.

May be the core maintainers have a better knowledge of the current affiliation of these folks.

May be Serge should be added as well to the list.

Other than that, I am aware that this is hard to maintain. Some folks [[1]](https://github.com/RichardLitt/name-your-contributors) [[2]](https://github.com/kentcdodds/all-contributors) have developed scripts to gather the contributor's list automatically from the GitHub API, but that does not solve the affiliation issue (unless the API provides a way to get it if specified at all).

skoudoro:
Sure, you can add me :-). Thanks ! Currently, we use `git shortlog -ns` for updating this list. I do not think a new tool is necessary.

I will ask @garyfallidis if he knows more, otherwise, we can merge it.

jhlegarreta:
While adding Serge, looks like I squashed your commit 7879b49967359df8dd263c0f68b1dc854b31443a, @gabknight Sorry. Could you please add again your work? Thanks.

jhlegarreta:
OK, I cherry-picked it and pushed, so @gabknight no need to do anything.

@skoudoro ready to be merged once checks are OK.

gabknight:
thx @jhlegarreta

skoudoro:
Thanks @jhlegarreta and @gabknight !

PR #1369, opened 2017-10-26
MRG: add procedure for building, uploading wheels

matthew-brett:
Test and upload wheels before uploading the source distribution.


arokem:
Thanks!

PR #1375, opened 2017-11-14
Fix piesno type

nilgoyette:
I found a problem with piesno. The line 

    np.sum(data**2, axis=-1, dtype=np.float32)

doesn't do what you think it does. For a dwi for example, the `dtype` is `int16` and some values^2 are greater then 2^15.

The problem is that `data**2` is still in `int16`, then it's converted to a `float32` so this image is curiously bounded with the `int16` bounds :)

Moreover, I changed `axis=-1` to `axis=2` because we always know the right axis... it's 2.

### DIFF
--- a/dipy/denoise/nlmeans.py
+++ b/dipy/denoise/nlmeans.py
@@ -1,17 +1,15 @@
 from __future__ import division, print_function
 
+from datetime import datetime
+
+import nibabel as nib
 import numpy as np
 from dipy.denoise.denspeed import nlmeans_3d
-# from warnings import warn
-# import warnings
-
-# warnings.simplefilter('always', DeprecationWarning)
-# warn(DeprecationWarning("Module 'dipy.denoise.nlmeans' is deprecated,"
-#                         " use module 'dipy.denoise.non_local_means' instead"))
+from dipy.denoise.noise_estimate import estimate_sigma, piesno
 
 
 def nlmeans(arr, sigma, mask=None, patch_radius=1, block_radius=5,
-            rician=True, num_threads=None):
+            rician=False, num_threads=None):
     r""" Non-local means for denoising 3D and 4D images
 
     Parameters
@@ -45,12 +43,6 @@ def nlmeans(arr, sigma, mask=None, patch_radius=1, block_radius=5,
                       HARDI, MICCAI 2008
 
     """
-
-    # warn(DeprecationWarning("function 'dipy.denoise.nlmeans'"
-    #                        " is deprecated, use module "
-    #                        "'dipy.denoise.non_local_means'"
-    #                        " instead"))
-
     if arr.ndim == 3:
         sigma = np.ones(arr.shape, dtype=np.float64) * sigma
         return nlmeans_3d(arr, mask, sigma,
@@ -67,6 +59,7 @@ def nlmeans(arr, sigma, mask=None, patch_radius=1, block_radius=5,
             sigma = np.ones(arr.shape, dtype=np.float64) * sigma
 
         for i in range(arr.shape[-1]):
+            start = datetime.now()
             denoised_arr[..., i] = nlmeans_3d(arr[..., i],
                                               mask,
                                               sigma[..., i],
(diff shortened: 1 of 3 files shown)


arokem:
Good catch. Would you mind writing a test that fails without this fix, so that we can avoid reintroducing it? I'd say keep `axis=-1` for the cases where I feed in 2D arrays as data.

nilgoyette:
I'll see what I can do.

> for the cases where I feed in 2D arrays as data.

    # This method works on a 2D array with repetitions as the third dimension,
    # so process the dataset slice by slice.
    if data.ndim < 3:
        e_s = "This function only works on datasets of at least 3 dimensions."
        raise ValueError(e_s)

My change is in `_piesno_3D` which will always receive 3D data.

arokem:
Duh, of course. Thanks!

nilgoyette:
I don't know what to do with those 10 python 3.5 errors, they are all "ok".

    Expected:
        array([ 1. ,  1.5,  2.5,  3.5,  4.5,  5. ])
    Got:
        array([1. , 1.5, 2.5, 3.5, 4.5, 5. ])

Do you ignore them?

nilgoyette:
I tried to use the normal `test_piesno` image for the test but I always had the same answers. I created my own and took `N=2` to see a difference.

dipy/denoise/nlmeans.py
```diff
@@ -1,17 +1,15 @@
...
-
-# warnings.simplefilter('always', DeprecationWarning)
-# warn(DeprecationWarning("Module 'dipy.denoise.nlmeans' is deprecated,"
-#                         " use module 'dipy.denoise.non_local_means' instead"))
+from dipy.denoise.noise_estimate import estimate_sigma, piesno
 
 
 def nlmeans(arr, sigma, mask=None, patch_radius=1, block_radius=5,
-            rician=True, num_threads=None):
+            rician=False, num_threads=None):
```
arokem: Why did you change this default?

dipy/denoise/nlmeans.py
```diff
@@ -67,15 +59,49 @@ def nlmeans(arr, sigma, mask=None, patch_radius=1, block_radius=5,
...
                                               num_threads).astype(arr.dtype)
+            print(datetime.now() - start)
 
         return denoised_arr
 
     else:
         raise ValueError("Only 3D or 4D array are supported!", arr.shape)
+
+
+def main():
```
arokem: Please remove this code. I think that is causing the current failures.

dipy/denoise/nlmeans.py
```diff
@@ -67,15 +59,49 @@ def nlmeans(arr, sigma, mask=None, patch_radius=1, block_radius=5,
             sigma = np.ones(arr.shape, dtype=np.float64) * sigma
 
         for i in range(arr.shape[-1]):
+            start = datetime.now()
```
arokem: Please remove the timing code from here.

dipy/denoise/nlmeans.py
```diff
@@ -67,15 +59,49 @@ def nlmeans(arr, sigma, mask=None, patch_radius=1, block_radius=5,
...
             denoised_arr[..., i] = nlmeans_3d(arr[..., i],
                                               mask,
                                               sigma[..., i],
                                               patch_radius,
                                               block_radius,
                                               rician,
                                               num_threads).astype(arr.dtype)
+            print(datetime.now() - start)
```
arokem: And this as well

nilgoyette:
**Damn, sorry, my bad.** These change are unrelated to the current topic. I did a `commit -a` without thinking. Will remove tomorrow at work.

arokem:
Hi @nilgoyette : the failures you see on the last bot are probably unrelated to your work here. This bot tests new releases of numpy/scipy before they go live, so that we can fix things in advance. We'll need to fix these on a separate PR, before we can merge yours (which I believe is fine).

arokem:
OK - now that #1377 is merged, this failure should go away. 

Could you please rebase your branch on top of master?

nilgoyette:
I did a rebase/push as suggested but it doesn't seem to start the CI. Is there a way to force a tests-run?

arokem:
Maybe try again? I just merged #1378, so if you rebase again, that might sort this one out.

arokem:
Off we go! Thanks!

Issue #1376, opened 2017-11-16
New test failures with pre-release numpy


arokem:
On new PR #1375, we stumbled on these failures on the pre-release Travis bot: https://travis-ci.org/nipy/dipy/jobs/302550184

skoudoro:
Hi @arokem, Thanks for this information ! 

It seems to be the same issue as #1352 (print formatting with doctest)

We just need to add `from .testing import setup_test` where tests failed.

arokem:
Closed through #1377

PR #1377, opened 2017-11-16
FIX: update printing format for numpy 1.14

skoudoro:
This PR is for resolving #1376  due to the change [here](https://github.com/numpy/numpy/commit/734b907fc2f7af6e40ec989ca49ee6d87e21c495#diff-6c34f1538350ce17eb53b35d2f0282a1R512) on Numpy


skoudoro:
It seems to be ok @arokem. you can merge it even if I do not understand why Codecov is not happy.

PR #1347, opened 2017-10-04
Replacing fvtk by the new viz API

skoudoro:
The goal of this PR is to replace `fvtk` by the new `viz` API in all examples.  Most of the work is done.

Last work to do:
 
- [x]  create `tensor_slicer` in actor instead of using `fvtk.tensor`
- [x] replace `fvtk.contour` by something else. Maybe by #1165 from @kesshijordan in some case.

PR #1379, opened 2017-11-21
New streamlines API integration on dipy examples

### LINKED ISSUE #1339: Deprecate dipy.io.trackvis?
At the very least, we need to update this line in the documentation: https://github.com/nipy/dipy/blob/master/dipy/io/trackvis.py#L10

See for example conversation here: https://mail.python.org/pipermail/neuroimaging/2017-September/001525.html

An easy fix for now would be to replace that function with a call out to nibabel and a deprecation warning.

skoudoro:
The goal of this PR is to integrate the new streamlines API. it should resolve #1339.


dipy/io/streamline.py
```diff
@@ -0,0 +1,56 @@
+import nibabel as nib
+from nibabel.streamlines import Field
+from nibabel.orientations import aff2axcodes
+
+
+def save_trk(fname, streamlines, affine, vox_size=None, shape=None, header=None):
+    """ function Helper for saving trk files.
```
arokem: I don't think this is a helper function. 

How about just "Saves trk files"?

dipy/io/streamline.py
```diff
@@ -0,0 +1,56 @@
...
+
+    Parameters
+    ----------
+    fname : str
+        output trk filename
+    streamlines : list of 2D arrays
+        Each 2D array represents a sequence of 3D points (points, 3).
+    affine : array_like (4, 4)
+        The mapping from voxel coordinates to streamline points.
+    vox_size : array_like (3,)
```
arokem: optional. What is the default?

dipy/io/streamline.py
```diff
@@ -0,0 +1,56 @@
...
+        The sizes of the voxels in the reference image.
+    shape : array, shape (dim,)
```
arokem: Optional + what is the default

dipy/io/streamline.py
```diff
@@ -0,0 +1,56 @@
...
+        The shape of the reference image.
+    header : dict
```
arokem: Same here.

dipy/io/streamline.py
```diff
@@ -0,0 +1,56 @@
...
+        header[Field.VOXEL_ORDER] = "".join(aff2axcodes(affine))
+
+    tractogram = nib.streamlines.Tractogram(streamlines)
+    tractogram.affine_to_rasmm = affine
+    trk_file = nib.streamlines.TrkFile(tractogram, header=header)
+    nib.streamlines.save(trk_file, fname)
+
+
+def load_trk(filename):
+    """ function Helper for Loading trk files.
```
arokem: Same here -- how about "Loads trk files"?

dipy/io/trackvis.py
```diff
@@ -1,14 +1,27 @@
 import nibabel as nib
 import numpy as np
+import warnings
 
 from dipy.tracking import utils
 
+warnings.simplefilter('always', DeprecationWarning)
+warnings.warn("The `dipy.io.trackvis` module is deprecated as of version" +
+              " 0.13 of Dipy and will be removed in a future " +
```
arokem: 0.14? I believe 0.13 is already out!

dipy/io/trackvis.py
```diff
@@ -1,14 +1,27 @@
...
+              DeprecationWarning)
+
 
 def save_trk(filename, points, vox_to_ras, shape):
     """A temporary helper function for saving trk files.
 
     This function will soon be replaced by better trk file support in nibabel.
     """
+    warnings.warn("The `dipy.io.trackvis.save_trk` function is deprecated as of version" +
+                  " 0.13 of Dipy and will be removed in a future " +
```
arokem: Same here

doc/examples/fiber_to_bundle_coherence.py
```diff
@@ -142,8 +142,9 @@
 
 streamlines = LocalTracking(prob_dg, classifier, seeds, affine, step_size=.5)
 
-# Compute streamlines and store as a list.
-streamlines = list(streamlines)
+# Compute streamlines.
+from dipy.tracking.streamline import Streamlines
```
arokem: Maybe import from nibabel instead?

doc/examples/introduction_to_basic_tracking.py
```diff
@@ -96,12 +96,13 @@
 from dipy.tracking.local import LocalTracking
 from dipy.viz import fvtk
 from dipy.viz.colormap import line_colors
+from dipy.tracking.streamline import Streamlines
 
 # Initialization of LocalTracking. The computation happens in the next step.
-streamlines = LocalTracking(csa_peaks, classifier, seeds, affine, step_size=.5)
+streamlines_generator = LocalTracking(csa_peaks, classifier, seeds, affine, step_size=.5)
```
arokem: Great - much clearer this way

doc/examples/introduction_to_basic_tracking.py
```diff
@@ -96,12 +96,13 @@
...
 
 # Compute streamlines and store as a list.
-streamlines = list(streamlines)
+streamlines = Streamlines(streamlines_generator)
```
arokem: Might want to change the comment above this line

doc/examples/introduction_to_basic_tracking.py
```diff
@@ -187,11 +190,11 @@
 callosum.
 """
 
-streamlines = LocalTracking(prob_dg, classifier, seeds, affine,
+streamlines_generator = LocalTracking(prob_dg, classifier, seeds, affine,
                             step_size=.5, max_cross=1)
 
 # Compute streamlines and store as a list.
-streamlines = list(streamlines)
+streamlines = Streamlines(streamlines_generator)
```
arokem: Change the comment above this line too.

doc/examples/probabilistic_fiber_tracking.py
```diff
@@ -65,8 +66,8 @@
 pmf = fod.clip(min=0)
 prob_dg = ProbabilisticDirectionGetter.from_pmf(pmf, max_angle=30.,
                                                 sphere=small_sphere)
-streamlines = LocalTracking(prob_dg, classifier, seeds, affine, step_size=.5)
-save_trk("probabilistic_small_sphere.trk", streamlines, affine, labels.shape)
+streamlines_generator = LocalTracking(prob_dg, classifier, seeds, affine, step_size=.5)
+save_trk("probabilistic_small_sphere.trk", Streamlines(streamlines_generator), affine, labels.shape)
```
arokem: What happens if `save_trk` gets the generator? Maybe we should allow that and generate inside of `save_trk`?

doc/examples/tracking_tissue_classifier.py
```diff
@@ -121,10 +122,10 @@
          affine,
          labels.shape)
 
-streamlines = [sl for sl in all_streamlines_threshold_classifier]
+streamlines_generator = Streamlines(all_streamlines_threshold_classifier)
```
arokem: Here it's not the generator, but the streamlines object, right? So the name isn't accurate.

doc/examples/tracking_tissue_classifier.py
```diff
@@ -188,9 +189,9 @@
          affine,
          labels.shape)
 
-streamlines = [sl for sl in all_streamlines_binary_classifier]
+streamlines_generator = Streamlines(all_streamlines_binary_classifier)
 fvtk.clear(ren)
-fvtk.add(ren, fvtk.line(streamlines, line_colors(streamlines)))
+fvtk.add(ren, fvtk.line(streamlines_generator, line_colors(streamlines_generator)))
```
arokem: Same comment here -- should probably be `streamlines`, instead of `streamlines_generator`?

doc/examples/tracking_tissue_classifier.py
```diff
@@ -275,10 +276,10 @@
          affine,
          labels.shape)
 
-streamlines = [sl for sl in all_streamlines_act_classifier]
+streamlines_generator = Streamlines(all_streamlines_act_classifier)
```
arokem: Same comment here.

doc/examples/tracking_tissue_classifier.py
```diff
@@ -301,10 +302,10 @@
          affine,
          labels.shape)
 
-streamlines = [sl for sl in valid_streamlines_act_classifier]
+streamlines_generator = Streamlines(valid_streamlines_act_classifier)
```
arokem: And here

gabknight:
Thx @skoudoro. 

The support for generator would be very usefull for huge streamline file. A `fiber compression` parameter would also be very nice here. Saving compressed streamlines at the time of generating them would be ideal!

The Nibabel lib support both trk and tck formats. I suggest we take advantage of this. We should depricate `save/load_trk(.)` and create new `save/load_streamlines(.)`and use the file extension to determine the file format using `nib.streamlines.detect_format(.)` or `nib.streamlines.is_supported`. `save/*load_trk(.)` could be just a call to `save/load_streamlines(.)`.

What do you think?

dipy/io/trackvis.py
skoudoro: good catch! Thanks!

doc/examples/fiber_to_bundle_coherence.py
skoudoro: For a tutorial, I do not find really user friendly the syntax `from nibabel.streamlines import ArraySequence as Streamlines`

However, I do prefer the name `Streamlines` over `ArraySequence`. So I just imported from DIPY.

doc/examples/probabilistic_fiber_tracking.py
skoudoro: it works, I update this. Thanks !

doc/examples/tracking_tissue_classifier.py
skoudoro: good catch, Thanks !

skoudoro:
Thanks @arokem and @gabknight for this review and sorry for my late reply.

Concerning save_trk, the default implementation accepts streamlines generator so I made a mistake.

Concerning `fiber compression`, It is a good idea and this improvement deserves its own PR.

During this PR, I only focus on example. I plan to go deeper into Dipy later. Let me know if you want me to update something else.

skoudoro:
I would like to see #1347 merge before this one. There is some conflict between both so I need to rebase this one first.

Garyfallidis:
@gabknight having a load_streamlines function is interesting. But it is not that simple. The different file formats need different inputs. For example trk can have voxel_size as an input. Let's make independent functions load/save_trk or load/save_tck for now and let's first play a bit with these to see if we can make a common interface. However, what we need in reality is a properly designed new file format. But that is a different discussion.

gabknight:
@skoudoro I updated the tracking tests with the new streamline API in #1431. dipy/tracking/local/tests/test_tracking.py

skoudoro:
Good to know, Thanks @gabknight!

doc/examples/tracking_tissue_classifier.py
```diff
@@ -124,7 +125,7 @@
          affine,
          labels.shape)
 
-streamlines = [sl for sl in all_streamlines_threshold_classifier]
+streamlines = Streamlines(all_streamlines_threshold_classifier)
 
 window.clear(ren)
 ren.add(actor.line(streamlines, line_colors(streamlines)))
```
gabknight: Can you add the line ```if window.have_vtk:``` for the call to ```actor.line(...)```

skoudoro:
This one is ready to go, can you have a look @gabknight.
Concerning `missing figures` and  `renaming example files`, I think it is better to do it in a new PR because this one start to be big

Garyfallidis:
Hey @skoudoro can you make codecov/patch green too? This PR needs to be super solid as many others will depend on it. Thanks.

Garyfallidis:
This looks good. The changes are straightforward. I will go ahead and merge this. 
We need new PRs for a) correcting fornix dataset, b) use what @MarcCote suggested allow  LazyTractograms during saving. @arokem  the generators are working now in this PR for all operating systems. Let's move with the release! Thanks Serge!

PR #1384, opened 2017-11-28
NF - Particle Filtering Tractography (merge)

gabknight:
This PR replaces #1340, and follows the changes of PR #1342.

This PR add a new tissue classifier (CMC) and a new tractography algorithm (PFT). Both methods are described in [Girard et al., NImg, 2014].

CMC - new tissue Classifier. CmcTissueClassifier used probabilities and the tissue partial volume fraction maps to determine when the tracking stops. As it is very similar to ActTissueClassifier, both classes now extend an a new abstract class ConstrainedTissueClassifier.

PFT - new tractography method. ParticleFilteringTracking is inherited from the LocalTracking class. The particle filter is used to find a valid streamline path when the streamline being traced reach the state INVALIDPOINT.

An example is giving in doc/examples/particle_filtering_fiber_tracking.py

References
----------
Girard, G., Whittingstall, K., Deriche, R., & Descoteaux, M. Towards quantitative connectivity analysis: reducing tractography biases. NeuroImage, 98, 266-278, 2014.

### DIFF
--- a/dipy/tracking/local/__init__.py
+++ b/dipy/tracking/local/__init__.py
@@ -1,9 +1,15 @@
-from .localtracking import LocalTracking
-from .tissue_classifier import (
-    ActTissueClassifier, BinaryTissueClassifier, ThresholdTissueClassifier,
-    TissueClassifier)
 from .direction_getter import DirectionGetter
+from .localtracking import LocalTracking, ParticleFilteringTracking
+from .tissue_classifier import (ActTissueClassifier,
+                                BinaryTissueClassifier,
+                                CmcTissueClassifier,
+                                ConstrainedTissueClassifier,
+                                ThresholdTissueClassifier,
+                                TissueClassifier)
+
 from dipy.tracking import utils
 
-__all__ = ["ActTissueClassifier", "BinaryTissueClassifier", "LocalTracking",
-           "ThresholdTissueClassifier"]
+__all__ = ["ActTissueClassifier", "BinaryTissueClassifier",
+           "CmcTissueClassifier", "ConstrainedTissueClassifier",
+           "DirectionGetter", "LocalTracking", "ParticleFilteringTracking",
+           "ThresholdTissueClassifier", "TissueClassifier"]
--- a/dipy/tracking/local/localtrack.pyx
+++ b/dipy/tracking/local/localtrack.pyx
@@ -7,7 +7,7 @@ import numpy as np
 from .direction_getter cimport DirectionGetter
 from .tissue_classifier cimport(
     TissueClass, TissueClassifier, ConstrainedTissueClassifier,
-    TRACKPOINT, ENDPOINT, OUTSIDEIMAGE, INVALIDPOINT)
+    TRACKPOINT, ENDPOINT, OUTSIDEIMAGE, INVALIDPOINT, PYERROR)
 from dipy.tracking.local.interpolation cimport trilinear_interpolate4d_c
 from dipy.utils.fast_numpy cimport cumsum, where_to_insert
 
@@ -211,7 +211,7 @@ def pft_tracker(
         np.float_t[:, :] directions,
         double step_size,
         int pft_max_nbr_back_steps,
-        int pft_max_steps,
+        int pft_max_nbr_front_steps,
         int pft_max_trials,
         int particle_count,
         np.float_t[:, :, :, :] particle_paths,
@@ -248,8 +248,8 @@ def pft_tracker(
     pft_max_nbr_back_steps : int
         Number of tracking steps to back track before starting the particle
         filtering tractography.
-    pft_max_steps : int
-        Maximum number of steps for the particle filtering tractography.
+    pft_max_nbr_front_steps : int
+        Number of additional tracking steps to track.
     pft_max_trials : int
         Maximum number of trials for the particle filtering tractography
         (Prevents infinite loops).
@@ -290,7 +290,7 @@ def pft_tracker(
 
     i = _pft_tracker(dg, tc, seed, dir, vs, streamline,
                      directions, step_size, &tissue_class, pft_max_nbr_back_steps,
-                     pft_max_steps, pft_max_trials, particle_count,
+                     pft_max_nbr_front_steps, pft_max_trials, particle_count,
                      particle_paths, particle_dirs, particle_weights,
                      particle_states)
     return i, tissue_class
@@ -309,7 +309,7 @@ cdef _pft_tracker(DirectionGetter dg,
                   double step_size,
                   TissueClass * tissue_class,
                   int pft_max_nbr_back_steps,
-                  int pft_max_steps,
+                  int pft_max_nbr_front_steps,
                   int pft_max_trials,
                   int particle_count,
                   np.float_t[:, :, :, :] particle_paths,
@@ -317,8 +317,8 @@ cdef _pft_tracker(DirectionGetter dg,
                   np.float_t[:] particle_weights,
                   np.int_t[:, :, :] particle_states):
     cdef:
-        int i, pft_trial, pft_streamline_i, pft_nbr_back_steps, strl_array_len
-        int pft_nbr_steps
+        int i, pft_trial, pft_streamline_i, back_steps, front_steps
+        int strl_array_len
         double point[3]
         double voxdir[3]
         void (*step)(double* , double*, double) nogil
@@ -344,34 +344,40 @@ cdef _pft_tracker(DirectionGetter dg,
             copypoint(dir, &directions[i, 0])
             tissue_class[0] = tc.check_point_c(point)
             i += 1
-
         if tissue_class[0] == TRACKPOINT:
+            # The tracking continues normally
             continue
-        elif tissue_class[0] == ENDPOINT:
-            break
         elif tissue_class[0] == INVALIDPOINT:
             if pft_trial < pft_max_trials and i > 1:
-                pft_nbr_back_steps = min(i - 1, pft_max_nbr_back_steps)
-                pft_nbr_steps = min(pft_max_steps,
-                                    strl_array_len - i - pft_nbr_back_steps - 1)
-                pft_nbr_steps = max(pft_nbr_steps, 1)
-                i = _pft(streamline, i - pft_nbr_back_steps, directions, dg, tc,
-                         voxel_size, step_size, tissue_class, pft_nbr_steps,
-                         particle_count, particle_paths, particle_dirs,
-                         particle_weights, particle_states)
-
+                back_steps = min(i - 1, pft_max_nbr_back_steps)
+                front_steps = min(strl_array_len - i - back_steps - 1,
+                                  pft_max_nbr_front_steps)
+                front_steps = max(0, front_steps)
+                i = _pft(streamline, i - back_steps, directions, dg, tc,
+                         voxel_size, step_size, tissue_class,
+                         back_steps + front_steps, particle_count,
+                         particle_paths, particle_dirs, particle_weights,
+                         particle_states)
                 pft_trial += 1
                 # update the current point with the PFT results
                 copypoint(&streamline[i-1, 0], point)
                 copypoint(&directions[i-1, 0], dir)
 
-
-                if not tissue_class[0] == TRACKPOINT:
+                if tissue_class[0] != TRACKPOINT:
+                    # The tracking stops. PFT returned a valid stopping point
+                    # (ENDPOINT, OUTSIDEIMAGE) or failed to find one
+                    # (INVALIDPOINT, PYERROR)
                     break
             else:
+                # PFT was run more times than `pft_max_trials` without finding
+                # a valid stopping point. The tracking stops with INVALIDPOINT.
                 break
+        else:
+            # The tracking stops with a valid point (ENDPOINT, OUTSIDEIMAGE)
+            # or an invalid point (PYERROR)
+            break
 
-    if tissue_class[0] == OUTSIDEIMAGE:
+    if tissue_class[0] == OUTSIDEIMAGE or tissue_class[0] == PYERROR:
         i -= 1
     return i
 
@@ -402,7 +408,7 @@ cdef _pft(np.float_t[:, :] streamline,
         int s, p, j
 
     if pft_nbr_steps <= 0:
-        return INVALIDPOINT, streamline_i
+        return streamline_i
 
     for p in range(particle_count):
         copypoint(&streamline[streamline_i, 0], &particle_paths[0, p, 0, 0])
@@ -413,7 +419,7 @@ cdef _pft(np.float_t[:, :] streamline,
 
     for s in range(pft_nbr_steps):
         for p in range(particle_count):
-            if not particle_states[0, p, 0] == TRACKPOINT:
+            if particle_states[0, p, 0] != TRACKPOINT:
                 for j in range(3):
                     particle_paths[0, p, s, j] = 0
                     particle_dirs[0, p, s, j] = 0
@@ -423,6 +429,7 @@ cdef _pft(np.float_t[:, :] streamline,
 
             if dg.get_direction_c(point, dir):
                 particle_states[0, p, 0] = INVALIDPOINT
+                particle_weights[p] = 0
             else:
                 for j in range(3):
                     voxdir[j] = dir[j] / voxel_size[j]
--- a/dipy/tracking/local/localtrack.pyx
+++ b/dipy/tracking/local/localtrack.pyx
@@ -404,7 +404,7 @@ cdef _pft(np.float_t[:, :] streamline,
         double point[3]
         double dir[3]
         double voxdir[3]
-        double eps = np.finfo(np.float).eps
+        double eps = 1e-16
         int s, p, j
 
     if pft_nbr_steps <= 0:
@@ -448,51 +448,56 @@ cdef _pft(np.float_t[:, :] streamline,
         sum_weights = 0
         for p in range(particle_count):
             sum_weights += particle_weights[p]
-        sum_squared = 0
-        for p in range(particle_count):
-            particle_weights[p] = particle_weights[p] / sum_weights
-            sum_squared += particle_weights[p] * particle_weights[p]
 
-        # Resample the particles if the weights are too uneven.
-        # Particles with negligable weights are replaced by duplicates of
-        # those with high weigths through resampling
-        N_effective = 1. / sum_squared
-        if N_effective < particle_count / 10.:
-            # copy data in the temp arrays
-            for pp in range(particle_count):
-                for ss in range(pft_nbr_steps):
-                    copypoint(&particle_paths[0, pp, ss, 0],
-                              &particle_paths[1, pp, ss, 0])
-                    copypoint(&particle_dirs[0, pp, ss, 0],
-                              &particle_dirs[1, pp, ss, 0])
-                particle_states[1, pp, 0] = particle_states[0, pp, 0]
-                particle_states[1, pp, 1] = particle_states[0, pp, 1]
+        if sum_weights > 0:
+            sum_squared = 0
+            for p in range(particle_count):
+                particle_weights[p] = particle_weights[p] / sum_weights
+                sum_squared += particle_weights[p] * particle_weights[p]
 
-            # sample N new particle
-            cumsum(&particle_weights[0],
-                   &particle_weights[0],
-                   particle_count)
-            for pp in range(particle_count):
-                rdm_sample = random() * particle_weights[particle_count - 1]
-                p_source = where_to_insert(&particle_weights[0],
-                                           rdm_sample,
-                                           particle_count)
-                for ss in range(pft_nbr_steps):
-                    copypoint(&particle_paths[1, p_source, ss, 0],
-                              &particle_paths[0, pp, ss, 0])
-                    copypoint(&particle_dirs[1, p_source, ss, 0],
-                              &particle_dirs[0, pp, ss, 0])
-                particle_states[0, pp, 0] = particle_states[1, p_source, 0]
-                particle_states[0, pp, 1] = particle_states[1, p_source, 1]
-            for pp in range(particle_count):
-                particle_weights[pp] = 1. / particle_count
+            # Resample the particles if the weights are too uneven.
+            # Particles with negligible weights are replaced by duplicates of
+            # those with high weigths through resampling
+            N_effective = 1. / sum_squared
+            if N_effective < particle_count / 10.:
+                # copy data in the temp arrays
+                for pp in range(particle_count):
+                    for ss in range(pft_nbr_steps):
+                        copypoint(&particle_paths[0, pp, ss, 0],
+                                  &particle_paths[1, pp, ss, 0])
+                        copypoint(&particle_dirs[0, pp, ss, 0],
+                                  &particle_dirs[1, pp, ss, 0])
+                    particle_states[1, pp, 0] = particle_states[0, pp, 0]
+                    particle_states[1, pp, 1] = particle_states[0, pp, 1]
+
+                # sample N new particle
+                cumsum(&particle_weights[0],
+                       &particle_weights[0],
+                       particle_count)
+                for pp in range(particle_count):
+                    rdm_sample = random() * particle_weights[particle_count - 1]
+                    p_source = where_to_insert(&particle_weights[0],
+                                               rdm_sample,
+                                               particle_count)
+                    for ss in range(pft_nbr_steps):
+                        copypoint(&particle_paths[1, p_source, ss, 0],
+                                  &particle_paths[0, pp, ss, 0])
+                        copypoint(&particle_dirs[1, p_source, ss, 0],
+                                  &particle_dirs[0, pp, ss, 0])
+                    particle_states[0, pp, 0] = particle_states[1, p_source, 0]
+                    particle_states[0, pp, 1] = particle_states[1, p_source, 1]
+                for pp in range(particle_count):
+                    particle_weights[pp] = 1. / particle_count
 
     # update the streamline with the trajectory of one particle
     cumsum(&particle_weights[0],
            &particle_weights[0],
            particle_count)
-    rdm_sample = random() * particle_weights[particle_count - 1]
-    p = where_to_insert(&particle_weights[0], rdm_sample, particle_count)
+    if particle_weights[particle_count - 1] > 0:
+        rdm_sample = random() * particle_weights[particle_count - 1]
+        p = where_to_insert(&particle_weights[0], rdm_sample, particle_count)
+    else:
+        p = 0
 
     for s in range(1, particle_states[0, p, 1]):
         copypoint(&particle_paths[0, p, s, 0], &streamline[streamline_i + s, 0])
--- a/dipy/tracking/local/localtrack.pyx
+++ b/dipy/tracking/local/localtrack.pyx
@@ -217,7 +217,8 @@ def pft_tracker(
         np.float_t[:, :, :, :] particle_paths,
         np.float_t[:, :, :, :] particle_dirs,
         np.float_t[:] particle_weights,
-        np.int_t[:, :, :]  particle_states):
+        np.int_t[:, :]  particle_steps,
+        np.int_t[:, :]  particle_tissue_classes):
     """Tracks one direction from a seed using the particle filtering algorithm.
 
     This function is the main workhorse of the ``ParticleFilteringTracking``
@@ -261,8 +262,10 @@ def pft_tracker(
         Temporary array for directions followed by particles.
     particle_weights : array, float, 1d (particle_count)
         Temporary array for the weights of particles.
-    particle_states : array, float, (2, particle_count, 2)
-        Temporary array for the states of particles.
+    particle_steps : array, float, (2, particle_count)
+        Temporary array for the number of steps of particles.
+    particle_tissue_classes : array, float, (2, particle_count)
+        Temporary array for the tissue classes of particles.
 
     Returns
     -------
@@ -292,7 +295,7 @@ def pft_tracker(
                      directions, step_size, &tissue_class, pft_max_nbr_back_steps,
                      pft_max_nbr_front_steps, pft_max_trials, particle_count,
                      particle_paths, particle_dirs, particle_weights,
-                     particle_states)
+                     particle_steps, particle_tissue_classes)
     return i, tissue_class
 
 
@@ -315,7 +318,8 @@ cdef _pft_tracker(DirectionGetter dg,
                   np.float_t[:, :, :, :] particle_paths,
                   np.float_t[:, :, :, :] particle_dirs,
                   np.float_t[:] particle_weights,
-                  np.int_t[:, :, :] particle_states):
+                  np.int_t[:, :] particle_steps,
+                  np.int_t[:, :] particle_tissue_classes):
     cdef:
         int i, pft_trial, pft_streamline_i, back_steps, front_steps
         int strl_array_len
@@ -357,7 +361,7 @@ cdef _pft_tracker(DirectionGetter dg,
                          voxel_size, step_size, tissue_class,
                          back_steps + front_steps, particle_count,
                          particle_paths, particle_dirs, particle_weights,
-                         particle_states)
+                         particle_steps, particle_tissue_classes)
                 pft_trial += 1
                 # update the current point with the PFT results
                 copypoint(&streamline[i-1, 0], point)
@@ -398,7 +402,8 @@ cdef _pft(np.float_t[:, :] streamline,
           np.float_t[:, :, :, :] particle_paths,
           np.float_t[:, :, :, :] particle_dirs,
           np.float_t[:] particle_weights,
-          np.int_t[:, :, :] particle_states):
+          np.int_t[:, :] particle_steps,
+          np.int_t[:, :] particle_tissue_classes):
     cdef:
         double sum_weights, sum_squared, N_effective, rdm_sample
         double point[3]
@@ -414,12 +419,12 @@ cdef _pft(np.float_t[:, :] streamline,
         copypoint(&streamline[streamline_i, 0], &particle_paths[0, p, 0, 0])
         copypoint(&directions[streamline_i, 0], &particle_dirs[0, p, 0, 0])
         particle_weights[p] = 1. / particle_count
-        particle_states[0, p, 0] = TRACKPOINT
-        particle_states[0, p, 1] = 0
+        particle_tissue_classes[0, p] = TRACKPOINT
+        particle_steps[0, p] = 0
 
     for s in range(pft_nbr_steps):
         for p in range(particle_count):
-            if particle_states[0, p, 0] != TRACKPOINT:
+            if particle_tissue_classes[0, p] != TRACKPOINT:
                 for j in range(3):
                     particle_paths[0, p, s, j] = 0
                     particle_dirs[0, p, s, j] = 0
@@ -428,7 +433,7 @@ cdef _pft(np.float_t[:, :] streamline,
             copypoint(&particle_dirs[0, p, s, 0], dir)
 
             if dg.get_direction_c(point, dir):
-                particle_states[0, p, 0] = INVALIDPOINT
+                particle_tissue_classes[0, p] = INVALIDPOINT
                 particle_weights[p] = 0
             else:
                 for j in range(3):
@@ -436,14 +441,14 @@ cdef _pft(np.float_t[:, :] streamline,
                 fixed_step(point, voxdir, step_size)
                 copypoint(point, &particle_paths[0, p, s + 1, 0])
                 copypoint(dir, &particle_dirs[0, p, s + 1, 0])
-                particle_states[0, p, 0] = tc.check_point_c(point)
-                particle_states[0, p, 1] = s + 1
+                particle_tissue_classes[0, p] = tc.check_point_c(point)
+                particle_steps[0, p] = s + 1
                 particle_weights[p] *= 1 - tc.get_exclude_c(point)
                 if particle_weights[p] < eps:
                     particle_weights[p] = 0
-                if (particle_states[0, p, 0] == INVALIDPOINT and
+                if (particle_tissue_classes[0, p] == INVALIDPOINT and
                         particle_weights[p] > 0):
-                    particle_states[0, p, 0] = TRACKPOINT
+                    particle_tissue_classes[0, p] = TRACKPOINT
 
         sum_weights = 0
         for p in range(particle_count):
@@ -467,8 +472,9 @@ cdef _pft(np.float_t[:, :] streamline,
                                   &particle_paths[1, pp, ss, 0])
                         copypoint(&particle_dirs[0, pp, ss, 0],
                                   &particle_dirs[1, pp, ss, 0])
-                    particle_states[1, pp, 0] = particle_states[0, pp, 0]
-                    particle_states[1, pp, 1] = particle_states[0, pp, 1]
+                    particle_tissue_classes[1, pp] = \
+                            particle_tissue_classes[0, pp]
+                    particle_steps[1, pp] = particle_steps[0, pp]
 
                 # sample N new particle
                 cumsum(&particle_weights[0],
@@ -484,8 +490,9 @@ cdef _pft(np.float_t[:, :] streamline,
                                   &particle_paths[0, pp, ss, 0])
                         copypoint(&particle_dirs[1, p_source, ss, 0],
                                   &particle_dirs[0, pp, ss, 0])
-                    particle_states[0, pp, 0] = particle_states[1, p_source, 0]
-                    particle_states[0, pp, 1] = particle_states[1, p_source, 1]
+                    particle_tissue_classes[0, pp] = \
+                            particle_tissue_classes[1, p_source]
+                    particle_steps[0, pp] = particle_steps[1, p_source]
                 for pp in range(particle_count):
                     particle_weights[pp] = 1. / particle_count
 
@@ -499,8 +506,8 @@ cdef _pft(np.float_t[:, :] streamline,
     else:
         p = 0
 
-    for s in range(1, particle_states[0, p, 1]):
+    for s in range(1, particle_steps[0, p]):
         copypoint(&particle_paths[0, p, s, 0], &streamline[streamline_i + s, 0])
         copypoint(&particle_dirs[0, p, s, 0], &directions[streamline_i + s, 0])
-    tissue_class[0] = <TissueClass> particle_states[0, p, 0]
-    return streamline_i + particle_states[0, p, 1]
+    tissue_class[0] = <TissueClass> particle_tissue_classes[0, p]
+    return streamline_i + particle_steps[0, p]
--- a/dipy/tracking/local/localtrack.pyx
+++ b/dipy/tracking/local/localtrack.pyx
@@ -91,37 +91,12 @@ cdef inline void copypoint(double * a, double * b) nogil:
 def local_tracker(
         DirectionGetter dg,
         TissueClassifier tc,
-        np.float_t[:] seed,
+        np.float_t[:] seed_pos,
         np.float_t[:] first_step,
         np.float_t[:] voxel_size,
         np.float_t[:, :] streamline,
         double step_size,
         int fixedstep):
-    cdef:
-        int i
-        TissueClass tissue_class
-
-    if (seed.shape[0] != 3 or first_step.shape[0] != 3 or
-            voxel_size.shape[0] != 3 or streamline.shape[1] != 3):
-        raise ValueError()
-
-    i = _local_tracker(dg, tc, seed, first_step, voxel_size, streamline,
-                       step_size, fixedstep, &tissue_class)
-    return i, tissue_class
-
-
-@cython.boundscheck(False)
-@cython.wraparound(False)
-@cython.cdivision(True)
-cdef int _local_tracker(DirectionGetter dg,
-                        TissueClassifier tc,
-                        np.float_t[:] seed,
-                        np.float_t[:] first_step,
-                        np.float_t[:] voxel_size,
-                        np.float_t[:, :] streamline,
-                        double step_size,
-                        int fixedstep,
-                        TissueClass* tissue_class):
     """Tracks one direction from a seed.
 
     This function is the main workhorse of the ``LocalTracking`` class defined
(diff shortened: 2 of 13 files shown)


dipy/tracking/local/localtrack.pyx
```diff
@@ -208,3 +210,263 @@ cdef int _local_tracker(
...
+           particle_count)
+    rdm_sample = random() * particle_weights[0, particle_count - 1]
+    p = where_to_insert(& particle_weights[0, 0], rdm_sample, particle_count)
+
+    for s in range(particle_states[0, p, 1]):
+        for j in range(3):
+            streamline[streamline_i + s, j] = particle_paths[0, p, s, j]
+            directions[streamline_i + s, j] = particle_dirs[0, p, s, j]
+
+    ###TODO fix this
```
gabknight: @nilgoyette, do you know why cython doesn't let me do `tissue_class[0] = TissueClass(particle_states[0, p, 0])`
```
NameError: name 'TissueClass' is not defined
```
I do have
```
from .tissue_classifier cimport TissueClass
```

nilgoyette: It seems that a cython enum doesn't "exists". It's simply defining the constants in the enum. You can cast it to the right type though:

    tissue_class[0] = <TissueClass>particle_states[0, p, 0]

I tested and it compiles and all the tests pass. There's no way to check if the `particle_states` returns a valid `TissueClass` but hey, it's cython!

gabknight:
Thanks for the quick feedback @nilgoyette. That's was the line I needed :)

@StongeEtienne it would be awesome if you could test this PFT implementation with SET. You will notice the speedup!

skoudoro:
Hi @gabknight! 

Currently, I cannot test the code because the compilation failed after a `pip install --user -e .`

```
dipy\direction\probabilistic_direction_getter.c(2154): error C2433: '__pyx_f_4dipy_5utils_10fast_numpy_where_to_insert': 'inline' not permitted on data declarations
    dipy\direction\probabilistic_direction_getter.c(2155): error C2433: '__pyx_f_4dipy_5utils_10fast_numpy_cumsum': 'inline' not permitted on data declarations
dipy\direction\probabilistic_direction_getter.c(4993): warning C4267: 'function': conversion from 'size_t' to 'int', possible loss of data
    dipy\direction\probabilistic_direction_getter.c(5079): warning C4267: 'function': conversion from 'size_t' to 'int', possible loss of data
    dipy\direction\probabilistic_direction_getter.c(28306): warning C4244: 'initializing': conversion from 'double' to 'float', possible loss of data
    dipy\direction\probabilistic_direction_getter.c(28312): warning C4244: 'initializing': conversion from 'double' to 'float', possible loss of data
    warning: dipy\utils\fast_numpy.pxd:8:31: Declarations should not be declared inline.
    warning: dipy\utils\fast_numpy.pxd:13:23: Declarations should not be declared inline.
```

I will look more on this problem next week, but if you have any idea before that, let me know :-)

Env : Windows 10  - python3 - numpy 1.12.1 - cython 0.25.1 - Anaconda

skoudoro:
it seems this problem comes from #1342 and I missed it before.  I have to use Windows more often :-) 

Indeed, we can see this error on [nipy bots](https://nipy.bic.berkeley.edu/builders/dipy-bdist32-33/builds/347/steps/shell_10/logs/stdio) too.

Can you have a look too @nilgoyette? Thanks !

nilgoyette:
Simply removing the 4 `inline`s (declarations and implementations) and deleting `direction_getter.c` kind-of fixed it. I hesitate to create a PR because I have other errors

    $ python.exe setup.py sdist
    TypeError: Unicode-objects must be encoded before hashing
    $ pip install --user -e .
    UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe9 in position 1: invalid continuation byte

Maybe it will pass on your server. I'll send a patch when I'm on my work computer, if nobody does it before

dipy/tracking/local/localtrack.pyx
```diff
@@ -208,3 +210,254 @@ cdef int _local_tracker(
...
+        np.float_t[:,:,:,:] particle_dirs,
+        np.float_t[:,:] particle_weights,
+        np.int_t[:,:,:]  particle_states):
+    cdef:
+        int i
+        TissueClass tissue_class
+
+    if (seed.shape[0] != 3 or first_step.shape[0] != 3 or
+            voxel_size.shape[0] != 3 or streamline.shape[1] != 3):
+        raise ValueError()
```
nilgoyette: IMO, this should be in the python code, probably in local_tracking, or simply removed (replaced by documentation?) One time would be ok but these checks are uselessly done thousands of times.

dipy/tracking/local/localtrack.pyx
```diff
@@ -208,3 +210,254 @@ cdef int _local_tracker(
...
+        np.float_t[:,:] directions,
+        double step_size,
+        int pft_nbr_back_steps,
+        int pft_max_steps,
+        int pft_max_trial,
+        int particle_count,
+        np.float_t[:,:,:,:] particle_paths,
+        np.float_t[:,:,:,:] particle_dirs,
+        np.float_t[:,:] particle_weights,
+        np.int_t[:,:,:]  particle_states):
```
nilgoyette: 2 spaces. Below too. And there's several in `_pft`

dipy/tracking/local/localtrack.pyx
```diff
@@ -208,3 +210,254 @@ cdef int _local_tracker(
...
+                  int pft_max_trial,
+                  int particle_count,
+                  np.float_t[:,:,:,:] particle_paths,
+                  np.float_t[:,:,:,:] particle_dirs,
+                  np.float_t[:,:] particle_weights,
+                  np.int_t[:,:,:]  particle_states):
+
+    if (seed.shape[0] != 3 or first_step.shape[0] != 3 or
+            voxel_size.shape[0] != 3 or streamline.shape[1] != 3):
+        raise ValueError()
```
nilgoyette: Remove this please. I could understand in the `def` method but I don't see any advantage here.

dipy/tracking/local/localtrack.pyx
```diff
@@ -208,3 +210,254 @@ cdef int _local_tracker(
...
+    for j in range(3):
+        streamline[0, j] = point[j] = seed[j]
+        dir[j] = first_step[j]
+        vs[j] = voxel_size[j]
+    copypoint(dir, &directions[0, 0])
+
+    tissue_class[0] = TRACKPOINT
+    i = 0
+    while i < streamline.shape[0] - 1:
+        if dg.get_direction(point, dir):
```
nilgoyette: `get_direction_c` is the fast call (from cython to cython instead of cython to python)

dipy/tracking/local/localtrack.pyx
```diff
@@ -208,3 +210,254 @@ cdef int _local_tracker(
...
+          double step_size,
+          TissueClass* tissue_class,
+          int pft_nbr_steps,
+          int particle_count,
+          np.float_t[:,:,:,:] particle_paths,
+          np.float_t[:,:,:,:] particle_dirs,
+          np.float_t[:,:] particle_weights,
+          np.int_t[:,:,:]  particle_states):
+    cdef:
+        double sum_weights, sum_squared, N_effective, rdm_sample#
```
nilgoyette: `#`?

dipy/tracking/local/localtrack.pyx
```diff
@@ -208,3 +210,254 @@ cdef int _local_tracker(
...
+        for p in range(particle_count):
+            if not particle_states[0, p, 0] == TRACKPOINT:
+                for j in range(3):
+                    particle_paths[0, p, s, j] = 0
+                    particle_dirs[0, p, s, j] = 0
+                continue  # move to the next particle
+            for j in range(3):
+                point[j] = particle_paths[0, p, s, j]
+                dir[j] = particle_dirs[0, p, s, j]
+            if dg.get_direction(point, dir):
```
nilgoyette: `get_direction_c`

dipy/tracking/local/localtrack.pyx
```diff
@@ -208,3 +210,254 @@ cdef int _local_tracker(
...
+    cdef:
+        double sum_weights, sum_squared, N_effective, rdm_sample#
+        double point[3], dir[3], vs[3], voxdir[3]
+        int s, p, j
+
+    if pft_nbr_steps <= 0:
+        return INVALIDPOINT, streamline_i
+
+    for j in range(3):
+        vs[j] = voxel_size[j]
```
nilgoyette: It's done everywhere like this but... isn't this a `copypoint`?

dipy/tracking/local/tests/test_tracking.py
```diff
@@ -260,7 +261,75 @@ def allclose(x, y):
         npt.assert_(np.allclose(sl, expected[1]))
 
 
-def test_maximum_deterministic_tracker():
+def test_ParticleFilteringTractography():
+    """This tests that the ParticleFilteringTracking produces
+    more streamlines connecting the gray matter then LocalTracking.
```
jchoude: then -> than

dipy/tracking/local/tests/test_tracking.py
```diff
@@ -260,7 +261,75 @@ def allclose(x, y):
...
+    """
+    #sphere = HemiSphere.from_sphere(unit_octahedron)
```
jchoude: If not used, should be removed.

dipy/tracking/local/tissue_classifier.pyx
```diff
@@ -95,7 +95,78 @@ cdef class ThresholdTissueClassifier(TissueClassifier):
             return ENDPOINT
 
 
-cdef class ActTissueClassifier(TissueClassifier):
+cdef class ConstrainedTissueClassifier(TissueClassifier):
+    r"""
+    Abstract class that takes as input included and excluded tissue maps.
+    The 'include_map' defines when the streamline reached a 'valid' stopping
+    region (e.g. gray matter partial volume estimation (PVE) map) and the
```
jchoude: In some cases, `gray` is used, whereas `grey` is used in other cases. Should be uniform.

doc/examples/particle_filtering_fiber_tracking.py
```diff
@@ -0,0 +1,137 @@
+"""
+=================================================
+Particle Filtering Tractography
+=================================================
+Particle Filtering Tractography (PFT) [Girard2014]_ uses tissue partial
+volume estimation (PVE) to reconstruct tracjectories connecting the gray matter,
+and not incorrectly stopping in the white matter or in the corticospinal fluid.
+It relies on a tissue classifier that identify the tissue where the streamline
```
jchoude: identify -> identifies

doc/examples/particle_filtering_fiber_tracking.py
```diff
@@ -0,0 +1,137 @@
...
+stopped. If the streamline correctly stopped in the gray matter, the trajectory
+is keep. If the streamline incorrecly stopped in the white matter or in the
```
jchoude: keep -> kept

doc/examples/particle_filtering_fiber_tracking.py
```diff
@@ -0,0 +1,137 @@
...
+                                               sphere=default_sphere)
+
+
+"""
+CMC/ACT Tissue Classifiers
+---------------------
+Continuous map criterion (CMC) [Girard2014]_ and Anatomically-constrained
+tractography (ACT) [Smith2012]_ both uses PVEs information from
+anatomical images to determine when the tractography stops.
+The both tissue classifier use a trilinear interpolation
```
jchoude: Remove the first "The". classifier -> classifiers

doc/examples/particle_filtering_fiber_tracking.py
```diff
@@ -0,0 +1,137 @@
...
+at the tracking position. CMC tissue classifier uses a probability derived from
+the PVE maps to determine if the streamline reach a 'valid' or 'invalid' region.
```
jchoude: reach -> reaches

doc/examples/particle_filtering_fiber_tracking.py
```diff
@@ -0,0 +1,137 @@
...
+ACT uses a fixed threshold on the PVE maps. Both tissue classifier can be used
```
jchoude: classifier -> classifiers

doc/examples/particle_filtering_fiber_tracking.py
```diff
@@ -0,0 +1,137 @@
...
+in conjunction PFT. In this exemple, we used CMC.
```
jchoude: conjunction WITH PFT.

exemple -> example

dipy/tracking/local/localtrack.pyx
```diff
@@ -208,3 +210,254 @@ cdef int _local_tracker(
...
+    cdef:
+        int i, pft_trial, pft_streamline_i, pft_nbr_steps
+        double point[3], dir[3], vs[3], voxdir[3]
+
+        void (*step)(double * , double*, double) nogil
+    pft_trial = 0
+
+    for j in range(3):
+        streamline[0, j] = point[j] = seed[j]
+        dir[j] = first_step[j]
```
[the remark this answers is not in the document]
gabknight: `first_step` and `voxel_size` are of type `float_t[:]`, while `copypoint(.,.)` needs `double *`. Is there a more elegant way to do it?

dipy/tracking/local/localtrack.pyx
gabknight: @jchoude had the same comment below.

here`voxel_size` is of type float_t[:], while copypoint(.,.) needs double *. Is there a more elegant way to do it?

I'm changing all the `for j in range(3):` everywhere with types `double*`. Thx.

dipy/tracking/local/localtrack.pyx
gabknight: This is done once per streamline. Each of these parameters is not specific to the tracker, but to the streamline that will be track. Moving it to the python code would be the same number of calls. Wouldn't 1,000,000 calls (streamlines) to this be negligible? What do you think?

dipy/tracking/local/__init__.py
```diff
@@ -1,9 +1,15 @@
...
-    ActTissueClassifier, BinaryTissueClassifier, ThresholdTissueClassifier,
-    TissueClassifier)
 from .direction_getter import DirectionGetter
+from .localtracking import LocalTracking, ParticleFilteringTracking
+from .tissue_classifier import (ActTissueClassifier,
+                                BinaryTissueClassifier,
+                                CmcTissueClassifier,
+                                ConstrainedTissueClassifier,
+                                ThresholdTissueClassifier,
+                                TissueClassifier)
```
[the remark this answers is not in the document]
gabknight: I added it to `__all__`. Thx

dipy/tracking/local/tissue_classifier.pyx
gabknight: fixed! thx.

gabknight:
@skoudoro Do you have any idea where this could come from? This happens only with python 3.

```
ERROR: dipy.reconst.tests.test_dti.test_mask
    self.test(*self.arg)
  File "/home/travis/build/nipy/dipy/venv/lib/python3.5/site-packages/dipy/reconst/tests/test_dti.py", line 520, in test_mask
    assert_almost_equal(dtifit_w_mask.S0_hat[0, 0, 0], dtifit.S0_hat[0, 0, 0])
  File "/opt/python/3.5.4/lib/python3.5/unittest/case.py", line 870, in assertAlmostEqual
    if round(abs(second-first), places) == 0:
```

skoudoro:
Not yet @gabknight. I try to figure out what are the recent changes

nilgoyette:
The line just above looks identical, but there's an important difference

    # dtifit_w_mask.fa[0, 0, 0] == numpy.float64
    assert_almost_equal(dtifit_w_mask.fa[0, 0, 0], dtifit.fa[0, 0, 0])

    # dtifit_w_mask.S0_hat[0, 0, 0] == numpy.ndarray, (1,)
    assert_almost_equal(dtifit_w_mask.S0_hat[0, 0, 0], dtifit.S0_hat[0, 0, 0])

I tested more and using `assert_almost_equal` is not a good idea with `numpy.ndarray` because it **almost** aways crash with

    ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

I say _almost_ because in the case where the shape of the array is `(1,)`, it crashes with 

No, I don't know why. But at least we know that it can be fixed

    assert_almost_equal(dtifit_w_mask.S0_hat[0, 0, 0, 0], dtifit.S0_hat[0, 0, 0, 0])

arokem:
Since this is currently affecting all our PRs, I've posted an issue here: #1399, so we can centralize the discussion.

dipy/tracking/local/localtrack.pyx
nilgoyette: We don't know how well the branch prediction is working but we can probably assume it's working as well as it would on a C program, so it's 4 conditions for each streamlines. One of them is checking a constant (`voxel_size`) Another is checking is checking a 2D array (`streamline`) with a constant width. The 2 others are a 3D point (`seed`) and a 3D vector (`first_step`), which seems obvious.

My opinion on this is:
- Cython is horrible but fast, lets not do anything useless
- Duck Typing can mean different things, but there's a concept I often see in Python "The most Pythonic way to check the type of an object is... not to check it." 

But, hey, it's up to you! As you wrote, it's negligible compared to the rest of the processing :)

dipy/tracking/local/localtrack.pyx
nilgoyette: Thank you for the `copy_point` modif, it's much better. For `voxel_size`, it would be better if it was converted to float64 one time in the `main()`. Again, it's probably negligible, but it's a waste anyway.

skoudoro:
Hi @gabknight,

On windows, I still have the previous issue (below).  Can you rebase?  it should fix this (via #1386) and the `__round__` problem (via #1400)

Thanks ! 

```
dipy\direction\probabilistic_direction_getter.c(2154): error C2433: '__pyx_f_4dipy_5utils_10fast_numpy_where_to_insert': 'inline' not permitted on data declarations
    dipy\direction\probabilistic_direction_getter.c(2155): error C2433: '__pyx_f_4dipy_5utils_10fast_numpy_cumsum': 'inline' not permitted on data declarations
dipy\direction\probabilistic_direction_getter.c(4993): warning C4267: 'function': conversion from 'size_t' to 'int', possible loss of data
    dipy\direction\probabilistic_direction_getter.c(5079): warning C4267: 'function': conversion from 'size_t' to 'int', possible loss of data
    dipy\direction\probabilistic_direction_getter.c(28306): warning C4244: 'initializing': conversion from 'double' to 'float', possible loss of data
    dipy\direction\probabilistic_direction_getter.c(28312): warning C4244: 'initializing': conversion from 'double' to 'float', possible loss of data
    warning: dipy\utils\fast_numpy.pxd:8:31: Declarations should not be declared inline.
    warning: dipy\utils\fast_numpy.pxd:13:23: Declarations should not be declared inline.
```

gabknight:
Thank @skoudoro. 

How should I proceed here? There are tons of conflicts if I `git rebase` and no conflict if I `git merge`. But if I `git merge` and update this branch, all commits made since June 2016 (first commit of this PR) will appear in this PR. If I `git merge`, then make a new PR, only the commits listed above will be shown. Is there a way to do the merge, update this branch and only show commits no already in master?

arokem:
This might be the way:
https://til.hashrocket.com/posts/ce7bff8134-accept-your-own-changes-during-git-rebase

gabknight:
thanks @arokem. The rebased didn't work. I managed to do it with a merge following https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch

tldr; updating the base branch of the PR updates the target branch state of the PR, but this leaves the log entries above.

dipy/tracking/local/localtrack.pyx
gabknight: Thx. I don't know how this got there.

jchoude:
To me, this seems good. Tested the merged version against the classical implementation, and everything works as expected.

@gabknight is there something specific you would like me to recheck since your last merge from master?

gabknight:
Thank @jchoude for the testing. It works as expected on my side too. Although, I tested it on only a few cases.  @StongeEtienne it would be great if you can give it a try with SET. 

Otherwise, I'm happy with the current state of the PR.

dipy/tracking/local/localtrack.pyx
```diff
@@ -88,39 +89,136 @@ cdef inline void copypoint(double * a, double * b) nogil:
...
+        size_t i
         TissueClass tissue_class
+        double dir[3]
+        double vs[3]
+        double seed[3]
 
-    if (seed.shape[0] != 3 or first_step.shape[0] != 3 or
+    if (seed_pos.shape[0] != 3 or first_step.shape[0] != 3 or
             voxel_size.shape[0] != 3 or streamline.shape[1] != 3):
         raise ValueError()
```
skoudoro: Can you add a message to this `ValueError`? Thanks!

dipy/tracking/local/localtrack.pyx
```diff
@@ -131,80 +229,271 @@ cdef int _local_tracker(
...
+        double vs[3]
+        double seed[3]
 
-    if fixedstep:
-        step = fixed_step
-    else:
-        step = step_to_boundary
+    if (seed_pos.shape[0] != 3 or first_step.shape[0] != 3 or
+            voxel_size.shape[0] != 3 or streamline.shape[1] != 3):
+        raise ValueError()
```
skoudoro: Can you add a message to this ValueError? Thanks!

doc/examples/particle_filtering_fiber_tracking.py
```diff
@@ -0,0 +1,137 @@
...
+
+from dipy.data import (read_stanford_labels, default_sphere,
+                       read_stanford_pve_maps)
+from dipy.direction import ProbabilisticDirectionGetter
+from dipy.io.trackvis import save_trk
+from dipy.reconst.csdeconv import (ConstrainedSphericalDeconvModel,
+                                   auto_response)
+from dipy.tracking.local import LocalTracking, ParticleFilteringTracking
+from dipy.tracking import utils
+from dipy.viz import fvtk
```
skoudoro: `fvtk` will be deprecated, can you use the new viz API : 

`from dipy.viz import window, actor`

Garyfallidis:
Hey @gabknight, we want to merge this nice PR. Hopefully this coming week. Can you look into the requested changes? Thanks!

dipy/tracking/local/tests/test_tracking.py
```diff
@@ -260,7 +261,75 @@ def allclose(x, y):
...
+
+    # Test with wrong tissueclassifier type
+    tc_bin = BinaryTissueClassifier(simple_wm)
+
+    def test_PFT_init():
+        pft_streamlines = ParticleFilteringTracking(dg, tc_bin, seeds,
+                                                    np.eye(4), 0.2)
+        pft_streamlines = ParticleFilteringTracking(dg, tc_bin, seeds,
+                                                    np.eye(4), 0.2,
+                                                    pft_nbr_steps=0)
```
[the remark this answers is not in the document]
gabknight: I wanted to raise the `ValueError` for an invalid number of pft steps. That was incorrectly done. Fixed now, with additional tests.

gabknight:
Thanks @skoudoro, I addressed all your comments in the commit above. @Garyfallidis if there is no further comment, this PR can be merged!

Thanks @nilgoyette @jchoude @skoudoro @arokem @Garyfallidis for your feedback and contributions.

dipy/tracking/local/localtrack.pyx
```diff
@@ -208,3 +199,301 @@ cdef int _local_tracker(
...
+        double voxdir[3]
+        void (*step)(double* , double*, double) nogil
+
+    copypoint(seed, point)
+    copypoint(seed, &streamline[0,0])
+    copypoint(dir, &directions[0, 0])
+
+    tissue_class[0] = TRACKPOINT
+    pft_trial = 0
+    i = 0
```
nilgoyette: I have a hard time following `i`. If I understand correctly, It represents the real number of points in `streamline` (and not `len(streamline)` which never change). In that case, why is it not starting at `1`? Would it be simpler instead of all the `i += 1; break;`?

Lets say `get_direction` returns False (which is good, wtf?), we add the current point and i++, then ``check_point` returns `ENDPOINT`, i++ again and break. This is strange for the reader to see another i++ here because you haven't added a new point.

dipy/tracking/local/localtrack.pyx
```diff
@@ -208,3 +199,301 @@ cdef int _local_tracker(
...
+        if dg.get_direction_c(point, dir):
+            # no valid diffusion direction to follow
+            tissue_class[0] = INVALIDPOINT
+        else:
+            for j in range(3):
+                voxdir[j] = dir[j] / voxel_size[j]
+            i += 1
+            fixed_step(point, voxdir, step_size)
+            copypoint(point, &streamline[i, 0])
+            copypoint(dir, &directions[i, 0])
```
nilgoyette: Why are you keeping all directions? You don't seem to be using them in `ParticleFilteringTracking`, or anywhere else.

nilgoyette: Ah, ok, you are putting some of them in `particle_dirs`, so keeping them all is simpler.

nilgoyette: Dammit, sorry, I answer too fast then I see something else :) It seems you are filling `particle_dirs` using only the last direction, so I still don't understand why you are keeping all directions.

    for p in range(particle_count):
        copypoint(&directions[streamline_i, 0], &particle_dirs[0, p, 0, 0])

gabknight: I need the direction for the backtracking step. Before starting PFT, we backtrack a few steps (`pft_nbr_back_steps`, e.g. 2mm), then we start the tracking with the particle filter. At this time, we need to know the direction followed to reach the current position. 

Because of the backtracking, streamline segments obtained with PFT can overlap, thus I need to save the direction of each of the streamline point. Also, the backtracking distance is not always the same (e.g. shorter if the streamline is less than 2mm).

dipy/tracking/local/localtrack.pyx
```diff
@@ -208,3 +199,301 @@ cdef int _local_tracker(
...
+    pft_max_trials : int
+        Maximum number of trials for the particle filtering tractography
+        (Prevents infinite loops).
+    particle_count : int
+        Number of particles to use in the particle filter.
+    particle_paths : array, float, 4d, (2, particle_count, pft_max_steps, 3)
+        Temporary array for paths followed by all particles.
+    particle_dirs : array, float, 4d, (2, particle_count, pft_max_steps, 3)
+        Temporary array for directions followed by particles.
+    particle_weights : array, float, 2d, (2, particle_count)
```
nilgoyette: I don't understand why you need the starting `(2` here. I see that you used it in the Resampling part, but all weights are copied from 0 to 1, cumsum+other in 1, then you overwrite 0 with new data. Why couldn't you work in 0 directly?

It seems to be required for the other temporary arrays though so I won't complain :)

dipy/tracking/local/localtrack.pyx
gabknight: Here. `i+=1` is done when we want to keep the last point (`tissue_class == ENDPOINT or INVALIDPOINT`), not when `tissue_class == OUTSIDEIMAGE`.  I tried to clarify the code. Thx.

For the `get_direction(.)`, It is a function name issue here. It should be called `update_direction(point, direction)`, returning 0 if there is no error, not 0 otherwise. With the current name, I would rather be having it returning True when the get_direction succeed. We can open an issue for this to be done in an other PR.

gabknight:
Thanks for your comment on `i += 1` @nilgoyette. I think what going on is more clear now. It allowed me to notice a bug when the streamline `maxlen` is smaller than backtracking distance. Now it is fixed and tested. Looking around for the bug I also notice a very corner case error that was not properly handled by both the `ConstrainedTissueClassifier` and the `PmfGen` objects.

dipy/tracking/local/localtrack.pyx
```diff
@@ -208,3 +199,296 @@ cdef int _local_tracker(
...
+    for p in range(particle_count):
+        copypoint(&streamline[streamline_i, 0], &particle_paths[0, p, 0, 0])
+        copypoint(&directions[streamline_i, 0], &particle_dirs[0, p, 0, 0])
+        particle_weights[p] = 1. / particle_count
+        particle_states[0, p, 0] = TRACKPOINT
+        particle_states[0, p, 1] = 0
+
+    for s in range(pft_nbr_steps):
+        for p in range(particle_count):
+            if not particle_states[0, p, 0] == TRACKPOINT:
```
nilgoyette: `not .. == ..`?

dipy/tracking/local/localtrack.pyx
```diff
@@ -208,3 +199,296 @@ cdef int _local_tracker(
...
+            for j in range(3):
+                voxdir[j] = dir[j] / voxel_size[j]
+
+            fixed_step(point, voxdir, step_size)
+            copypoint(point, &streamline[i, 0])
+            copypoint(dir, &directions[i, 0])
+            tissue_class[0] = tc.check_point_c(point)
+            i += 1
+
+        if tissue_class[0] == TRACKPOINT:
```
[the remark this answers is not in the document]
nilgoyette: I tested more and I don't understand what's going on. I  added a print on line 347

    print 'Loop', i, point, class

and I obtain

    Loop 2 0.669794495731 2.11007159921 1.86826715409 OUTSIDEIMAGE
    Loop 3 0.589588991463 1.97014319842 1.98653430818 OUTSIDEIMAGE
    Loop 4 0.57799543161 1.82547363239 2.12414355945 OUTSIDEIMAGE
    Loop 5 0.566401871757 1.68080406636 2.26175281072 OUTSIDEIMAGE
    Loop 6 0.554808311905 1.53613450033 2.39936206199 OUTSIDEIMAGE
    Loop 7 0.543214752052 1.3914649343 2.53697131326 ENDPOINT
    END ENDPOINT 7
    Loop 2 0.830205504269 2.38992840079 1.63173284591 OUTSIDEIMAGE
    Loop 3 0.910411008537 2.52985680158 1.51346569182 INVALIDPOINT
    Starting pft on 3 8
    ...

I checked and all `OUTSIDEIMAGE` are in the final streamline. Why are they considered valid if they are classified as Outside the image?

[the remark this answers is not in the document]
gabknight: `OUTSIDEIMAGE` is equal to `ENDPOINT` for PFT. Since we don't have information at that point, we stop the tracking and keep the streamline. This is especially important to keep streamlines exiting the image volume through the spinal cord. 

There is bug here as the tracking should stop when it reach `OUTSIDEIMAGE`. Thank you for pointing this out. I fixed it and I added comments to clarify why the algorithm reaches a `break` or a `continue` staetment.

dipy/tracking/local/localtrack.pyx
```diff
@@ -208,3 +199,296 @@ cdef int _local_tracker(
...
+                         particle_count, particle_paths, particle_dirs,
+                         particle_weights, particle_states)
+
+                pft_trial += 1
+                # update the current point with the PFT results
+                copypoint(&streamline[i-1, 0], point)
+                copypoint(&directions[i-1, 0], dir)
+
+
+                if not tissue_class[0] == TRACKPOINT:
```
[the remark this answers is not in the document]
gabknight: I agree, in this case, with the `TissueTypes`, it is more clear with `!=`

dipy/tracking/local/tests/test_tracking.py
```diff
@@ -259,6 +258,125 @@ def allclose(x, y):
...
+    simple_gm = np.dstack([np.zeros(simple_gm.shape),
+                           simple_gm,
+                           simple_gm,
+                           simple_gm,
+                           np.zeros(simple_gm.shape)])
+    simple_csf = np.ones(simple_wm.shape) - simple_wm - simple_gm
+    tc = ActTissueClassifier.from_pve(simple_wm, simple_gm, simple_csf)
+    seeds = seeds_from_mask(simple_wm, density=2)
+
+    # Random pfm in every voxels
```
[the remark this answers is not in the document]
gabknight: fixed. thx.

dipy/tracking/local/tests/test_tracking.py
```diff
@@ -259,6 +258,125 @@ def allclose(x, y):
...
+    shape_img = list(simple_wm.shape)
+    shape_img.extend([sphere.vertices.shape[0]])
+    np.random.seed(0)  # fix the randominity
```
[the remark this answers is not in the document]
gabknight: yes :). I went with `np.random.seed(0)  # Random number generator initialization` to be safe here

thx.

gabknight:
@skoudoro, did you see that error before?
```
FAIL: dipy.reconst.tests.test_dki.test_multi_voxel_kurtosis_maximum
  File "/home/travis/build/nipy/dipy/venv/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/build/nipy/dipy/venv/lib/python2.7/site-packages/dipy/reconst/tests/test_dki.py", line 723, in test_multi_voxel_kurtosis_maximum
    assert_almost_equal(k_max, RK, decimal=5)
  File "/home/travis/build/nipy/dipy/venv/lib/python2.7/site-packages/numpy/testing/nose_tools/utils.py", line 565, in assert_almost_equal
    return assert_array_almost_equal(actual, desired, decimal, err_msg)
  File "/home/travis/build/nipy/dipy/venv/lib/python2.7/site-packages/numpy/testing/nose_tools/utils.py", line 963, in assert_array_almost_equal
    precision=decimal)
  File "/home/travis/build/nipy/dipy/venv/lib/python2.7/site-packages/numpy/testing/nose_tools/utils.py", line 779, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Arrays are not almost equal to 5 decimals
(mismatch 12.5%)
 x: array([[[  1.28571,   1.41176],
        [  8.53846,   3.12245]],
...
 y: array([[[  1.28571,   1.41176],
        [  8.53846,   3.12245]],
...
Ran 729 tests in 689.618s
FAILED (SKIP=25, failures=1)
```

skoudoro:
No @gabknight and It is weird that it appears only on one build. I will restart this build to be sure that it is reproducible before investigating it

gabknight:
Thank again @nilgoyette for noticing this bug. I much happier to find it now :p 

Here is the whole brain tracking (stanford_hardi) with and without PFT, same seeding, same tracking parameters.

pft

dipy/tracking/local/localtrack.pyx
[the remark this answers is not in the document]
jchoude: @gabknight I think we had a similar discussion in the prototype version of this code: you say you want to keep streamlines exiting the image volume through the spinal cord. So, is the chosen behavior for now to keep the additional point? If so, won't that bring back some issues when trying to project the streamlines to a grid / volume? I.e. the point outside won't match a possible voxel coordinate, and might cause undefined behavior?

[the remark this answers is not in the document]
gabknight: @jchoude I don't keep the `OUTSIDEIMAGE` point. My answer was misleading. `ENDPOINT` and `OUTSIDEIMAGE` show  **almost** equal behaviors, except at line 381 (localtrack.pyx), where the last point is removed for the `OUTSIDEIMAGE` case. I'm not sure we test that all points are within the image volume. I will add this test.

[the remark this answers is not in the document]
jchoude: Oh ok all good then, I just misunderstood.

Thanks a lot for all this work!

dipy/tracking/local/localtrack.pyx
```diff
@@ -208,3 +199,303 @@ cdef int _local_tracker(
...
+                particle_weights[p] *= 1 - tc.get_exclude_c(point)
+                if particle_weights[p] < eps:
+                    particle_weights[p] = 0
+                if (particle_states[0, p, 0] == INVALIDPOINT and
+                        particle_weights[p] > 0):
+                    particle_states[0, p, 0] = TRACKPOINT
+
+        sum_weights = 0
+        for p in range(particle_count):
+            sum_weights += particle_weights[p]
```
nilgoyette: Just to be sure, are you aware that sometimes your weights all equal NaN?

dipy/tracking/local/localtrack.pyx
```diff
@@ -208,3 +199,303 @@ cdef int _local_tracker(
...
+        sum_squared = 0
+        for p in range(particle_count):
+            particle_weights[p] = particle_weights[p] / sum_weights
+            sum_squared += particle_weights[p] * particle_weights[p]
+
+        # Resample the particles if the weights are too uneven.
+        # Particles with negligable weights are replaced by duplicates of
```
nilgoyette: negligible

dipy/tracking/local/localtrack.pyx
```diff
@@ -208,3 +199,303 @@ cdef int _local_tracker(
...
+        # those with high weigths through resampling
+        N_effective = 1. / sum_squared
+        if N_effective < particle_count / 10.:
```
nilgoyette: **Using your test** (without the fixed random), I couldn't get this to happen. In fact, everytime I tested, all elements in the arrays (paths, dirs, states, weights) are perfectly equal in the `p` dimension. Looking at the code, I don't see how they could be different. If they are actually always equal, then this algo is much too complicated!

I probably don't understand well enough. Can you please explain how it can happen? Is it only because I'm using test data?

dipy/tracking/local/localtrack.pyx
```diff
@@ -208,3 +199,303 @@ cdef int _local_tracker(
...
+            if dg.get_direction_c(point, dir):
+                particle_states[0, p, 0] = INVALIDPOINT
+                particle_weights[p] = 0
+            else:
+                for j in range(3):
+                    voxdir[j] = dir[j] / voxel_size[j]
+                fixed_step(point, voxdir, step_size)
+                copypoint(point, &particle_paths[0, p, s + 1, 0])
+                copypoint(dir, &particle_dirs[0, p, s + 1, 0])
+                particle_states[0, p, 0] = tc.check_point_c(point)
```
nilgoyette: Any reason why you shared the `particle tissue_class` with the `particle length`? IMO, it's harder to understand for no real gain. Shouldn't they be 2 arrays?

dipy/tracking/local/localtrack.pyx
```diff
@@ -208,3 +199,303 @@ cdef int _local_tracker(
...
+          np.float_t[:, :, :, :] particle_paths,
+          np.float_t[:, :, :, :] particle_dirs,
+          np.float_t[:] particle_weights,
+          np.int_t[:, :, :] particle_states):
+    cdef:
+        double sum_weights, sum_squared, N_effective, rdm_sample
+        double point[3]
+        double dir[3]
+        double voxdir[3]
+        double eps = np.finfo(np.float).eps
```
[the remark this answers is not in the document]
gabknight: float64 should be fine as we always use `double` in `trilinear_interpolate4d`. It might be safer to put eps slightly higher as we do multiple additions in `trilinear_interpolate4d`.

dipy/tracking/local/localtracking.py
```diff
@@ -68,66 +71,181 @@ def __init__(self, direction_getter, tissue_classifier, seeds, affine,
...
+        """
+
+        if not isinstance(tissue_classifier, ConstrainedTissueClassifier):
+            raise ValueError("expecting ConstrainedTissueClassifier")
+
+        self.pft_max_nbr_back_steps = int(np.ceil(pft_back_tracking_dist
+                                                  / step_size))
+        self.pft_max_nbr_front_steps = int(np.ceil(pft_front_tracking_dist
+                                                   / step_size))
+        self.pft_max_steps = (self.pft_max_nbr_back_steps
```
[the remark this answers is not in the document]
gabknight: `back_steps + front_steps` is variable (depending on the current tracking point). I removed it from self. thx.

dipy/tracking/local/localtrack.pyx
gabknight: thx.

dipy/tracking/local/localtrack.pyx
gabknight: I wasn't. This happens when all the weighs of the particle are equal to 0. In this case, all particle paths are equally bad and we return the path of the first particle, with the `TissueClass` `INVALIDPOINT`. i.e. PFT failed to find an alternative valid path. Having weights = 0 or nan produces the same output. I now removed the division that made the nans. thx.

dipy/tracking/local/localtrack.pyx
gabknight: On my side, the resampling happens and the weights, dirs, states, path are different.
try running `nosetests --nocapture -v` in `dipy/tracking/local/`
and adding a print at line localtrack.pyx:463 `print("Resampling")`
add adding a print at line localtracking.py:239 `print(self.particle_particle_weights)`

You can also try running the example: `python doc/example/particle_filtering_tractography.py` to check if `pft_streamlines` and `'local_prob_streamlines` are different.


----------------

In `_pft(.)`, we do probabilistic tractography for P particles and N steps. at each tracking step, the P particle weights are updates following the information provided by the `TissueClassifier` (localtrack.pyx:line 441). After the N steps, we sample 1 particle from the P particles using their weight (higher weight, more likely to be sampled). The path of this particle is the one used in the final streamline.

The thing here is that P particles might be very few particles to properly explore a given configuration. To overcome that, we focus the _exploration_ in regions more promising (i.e. particles with higher weights). To do that, we discard particles with low weight and duplicate particles with high weights. This is done when the weight distribution is unbalanced (e.g. when `1. / sum_squared < particle_count / 10`).

dipy/tracking/local/localtrack.pyx
gabknight: I thought that would be more clear to group them to reduce the number of arrays... Let me try with 2 arrays.

dipy/tracking/local/localtracking.py
```diff
@@ -68,66 +71,187 @@ def __init__(self, direction_getter, tissue_classifier, seeds, affine,
...
+                                                  / step_size))
+        self.pft_max_nbr_front_steps = int(np.ceil(pft_front_tracking_dist
+                                                   / step_size))
+        pft_max_steps = (self.pft_max_nbr_back_steps +
+                         self.pft_max_nbr_front_steps)
+
+        if (self.pft_max_nbr_front_steps < 0
+                or self.pft_max_nbr_back_steps < 0
+                or pft_max_steps < 1):
+            raise ValueError("The number of PFT steps must be greater than 0.")
```
nilgoyette: Wouldn't `<= 0.0` be better for `pft_max_nbr_front_steps` and `pft_max_nbr_back_steps`? It seems to me that `pft_back_tracking_dist` and `pft_front_tracking_dist` should be > 0.0. No?

gabknight: At least one of the 2 must be `>0`, otherwise, there is no PFT.  Putting `pft_max_nbr_front_steps = 0` is conceivable, with a `pft_back_tracking_dist > 0`, or the other way around. 

With `CMC` (probabilistic stopping criteria), `pft_back_tracking_dist` can be 0 without issue, that wouldn't work for `ACT` (deterministic stopping criteria), i.e. with `ACT` and `pft_back_tracking_dist = 0`, the output would be the same as without PFT.

skoudoro:
Any other review or request? @Garyfallidis ? 

If not, I will merge it this Friday.

jchoude:
Retested integration in our normal work, works well. It is a +1 from me.

Good job @gabknight and everyone!

Garyfallidis:
Give me time to review this until this Sunday.

doc/examples/particle_filtering_fiber_tracking.py
```diff
@@ -0,0 +1,151 @@
...
+tractography (ACT) [Smith2012]_ both uses PVEs information from
+anatomical images to determine when the tractography stops.
+Both tissue classifiers use a trilinear interpolation
+at the tracking position. CMC tissue classifier uses a probability derived from
+the PVE maps to determine if the streamline reaches a 'valid' or 'invalid'
+region. ACT uses a fixed threshold on the PVE maps. Both tissue classifiers can
+be used in conjunction with PFT. In this example, we used CMC.
+"""
+
+import matplotlib.pyplot as plt
```
Garyfallidis: Unused import

doc/examples/particle_filtering_fiber_tracking.py
```diff
@@ -0,0 +1,151 @@
...
+# Local Probabilistic Tractography
+probabilistic_streamlines = LocalTracking(dg,
+                                          cmc_classifier,
+                                          seeds,
+                                          affine,
+                                          max_cross=1,
+                                          step_size=step_size,
+                                          maxlen=1000,
+                                          return_all=False)
+probabilistic_streamlines = [s for s in probabilistic_streamlines]
```
Garyfallidis: This is the old standard for generating streamlines which has memory issues what you should use is 
```
from dipy.tracking.streamline import Streamlines
streamlines = Streamlines(probabilistic_streamlines)
```
See example `tracking_quick_start.py` for more.

doc/examples/particle_filtering_fiber_tracking.py
```diff
@@ -0,0 +1,151 @@
...
+save_trk("probabilistic_streamlines.trk", probabilistic_streamlines, affine,
```
Garyfallidis: This should also be updated by a new function. But @skoudoro will take care of that in his Streamlines API update PR.
#1379

doc/examples/particle_filtering_fiber_tracking.py
```diff
@@ -0,0 +1,151 @@
...
+                                            seeds,
+                                            affine,
+                                            max_cross=1,
+                                            step_size=step_size,
+                                            maxlen=1000,
+                                            pft_back_tracking_dist=2,
+                                            pft_front_tracking_dist=1,
+                                            particle_count=15,
+                                            return_all=False)
+streamlines = [s for s in pft_streamlines]
```
Garyfallidis: See below.

Garyfallidis:
Apart from a few comments. I think this is ready.

gabknight:
Thank you for the feedback @Garyfallidis, I made the changes. Thank you all for your help and feedback on this PR!

Garyfallidis:
Merging :star2: ! Many years on the making for this PR. Good luck! And @gabknight lets make a workflow :)

jchoude:
:beer: thanks @gabknight for the hard work and everyone (@nilgoyette ) for the reviews!

PR #1386, opened 2017-12-04
Possible fix for the inline compilation problem

nilgoyette:
I'm not sure this will actually fix the problem because I have several others problems on Windows here. This PR is more of a way to test on your server.

### DIFF
--- a/dipy/utils/fast_numpy.pxd
+++ b/dipy/utils/fast_numpy.pxd
@@ -5,12 +5,12 @@
 cimport numpy as np
 
 # Replaces a numpy.searchsorted(arr, number, 'right')
-cdef inline int where_to_insert(
+cdef int where_to_insert(
         np.float_t* arr,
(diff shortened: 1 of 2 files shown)


skoudoro:
I try it, and it resolved the compilation error, but you have to add this change on the `pyx` file too.

nilgoyette:
I did as you requested, but why did you tell me `to add this change on the pyx file too`? Was there a warning or an error for one of the OS?

nilgoyette:
The red X only mean that the test coverage decreased, right?

Is there anything else I can do? If not, we (Imeka) look forward to work with PFT so it would be nice if this commit can get merged.

skoudoro:
Hi @nilgoyette,

> Was there a warning or an error for one of the OS?

warning on my Windows machine, I would like to understand why because it is really strange. Anyway, merging.

PR #1388, opened 2017-12-07
Mapmri workflow

aarya22:


arokem:
This should probably supersede #1350 ?

aarya22:
I opened this by accident. It should be the same as #1350 now. I just thought we should keep all the related conversations on the MAPMRI workflow in one place in #1350 so I closed this request.

PR #1391, opened 2017-12-14
Re-entering conflict-free typos from deleted PR 1331

justbennet:
I believe these are all the corrections that are left that were not addressed by the documentation in the current master branch.  I am not able to run `make` on this, as that seems to want to download data and run dypy.  If there is a way to make _just_ the documentation and you can send me instructions and/or a link to the instructions for making just the docs, then I will be happy to give it a try.

Sorry for all the hassle this has caused.

### DIFF
--- a/doc/examples/combined_workflow_creation.py
+++ b/doc/examples/combined_workflow_creation.py
@@ -138,11 +138,3 @@ with the parameters and see the results::
 .. include:: ../links_names.inc
 
 """
-
-
-
-
-
-
-
-
--- a/doc/examples/gradients_spheres.py
+++ b/doc/examples/gradients_spheres.py
@@ -78,7 +78,7 @@ fvtk.record(ren, out_path='full_sphere.png', size=(300, 300))
 
    Full sphere.
 
-It is time to create the Gradients. For this reason we will need to use the
+It is time to create the Gradients. For this purpose we will use the
 function ``gradient_table`` and fill it with the ``hsph_updated`` vectors that 
 we created above.
 """
@@ -98,7 +98,7 @@ bvecs = np.vstack((vertices, vertices))
 bvals = np.hstack((1000 * values, 2500 * values))
 
 """
-We can also add some b0s. Let's add one in the beginning and one at the end.
+We can also add some b0s. Let's add one at the beginning and one at the end.
 """
 
 bvecs = np.insert(bvecs, (0, bvecs.shape[0]), np.array([0, 0, 0]), axis=0)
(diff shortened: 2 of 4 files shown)


skoudoro:
Thanks for doing this @justbennet  and Thanks for the review @jchoude !

Unfortunately, there is no way to generate the documentation gallery without running examples but it should be easy to add a command for doing that.

Otherwise, everything else can be checked via `make api` and then `make html-after-examples`.

jchoude:
@skoudoro Want to have a look, or we can merge? It is straight forward. I looked, all good. Was just waiting for the checks to finish, and they just did.

justbennet:
@jchoude Did the annotation at the bottom work out OK?

I don't know how many people there are like me who don't run the software but do read the docs.  I help other people, and my laptop isn't really suitable to run Dipy.  I do use it to read documentation, though.  I think the software needed to make just the documenation would be supportable.  Up to you folks.

skoudoro:
There is the same error on these 3 files:
* [reconst_csa.py](https://github.com/nipy/dipy/blob/master/doc/examples/reconst_csa.py#L129)
* [workflow_creation.py](https://github.com/nipy/dipy/blob/master/doc/examples/workflow_creation.py#L128)
* [combined_workflow_creation.py](https://github.com/nipy/dipy/blob/master/doc/examples/combined_workflow_creation.py#L141)

It would be great if you can you remove all spaces from these files @justbennet  :-)

After that, we can merge it. 

Thanks !

justbennet:
I don't understand.  I did not touch those file at all.

justbennet:
I touched one file only, doc/examples/gradients_spheres.py

jchoude:
I think that @skoudoro meant that, at the same time, if you could additionnally fix it in the other files, he would be appreciative :)

And I understand your comment about just generating the documentation without running everything. The situation is that all images and screenshots are generated directly using the software, so, for the example to completly render, they need to do some processing.

skoudoro:
Exactly, Thanks @jchoude , and sorry if I was not clear.

justbennet:
Ah.  OK.  What spaces?  The trailing space at the end of those files?

justbennet:
I believe I did as you requested.  Sorry for the delay.  I am at a different machine and had to reclone the repository.

jchoude:
No problem! Thanks a lot for your patience with us!

I'm just waiting on the CI to finish.

jchoude:
I generated the rst files locally, and the all render correctly. Merging!

Thanks @justbennet for your work!

justbennet:
You're welcome.

Issue #1387, opened 2017-12-07
Assert equality, instead of asserting that `a == b` is true


arokem:
For example, here: https://github.com/nipy/dipy/blob/master/dipy/workflows/tests/test_reconst_dti.py#L45

This is because asserting equality gives more informative error messages.

thechargedneutron:
I am new to dipy and would like to take up this issue. Should I also replace `assert_true(x <= y)` with `assert_less(x, y)`?

skoudoro:
yes @thechargedneutron, thanks for doing this !

thechargedneutron:
@arokem  @skoudoro  Please review. :smile:

PR #1393, opened 2018-01-01
Adds a DKI workflow.

arokem:
Also, some PEP8 fixes in the `workflows/reconst.py` file


arokem:
Rebased on master after #1400. 

Anyone want to review this?

dipy/workflows/reconst.py
```diff
@@ -504,6 +506,185 @@ def run(self, input_files, bvalues, bvectors, mask_files,
...
+            multiple masks at once. (default: No mask used)
+        b0_threshold : float, optional
+            Threshold used to find b=0 directions (default 0.0)
+        save_metrics : variable string, optional
+            List of metrics to save.
+            Possible values: fa, ga, rgb, md, ad, rd, mode, tensor, evec, eval
+            (default [] (all))
+        out_dir : string, optional
+            Output directory (default input file directory)
+        out_tensor : string, optional
```
skoudoro: Unexpected parameter `out_tensor` in docstring. The function `run` has  `out_dt_tensor` and `out_dk_tensor`

dipy/workflows/reconst.py
```diff
@@ -504,6 +506,185 @@ def run(self, input_files, bvalues, bvectors, mask_files,
...
+                nib.save(evecs_img, oevecs)
+
+            if 'eval' in save_metrics:
+                evals_img = nib.Nifti1Image(dkfit.evals.astype(np.float32),
+                                            affine)
+                nib.save(evals_img, oevals)
+
+            if 'mk' in save_metrics:
+                mk_img = nib.Nifti1Image(dkfit.mk().astype(np.float32),
+                                     affine)
```
skoudoro: PEP8: continuation of line

dipy/workflows/reconst.py
```diff
@@ -504,6 +506,185 @@ def run(self, input_files, bvalues, bvectors, mask_files,
...
+                nib.save(mk_img, omk)
+
+            if 'ak' in save_metrics:
+                ak_img = nib.Nifti1Image(dkfit.ak().astype(np.float32),
+                                     affine)
```
skoudoro: PEP8: continuation of line

dipy/workflows/reconst.py
```diff
@@ -504,6 +506,185 @@ def run(self, input_files, bvalues, bvectors, mask_files,
...
+                nib.save(ak_img, oak)
+
+            if 'rk' in save_metrics:
+                rk_img = nib.Nifti1Image(dkfit.rk().astype(np.float32),
+                                     affine)
```
skoudoro: PEP8: continuation of line

dipy/workflows/tests/test_reconst_dki.py
```diff
@@ -0,0 +1,94 @@
...
+        assert_true(evecs_data.shape[:-2] == volume.shape[:-1])
+
+        evals_path = dki_flow.last_generated_outputs['out_eval']
+        evals_data = nib.load(evals_path).get_data()
+        assert_true(evals_data.shape[-1] == 3)
+        assert_true(evals_data.shape[:-1] == volume.shape[:-1])
+
+
+
+if __name__ == '__main__':
```
skoudoro: PEP8:  Too many blank line

arokem:
I think that addressed all your comments, @skoudoro ?

dipy/workflows/tests/test_reconst_dki.py
```diff
@@ -0,0 +1,93 @@
...
+
+        dki_flow.run(*args, out_dir=out_dir)
+
+        fa_path = dki_flow.last_generated_outputs['out_fa']
+        fa_data = nib.load(fa_path).get_data()
+        assert_true(fa_data.shape == volume.shape[:-1])
+
+        tensor_path = dki_flow.last_generated_outputs['out_dt_tensor']
+        tensor_data = nib.load(tensor_path)
+        assert_true(tensor_data.shape[-1] == 6)
```
skoudoro: I think you prefer  `assert_equal` than `assert_true` like you said on #1387 ;-)

dipy/workflows/reconst.py
```diff
@@ -504,6 +504,193 @@ def run(self, input_files, bvalues, bvectors, mask_files,
...
-    def get_tensor_model(self, gtab):
-        return TensorModel(gtab, fit_method="RT", sigma=self.sigma,
-                           jac=self.jacobian)
+
+class ReconstDkiFlow(Workflow):
+    @classmethod
+    def get_short_name(cls):
+        return 'dki'
+
+    def run(self, input_files, bvalues, bvectors, mask_files, b0_threshold=0.0,
```
skoudoro: After playing with this workflow, I wonder if `mask_files` can be optional.

It was a constraint to generate a mask before running this model when I just want to test it on the whole image (or I was lazy :-))

arokem:
Renamed the script and rebased!

dipy/workflows/reconst.py
arokem: I think that setting default to `None` (i.e., use all voxels) would be fine. 

But it needs to be consistently applied across all the workflows (see for example the `dipy_dti_fit` flow, see: https://github.com/nipy/dipy/blob/master/dipy/workflows/reconst.py#L30).

skoudoro: Agree with consistency, so no need to change it.  Apparently, we want to give good habits to users

arokem:
No problem. If we decide otherwise, we can also make this change across all
the workflows in one PR. It wouldn't even break backwards compatibility.

Anything else here?

skoudoro:
No, everything good, I just tested it again on DKI dataset and the workflow works well.

Can you just rebase? Then I can merge it.

Thank you @arokem

dipy/workflows/reconst.py
```diff
@@ -120,9 +121,7 @@ def run(self, input_files, bvalues, bvectors, mask_files, b0_threshold=0.0,
             data = img.get_data()
             affine = img.affine
 
-            if mask is None:
-                mask = None
-            else:
+            if mask is not None:
```
arokem: OK - I am not sure why this is here. @Garyfallidis : under what circumstances would the mask be a `None`? It's a required input into the workflow?

arokem:
Rebased!

dipy/workflows/reconst.py
Garyfallidis: When you run the Flow from a Python script it is easy to put a None.  But not easy from the command line. It may be good to enforce the same behavior for consistency.

arokem: That makes sense! What do you think about defaulting to `None`, so that if users don't pass a mask, we use this option? We can do that on a separate PR, to make sure to do this uniformly across the different flows.

arokem: All that's to say: I think this PR is ready to go! @skoudoro : would you mind giving this another look?

dipy/workflows/reconst.py
```diff
@@ -504,3 +503,194 @@ def run(self, input_files, bvalues, bvectors, mask_files, sh_order=6,
...
+            (default 'evecs.nii.gz')
+        out_eval : string, optional
+            Name of the eigenvalues to be saved (default 'evals.nii.gz')
+        out_mk : string, optional
+            Name of the mean kurtosis to be saved (default: 'mk.nii.gz')
+        out_ak : string, optional
+            Name of the axial kurtosis to be saved (default: 'ak.nii.gz')
+        out_rk : string, optional
+            Name of the radial kurtosis to be saved (default: 'rk.nii.gz')
+        """
```
skoudoro: Can you add references/citations?

dipy/workflows/reconst.py
skoudoro: I think it will be good to just raise an exception if there is no mask. In that way, 
1. we keep the consistency with the command line
2. we give good habits to the user by forcing him to generate a mask

But I agree with @arokem, we can do that in a separate PR

skoudoro:
LGTM, I let you fix my last comment, and I can merge it today

arokem:
Yeah - good call. I also fixed up other references in this module for the
sphinx docs.

dipy/workflows/reconst.py
```diff
@@ -581,6 +581,18 @@ def run(self, input_files, bvalues, bvectors, mask_files, b0_threshold=0.0,
             Name of the axial kurtosis to be saved (default: 'ak.nii.gz')
         out_rk : string, optional
             Name of the radial kurtosis to be saved (default: 'rk.nii.gz')
+
+        Notes
```
skoudoro: `References` instead of `Notes`

arokem:
I believe that according to the numpy docstring standard, references should
be in the "Notes" section.

Do we want to diverge from the numpy docstring standard? I'm fine if we do,
but we should document that somewhere.

skoudoro:
I just checked on [numpy docstring standard](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt) and we do not diverge from it. `References` are not in `Notes` section.
I do not know if it is something new or not

arokem:
My reading of this line in the standard:
https://github.com/numpy/numpy/blame/master/doc/HOWTO_DOCUMENT.rst.txt#L375,
was that references should be cited in the notes section. But you are
right, there seem to be numpy docstring that have a separate section for
references (e.g., here:
https://docs.scipy.org/doc/numpy-1.13.0/reference/generated/numpy.linalg.solve.html#numpy.linalg.solve).
Correcting...

skoudoro:
Thank you for completing this information

skoudoro:
Thanks for this workflow @arokem !  merging!

PR #1395, opened 2018-01-03
Mapmri workflow

aarya22:
Supersedes #1350


dipy/workflows/reconst.py
```diff
@@ -21,6 +21,330 @@
...
 
+from dipy.reconst import mapmri
+
+class ReconstMAPMRIFlow(Workflow):
+    @classmethod
+    def get_short_name(cls):
+        return 'mapmri'
+
+    def run(self, data_file, data_bvecs, data_bvals, model_type='', out_rtop='rtop.nii.gz',
+            out_lapnorm='lapnorm.nii.gz',out_msd='msd.nii.gz', out_qiv='qiv.nii.gz', out_rtap='rtap.nii.gz',
```
arokem: PEP8: line is longer than 80 characters

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,330 @@
...
+            out_rtpp='rtpp.nii.gz', small_delta=0.0129, big_delta=0.0218, save_metrics=[], out_dir=''):
```
arokem: PEP8: line is longer than 80 characters

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,330 @@
...
+        """ Workflow for the app-dipy-mapmri on Brain-Life (www.brain-life.org). 
+        Generates rtop, lapnorm, msd, qiv, rtap, rtpp saved in a nifti format in input files provided by
```
arokem: PEP8: line is longer than 80 characters

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,330 @@
...
+        `data_file` and saves the png file to an output directory specified by 
+        `out_dir`.
+
+        Parameters
+        ----------
+        data_file : string
+            Path to the input volume.
+        data_bvecs : string
+            Path to the bvec files.
+        data_bvals :
```
arokem: => `data_bvals` : string

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,330 @@
...
+
+        Parameters
+        ----------
+        data_file : string
+            Path to the input volume.
+        data_bvecs : string
+            Path to the bvec files.
+        data_bvals : 
+            Path to the bval files.
+        small_delta :
```
arokem: => `small_delta : float`

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,330 @@
...
+            Small delta value used in generation of gradient table of provided
+            bval and bvec. (default: 0.0129)
+        big_delta :
```
arokem: => `big_delta : float`

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,330 @@
...
+            Big delta value used in generation of gradient table of provided 
+            bval and bvec. (default: 0.0218)
+        model_type : string
+            Model type to fit.
+            Possible values: laplacian, positivity, both
+            (default: both)
+        save_metrics :
```
arokem: => `save_metrics : list of strings`

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,330 @@
...
+            Name of the msd to be saved
+        out_qiv : string, optional
+            Name of the qiv to be saved
+        out_rtap : string, optional
+            Name of the rtap to be saved
+        out_rtpp : string, optional
+            Name of the rtpp to be saved
+        """
+        io_it = self.get_io_iterator()
+        for dwi, bval, bvec, out_rtop, out_lapnorm, out_msd, out_qiv, out_rtap, out_rtpp in io_it:
```
arokem: PEP8: Longer than 80 characters

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,330 @@
...
+
+            logging.info('Computing DTI metrics for {0}'.format(dwi))
+            img = nib.load(dwi)
+            data = img.get_data()
+            affine = img.affine
+            bvals, bvecs = read_bvals_bvecs(bval, bvec)
+
+            gtab = gradient_table(bvals=bvals, bvecs=bvecs,
+                                  small_delta=small_delta,
+                                  big_delta=big_delta, b0_threshold=50)
```
arokem: Why is the b0_threshold here set to 50?

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,330 @@
...
+                model_type = 'both'
+
+            if not save_metrics:
+                save_metrics = ['rtop', 'laplacian_signal', 'msd', 'qiv', 'rtap', 'rtpp']
+
+            # print('data.shape (%d, %d, %d, %d)' % data.shape)
+
+            radial_order = 6
+            
+            # How do I save the rtop in a useful form? A picture? A numpy array?
```
arokem: Please remove these comments

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,330 @@
...
+                mapfit_aniso = map_model_aniso.fit(data)
+
+            elif model_type is 'both':
+                map_model_aniso = mapmri.MapmriModel(gtab, radial_order=radial_order,
+                                                     laplacian_regularization=True,
+                                                     laplacian_weighting=.05,
+                                                     positivity_constraint=True)
+                mapfit_aniso = map_model_aniso.fit(data)
+
+            # Not sure where to get affine or metadata?
```
arokem: Please remove this comment

dipy/workflows/reconst.py
arokem: I believe that `b0_threshold` should be an input, like here: https://github.com/nipy/dipy/blob/master/dipy/workflows/reconst.py#L30

aarya22:
@arokem I made the changes you had asked me to including the laplacian and positivity boolean arguments as well as the ng, ng parallel, and ng perpendicular. Let's see what travis says.

aarya22:
@arokem Travis is giving me the following error: 

ERROR: dipy.reconst.tests.test_dti.test_mask
    self.test(*self.arg)
  File "/home/travis/build/nipy/dipy/venv/lib/python3.5/site-packages/dipy/reconst/tests/test_dti.py", line 520, in test_mask
    assert_almost_equal(dtifit_w_mask.S0_hat[0, 0, 0], dtifit.S0_hat[0, 0, 0])
  File "/opt/python/3.5.4/lib/python3.5/unittest/case.py", line 870, in assertAlmostEqual
    if round(abs(second-first), places) == 0:

I didn't work on this file at all. do you know what this might be?

arokem:
Don't worry about it. This is something that is currently plaguing all of our PRs. 

Seems to be a recent change in numpy (or nose?) in the way that arrays are compared. 

So - has nothing to do with your code!

aarya22:
Hm.. maybe we can do np.round, instead of round?

aarya22:
This should be ready, could somebody take a look?

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,165 @@
 from dipy.reconst.shm import CsaOdfModel
 from dipy.workflows.workflow import Workflow
 
+from dipy.reconst import mapmri
+
```
arokem: Add another line of whitespace here

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,165 @@
...
+    def get_short_name(cls):
+        return 'mapmri'
+
+    def run(self, data_file, data_bvecs, data_bvals, b0_threshold=0.0,
+            laplacian=True, positivity=True, out_rtop='rtop.nii.gz', out_lapnorm='lapnorm.nii.gz',
+            out_msd='msd.nii.gz', out_qiv='qiv.nii.gz', out_rtap='rtap.nii.gz',
+            out_rtpp='rtpp.nii.gz', out_ng='ng.nii.gz', out_perng='perng.nii.gz',
+            out_parng='parng.nii.gz', small_delta=0.0129, big_delta=0.0218, save_metrics=[],
+            out_dir='', laplacian_weighting=0.05):
+        """ Workflow for the app-dipy-mapmri on Brain-Life (www.brain-life.org).
```
arokem: Please remove the mention of brainlife. 

The correct description here is: "Workflow for fitting the MAPMRI model (with optional Laplacian regularization)."

It happens to also work on brainlife, but that's not the only place it could work.

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,165 @@
...
+        return 'mapmri'
+
+    def run(self, data_file, data_bvecs, data_bvals, b0_threshold=0.0,
+            laplacian=True, positivity=True, out_rtop='rtop.nii.gz', out_lapnorm='lapnorm.nii.gz',
+            out_msd='msd.nii.gz', out_qiv='qiv.nii.gz', out_rtap='rtap.nii.gz',
+            out_rtpp='rtpp.nii.gz', out_ng='ng.nii.gz', out_perng='perng.nii.gz',
+            out_parng='parng.nii.gz', small_delta=0.0129, big_delta=0.0218, save_metrics=[],
+            out_dir='', laplacian_weighting=0.05):
+        """ Workflow for the app-dipy-mapmri on Brain-Life (www.brain-life.org). 
+        Generates rtop, lapnorm, msd, qiv, rtap, rtpp saved
```
arokem: I think the description of the outputs is out-of-date. It also generates non-Gaussianity, right?

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,165 @@
...
+                                                     laplacian_regularization=False,
+                                                     positivity_constraint=True)
+                mapfit_aniso = map_model_aniso.fit(data)
+
+            elif laplacian:
+                map_model_aniso = mapmri.MapmriModel(gtab, radial_order=radial_order,
+                                                     laplacian_regularization=True,
+                                                     laplacian_weighting=laplacian_weighting)
+                mapfit_aniso = map_model_aniso.fit(data)
+
```
skoudoro: Thanks for this work @aarya22. I just have a couple of questions:

- When `laplacian=True` and `posivity=True`, the first `if` is overridden by the others... there is something wrong here
- What happens when `laplacian=False` ?
- What happens when `laplacian=False` and `posivity=False` ? We never know how it will be used...

aarya22: If laplacian=True and positivity=True, the first if should override the next two if statements right? This is supposed to happen, in order to create the correct mapmri model type given the inputs.

Perhaps it is not quite clear, but either the laplacian or positivity or both must be true in order for the workflow to function. If they are both false then the workflow will not work. I can add that in the description.

skoudoro: Ok, Thanks ! Maybe you should warn the users by generating a warning or print a message when both value are `False`

aarya22: Good idea! I just added that as an error message.

aarya22:
@arokem I've fixed a majority of the PEP8 issues. Let me know if there's anything more I need to fix.

@skoudoro Let me know if there's anything else I can change here.

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,184 @@
...
+                                  small_delta=small_delta,
+                                  big_delta=big_delta,
+                                  b0_threshold=b0_threshold)
+
+            if not save_metrics:
+                save_metrics = ['rtop', 'laplacian_signal', 'msd',
+                                'qiv', 'rtap', 'rtpp',
+                                'ng', 'perng', 'parng']
+
+            radial_order = 6
```
skoudoro: Do you know if this static variable works for all dataset? I think `radial_order` might be an input parameter? what do you think?

aarya22: @arokem What do you think, should this be an input parameter?

Not sure what radial order is.

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,184 @@
...
+                ng = nib.nifti1.Nifti1Image(n.astype(np.float32), affine)
+                nib.save(ng, out_perng)
+
+            if 'parng' in save_metrics:
+                n = mapfit_aniso.ng_parallel()
+                ng = nib.nifti1.Nifti1Image(n.astype(np.float32), affine)
+                nib.save(ng, out_parng)
+
+            logging.info('MAPMRI saved in {0}'.
+                         format(os.path.dirname(out_rtpp)))
```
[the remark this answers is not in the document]
aarya22: Ah thanks for pointing that out will fix it.

arokem:
Yes - that should be an input paramter. 6 seems like a reasonable default.

aarya22:
@arokem By default the bval_threshold for mapmri model is set to np.inf. This gives the warning that Serge gets: 

C:\Users\skoudoro\Devel\dipy\dipy\reconst\mapmri.py:786: UserWarning: model bval_threshold must be lower than 2000 for the non_Gaussianity to be physically meaningful [2]. warn(msg)

Is there a specific value for bval_threshold that I should set that is under 2000, or just any value under 2000 will do?

dipy/workflows/reconst.py
rutgerfick: radial_order defines how many basis functions mapmri uses to approximate the data. radial_order=0 means it only uses the first, which is just an anisotropic gaussian. radial_order=6 makes it use 50 basis functions.

Changing radial_order therefore directly influences the estimates q-space indices. You can default it to e.g. 6, but you definitely shouldn't hardcore it (maybe it makes sense to set it to 4 for very noisy or small data sets for example)

dipy/workflows/reconst.py
rutgerfick: Why not add a line with "elif not laplacian and not positivity" and then just set laplacian_regularization to False in the model call? There's no reason to put an exception here since the model will just do least squares in this case.

rutgerfick:
Hi everyone,

Do you guys want to somehow put propagator anisotropy into mapmri? I have a script that does it using the old cvxopt code, so someone will have to just click it into your new workflow (it requires simultaneous fitting of anisotropic and isotropic parts of the mapmri basis). Is this interesting for you?

Best,
Rutger

arokem:
Any chance you might refactor it to us cvxpy? We'd like to avoid GPL code.

rutgerfick:
Sure that's no issue. Can you handle it once I give you a cvxpy compatible
version?
You'll have to think a bit how to put it in your desired workflow since
PA-estimation requires fitting of both isotropic and anisotropic parts of
the the mapmri basis.

arokem:
Maybe we can integrate that into the reconst module first?

Could you say a little bit more about what "fitting of both isotropic and
anisotropic parts of the the mapmri basis" entails? What are the
complications you foresee?

rutgerfick:
In a nutshell, propagator anisotropy is just a measure difference between
the anisotropic approximation of a multi-shell signal (so regular mapmri
fit), and it's isotropic approximation.
In the reconstmodel this means that when you fit the basis coefficients for
the anisotropic basis up to a radial_order N, you ALSO need to separately
fit the (100, 200, ...., (N/2)00) coefficients of the isotropic basis.
PA is then some complicated function of these anisotropic and isotropic
coefficients, which is already coded and ready so you don't need to care
about that.

So it's not so much complications that I foresee, but you just need to
weigh for yourself if you want to add this feature at the cost of a
somewhat slower fitting process.
What do you say?

dipy/workflows/reconst.py
aarya22: @rutgerfick Do you mean to make a model with the following parameters in the case of not laplacian and not positivity?
 
`mapmri.MapmriModel(gtab, radial_order=radial_order, laplacian_regularization=False)`

aarya22:
@rutgerfick Is the fitting of the coefficients on the isotropic basis something that we will have to code into the workflow or does your code do that automatically?

dipy/workflows/reconst.py
rutgerfick: if positivity_constraint's default is False then that should do it :-)

rutgerfick:
It happens inside the ReconstModel, so the workflow would be inaffected.
The only difference on the outside would be that you have mapfit.pa() as a
new estimatable feature.

dipy/workflows/reconst.py
aarya22: Sounds good, will make the change. @rutgerfick Thanks for the suggestion.

aarya22:
Sounds good to me. @arokem What do you think? Maybe this should be an issue on mapmri.py.

dipy/workflows/reconst.py
```diff
@@ -124,32 +133,38 @@ def run(self, data_file, data_bvecs, data_bvals, b0_threshold=0.0,
                                 'qiv', 'rtap', 'rtpp',
                                 'ng', 'perng', 'parng']
 
-            radial_order = 6
+            radial_order = radial_order
```
arokem: I don't think this line is neccessary 😄

aarya22: You're absolutely right.

arokem:
Great! I think this is ready to go now. If you want to take a look before I merge, please do so in the next couple of days.

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,197 @@
...
+    def run(self, data_file, data_bvecs, data_bvals, b0_threshold=0.0,
+            laplacian=True, positivity=True, bval_threshold=2000,
+            out_rtop='rtop.nii.gz', out_lapnorm='lapnorm.nii.gz',
+            out_msd='msd.nii.gz', out_qiv='qiv.nii.gz',
+            out_rtap='rtap.nii.gz',
+            out_rtpp='rtpp.nii.gz', out_ng='ng.nii.gz',
+            out_perng='perng.nii.gz',
+            out_parng='parng.nii.gz', small_delta=0.0129,
+            big_delta=0.0218, save_metrics=[],
+            out_dir='', laplacian_weighting=0.05, radial_order=6):
```
skoudoro: Last wednesday, I learned that the parameter order has a crucial importance.  it should be : 
1. input ->  `data_file, data_bvecs, data_bvals`
2. optional parameter -> `b0_threshold=0.0, laplacian=True, positivity=True, bval_threshold=2000, small_delta=0.0129, big_delta=0.0218, save_metrics=[], laplacian_weighting=0.05, radial_order=6`
3. `out_dir`
4. All `out_*`

Can you do this change and move `small_delta=0.0129, big_delta=0.0218, save_metrics=[], out_dir='', laplacian_weighting=0.05, radial_order=6` in the right place ?

Can you do the same below too (in the documentation)

Thanks !

rutgerfick:
Hello,

I don't think small_delta and big_delta should have preset values in any
version of MAPMRI workflow. The mapmri model internally uses q-values,
which are derived from b-values using delta/Delta. This means that if
delta/Delta are not corresponding to the data you're fitting, then q-values
are misestimated, which means estimated q-space indices will be biased
(either higher or lower) depending on how wrong delta/Delta were.

Best,
Rutger

dipy/workflows/reconst.py
aarya22: Got it, I'll make the change shortly. Thanks for the note.

aarya22:
@rutgerfick Is there a way to calculate or derive delta/Delta based on the data?

rutgerfick:
I'm afraid not :-(

rutgerfick:
It's the same as asking whether you can estimate the gradient strength from
the b-value. It's part of the acquisition parameters that makes up a data
set. They are all needed to make a proper assessment of the data, and
should be known by the user.

arokem:
Let's make small_delta/big_delta required inputs.

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,198 @@
...
+            (default: [] (all))
+        laplacian_weighting : float
+            Weighting value used in fitting the MAPMRI model in the laplacian
+            and both model types. (default: 0.05)
+        radial_order : unsigned int
+            Even value used to set the order of the basis
+            (default: 6)
+        out_dir : string, optional
+            Output directory (default: input file directory)
+        laplacian : boolean
```
skoudoro: Thanks for this last commit @aarya22!

`laplacian` and `positivity` should be before `out_dir`. can you move them? After that, it should be ready to go!

aarya22: @skoudoro Thanks for pointing that out. Should be good to go now.

skoudoro:
Any other comments on this workflow?  If no, I will merge it tomorrow

Garyfallidis:
#1358 needs to be merged first. Before this one. I suggest waiting for this to happen.

dipy/workflows/reconst.py
```diff
@@ -21,6 +21,195 @@
...
 
+from dipy.reconst import mapmri
+
+
+class ReconstMAPMRIFlow(Workflow):
+    @classmethod
+    def get_short_name(cls):
+        return 'mapmri'
+
+    def run(self, data_file, data_bvecs, data_bvals, small_delta, big_delta,
```
skoudoro: Hi @aarya22, can you switch `data_bvecs` and `data_bvals`.  When I run the workflows, I get an error because it does not match with the line [114](https://github.com/nipy/dipy/pull/1395/files#diff-678dce6086d38985220e2ded56251680R114) where `bval` is first and `bvect` the second parameter.
Morever, it keeps the consistency with the others workflows.

skoudoro:
Thanks @aarya22 !  Can you rebase ? it is ready to be merged, just missing that.

skoudoro:
closing in favor of #1428. Thank you again @aarya22!

PR #1406, opened 2018-01-26
[MRG] assert_true which checks for equality replaced with assert_equal

### LINKED ISSUE #1387: Assert equality, instead of asserting that `a == b` is true

thechargedneutron:
Fixes #1387

### DIFF
--- a/dipy/reconst/tests/test_shm.py
+++ b/dipy/reconst/tests/test_shm.py
@@ -3,7 +3,7 @@ import warnings
 import numpy as np
 import numpy.linalg as npl
 
-from nose.tools import assert_equal, assert_raises, assert_less
+from nose.tools import assert_equal, assert_raises, assert_true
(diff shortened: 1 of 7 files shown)


thechargedneutron:
@skoudoro Thanks for the review.
However, `test_utils.py` contains `assert_true(x is y)` and not `assert_true(x==y)`. Are these two equivalent in that context?

skoudoro:
No, sorry, it seems we are testing the reference equality and not the value. 
Thanks for pointing that @thechargedneutron.

LGTM, anyone for reviewing this PR before I merge ?

skoudoro:
Thanks @thechargedneutron!

PR #1408, opened 2018-01-31
Updated color map in reconst_csa.py and reconst_forecast.py 

mrfox321:
... from 'jet' to 'plasma'


skoudoro:
Thanks for contributing and this change @mrfox321!

PR #1418, opened 2018-02-06
Adding parallel_voxel_fit decorator

skoudoro:
This PR is an updated version of PR #1135 and #642. I have simplified and refactored the code of @MrBago and @sahmed95.

You just need to add this decorator (`@parallel_voxel_fit` ) to your model for splitting your data into small chunks and run them through several processors.

To define the number of processors, you just need to do: `model.fit(data, nb_processes=6)`

Currently, all the code is in `reconst/multi_voxe.py` but I suppose that this is not the right place and any ideas are welcome.

Can you have a look @nilgoyette @MrBago @arokem  @sahmed95 @Garyfallidis?

Thanks! 

***Note:*** In order for multiprocessing to be beneficial, the fitting task needs to be significant. Significant means bigger than the overhead created by multiprocessing during inter-process communication. If it is not the case, use `multi_voxel_fit` decorator.

***Note:*** `multiprocessing` is using `fork()` on Linux when it starts a new child process. On Windows -- which does not support fork() -- multiprocessing is using the win32 API call CreateProcess. It creates an entirely new process from any given executable (which is quite slow)

### DIFF
--- a/dipy/core/parallel.py
+++ b/dipy/core/parallel.py
@@ -1,0 +1,89 @@
+
+import ctypes
+import numpy as np
+from multiprocessing import Array
+
+
+_ctypes_to_numpy = {ctypes.c_char: np.dtype(np.uint8),
+                    ctypes.c_wchar: np.dtype(np.int16),
+                    ctypes.c_byte: np.dtype(np.int8),
+                    ctypes.c_ubyte: np.dtype(np.uint8),
+                    ctypes.c_short: np.dtype(np.int16),
+                    ctypes.c_ushort: np.dtype(np.uint16),
+                    ctypes.c_int: np.dtype(np.int32),
+                    ctypes.c_uint: np.dtype(np.uint32),
+                    ctypes.c_long: np.dtype(np.int64),
+                    ctypes.c_ulong: np.dtype(np.uint64),
+                    ctypes.c_float: np.dtype(np.float32),
+                    ctypes.c_double: np.dtype(np.float64),
+                    ctypes.c_void_p: np.dtype(object)}
+
+
+_numpy_to_ctypes = dict(zip(_ctypes_to_numpy.values(),
+                            _ctypes_to_numpy.keys()))
+
+
+def mparray_as_ndarray(mp_array, shape=None):
+    """
+    Given a multiprocessing.Array, returns an ndarray pointing to
+    the same data.
+
+    Parameters
+    ----------
+    mp_array : array
+        multiprocessing.Array that you want to convert.
+    shape : tuple, optional
+        tuple of array dimensions (default: None)
+
+    Returns
+    --------
+    numpy_array : ndarray
+        ndarray pointing to the same data
+    """
+
+    # support SynchronizedArray:
+    if not hasattr(mp_array, '_type_'):
+        mp_array = mp_array.get_obj()
+
+    dtype = _ctypes_to_numpy[mp_array._type_]
+    result = np.frombuffer(mp_array, dtype)
+
+    if shape is not None:
+        result = result.reshape(shape)
+
+    return np.asarray(result)
+
+
+def ndarray_to_mparray(arr, lock=False):
+    """
+    Generate an 1D multiprocessing.Array containing the data from
+    the passed ndarray.  The data will be *copied* into shared
+    memory.
+
+    Parameters
+    ----------
+    arr : ndarray
+        numpy ndarray that you want to convert.
+    lock : boolean, optional
+        controls the access to the shared array. When your shared
+        array has a read only access, you do not need lock. Otherwise,
+        any writing access need to activate the lock to be
+        process-safe(default: False)
+
+    Returns
+    --------
+    shm : multiprocessing.Array
+        copied shared array
+    """
+
+    array1d = arr.ravel(order='A')
+
+    try:
+        c_type = _numpy_to_ctypes[array1d.dtype]
+    except KeyError:
+        c_type = _numpy_to_ctypes[np.dtype(array1d.dtype)]
+
+    result = Array(c_type, array1d.size, lock=lock)
+    mparray_as_ndarray(result)[:] = array1d
+
+    return result
--- a/dipy/reconst/multi_voxel.py
+++ b/dipy/reconst/multi_voxel.py
@@ -114,7 +114,7 @@ def parallel_fit_worker(arguments):
         return a list of tuple(voxel index, model fitted instance)
     """
     model, input_queue, args, kwargs = arguments
-    return [(idx, model.fit(shared_arr[idx], args, kwargs)) for idx in input_queue]
+    return [(idx, model.fit(shared_arr[idx], *args, **kwargs)) for idx in input_queue]
 
 
 def _init_parallel_fit_worker(arr_to_populate, shape):
@@ -167,8 +167,11 @@ def parallel_voxel_fit(single_voxel_fit):
     True
     """
 
-    def new_fit(model, data, mask=None, *args, **kwargs):
+    def new_fit(model, data, *args, **kwargs):
         """Fit method in parallel for every voxel in data """
+        # Pop the mask, if there is one
+        mask = kwargs.pop('mask', None)
+        # print(mask, mask.shape)
         if data.ndim == 1:
             return single_voxel_fit(model, data, *args, **kwargs)
         if mask is None:
@@ -177,7 +180,7 @@ def parallel_voxel_fit(single_voxel_fit):
             raise ValueError("mask and data shape do not match")
 
         # Get number of processes
-        nb_processes = int(kwargs.get('nb_processes', '0'))
+        nb_processes = int(kwargs.pop('nb_processes', '0'))
         nb_processes = nb_processes if nb_processes >= 1 else cpu_count()
 
         if nb_processes == 1:
--- a/dipy/reconst/multi_voxel.py
+++ b/dipy/reconst/multi_voxel.py
@@ -115,9 +115,9 @@ def parallel_fit_worker(arguments):
     result: list of tuple
         return a list of tuple(voxel index, model fitted instance)
     """
-    model, input_queue, args, kwargs = arguments
-    return [(idx, model.fit(shared_arr[idx], *args, **kwargs))
-            for idx in input_queue]
+    model, chunks = arguments
+    return [(idx, model.fit(shared_arr[idx]))
+            for idx in chunks]
 
 
 def _init_parallel_fit_worker(arr_to_populate, shape):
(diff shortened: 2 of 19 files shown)


dipy/reconst/tests/test_multi_voxel.py
```diff
@@ -104,41 +105,71 @@ def test_CallableArray():
...
 
 
-def test_multi_voxel_fit():
+class SillyParallelModel(object):
+    @parallel_voxel_fit
+    def fit(self, data, mask=None):
+        return SillyFit(self, data)
+
+    def predict(self, S0):
+        return np.ones(10) * S0
```
nilgoyette: Why not use `numpy.full()`? Or if `S0` is an array, `10 * S0`?

dipy/reconst/tests/test_multi_voxel.py
```diff
@@ -104,41 +105,71 @@ def test_CallableArray():
...
-        def predict(self, S0):
-            return np.ones(self.data.shape) * S0
+    def predict(self, S0):
+        return np.ones(self.data.shape) * S0
 
+
+def test_parallel_voxel_fit():
+    voxel_fit(SillyParallelModel, SillyFit)
+    model = SillyParallelModel()
+    # Test with a mask and few processors
```
nilgoyette: Do you mean 'processes'?

dipy/reconst/multi_voxel.py
```diff
@@ -5,6 +5,215 @@
...
+        # Get number of processes
+        nb_processes = int(kwargs['nb_processes']) if 'nb_processes' in kwargs else cpu_count()
+        nb_processes = cpu_count() if nb_processes < 1 else nb_processes
+
+        if nb_processes == 1:
+            return single_voxel_fit(model, data, *args, **kwargs)
+
+        # Get non null index from mask
+        indexes = np.argwhere(mask)
+        # convert indexes to tuple
```
nilgoyette: You sometime start with a

    # Majuscule
    # miniscule

Could you please keep it standard? Here and elsewhere.

dipy/reconst/multi_voxel.py
```diff
@@ -5,6 +5,215 @@
...
+        if data.ndim == 1:
+            return single_voxel_fit(model, data, *args, **kwargs)
+        if mask is None:
+            mask = np.ones(data.shape[:-1], bool)
+        elif mask.shape != data.shape[:-1]:
+            raise ValueError("mask and data shape do not match")
+
+        # Get number of processes
+        nb_processes = int(kwargs['nb_processes']) if 'nb_processes' in kwargs else cpu_count()
+        nb_processes = cpu_count() if nb_processes < 1 else nb_processes
```
nilgoyette: I think it's clearer this way.

    nb_processes = int(kwargs.get('nb_processes', '0'))
    nb_processes = nb_processes if nb_processes >= 1 else cpu_count()

dipy/reconst/multi_voxel.py
```diff
@@ -5,6 +5,215 @@
...
+
+        if nb_processes == 1:
+            return single_voxel_fit(model, data, *args, **kwargs)
+
+        # Get non null index from mask
```
nilgoyette: Not singular, indexes or indices.

dipy/reconst/multi_voxel.py
```diff
@@ -5,6 +5,215 @@
...
+    """
+    model, input_queue, args, kwargs = arguments
+    return [(idx, model.fit(shared_arr[idx], args, kwargs)) for idx in input_queue]
+
+
+def _init_parallel_fit_worker(arr_to_populate, shape):
+    """
+    Each pool process calls this initializer.
+    Load the array to be populated into
+    that process's global namespace
```
nilgoyette: Why the \n here? The doc can also reach 79 characters.

dipy/reconst/multi_voxel.py
```diff
@@ -5,6 +5,215 @@
...
+    the passed ndarray.  The data will be *copied* into shared
+    memory.
+
+    Parameters
+    ----------
+    arr : ndarray
+        numpy ndarray that you want to convert.
+    lock : boolean
+        controls the access to the shared array. When you shared
+        array is a read only access, you do not need lock. Otherwise,
```
nilgoyette: When *your* shared array *has* a ... ? need *a* lock?

skoudoro:
Thank you @nilgoyette for the review, I made the changes.

nilgoyette:
Oh, damn, I wasn't aware that you can't use `np.full`! I'll stop advising people to use it in DiPy.

dipy/reconst/multi_voxel.py
```diff
@@ -5,6 +5,224 @@
...
+    ctypes.c_void_p: np.dtype(object)}
+
+
+_numpy_to_ctypes = dict(zip(_ctypes_to_numpy.values(),
+                            _ctypes_to_numpy.keys()))
+
+global shared_arr
+
+
+def shm_as_ndarray(mp_array, shape=None):
```
arokem: I wonder about the naming here: there's nothing specific to spherical harmonics here, is there? Or does "shm" stand for something else here?

dipy/reconst/multi_voxel.py
```diff
@@ -5,6 +5,224 @@
...
+    dtype = _ctypes_to_numpy[mp_array._type_]
+    result = np.frombuffer(mp_array, dtype)
+
+    if shape is not None:
+        result = result.reshape(shape)
+
+    return np.asarray(result)
+
+
+def ndarray_to_shm(arr, lock=False):
```
arokem: And same here

dipy/reconst/multi_voxel.py
```diff
@@ -5,6 +5,224 @@
...
+    shape: tuple
+        tuple of array dimensions
+    """
+    global shared_arr
+    with warnings.catch_warnings():
+        warnings.simplefilter('ignore', RuntimeWarning)
+        # shared_arr = np.ctypeslib.as_array(arr_to_populate)
+        # shared_arr = shared_arr.reshape(shape)
+        shared_arr = shm_as_ndarray(arr_to_populate, shape)
+
```
arokem: I would love to have some unit tests for the functions above. If I understand correctly, right now they are tested only through the decorator.

dipy/reconst/multi_voxel.py
skoudoro: right, bad naming, `shm` for `shared multiprocessing array`. I will changed that

dipy/reconst/multi_voxel.py
skoudoro: That's true, I will add them.

BTW, I still wonder if these functions (`shm_as_array` and `ndarray_to_shm`) are in the right place because they can be used for other purposes like `peaks_from_model`

arokem: Maybe in a new `dipy.core.parallel` module?

dipy/reconst/multi_voxel.py
```diff
@@ -3,8 +3,140 @@
...
+        tuple should contains a model and
+        list of indexes
+
+    Returns
+    --------
+    result: list of tuple
+        return a list of tuple(voxel index, model fitted instance)
+    """
+    model, chunks = arguments
+    return [(idx, model.fit(shared_arr[idx]))
```
Garyfallidis: Check list in case of old memory issue.

dipy/reconst/multi_voxel.py
```diff
@@ -3,8 +3,140 @@
...
+        # print(mask, mask.shape)
+        if data.ndim == 1:
+            return single_voxel_fit(model, data)
+        if mask is None:
+            mask = np.ones(data.shape[:-1], bool)
+        elif mask.shape != data.shape[:-1]:
+            raise ValueError("mask and data shape do not match")
+
+        # Get number of processes
+        nb_processes = int(kwargs.pop('nb_processes', '0'))
```
Garyfallidis: Maybe the default should be None for all available cpus.

dipy/reconst/multi_voxel.py
```diff
@@ -3,8 +3,140 @@
...
+        indexes = np.argwhere(mask)
+        # Convert indexes to tuple
+        indexes = [tuple(v) for v in indexes]
+        # Create chunks
+        chunks_spacing = np.linspace(0, len(indexes), num=nb_processes + 1)
+        chunks = [(indexes[int(chunks_spacing[i - 1]): int(chunks_spacing[i])])
+                  for i in range(1, len(chunks_spacing))]
+
+        # Create shared array
+        shared_arr_in = ndarray_to_mparray(data)
```
Garyfallidis: Add in the comment that you make a single copy of the data here. If you make a copy.

dipy/core/parallel.py
```diff
@@ -0,0 +1,89 @@
...
+        any writing access need to activate the lock to be
+        process-safe(default: False)
+
+    Returns
+    --------
+    shm : multiprocessing.Array
+        copied shared array
+    """
+
+    array1d = arr.ravel(order='A')
```
Garyfallidis: This copy is not needed. Use directly arr.

dipy/reconst/multi_voxel.py
Borda: write it in a single line?

dipy/reconst/tests/test_ivim.py
```diff
@@ -25,67 +25,77 @@
...
+    noisy_single = np.array([4243.71728516, 4317.81298828, 4244.35693359,
+                             4439.36816406, 4420.06201172, 4152.30078125,
+                             4114.34912109, 4104.59375, 4151.61914062,
+                             4003.58374023, 4013.68408203, 3906.39428711,
+                             3909.06079102, 3495.27197266, 3402.57006836,
+                             3163.10180664, 2896.04003906, 2663.7253418,
+                             2614.87695312, 2316.55371094, 2267.7722168])
+
+    noisy_multi = np.zeros((2, 2, 1, len(gtab.bvals)))
+    noisy_multi[0, 1, 0] = noisy_multi[
```
Borda: this break line is not very good for reading...

skoudoro: I agree, I will change that

dipy/reconst/multi_voxel.py
skoudoro: I do not see the benefit. it will be harder to read and understand if I change that

oesteban:
> _**Note:**_ `multiprocessing` is using `fork()` on Linux when it starts a new child process. On Windows -- which does not support fork() -- multiprocessing is using the win32 API call CreateProcess. It creates an entirely new process from any given executable (which is quite slow)

I would expect an enormous overhead in either platform - would it be worth trying to parallelize using threading? (that said, I have externally tried with threads, and it doesn't seem to yield any gains, which is surprising at the very least)

arokem:
See also discussion in https://github.com/nipreps/eddymotion/issues/101. Indeed, some models (DTI and DKI, in particular) will not benefit from more parallelization, because we're already using multi-threaded matrix operations at the numpy level. This kind of parallelization would be most beneficial for models that can't utilize that kind of parallelization, or at least can't do that easily.

arokem:
Incidentally, is this PR superseded by https://github.com/dipy/dipy/pull/2539?

skoudoro:
> Incidentally, is this PR superseded by https://github.com/dipy/dipy/pull/2539?

Yes, closing in favor of #2539

PR #1314, opened 2017-08-01
DOC: Fix typos and formatting in .rst files and Python examples.

jhlegarreta:
Fix typos and formatting in .rst files and Python examples'
documentations:
- Fix typos in files.
- Replace tabs for two white spaces.
- Fix errors in citation and DIPY internal cross-refs.
- Add missing cross-refs (e.g. Aganj MRM 2010).
- Add links where necessary (ISBI HARDI Contest 2013 and FreeSurfer).
- Fix LaTeX math formatting errors.
- Make the example/DIPY code object markdown consistent (inverted commas).
- Capitalize the acronym long names' first/corresponding letters (e.g.
  Constrained Spherical Deconvolution).
- Capitalize acronyms (e.g. FA).
- Make the writing of terms consistent in terms of uppercase/lowercase
  (e.g. b0 vs. B0; cospus callosum vs. Corpus Callosum, etc.).
- Use the [NameYear] convention for citations.
- Place all citations under a 'References' section, and use the subsection
  markdown for the section.
- Use inverted commas consistently to reference code objects.
- Make the b-value units consistent across files (s/mm^2).
- Create and include missing figures for 'reconst_fwdti.py'.
- Write DIPY instead of any of its variants in the documentation.
- Link the first appearance of dipy in every file (use dipy_).

PR #1421, opened 2018-02-10
DOC: Fix reStructuredText formatting issues in coding style guideline.

### LINKED ISSUE #1390: coding_style_guideline.rst does not render correctly
## Description
When viewing the `coding_style_guideline.rst` file directly on the web, the markup doesn't seem to correctly render. For example, some links are not formatted as links, and the lists do not generate lists.

## Way to reproduce
* Go [here](https://github.com/nipy/dipy/blob/master/doc/devel/coding_style_guideline.rst)
* Check links and lists.

jhlegarreta:
Fix RST markup issues in the `coding_style_guideline.rst` file.

Fixes #1390.

### DIFF
--- a/doc/devel/coding_style_guideline.rst
+++ b/doc/devel/coding_style_guideline.rst
@@ -5,20 +5,21 @@ DIPY Coding Style Guideline
 ===========================
 
 The main principles behind dipy_ development are:
-* **Robustness**: the results of a piece of code must be verified
-systematically, and hence stability and robustness of the code must be ensured,
-reducing code redundancies.
-* **Readability**: the code is written and read by humans, and it is read much
-more frequently than it is written.
-* **Consistency**: following these guidelines will ease reading the code, and
-will make it less error-prone.
-* **Documentation**: document the code. Documentation is essential as it is one
-of the key points for the adoption of DIPY as the toolkit of choice in
-diffusion by the scientific community. Documenting helps clarifying certain
-choices, helps avoiding obscure places, and is a way to allow other members
-*decode* it with less effort.
-* **Language**: the code must be written in English. Norms and spelling should
-be abided by.
+
+  * **Robustness**: the results of a piece of code must be verified
+    systematically, and hence stability and robustness of the code must be
+    ensured, reducing code redundancies.
+  * **Readability**: the code is written and read by humans, and it is read
+    much more frequently than it is written.
+  * **Consistency**: following these guidelines will ease reading the code,
+    and will make it less error-prone.
+  * **Documentation**: document the code. Documentation is essential as it is
+    one of the key points for the adoption of DIPY as the toolkit of choice in
+    diffusion by the scientific community. Documenting helps clarifying
+    certain choices, helps avoiding obscure places, and is a way to allow
+    other members *decode* it with less effort.
+  * **Language**: the code must be written in English. Norms and spelling
+    should be abided by.
 
 ------------
 Coding style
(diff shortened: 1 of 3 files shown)


jhlegarreta:
@jchoude Thanks for reporting the issue !

Rather than checking the [GitHub file](https://github.com/nipy/dipy/blob/master/doc/devel/coding_style_guideline.rst), it looks like checking the [dipy website version](http://nipy.org/dipy/devel/coding_style_guideline.html) gives a more intuitive idea of what's going on.

The [mission file](https://github.com/nipy/dipy/blob/master/doc/mission.rst) seems to have issues as well, but it's [rendered correctly](http://nipy.org/dipy/mission.html).

Please, correct me if I'm wrong, and if both versions should be rendered correctly.

doc/devel/coding_style_guideline.rst
```diff
@@ -52,56 +54,60 @@ example, interleaved with multi-line comments that contain explanations of the
...
+following guidelines:
+
+  * The acronym for the Diffusion Imaging in Python toolkit should be written
+    as **DIPY**.
+  * The classes, objects, and any other construct referenced from the code
+    should be written with inverted commas, such as in *In DIPY, we use an
+    object called ``GradientTable`` which holds all the acquisition specific
+    parameters, e.g. b-values, b-vectors, timings and others.*
+  * Cite the relevant papers. Use the *[NameYear]* convention for
+    cross-referencing them, such as in [Garyfaillidis2014]_, and put them
```
skoudoro: `Garyfallidis`instead of `Garyfaillidis`

doc/devel/coding_style_guideline.rst
```diff
@@ -52,56 +54,60 @@ example, interleaved with multi-line comments that contain explanations of the
...
+    for filenames, labels for references, etc.
+  * When including figures, use the regular font for captions (i.e. do not use
+    bold faces), unless otherwise required for a specific text part (e.g. a
+    DIPY object, etc.).
+  * When referring to relative paths, use the backquote inline markup
+    convention, such as in ``doc/devel``. Do not add the
+    greater-than/less-than signs to enclose the path.
 
 
 """
```
skoudoro: Do we really need this quotes?  I think you can remove them because they look strange as you can see on [this page](http://nipy.org/dipy/devel/coding_style_guideline.html)

doc/devel/coding_style_guideline.rst
```diff
@@ -117,4 +123,5 @@ References
    <http://journal.frontiersin.org/Journal/10.3389/fninf.2014.00008/abstract>`_
    Frontiers in Neuroinformatics, vol.8, no.8.
 
+.. include:: ../links_names.inc
 """
```
skoudoro: same comment as above

skoudoro:
Concerning the rendering, I wonder why GitHub does not take in consideration `link_names.inc`

doc/devel/coding_style_guideline.rst
jhlegarreta: Thanks ! I inspected this visually yesterday without being able to tell why it was not being linked 😬

doc/devel/coding_style_guideline.rst
jhlegarreta: I think they make easier reading the text, standing out what is a  `code snippet`  or *any* other construct that belong to the structure of the toolkit. The reason why they are double inverted quotes instead of simple inverted quotes is due to the Sphinx syntax in the docstrings (if I remember well).

A question relative to the way dirs had to be described was raised in a [comment](https://github.com/nipy/dipy/pull/1314#issuecomment-321590604) #1314 .

jhlegarreta:
Yes, I also realized that GitHub does not link the cross-refs in `link_names.inc` but have not investigated 😟

doc/devel/coding_style_guideline.rst
skoudoro: I think we are not talking about the same line. I am talking about the quotes [here](https://github.com/nipy/dipy/pull/1421/files#diff-e1343e5bdca90449ac1bf352283b04a4R113), line 113

doc/devel/coding_style_guideline.rst
skoudoro: Same comment as above. The double quotes are not necessary here.

skoudoro:
It seems that github don't allow the include directive for security reason (look [here](https://github.com/github/markup/issues/172)).

doc/devel/coding_style_guideline.rst
jhlegarreta: OK, now I got it. Sorry. My bad.

jhlegarreta:
OK. Fixed the quotes. Thanks for the explanation on the `include` directive issue @skoudoro

skoudoro:
- Can you add release 0.12 and 0.13 on [this page](https://github.com/nipy/dipy/blob/master/doc/stateoftheart.rst)
- Can you replace `HBM 2015`by `ÒHBM 2015` on [this page](https://github.com/nipy/dipy/blob/master/doc/index.rst) 

It will be nice if you can fix these additional issues on this PR. Thank you !

jhlegarreta:
Sorry Serge. I missed that in your first comment. Done.

skoudoro:
Thank you @jhlegarreta, it looks better now after test.

merging !

Issue #1291, opened 2017-07-02
Existing MAPMRI tutorial does not render correctly and MAPL looks hidden in the existing tutorial.


Garyfallidis:
http://nipy.org/dipy/examples_built/reconst_mapmri.html#example-reconst-mapmri

arokem:
That example does include MAPL. It was changed in the PR that introduced MAPL: 

https://github.com/nipy/dipy/pull/1153/files#diff-aa743bb3a0919f172e5dca3244da2ab3 

Rendering's borked though. That should be fixed.

Garyfallidis:
Okay, cool.

Garyfallidis:
Also this tutorial seems like needing better datasets. Which is something we talked about in the past. Real datasets in  high resolution that can expose the power of Mapmri.

arokem:
Yeah - we should switch this to use the recently added CFIN data: https://github.com/nipy/dipy/blob/master/dipy/data/fetcher.py#L1001

skoudoro:
It seems to be fixed by #1424, I close this issue

PR #1424, opened 2018-02-11
ENH: Use the CFIN dataset rather than the CENIR dataset.

jhlegarreta:
Following the discussion in #1291, use the ``CFIN`` dataset instead of
the ``CENIR`` dataset.

Also, make an explicit mention to the `MAPL` acronym so that the use of
the Laplacian-regularized MAP-MRI method seems no longer to be hidden.

### DIFF
--- a/doc/examples/reconst_mapmri.py
+++ b/doc/examples/reconst_mapmri.py
@@ -21,33 +21,32 @@ the parallel and perpendicular Non-Gaussianity.
 The estimation of these properties from noisy DWIs requires that the
 fitting of the MAPMRI basis is regularized. We will show that this can
 be done using both constraining the diffusion propagator to positive
-values [Ozarslan2013]_ and analytic Laplacian Regularization [Fick2016a]_.
+values [Ozarslan2013]_ and analytic Laplacian Regularization (MAPL)
+[Fick2016a]_.
 
 First import the necessary modules:
 """
 
 from dipy.reconst import mapmri
 from dipy.viz import window, actor
-from dipy.data import fetch_cenir_multib, read_cenir_multib, get_sphere
+from dipy.data import fetch_cfin_multib, read_cfin_dwi, get_sphere
 from dipy.core.gradients import gradient_table
 import matplotlib.pyplot as plt
+import numpy as np
 from mpl_toolkits.axes_grid1 import make_axes_locatable
 
 """
 Download and read the data for this tutorial.
 
 MAPMRI requires multi-shell data, to properly fit the radial part of the basis.
-The total size of the downloaded data is 1760 MBytes, however you only need to
-fetch it once. Parameter ``with_raw`` of function ``fetch_cenir_multib`` is set
-to ``False`` to only download eddy-current/motion corrected data:
+to ``False`` to only download eddy-current/motion corrected data.
+The total size of the downloaded data is 187.66 MBytes, however you only need
+to fetch it once.
 """
 
-fetch_cenir_multib(with_raw=False)
+fetch_cfin_multib()
 
 """
-For this example we select only the shell with b-values equal to the one of the
-Human Connectome Project (HCP).
-
 ``data`` contains the voxel data and ``gtab`` contains a ``GradientTable``
 object (gradient information e.g. b-values). For example, to show the b-values
 it is possible to write::
@@ -59,15 +58,14 @@ explicitly state the ``big_delta`` and ``small_delta`` parameters in the
 gradient table.
 """
 
-bvals = [1000, 2000, 3000]
-img, gtab = read_cenir_multib(bvals)
+img, gtab = read_cfin_dwi()
 big_delta = 0.0365  # seconds
 small_delta = 0.0157  # seconds
 gtab = gradient_table(bvals=gtab.bvals, bvecs=gtab.bvecs,
                       big_delta=big_delta,
                       small_delta=small_delta)
 data = img.get_data()
-data_small = data[40:65, 50:51, 35:60]
+data_small = data[40:65, 50:51]
 
 print('data.shape (%d, %d, %d, %d)' % data.shape)
 
@@ -118,6 +116,10 @@ A suggested strategy is to use a low Laplacian weight together with the
 positivity constraint. In this way both desired properties are guaranteed in
 the final solution.
 
+We use package CVXPY (http://www.cvxpy.org/) to solve convex optimization
+problems when "positivity_constraint=True", so we need to first install
+CVXPY.
+
 For now we will generate the anisotropic models for all combinations.
 """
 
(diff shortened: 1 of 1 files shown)


jhlegarreta:
@Garyfallidis @arokem 

I removed the 
```
For this example we select only the shell with b-values equal to the one of the Human Connectome Project (HCP).
```
sentence from the documentation since `read_cfin_dwi` does not accept the b-value list as an argument.

Finally, I got an error while debugging the example:
```
  File "C:/SDKs/dipy/doc/examples/reconst_mapmri.py", line 124, in <module>
    positivity_constraint=True)
  File "C:\SDKs\dipy\dipy\reconst\mapmri.py", line 251, in __init__
    'CVXPY package needed to enforce constraints')
ValueError: CVXPY package needed to enforce constraints
```
at
```
map_model_positivity_aniso = mapmri.MapmriModel(gtab,
                                                radial_order=radial_order,
                                                laplacian_regularization=False,
                                                positivity_constraint=True)
```

If CI builds do not report issues, I can assume this is a local issue, and that our `requirements.txt` does not need to be updated to include the `CVXPY` package. Correct me if I'm wrong.

jhlegarreta:
Looks like CircleCI containers do have the `CVXPY` package installed. Any comments on requiring it?
 
I've noticed that the `CENIR` data has **also** data for **b-vals** **200** and **400** s/mm2. The **plots in the example will not show data for all data** since they only contain three subplots; **should we show all data, or just take into account only the 1000, 2000 and 3000** s/mm2, as the previous version was doing?

skoudoro:
Can you review this PR @arokem?

arokem:
Will do!

arokem:
Moving to this new dataset will require some additional changes. See https://github.com/jhlegarreta/dipy/pull/1

jhlegarreta:
Thanks for the review @arokem !

arokem:
Thank you! I am +1 for merging this. If someone else could take a look, that would be great.

RicciWoo:
After PR: [mapmri using cvxpy instead of cvxopt #1285](https://github.com/nipy/dipy/pull/1285), we are now using package CVXPY instead of CVXOPT, I think we need to update the comment on line 27-31 in reconst_mapmripy for this.

skoudoro:
Hi @jhlegarreta, I got this error when I tried to run this example with the new dataset:
```
  File "reconst_mapmri.py", line 162, in <module>
    vmin=0, vmax=5e7)
  File "C:\Users\skoudoro\AppData\Local\Continuum\Anaconda3\envs\dipy-dev-3\lib\site-packages\matplotlib\__init__.py", line 1897, in inner
    return func(ax, *args, **kwargs)
  File "C:\Users\skoudoro\AppData\Local\Continuum\Anaconda3\envs\dipy-dev-3\lib\site-packages\matplotlib\axes\_axes.py", line 5124, in imshow
    im.set_data(X)
  File "C:\Users\skoudoro\AppData\Local\Continuum\Anaconda3\envs\dipy-dev-3\lib\site-packages\matplotlib\image.py", line 596, in set_data
    raise TypeError("Image data can not convert to float")
TypeError: Image data can not convert to float
```
I am working on Windows. Do you get this error ? Can you update the plotting part.
Thank you

jhlegarreta:
@skoudoro thanks for the review. I did get the same error.

Please, check if commit 795aa4f solves the plot part issues.

Also, I am getting an error at
```
/dipy/doc/examples/reconst_mapmri.py", line 393, in <module>
    print('odf.shape (%d, %d, %d, %d)' % odf.shape)
```
telling me that
```
TypeError: not enough arguments for format string
```

Let's see what CI says, or whether you do get the same error; if you do, I'll amend with
```
print('odf.shape ({0})'.format(odf.shape))
```

BTW, although I know other parts of the code use `cvxpy`, in order to run the exampled I had to manually install it (pip was not listing it either as installed), so I'm wondering whether `cvxpy` should be added to the `requirements.txt` file.

May be this is a Windows-specific issue (?)

jhlegarreta:
Looks like travis is happy; may be because no image is run on Windows machines (?). @skoudoro let me know when you check the patch set.

arokem:
It's because the examples are not run as part of the CI. That would take
way too long.

skoudoro:
Thanks @jhlegarreta. Today, I do not have access to my windows machine so I will try it tomorrow.

Concerning `cvxpy`, I think we want to let it as an optional requirement so do not add it to `requirement.txt`

jhlegarreta:
@skoudoro 3eb7ddb was rebased on master.

As for the data region, just to make sure, you mean using a slice instead of a volume for the example now?

Regardless of the data region, not sure whether the result images are correct: I get all 3 images in `MAPMRI_maps_regularization.png` black.

skoudoro:
> As for the data region, just to make sure, you mean using a slice instead of a volume for the example now?

`print(data.shape)` -> (96,96,19,496)
`data_small = data[40:65, 50:51, 35:60]` -> as you can see here `35:60`do not exist on the new dataset. In both case (old, and new dataset), it seems we work with one slice.

>  I get all 3 images in MAPMRI_maps_regularization.png black

In interactive mode, all results seem good when we display them, so maybe a saving issue, I will look at that

jhlegarreta:
OK. Let me know if you are able to reproduce the issue (working on Win 10); if not, may be it's local to my environment and we can ignore it. Thanks Serge.

doc/examples/reconst_mapmri.py
```diff
@@ -161,21 +163,23 @@
...
 
 ax2 = fig.add_subplot(1, 3, 2, title=r'RTOP - Positivity')
 ax2.set_axis_off()
-ind = ax2.imshow(mapfit_positivity_aniso.rtop()[:, 0, :].T,
-                 interpolation='nearest', origin='lower', cmap=plt.cm.gray,
-                 vmin=0, vmax=5e7)
+rtop_positivity = np.array(mapfit_positivity_aniso.rtop()[:, 0, :].T,
+                           dtype=float)
+ind = ax2.imshow(rtop_positivity, interpolation='nearest',
+                 origin='lower', cmap=plt.cm.gray, vmin=0, vmax=5e7)
```
skoudoro: `vmin` and `vmax` are incorrect. This wrong range is the reason for your black image. Can you just remove this 2 parameters and let  `matplotlib` manage it

doc/examples/reconst_mapmri.py
```diff
@@ -161,21 +163,23 @@
...
 
 ax3 = fig.add_subplot(1, 3, 3, title=r'RTOP - Both')
 ax3.set_axis_off()
-ind = ax3.imshow(mapfit_both_aniso.rtop()[:, 0, :].T,
-                 interpolation='nearest', origin='lower', cmap=plt.cm.gray,
-                 vmin=0, vmax=5e7)
+rtop_both = np.array(mapfit_both_aniso.rtop()[:, 0, :].T, dtype=float)
+ind = ax3.imshow(rtop_both, interpolation='nearest', origin='lower',
+                 cmap=plt.cm.gray, vmin=0, vmax=5e7)
```
skoudoro: `vmin` and `vmax` are incorrect. This wrong range is the reason for your black image. Can you just remove this 2 parameters and let  `matplotlib` manage it

skoudoro:
Due to the release, I will merge it and make the last correction on new PR.

Thank you @jhlegarreta for this PR and @arokem for this review!

PR #1434, opened 2018-02-16
QuickBundlesX

Garyfallidis:
This PR is replacing https://github.com/nipy/dipy/pull/1380 
**This PR includes the fastest algorithm ever for clustering tractograms!!! :fire:** 

Garyfallidis E. et al. QuickBundlesX: Sequential clustering of millions of streamlines in multiple levels of detail at record execution time. Proceedings of the, International Society of Magnetic Resonance in Medicine (ISMRM). Singapore, 4187, 2016.

PR #1439, opened 2018-03-01
BF: Avoid using memview in struct (Cython 0.28)

MarcCote:
Replace memview's object in Centroid struct with a memview pointer.
Added a unit test that uses a Python metric with QuickBundles clustering.
Fixed bench_streamlines.py script.

@skoudoro we can drop this PR in favor of #1434 since you already made the necessary modification (and it is much cleaner than what I have). However, you might want to use the fixed `bench_streamlines.py` script, the unit test `test_quickbundles_with_python_metric` in `test_quickbundles.py` and the fix in `metricspeed.pyx`

### DIFF
--- a/dipy/segment/tests/test_quickbundles.py
+++ b/dipy/segment/tests/test_quickbundles.py
@@ -151,6 +151,41 @@ def test_quickbundles_streamlines():
         assert_equal(clusters.centroids[0].dtype, np.float32)
 
 
+def test_quickbundles_with_python_metric():
+
+    class MDFpy(dipymetric.Metric):
+        def are_compatible(self, shape1, shape2):
+            return shape1 == shape2
+
+        def dist(self, features1, features2):
+            dist = np.sqrt(np.sum((features1 - features2)**2, axis=1))
+            dist = np.sum(dist / len(features1))
+            return dist
+
+    rdata = streamline_utils.set_number_of_points(data, 10)
+    qb = QuickBundles(threshold=2 * threshold, metric=MDFpy())
+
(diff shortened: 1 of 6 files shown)


dipy/segment/tests/test_quickbundles.py
```diff
@@ -151,6 +151,41 @@ def test_quickbundles_streamlines():
...
+    # Set `refdata` to return indices instead of actual data points.
+    clusters.refdata = None
+    assert_array_equal(list(itertools.chain(*clusters)),
+                       list(itertools.chain(*clusters_truth)))
+
+    # Cluster read-only data
+    for datum in rdata:
+        datum.setflags(write=False)
+
+    clusters = qb.cluster(rdata)
```
skoudoro: unused, what is the goal here ?

skoudoro:
Ok, I see why for `metricspeed.pyx`

dipy/segment/tests/test_quickbundles.py
MarcCote: It was just to check the code can run without segfault even if data is read only.

MarcCote:
Yeah, we need to that trick in metricspeed.pyx (similar to what is done in clusters_centroid2clustermap_centroid) because the memview we created manually misses the reference to its Python counterpart (`memview` pointer in memview_slice struct).

dipy/segment/tests/test_quickbundles.py
skoudoro: it makes sense now, thank you

MarcCote:
Close in favor of #1434

PR #1440, opened 2018-03-03
Raise Error when MDFmetric is used in QB or QBX

Garyfallidis:
…Actually the AveragePointwiseEuclideanMetric has the correct behavior of the MDF metric as explained in the QuickBundles paper.  @MarcCote this one is for you to check.

### DIFF
--- a/doc/examples/segment_clustering_metrics.py
+++ b/doc/examples/segment_clustering_metrics.py
@@ -117,11 +117,6 @@ print("Cluster sizes:", map(len, clusters))
 
     Cluster sizes: [64, 191, 44, 1]
 
-.. _clustering-examples-MinimumAverageDirectFlipMetric:
-
-::
-
-.. _clustering-examples-MinimumAverageDirectFlipMetric:
 
 Cosine Metric
 =============
--- a/doc/examples/segment_clustering_metrics.py
+++ b/doc/examples/segment_clustering_metrics.py
@@ -119,41 +119,7 @@ print("Cluster sizes:", map(len, clusters))
 
 .. _clustering-examples-MinimumAverageDirectFlipMetric:
 
-Minimum Average Direct Flip Metric (MDF)
-========================================
-**What:** It is the metric used in the QuickBundles algorithm [Garyfallidis12]_.
-Instances of `MinimumAverageDirectFlipMetric` first compute the
-direct distance *d1* by taking the average of the pointwise
-Euclidean distances between two sequences *of same length*. Reverse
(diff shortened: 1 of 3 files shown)


doc/examples/segment_clustering_metrics.py
```diff
@@ -69,6 +73,8 @@ def get_streamlines():
 print("Cluster sizes:", map(len, clusters))
 
 """
+If there are the data do not need ordering checks then make sure that
```
MarcCote: wording

doc/examples/segment_clustering_metrics.py
```diff
@@ -47,7 +47,11 @@ def get_streamlines():
 `Feature` is the `IdentityFeature` the streamlines won't be resampled thus
 saving some computational time.
 
-**Note:** Inputs must be sequences of same length.
+**Note:** Inputs must be sequences of same length. Also when the input
+feature of this metric ``is_order_invariant=True`` then the metric is
```
MarcCote: That's misleading. It's not because the `Features` object has `is_order_invariant=True` that this metric becomes the MDF. However, it is true it would produce the same result.

That said, I don't see how you can design a `Feature` object that would allow you to produce the same results as the MDF described in `[Garyfallidis12]_`.

Garyfallidis:
The main problem is that if someone reads the paper may go and use the MDF metric as input to QuickBundles although one should be using the AverageEuclidian metric. So, I think we should at least remove the MDF metric from the same tutorial with AverageEuclidian.

And I am not sure with using is isinstance(metric, MDF) but I can try it and see how it goes. Let me update this PR.... one sec...

Garyfallidis:
What about now @MarcCote ?

doc/examples/segment_clustering_metrics.py
```diff
@@ -119,42 +119,8 @@ def get_streamlines():
...
-
-print("MDF distance between the first two streamlines: ", d)
-
-"""
-
 ::
 
-    MDF distance between the first two streamlines: 11.681308709622542
-
 .. _clustering-examples-MinimumAverageDirectFlipMetric:
```
MarcCote: This line should be removed too, no?

doc/examples/segment_clustering_metrics.py
```diff
@@ -119,42 +119,8 @@ def get_streamlines():
 
 .. _clustering-examples-MinimumAverageDirectFlipMetric:
```
MarcCote: You have the same line below.

Garyfallidis:
Thanks! Done with removing the two lines.

PR #1441, opened 2018-03-04
Cleaning UI and improving positioning of Panel2D

MarcCote:
This PR exists to reduce changes in #1355.

Remove unused attributes in `UI`.
Add a better way of handling absolute/relative coordinates in `Panel2D`.
Remove `FileSelectMenu2D` that will be replaced with ListBox2D (will be added in #1355).


dipy/viz/ui.py
```diff
@@ -656,40 +642,46 @@ def add_to_renderer(self, ren):
...
-            element.set_center((position[0], position[1]))
-        else:
-            raise ValueError("Position can only be absolute or relative")
+        coords = np.array(coords)
+
+        if np.issubdtype(coords.dtype, np.floating):
+            if np.any(coords < 0) or np.any(coords > 1):
+                raise ValueError("Normalized coordinates must be in [0,1].")
+
+            coords = coords * self.size
```
ranveeraggarwal: Are we sure we want to do this instead of an extra parameter?

Here's another idea. If the parameter is specified as `relative`, we can normalize the values ourselves and go for a relative placement in this fashion.

dipy/viz/ui.py
```diff
@@ -656,40 +642,46 @@ def add_to_renderer(self, ren):
...
+
+        #TODO: Check if coords is outside the panel.
+
```
ranveeraggarwal: I think coordinates going outside the panel is fine (in the absolute case). We should just throw a warning if this happens.

dipy/viz/ui.py
MarcCote: As an end user, you would have to know what pixel coordinate corresponds to let say 0.5 of the panel. I found it a bit annoying compared to just use 0.5. Also, if we ever made the panel resizable, specifying 0.5 is more intuitive I find.

Are there any cons I don't see?

ranveeraggarwal: Yeah, I guess let's go with this for now. This is intuitive enough.

doc/examples/viz_ui.py
```diff
@@ -123,8 +123,8 @@ def modify_button_callback(i_ren, obj, button):
 
 panel = ui.Panel2D(center=(440, 90), size=(300, 150), color=(1, 1, 1),
                    align="right")
-panel.add_element(button_example, 'relative', (0.2, 0.2))
-panel.add_element(second_button_example, 'absolute', (480, 100))
+panel.add_element(button_example, (0.2, 0.2))
+panel.add_element(second_button_example, (480, 100))
```
skoudoro: Now, this button seems to be outside the panel, I think you should update the position if it is not on purpose.

MarcCote: Done

skoudoro:
I think it is ready to go. is it ok for you @ranveeraggarwal? Can you have a look @Garyfallidis?

MarcCote:
Wait! I think it would be better to merge #1448 instead of this one.

skoudoro:
so can you close this one?

MarcCote:
Sure.

Issue #1420, opened 2018-02-09
window.record() resolution limit


dmreagan:
## Description
`window.record()` can be used to save a render to a PNG image. If you set the `size` to be higher than your monitor size, the resulting image will be capped to just under your monitor's resolution. For example, if you set `size=(5000, 5000)`, but your monitor's resolution is (1920, 1080), the image width will end up being something like 1916 pixels. 

## Way to reproduce
Run any viz example that saves a render with `window.record()`, but change the `size` to be very high. Tested on Windows and Mac with VTK 7.

ShreyasFadnavis:
Can you review my pull request for this? #1453

dmreagan:
@ShreyasFadnavis  and @karandeepSJ report that they had no trouble with large resolutions on their systems.  Can you both report which OS, Python, and VTK versions you are using to get these successful results?

ShreyasFadnavis:
@dmreagan I tested this on both Windows 10 and Ubuntu 16.04. 
Python Version: 3.6.4
VTK Version: 7.1.1

karandeepSJ:
Ubuntu 16.04
Python Version: 3.5.2
VTK Version: 8.1.0

thechargedneutron:
@dmreagan  Is this issue on Ubuntu as well? Upon changing the size to (6000, 6000) I got a file with a size of (6000, 6000).

thechargedneutron:
@skoudoro  very nicely figured the problem. @dmreagan  I have added a patch in my branch [here](https://github.com/thechargedneutron/dipy/tree/window_resolution). Can you please verify if this is working fine since this issue is not there on Ubuntu I guess.

skoudoro:
This issue was moved to fury-gl/fury#13

PR #1453, opened 2018-03-10
changed window.record() to a large value

ShreyasFadnavis:


ShreyasFadnavis:
Maybe I misunderstood?

karandeepSJ:
Did you try the 5000x5000 image on Linux? window.record() works fine on Linux. The problem is on Windows

ShreyasFadnavis:
Yes, I did.. It works fine on my Windows PC!

jchoude:
Is this for issue #1420? IF so, I don't think that's the fix that @dmreagan intended... This hardcoded value is not better, and might fail depending on the driver and GPU of the user. I guess that @dmreagan wanted something that could check that max resolution supported by the user's machine and work within those boundaries.

dmreagan:
@jchoude is correct that I wasn't looking for someone to change the values in the examples. I was hoping the underlying VTK functionality just wasn't being configured correctly for the case described in #1420. But it is good feedback from @ShreyasFadnavis and @karandeepSJ that large resolutions work correctly on your systems. I think it's probaby best to close this PR and continue the discussion in the issue.

PR #1455, opened 2018-03-13
Using pyautogui to adapt to users' monitor size in viz examples

ShreyasFadnavis:


ShreyasFadnavis:
@dmreagan does this make sense/ is similar to what you wanted?

dmreagan:
Thanks for contributing, @ShreyasFadnavis, but this isn't what I'm looking for. It seems I should reword the issue description to make it more clear. I want to find out why I can't save large images on my system. Since you aren't seeing the issue on your system, it might be something specific to my setup.

skoudoro:
I have the same issue so I do not think, it is something speficic to you @dmreagan 

I think you understood the problem @ShreyasFadnavis, but we do not want any other dependencies. You have to find a way to do the same with vtk and python standard library, then update `record` function where you can find [here](https://github.com/nipy/dipy/blob/master/dipy/viz/window.py#L655). So sorry, you can forget `pyautogui`

dmreagan:
>You have to find a way to do the same with vtk and python standard library

I don't think that's correct. I don't want the examples to create renders at the detected monitor size. I want to find out why `record` isn't allowing you and me to render to sizes larger than our monitors.

skoudoro:
oh ok, Thanks for the update!

Issue #1454, opened 2018-03-13
Sliders in examples don't react properly to clicks


dmreagan:
## Description
Some viz examples use `LineSlider2D` and `DiskSlider2D` to change properties of actors in the scene. The sliders work correctly if you drag the handle. But if you click directly on the path, the value of the slider changes without updating the corresponding actor. 

## Way to reproduce
Run an example like *viz_ui.py* in interactive mode. Click directly on the slider path instead of dragging the handle.

skoudoro:
done in #1457, closing

PR #1457, opened 2018-03-13
Fix for "Sliders in examples don't react properly to clicks"

karandeepSJ:
This PR fixes issue #1454 
- A callback was added to call `translate_green_cube` on `"LeftButtonPressEvent"`. 
- Priority given to ` self.line_click_callback` in `ui.py` to ensure that the position of disk gets updated before moving the cube.

### DIFF
--- a/dipy/viz/ui.py
+++ b/dipy/viz/ui.py
@@ -1851,7 +1851,7 @@ class LineSlider2D(UI):
 
         """
         self.add_callback(self.slider_line, "LeftButtonPressEvent",
-                          self.line_click_callback)
+                          self.line_click_callback,1)
         self.add_callback(self.slider_disk, "LeftButtonPressEvent",
                           self.disk_press_callback)
         self.add_callback(self.slider_disk, "MouseMoveEvent",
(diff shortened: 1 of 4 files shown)


dipy/viz/ui.py
```diff
@@ -1851,7 +1851,7 @@ def handle_events(self, actor):
 
         """
         self.add_callback(self.slider_line, "LeftButtonPressEvent",
-                          self.line_click_callback)
+                          self.line_click_callback,1)
```
ranveeraggarwal: Add a space after the `,` in accordance with PEP8.

doc/examples/viz_ui.py
```diff
@@ -143,14 +143,17 @@ def translate_green_cube(i_ren, obj, slider):
...
 
-
 line_slider = ui.LineSlider2D(initial_value=-2,
                               min_value=-5, max_value=5)
 
 line_slider.add_callback(line_slider.slider_disk,
                          "MouseMoveEvent",
                          translate_green_cube)
 
+line_slider.add_callback(line_slider.slider_line,
```
ranveeraggarwal: Can you do the same for the disk slider also? Let's merge both these into this PR.

dmreagan:
Please also update *viz_advanced.py*. Once that is done, I think this is ready to merge.

skoudoro:
is it ready to go @dmreagan?

Issue #1458, opened 2018-03-13
DiskSlider does not rotate actor in opposite direction


karandeepSJ:
## Description
Some viz examples use 'DiskSlider2D' to rotate actors. The slider rotates the actor in a single direction, irrespective of whether you slide clockwise or counter-clockwise.

## Way to reproduce
Run an example like _viz_ui.py_ in interactive mode. Click on the rotation slider and move your mouse in clockwise and counter-clockwise directions.

skoudoro:
fix by #1460. closing

PR #1460, opened 2018-03-13
Fix for "DiskSlider does not rotate actor in opposite direction"

karandeepSJ:
This PR fixes issue #1458 
- The problem was that the actor was being rotated every-time by the angle between the line joining center of `base_disk` and `handle` and the horizontal axis. It should have been rotated by angle equal to the change in angle between the line and horizontal axis.
- To implement this, an attribute - `pvalue` was added to the `DiskSlider2D` class to store the previous value of rotation. This is used to calculate the angle by which the actor must be rotated.

### DIFF
--- a/dipy/viz/ui.py
+++ b/dipy/viz/ui.py
@@ -1882,6 +1882,8 @@ class DiskSlider2D(UI):
         Inner radius of the slider's handle.
     previous_value: float
         Value of Rotation of the actor before the current value.
+    initial_value: float
+        Initial Value of Rotation of the actor assigned on creation of object.
 
     """
     def __init__(self, position=(0, 0),
--- a/dipy/viz/ui.py
+++ b/dipy/viz/ui.py
@@ -1921,7 +1921,7 @@ class DiskSlider2D(UI):
         self.center = np.array(position)
         self.min_value = min_value
         self.max_value = max_value
-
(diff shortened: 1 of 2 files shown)


dipy/viz/ui.py
```diff
@@ -1921,7 +1921,7 @@ def __init__(self, position=(0, 0),
         self.center = np.array(position)
         self.min_value = min_value
         self.max_value = max_value
-
+        self.initial_value = initial_value
```
ranveeraggarwal: This needs to be added to the class docstring.

dipy/viz/ui.py
```diff
@@ -1939,6 +1939,7 @@ def __init__(self, position=(0, 0),
 
         # By setting the value, it also updates everything.
         self.value = initial_value
+        self.pvalue = initial_value
```
ranveeraggarwal: This too needs to be added to the docstring. 
Also, IMO an explicit `previous_value` would be a better name for this attribute.

dipy/viz/ui.py
ranveeraggarwal: @karandeepSJ please fix this too.

ranveeraggarwal:
This looks good to me. @skoudoro @dmreagan I think this is ready for merge.

PR #1468, opened 2018-03-19
DOC: correct error when doing 'make html'

RicciWoo:
This PR is to correct the error when doing 'make html':
ValueError: The section Notes appears twice in the docstring of None in None.
With the help of @skoudoro & @Garyfallidis, we find the bug and fix it.

PR #1477, opened 2018-03-25
BF: Fix bug in example of reconst_csd.py

RicciWoo:
This is to fix the error of example reconst_csd.py.
1. transform image data from 1D to 4D before calling actor.odf_slicer();
2. remove unexpected argument ‘scale' when calling actor.peak_slicer().


doc/examples/reconst_csd.py
```diff
@@ -225,7 +227,7 @@
                              parallel=True)
 
 window.clear(ren)
-fodf_peaks = actor.peak_slicer(csd_peaks.peak_dirs, csd_peaks.peak_values, scale=1.3)
+fodf_peaks = actor.peak_slicer(csd_peaks.peak_dirs, csd_peaks.peak_values)
```
skoudoro: Why are you changing the scale here?

doc/examples/valid_examples.txt
```diff
@@ -1,62 +1,62 @@
- quick_start.py
- tracking_quick_start.py
- brain_extraction_dwi.py
- reconst_csa_parallel.py
- reconst_csa.py
- reconst_csd_parallel.py
+# quick_start.py
```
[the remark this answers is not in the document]
RicciWoo: file removed.

RicciWoo:
Do I need to remove the change in dti.py, because it's already in PR #1468

doc/examples/reconst_csd.py
RicciWoo: The same reason apply to PR #1479 in example sfm_reconst.py.

Issue #1469, opened 2018-03-20
Errors in generating documents (.rst) of examples


RicciWoo:
When I used "make clean", then "make html" to generate .rst documents of examples,
I encountered errors for several examples:
- sfm_tracking.py
- streamline_length.py
- linear_fascicle_evaluation.py
- segment_extending_clustering_framework.py
- segment_clustering_features.py
- segment_clustering_metrics.py
- bundle_registration.py

===========================================
In details for each one:
===========================================
- sfm_tracking.py:
Traceback (most recent call last):
  File "../../tools/make_examples.py", line 160, in <module>
    run_script()
  File "../../tools/make_examples.py", line 144, in run_script
    exec(open(script).read(), namespace)
  File "<string>", line 135, in <module>
  File "~/dipy/dipy/viz/actor.py", line 448, in streamtube
    poly_mapper.GlobalImmediateModeRenderingOn()
AttributeError: 'vtkRenderingOpenGL2Python.vtkOpenGLPolyDataMapper' object has no attribute 'GlobalImmediateModeRenderingOn'
===========================================
- streamline_length.py:
Traceback (most recent call last):
  File "../../tools/make_examples.py", line 160, in <module>
    run_script()
  File "../../tools/make_examples.py", line 144, in run_script
    exec(open(script).read(), namespace)
  File "<string>", line 121, in <module>
  File "~/dipy/dipy/viz/actor.py", line 448, in streamtube
    poly_mapper.GlobalImmediateModeRenderingOn()
AttributeError: 'vtkRenderingOpenGL2Python.vtkOpenGLPolyDataMapper' object has no attribute 'GlobalImmediateModeRenderingOn'
===========================================
- linear_fascicle_evaluation.py:
Traceback (most recent call last):
  File "../../tools/make_examples.py", line 160, in <module>
    run_script()
  File "../../tools/make_examples.py", line 144, in run_script
    exec(open(script).read(), namespace)
  File "<string>", line 28, in <module>
  File "~/dipy/doc/examples_built/streamline_tools.py", line 240, in <module>
    save_trk("lr-superiorfrontal.trk", lr_sf_trk, shape=shape, vox_size=voxel_size)
TypeError: save_trk() missing 1 required positional argument: 'affine'
===========================================
- segment_extending_clustering_framework.py:
Traceback (most recent call last):
  File "../../tools/make_examples.py", line 160, in <module>
    run_script()
  File "../../tools/make_examples.py", line 144, in run_script
[...]

PR #1478, opened 2018-03-25
DOC: Add comment about package CVXPY in example of reconst_mapmri.py

RicciWoo:
This is to solve the second problem in issue #1469: add comment about package CVXPY in example of reconst_mapmri.py.
Only modification in files reconst_mapmri.py & links_names.inc are necessary.


arokem:
I believe this is already handled in https://github.com/nipy/dipy/pull/1424

RicciWoo:
Yes, thank you for mention that, I will close this PR. ^-^

PR #1480, opened 2018-03-26
BF: Fix bug in example of streamline_tool.py

RicciWoo:
This is to fix error in example of streamline_tool.py:
Add necessary argument 'affine' when calling save_trk().


MarcCote:
IMHO it's better to rebase on top of master whenever it is possible compared to merging master back into your branch. I would expect a single commit for a change to a single line of code.

PR #1481, opened 2018-03-26
BF: Fix bug of no attribute 'GlobalImmediateModeRenderingOn' in actor.py

RicciWoo:
It seems like in VTK 8, there is no long attribute 'GlobalImmediateModeRenderingOn' in vtkMapper.
This PR is just to remove it, that is comment out line 448 in actor.py:
    #poly_mapper.GlobalImmediateModeRenderingOn()
This solves the problem of no attribute 'GlobalImmediateModeRenderingOn' in issue [#1469](https://github.com/nipy/dipy/issues/1469).


dipy/viz/actor.py
```diff
@@ -445,7 +445,7 @@ def streamtube(lines, colors=None, opacity=1, linewidth=0.1, tube_sides=9,
     poly_mapper.ScalarVisibilityOn()
     poly_mapper.SetScalarModeToUsePointFieldData()
     poly_mapper.SelectColorArray("Colors")
-    poly_mapper.GlobalImmediateModeRenderingOn()
+    #poly_mapper.GlobalImmediateModeRenderingOn()
```
skoudoro: Hi @RicciWoo, It does not really fix the problem. It would be better to add a vtk version test here. Can you do this change? Thanks

RicciWoo: I would like to try, but I don't know how to add a test, can you give me some hint?

dmreagan:
Replacing with #1493

PR #1484, opened 2018-03-26
DOC: Use the correct punctuation marks for `et al.`.

jhlegarreta:
Change `et. al` for `et al.`.


jhlegarreta:
The reported error
```
FAIL: dipy.reconst.tests.test_dki_micro.test_single_fiber_model
Arrays are not almost equal to 7 decimals
```
for `Python 3.4` is unrelated to the changes introduced in this topic. Only documentation has been changed. @arokem @skoudoro I'd dare to say that this is ready to be merged.

arokem:
Yep. Let's see if it persists when I merge this on master. We might have to deal with this on a separate issue.

PR #1485, opened 2018-03-26
BF: Fix bug in example of segment_quickbundles.py

RicciWoo:
It looks like the last modification of segment_quickbundles.py makes this error:

    File "dipy/segment/clustering_algorithms.pyx", line 108, in 
    dipy.segment.clustering_algorithms.quickbundles
    ValueError: setting an array element with a sequence.

so I changed back to where it was. Then it turns the problem of VTK 8 no attribute:

    "~/dipy/dipy/viz/actor.py", line 448, in streamtube
    poly_mapper.GlobalImmediateModeRenderingOn()
    AttributeError: 'vtkRenderingOpenGL2Python.vtkOpenGLPolyDataMapper' object 
    has no attribute 'GlobalImmediateModeRenderingOn'

This is discussed in issue #1490 .


doc/examples/segment_quickbundles.py
```diff
@@ -30,7 +30,7 @@
 
 streams, hdr = tv.read(fname)
```
MarcCote: We should be using `dipy.io.streamline.load_trk` instead of `tv.read`. The best part is that `load_trk` returns a `Streamlines` object and the header. That means, there is no need for line 33 at all :-).

MarcCote:
That's a lot of commits for a single line change :P. It would be better to branch anew from master, make you change and overwrite (force push) the branch associated to this PR, i.e. fix-bug-segment_quickbundles.

Also, check my comment on how to fix the example.

doc/examples/segment_quickbundles.py
RicciWoo: @skoudoro mentioned that if you use dipy.io.streamline.load_trk to read the dataset of 'fornix', will cause some problem. This may need some further effort beyond my knowledge right now. So I just leave this problem there.

RicciWoo:
Thanks for your comment. I'm new for github, but I learnt a lot recently on this project. ^-^

doc/examples/segment_quickbundles.py
MarcCote: @skoudoro what's the problem exactly?

MarcCote:
@RicciWoo I see that you opened a new PR (and closed this one). What you could have done is "force push" your new branch into this PR using `git push --force-with-lease origin local_branch:remote_branch`. In your case, here's what it would look like.

Inside your local repository (your machine):
```
git push --force-with-lease origin bf-ex-segment_quickbundle:fix-bug-segment_quickbundles
```
That line will push and overwrite (`--force-with-lease`) the content of the **local** branch `bf-ex-segment_quickbundle` into the **remote** branch `fix-bug-segment_quickbundles` of the **remote** repository named `origin` (assuming `origin` points to your Github's dipy repo).

doc/examples/segment_quickbundles.py
skoudoro: `fornix.trk` has an old header that misses some information @MarcCote. I do not remember which information exactly is missing but it crashes when you try to load it with `dipy.io.streamline.load_trk`. (more details, tomorrow :-) )

 We just need to update this data. I plan to do it on a new PR.

MarcCote: Ok, I see. It seems the header of the fornix file was zeroed out.

PR #1486, opened 2018-03-27
Parichit dipy flows

parichit:
Created a basic DIPY workflow to probe the status of the files in the system.  Tested the behavior on the local system to identify bugs and performance.

PR #1487, opened 2018-03-27
Commit for updated check_scratch.py script.

parichit:
1) Not the ascript accpets a directory path and list all the contents in that specified directory.
2) Don't understand why the script still throws errors when the condition fails, it should
just execute the else block.

### DIFF
--- a/dipy/workflows/check_scratch.py
+++ b/dipy/workflows/check_scratch.py
@@ -1,0 +1,52 @@
+###################
+#This is the modified workflow script motivated from the example given on the following link.
+#http://nipy.org/dipy/examples_built/workflow_creation.html#example-workflow-creation
+#This script now accepts a folder path and lists all the files inside that folder.
+
+#Current Testing Status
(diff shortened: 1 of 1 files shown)


skoudoro:
same comment as #1486 
I do not think this workflow is necessary but I think you should update all the current workflows with your test. 
I recommend you to close this PR and create a new one with this update.
There is a lot of pep8 (comma, space, ...) issue in your code so I recommend you to have a look at [python coding style](https://www.python.org/dev/peps/pep-0008/)
Do not forget to add some tests

parichit:
Hey Serge,

Thanks so much for responding. I will close this PR and apply to all other
workflows.

Only thing I don't understand is that my edotor is PEP8 compliant and it
does not show any errors with respect to code formatting.

But I will try couple of other editors and see if there are any issues in
the formatting.

Thanks so much :)

parichit:
The test in this workflow script is not versatile enough to handle all the input cases. 

Closing this PR and will open a new PR for different test cases.

PR #1494, opened 2018-04-11
BF: fix bug in example of reconst_csd.py

RicciWoo:
This is to fix the bugs in example of reconst_csd.py:

    File “~/dipy/dipy/viz/actor.py", line 717, in odf_slicer
        (szx, szy, szz) = odfs.shape[:3]
    ValueError: not enough values to unpack (expected 3, got 1)

and

    TypeError: peak_slicer() got an unexpected keyword argument ‘scale’

The modifications are:
1. change image data 'response_signal' from 1D to 4D before calling to 'actor.odf_slicer()';
2. reduce the parameter 'scale' to 'actor.odf_slicer()' because it's too large in 'csd_odfs.png';
3. remove unexpected argument ‘scale' when calling 'actor.peak_slicer()'.

### DIFF
--- a/doc/examples/reconst_csd.py
+++ b/doc/examples/reconst_csd.py
@@ -151,6 +151,8 @@ like  a pancake:
 """
 
 response_signal = response.on_sphere(sphere)
+# transform our data from 1D to 4D
+response_signal = response_signal[None, None, None, :]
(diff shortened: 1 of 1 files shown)


skoudoro:
Thanks @RicciWoo

PR #1495, opened 2018-04-12
BF: fix bug in example of sfm_reconst.py

RicciWoo:
This is to fix the bug in example of sfm_reconst.py:

    TypeError: peak_slicer() got an unexpected keyword argument ‘scale’

The modifications are:
1. reduce the parameter 'scale' to 'actor.odf_slicer()' because it's too large in 'sf_both.png';
2. remove unexpected argument ‘scale' when calling 'actor.peak_slicer()'.


skoudoro:
Thanks @RicciWoo

PR #1474, opened 2018-03-23
DOC: Fix typo on website & examples

RicciWoo:
1. change Dipy & dipy to DIPY on documents
2. change some acronyms to be all capitals
3. fix some typos and fake hyperlinks
4. fix some makeup styles and indentations

PR #1496, opened 2018-04-12
BF: fix bug in example of streamline_tools.py

RicciWoo:
This is to fix the bug in example of streamline_tools.py:

>  TypeError: save_trk() missing 1 required positional argument: 'affine'

The modification is just adding the argument 'affine=affine' to 'save_trk()'.


skoudoro:
LGTM too, waiting for #1474

skoudoro:
Thanks @RicciWoo

PR #1493, opened 2018-04-05
Fix surface rendering bugs in VTK8

dmreagan:
Update some defaults and remove a deprecated feature for the new OpenGL2 rendering backend in VTK7+. 

Also removed an unused `light` parameter that is missing from the doc string.

This PR fixes #1490 and replaces #1481.

PR #1498, opened 2018-04-12
BF: fix bug in example of segment_quickbundles.py

RicciWoo:
The last modification of segment_quickbundles.py makes this error:

    File "dipy/segment/clustering_algorithms.pyx", line 108, in 
    dipy.segment.clustering_algorithms.quickbundles
    ValueError: setting an array element with a sequence.

So to fix the bug, I just changed back to where it was. 
Then it turns to the attribute error of VTK 8.1:

    "~/dipy/dipy/viz/actor.py", line 448, in streamtube
    poly_mapper.GlobalImmediateModeRenderingOn()
    AttributeError: 'vtkRenderingOpenGL2Python.vtkOpenGLPolyDataMapper' object 
    has no attribute 'GlobalImmediateModeRenderingOn'

And I checked that the PR #1493 solves this problem. The resulting pictures look good.

### DIFF
--- a/doc/examples/segment_quickbundles.py
+++ b/doc/examples/segment_quickbundles.py
@@ -30,7 +30,7 @@ Load fornix streamlines.
 
 streams, hdr = tv.read(fname)
 
-streamlines = Streamlines(streams)
+streamlines = [i[0] for i in streams]
 
 """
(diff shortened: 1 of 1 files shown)


skoudoro:
For the moment, I will merge that one but we need to update 'fornix' dataset. Thank you @RicciWoo

PR #1503, opened 2018-04-30
fix plotting issue on Mapmri example

skoudoro:
This follows PR #1424

here the correction made:
- remove range of plot
- rotate odf_actor


Garyfallidis:
This is a change in an example and travis cannot catch such a change. Need to merge this to move forward with the release.

PR #1504, opened 2018-04-30
Fix test_whole_brain_slr precision

skoudoro:
fix tests precision for test_whole_brain_slr.
This needs to be improved in a close future


Garyfallidis:
Sound good. Thanks. Just to say here that this test had some issue on Macosx taking a bit longer time than in other PRs.

PR #1508, opened 2018-05-03
Adds whitespace, to appease the sphinx.

arokem:
This should at least partially address #1507

PR #1509, opened 2018-05-03
DOC: fix math rendering for some dki functions

matthew-brett:
See: https://github.com/nipy/dipy/issues/1507


arokem:
Jinx!

arokem:
See #1508

PR #1514, opened 2018-05-07
Adding pep8speak config file

skoudoro:
For checking pep8 issues automatically, I think we should try to use pep8speaks bots.

To get more information about this service, [Here the link](https://github.com/OrkoHunter/pep8speaks#private-repos).

I would like your opinion @arokem, @Garyfallidis.


arokem:
+1 from me!

PR #1520, opened 2018-05-14
Updating the documentation for the workflow creation tutorial.

parichit:
1) This change will be reflected on the page http://nipy.org/dipy/examples_built/workflow_creation.html#example-workflow-creation
2) Added one more import statement to import the AppendTextFlow method from the workflow class.
3) Added few lines to explain why we need to import the methods.

Added the desired import statement into the workflow_creation.py file.

The original tutorial mentions that the lines need to be added in a separate file in the "bin" directory but did not mention that this will also require importing the AppendTextFlow method from the newly created class.

Tested the file in ```bin``` directory after adding the import statement and the script executes perfectly.

### DIFF
--- a/doc/examples/workflow_creation.py
+++ b/doc/examples/workflow_creation.py
@@ -13,7 +13,7 @@ line::
 """
 
 """
-First create your workflow. Usually this would be in its own python file in
+First create your workflow (let's name this workflow file as my_workflow.py). Usually this is a python file in
 the ``<../dipy/workflows>`` directory.
 """
 
@@ -98,13 +98,12 @@ executable file located in ``bin``.
 
 """
 
-These lines will import the required files and classes into the python file. The
-first line imports the run_flow method from the flow_runner class and the second
-line imports the AppendTextFlow method from the newly created workflow class.
+The first line imports the run_flow method from the flow_runner class and the second
+line imports the AppendTextFlow class from the newly created my_workflow.py file.
 """
 
 from dipy.workflows.flow_runner import run_flow
-from dipy.workflows.AppendTextFlow import AppendTextFlow
+from dipy.workflows.my_workflow import AppendTextFlow
 """
 This is the method that will wrap everything that is needed to make a flow
 command line ready then run it.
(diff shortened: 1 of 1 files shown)


parichit:
Edited the workflow_creation.py file to resolve the PEP8 warnings about the trailing whitespaces.

However, doing so has created another warning about the maximum width of the line. It is more than 79 characters now. I noticed that several lines are more than 79 characters in width in the workflow_creation.py file so I kept my changes that way to solve the trailing whitespaces warning.

doc/examples/workflow_creation.py
```diff
@@ -96,7 +96,15 @@ def run(self, input_files, text_to_append='dipy', out_dir='',
 executable file located in ``bin``.
 """
 
+"""
+
+These lines will import the required files and classes into the python file. The
+first line imports the run_flow method from the flow_runner class and the second
+line imports the AppendTextFlow method from the newly created workflow class.
```
skoudoro: - I think the first sentence is not necessary (`These lines... python files.`)
- `AppendTextFlow method` -> `AppendTextFlow class`

doc/examples/workflow_creation.py
```diff
@@ -96,7 +96,15 @@ def run(self, input_files, text_to_append='dipy', out_dir='',
...
+"""
+
 from dipy.workflows.flow_runner import run_flow
+from dipy.workflows.AppendTextFlow import AppendTextFlow
```
skoudoro: Can you precise above that we create a file named `my_workflow.py` in `dipy/workflows` folder ?
After that, Can you change `from dipy.workflows.AppendTextFlow import AppendTextFlow` to `from dipy.workflows.my_workflow import AppendTextFlow` ?

parichit: Hey Serge, thanks for going through the changes.

Made the change to 

from dipy.workflows.my_workflow import AppendTextFlow

doc/examples/workflow_creation.py
parichit: Hey, thanks.

I removed the first line and changed the comments to explain more clearly.

PR #1522, opened 2018-05-15
Image element in dipy.viz.ui

karandeepSJ:
Implemented a new UI element that is capable of displaying an image when provided with the image path.
Tests written are passing.
Currently supports JPEG/JPG and PNG images.

### Features to be added
- Dragging and Resizing by mouse
- Browse file system to select and change image


ranveeraggarwal:
We have some pending changes in #1492 which would change the base UI class. Therefore, we'll need some changes to this PR when #1492 is merged. 
I am currently reviewing it - we should be able to merge it by this week. 
cc: @skoudoro @Garyfallidis

ranveeraggarwal:
That's a good idea. Once this is merged, we can look at modifying the Button2D element to use this.

dipy/viz/ui.py
```diff
@@ -2618,3 +2619,186 @@ def set_center(self, position):
...
+        tc.InsertComponent(1, 0, 1.0)
+        tc.InsertComponent(1, 1, 0.0)
+        tc.InsertComponent(2, 0, 1.0)
+        tc.InsertComponent(2, 1, 1.0)
+        tc.InsertComponent(3, 0, 0.0)
+        tc.InsertComponent(3, 1, 1.0)
+        self.texture_polydata.GetPointData().SetTCoords(tc)
+
+        texture_mapper = vtk.vtkPolyDataMapper2D()
+        if major_version <= 5:
```
nilgoyette: I know it's not your fault, but `major_version` is a terrible name. The major version of what? @skoudoro do you have an opinion on this? Should we rename this? And maybe create a function with the patterns

    if version SetInput else SetInputData
because it's used 4 times only in this file. We might also wonder if there's these still some people using vtk5. It's been 5 years since vtk6.

ranveeraggarwal:
@karandeepSJ now that Marc's PR is merged, can you refactor this code to go with the new base UI element? And #1534 too.

dipy/viz/ui.py
```diff
@@ -2161,3 +2163,185 @@ def handle_move_callback(self, i_ren, obj, slider):
         self.move_handle(click_position=click_position)
         i_ren.force_render()
         i_ren.event.abort()  # Stop propagating the event.
+
+
+class ImageHolder(UI):
```
ranveeraggarwal: `ImageHolder2D` probably be a better name.

dipy/viz/ui.py
```diff
@@ -2161,3 +2163,185 @@ def handle_move_callback(self, i_ren, obj, slider):
...
+    - png and jpg/jpeg images
+
+    Attributes
+    ----------
+    size: (float, float)
+        Image size (width, height) in pixels.
+
+    """
+
+    def __init__(self, imgPath, position=(0, 0), size=(100, 100)):
```
ranveeraggarwal: Parameter names are snake_cased. Replace `imgPath` with `img_path`

dipy/viz/ui.py
```diff
@@ -2161,3 +2163,185 @@ def handle_move_callback(self, i_ren, obj, slider):
...
+        ----------
+        imgPath : string
+            Path of the image
+        position : (float, float), optional
+            Absolute coordinates (x, y) of the lower-left corner of the image.
+        size : (int, int), optional
+            Width and height in pixels of the image.
+        """
+        super(ImageHolder, self).__init__(position)
+        self.img = self._build_image(imgPath)
```
ranveeraggarwal: Add a description of this attribute to the docstring.

dipy/viz/ui.py
ranveeraggarwal: +1 @karandeepSJ please look into this.

dipy/viz/ui.py
```diff
@@ -2161,3 +2163,185 @@ def handle_move_callback(self, i_ren, obj, slider):
...
+    def _add_to_renderer(self, ren):
+        """ Add all subcomponents or VTK props that compose this UI component.
+
+        Parameters
+        ----------
+        ren : renderer
+        """
+        ren.add(self.actor)
+
+    def resize(self, size):
```
ranveeraggarwal: I think this could be refactored to just use size as a property. Let me know what you think @karandeepSJ

dipy/viz/ui.py
```diff
@@ -2161,3 +2163,185 @@ def handle_move_callback(self, i_ren, obj, slider):
...
+
+    def set_img(self, img):
+        """ Modifies the image used by the vtkTexturedActor2D.
+
+        Parameters
+        ----------
+        img : imageDataGeometryFilter
+
+        """
+        if major_version <= 5:
```
ranveeraggarwal: Same as above, use something like `vtk_major_version` instead of `major_version`

karandeepSJ: Should I change this for all classes?

dipy/viz/ui.py
karandeepSJ: Size is not an attribute in the UI class or in any other classes. It would be best to maintain that uniformity and keep it as `resize()`

dipy/viz/ui.py
ranveeraggarwal: Sure, that would be great

dipy/viz/ui.py
ranveeraggarwal: You're correct. The way I thought of this is that size is an important attribute for some elements. 
For instance, in the TextBlock/TextBox, text is a major attribute and therefore we make that a property. Why not do the same for size? 
Maybe we can keep it the way it is for now and later refactor all elements if it makes sense.

dipy/viz/ui.py
dmreagan: This looks like a good use for the viz utility function [set_input](https://github.com/nipy/dipy/blob/4484514fb039bc212bc4e503829195c02e35ab95/dipy/viz/utils.py#L17)

dipy/viz/ui.py
karandeepSJ: I'll do it in a separate PR using the `set_input` function from utils

dipy/viz/ui.py
karandeepSJ: I'll create a separate PR for this.

dipy/viz/ui.py
dmreagan: I don't love "holder." Maybe `ImageContainer2D`?

dipy/viz/ui.py
```diff
@@ -2161,3 +2163,185 @@ def handle_move_callback(self, i_ren, obj, slider):
         self.move_handle(click_position=click_position)
         i_ren.force_render()
         i_ren.event.abort()  # Stop propagating the event.
+
+
+class ImageHolder(UI):
+    """ A 2D container to hold an image and is of type vtkTexturedActor2D.
```
dmreagan: I don't think it's necessary to say what type of actor is used. Just leave it at "A 2D container to hold an image."

dmreagan:
For a future PR: it would be nice to have something like CSS's [object-fit](https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit) to handle mismatches between the image size and the container size.

PR #1524, opened 2018-05-16
Values from streamlines object

arokem:
Small bug fix, that accommodates use of the new nibabel Streamlines API with this function.

The first commit demonstrates the bug with a test.

The second commit is a one-line change that fixes it.

### DIFF
--- a/dipy/tracking/streamline.py
+++ b/dipy/tracking/streamline.py
@@ -29,7 +29,6 @@ else:
 
     MEGABYTE = 1024 * 1024
 
-
     class _BuildCache(object):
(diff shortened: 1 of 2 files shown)


skoudoro:
great! I saw you made a part of this change on #1515. 

After making pep8speak happy, I think it is ready to go.

arokem:
Be careful what you ask for :-)

I fixed PEP8 for this entire file.

skoudoro:
😄 hahaha, Thanks for the advise, I will try to be careful and not listen too much pep8speak 😄

PR #1528, opened 2018-05-20
Show spheres with different radii, colors and opacities + add timers + add exit a + resolve issue with imread

Garyfallidis:
- [x] Nice and simple to use function for visualizing many spheres with different radii and colors. RGB and RGBA colors are allowed.  A (opacity) component can be different for each sphere.

- [x] Adding  timers in ShowManager for building animations. 

- [x] Allowing exiting (closing) a window from inside ShowManager.

- [x] Resolving issue with imread as it is not available anymore in scipy.misc.


dipy/viz/actor.py
```diff
@@ -1337,6 +1339,79 @@ def point(points, colors, opacity=1., point_radius=0.1, theta=8, phi=8):
...
+
+    polyData = vtk.vtkPolyData()
+    polyData.SetPoints(pts)
+    polyData.GetPointData().AddArray(radii_fa)
+    polyData.GetPointData().SetActiveScalars('rad')
+    polyData.GetPointData().AddArray(cols)
+
+    glyph = vtk.vtkGlyph3D()
+    glyph.SetSourceConnection(src.GetOutputPort())
+    if major_version <= 5:
```
ranveeraggarwal: As mentioned by @nilgoyette in #1522, we should decide on an alternate name for this variable. Like `vtk_major_version`.

ranveeraggarwal:
Looks good! A couple of points:

1. Let's merge the [changes](https://github.com/Garyfallidis/dipy/pull/56) made by @thechargedneutron before we merge this in.
2. It will be great if we can have an example for this in doc/examples.

dipy/viz/actor.py
Garyfallidis: Sure but this is a topic for a different PR and see here https://github.com/nipy/dipy/issues/1544 Support of old vtk versions should be removed all together. More of a problem than a solution.

Garyfallidis:
After travis passes @ranveeraggarwal, @karandeepSJ, @dmreagan and @MarcCote this should be ready to be merged.

doc/examples/viz_timers.py
```diff
@@ -0,0 +1,69 @@
...
+    if cnt > 100:
+        showm.exit()
+
+
+renderer.add(tb)
+
+# Run every 200 milliseconds
+showm.add_timer_callback(True, 200, timer_callback)
+
+showm.start()
```
MarcCote: This should be commented?

doc/examples/viz_timers.py
```diff
@@ -0,0 +1,69 @@
+"""
+===============
+Using a timer
```
MarcCote: Go this warning when running the example.

```
/home/macote/.local/venvs/dipy/lib/python3.6/site-packages/vtk/util/numpy_support.py:137: FutureWarning: Conversion of the second argument of issubdtype from `complex` to `np.complexfloating` is deprecated. In future, it will be treated as `np.complex128 == np.dtype(complex).type`.
  assert not numpy.issubdtype(z.dtype, complex), \
```

doc/examples/viz_timers.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+renderer.add(tb)
+
+# Run every 200 milliseconds
+showm.add_timer_callback(True, 200, timer_callback)
+
+showm.start()
+
+window.record(showm.ren, size=(900, 768), out_path="viz_timer.png")
```
MarcCote: The screenshot recorded says `Let's count up to 100 and exit: 101`. I would have expected to see 100 instead of 101.

dipy/viz/window.py
```diff
@@ -591,6 +586,28 @@ def add_window_callback(self, win_callback):
         self.window.AddObserver(vtk.vtkCommand.ModifiedEvent, win_callback)
         self.window.Render()
 
+    def add_timer_callback(self, repeat, duration, timer_callback):
+        self.iren.AddObserver("TimerEvent", timer_callback)
```
MarcCote: Rather than using `timer_callback` directly, we could wrap it with our own method/function that way we can control what arguments will be provided to `timer_callback`.

For instance, at the moment `timer_callback` is a function that expects two arguments: a `vtkRenderWindowInteractor` and a `str` (the event's name). I doubt common users would use them but they are still forced to provide a callback function that accepts those two arguments. What I propose is the following:

```python
def _wrap_timer_callback(self, timer_callback):
    def _wrapped_timer_callback(iren, event):
        timer_callback(self, ...)

   return _wrapped_timer_callback
```
then replace line 590 with
```python
self.iren.AddObserver("TimerEvent", self._wrap_timer_callback(timer_callback))
```
Note `...` means we could provide any Python object we deem relevant in addition to the `ShowManager` instance (i.e. `self`).

doc/examples/viz_timers.py
Garyfallidis: lol okay sure. Will fix that. Sharp eyes @MarcCote!

doc/examples/viz_timers.py
Garyfallidis: Nope, start does not need to be commented @MarcCote :) That is the great thing here we can close the application whenever we want from inside the timer callback. Perfect for testing viz, building animations etc. This is a really nice new feature. Don't you agree? :)

doc/examples/viz_timers.py
Garyfallidis: Interesting. As far as I know we do not use complex numbers

Garyfallidis: @skoudoro would be great if we can track down these FutureWarnings appearing when running this example.

dipy/viz/window.py
Garyfallidis: Okay, this is an important point. But does not make much sense changing that in this PR. We need to make decisions about callbacks in a developers meeting. I know that each one of us @karandeepSJ, @dmreagan, @ranveeraggarwal and myself have different suggestions of what will be the optimal strategy with callbacks.  My plan is to have that discussion as soon as possible. Possibly after the first split of viz modules. Will send a doodle about that. But for now I suggest moving ahead with this PR. Let me know if you disagree.

doc/examples/viz_timers.py
MarcCote: Oh nice! :relaxed:

doc/examples/viz_timers.py
MarcCote: ;)

Garyfallidis:
@skoudoro this is ready to be merged. Please do merge it. Also, check why codecov/patch reports coverage on tests. That looks incorrect.

skoudoro:
After some tests, LGTM too. merging

PR #1530, opened 2018-05-21
Wrong default value for parameter 'symmetric' connectivity_matrix function

albayenes:
Hi,
There were a small mistake in documentation of connectivity_matrix function parameter symmetric. Its default value is True but it was given as False. This PR fix it.


arokem:
Good catch!

PR #1533, opened 2018-05-24
Show message if number of arguments mismatch between the doc string and the run method.

parichit:
A simple check is implemented in the base.py script to display a message if the number of arguments in the doc string and the run method mismatches.

### DIFF
--- a/dipy/workflows/base.py
+++ b/dipy/workflows/base.py
@@ -112,20 +112,21 @@ class IntrospectiveArgumentParser(arg.ArgumentParser):
         output_args = \
             self.add_argument_group('output arguments(optional)')
 
+        len_args = len(args)
+        len_defaults = len(defaults)
+
         #  This check simply shows a helpful message to the user if there
         #  is a mismatch in the number of arguments in the run method and
         #  the doc string. Doc string refers to the parameter help written
         # in the workflow python script.
 
-        if len(args) != len(self.doc):
+        if len_args != len(self.doc):
             print(self.prog+": Number of parameters in the doc string and "
                             "run method does not match. Please ensure that"
                             " the number of parameters in the run method is"
                             " the same as the doc string.")
             exit(1)
 
-        len_args = len(args)
-        len_defaults = len(defaults)
 
         for i, arg in enumerate(args):
             prefix = ''
--- a/dipy/workflows/base.py
+++ b/dipy/workflows/base.py
@@ -109,9 +109,6 @@ class IntrospectiveArgumentParser(arg.ArgumentParser):
 
         args, defaults = get_args_default(workflow.run)
 
-        len_args = len(args)
-        len_defaults = len(defaults)
-
         output_args = \
             self.add_argument_group('output arguments(optional)')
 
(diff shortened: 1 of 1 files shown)


parichit:
Closing this pull request due to the usage of exit() statement. 
Will create a new pull request by replacing the exit with an error

skoudoro:
you do not need to close this PR @parichit. You just have to commit on this branch and it will be ok

parichit:
@skoudoro : Thanks for the update. I have committed to the branch and re-opened the PR.

dipy/workflows/base.py
```diff
@@ -101,19 +101,33 @@ def add_workflow(self, workflow):
             ref_text = [text if text else "\n" for text in npds['References']]
             ref_idx = self.epilog.find('References: \n') + len('References: \n')
             self.epilog = "{0}{1}\n{2}".format(self.epilog[:ref_idx],
-                                               ''.join([text for text in ref_text]),
+                                        ''.join([text for text in ref_text]),
```
nilgoyette: Can't you write `''.join(ref_text),`?

dipy/workflows/base.py
```diff
@@ -101,19 +101,33 @@ def add_workflow(self, workflow):
...
                                                self.epilog[ref_idx:])
 
         self.outputs = [param for param in npds['Parameters'] if
                         'out_' in param[0]]
 
         args, defaults = get_args_default(workflow.run)
 
+        output_args = \
+            self.add_argument_group('output arguments(optional)')
```
nilgoyette: This fits on one line (col 76)

dipy/workflows/base.py
```diff
@@ -101,19 +101,33 @@ def add_workflow(self, workflow):
...
+        #  the doc string. Doc string refers to the parameter help written
+        # in the workflow python script.
+
+        if len_args != len(self.doc):
+            raise ValueError(
+                            self.prog +
+                            ": Number of parameters in the "
+                            "doc string and run method does not match. "
+                            "Please ensure that the number of parameters "
+                            "in the run method is same as the doc string.")
```
nilgoyette: Comments above and this text agree, which is good, but what's the point of the comment then? We usually add comments to explain something more clearly. IMO, it's not required here.

Also, your indentation is weird here.

parichit: Hey,

Thanks for the comments. I will discuss this with the team and remove the comments. So the reason for putting the comment is to ensure some space in case the message needs to be changed in future (a new developer to DIPY, like me) will find the comment useful to understand the piece of code.

About the indentation, I made it strictly compliant to PEP8 since the dry run of PEP8 was complaining about the under or over indentation. So I made it strictly according to the PEP8 standards. 

I agree with you, the indentation is not good enough but since the change was in the core file of DIPY workflows so I chose to follow the PEP8 coding style to avoid PEP8 messages in the PR checks.

Thanks :)

dipy/workflows/base.py
parichit: Hi,
Thanks for the update. I have updated it to ''.join(ref_text).

dipy/workflows/base.py
parichit: Hey,
Thanks for the update. I updated the code to adjust in a single line.

dipy/workflows/base.py
parichit: Hey,

I updated the indentation.

Thanks.

skoudoro:
ok, LGTM, I will merge tomorrow if there is no other comment

PR #1534, opened 2018-05-25
Changed the icon set in Button2D from Dictionary to List of Tuples

karandeepSJ:
The `Button2D` class used a dictionary to store icons. The random ordering of dictionaries in Python was causing problems when a particular order was required for the button icons.
This PR changes the dictionary to a list of tuples of the form `((icon_name, icon), (icon_name, icon), ....)`


karandeepSJ:
There seems to be some issue with how I rebased. I'll make a new PR.

PR #1535, opened 2018-05-25
removed some unnecessary imports from sims example

ShreyasFadnavis:


.ipynb_checkpoints/Untitled-checkpoint.ipynb
```diff
@@ -0,0 +1,6 @@
+{
```
Garyfallidis: Hi @ShreyasFadnavis I believe you have accidentally added these notebooks here.

ShreyasFadnavis: Got it! Thanks..

PR #1539, opened 2018-05-29
Changing the default behavior of the workflows to create the output file(s) in the current working directory.

parichit:
The default value of the output strategy parameter is updated to 'absolute'. It was 'append' earlier. Now, with absolute, the output files are being created in the current working directory.

This behavior can be changed by passing other values for --out_strat parameter to create the output files in either,

A) Current working directory (default: 'absolute')
B) In the same path as the input files ('append')


parichit:
I am closing this pull request because there are few extra lines of code that got committed to this branch. 

I will be creating another pull request with only the desired changes.

PR #1542, opened 2018-05-30
fix for using cvxpy solver

Borda:
updating the optimization problems since the cvxpy does not support the strict inequalities


Borda:
well, there is still an issue with building required packages for certain configurations
"Failed building wheel for osqp"

Borda:
I have fixed the strict inequalities which are not supported in cvxpy

skoudoro:
Nice! Thanks @Borda ! merging

PR #1548, opened 2018-06-04
Fixed references per request of @garyfallidis.

kesshijordan:
@Garyfallidis: Here is the modified bootstrap tutorial with @MrBago dissertation added and citing the Berman 2008 paper with a functional link.


Garyfallidis:
thanks @kesshijordan !

PR #1554, opened 2018-06-07
Checking if the input file or directory exists when running a workflow

parichit:
1) Added a utility function 'file_existence_check' to
check if the input file or directory exists or not.

2) The workflows now report a friendly message to the user
if the input file or directory does not exist.

3) The error stack is still reported in addition to the message.

### DIFF
--- a/dipy/workflows/multi_io.py
+++ b/dipy/workflows/multi_io.py
@@ -197,7 +197,6 @@ def io_iterator_(frame, fnc, output_strategy='absolute', mix_names=False):
     return io_iterator(inputs, out_dir, outputs, output_strategy, mix_names,
                        out_keys=out_keys)
 
-
 class IOIterator(object):
     """ Create output filenames that work nicely with multiple input files from
     multiple directories (processing multiple subjects with one command)
@@ -214,6 +213,7 @@ class IOIterator(object):
 
 
     def set_inputs(self, *args):
+        self.file_existence_check(args)
         self.input_args = list(args)
         self.inputs = [sorted(glob(inp)) for inp in self.input_args if type(inp) == str]
 
@@ -253,3 +253,11 @@ class IOIterator(object):
         IO = np.concatenate([I, O], axis=1)
         for i_o in IO:
             yield i_o
+
+    @staticmethod
+    def file_existence_check(args):
+        input_args = list(args)
+        for input_path in sorted(input_args):
+            if len(glob(input_path)) == 0:
+                raise OSError('File not found: '+input_path)
+
--- a/dipy/workflows/multi_io.py
+++ b/dipy/workflows/multi_io.py
@@ -254,7 +254,7 @@ class IOIterator(object):
         for i_o in IO:
             yield i_o
 
-    def file_existence_check(args):
+    def file_existence_check(self, args):
         input_args = list(args)
         for input_path in sorted(input_args):
             if len(glob(input_path)) == 0:
--- a/dipy/workflows/multi_io.py
+++ b/dipy/workflows/multi_io.py
@@ -7,6 +7,7 @@ from glob import glob
 from dipy.utils.six import string_types
 from dipy.workflows.base import get_args_default
 
+
 def common_start(sa, sb):
     """ Returns the longest common substring from the beginning of sa and sb """
     def _iter():
@@ -83,7 +84,6 @@ def connect_output_paths(inputs, out_dir, out_files, output_strategy='absolute',
 
         elif output_strategy == 'append':
             dname = path.join(inp_dirname, out_dir)
-
 
         else:
             dname = out_dir
@@ -197,6 +197,7 @@ def io_iterator_(frame, fnc, output_strategy='absolute', mix_names=False):
     return io_iterator(inputs, out_dir, outputs, output_strategy, mix_names,
                        out_keys=out_keys)
 
+
 class IOIterator(object):
     """ Create output filenames that work nicely with multiple input files from
     multiple directories (processing multiple subjects with one command)
@@ -210,7 +211,6 @@ class IOIterator(object):
         self.mix_names = mix_names
         self.inputs = []
         self.out_keys = None
-
 
     def set_inputs(self, *args):
         self.file_existence_check(args)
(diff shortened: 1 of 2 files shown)


Garyfallidis:
@parichit you need to provide a test.

skoudoro:
@parichit, can you rebase too ?

parichit:
@Garyfallidis: Thank you for the suggestion, The function is covered by all the workflows (by default) because whenever an IOIterator is created, this function is being invoked by it that's why I have not created a test for this addition. 

@serge: Thank for the good advice (you are correct, i should rebase). I tried rebasing but it says that the branch is up-to-date. I am closing this pull request and will open a new request with a clean history because this branch has also picked up the history from older commits.

parichit:
@serge: Sorry about this but can you please check if this branch is okay now? I did rebased it after fetching the upstream. If this is not yet rebased then please tell me and I will create a new branch for this feature.

parichit:
@Serge and @Garyfallidis: Thanks for pointing the test case out. I will implement the test case for this PR and recommit.

parichit:
Added the test case for this PR.
Thanks :)

dipy/workflows/multi_io.py
```diff
@@ -253,3 +253,10 @@ def __iter__(self):
         IO = np.concatenate([I, O], axis=1)
         for i_o in IO:
             yield i_o
+
+    @staticmethod
```
skoudoro: Why you need it as a static method?  I think you can remove it

dipy/workflows/multi_io.py
```diff
@@ -253,3 +253,10 @@ def __iter__(self):
...
+    def file_existence_check(args):
+        input_args = list(args)
+        for input_path in sorted(input_args):
+            if len(glob(input_path)) == 0:
```
skoudoro: This condition works but it is not clear. Can you change it and use something more explicit like:
`if not os.path.isfile(input_path) or not os.path.isdir(input_path)`

dipy/workflows/multi_io.py
```diff
@@ -253,3 +253,10 @@ def __iter__(self):
...
+                raise OSError('File not found: '+input_path)
```
skoudoro: - Can you replace `OSError` by `IOError`?
- pep8: Can you add space around the operator `+`

dipy/workflows/multi_io.py
parichit: Thanks for the suggestion. I have removed the @staticmethod.

dipy/workflows/multi_io.py
parichit: Yes, this is a logical suggestion. Thank you for pointing this out. I have replaced the OSError by IOError. 

I used OSError because the file/directory related errors/exceptions are placed inside the OSError on the python's exception page here https://docs.python.org/3/library/exceptions.html#os-exceptions.

I have changed the OSError to IOError. :)

dipy/workflows/multi_io.py
parichit: @serge thank you for checking out the code. So you are right the way I am doing it is not intuitive but below are  the reasons why I am doing it this way, So I am writing the code to do what you suggested,

        for input_path in sorted(input_args):
            if os.path.isdir(input_args):
                if not os.path.isdir(input_path):
                    err = True
            elif os.path.isfile(input_path):
                if not os.path.isfile(input_path):
                    err = True
            if err:
                raise IOError('File not found: ' + input_path)

You can see that O have to separately check for the file and a directory and accordingly decide what is missing, this increases the number of IF-ELSE conditions for this small check and since this check is performed for all the workflows so this will cause the slowdown when the workflows are executed multiple times.

So, what i did is to keep the default glob object and check if it's length is 0 so that there is just one IF condition to check.

Please tell if there is a better way to implement this check that I can use? otherwise, I will implement the above solution which is more intuitive but a bit slower.

Thanks.

parichit: I will wait for your suggestion and only commit the changes after that.

parichit: Hey serge, thanks for your suggestion, here is where I think this check can fail,

We know that currently, the workflows work for both absolute file/directory paths and also for wildcards, in case the user gives absolute file or directory names then this is fine but when the user supplies a wildcard for example, in the apply_affine workflow to provide all the input files at once,
For example:
dipy_apply_affine *input --affine=affine.txt

In this case, you can see the above check will always fail since this check will simply look for file *input but actually, it can be a file called input1.nii.gz or input2.nii.gz etc. and so it will always fail with the file not found.

I tested this behavior when I wrote this check and realized that a simple condition won't be enough. So, to keep the code to the minimum, I used the len(glob(input_path)) so that for both absolute file names and wildcard based file names the check can work because the glob will automatically expand the wildcard to the actual file names.

I realized this when I was starting to make the changes again based on your suggestion.

Let me know what you think about it?

skoudoro: Ok, I see, I totally forgot the wildcards. Thank you for all this explanation. I understand why you use `glob`.

No need to change, You can let it as it is.

skoudoro:
Hi @parichit, this PR is failing, can you fix [this error](https://travis-ci.org/nipy/dipy/jobs/419825209#L3751) please?  I think, you are just missing a `*` on `args`

parichit:
hey, @skoudoro, I have added the '**_self_**' keyword in the file_existense_check() function, the error was occurring because the number of arguments was mismatching. I have tested the code manually on the terminal, the relevant test case for this PR appears in the test_workflow.py file. Please let me know if you think we need more changes.

dipy/workflows/multi_io.py
```diff
@@ -253,3 +253,9 @@ def __iter__(self):
         IO = np.concatenate([I, O], axis=1)
         for i_o in IO:
             yield i_o
+
+    def file_existence_check(self, args):
+        input_args = list(args)
```
skoudoro: Thank you for this update @parichit. 

As you can see [here](https://travis-ci.org/nipy/dipy/jobs/424453062#L3493), the tests are still failing. I think you can replace this line with something like:
`input_args =  [fname for fname in list(args) if isinstance(fname, str)]`

it allow to keep only string value and not integers. Indeed, we can have integers as positional argument sometimes

parichit: Thanks for this suggestion @skoudoro , I wasn't sure why the PR was failing even after I made the change. You are right, there were other arguments type than string. I changed the line to accept only strings. Waiting for the checks to complete.

skoudoro:
Thank you @parichit! merging

PR #1557, opened 2018-06-08
Range slider element for viz.ui

karandeepSJ:
This PR adds two elements to the viz.ui module.

LineDoubleSlider2D
-----------------
This element allows the user to have two handles on the same slider. These handles can slide on the track, while not crossing each other at any point.
This element is useful for setting a range for a parameter.
For example, in CT images, this can be used to select a mapping for some window of pixel values to values between 0-255.
double slider

RangeSlider
---------------
This element uses a `LineDoubleSlider2D` to select a range which restricts a `LineSlider2D` to move within that range. This can be useful when a user wants to fine tune the value of a parameter by narrowing the permissible range for that parameter.
rangeslider

Shape of Handles
-----------------
This PR also gives the user an option to choose the shape of the handles of the sliders. Currently, `disk` and `square` are supported. This option is available in `LineSlider2D`, `LineDoubleSlider2D` and `RangeSlider`.

### DIFF
--- a/dipy/viz/test.py
+++ b/dipy/viz/test.py
@@ -1,0 +1,6 @@
+import ui
+import window
+ls = ui.RangeSlider()
+sm=window.ShowManager(size=(600,600))
+sm.ren.add(ls)
+sm.start()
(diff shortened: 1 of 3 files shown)


Garyfallidis:
Okay this good too @karandeepSJ . 
Something minor rather than shape I do prefer handle_shape. Do change this in one of your upcoming PRs.

PR #1547, opened 2018-06-04
 Changed the icon set in Button2D from Dictionary to List of Tuples

karandeepSJ:
The `Button2D` class used a dictionary to store icons. The random ordering of dictionaries in Python was causing problems when a particular order was required for the button icons.
This PR changes the dictionary to a list of tuples of the form `((icon_name, icon), (icon_name, icon), ....)`

PR #1559, opened 2018-06-10
Checkbox and RadioButton elements for viz.ui

karandeepSJ:
This PR implements 3 classes.
Need to merge  #1547 first 

`Option`
---------
This is a  set of a Button2D and a TextBlock2D to act as a single option for check-boxes and radio buttons.
It aligns the button with the label using the number of `\n` in the label.

`Checkbox`
-------------
This class uses a list of objects of `Option` class to implement the checkbox element. It is a group of options where multiple options can be checked.

checkbox

`RadioButton`
-----------------
This class implements the radio button element where only a single option can be selected at a time.
It inherits from the `Checkbox` class with a different `toggle` function to uncheck all other options when one is clicked.

radiobutton


Garyfallidis:
@karandeepSJ please resolve the conflicts here.

dipy/viz/ui.py
```diff
@@ -2952,6 +2952,278 @@ def set_img(self, img):
...
+
+        Parameters
+        ----------
+        ren : renderer
+        """
+        self.button.add_to_renderer(ren)
+        self.text.add_to_renderer(ren)
+
+    def _get_size(self):
+        width = self.button.size[0] + 10 + self.text.size[0]
```
dmreagan: 10 is a magic number here and in line 3035. It should be pulled out into a variable so the two don't get out of sync.

dipy/viz/ui.py
```diff
@@ -2952,6 +2952,278 @@ def set_img(self, img):
...
+
+        Parameters
+        ----------
+        coords: (float, float)
+            Absolute pixel coordinates (x, y).
+        """
+        num_newlines = self.label.count('\n')
+        self.button.position = coords + \
+            (0, num_newlines * self.font_size * 0.5)
+        offset = (self.button.size[0] + 10, 0)
```
dmreagan: Other instance of magic number.

dipy/viz/ui.py
```diff
@@ -2952,6 +2952,278 @@ def set_img(self, img):
...
+        """ Setup this UI component.
+        """
+        self.options = []
+        button_y = self.position[1]
+        for option_no in range(self.num_options):
+            option = Option(label=self.labels[option_no],
+                            font_size=self.font_size,
+                            position=(self.position[0], button_y))
+            button_y = button_y + self.font_size * \
+                (self.labels[option_no].count('\n') + 1) * 1.2 + self.padding
```
dmreagan: Do something about magic number 1.2, which also appears on lines 2987 and 3149.

dipy/viz/ui.py
```diff
@@ -2952,6 +2952,278 @@ def set_img(self, img):
...
+        self._font_size = font_size
+        self.font_family = font_family
+        super(Checkbox, self).__init__(position)
+
+    def _setup(self):
+        """ Setup this UI component.
+        """
+        self.options = []
+        button_y = self.position[1]
+        for option_no in range(self.num_options):
```
dmreagan: Is `num_options` necessary in this class? It seems like you could write this loop as

```python
for label in self.labels
```

Otherwise, it is only used in `_get_size()`

skoudoro:
Hi @karandeepSJ, 

As requested before, can you rebase this PR and add an example. #1547 has been merged.

Thank you

dipy/viz/ui.py
```diff
@@ -2952,6 +2952,276 @@ def set_img(self, img):
...
+        """
+        event = []
+        button.next_icon()
+        for option in self.options:
+            if option.button == button:
+                option.checked = not option.checked
+            if option.checked is True:
+                event.append(option.label)
+        i_ren.force_render()
+        print(event)
```
dmreagan: Remove prints

dipy/viz/ui.py
```diff
@@ -2952,6 +2952,276 @@ def set_img(self, img):
...
+                    option.checked = True
+                    option.button.next_icon()
+                event = option.label
+
+            elif option.checked is True:
+                option.checked = False
+                option.button.next_icon()
+
+        i_ren.force_render()
+        print(event)
```
dmreagan: Remove prints

dmreagan:
Can we change this so the options are drawn starting from the top instead of the bottom? Right now, the first option is on the bottom and the last option is on the top.

dmreagan:
@MarcCote any thoughts on this before it goes in?

dipy/viz/ui.py
```diff
@@ -2952,6 +2952,283 @@ def set_img(self, img):
         self.texture = set_input(self.texture, img)
 
 
+class Option(UI):
```
MarcCote: I was expecting the Option class to have a `select()` and a `deselect()` methods which would take care of changing the option's appearance accordingly.

dipy/viz/ui.py
```diff
@@ -2952,6 +2952,283 @@ def set_img(self, img):
...
+        for label in self.labels:
+            option = Option(label=label,
+                            font_size=self.font_size,
+                            position=(self.position[0], button_y))
+            line_spacing = option.text.actor.GetTextProperty().GetLineSpacing()
+            button_y = button_y + self.font_size * \
+                (label.count('\n') + 1) * (line_spacing + 0.1) + self.padding
+            option.button.on_left_mouse_button_pressed = self.toggle_check
+            self.options.append(option)
+            option.button.add_callback(option.text.actor,
```
MarcCote: I think we can all agree on the way we are handling events is unintuitive (and probably bogus) right? @Garyfallidis @dmreagan @ranveeraggarwal 
In this particular instance, we are adding a callback (which is a method of a **CheckBox** object) to the VTKActor of a **TextBlock2D** object using the add_callback of a **Button2D** object!

I think the proper way would have been something like:
`option.on_change = self._handle_new_selection`.

karandeepSJ: I did this to avoid looping through all the options in the callback. This sends the button of the text clicked as an argument to the `add_callback()` function, so that its icon can be changed.

MarcCote: I see. Wouldn't it be better if the `Option` object was handling that for you? To me, whether the `TextBlock2D` is associated with a particular `Button2D` shouldn't influence the behavior of `CheckBox` and `RadioBox` classes.

dipy/viz/tests/test_ui.py
```diff
@@ -646,6 +803,15 @@ def test_ui_image_container_2d(interactive=False):
...
         test_ui_range_slider(interactive=False)
 
+    if len(sys.argv) <= 1 or sys.argv[1] == "test_ui_option":
+        test_ui_option(interactive=False)
+
+    if len(sys.argv) <= 1 or sys.argv[1] == "test_ui_checkbox":
+        test_ui_checkbox(interactive=False)
+
+    if len(sys.argv) <= 1 or sys.argv[1] == "test_ui_radio_button":
+        test_ui_radio_button(interactive=False)
```
MarcCote: It should be `interactive=True` though. When the script is called directly, interactive should be true.

dipy/viz/ui.py
MarcCote: See https://github.com/karandeepSJ/dipy/pull/1 for a way of doing what I described above.

MarcCote:
@dmreagan feel free to merge this one whenever you want.

PR #1563, opened 2018-06-11
Replacing major_version in viz.ui

karandeepSJ:
This PR changes the variable `major_version` used to check the user's version of the vtk library.
When it need to be used to provide the input to the texture mapper, the `set_input` function from `viz.utils` has been used.
At other places, `major_version` has been replaced with `VTK_MAJOR_VERSION`.


Garyfallidis:
Cool, thanks. We are going to deprecate old vtk versions soon. But hey good to keep the standards right.

Issue #1565, opened 2018-06-14
New warnings with `PRE=1`


arokem:
See: https://travis-ci.org/nipy/dipy/jobs/391291194

We are getting a lot (!) of these: 

```
home/travis/build/nipy/dipy/venv/lib/python3.5/site-packages/scipy/fftpack/basic.py:160: FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.
  z[index] = x
```

This could be numpy 1.15, or a new version of scipy? 

Looks like maybe it's coming out of `imaffine`?

skoudoro:
Even if I find them strange because they work with a short example, I use the opportunity to add all errors we get on pre:

```
======================================================================
ERROR: dipy.align.tests.test_parzenhist.test_parzen_joint_histogram
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/koudoro/anaconda/envs/dipy-pre-env-3/lib/python3.6/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/Users/koudoro/Software/dipy/dipy/align/tests/test_parzenhist.py", line 136, in test_parzen_joint_histogram
    P.setup(static, moving, static_mask, moving_mask)
  File "dipy/align/parzenhist.pyx", line 105, in dipy.align.parzenhist.ParzenJointHistogram.setup
    self.smin = np.min(static[smask != 0])
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

======================================================================
ERROR: dipy.align.tests.test_parzenhist.test_parzen_densities
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/koudoro/anaconda/envs/dipy-pre-env-3/lib/python3.6/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/Users/koudoro/Software/dipy/dipy/align/tests/test_parzenhist.py", line 193, in test_parzen_densities
    parzen_hist.setup(static, moving)
  File "dipy/align/parzenhist.pyx", line 105, in dipy.align.parzenhist.ParzenJointHistogram.setup
    self.smin = np.min(static[smask != 0])
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

======================================================================
[...]

PR #1570, opened 2018-06-17
Added File Menu element to viz.ui

karandeepSJ:
Added a `FileMenu2D` that uses a `ListBox2D` to display all files and directories in the file system. Files are shown with Blue color and directories with Green. Clicking on a directory will take the user inside it.

filemenu


MarcCote:
I think this code could be more concise if it was not reimplementing scrolling and selection logic as it is already managed by the ListBox component.

Here my version (still WIP) that could use as a start point: https://github.com/nipy/dipy/compare/master...MarcCote:enh_ui_filedialog3

karandeepSJ:
It's not reimplementing them. There are additional tasks that need to be done on scrolling and clicks, so I added more callbacks. For scrolling, we need to update the colors of the text in the slots, so the `scroll_callback` is handling that.

MarcCote:
I see. Maybe it would be better (more modular) to have that logic in ListBox no? 

We could have a list of "representation" along the values for all items in the list. Then, whenever an item needs to be displayed, we update its slots according to its values (already doing this) and representation (need to be implemented). That way would offer a modular approach to use special representation for files/folders in the FileMenu.

karandeepSJ:
What could be the possible uses for a ListBox that require different representations?

MarcCote:
It could allow a user to define which colors should be used to highlight the different items in the listbox (like you are doing for FileMenu).

karandeepSJ:
Should I keep representation as a list of colors?

arokem:
Indeed, this fixes the error that was on master. The remaining error is due to #1565, so I will go ahead and merge this.

karandeepSJ:
Wrong PR :P. The one you want is #1574

arokem:
Yeah - sorry about that. See #1577 for the reversion of this. 

I think the right thing to do know is to merge #1577 and then reopen this PR. Again - so sorry about my confusion. I had two windows open simultaneously, and then acted in the wrong one.

skoudoro:
Revert is done, can you reopen this PR @karandeepSJ?

PR #1578, opened 2018-06-28
BF - bad condition in maximum dg

### LINKED ISSUE #1566: Bad condition in local tracking
In [probabilistic_direction_getter.pyx](https://github.com/nipy/dipy/blob/master/dipy/direction/probabilistic_direction_getter.pyx#L170), you have this block:

    max_idx = 0
    max_value = 0.0
    for i in range(_len):
        if bool_array[i] > 0 and pmf[i] > max_value:
            max_idx = i
            max_value = pmf[i]

    if pmf[max_idx] == 0:
        return 1
        
    good direction!

I think there's a problem with the last condition.

- In simple terms, the loop is a argmax with a `zip(bool_array)`, so we find the argmax if it's a maximal value AND there's a `1` on the `bool_array` at the right place.
- The condition doesn't check if a `max_value` was found for real. It doesn't care if the inner condition is always False. It simply check if `pmf[max_idx] == 0`, which can happen even with a zeroed-out `bool_array`.

We would reach `good direction!` without problem with this hypothetical case:

    pmf        0.4 0.1 0.0 0.0 ... 0.0
    bool_array  F   F   ...wathever...

After a discussion with @jchoude, we think that `if max_value > 0.0` is a good fix. It would force the argmax to have found an actual maximal value. Can someone confirm that this would be a correct/valid solution? I mean, in the anatomical sense. I'm no tracking expert myself so I prefer asking.

gabknight:
this fix #1566.

Thanks @nilgoyette @jchoude

### DIFF
--- a/dipy/direction/probabilistic_direction_getter.pyx
+++ b/dipy/direction/probabilistic_direction_getter.pyx
@@ -180,7 +180,7 @@ cdef class DeterministicMaximumDirectionGetter(ProbabilisticDirectionGetter):
                 max_idx = i
                 max_value = pmf[i]
 
-        if pmf[max_idx] == 0:
+        if max_value <= 0:
             return 1
 
         newdir = self.vertices[max_idx]
--- a/dipy/direction/tests/test_prob_direction_getter.py
+++ b/dipy/direction/tests/test_prob_direction_getter.py
@@ -3,7 +3,8 @@ import numpy.testing as npt
 
 from dipy.core.sphere import unit_octahedron
 from dipy.reconst.shm import SphHarmFit, SphHarmModel
-from dipy.direction import ProbabilisticDirectionGetter
+from dipy.direction import (DeterministicMaximumDirectionGetter,
+                            ProbabilisticDirectionGetter)
 
 
 def test_ProbabilisticDirectionGetter():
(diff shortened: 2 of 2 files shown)


nilgoyette:
If all tests pass, lgtm.

jchoude:
One check if failing, but it seems to not be related to this fix. The error is:

```
  z[index] = x
/home/travis/build/nipy/dipy/venv/lib/python3.5/site-packages/scipy/fftpack/basic.py:160: FutureWarning: Using a non-tuple sequence for multidimensional indexing is deprecated; use `arr[tuple(seq)]` instead of `arr[seq]`. In the future this will be interpreted as an array index, `arr[np.array(seq)]`, which will result either in an error or a different result.
```

jchoude:
Also LGTM.

arokem:
That warning is apparently a bug in the scipy pre-release version. See #1565

arokem:
Could you please add a test that fails with current `master`, but passes when this fix is implemented?

jchoude:
Ok. If no one else wants to check this micro PR, I could go ahead and merge it in a few days.

Anyone wants to test it?

jchoude:
Oups my bad didn't see your last email. I'll let @gabknight and @nilgoyette coordinate on this.

gabknight:
I added the test. thx @arokem @nilgoyette @jchoude

jchoude:
LGTM. Thanks @gabknight @nilgoyette

arokem:
Yep - nice job - thanks for doing that! Per our conversation on #1579, I'm going to go ahead and merge this, despite the failure on the `PRE=1` machine.

Issue #1579, opened 2018-06-28
Error on PRE=1 (cython / numpy) 


skoudoro:
## Description

Equality operator generate a ValueError with the pre cython version. here a small example: 

`simple_mp.pyx` file:
````
import numpy as np
import cython

def masking_test():
	simple_wm = np.array([[0, 0, 0, 0, 0, 0],
                          [0, 0, 1, 0, 0, 0],
                          [0, 1, 1, 1, 0, 0],
                          [0, 1, 1, 1, 0, 0],
                          [0, 0, 0, 0, 0, 0]])
	mask = np.zeros(simple_wm.shape)

	simple_wm[mask==0] = 1
````
Result in a python terminal: 
````
>>> import simple_mp as m
>>> m.masking_test()
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "simple_mp.pyx", line 13, in simple_mp.masking_test
    simple_wm[mask==0] = 1
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
````

Python 3.6
cython 0.29a0
numpy 1.16.0.dev0+2804c03

PRE Error:
````
======================================================================
ERROR: dipy.align.tests.test_parzenhist.test_parzen_joint_histogram
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/koudoro/anaconda/envs/dipy-pre-env-3/lib/python3.6/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/Users/koudoro/Software/dipy/dipy/align/tests/test_parzenhist.py", line 136, in test_parzen_joint_histogram
    P.setup(static, moving, static_mask, moving_mask)
  File "dipy/align/parzenhist.pyx", line 105, in dipy.align.parzenhist.ParzenJointHistogram.setup
    self.smin = np.min(static[smask != 0])
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()

======================================================================
ERROR: dipy.align.tests.test_parzenhist.test_parzen_densities
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/Users/koudoro/anaconda/envs/dipy-pre-env-3/lib/python3.6/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/Users/koudoro/Software/dipy/dipy/align/tests/test_parzenhist.py", line 193, in test_parzen_densities
    parzen_hist.setup(static, moving)
[...]

PR #1580, opened 2018-06-29
Allow PRE=1 job to fail  

skoudoro:
To follow the suggestion on #1579,  This PR allow `PRE=1` to fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -70,6 +70,8 @@ matrix:
       env:
         - INSTALL_TYPE=requirements
         - DEPENDS=""
+        
+  allow_failures:
     - python: 3.5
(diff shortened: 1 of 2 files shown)


arokem:
What do you think, @matthew-brett? Would this be too subtle?

matthew-brett:
I think that's fine, as long as we check from time to time...  Obviously before a release.

skoudoro:
If we all agreed, I think this PR is ready to go @arokem. We need to put a note on the documentation.

arokem:
Would you mind adding a note in https://github.com/nipy/dipy/blob/master/doc/devel/make_release.rst ?

arokem:
Thanks!

PR #1581, opened 2018-07-01
Image registration Workflow with quality metrices

parichit:
1) Opening the PR for the image registration workflow.
2) The tests have been written for both correct and incorrect input.
3) The older branches have been removed and the code is included in the align.py and test_align.py files for code consistency.
4) Also, added the reporting of quality metrics such as the distance between the registered images and the metric of optimal parameters (to assist in quantitative quality assessment).


dipy/workflows/align.py
```diff
@@ -45,16 +54,464 @@ def run(self, input_files, new_vox_size, order=1, mode='constant', cval=0,
...
+
+        img_registration = transform_centers_of_mass(static,
+                                                     static_grid2world,
+                                                     moving,
+                                                     moving_grid2world)
+
+        transformed = img_registration.transform(moving)
+        return transformed, img_registration.affine
+
+    def translate(self, static, static_grid2world, moving,
```
nilgoyette: There's clearly a pattern here. `translate`, `rigid` and `affine` all have the ~same block of code. Only the transform type changes. Maybe you could create a common function for these functions?

dipy/workflows/align.py
```diff
@@ -45,16 +54,464 @@ def run(self, input_files, new_vox_size, order=1, mode='constant', cval=0,
...
+            static = np.array(image.get_data())
+            static_grid2world = image.affine
+
+            image = nib.load(mov_img)
+            moving = np.array(image.get_data())
+            moving_grid2world = image.affine
+
+            self.check_dimensions(static, moving)
+
+            if transform.lower() == 'com':
```
nilgoyette: Instead of writing `transform.lower()` 4 times, you could put it in a variable.

dipy/workflows/align.py
```diff
@@ -45,16 +54,464 @@ def run(self, input_files, new_vox_size, order=1, mode='constant', cval=0,
...
+        setting up the io iterator to gobble the input and output paths
+        """
+        io_it = self.get_io_iterator()
+
+        for static_img, mov_img, moved_file, affine_matrix_file, \
+                qual_val_file in io_it:
+
+            """
+            Load the data from the input files and store into objects.
+            """
```
nilgoyette: It's the first time I see random comments in `"""` I thought it was illegal because it's supposed to be reserved for documentation. I was wrong. IMO, I think we should stick to normal `#` comments.

dipy/workflows/tests/test_align.py
```diff
@@ -20,12 +31,127 @@ def test_reslice():
...
+        moving_image_file = pjoin(temp_out_dir, 't1.nii.gz')
+
+        image_registeration_flow = ImageRegistrationFlow()
+
+        def read_distance(qual_fname):
+            temp_val = 0
+            with open(pjoin(temp_out_dir, qual_fname), 'r') as f:
+                for line in f:
+                    pass
+                temp_val = line
```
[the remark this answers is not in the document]
parichit: @nilgoyette: Thanks for this one, you are absolutely right, I could have done that. Will update the code.

dipy/workflows/tests/test_align.py
```diff
@@ -20,12 +31,127 @@ def test_reslice():
...
+            npt.assert_raises(ValueError, image_registeration_flow.run,
+                              static_image_file,
+                              moving_image_file,
+                              metric='wrong_metric')
+
+        test_com()
+        test_translation()
+        test_rigid()
+        test_affine()
+        test_err()
```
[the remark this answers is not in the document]
parichit: @nilgoyette: Thank again for this suggestion. Here I am a bit confused, please correct me if I am wrong. From what I understand is that during Travis checks, it automatically looks for all the files starting from 'test_' and executes them. I had that syntax (that you have suggested) earlier but then I removed it because I came to know that I don't need it for the test to execute.

However, I will definitely look into it if changing this one will help in both running the test independently and also in the travis checks.

Thanks again.

dipy/workflows/align.py
parichit: @nilgoyette: Yes, you are right. Thanks for this one. Much appreciate it. Will update the code.

dipy/workflows/align.py
parichit: @nilgoyette: I think I just overlooked this one. It makes sense to follow the '#' based commenting for this one. Really appreciate this. Thanks.

dipy/workflows/tests/test_align.py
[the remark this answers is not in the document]
nilgoyette: You're right, Travis and other tools don't need the `run_module_suite()`. However, it's practical for us when we run this file in command line or with the play button in our favorite IDE.

dipy/workflows/tests/test_align.py
```diff
@@ -7,6 +6,18 @@
 from dipy.data import get_data
 from dipy.workflows.align import ResliceFlow
 
+import os.path
+from os.path import join as pjoin
+import shutil
+
+from dipy.align.tests.test_parzenhist import setup_random_transform
+from dipy.align.transforms import (Transform,
+                                   regtransforms)
```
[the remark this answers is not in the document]
parichit: @nilgoyette:  Done, Thanks for pointing this out.

dipy/workflows/tests/test_align.py
```diff
@@ -20,12 +31,127 @@ def test_reslice():
...
+                                         transform='trans',
+                                         out_dir=temp_out_dir,
+                                         out_moved=out_moved,
+                                         out_affine=out_affine,
+                                         save_metric=True,
+                                         level_iters=[100, 10, 1],
+                                         out_quality='trans_q.txt')
+
+            dist = read_distance('trans_q.txt')
+            npt.assert_equal('%.2f' % dist, '%.2f' % -0.3953547764454917)
```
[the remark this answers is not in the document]
parichit: Thanks for this one, yes you are correct, here almost_equal() improves the code readability.

dipy/workflows/align.py
```diff
@@ -45,16 +54,464 @@ def run(self, input_files, new_vox_size, order=1, mode='constant', cval=0,
...
+    registration will be slower but will improve the quality.
+
+    This can be controlled by using the progressive flag (True by default).
+    """
+
+    def center_of_mass(self, static, static_grid2world,
+                       moving, moving_grid2world):
+
+        """ Function for the center of mass based image
+        registration.
```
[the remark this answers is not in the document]
parichit: Yup, fits within 70 characters. Done. Thanks for pointing this out.

dipy/workflows/align.py
parichit: Done.

dipy/workflows/tests/test_align.py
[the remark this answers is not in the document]
parichit: Done, It looks better now. I did not know that the structure returned by readlines() can be directly index like readlines()[-1], thanks for this one.

dipy/workflows/tests/test_align.py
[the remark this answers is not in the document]
parichit: Using run_module_suite() now. Thanks.

dipy/workflows/align.py
parichit: @nilgoyette: You are correct, there is clearly a pattern here but the thing is that this pattern can get a bit more complicated depending on how the user wants to run the analysis. So, I simply separated each type of registration in its own function. These functions, in turn, are calling other dependent functions to generate some intermediate results (depending on the type of analysis the user wants to perform.) So, translate can call center of mass or rigid can call translate and center of mass or don't call them at all.

With a single function to abstract the low-level details, it will become a bit more redundant code because it will also call different functions (like it's being done now). 

But, I am thinking about this one more, if there is a way to further abstract away the code then I surely will do that. It will structure the code even more nicely.

Thanks.

parichit: I have tried to reduce the code duplication by moving the common code from (translate, rigid and affine) functions in a separate function. Thanks for this review.

dipy/workflows/tests/test_align.py
```diff
@@ -20,12 +31,127 @@ def test_reslice():
         out_path = reslice_flow.last_generated_outputs['out_resliced']
         out_img = nib.load(out_path)
         resliced = out_img.get_data()
-        
+
         npt.assert_equal(resliced.shape[0] > volume.shape[0], True)
         npt.assert_equal(resliced.shape[1] > volume.shape[1], True)
         npt.assert_equal(resliced.shape[2] > volume.shape[2], True)
         npt.assert_equal(resliced.shape[-1], volume.shape[-1])
```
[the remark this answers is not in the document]
nilgoyette: @skoudoro and others, do you have an opinion on the use of `assert`? I don't like statements much, but I like `assert_equal(wathever, True)` even less.

dipy/workflows/tests/test_align.py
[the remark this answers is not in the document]
nilgoyette: Good thing you're using `run_module_suite()` but why are you still calling `test_com()`, `test_translation()`, etc.?

dipy/workflows/align.py
```diff
@@ -4,17 +4,26 @@
...
 from dipy.workflows.workflow import Workflow
 
+import numpy as np
+import nibabel as nib
+from dipy.align.imaffine import (transform_centers_of_mass, AffineMap,
+                                 MutualInformationMetric, AffineRegistration)
+from dipy.align.transforms import (TranslationTransform3D, RigidTransform3D,
+                                   AffineTransform3D)
+from dipy.io.image import save_nifti, save_affine_matrix, \
+    save_quality_assur_metric
```
nilgoyette: I don't know if DiPy has a standard or not, but since it's your own code, can you stick to one way? Either `import ()` or `import \`, but not both.

dipy/workflows/align.py
```diff
@@ -45,16 +54,492 @@ def run(self, input_files, new_vox_size, order=1, mode='constant', cval=0,
...
+        affreg : An object of the image registration class.
+
+        params0 : array, shape (n,)
+            parameters from which to start the optimization. If None, the
+            optimization will start at the identity transform. n is the
+            number of parameters of the specified transformation.
+
+        """
+        moved, affine = self.center_of_mass(static, static_grid2world,
+                                            moving, moving_grid2world)
```
nilgoyette: Looks like you're not using `moved` here and below. Simply use `_, affine = ...`

dipy/workflows/align.py
```diff
@@ -45,16 +54,492 @@ def run(self, input_files, new_vox_size, order=1, mode='constant', cval=0,
...
+            optimization will start at the identity transform. n is the
+            number of parameters of the specified transformation.
+
+        progressive : boolean
+            Flag to enable or disable the progressive registration. (defa
+            ult True)
+
+        """
+        if progressive:
+            moved, affine, xopt, fopt = self.rigid(static, static_grid2world,
```
nilgoyette: Idem. When not using an answer, use `_`

dipy/workflows/align.py
```diff
@@ -45,16 +54,492 @@ def run(self, input_files, new_vox_size, order=1, mode='constant', cval=0,
...
+    registration will be slower but will improve the quality.
+
+    This can be controlled by using the progressive flag (True by default).
+    """
+
+    def perform_transformation(self, static, static_grid2world, moving,
+                               moving_grid2world,
+                               affreg, params0, transform, starting_affine):
+
+        """ Function for translation based registration.
```
nilgoyette: Copy pasting is often a trap in programming.

dipy/workflows/align.py
```diff
@@ -45,16 +54,492 @@ def run(self, input_files, new_vox_size, order=1, mode='constant', cval=0,
...
+
+        starting_affine : Affine matrix to be used as starting affine
+            for the optimizer.
+
+        """
+        img_registration, \
+            xopt, fopt = affreg.optimize(static, moving, transform, params0,
+                                         static_grid2world,
+                                         moving_grid2world,
+                                         starting_affine=starting_affine,
```
nilgoyette: I'm not too sure where `optimize` is defined so I'm sure, but check if you actually need to specify the name. `starting_affine=starting_affine,` is kind of useless to read.

dipy/workflows/align.py
```diff
@@ -45,16 +54,492 @@ def run(self, input_files, new_vox_size, order=1, mode='constant', cval=0,
...
+    For example, using affine registration with progressive set to 'True'
+    will involve center of mass, translation, rigid body and full affine
+    registration. Whereas, when progressive is False the registration will
+    include only center of mass and affine registration. The progressive
+    registration will be slower but will improve the quality.
+
+    This can be controlled by using the progressive flag (True by default).
+    """
+
+    def perform_transformation(self, static, static_grid2world, moving,
```
nilgoyette: I understand now why you didn't create this function on the first time. Yes, there was some code copy, but there's so much parameters and comments that you're probably not saving much space. **If** you prefer your old version, you can revert your changes. As you wish, I won't complain on this matter anymore :)

dipy/workflows/tests/test_align.py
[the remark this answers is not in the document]
skoudoro: I am not against `assert` and I think we should use it here

dipy/workflows/tests/test_align.py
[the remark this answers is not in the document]
skoudoro: I was confused too @nilgoyette but when you look at the indentation, these calls are inside `test_image_registration()` function and call local functions.

dipy/workflows/align.py
parichit: Done, Thanks. Actually, part of this file was contributed was someone else earlier and so I did not fiddle with the parts that were already there but you have a valid point, I also support code consistency and standard. It's better to just stick to a single standard rather than using different ones. Good catch, I did not even notice it until you pointed this out.

dipy/workflows/align.py
parichit: Done. I forgot to take advantage of this feature of Python :) Thanks for reminding me.

dipy/workflows/align.py
parichit: Done.

dipy/workflows/align.py
parichit: Yeah, totally agree. Sorry for this. Updated it. :)

dipy/workflows/align.py
parichit: Yeah, I tried and thought about somehow reducing the duplication but figured that it's just not possible given the current specification of parameters and different analysis that can be done with this workflow. If I remove the code from one part then it has to come up in some other part (and it's the same code), Phewww.  But, I haven't given up, still thinking about how to make it less redundant. :)

dipy/align/imaffine.py
```diff
@@ -954,7 +954,7 @@ def _init_optimizer(self, static, moving, transform, params0,
 
     def optimize(self, static, moving, transform, params0,
                  static_grid2world=None, moving_grid2world=None,
-                 starting_affine=None):
+                 starting_affine=None, ret_metric=False):
```
parichit: @nilgoyette: Thanks for pointing out the thing about 'starting_affine'. Here is the optimize function definition, So the 'starting_affine' is there as an optional parameter but I think it is position independent so passing a value to that argument won't work (please correct me if I am wrong). So, that's why I stick to the name 'starting_affine'. But I have changed it to 'starting_affine=affine' from what it was earlier (aesthetically it looks better now). A very good point though, I really appreciate all the useful comments that you have made, it really helped me to improve the quality of the code. Thanks.

dipy/workflows/tests/test_align.py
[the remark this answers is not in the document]
parichit: Done, Updated the code to use native python assert. Also, bought the code duplication down a bit by separating the checks in a separate function.
@nilgoyette and @skoudoro:  Thanks.

Garyfallidis:
This PR needs rebasing on top of master.

dipy/io/image.py
```diff
@@ -24,3 +25,34 @@ def load_nifti(fname, return_img=False, return_voxsize=False,
...
+        File name to save the metric values.
+    xopt: numpy array
+        The metric containing the
+        optimal parameters for
+        image registration.
+    fopt: int
+        The distance between the registered images.
+    """
+    np.savetxt(fname, xopt, header="Optimal Parameter metric")
+    with open(fname, 'a') as f:
```
Garyfallidis: Is 'a' the correct parameter or should it be 'w'?

bin/dipy_align_affine
```diff
@@ -0,0 +1,17 @@
+#!python
+
+"""
+Importing the flow runner method and the ImageRegistrationFlow
+method from the workflows.
```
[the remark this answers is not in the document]
parichit: Thanks for pointing this out. Made the changes.

bin/dipy_align_affine
```diff
@@ -0,0 +1,17 @@
...
+"""
+
+from dipy.workflows.flow_runner import run_flow
+from dipy.workflows.align import ImageRegistrationFlow
+
+"""
+This is the method that will wrap everything that is needed to make a flow
+command line ready then run it.
+"""
```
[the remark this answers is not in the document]
parichit: Removed the comment. Thank you.

dipy/io/image.py
Garyfallidis: No, that is correct. Ignore.

parichit:
@Garyfallidis: Kindly take a brief look at the parameter description in the run method, I have tried to write the description as concise as possible but I think that it can be a bit verbose, i just don't the correct terminology to use. 

Thanks.

dipy/reconst/shm.py
```diff
@@ -997,9 +997,20 @@ def calculate_max_order(n_coeffs):
         Finally, the positive value is chosen between the two options.
         """
 
-        L1 = (-3 + np.sqrt(1 + 8 * n_coeffs)) / 2
-        L2 = (-3 - np.sqrt(1 + 8 * n_coeffs)) / 2
-        return np.int(max([L1, L2]))
+        # L2 is negative for all positive values of n_coeffs, so we don't
```
Garyfallidis: Not sure why this is here. Seems irrelevant to registration. I suppose we need to rebase?

parichit: I already did a rebase and now when I check, it says that the branch is up-to-date. So, I am a bit lost as to what is going wrong in my Git flow. Trying to figure it out.

parichit: You were right, I checked again and the earlier rebase was not complete, there were conflicts remaining. I did a rebase again (after pulling the latest changes from the remote origin) and merged the conflicts, manually. This branch should be okay now :)

skoudoro:
closing in favor of #1604

PR #1583, opened 2018-07-05
Fix the relative SF threshold Issue

### LINKED ISSUE #1582: SF threshold in PMF is not relative
When using the LocalTracking a _ProbabilisticDirectionGetter_ needs to be provided, _ProbabilisticDirectionGetter_ requires a _relative_peak_threshold_. Which is supposed to ensure that a linear scaling of FODf amplitude does not affect the tracking.

Example:
Same datasets, artificially scaled FODf (factor 0.1, 1, 10) should produced identical tracking. This is not the case. The results are that tracking using the factor 0.1 results in 0 streamlines and a large difference is observable between the local tracking using the factor 1 and the factor 10.

https://drive.google.com/open?id=1j-V6xx4Y18e1B_LsKgdGT0t8qdUYkVCz
@gabknight is the sfthreshold aimed to cover this situation ?

frheault:
Following the 4D interpolation and the fit, the amplitude of the values on the sphere are not always between 0 and 1, they can be very large or very small. Right after, a condition using a threshold (between 0 and 1) is performed.
The values on the sphere need to be scaled between 0 and 1 before applying the threshold.

As mentionned in #1582 , without this a simple linear scaling of the same FODf affect the tracking (ranging from a complete absence of streamlines to a broad tracking where most direction are acceptable)

### DIFF
--- a/dipy/direction/closest_peak_direction_getter.pyx
+++ b/dipy/direction/closest_peak_direction_getter.pyx
@@ -102,8 +102,9 @@ cdef class BaseDirectionGetter(DirectionGetter):
         pmf = self.pmf_gen.get_pmf_c(point)
         _len = pmf.shape[0]
         max_pmf = np.max(pmf)
+        relative_pmf_threshold = self.pmf_threshold*max_pmf
         for i in range(_len):
-            if pmf[i] < self.pmf_threshold*max_pmf:
+            if pmf[i] < relative_pmf_threshold:
                 pmf[i] = 0.0
         return pmf
 
--- a/dipy/direction/closest_peak_direction_getter.pyx
+++ b/dipy/direction/closest_peak_direction_getter.pyx
@@ -98,11 +98,12 @@ cdef class BaseDirectionGetter(DirectionGetter):
         cdef:
             size_t _len, i
             double[:] pmf
+            double relative_pmf_threshold
 
         pmf = self.pmf_gen.get_pmf_c(point)
         _len = pmf.shape[0]
-        max_pmf = np.max(pmf)
-        relative_pmf_threshold = self.pmf_threshold*max_pmf
+
+        relative_pmf_threshold = self.pmf_threshold*np.max(pmf)
         for i in range(_len):
             if pmf[i] < relative_pmf_threshold:
                 pmf[i] = 0.0
--- a/dipy/direction/closest_peak_direction_getter.pyx
+++ b/dipy/direction/closest_peak_direction_getter.pyx
@@ -98,14 +98,14 @@ cdef class BaseDirectionGetter(DirectionGetter):
         cdef:
             size_t _len, i
             double[:] pmf
-            double relative_pmf_threshold
+            double absolute_pmf_threshold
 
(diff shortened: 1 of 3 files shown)


dipy/direction/closest_peak_direction_getter.pyx
```diff
@@ -101,8 +101,9 @@ cdef class BaseDirectionGetter(DirectionGetter):
 
         pmf = self.pmf_gen.get_pmf_c(point)
         _len = pmf.shape[0]
+        max_pmf = np.max(pmf)
         for i in range(_len):
-            if pmf[i] < self.pmf_threshold:
+            if pmf[i] < self.pmf_threshold*max_pmf:
```
jchoude: You could factor the multiplication out of the loop.

jchoude:
@gabknight Would you have a few minutes to look at this? I'm pretty confident in the identification of the issue and @frheault 's fix, but since you're more familiar than me in Dipy's tracking code, I'd like a second opinion.

dipy/direction/closest_peak_direction_getter.pyx
```diff
@@ -101,8 +101,10 @@ cdef class BaseDirectionGetter(DirectionGetter):
 
         pmf = self.pmf_gen.get_pmf_c(point)
         _len = pmf.shape[0]
+        max_pmf = np.max(pmf)
+        relative_pmf_threshold = self.pmf_threshold*max_pmf
```
gabknight: Can you add ```double relative_pmf_threshold``` to the cdef declarations

I would also remove ```max_pmf``` and do:
``` relative_pmf_threshold = self.pmf_threshold * np.max(pmf)```

**Comment by gabknight** [comment]
Hi @frheault @jchoude,

Thanks for spotting this. We are currently using an absolute threshold not a relative threshold despite the documentation saying relative threshold. Your change fixes this. 

- Can you look into the failing tests? I suspect the toy PMFs used in the tests are not between 0 and 1. You can probably put the threshold to 0 for most tests, instead of the default 0.1.

- Can you also add a test that fails on current master and passes with this PR? E.g. multiplying a pmf by 1 or 10, resulting in different ```ProbabilisticDirectionGetter``` directions with the same relative pmf threshold.

File: dipy/direction/closest_peak_direction_getter.pyx [hunk 3]
```diff
@@ -98,11 +98,14 @@ cdef class BaseDirectionGetter(DirectionGetter):
         cdef:
             size_t _len, i
             double[:] pmf
+            double relative_pmf_threshold
 
         pmf = self.pmf_gen.get_pmf_c(point)
         _len = pmf.shape[0]
+
+        relative_pmf_threshold = self.pmf_threshold*np.max(pmf)
```
jchoude: Just to make sure, can it happen that `pmf==0` everywhere?

jchoude: Also, is it still relative if you scale it?

jchoude:
LGTM. 

Should we go ahead and merge this by monday morning, August 20th?

arokem:
👍 from me!

PR #1584, opened 2018-07-06
Added Examples

karandeepSJ:
This PR adds examples for the new elements in viz.ui. Currently, this includes the examples for the merged elements. I will add more as the other PRs get merged.


dmreagan:
I'm not sure we want a second UI example. I think I'd prefer to clean up the existing example and add these to it. By clean up, I mean:

- make the window larger so you can see everything
- line all the elements up in a nice grid

Something like [this](https://semantic-ui.com/kitchen-sink.html)

karandeepSJ:
New Examples:
examples

dmreagan:
Huge improvement over the previous mess! Ultimately we'll want this to replace _viz_ui.py_ instead of adding a second file. We'll also want to edit the text so it provides a coherent tutorial for the user.

skoudoro:
I feel like this PR is ready to go. Any other comment? I will wait until tomorrow for merging it

doc/examples/viz_ui.py
```diff
@@ -4,11 +4,10 @@
...
 
-This example shows how to use the UI API.
-Currently includes button, textbox, panel, and line slider.
+This example shows how to use the UI API. We will demonstrate how to create
+several DIPY UI elements, then use a list box to toggle which element is shown.
 
 First, a bunch of imports.
-
 """
 import os
```
skoudoro: can you separate `"""` and `import os` by an empty line. 

We need it to render the documentation correctly

skoudoro:
Thank you @karandeepSJ and @dmreagan. Merging!

Issue #1588, opened 2018-07-16
Dipy.reconst.shm.calculate_max_order only works on specific cases.


jchoude:
## Description
Function `calculate_max_order()` from `dipy.reconst.shm` doesn't work as expected.

From it's use and history, I think it should be called `find_sh_order()`, and it should be explained that it only works (at least, in its current form), for number of coefficients that are really yielded from a correctly formed model (ex: order 6 == 28 coeffs, etc).

The first issue is that, in all cases, `L2` will always yield a complex number. This, of course, is not useful, and will always be discarded when choosing the value to return. I know this comes from finding the roots of the polynomial, but we could just not compute it, since it serves no real purpose.

The second issue depends on the intended use of that function. From its use, we want to guess the maximal SH order that could have been used with a specific number of coefficients. For values that are the real number of coefficients (for example: 28 for order 6, 45 for order 8, etc), the returned value is exactly the order associated. However, what is the intended behavior when the number of coefficients is incorrect?

Say you provide `n_coeffs=29`, in the end, we get a result of 6. However, this is due to the cast to `int`. Normally, 29 coefficients is not a correct order 6 fODF file.

Even worse, let's say we give `n_coeffs=43`, we get a result of `sh_order=7`. Normally, we only use even SH orders, so that doesn't really make sense.

Tagging @arokem since last commit on this comes from his work on PR #724 . Any opinion on refactoring or restricting the use of this to valid SH coefficients number?

arokem:
We should definitely not give back wrong answers :-)

Regarding L2: are you sure it's always complex? It's possible that it's
always negative, but I am not sure about complex. Unless `n_coeffs` is set
to a negative number.

Regarding the name, let me walk you through my logic: the reason I am using
`max_order`  in the name is that I think of a set of spherical harmonics as
having multiple functions each of which is defined by its degree and order.
The number we are interested in here is the highest order in this set of
spherical harmonic functions. Does that make sense?

Regarding a potential solution: one option would be to store a lookup table
for a (large!) range of cases and raise an error when an invalid value is
given. What do you think about that?

jchoude:
Thanks for having a look! I don't think it's an urgent issue but I noticed it so I wanted to check with you :)

For `L2`, I agree that the number will not be a complex number, I looked too quickly. The result would still always be negative for n_coeffs >= 0. I see that you adressed that in your PR.

[...]

PR #1589, opened 2018-07-18
Fix calculation of highest order for a sh basis set

arokem:
Starting to think about addressing gh-1588.


jchoude:
LGTM!

Thanks @arokem .

I'll wait until Travis finishes, and if everything is good I'll go ahead with this since it's a really well contained change. Except if you think someone else should also review it?

jchoude:
Since there was no additional feedback and this was really small, I'll go ahead and merge.

Thanks @arokem

PR #1592, opened 2018-07-24
Added File Menu element to viz.ui

karandeepSJ:
This PR reopens #1570 
Added a FileMenu2D that uses a ListBox2D to display all files and directories in the file system. Files are shown with Blue color and directories with Green. Clicking on a directory will take the user inside it.
filemenu


dipy/viz/ui.py
```diff
@@ -3303,3 +3306,225 @@ def left_button_clicked(self, i_ren, obj, list_box_item):
...
+        -------
+        file_names: list(string)
+            List of all file names as string.
+        """
+        # A list of file names with extension in the current directory
+        for (_, _, files) in os.walk(self.current_directory):
+            break
+
+        file_names = []
+        if "*" in self.extensions:
```
dmreagan: It should also probably treat an empty string "" the same as "*". Then it can accept `None` for directories-only.

dmreagan:
Looks good. Any thoughts, @ranveeraggarwal and @MarcCote ?

dipy/viz/tests/test_ui.py
```diff
@@ -627,6 +627,78 @@ def test_ui_image_container_2d(interactive=False):
...
+    show_manager.ren.add(filemenu)
+
+    # Recorded events:
+    #  1. Click on 'testfile.txt'
+    #  2. Click on 'tempdir/'
+    #  3. Click on 'test0.txt'.
+    #  4. Shift + Click on 'test6.txt'.
+    #  5. Click on '../'.
+    #  2. Click on 'testfile.txt'.
+    show_manager.play_events_from_file(recording_filename)
```
MarcCote: Where does the recording file come from? I mean how do you record it? I don't see any `EventCounter.save`.

dipy/viz/ui.py
```diff
@@ -3230,8 +3231,10 @@ def _setup(self):
                                      vertical_justification="middle")
 
         # Add default events listener for this UI component.
-        self.textblock.on_left_mouse_button_clicked = self.left_button_clicked
-        self.background.on_left_mouse_button_clicked = self.left_button_clicked
+        self.add_callback(self.textblock.actor, "LeftButtonPressEvent",
```
MarcCote: Is this really necessary, we should avoid directly adding callbacks to VtkProp whenever is possible.

dipy/viz/ui.py
```diff
@@ -3300,3 +3303,226 @@ def left_button_clicked(self, i_ren, obj, list_box_item):
...
+        if self.reverse_scrolling:
+            up_event, down_event = down_event, up_event  # Swap events
+
+        self.add_callback(self.listbox.panel.background.actor, up_event,
+                          self.scroll_callback)
+        self.add_callback(self.listbox.panel.background.actor, down_event,
+                          self.scroll_callback)
+
+        # Handle mouse wheel events on the slots.
+        for slot in self.listbox.slots:
```
MarcCote: Still not a fan of reimplementing things that are already in `ListBox2D` (see https://github.com/nipy/dipy/pull/1570#issuecomment-397904982 and https://github.com/nipy/dipy/pull/1570#issuecomment-397906003) but I don't have time to think it through and already block you for long enough.

dipy/viz/tests/test_ui.py
karandeepSJ: The events are already recorded and saved. I removed the recording part because in the earlier implementation, there was no `interactive` parameter and if a user wanted to interact with the elements, they would have to set `recording = True` which changes the log files.
Maybe we can have two parameters `recording` and `interactive`?

MarcCote: That's a good idea. We could have a single parameter `mode=None|'interactive|'recording'`. I found the recording useful when we change some core functionality of the UI framework and have to redo the recording. But to be fair that happens less often nowadays.

karandeepSJ: I'll do that in another PR since all the tests will have to be changed.
@dmreagan @Garyfallidis @ranveeraggarwal should I go ahead with this?

dmreagan:
I agree that this one has been held up long enough. I'll merge in 48 hours if there are no further objections. @MarcCote can you make an issue with your remaining concerns so we can remember to address them later?

MarcCote:
@dmreagan sure (see https://github.com/nipy/dipy/issues/1618).

PR #1595, opened 2018-07-25
Workflow for visualizing the quality of the registered data with DIPY

parichit:
This branch is created from the older 'affine_visual' branch and only focuses on integrating the visualization support for the registered images.

These changes will form the part of the visual quality assessment in the DIPY workflows and will complement the quantitative assessment from the affine_registration branch.


Garyfallidis:
This should be merged after #1581 is merged.

dipy/workflows/vis_registeration.py
```diff
@@ -0,0 +1,278 @@
...
+                                                       moved_img)
+
+        renderer = window.Renderer()
+        renderer.background((0.5, 0.5, 0.5))
+
+        slice_actor = actor.slicer(overlay, affine, value_range)
+        renderer.projection('parallel')
+
+        cnt = 0
+        X, Y, num_slices = slice_actor.shape[:3]
```
nilgoyette: I thought PEP8 didn't like upper case names. Isn't there a warning?

dipy/workflows/vis_registeration.py
```diff
@@ -0,0 +1,278 @@
...
+
+        if sli_type == 'saggital':
+            num_slices = x
+            slice_type = 0
+        elif sli_type == 'coronal':
+            num_slices = y
+            slice_type = 1
+        elif sli_type == 'axial':
+            num_slices = z
+            slice_type = 2
```
nilgoyette: It's `sagittal`. Also, what happens when it's something else than those three names? Ok, you write `None (default).` but it looks like this function is doing nothing on `None`.

dipy/workflows/vis_registeration.py
```diff
@@ -0,0 +1,278 @@
...
+                                      sli_type, fname, affine):
+
+        """
+        Function for creating the animated GIF from the slices of the
+        registered image.
+        It uses the renderer object to control the dimensions of the
+        created GIF and for correcting the orientation.
+        Patameters
+        ----------
+        sli_type : str (optional)
```
nilgoyette: You're only saving 2 characters. It's not worth it. Can you please name it `slice_type`?

dipy/workflows/vis_registeration.py
```diff
@@ -0,0 +1,278 @@
...
+            snap1 = self.adjust_color_range(snap)
+            slices.append(snap1)
+
+        # Writing the GIF below
+        write_gif(slices, fname, fps=10)
+
+    @staticmethod
+    def check_dimensions(static, moving):
+
+        """Check the dimensions of the input images."""
```
nilgoyette: What's your standard? You sometime glue the documentation to your function, sometime not. I would advise removing the useless \n, but wathever you do, can you please stick to one?

dipy/workflows/vis_registeration.py
```diff
@@ -0,0 +1,278 @@
...
+
+        if len(static.shape) != len(moving.shape):
+            raise ValueError('Dimension mismatch: The'
+                             ' input images must have same number of '
+                             'dimensions.')
```
nilgoyette: Again, stop using \n everywhere. This string could easily fit one 2 lines. It's harder to read for no reason.

dipy/workflows/vis_registeration.py
```diff
@@ -0,0 +1,278 @@
...
+
+        animate_file : string, optional
+            Name of the GIF file for saving the animation
+            (default animation.gif).
+
+        """
+
+        io_it = self.get_io_iterator()
+
+        for static_img, mov_img, affine_matrix_file in io_it:
```
nilgoyette: Any reason to create a var with `io_it`? It doesn't help comprehension. You can simply use it directly in the for loop.

dipy/workflows/tests/test_align.py
```diff
@@ -7,6 +6,16 @@
 from dipy.data import get_data
 from dipy.workflows.align import ResliceFlow
 
+import os.path
+from os.path import join as pjoin
+import shutil
+
+from dipy.align.tests.test_parzenhist import setup_random_transform
+from dipy.align.transforms import (Transform,
+                                   regtransforms)
```
nilgoyette: Useless \n and ().

dipy/workflows/tests/test_align.py
```diff
@@ -20,12 +29,105 @@ def test_reslice():
...
+        out_moved = pjoin(temp_out_dir, "affine_moved.nii.gz")
+        out_affine = pjoin(temp_out_dir, "affine_affine.txt")
+
+        image_registeration_flow.run(static_image_file, moving_image_file,
+                                     transform='affine', out_dir=temp_out_dir,
+                                     out_moved=out_moved,
+                                     out_affine=out_affine,
+                                     level_iters=[100, 10, 1])
+
+        npt.assert_equal(os.path.exists(out_moved), True)
```
nilgoyette: As advised in another review, simply use the `assert` statement when writing `npt.assert_equal(..., True)`

dipy/workflows/core.py
```diff
@@ -0,0 +1,426 @@
...
+    """Convert the NumPy array to two nested lists with r,g,b tuples."""
+    dim, nrow, ncol = dataset.shape
+    uint8_dataset = dataset.astype('uint8')
+    if not (uint8_dataset == dataset).all():
+        message = (
+            "\nYour image was cast to a `uint8` (`<img>.astype(uint8)`), "
+            "but some information was lost.\nPlease check your gif and "
+            "convert to uint8 beforehand if the gif looks wrong."
+        )
+        warnings.warn(message)
```
nilgoyette: Maybe this is more a question of style, so don't feel forced to change it. I see no advantage in creating a message from a string, then calling a warning with it. Why don't you simply write it in the argument? It takes less place. Idem for this pattern at other places.

    warnings.warn(
        "\nYour image was cast to a `uint8` (`.astype(uint8)`), "
        "but some information was lost.\nPlease check your gif and "
        "convert to uint8 beforehand if the gif looks wrong.")

Also, it's the first time I see a warning starting by `\n`. Is this normal in DiPy?

dipy/workflows/core.py
```diff
@@ -0,0 +1,426 @@
...
+        dataset = try_fix_dataset(dataset)
+        check_dataset(dataset)
+    delay_time = 100 // int(fps)
+
+    def encode(d):
+        four_d = isinstance(dataset, numpy.ndarray) and len(dataset.shape) == 4
+        if four_d or not isinstance(dataset, numpy.ndarray):
+            return _make_animated_gif(d, delay_time=delay_time)
+        else:
+            return _make_gif(d)
```
nilgoyette: Again, maybe a question of style, but a `else` after a `return` is useless.

dipy/workflows/align.py
```diff
@@ -1,20 +1,36 @@
 from __future__ import division, print_function, absolute_import
 import logging
 from dipy.align.reslice import reslice
-from dipy.io.image import load_nifti, save_nifti
 from dipy.workflows.workflow import Workflow
 
+import numpy as np
+import nibabel as nib
+from dipy.viz.regtools import overlay_slices
+from dipy.viz import (window, actor)
```
nilgoyette: Useless ()

dipy/workflows/align.py
```diff
@@ -45,16 +61,576 @@ def run(self, input_files, new_vox_size, order=1, mode='constant', cval=0,
...
+            num_slices += 1
+
+        for i in range(5, num_slices):
+
+            if num_slices % i == 0:
+                rows = i
+                cols = num_slices // i
+                break
+
+        return rows, cols
```
nilgoyette: Hey, I already commented on this function in another file! We rarely have to copy code. DRY. In fact, I recognize some of the code in this file. What's happening? Can't you use common functions?

dipy/workflows/align.py
```diff
@@ -45,16 +61,576 @@ def run(self, input_files, new_vox_size, order=1, mode='constant', cval=0,
...
+        Patameters
+        ----------
+        slice_type : str (optional)
+            the type of slice to be extracted:
+            sagital, coronal, axial, None (default).
+        fname: str, optional
+            Filename for saving the GIF (default 'animation.gif').
+        """
+
+        if sli_type == 3:
```
nilgoyette: Why would `slice_type : str (optional)` be equal to 3?

dipy/workflows/vis_registeration.py
```diff
@@ -0,0 +1,278 @@
...
+        """
+        Function for creating the mosaic of the moved image. It
+        currently only supports slices from the axial plane.
+
+        fname: str, optional
+            Filename to be used for saving the mosaic (default 'mosaic.png').
+        """
+
+        overlay, value_range = self.process_image_data(static_img,
+                                                       moved_img)
```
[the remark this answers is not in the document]
parichit: Yup, It fits in 67 characters, fixed it. Thanks.

dipy/workflows/vis_registeration.py
parichit: Yes, PEP8 don't like capital letter names but I kept it that way to remind me that these are the image dimensions. I was experimenting so there were small case x,y and z too. Reverted to the small case now.

dipy/workflows/vis_registeration.py
parichit: Yup, it is 'sagittal', corrected the typo. Also, earlier I was checking the 'None' inside the functions and returning immediately if that was true. Later I decided to check for this in the run method but forgot to remove the 'None' from the function documentation. Good point, though. Thanks.

dipy/workflows/vis_registeration.py
parichit: Yes, you are right, though, I was already using the slice_type in the function (for storing the integer indices for the slice type and so sli_type was just a string name for the convenience of the user). Now, I am just overriding the variable. I did not know that I can also override it. I tried it now and it is working. Thank you.

dipy/workflows/tests/test_align.py
parichit: Done. Thanks.

dipy/workflows/align.py
parichit: yes, thanks.

dipy/workflows/align.py
parichit: So here is the thing: This branch was originally part of the affine_registration PR 1581 and so you will remember that you suggested a good number of improvements to that PR. I did all those changes to that PR and I was hoping that it will be merged. Meanwhile, I continued working on this branch thinking that when PR 1581 will be merged, I can just rebase and use all the latest changes here. 

But, this branch reached its's completion and the PR 1581 is still not merged so a large number of changes that you said, for example, using assert and not npt.assert_equal or not using () while importing or sticking to just one standard for documentation in the functions have already been done in PR 1581. nevertheless, i manually copied all the changes from my local repo for PR 1581 and put them here so that they align with all the improvements that I made in PR 1581. 

Sorry for the inconvenience and thanks very much for the useful comments :).

dipy/workflows/vis_registeration.py
```diff
@@ -0,0 +1,278 @@
...
+            num_slices += 1
+
+        for i in range(5, num_slices):
+
+            if num_slices % i == 0:
+                rows = i
+                cols = num_slices // i
+                break
+
+        return rows, cols
```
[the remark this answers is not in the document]
parichit: yeah, I have to find a way to improve this function. The way it is right now it works well for most inputs but it needs more improvement. Not just the conciseness as you suggested but the mathematics itself can be optimized much. I am simply trying to factor a number into its 2 factors so that the number of rows and columns that I get back can be used in the create_mosaic function.

So, if there are 74 slices (for example in an image, I increase it to 75 to make it a multiple of 5). It can be broken down into either 25 *3 or 5 * 15 or 75 * 1 etc. but you will agree that having 75 rows and 1 column or 1 row and 75 column is not the best way to show the slices side by side. Similarly, 25 rows and 3 columns is also not a good option. But, I can use 5 * 15 means 5 rows and 15 columns for visualizing all the slices side by side. 

Arriving at this combination is made difficult by the fact that I can different number of slices in different inputs and what is the most optimal number of rows and columns to create the mosaic?

I need to come up either with a better algorithm or a clever trick to achieve this.

But many thanks for sharing your concerns with me. I really appreciate your help.

skoudoro:
closing in favor of #1606

PR #1601, opened 2018-07-27
Fix: Decrease Nosetest warning

skoudoro:
The goal of this PR is to reduce the number of warnings and avoid what you can see [here](https://travis-ci.org/nipy/dipy/jobs/408195332#L2879).

This should make Travis happy.

more information on #1565

### DIFF
--- a/dipy/reconst/dsi.py
+++ b/dipy/reconst/dsi.py
@@ -5,6 +5,7 @@ from dipy.reconst.odf import OdfModel, OdfFit
 from dipy.reconst.cache import Cache
 from dipy.reconst.multi_voxel import multi_voxel_fit
 
+from dipy.testing import setup_test
 
 class DiffusionSpectrumModel(OdfModel, Cache):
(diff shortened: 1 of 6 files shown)


arokem:
LGTM! Let's see what that looks like on Travis.

skoudoro:
Everything is green! Travis seems to be happy! it is ready to be merged @arokem 

By default, Nosetest set up the warning filter as "always" which is the highest level. I switch it to "default" which means it will print the first occurrence of matching warnings for each location (module + line number) where the warning is issued

More info [here](https://docs.python.org/3/library/warnings.html)

Garyfallidis:
I agree to merge. Thousands of repeated warning messages are really not useful. And we need to have master passing again, so that the other PRs can be rebased etc. 
So, +1 on my side. Thanks.

arokem:
Since we seem to have agreement, I will go ahead with the merge.

Issue #1596, opened 2018-07-25
Reproducibility in PFT tracking


GuillaumeTh:
After some tests with @jchoude on the PFT tracking, we saw a issue. If we run twice the algorithm with maps slightly differents (1 voxel), the tracking is not reproducible.

I think it could be cool to have a reproducibility of 99% on the tracking if the maps are reproducible at 99%. 

We need to set the random seed follow the voxel and the global random seed( random.seed(voxel_id * random_seed) ).

I checked the files where we need this and they are:

tracking/utils.py -> random_seeds_from_mask
tracking/local/localtrack.pyx -> _pft line 482 and line 502 
direction -> probabilitic_direction_getter.pyx line 120

What do you think about this feature @gabknight

gabknight:
@GuillaumeTh This will be nice to have, but this isn't limited to PFT. This should be implemented for all tracking algorithms. I think this should go in tracking/local/direction_getter as a function setting the random state, which should be called in LocalTracking and ParticleFilteringTracking before starting the tracking from an initial seed position. The random seed should not only be based on the voxel_id, otherwise all streamline started at the same voxel will be will be identical or biased in one direction. Maybe voxel id and seed_count? This could also be floating point 3D position of the seed, thus an identical seed will then always give an identical streamline, but then this removes the option to generate e.g. 1k streamlines from the very same seeding position.

I'm not sure here how this should be done...

GuillaumeTh:
I think the random seed for each seed tracking could be what you said. I will try to have a draft and check if it's easy to do or not.

gabknight:
Ok thx let me know if I can help with this. 

I wouldn't "reset" the random seed in the middle of streamline being traced (localtracking.pyx:502), as this would biais the overall reconstruction.

GuillaumeTh:
So I tested something and It seems to work.

Finally I changed the function `random_seed_from_mask` to add a random_seed argument. Moreover I compute the same seeds in a same voxels follow the random_seed argument.

Before the initial_direction, I setted the numpy.random and random seed. With these modifications, I am reproducible at 99% if the masks are reproducible at 99%.

GuillaumeTh:
So @gabknight, I can create a PR if you want to see.

@skoudoro Are you ok to review this functionality ?

skoudoro:
No problem @GuillaumeTh, but first, I just need to fix our problem with Travis.

GuillaumeTh:
@skoudoro Can I create the PR with the Travis problem ?

PR #1602, opened 2018-07-28
Fix random seed in tracking

GuillaumeTh:
Fix random seed by voxels during the tracking to be reproducible in each voxels. This fix impact the PFT and the local tracking.

More details in issue #1596 .

@gabknight, @skoudoro everything looks fine for the PFT results.


dipy/tracking/utils.py
```diff
@@ -483,13 +483,19 @@ def random_seeds_from_mask(mask, seeds_count=1, seed_count_per_voxel=True,
     else:
         seeds_per_voxel = seeds_count
 
-    # Generate as many random triplets as the number of seeds needed
-    grid = np.random.random([seeds_per_voxel * num_voxels, 3])
-    # Repeat elements of 'where' so that it can be added to grid
-    where = np.repeat(where, seeds_per_voxel, axis=0)
-    seeds = where + grid - .5
+    seeds = []
+    for i in range(1, seeds_per_voxel + 1):
```
Garyfallidis: This looks much slower than before

dipy/tracking/local/localtracking.py
```diff
@@ -116,6 +117,9 @@ def _generate_streamlines(self):
         B = F.copy()
         for s in self.seeds:
             s = np.dot(lin, s) + offset
+            # Fix the random seed in numpy and random
+            random.seed(np.sum(s))
```
Garyfallidis: @GuillaumeTh Not sure what is the fix here.

GuillaumeTh: We fix the random seed with the sum of seed coordinates.

dipy/tracking/utils.py
GuillaumeTh: It is possible but you call this function one time so I think if it's slower it's not a real problem.

GuillaumeTh: I will benchmark before and after the fix

GuillaumeTh: Yes, it is slower I check to be faster.

dipy/tracking/local/localtracking.py
GuillaumeTh: Is it more clear if I say " Set the random seed in numpy and random" ?

dipy/tracking/utils.py
GuillaumeTh: So, @Garyfallidis I think I can't be more faster the slower line is to set the random seed in numpy.

dipy/tracking/utils.py
```diff
@@ -483,13 +483,19 @@ def random_seeds_from_mask(mask, seeds_count=1, seed_count_per_voxel=True,
...
+        for s in where:
+            # Set the random seed with the current seed, the current value of
+            # seeds per voxel and the global random seed.
+            np.random.seed((s + 1) * i + random_seed)
```
gabknight: For consistency, I think this should be `np.random.seed(hash((np.sum(s) + 1) * i + random_seed))`.

dipy/tracking/utils.py
```diff
@@ -483,13 +483,19 @@ def random_seeds_from_mask(mask, seeds_count=1, seed_count_per_voxel=True,
...
+            # Generate random triplet
+            grid = np.random.random(3)
+            seed = s + grid - .5
+            seeds.append(seed)
     seeds = asarray(seeds)
 
+    np.random.seed(random_seed)
     if not seed_count_per_voxel:
         # Randomize the seeds and select the requested amount
         np.random.shuffle(seeds)
```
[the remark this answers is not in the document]
GuillaumeTh: I'll modify the function to have this feature.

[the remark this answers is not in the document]
GuillaumeTh: But if we want less than len(seeds) we have not the same seeds. Something that could be correct this problem is to shuffle the data first and then compute the seeds. 

@gabknight Do you agree ?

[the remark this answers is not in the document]
GuillaumeTh: I will push this feature. I tested it and it works correctly for npv and nt

GuillaumeTh:
@gabknight after the dipy tests I saw that is know a good idea to use directly a hash because the hash could be superior than 2**32 -1 and it crash when we set the random seed. Maybe just remove the hash for random_seeds_from_mask and _generate_streamlines ?

dipy/tracking/utils.py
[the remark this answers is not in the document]
gabknight: OK. I think you solved it but by shuffling the indices instead of the seeds. 

Can you add a test comparing the resulting seed positions, with the same `random_seed` but various npv and nt, e.g. with a mask of 100 voxel:
```
random_seeds_from_mask(2, True)[:150] 
== random_seeds_from_mask(3, True)[:150] 
== random_seeds_from_mask(150, False)[:150] 
== random_seeds_from_mask(500, False)[:150]
```

gabknight:
@GuillaumeTh OK, I don't know what is best for the hash. I suggested to use `hash(.)` because this is what happen under the hood for random.seed(.): (from https://docs.python.org/2/library/random.html)
```
random.seed(a=None)¶
Initialize internal state of the random number generator.

None or no argument seeds from current time or from an operating system specific randomness source if available (see the os.urandom() function for details on availability).

If a is not None or an int or a long, then hash(a) is used instead. Note that the hash values for some types are nondeterministic when PYTHONHASHSEED is enabled.
``` 
Adding hash(.)%(2^32 - 1) is probabably OK.

GuillaumeTh:
I added the test and the hash correction. 

Any things else @gabknight @Garyfallidis @skoudoro ?

GuillaumeTh:
Hi,

I would like to use this code quickly. Is it possible to say me if everything is ok @skoudoro @gabknight ?

gabknight:
LGTM. thx @GuillaumeTh

dipy/tracking/local/localtracking.py
```diff
@@ -1,4 +1,5 @@
 import numpy as np
+import random
```
jchoude: If we want to be really "strict", this import should be first in its own block since this is a core Python import.

dipy/tracking/local/tests/test_tracking.py
```diff
@@ -195,15 +195,13 @@ def test_probabilistic_odf_weighted_tracker():
     def allclose(x, y):
         return x.shape == y.shape and np.allclose(x, y)
 
-    path = [False, False]
```
jchoude: I might be missing something, but why was this modified?

dipy/tracking/utils.py
jchoude: This will definitively be a bit slower than before. However, I think that, as @GuillaumeTh said, this is called only once at the beginning, to generate the seeds. In the overall processing time, this should be negligible, and has the big advantage of allowing a truly reproducible tracking (between runs on the same dataset).

dipy/tracking/utils.py
```diff
@@ -413,16 +413,15 @@ def seeds_from_mask(mask, density=[1, 1, 1], voxel_size=None, affine=None):
 
 
 def random_seeds_from_mask(mask, seeds_count=1, seed_count_per_voxel=True,
-                           affine=None):
+                           affine=None, random_seed=0):
```
jchoude: @GuillaumeTh Could you test how complicated it would be to also manage the case where `random_seed=None`, i.e. set to the previous default behavior? In case someone wants to keep the original behavior.

jchoude:
LGTM. Since @gabknight has already given a thumbs up, I'll wait until tuesday evening to merge.

Issue #1500, opened 2018-04-28
Updating streamlines API in streamlinear.py


Garyfallidis:
This a note to myself to update line 901 of streamlinear.py with the following masking trick using the new Streamlines API.

Here is an example of what needs to be done rather than using list comprehension.

```
import numpy as np
from dipy.tracking.streamline import Streamlines
s = Streamlines()
initial = Streamlines([np.randon.rand(5, 3), np.random.rand(2, 3), np.ones((5, 3))])
initial = Streamlines([np.random.rand(5, 3), np.random.rand(2, 3), np.ones((5, 3))])
new = initial[[True, False, True]]
```

skoudoro:
This issue is fixed, there is no more list comprehension in `streamlinear.py`.

Closing

PR #1617, opened 2018-08-14
WIP: added streamline clipping function to utils

kesshijordan:
Added a function that clips streamlines at the edge of an ROI (discards points past the first encounter of the ROI, keeping the longer end outside of the target ROI). I still need to write tests, etc. but I thought I'd solicit feedback now. @arokem you need this for pyAFQ, right? Do you have any thoughts on this?


arokem:
Yes: would be useful to have something like this. I wonder whether there is any way to edit a nibabel `ArraySequence` objects in-place to do this, rather than yielding into a new sequence. Regarding the specifics of your code -- I wonder whether we could do something else rather than raise those errors. The first one seems reasonable: maybe it's OK to require that users check that the streamlines touch the ROI, before passing them to this function. The second seems a tad draconic: how would users know in advance that none of their streamlines would fall into this condition?

kesshijordan:
Thanks for the feedback, @arokem. That's fair... I suppose the user would notice if they get the wrong part of the track returned : ) The first error is useful because it tells the user when they have the affine wrong, but I'll remove the second one. Regarding the nibabel streamline API, I think if we do that it would make sense to migrate targeting over at the same time, since they're so similar. Is there a larger ongoing effort underway to better integrate the streamline API that this falls into? Also, do you have an opinion on @dmreagan 's suggestion?

arokem:
I think that a couple of simple tests would be helpful to understand what this does (and of course necessary at a later stage anyway).

Regarding the ArraySequence API -- we are trying to incorporate this in as much as possible (e.g., see #1379, #1500), so it would be good for new functionality to be aware of this API and take advantage of it. It has performance benefits. 

I don't think that I understood @dmreagan's suggestion -- would the VTK filters work outside of the visualization context?

kesshijordan:
@dmreagan could you comment on this?

dmreagan:
On the vis side, we were thinking of a way to filter out streamlines that do not pass through a ROI. After looking into it, it doesn't seem like there is currently a VTK filter that does exactly what we were imagining. I think we could get something working, but I doubt it'll happen soon. Regardless, it likely would be of limited value beyond the visualization context

dipy/tracking/tests/test_utils.py
```diff
@@ -292,6 +293,60 @@ def _target(target_f, streamlines, voxel_both_true, voxel_one_true,
...
+
+    roi5 = np.zeros([20, 20, 20])
+    roi5[1:2, 0:5, 0:5] = 1
+    new5 = list(clip_streamlines_to_target(streamlines, roi5, affine=aff))
+    assert_array_equal([len(i) for i in new5], np.array([4, 4, 4]))
+
+    # check value error for streamline outside of target mask dimensions
+    bad = clip_streamlines_to_target(streamlines_outside, roi1, affine=aff)
+    assert_raises(ValueError, list, bad)
+
```
arokem: Please add tests with an affine that isn't `np.eye(4)`

skoudoro:
ok, this PR is good to go. merging!

ps: Concerning `ArraySequence` support, it is ok if you pass `ArraySequence.streamlines` 

A better support can be done, with `StatefulTractogram` and `ArraySequence` but it should be done more generally in this whole `utils` file (in a new PR).

PR #1621, opened 2018-08-21
workflows : warn user for strange b0 threshold 

skoudoro:
I just added a warning when b0 threshold is smaller than all b-values.

I had to duplicate the code many times. It is not the first time. This means a new PR is needed to create a parent class (ReconstWorkflow) and avoid the redundancy.

### DIFF
--- a/dipy/workflows/reconst.py
+++ b/dipy/workflows/reconst.py
@@ -118,7 +118,11 @@ class ReconstMAPMRIFlow(Workflow):
             data = img.get_data()
             affine = img.affine
             bvals, bvecs = read_bvals_bvecs(bval, bvec)
-
+            if b0_threshold < bvals.min():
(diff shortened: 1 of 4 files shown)


arokem:
Nice. Any chance to get a test of this behavior?

skoudoro:
- What is the good way to test logging module @arokem?

- When we enter into this specific condition, an error is always generated later in the code. It seems I should convert this warning into an error. What do you think @arokem @Garyfallidis?

skoudoro:
I added the test so this one is ready to go @arokem

skoudoro:
Hi @arokem, can you check this one?

Thank you

arokem:
Yep. Ready to go!

skoudoro:
Thanks!

PR #1622, opened 2018-08-22
[Fix] viz slice example

skoudoro:
viz slice example was broken because of the `viz` new grid system so this is just an update.


doc/examples/viz_bundles.py
```diff
@@ -170,8 +170,8 @@
 
 lengths = length(bundle_native)
 
-hue = [0.5, 0.5]  # red only
```
arokem: I am confused by the comment here. This is the same as above, so how could they both be `red only`?

skoudoro: Good catch! should be blue here

PR #1624, opened 2018-08-24
NF: Outlier scoring

kesshijordan:
This is an outlier scoring method that compares the pathways of each streamline in a bundle (pairwise) and scores each streamline by how many other streamlines have similar pathways. I still need to write tests, but any initial feedback would be appreciated.

Here is a notebook demonstrating the use to filter streamlines
https://github.com/kesshijordan/shared_notebooks/blob/master/CCI_Demo.ipynb

Here is a paper about the method
https://onlinelibrary.wiley.com/doi/abs/10.1111/jon.12467


skoudoro:
@pep8speaks Resume now

dipy/tracking/tests/test_utils.py
```diff
@@ -37,6 +37,60 @@ def make_streamlines():
     return streamlines
 
 
+def test_cci():
+    # two identical streamlines should raise an error
+    mysl = np.array([np.arange(10)] * 3).T
+    test_streamlines = list([mysl])*2
```
skoudoro: Due to memory issue, you should avoid having a list of `np.array` as much as possible. This is for this reason, we created `Streamlines` object. Can you replace `list` with `Streamlines`?

dipy/tracking/tests/test_utils.py
```diff
@@ -37,6 +37,60 @@ def make_streamlines():
...
+    assert_raises(ValueError, calculate_cci, test_streamlines)
+
+    # 3 offset collinear streamlines
+    test_streamlines = list([mysl])+[mysl+1]+[mysl+2]
```
skoudoro: same as above

dipy/tracking/tests/test_utils.py
```diff
@@ -37,6 +37,60 @@ def make_streamlines():
...
+    mysl3 = mysl.copy()
+    mysl3[:, 1] = 2
+    mysl4 = mysl.copy()
+    mysl4[:, 1] = 4
+    mysl5 = mysl.copy()
+    mysl5[:, 1] = 5000
+
+    test_streamlines_p1 = list([mysl])+[mysl2]+[mysl3]
+    test_streamlines_p2 = list([mysl])+[mysl3]+[mysl4]
+    test_streamlines_p3 = list([mysl])+[mysl2]+[mysl3]+[mysl5]
```
skoudoro: same as above

dipy/tracking/utils.py
```diff
@@ -921,6 +923,79 @@ def unique_rows(in_array, dtype='f4'):
...
+
+    [Garyfallidis12] Garyfallidis E. et al., QuickBundles a method for
+    tractography simplification, Frontiers in Neuroscience,
+    vol 6, no 175, 2012.
+
+    """
+
+    # error if any streamlines are shorter than 20mm
+    lengths = list(length(streamlines))
+    if np.array(lengths).min() < 20 and not override:
```
skoudoro: I think you do not need to convert it as `np.array`, you could just write `min(lengths)`.

dipy/tracking/utils.py
```diff
@@ -921,6 +923,79 @@ def unique_rows(in_array, dtype='f4'):
...
+
+    cci_score_mtrx = np.zeros([len(subsamp_sls)])
+
+    for i, sl in enumerate(subsamp_sls):
+        mdf_mx = bundles_distances_mdf([subsamp_sls[i]], subsamp_sls)
+        if (1 * mdf_mx == 0).sum() > 1:
+            raise ValueError('Identical streamlines. CCI calculation invalid')
+        mdf_mx_oi = (mdf_mx > 0) & (mdf_mx < max_mdf) & ~ np.isnan(mdf_mx)
+        mdf_mx_oi_only = mdf_mx[mdf_mx_oi]
+        cci_score = np.sum(np.divide(1, np.power(mdf_mx_oi_only, power)))
```
skoudoro: Any chance to have `mdf_mx_oi_only` equal to 0?  if yes, make sure to avoid the `DivisionZeroError`.

dipy/tracking/utils.py
```diff
@@ -921,6 +923,79 @@ def unique_rows(in_array, dtype='f4'):
...
+    # calculate the pairwise MDF distance between all streamlines in dataset
+    subsamp_sls = set_number_of_points(streamlines, subsample)
+
+    cci_score_mtrx = np.zeros([len(subsamp_sls)])
+
+    for i, sl in enumerate(subsamp_sls):
+        mdf_mx = bundles_distances_mdf([subsamp_sls[i]], subsamp_sls)
+        if (1 * mdf_mx == 0).sum() > 1:
+            raise ValueError('Identical streamlines. CCI calculation invalid')
+        mdf_mx_oi = (mdf_mx > 0) & (mdf_mx < max_mdf) & ~ np.isnan(mdf_mx)
```
skoudoro: nice! One day, I should try to see if this syntax is more efficient then using `numpy` function

dipy/tracking/utils.py
```diff
@@ -64,6 +64,8 @@
 from numpy import (asarray, ceil, dot, empty, eye, sqrt)
 from dipy.io.bvectxt import ornt_mapping
 from dipy.tracking import metrics
+from dipy.tracking.streamline import set_number_of_points
```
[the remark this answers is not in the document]
arokem: I think this means that the CCI calculation is not a utility. I would put it in the streamlines module

dipy/tracking/utils.py
```diff
@@ -921,6 +923,79 @@ def unique_rows(in_array, dtype='f4'):
...
+        The number of points that are considered for each streamline
+        in the calculation. To save on calculation time, each
+        streamline is subsampled to subsampleN points.
+    power: int
+        The power to which the MDF distance for each streamline
+        will be raised to determine how much it contributes to
+        the cci. High values of power make the contribution value
+        degrade much faster. Example: a streamline with 5mm MDF
+        similarity contributes 1/5 to the cci if power is 1, but
+        only contributes 1/5^2 = 1/25 if power is 2.
```
arokem: `override` is not documented

skoudoro:
+1 for the naming suggestion

dipy/tracking/utils.py
```diff
@@ -921,6 +923,79 @@ def unique_rows(in_array, dtype='f4'):
...
+                   'To continue with short streamlines set override=True')
+
+    # calculate the pairwise MDF distance between all streamlines in dataset
+    subsamp_sls = set_number_of_points(streamlines, subsample)
+
+    cci_score_mtrx = np.zeros([len(subsamp_sls)])
+
+    for i, sl in enumerate(subsamp_sls):
+        mdf_mx = bundles_distances_mdf([subsamp_sls[i]], subsamp_sls)
+        if (1 * mdf_mx == 0).sum() > 1:
```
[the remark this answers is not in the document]
kesshijordan: I guess I don't... I did that to make sure the boolean is converted to an integer so it gets added correctly but I just tried it and True + True = 2. Thanks!

dipy/tracking/utils.py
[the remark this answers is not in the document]
kesshijordan: Thanks, @skoudoro and @arokem. Could one of you please explain this point a bit? I'm not following.

dipy/tracking/utils.py
kesshijordan: I don't think so... here's my reasoning: 
- the mdf_mx can only be zero when two streamlines are identical (it's the mdf distance, so based on summing euclidean distances between subsampled points on streamlines.. this can't be negative so the only way to get to zero would be if all of the points were identical). 
- Since we raise a ValueError when there are identical streamlines, we should never get to the point where mdf_mx_oi_only since that's a subset of mdf_mx
- I added the ValueError for identical streamlines because, if you think about what the purpose of this outlier calculation is, it's not really telling you anything when streamlines are identical because they would give strong support to a pathway that was probably gained "unfairly"... e.g. 
1) this is being used for deterministic tracking with two seeds that are on top of each other, which wouldn't actually give you confidence in that pathway... only a seed that's a bit further away would be informative in a deterministic system... 2) it's really intended for probabilistic tracking, which shouldn't have this identical streamline situation unless something is wrong with the way the bundle was generated (e.g. duplicated samples), in which case the user needs to address that to avoid getting misleading results)

Does this sound reasonable?

dipy/tracking/utils.py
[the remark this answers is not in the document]
kesshijordan: I see. Thank you for explaining, @arokem. So this is a decision made about the structure of the tracking module as a whole. Utils gets imported by other tracking sub-modules but should only import from other modules (unless it's from `_utils` since that is lower on the hierarchy). I think this explains some other things I've been puzzled about debugging imports... I might be an offender with respect to this point so let me know if I can help fix the module if it's decided to enforce the hierarchy  : )

dipy/tracking/tests/test_utils.py
kesshijordan: Thanks, @skoudoro.

Something I noticed while fixing this: you get an error if the type isn't double (my toy streamlines in tests were initially integers)... I just fixed it in the test to make the toys floats. Can we assume that inputs to set_number_of_points will always be floats or is this an issue?

dipy.tracking.streamlinespeed.set_number_of_points()
ValueError: Buffer dtype mismatch, expected 'double' but got 'long'

kesshijordan:
Thanks so much for all your help, @skoudoro or @arokem! I addressed most of your comments (though I still need to write the example and finish formatting, see below). 

I have a few formatting questions:
- Do you have a suggestion for the best way to screen for pep8 errors? I have flake8 in my editor and I added autopep8 (Atom), but it isn't catching what @skoudoro pointed out. It catches missing spaces after commas, but not around operators. What are you using, @skoudoro?
- I'm getting a Flake8 warning on some things that are imported but not used, but they are not part of my updates so I want to make sure I understand before changing anything (run_module_suite in test_streamline, import nose in test_utils, setup_test in utils, setup_test & compress_streamlines in streamline) Is it safe to delete these or are they there for a reason?

dipy/tracking/utils.py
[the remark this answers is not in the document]
kesshijordan: I took your suggestion, @arokem, and moved the function to the streamlines sub-module in tracking

kesshijordan:
I had forgotten to raise the streamline length ValueError... I wanted to get the last line of test coverage so I found the error!

dipy/tracking/tests/test_utils.py
skoudoro: Indeed, I think this is an issue. But I would like to have @Garyfallidis opinion.

dipy/tracking/utils.py
skoudoro: yep, good for me. Thank you for your explanation

kesshijordan:
No problem, @skoudoro. Thanks for all the time you put into reviewing my code! I added a tutorial and rebased on master.

dipy/tracking/streamline.py
```diff
@@ -450,6 +451,83 @@ def select_by_rois(streamlines, rois, include, mode=None, affine=None,
...
+                         ' override=True')
+
+    # calculate the pairwise MDF distance between all streamlines in dataset
+    subsamp_sls = set_number_of_points(streamlines, subsample)
+
+    cci_score_mtrx = np.zeros([len(subsamp_sls)])
+
+    for i, sl in enumerate(subsamp_sls):
+        mdf_mx = bundles_distances_mdf([subsamp_sls[i]], subsamp_sls)
+        if (1 * mdf_mx == 0).sum() > 1:
```
skoudoro: Finally, you keep this multiplication by 1. Why ?

doc/examples/cluster_confidence.py
```diff
@@ -0,0 +1,188 @@
...
+"""
+Now we calculate the Cluster Confidence Index using the corpus callosum
+streamline bundle and visualize them.
+"""
+
+
+cci = cluster_confidence(long_streamlines)
+
+# Visualize the streamlines, colored by cci
+ren = window.renderer()
```
skoudoro: Capital *R* here.  `window.Renderer()` (class object) instead of `window.renderer()` (deprecated function)

doc/examples/cluster_confidence.py
```diff
@@ -0,0 +1,188 @@
...
+streamlines = Streamlines(streamlines)
+
+
+"""
+We do not want our results inflated by short streamlines, so we remove
+streamlines shorter than 40mm prior to calculating the CCI.
+"""
+
+lengths = list(length(streamlines))
+long_streamlines = []
```
skoudoro: Can replace `long_streamlines = []` by `long_streamlines = Streamlines()`? In general, try to avoid list for streamlines object. They work like a list but improve memory management.

doc/examples/cluster_confidence.py
```diff
@@ -0,0 +1,188 @@
...
+.. figure:: cci_histogram.png
+   :align: center
+
+   Histogram of Cluster Confidence Index values.
+
+Now we threshold the CCI, defining outliers as streamlines that score below 1.
+
+"""
+
+keep_streamlines = []
```
skoudoro: same as above: Can replace `keep_streamlines = []` by `keep_streamlines = Streamlines()`?

doc/examples/cluster_confidence.py
```diff
@@ -0,0 +1,188 @@
...
+for i, sl in enumerate(long_streamlines):
+    if cci[i] >= 1:
+        keep_streamlines.append(sl)
+
+# Visualize the streamlines we kept
+ren = window.renderer()
```
skoudoro: Same as above, capital *R* here.  `window.Renderer()` (class object) instead of `window.renderer()` (deprecated function)

skoudoro:
Any chance to fix this quick request and rebase this PR @kesshijordan? it would be great to see this PR on the next release this month. Thanks !

kesshijordan:
what's the release date, @skoudoro?

skoudoro:
we do not have a date yet, but 100% sure this month. Sounds like last week of October...

dipy/tracking/streamline.py
kesshijordan: I did that to change the boolean to an integer for summing... tested on ipython and not necessary... True sums like 1 : ) Thanks for the pointer, @skoudoro !

skoudoro:
Awesome! Thank you for this update @kesshijordan! 

I will merge this as soon as Travis and Appveyor are happy. So, if someone wants to add a comment, you have 1h

arokem:
+1 for the merge!

kesshijordan:
@skoudoro the travis error is in the forecast model, which I didn't touch in this PR. Is this a known issue?

FAIL: dipy.reconst.tests.test_forecast.test_forecast_positive_constrain
    self.test(*self.arg)
  File "/home/travis/build/nipy/dipy/venv/lib/python3.5/site-packages/numpy/testing/_private/decorators.py", line 155, in skipper_func
    return f(*args, **kwargs)
  File "/home/travis/build/nipy/dipy/venv/lib/python3.5/site-packages/dipy/reconst/tests/test_forecast.py", line 53, in test_forecast_positive_constrain
    assert_almost_equal(fodf[fodf < 0].sum(), 0, 2)
  File "/home/travis/build/nipy/dipy/venv/lib/python3.5/site-packages/numpy/testing/_private/utils.py", line 584, in assert_almost_equal
    raise AssertionError(_build_err_msg())
AssertionError: 
Arrays are not almost equal to 2 decimals
 ACTUAL: -0.015931951387651192
 DESIRED: 0

skoudoro:
It appears on my PR too, it seems "new" @kesshijordan. I need to look at this because it will block all our PR.

arokem:
I think this is because of the newly released version of cvxpy. See: https://github.com/nipy/dipy/issues/1654

skoudoro:
Hi @kesshijordan , we just fixed the issue, you can rebase your PR and we can merge it after that.

Thank you

kesshijordan:
Thanks, @skoudoro and @arokem! I think the Python27 environment on appveyor is failing because of a copied and pasted non-ASCII dash in a citation. 

I tried this in my Python 3 vs 2.7 environment and it only catches the dash in 2.7. 
https://stackoverflow.com/questions/21639275/python-syntaxerror-non-ascii-character-xe2-in-file

ERROR: Failure: SyntaxError (Non-ASCII character '\xe2' in file c:\projects\dipy\dipy\tracking\streamline.py on line 517, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details (streamline.py, line 516))

dipy/tracking/streamline.py
```diff
@@ -469,6 +470,83 @@ def select_by_rois(streamlines, rois, include, mode=None, affine=None,
...
+        though there are short streamlines in the dataset that may alter
+        expected behaviour.
+
+    Returns
+    -------
+    Returns an array of CCI scores
+
+    References
+    ----------
+    [Jordan17] Jordan K. Et al., Cluster Confidence Index: A Streamline‐Wise
```
arokem: Something about this reference is wonky and throwing 27 off. Could you rewrite it from scratch by typing it manually into the file, instead of copy-paste?

kesshijordan: thanks, @arokem. There was a rogue dash, but I replaced it and then 27 passed.

arokem: 🎉

skoudoro:
Thank you for all the hard work @kesshijordan!  and thank you for the review @arokem! merging! 🎉

kesshijordan:
Thanks, @skoudoro and @arokem for all the work you put into reviews!

PR #1625, opened 2018-08-25
WIP NF: Generate Recobundles Template from Cohort

kesshijordan:
Here is a WIP for my recobundles template generation code. Each template bundle represents the centroids of a group of bundles (e.g. from many subjects).

Demo:
https://github.com/kesshijordan/shared_notebooks/blob/master/Kesh_Templates_newAPI.ipynb

@Garyfallidis and @BramshQamar this is a nicer version of the code for the project I worked on at Brainhack Global

### DIFF
--- a/dipy/segment/templates.py
+++ b/dipy/segment/templates.py
@@ -1,0 +1,88 @@
+from dipy.tracking.streamline import set_number_of_points
+from dipy.align.streamlinear import StreamlineLinearRegistration
+from dipy.segment.clustering import QuickBundles
+from dipy.align.streamlinear import whole_brain_slr
+from dipy.tracking.utils import move_streamlines
+import numpy as np
+
+
+def bundle_to_wb(fixed_wb, moving_wb, moving_bundle):
+    # TODO:implement a 2-step registration in which you align whole brain, do
+    # recobundles, then SLR to that bundle to lock on to the local anatomy
+
+    moved, transform, qb_cents1, qb_cents2 = whole_brain_slr(fixed_wb,
+                                                             moving_wb,
+                                                             verbose=True,
+                                                             progressive=True)
+    bundle_xfmd = moving_bundle.copy().apply_affine(transform).streamline
+    # bundle_xfmd = nib.streamlines.tractogram.Tractogram(moving_bundle,
+    #             affine_to_rasmm = fa_aff).streamlines
+    return moved, transform, bundle_xfmd
+
+
+def make_rb_template(bundle_list, keystone_boi, qb_thresh=5., Nsubsamp=20,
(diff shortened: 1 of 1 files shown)


skoudoro:
Nice @kesshijordan! any chance to have tests ? :-)

kesshijordan:
Thanks, @skoudoro ! Tests are on my to-do list but I wanted to solicit feedback in the meantime  : )

Mainly:
1) Is wholebrain slr fine to put bundles in MNI space or should I add a secondary local registration (same idea as what I suggested for the uncinate to get Recobundles to perform better... essentially do a wholebrain SLR, then a local recobundles, then a secondary SLR with the output of recobundles to lock onto local anatomy)? 
PRO: we could probably do better at matching the MNI wholebrain template locally, especially in high-distortion areas
CON: if someone wanted to generate bundles that aren't well-represented in the MNI wholebrain they might get a weird result. Wholebrain is safer for the general user and it may not matter (seems to be doing fine with my initial tests)
NOTE: I can implement this anyway as a bundle-to-wholebrain registration since I think it would be useful, regardless of whether or not it should be used here

2) What to do with the centroids that make up the template... Intuitively, templates should be stored at a relatively high sampling, but we definitely don't want to carry unnecessarily high sampling for quickbundles and SLR calculations. 
- Is there any reason to sample them at a particular number of points? We could let the user set the output sampling and calculation sampling individually, but I'm not sure it's worth the trouble to figure out how to get a higher-resolution centroid out of quickbundles (I guess we could take the nearest original streamline to the bundle, which would be full-resolution, but that may not represent the whole bundle like the centroid would depending on the clustering parameters)
- At the moment I'm subsampling to 20 which seems to be an okay tradeoff on capturing the shape and not taking too long

kesshijordan:
Here is a comparison to the other atlas, for reference. This is the raw output (prior to outlier rejection or manual QC) representing 20 manually-cleaned healthy controls.

arcuate_comp

skoudoro:
@kesshijordan :  Do you plan to follow up on this work?

kesshijordan:
Hi, @skoudoro . Yes, I do plan to follow-up but it may take some time. If anyone else wants to contribute to speed it up they are welcome to. Is there a release goal you had in mind?

skoudoro:
> Is there a release goal you had in mind?

Dipy 1.0 in July !

> If anyone else wants to contribute to speed it up they are welcome to

It will would be good if @BramshQamar could look at this PR. It can be useful for her work

Thank you for your feedback kesshi

skoudoro:
Hi @BramshQamar!

Do you think I should clean and push this old PR forward or we should just close it. 

Looking forward for your PR. 

ps: any help would be very appreciated if you recommend to push it forward

PR #1627, opened 2018-08-27
WIP - NF - Tracking with Initial Directions and other tracking parameters

gabknight:
WIP - it needs a first review by @StongeEtienne and @jchoude. Does this will do for SET?

The aim of this PR is to add three new optional parameters to `LocalTracking`:
- `initial_directions` :  specify the initial direction for tractography to follow at the initial seed position. If `initial_directions` is `None` (default), `LocalTracking` calls `DirectionGetter.initial_direction(seed_position)` to get the initial directions (current behavior). This functions return the list of peak at the seed position. In a similar fashion, `initial_directions` contains a list of directions for every seed positions.
  - The tracking works with directions that correspond to vertices on the sphere. Thus, I had to select the vertice the closest to the input direction instead of the actual input direction. i.e. the closest sphere vertice is used as initial direction instead of the given initial directions.
- `unidirectional`: perform the tractography in a single direction from the seeding position, instead of both (default). This is especially useful in conjunction with `initial_direction`, where, for instance, those initial directions could be pointing inward from the cortical surface or outward from subcortical regions.
- `randomized_forward_direction`: Useful in conjunction with `unidirectional`, this parameter allows the tracking to start randomly either in the initial direction or in its opposite direction. If `undirectional is False`, this parameter will only reverse the points order of some streamlines.

### DIFF
--- a/dipy/direction/closest_peak_direction_getter.pxd
+++ b/dipy/direction/closest_peak_direction_getter.pxd
@@ -10,7 +10,6 @@ cdef int closest_peak(np.ndarray[np.float_t, ndim=2] peak_dirs,
 cdef class BaseDirectionGetter(DirectionGetter):
 
     cdef:
-        object sphere
         dict _pf_kwargs
         PmfGen pmf_gen
         double pmf_threshold
--- a/dipy/tracking/local/direction_getter.pxd
+++ b/dipy/tracking/local/direction_getter.pxd
@@ -3,6 +3,9 @@ cimport numpy as np
 
 cdef class DirectionGetter:
 
+    cdef:
+        public object sphere
+
     cpdef np.ndarray[np.float_t, ndim=2] initial_direction(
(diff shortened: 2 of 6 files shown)


gabknight:
@StongeEtienne, do you think those tracking parameters will do for your future mesh seeding integration?

guaje:
@gabknight I was checking your code and I realized you are importing HemiSphere but you are not using it. Can you please check that?

skoudoro:
What is the status of this PR @gabknight @StongeEtienne?

gabknight:
This needs a bit of work and not required for the next release. let keep this for Dipy 1.1

skoudoro:
Ok Sounds like plan. Thank you for the feedback.

gabknight:
Closed in favor of PRs #2821 and #2825.

PR #1639, opened 2018-09-13
matched shore computation to brainsuite

mattcieslak:
Hi Dipy community,

@dvya and I went through the Dipy 3dSHORE implementation and compared it to the BrainSuite implementation. We found a difference in the number of bases generated in Dipy and Brainsuite. For example with radial order set to 6 there were 50 bases in Dipy and 72 in BrainSuite. 

In the Merlet 2013 paper right after eq 18 they write 

> For SHORE basis, the radial order is n = 6 and because l is bounded by n, this gives 72 atoms.

However, the documentation for 3dSHORE in Dipy says that a modified version is being used, so maybe the difference is due to the modification?

### DIFF
--- a/dipy/reconst/shore.py
+++ b/dipy/reconst/shore.py
@@ -371,8 +371,8 @@ class ShoreFit():
         c_sh = np.zeros(J)
         counter = 0
 
-        for l in range(0, self.radial_order + 1, 2):
-            for n in range(l, int((self.radial_order + l) / 2) + 1):
+        for n in range(radial_order + 1):
+            for l in range(0, n + 1, 2):
                 for m in range(-l, l + 1):
 
                     j = int(l + m + (2 * np.array(range(0, l, 2)) + 1).sum())
@@ -543,19 +543,19 @@ def shore_matrix(radial_order, zeta, gtab, tau=1 / (4 * np.pi ** 2)):
     theta[np.isnan(theta)] = 0
     F = radial_order / 2
     n_c = int(np.round(1 / 6.0 * (F + 1) * (F + 2) * (4 * F + 3)))
-    M = np.zeros((r.shape[0], n_c))
-
+    #M = np.zeros((r.shape[0], n_c))
+    M = []
     counter = 0
-    for l in range(0, radial_order + 1, 2):
-        for n in range(l, int((radial_order + l) / 2) + 1):
+    for n in range(radial_order + 1):
+        for l in range(0, n + 1, 2):
             for m in range(-l, l + 1):
-                M[:, counter] = real_sph_harm(m, l, theta, phi) * \
+                M.append(real_sph_harm(m, l, theta, phi) * \
                     genlaguerre(n - l, l + 0.5)(r ** 2 / zeta) * \
                     np.exp(- r ** 2 / (2.0 * zeta)) * \
                     _kappa(zeta, n, l) * \
-                    (r ** 2 / zeta) ** (l / 2)
+                    (r ** 2 / zeta) ** (l / 2))
                 counter += 1
-    return M
+    return np.column_stack(M)
 
 
 def _kappa(zeta, n, l):
@@ -585,20 +585,21 @@ def shore_matrix_pdf(radial_order, zeta, rtab):
     theta[np.isnan(theta)] = 0
     F = radial_order / 2
     n_c = int(np.round(1 / 6.0 * (F + 1) * (F + 2) * (4 * F + 3)))
-    psi = np.zeros((r.shape[0], n_c))
+    #psi = np.zeros((r.shape[0], n_c))
+    psi = []
     counter = 0
-    for l in range(0, radial_order + 1, 2):
-        for n in range(l, int((radial_order + l) / 2) + 1):
+    for n in range(radial_order + 1):
+        for l in range(0, n + 1, 2):
             for m in range(-l, l + 1):
-                psi[:, counter] = real_sph_harm(m, l, theta, phi) * \
+                psi.append( real_sph_harm(m, l, theta, phi) * \
                     genlaguerre(n - l, l + 0.5)(4 * np.pi ** 2 *
                                                 zeta * r ** 2) *\
                     np.exp(-2 * np.pi ** 2 * zeta * r ** 2) *\
                     _kappa_pdf(zeta, n, l) *\
                     (4 * np.pi ** 2 * zeta * r ** 2) ** (l / 2) * \
-                    (-1) ** (n - l / 2)
+                    (-1) ** (n - l / 2))
                 counter += 1
-    return psi
+    return np.column_stack(psi)
 
 
 def _kappa_pdf(zeta, n, l):
@@ -631,17 +632,18 @@ def shore_matrix_odf(radial_order, zeta, sphere_vertices):
     F = radial_order / 2
     n_c = int(np.round(1 / 6.0 * (F + 1) * (F + 2) * (4 * F + 3)))
     upsilon = np.zeros((len(sphere_vertices), n_c))
+    upsilon = []
     counter = 0
-    for l in range(0, radial_order + 1, 2):
-        for n in range(l, int((radial_order + l) / 2) + 1):
+    for n in range(radial_order + 1):
+        for l in range(0, n + 1, 2):
             for m in range(-l, l + 1):
-                upsilon[:, counter] = (-1) ** (n - l / 2.0) * \
+                upsilon.append( (-1) ** (n - l / 2.0) * \
                     _kappa_odf(zeta, n, l) * \
                     hyp2f1(l - n, l / 2.0 + 1.5, l + 1.5, 2.0) * \
-                    real_sph_harm(m, l, theta, phi)
+                    real_sph_harm(m, l, theta, phi))
                 counter += 1
 
-    return upsilon
+    return np.column_stack(upsilon)
 
 
 def _kappa_odf(zeta, n, l):
@@ -655,13 +657,11 @@ def l_shore(radial_order):
     "Returns the angular regularisation matrix for SHORE basis"
     F = radial_order / 2
     n_c = int(np.round(1 / 6.0 * (F + 1) * (F + 2) * (4 * F + 3)))
-    diagL = np.zeros(n_c)
-    counter = 0
-    for l in range(0, radial_order + 1, 2):
-        for n in range(l, int((radial_order + l) / 2) + 1):
+    diagL = []
+    for n in range(radial_order + 1):
+        for l in range(0, n + 1, 2):
             for m in range(-l, l + 1):
-                diagL[counter] = (l * (l + 1)) ** 2
-                counter += 1
+                diagL.append( (l * (l + 1)) ** 2)
 
     return np.diag(diagL)
 
@@ -671,12 +671,11 @@ def n_shore(radial_order):
     F = radial_order / 2
     n_c = int(np.round(1 / 6.0 * (F + 1) * (F + 2) * (4 * F + 3)))
     diagN = np.zeros(n_c)
-    counter = 0
-    for l in range(0, radial_order + 1, 2):
-        for n in range(l, int((radial_order + l) / 2) + 1):
+    diagN = []
+    for n in range(radial_order + 1):
+        for l in range(0, n + 1, 2):
             for m in range(-l, l + 1):
-                diagN[counter] = (n * (n + 1)) ** 2
-                counter += 1
+                diagN.append( (n * (n + 1)) ** 2 )
 
     return np.diag(diagN)
 
@@ -762,7 +761,7 @@ def shore_indices(radial_order, index):
     n_i = 0
     l_i = 0
     m_i = 0
-
+    """
     if n_c < (index + 1):
         msg = "The index %s is higher than the number of" % index
         msg += " coefficients of the truncated basis,"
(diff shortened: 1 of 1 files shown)


skoudoro:
Thank you @mattcieslak and @mattcieslak!

I think this is @maurozucchelli who implemented this modified version of SHORE so it will be good if you could answer the @mattcieslak's question. Thank you!

mattcieslak:
This definitely breaks some of the unit tests. 

I see the differences from Merlet are described in the appendix of @maurozucchelli's 2016 paper. Is this just a case of differences in the interpretation of what radial_order should mean?

arokem:
I don't understand the issue deeply, but if it is indeed a difference in interpretation, I think that it would be great to have both options possible. I would make the current implementation the default, and then add a kwarg to the model object that allows to use this interpretation instead.

arokem:
Also, tiny detail: is there a reason you forgo the preallocations that were there before?

mattcieslak:
the n_c was evaluating to the truncated basis set (50) instead of the full set (72). I can change the calculation so it matches the new loops. Funny thing is the brainsuite code used n_c too, but since it's matlab it just kept appending as more than n_c values were calculated.

Do you think it makes sense to make a separate class? Like a `BrainSuiteShore` or `MerletShore` or something? It would need different methods to calculate all the rtop/rtap/etc and adding lots of ifs in the functions makes life messier.

arokem:
No - I think this is better handled through key-word arguments to one class that implements either.

dipy/reconst/shore.py
```diff
@@ -762,23 +761,24 @@ def shore_indices(radial_order, index):
...
+    """
+    counter = 0
+    for n in range(radial_order + 1):
+        for l in range(0, n + 1, 2):
+            for m in range(-l, l + 1):
+                if counter == index:
+                    n_i = n
+                    l_i = l
+                    m_i = m
+                counter += 1
```
nilgoyette: I know nothing about this code but it looks like you could find `n_i`, `l_i` and `m_i` using /, mod and -. No?

dipy/reconst/shore.py
```diff
@@ -817,8 +817,8 @@ def shore_order(n, l, m):
         counter_i = 0
 
         counter = 0
-        for l_i in range(0, radial_order + 1, 2):
-            for n_i in range(l_i, int((radial_order + l_i) / 2) + 1):
+        for n_i in range(radial_order + 1):
+            for l_i in range(0, n + 1, 2):
                 for m_i in range(-l_i, l_i + 1):
                     if n == n_i and l == l_i and m == m_i:
                         counter_i = counter
```
nilgoyette: I know it's not your code, but I would ask the same question: can `counter_i` be found using simple math?

mattcieslak: Yes in both cases. This looping is mostly unnecessary and I've replaced it in a local commit that I haven't pushed yet. 

It turned out that in order to totally match BrainSuite I needed to add another spherical harmonics function.  Their definition of spherical harmonics (from here https://www.ncbi.nlm.nih.gov/pubmed/23353603) and their Merlet-compatible definition of the SHORE basis has some really nice properties when fitting real data. 

I haven't pushed these changes yet because their code is released under a GPLv2 license and that's not going to work with Dipy. Not sure what the best way forward is. The BrainSuite authors might agree to letting a python version of these two functions be released under a more permissive license. 

I also added the L1 optimization from the Merlet paper, which has performs very impressively.

mattcieslak: You have to register on their website and then you can download an archive of their matlab code. It's all GPLv2. @dvya is one of the authors

arokem: Thanks! Unfortunately, you're right and the GPL is a non-starter for us :-( 

Unless what we have here is demonstrably wrong, I would recommend making this external to Dipy itself. You can grab our code and put it in a separate GPL-licensed repository. I'd be happy to help set that up.

One thing that I don't entirely understand: Is what we currently have flat-out wrong? Or is it really a difference of interpretation?

mattcieslak: It's not wrong, but also definitely not what's in the Merlet paper. At the minimum we might want change the docstrings so that they just reference the Zucchelli paper. I see differences in performance on synthetic datasets and the two methods have very different EAP response functions, with dipy's having considerably more artefact. 

If we get written permission from the authors to release a python implementation under a more permissive license, would that be ok?

dvya: Hello - Sorry for chiming in so late (in between moves). So here is some technical background as to why we started this pull request in the first place.  All these basis functions are continuous functions so we have to truncate and approximate it while implementing. So what is currently in dipy is one approximation and it just leaves out more bases functions. And as Matt pointed out the reference is actually Zucchelli's paper. What BrainSuite has currently implemented is Merlet's l2 version which has more bases and we can show theoretically (was demonstrated at Brainhack 2018 - Varadarajan et al Neuroimage 2017) that it has better characteristics. (more accurate). 

So basically at Brainhack we discussed starting this pull request  and bringing this version to dipy.

dvya: Regarding licenses - Can you please tell me exactly what information I need to get from BrainSuite to get this version of shore into dipy ?   If we cannot implement it for dipy then having a separate toolbox might not make sense because we have the implementation in BrainSuite - which is an open source toolbox. 

This is also important because I have a few of my methods implemented for dipy (testing ongoing), but these methods will be a part of next BrainSuite release.

Garyfallidis: Hi all, and terribly sorry for being late to respond to this.  @mattcieslak and @dvya I do not think that we really need written permission as the same developer @dvya is providing the code to both software and the change is small. Also @dvya has already talked about this to the core devs of Brainsuite. If more algorithms are on the way then what Ariel suggests makes sense. 

Let me know if this is clear.

Garyfallidis: Also @mattcieslak  definitely agree with @arokem with the class decision. Prefer to be one class if possible with two options new and old.

Garyfallidis:
@mattcieslak please also report speed difference between the two versions. Thank you! Really appreciate your help. :)

arokem:
Hey @mattcieslak: would be good to pick this back up. How do you feel about providing both options? This would be similar to what we do with DTI, so that people can fit it with different options: least-squares, weighted least-squares, RESTORE, and so on.

skoudoro:
Hi @mattcieslak, it would be good to have this feature for DIPY 1.0 release (mid July). 
Any chance to have an update with both options and a small test?
Let us know if you need any help to make it happen!

mattcieslak:
I think this is doable. I don't think it should be an option in the currently-implemented 3dSHORE class, though. Unlike DTI, where the many ways to fit always result in 6 coefficients that are used the same way for FA/MD/etc, the BrainSuiteShore functions for rtop/rtap/odf/pdf/etc would all need to be changed to check if the instance is using the Zucchelli bases or the BrainSuite bases. It will be easier to maintain and read if it's just a separate class. Would that work for you?

arokem:
Yes. That makes sense. It seems more similar to the distinction between fwdti and dti than between different ways of fitting dti.

arokem:
Hey @mattcieslak : is this ready for a review or are you still working on it?

mattcieslak:
It's almost there. I am working on the tests and getting rtop/rtap/etc. Should be ready in a couple days

skoudoro:
Hi @mattcieslak, just for your information, we plan to release this Wednesday (31rst). Let us know if you find time to complete this PR for a review. Thanks!

mattcieslak:
Sorry @skoudoro I'm still waiting to get the equations for rtop, rtap, etc for these bases. I'm not sure I'll have it ready to go in time :(

skoudoro:
ok, no worries and no problem! I will tag your PR for the release 1.1 in October.

dipy/reconst/brainsuite_shore.py
```diff
@@ -0,0 +1,668 @@
+from __future__ import division
+import warnings
```
Borda: in general, it is better to split the import into three visual groups
1. build-in packages
2, third party packages (external)
3. import from this package (internal)

dipy/reconst/brainsuite_shore.py
```diff
@@ -0,0 +1,668 @@
...
+from scipy.special import genlaguerre, gamma, hyp2f1
+from dipy.reconst.cache import Cache
+from dipy.reconst.multi_voxel import multi_voxel_fit
+from .shm import real_sym_sh_brainsuite
+from dipy.core.geometry import cart2sphere
+from dipy.utils.optpkg import optional_package
+
+cvxpy, have_cvxpy, _ = optional_package("cvxpy")
+sklearn, have_sklearn, _ = optional_package("sklearn")
+if have_sklearn:
```
Borda: what happens if there is no `sklearn`, whole fitting crashes? there is no alternative assignment of these functions...

dipy/reconst/brainsuite_shore.py
```diff
@@ -0,0 +1,668 @@
...
+        Image Analysis, 2013.
+
+        Examples
+        --------
+        In this example, where the data, gradient table and sphere tessellation
+        used for reconstruction are provided, we model the diffusion signal
+        with respect to the SHORE basis and compute the real and analytical
+        ODF.
+
+        from dipy.data import get_data,get_sphere
```
Borda: is this doctests example? I would say it should be... reformat to `>>> `

dipy/reconst/brainsuite_shore.py
```diff
@@ -0,0 +1,668 @@
...
+
+    def _l_shore(self):
+        ell = self.ind_mat[:, 1]
+        return np.diag((ell * (ell + 1))**2)
+
+    @multi_voxel_fit
+    def fit(self, data):
+
+        # Generate the SHORE basis
+        M = self.cache_get('shore_matrix', key=self.gtab)
```
Borda: Capital variables are not very supported in PEP8

dipy/reconst/brainsuite_shore.py
```diff
@@ -0,0 +1,668 @@
...
+        if M is None:
+            M = brainsuite_shore_basis(
+                self.radial_order, self.zeta, self.gtab, self.tau)
+            self.cache_set('shore_matrix', self.gtab, M)
+        MpseudoInv = self.cache_get('shore_matrix_reg_pinv', key=self.gtab)
```
Borda: Camel case is PEP8, use rather `mtx_pseudi_inv`

dipy/reconst/brainsuite_shore.py
```diff
@@ -0,0 +1,668 @@
...
+        self._alpha = alpha
+        self._r2 = r2
+        self._cnr = cnr
+        self._regularization = regularization
+        self.gtab = model.gtab
+        self.radial_order = model.radial_order
+        self.zeta = model.zeta
+
+    def pdf_grid(self, gridsize, radius_max):
+        r""" Applies the analytical FFT on $S$ to generate the diffusion
```
Borda: why the r""" here and not everywhere?

dipy/reconst/shm.py
```diff
@@ -240,6 +240,74 @@ def real_sph_harm(m, n, theta, phi):
...
+    # Adapted from BrainSuite's sph_harm_basis
+    ndirs = len(phi)
+    nbases = int(1 / 2 * (sh_order + 1) * (sh_order + 2))
+    S = np.zeros((ndirs, nbases))
+    S[:, 0] = 1 / np.sqrt(4 * np.pi)
+    L = np.zeros(nbases)
+    Z = np.zeros(nbases)
+    for ell in range(2, sh_order + 1, 2):
+        span = np.arange(ell + 1)
+        Pell = (np.ones((ndirs, ell + 1))
```
Borda: rather compute the elements and then join then having such complicated formulation in a single line...

dipy/reconst/shm.py
```diff
@@ -240,6 +240,74 @@ def real_sph_harm(m, n, theta, phi):
...
+        Pell = np.column_stack([
+            np.sqrt(2) * np.real(Pell[:, 1:])[:, ::-1],
+            Pell[:, 0],
+            np.sqrt(2) * (np.ones(ndirs)[:, None]*(-1) ** span2)
+            * np.imag(Pell[:, 1:])])
+        start_index = int(ell * (ell - 1) / 2)
+        end_index = int((ell + 1) * (ell + 2) / 2)
+        S[:, start_index:end_index] = Pell
+        L[start_index:end_index] = ell
+        Z[start_index:end_index] = np.arange(-ell, ell+1)
```
Borda: spacing, use `flack8` or `pylint`

Borda:
@skoudoro maybe you may remove tag "need review" since there already a couple of reviewers :)

skoudoro:
The tag has already been removed but Thank you for the notice! :)

skoudoro:
Hi @mattcieslak, 

> I'm still waiting to get the equations for rtop, rtap, etc for these bases

Any updates concerning this PR? Did you get the equations? How can we help?

mattcieslak:
@skoudoro do you know any of the brainsuite developers? Their message board was a dead end

skoudoro:
Hi @mattcieslak,

We plan to talk about your PR during the DIPY online meeting tomorrow (October 7th, 2021 | 1pm EST / 7pm CET / 10am PT).

It would be great if you could attend this meeting to provide us a short update. 

More information on https://github.com/dipy/dipy/discussions/2398

Thank you!

PR #1641, opened 2018-09-17
changed vertices to float64 in evenly_distributed_sphere_642.npz

mattcieslak:
The npz file for the symmetric642 sphere caused a value error when used for `peaks_from_model`. 

```
ValueError: Buffer dtype mismatch, expected 'float_t' but got 'float'
```

I changed the datatype to float64 and now it works.


arokem:
Thanks for doing this! Would you mind adding a test that fails without this change, so it doesn't creep back in?

arokem:
Could you take a look at #1642? Shouldn't that fail due to this issue? It passes on my machine (but let's also see what Travis says...).

mattcieslak:
All the other npz files had float64 vertices. The only sphere with float32's that caused errors came from ``get_sphere('symmetric642')``, which I believe is the default sphere used in DSI Studio. ``get_sphere('symmetric724')`` has always worked for me.

arokem:
Oh - gotcha! I was looking at the wrong sphere... 

I am not arguing that we shouldn't change this, but I worry that there might be other things afoot (probably not). See: https://github.com/nipy/dipy/pull/1642/commits/0ba1b635528218135cd975b056fba30feaf26a6e

That still passes on my machine. Are you doing more than that with your PAM?

mattcieslak:
I'm not doing anything with the pam yet, just trying to figure out how saving data works in dipy. Working on writing a pam5 to dsi studio converter

skoudoro:
Any other comment? I wait 24h before merging this PR and #1642 

> Working on writing a pam5 to dsi studio converter

look at [this test](https://github.com/nipy/dipy/blob/c8d2f179dac618d9249c82a0053327f1edf1700b/dipy/io/tests/test_io_peaks.py), maybe it can help you

mattcieslak:
I don't see how it could run with float32 data as soon as it gets to here https://github.com/nipy/dipy/blob/master/dipy/reconst/recspeed.pyx#L207, so I think this PR won't cause anything to break. I added some tests, too.

Thanks for the link @skoudoro!

arokem:
That probably means that the tests for PAM are not sufficiently exercising the code. Otherwise, #1642 
 should've failed. But yeah, go ahead and merge this.

PR #1642, opened 2018-09-17
Test that you can use the 724 symmetric sphere in PAM.

arokem:
Related to #1641


skoudoro:
maybe a hidden cast... I don't know where yet....

PR #1643, opened 2018-09-18
Add hash for SCIL b0 file

arokem:


dipy/data/fetcher.py
```diff
@@ -371,7 +371,7 @@ def fetcher():
     UW_RW_URL + "1773/38479/",
     ['datasets_multi-site_all_companies.zip'],
     ['datasets_multi-site_all_companies.zip'],
-    None,
+    "e9810fa5bf21b99da786647994d7d5b7",
```
skoudoro: MD5 should be in a list, it misses the bracket here. Otherwise you get the following error: 
```
FetcherError: The downloaded file, /Users/koudoro/.dipy/datasets_multi-site_all_companies.zip, does not have the expected md5
   checksum of "e". Instead, the md5 checksum was: "e9810fa5bf21b99da786647994d7d5b7". This could mean that
   something is wrong with the file or that the upstream file has been updated.
   You can try downloading the file again or updating to the newest version of dipy.
```

skoudoro:
Thank you @arokem for this fix!

PR #1650, opened 2018-10-16
Increase codecov tolerance

skoudoro:
Drops on the order 0.01% are typical even when no change occurs. So, Having this threshold set a little higher (0.1%) than that makes it a little more tolerant to fluctuations.


arokem:
LGTM. Merging.

PR #1651, opened 2018-10-16
Add save/load tck

skoudoro:
This PR just add:  
[x] `load_tck` function
[x] `save_tck`  function
[x] `lazy_load` option
[x] `lazy_save` option via `LazyTractogram`

In the end, this is an implementation of @gabknight [comment](https://github.com/nipy/dipy/pull/1379#issuecomment-346350769)


dipy/io/streamline.py
```diff
@@ -20,6 +24,11 @@ def save_trk(fname, streamlines, affine, vox_size=None, shape=None, header=None)
         The shape of the reference image (default: None)
     header : dict, optional
         Metadata associated to the tractogram file(*.trk). (default: None)
+    lazy_save : {False, True}, optional
+        If True, save streamlines in a lazy manner i.e. they will not be kept
+        in memory. Otherwise, load all streamlines in memory.
```
jchoude: `load` -> `keep all streamlines in memory until saving`.

dipy/io/streamline.py
```diff
@@ -20,6 +24,11 @@ def save_trk(fname, streamlines, affine, vox_size=None, shape=None, header=None)
         The shape of the reference image (default: None)
     header : dict, optional
         Metadata associated to the tractogram file(*.trk). (default: None)
+    lazy_save : {False, True}, optional
```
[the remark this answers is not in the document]
skoudoro: When you have a generator with a huge amount of streamlines, `nib.streamlines.Tractogram(streamlines)` put all the streamlines in memory before saving them and lead to a crash if you do not have enough memory.
By using the `lazy_save, you use `nib.streamlines.LazyTractogram(streamlines)` and this object keep your generator as it is and save your streamlines without any problem.

dipy/io/streamline.py
```diff
@@ -50,5 +68,20 @@ def load_trk(filename):
     hdr : dict
         header from a trk file
     """
-    trk_file = nib.streamlines.load(filename)
+    trk_file = nib.streamlines.load(filename, lazy_load)
     return trk_file.streamlines, trk_file.header
+
+
+load_tck = load_tractogram
```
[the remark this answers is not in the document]
skoudoro: I think just leave here for the moment. but any other opinion is welcomed

dipy/io/streamline.py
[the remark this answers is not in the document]
Garyfallidis: Yes, we should rename this variable. It means that it saves a few streamlines each time. Uses less memory.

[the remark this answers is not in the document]
skoudoro: I think `lazy_save`or `save_with_generator ` are not explicit enough. What do you think of `reduce_memory_usage`?

dipy/io/streamline.py
```diff
@@ -20,7 +27,23 @@ def save_trk(fname, streamlines, affine, vox_size=None, shape=None, header=None)
         The shape of the reference image (default: None)
     header : dict, optional
         Metadata associated to the tractogram file(*.trk). (default: None)
+    reduce_memory_usage : {False, True}, optional
```
arokem: Even if we go with this name (which I think is fine), I'd still want to make sure that users understand that this will only save memory if they provide an input that is a generator (that's right?), so I would add that to the description.

skoudoro: > will only save memory if they provide an input that is a generator (that's right?)

No, whatever the input streamlines object, we do the job under the hood [here](https://github.com/nipy/dipy/pull/1651/files#diff-58c02ec9366478aebd7f3ab2c80a62a6R56) and the user will save memory

skoudoro: by saving memory, I mean, it will not increase like a hell 😄

arokem: But if `streamlines` is a list on input, that wouldn't matter. It would just generate from that list, without saving any memory. Or am I missing something?

skoudoro: If you have a list on input,  `Tractogram` create `Streamlines object` which will create a copy of your list and explode the memory for big data. The `reduce_memory_usage` variable allows the conversion of this list in a generator so no copy and no extra memory.

skoudoro: Each time you create a `Streamlines`object from a list, a copy is done because they are not organize is the same way on the memory

arokem: Gotcha - that makes sense. Is there any performance trade-off? Why would anyone ever want to not use this? I mean, if there's no reason to set this to False, should we just eliminate this as an option?

skoudoro: Good question! no idea about the performance trade-off.  But in general, the generators are less faster

arokem:
OK. Looks good to me. I think this is ready to go in?

PR #1414, opened 2018-02-04
[WIP] Switching tests to pytest and removing nose dependencies

thechargedneutron:
Fixes #1280 

Please review if this is what you intended. I have added separate check function and used it.

PR #1652, opened 2018-10-16
Switching tests to pytest

### LINKED ISSUE #1280: Switch testing to pytest?
I have a feeling we've discussed this before. Nose is no longer being developed, and from what I have heard and experienced myself, it looks like pytest (https://docs.pytest.org/en/latest/) is a good way to go. Looks like maybe nose2 is also an option (http://nose2.readthedocs.io/en/latest/).

skoudoro:
This PR replace `nose` by `pytest`. 

Unfortunately `numpy.testing` depends on `nose` until 1.15 so we need to keep this dependency.  We just do not use it directly anymore

This PR should fix #1280 request and supersede #1414


skoudoro:
Todo:

- [ ] check `setup_module`. it does not work
- [ ] check `setup_test`. it does not work. should be replace by `setup_module`
- [ ] check `+ Skip if` in doctest. it does not work

skoudoro:
Hi @arokem, Pytest is failing because of [this error](https://travis-ci.org/nipy/dipy/jobs/446208261#L6205) which I find it strange because we can not reproduce it with Nosetests. 
I would like your opinion about this error. Thank you

arokem:
Hmmm. Have you been able to reproduce that error locally? I am trying it right now.

arokem:
OK - yeah - I see this locally as well. Will investigate.

arokem:
Hmm. I am now seeing this on master as well, with a new environment I made on my local machine. Did anything else change around this? I think there's a new release of nibabel. Maybe that's causing this somehow?

skoudoro:
>  Did anything else change around this?

Currently, No idea...

arokem:
Hey @skoudoro : would you mind rebasing this on master? I'd like to give this another look.

skoudoro:
Done, sorry for the delay.

arokem:
Looks like that test is still failing on Travis (but not on appveyor?!). But I still cannot replicate that locally... The Travis log is also endlessly growing because of a warning coming out of cvxpy, related to [this issue](https://github.com/cvxgrp/cvxpy/issues/567). I might have to make a PR on cvxpy to resolve that...

skoudoro:
> Looks like that test is still failing on Travis

Unfortunately, exactly the same error....

> (but not on appveyor?!)

because I did not change this file yet... I prefer to fix first Travis...

> I might have to make a PR on cvxpy to resolve that...

great! 👍

arokem:
OK: more information here. I am definitely seeing this error locally, but it's intermittent, happening maybe one out of 10 times that I run it. And I do also see this on master, so this is not a pytest specific thing. See also https://github.com/nipy/dipy/issues/1679

skoudoro:
New update: it seems related to the function `_streamlines_in_mask` which do not return always the same result. 

This function is recursive and I wonder whether there is a memory issue with a list of np.array...

I can not dig into today but I add it on my todo list

skoudoro:
One workaround could be to just use `Streamlines` object in this function and tests

skoudoro:
Finally, it is intermittent because of a `np.random.random`  [here](https://github.com/nipy/dipy/blob/master/dipy/tracking/tests/test_utils.py#L272). 

I think I will just create a static array for this test and it should work. What do you think @arokem?

arokem:
I just ran this on my mac laptop and am glad to report that everything is passing.
 
I still get a lot of PendingDeprecation warnings from cvxpy. I hope they review my PR soon.

skoudoro:
> I just ran this on my mac laptop and am glad to report that everything is passing.

Great! it seems to fail on python2.7 but this is an easy fix. My main concern is about our nose dependence due to `numpy.testing`. Should we just get rid of `numpy.testing` or we just accept it. Numpy drop `Nosetests` on version 1.15 and I can not imagine us having numpy 1.15 as minimal version 😄 which means we will keep nose for still a long time.

In any case, if we have to do this change, it will be for a new PR. What do you think @arokem or devs?

> I still get a lot of PendingDeprecation warnings from cvxpy. I hope they review my PR soon.

I hope so too!

arokem:
My inclination is to say that it's fine for us to keep indirectly depending on nose for testing as long as we still depend on versions of numpy that depend on nose for testing.

arokem:
Good news! Looks like my PRs on cvxpy were merged, so we shouldn't see these warnings anymore once they make a release.

CONTRIBUTING.md
```diff
@@ -39,9 +39,9 @@ branch points. To check how much coverage the tests have, you will need.
 
 When running:
 
-    nosetests --with-coverage --cover-package=dipy
+    coverage run -m pytest -s --doctest-modules --verbose dipy
```
arokem: Also get `Coverage.py warning: --include is ignored because --source is set (include-ignored)` when I run this locally.

Makefile
```diff
@@ -27,7 +27,7 @@ ext: recspeed.so propspeed.so vox2track.so \
     mrf.so
 
 test: ext
-	nosetests .
+	pytest .
```
arokem: Do we want to add the full incantation here?

dipy/conftest.py
```diff
@@ -0,0 +1,27 @@
+"""pytest initialization."""
```
arokem: Does this ever get used by anything? I couldn't find an import of this module anywhere.

dipy/testing/__init__.py
```diff
@@ -11,14 +11,30 @@
...
+    """Check Boolean statement."""
+    try:
+        if op == operator.is_:
+                value1 = bool(value1)
+        assert op(value1, value2)
+    except AssertionError:
+        raise AssertionError(msg.format(str(value2), str(value1)))
+
+
+assert_greater_equal = partial(assert_operator, op=operator.ge,
```
arokem: Are these defined here because pytest doesn't have these? Looks like something that might be useful in numpy?

dipy/workflows/tests/test_reconst_mapmri.py
```diff
@@ -47,39 +45,39 @@ def reconst_mmri_core(flow, lap, pos):
 
         rtop = mmri_flow.last_generated_outputs['out_rtop']
         rtop_data = nib.load(rtop).get_data()
-        eq_(rtop_data.shape, volume.shape[:-1])
+        assert_equal(rtop_data.shape, volume.shape[:-1])
```
arokem: If you want consistent style, you might go with `npt.assert_equal` here, as you had in other files (I think that's better -- more explicit).

doc/installation.rst
```diff
@@ -383,7 +383,7 @@ Testing
 
 If you want to run the tests::
 
-    sudo pip install nose
+    sudo pip install pytest nose
```
arokem: Do you really still need to install both pytest and nose?

scratch/very_scratch/gqsampling_stats.py
```diff
@@ -1,91 +1,89 @@
 import os
 import numpy as np
-from nose.tools import assert_true, assert_false, assert_equal, assert_raises
-from numpy.testing import assert_array_equal, assert_array_almost_equal
+from dipy.testing import assert_true, assert_false
+from numpy.testing import (assert_array_equal, assert_array_almost_equal,
+                           assert_equal, assert_raises)
 import time
-#import dipy.core.reconstruction_performance as rp
+# import dipy.core.reconstruction_performance as rp
```
arokem: Should we remove these commented out imports? They seem obsolete.

arokem: Oh sorry - this is in "very_scratch". I guess we can ignore it :-)

setup.py
```diff
@@ -50,13 +50,14 @@
     # filenames to .c filenames, and we probably don't have the .c files.
     sys.path.insert(0, pjoin(dirname(__file__), 'fake_pyrex'))
     # Set setuptools extra arguments
+    # We need nosetests for numpy.testing<=1.15
     extra_setuptools_args = dict(
-        tests_require=['nose'],
+        tests_require=['pytest', 'nose'],
```
arokem: Same question here: we do still need both nose and pytest?

arokem:
Nice work! This PR reminded me that I have been wanting to clean up "scractch", so I added https://github.com/nipy/dipy/issues/1713 to keep that in mind.

dipy/conftest.py
skoudoro: `Pytest` look for many setup files implicitly. `conftest.py` is one of them. more info [here](https://docs.pytest.org/en/2.7.3/plugins.html)

arokem:
I am happy to report that it's all passing on my laptop. We still do have a lot of warnings, but we should take those on separate PRs.

dipy/conftest.py
```diff
@@ -0,0 +1,27 @@
...
+ References
+-----------
+https://github.com/numpy/numpy/commit/710e0327687b9f7653e5ac02d222ba62c657a718
+https://github.com/numpy/numpy/commit/734b907fc2f7af6e40ec989ca49ee6d87e21c495
+https://github.com/nipy/nibabel/pull/556
+"""
+if LooseVersion(np.__version__) >= LooseVersion('1.14'):
+    np.set_printoptions(legacy='1.13')
+# Temporary fix until scipy release in October 2018
+# must be removed after that
```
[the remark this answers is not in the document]
skoudoro: yes, this refers to #1601

dipy/testing/__init__.py
skoudoro: `Numpy` and `Pytest` do not have them. they use `assert x > y`, but the error message is never explicit, it returns `True or False`. And yes, it can be useful for them.

dipy/workflows/tests/test_reconst_mapmri.py
skoudoro: I agree,

I will update that

doc/installation.rst
skoudoro: Unfortunately yes...  Until we have Numpy 1.15 as a minimal version.  `numpy.testing` depends on `nose` until that version

setup.py
skoudoro: Same answer: Unfortunately yes...  Until we have Numpy 1.15 as a minimal version.  `numpy.testing` depends on `nose` until that version

arokem: Gotcha. Yeah - I remember we discussed this before. Seems fine for now.

dipy/conftest.py
arokem: Huh. Interesting.

dipy/conftest.py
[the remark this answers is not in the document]
arokem: So should we remove this?

arokem:
OK. This looks good to me. Considering that it is a fairly crucial and sensitive transition, I think that it would be really good to get another review of this, before we consider merging it.

arokem:
Does anyone else have time to take a look through this? It's all fairly straightforward, but because it's our test suite, I think that it would be a good idea to have one more set of eyes on this, before we merge this. If I don't hear anything in the next week or so, I'll assume everyone is just too busy and I'll merge this as is.

arokem:
OK. Looks like no one else has the bandwidth to take a look, so I will go ahead and merge this. We'll want to keep and eye out for things that should fail and don't, or vice versa.

skoudoro:
Exciting! Thanks @arokem!

PR #1653, opened 2018-10-16
STYLE: Honor 'descoteaux' SH basis naming.

jhlegarreta:
Honor 'descoteaux' naming for SH basis: replace and deprecate 'fibernav'.

Use the author's name rather than the software/implementation's.


jhlegarreta:
The BIDS draft is already proposing this change.

The use of the former `fibernav` may need some deprecation cycle. Suggestions on this are welcome.

dipy/direction/peaks.py
```diff
@@ -423,10 +423,10 @@ def peaks_from_model(model, data, sphere, relative_peak_threshold,
     sh_order : int, optional
         Maximum SH order in the SH fit.  For `sh_order`, there will be
         ``(sh_order + 1) * (sh_order + 2) / 2`` SH coefficients (default 8).
-    sh_basis_type : {None, 'mrtrix', 'fibernav'}
-        ``None`` for the default dipy basis which is the fibernav basis,
+    sh_basis_type : {None, 'mrtrix', 'descoteaux'}
```
skoudoro: you should keep `'fibernav'` for the deprecation. just add `'descoteaux'` like you just did

dipy/direction/peaks.py
```diff
@@ -423,10 +423,10 @@ def peaks_from_model(model, data, sphere, relative_peak_threshold,
...
+        ``None`` for the default dipy basis which is the descoteaux basis,
         ``mrtrix`` for the MRtrix basis, and
-        ``fibernav`` for the FiberNavigator basis
+        ``descoteaux`` for the Descoteaux basis
```
skoudoro: same as above

dipy/reconst/csdeconv.py
```diff
@@ -702,8 +702,8 @@ def odf_sh_to_sharp(odfs_sh, sphere, basis=None, ratio=3 / 15., sh_order=8,
         array of odfs expressed as spherical harmonics coefficients
     sphere : Sphere
         sphere used to build the regularization matrix
-    basis : {None, 'mrtrix', 'fibernav'}
-        different spherical harmonic basis. None is the fibernav basis as well.
+    basis : {None, 'mrtrix', 'descoteaux'}
```
skoudoro: same as above

dipy/reconst/shm.py
```diff
@@ -861,10 +862,10 @@ def sf_to_sh(sf, sphere, sh_order=4, basis_type=None, smooth=0.0):
     sh_order : int, optional
         Maximum SH order in the SH fit.  For `sh_order`, there will be
         ``(sh_order + 1) * (sh_order_2) / 2`` SH coefficients (default 4).
-    basis_type : {None, 'mrtrix', 'fibernav'}
+    basis_type : {None, 'mrtrix', 'descoteaux'}
```
skoudoro: same as above

dipy/reconst/shm.py
```diff
@@ -861,10 +862,10 @@ def sf_to_sh(sf, sphere, sh_order=4, basis_type=None, smooth=0.0):
...
         ``None`` for the default dipy basis,
         ``mrtrix`` for the MRtrix basis, and
-        ``fibernav`` for the FiberNavigator basis
```
skoudoro: same as above

jchoude:
Also a quick question: is there some place (in the doc, maybe?) where the different bases are explained, or a least a reference to the paper? I would guess somewhere in the `sphm lookup` code? Just to make sure what is defined.

arokem:
I don't think there is a page in the docs with a clear explanation of this.
It would certainly be a welcome addition!

dipy/direction/peaks.py
jhlegarreta: Done in 11e6852.

dipy/direction/peaks.py
jhlegarreta: Done in 11e6852.

dipy/reconst/csdeconv.py
jhlegarreta: Done in 11e6852.

dipy/reconst/shm.py
jhlegarreta: Done in 11e6852.

jhlegarreta:
@skoudoro thanks for the review ! As for the docs, I can do it. Should this just go as a note in `shm.py` beside the dictionary `sph_harm_lookup`, or else/and should it be somewhere else in the docs/examples as well?

jhlegarreta:
Test errors are unrelated; some configurations report success.

arokem:
Yeah. We've been seeing these errors on other PRs in the last couple of days. See https://github.com/nipy/dipy/issues/1654

We're still trying to figure that one out.

As for documentation: I think that this issue would merit a page in the narrative documentation of the software. Or an entire example devoted just to an explanation of the differences between the basis sets (e.g., plotting the relationships between them, and how you might go between one and the other). It would be good to have something that we could point to whenever people ask us about this.

jchoude:
Personnally, LGTM.

Should we go ahead with this, and have another PR for the documentation?

jhlegarreta:
IMHO I think it's cleaner to do a separate topic for the documentation, since this kind of a feature/option-like addition, and in case of trouble, the documentation topic can be left aside/not blamed on. Just point where it should go, and I'll add that as soon as I can.

skoudoro:
All PR's are blocked because of #1654. I think we should not merge it until we fix this issue and make Travis happy.

jhlegarreta:
As for the place where to put the references, I think desirable would be where the list is [defined](https://github.com/nipy/dipy/blob/e24ad736de1856d645173ced3f0a7da517011cc2/dipy/reconst/shm.py#L331):
```
sph_harm_lookup = {None: real_sym_sh_basis,
                   "mrtrix": real_sym_sh_mrtrix,
                   "fibernav": real_sym_sh_basis}
```

But that does not make its way into the doc, so the alternative for `MRtrix3` would be in the method [real_sym_sh_mrtrix](https://github.com/nipy/dipy/blob/e24ad736de1856d645173ced3f0a7da517011cc2/dipy/reconst/shm.py#L242). I've asked to the MRtrix3 people which of the Tournier's (?) refs should be put, or else, whether we should directly ref to their documentation.

For, `descoteaux` it would be in the method [real_sym_sh_basis](https://github.com/nipy/dipy/blob/e24ad736de1856d645173ced3f0a7da517011cc2/dipy/reconst/shm.py#L285), removing the link to the software and replacing by a ref to the paper
```
Descoteaux, M. , Angelino, E. , Fitzgibbons, S. and Deriche, R. (2006), Apparent diffusion
coefficients from high angular resolution diffusion imaging: Estimation and applications.
Magn. Reson. Med., 56: 395-410. doi:10.1002/mrm.20948
```

Please correct me if I'm wrong.

skoudoro:
Why we do not do the same for `mrtrix`? removing the software and replace it by `tournier` and add the paper instead of the reference to their documentation?

jhlegarreta:
That's something that crossed my mind when thinking about the documentation, but I did not bring it into the table because I thought that the `MRtrix3` method was more likely to be changed over time/was better identified by the software, but it is definitely more consistent from our side to do the same for both cases.

So may be a new PR is then better if JC and Ariel are also for this change.

dipy/direction/peaks.py
```diff
@@ -423,10 +423,10 @@ def peaks_from_model(model, data, sphere, relative_peak_threshold,
     sh_order : int, optional
         Maximum SH order in the SH fit.  For `sh_order`, there will be
         ``(sh_order + 1) * (sh_order + 2) / 2`` SH coefficients (default 8).
-    sh_basis_type : {None, 'mrtrix', 'fibernav'}
-        ``None`` for the default dipy basis which is the fibernav basis,
+    sh_basis_type : {None, 'mrtrix', 'fibernav', 'descoteaux'}
+        ``None`` for the default dipy basis which is the descoteaux basis,
         ``mrtrix`` for the MRtrix basis, and
-        ``fibernav`` for the FiberNavigator basis
+        ``descoteaux`` for the Descoteaux basis
```
arokem: Need a reference here.

Also: should we remove `'fibernav'` from here, since this is the one that is getting deprecated? If people have this in their code, we don't want to break it, but we also don't necessarily want to advertise this as an option, so people don't add it to new code they are writing.

dipy/reconst/csdeconv.py
```diff
@@ -702,8 +702,8 @@ def odf_sh_to_sharp(odfs_sh, sphere, basis=None, ratio=3 / 15., sh_order=8,
         array of odfs expressed as spherical harmonics coefficients
     sphere : Sphere
         sphere used to build the regularization matrix
-    basis : {None, 'mrtrix', 'fibernav'}
-        different spherical harmonic basis. None is the fibernav basis as well.
+    basis : {None, 'mrtrix', 'fibernav', 'descoteaux'}
+        different spherical harmonic basis. None is the descoteaux basis as well.
```
arokem: Same comment here as above: remove the `'fibernav'` option, and refer to the paper here.

dipy/reconst/shm.py
```diff
@@ -861,10 +863,11 @@ def sf_to_sh(sf, sphere, sh_order=4, basis_type=None, smooth=0.0):
     sh_order : int, optional
         Maximum SH order in the SH fit.  For `sh_order`, there will be
         ``(sh_order + 1) * (sh_order_2) / 2`` SH coefficients (default 4).
-    basis_type : {None, 'mrtrix', 'fibernav'}
+    basis_type : {None, 'mrtrix', 'fibernav', 'descoteaux'}
```
arokem: And same comment here.

dipy/reconst/shm.py
```diff
@@ -861,10 +863,11 @@ def sf_to_sh(sf, sphere, sh_order=4, basis_type=None, smooth=0.0):
...
         ``None`` for the default dipy basis,
-        ``mrtrix`` for the MRtrix basis, and
-        ``fibernav`` for the FiberNavigator basis
+        ``mrtrix`` for the MRtrix basis,
+        ``fibernav`` (deprecated), and
+        ``descoteaux`` for the Descoteaux basis
         (default ``None``).
```
arokem: I would add here that `None` implies `'descoteaux'`

dipy/reconst/shm.py
```diff
@@ -900,10 +908,11 @@ def sh_to_sf(sh, sphere, sh_order, basis_type=None):
     sh_order : int, optional
         Maximum SH order in the SH fit.  For `sh_order`, there will be
         ``(sh_order + 1) * (sh_order_2) / 2`` SH coefficients (default 4).
-    basis_type : {None, 'mrtrix', 'fibernav'}
+    basis_type : {None, 'mrtrix', 'fibernav', 'descoteaux'}
```
arokem: Same comments here as above

dipy/reconst/tests/test_shm.py
```diff
@@ -364,8 +364,8 @@ def test_sf_to_sh():
     odf2 = sh_to_sf(odf_sh, sphere, 8, "mrtrix")
     assert_array_almost_equal(odf, odf2, 2)
 
-    odf_sh = sf_to_sh(odf, sphere, 8, "fibernav")
-    odf2 = sh_to_sf(odf_sh, sphere, 8, "fibernav")
+    odf_sh = sf_to_sh(odf, sphere, 8, "descoteaux")
+    odf2 = sh_to_sf(odf_sh, sphere, 8, "descoteaux")
     assert_array_almost_equal(odf, odf2, 2)
```
arokem: You might want to add a test here for the deprecation warning. 

See for example https://github.com/nipy/dipy/blob/master/dipy/reconst/tests/test_shm.py#L446 for how to catch a warning and test that it exists (in that example, we're checking that it doesn't exist).

jhlegarreta:
@arokem thanks for the comments !

> Regarding the naming of the mrtrix basis. What does the BIDS spec have to say about situations where the same author has proposed more than one basis set?

As for how BIDS foresees to handle multiple basis from the same author, as it may happen with Tournier's, I cannot say much: whether it is adding the year or some other appendix that allows to uniquely identify the basis in an informative way.

> Also: should we remove 'fibernav' from here, since this is the one that is getting deprecated? If people have this in their code, we don't want to break it, but we also don't necessarily want to advertise this as an option, so people don't add it to new code they are writing.

Makes sense to me. I'm fine with whatever you decide is better.

jchoude:
@arokem As for the naming of tournier / Mrtrix basis, Robert Smith (mrtrix core dev) commented that the Mrtrix-related bases should be `mrtrix_0.2` and `mrtrix3`. I agree with that because of the difference in normalization between the 2 versions of the software. Else, we could have `tournier` and `tournier_normalized`, but for basic users it might be less clear.

Garyfallidis:
I feel this change does not solve the problem. Neither makes things more clear.
I would really prefer if the name was related to the difference between the bases. Or if at least the name of the parameter would link to a specific paper and year. However, if we change this we should also change the mrtrix parameter.

Please also communicate before making changes in BIDS or otherwise concerning DIPY.
@jhlegarreta can you send me the link to the last BIDS update?

The parameters should be "descoteaux07" and "tournier07" rather than "fibernav" and "mrtrix". That I think clarifies the naming issue.

Garyfallidis:
@jchoude for me also important is to have the year of the basis introduced if that is possible. So Tournier07  and Descoteaux06 etc. Mrtrix3 for example is not a solution because the basis can change many times while mrtrix3 is evolving. We have seen this before.

Garyfallidis:
So, my suggestion is to have the authors' name and year were the basis was introduced.

Garyfallidis:
And that I think solves most of our problems.

jchoude:
@Garyfallidis I agree that the year might help. I have no strong opinion, I just think that having, somewhere easily accessible, a mapping telling people that the "old" Tournier basis == mrtrix 0.2, and the "new" Tournier is Mrtrix 3 would be quite helpful for users that don't know exactly the litterature.

Also, if I had to guess, I would bet that the Mrtrix 3 basis will not change anymore, since it's exactly the old Tournier basis with an added normalization. If they introduced some new fancy basis, I think it would have another different name.

Garyfallidis:
Also @jhlegarreta - @jchoude has already sent me the link so no need to resend.

arokem:
+1 to @Garyfallidis suggestions for naming (`descoteaux07` and `tournier07`). According to that logic, the mrtrix3 basis would be `tournier12`, I believe (pointing to this paper: https://onlinelibrary.wiley.com/doi/abs/10.1002/ima.22005)

I think that a documentation page that has the mapping, explains the mapping and lays out all the namings should help users navigate these issues.

skoudoro:
> Test errors are unrelated; some configurations report success

it is fixed, you can rebase

jhlegarreta:
Thanks for the discussion !

I will then submit the change in another topic (it is no longer only about `fibernav)` over the next few days: so `descoteaux07` and `tournier07`. Adding the corresponding deprecation messages and testing them.

Will subsequently submit another PR with the doc if you let me know the best location for doing so.

arokem:
Maybe in here? https://github.com/nipy/dipy/tree/master/doc/theory

jhlegarreta:
Looks the right place to me 👍. Thanks Ariel. Will do these over the weekend.

PR #1658, opened 2018-10-27
STYLE: Honor 'descoteaux'and 'tournier' SH basis naming.

jhlegarreta:
Honor the authors in the naming for SH basis choice: replace and deprecate
`fibernav` and `mrtrix` and use `descoteaux07` and `tournier07` instead.

The change includes:
- Using the author's name rather than the software/implementation's.
- Testing that the deprecation warning is effectively exercised.
- Adding the necesary references.


skoudoro:
> I had an answer from Tournier to a mail about the SH basis...

Thank for this info. The code looks good so, I think we can merge it after another review.

Thanks @jhlegarreta

arokem:
Looks good to me. Merging.

PR #1659, opened 2018-10-29
From dipy.viz to FURY

skoudoro:
* Remove fvtk module
*  `dipy.viz` give birth to [FURY](https://fury.gl) --> New Scientific visualization library

More information about **FURY**:
* Fury Github: https://github.com/fury-gl/fury
* Fury Documentation: https://fury.gl


arokem:
OK. This all looks good to me. 

Do you want to add some documentation here pointing to Fury? I think that you mention it where we previously pointed to VTK, but I would add a link there, so that people can easily find it.

skoudoro:
oh ok, It is not, you are right. I do not know yet where to put it but will do

arokem:
How about in the installation instructions?

dipy/io/vtk.py
```diff
@@ -15,8 +13,37 @@
     major_version = vtk.vtkVersion.GetVTKMajorVersion()
 
 
+def set_input(vtk_object, inp):
```
Garyfallidis: @skoudoro remind me something this function stays in DIPY because we want to have surface loading support? Or there is some other reason? I thought the idea was to remove direct dependencies to vtk. And only indirectly depend through FURY?

skoudoro:
Thanks @arokem and @Garyfallidis for the review.  I added the documentation link and fix `io/vtk`.

It looks like it is ready to go! Can you have a last check?

Garyfallidis:
Great work! To Inf+!

PR #1660, opened 2018-10-29
OPT - moved the tolerance check outside of the for loop

danielenricocahall:
I moved the conditional which assigned "tolerance" if it was None outside of the for-loop, as that check  only needs to be done once.


skoudoro:
Great! Thanks for this update @danielenricocahall.

I will merge it when Travis and Appeyor are happy

Issue #1661, opened 2018-11-01
Local Tracking - BinaryTissueClassifier - dropping the last point


jchoude:
## Description
@gabknight This is the old classical question: should the last point of a streamline, the point that makes it finished (eg: reaching a `0` value in a Binary map), be kept or not in the tractogram?

When using `local_tracking` with a `BinaryTissueClassifier`, once a streamline exits the `True` valued voxels of the mask, a `ENDPOINT` flag is raised, the number of steps (`i` in the loop) is incremented by 1. This means that the tracker will keep the last point, even if, in fact, it is outside the mask.

I think we should probably drop that last point. Else, semantically, it doesn't make a lot of sense. Let's say I track, and then just do a endpoint filtering with the same mask, I will be missing some streamlines, since the endpoints will be outside of the mask.

@gabknight do you agree that this should be the behavior? If so, @czotti could help implementing it.

gabknight:
@jchoude yes ok. Removing the `i += 1` in dipy/tracking/local/localtrack.pyx:189 should do the trick. Let's also remove the last point for the `INVALIDPOINT` case. 

I would do that for `local_tracker` but not for `pft_tracker`. (`BinaryTissueClassifier` cannot be used for PFT, only `ConstrainedTissueClassifier`). Removing the last point of streamlines stopping in a voxel with lower then 100% GM is not ideal. Moreover, I would rather have all streamlines extend further in the GM for connectivity analysis.  

The annoying thing with this is that PFTracking and localTracking will have different behavior when both use e.g. `CMCTissueClassifier`. I suggest clarifying what happen to the last points in the doc. What do you think?

@czotti Do you want to give it a go? The code modification is easy, but the modification will break all `localTracking` tests. :)

jchoude:
@gabknight Thanks for that! I agree that PFTracking shouldn't be modified.

skoudoro:
fixed by #1666. Closing this issue

PR #1662, opened 2018-11-02
Remove the points added oustide of a mask. Fix the related tests.

czotti:
This PR is to address the issue #1661.

- It change the way points are added to a streamline if the point is outside the mask.
- I changed all the needed test.

Unfortunately, one test is failing on the travis-ci when the `visu` is enabled. You are currently removing this from dipy (#1659) do you want it to be fixed (@skoudoro)?
@gabknight, @jchoude.


czotti:
I still need to make some changes because now some tests have streamlines of size 1 which is not a streamline.

czotti:
Ready for review, @gabknight, @jchoude.

skoudoro:
> do you want it to be fixed (@skoudoro)

ok, it's fine. Do not forget you rebase it when #1659 is in. 

FYI, DIPY release 0.15.0 is just waiting for #1659.  I would recommend @gabknight and @jchoude to review this PR as soon as possible is they want to see this one in.

PR #1667, opened 2018-11-22
Changing the default b0_threshold in gtab

ShreyasFadnavis:
More than often the b0s are not 0 and then if this is not set correctly in gradient_table that creates other problems across the pipeline for example getting response function with nans etc.


Garyfallidis:
+1000! for this.  The b0s are not necessarily exactly 0 but usually lower than 50 indeed.

Garyfallidis:
I need a second core dev. to approve this PR. Thank you @ShreyasFadnavis and all!

skoudoro:
Thank @ShreyasFadnavis for this. 

Can you update `b0_threshold` default value on all workflows like [here](https://github.com/nipy/dipy/blob/master/dipy/workflows/reconst.py#L35)? 

I think we should warn the user when they are doing something strange. So you should compare `bval` min and `b0_threshold` as I did on [this PR](https://github.com/nipy/dipy/pull/1621) and raise a warning. Can you add this behavior on this 5 functions?

jchoude:
I second the request to add a warning. Also, should there be an upper bound on b0_threshold?

Garyfallidis:
The upper bound is acquisition dependent.

jchoude:
Yes, but I mean, would it make sense to have a B=0 threshold higher than, say, 100? It wouldn't really be a B=0 in this case, no?

Garyfallidis:
Difficult to say. I have seen data with b0 at 100. But indeed it is rare.

Garyfallidis:
I suppose it is safe to have an upper bound at 200 and raise a warning if it goes beyond that.

ShreyasFadnavis:
@skoudoro @jchoude @Garyfallidis Thank you all! So, I will add change the `b0_threshold` in the workflows  as @skoudoro suggested. Also will add a warning and an upper bound of `200` to the b0_threshold?

Garyfallidis:
Also a warning for a lower bound. It cannot be less than 0. 0 is the minimum. Thanks!

skoudoro:
Do not forget to update the tests with this default value and add new one for the warning. Thanks!

ShreyasFadnavis:
> Also a warning for a lower bound. It cannot be less than 0. 0 is the minimum. Thanks!

@Garyfallidis the lower bound for `b0_threshold` should be the lowest b0 value right? not 0 as the min, because the `auto_response` may give an issue in that case..

Garyfallidis:
Nope!

Garyfallidis:
All I am saying is check if b0_threshold is accidentally negative. Less than 0. If that is the case you should raise an Error. Sorry for not being clear multi-tasking right now with baby and laptop. But hey no excuses. Keep it up @ShreyasFadnavis !

ShreyasFadnavis:
> I suppose it is safe to have an upper bound at 200 and raise a warning if it goes beyond that.

@Garyfallidis I am raising a warning if >= 200, but am not sure if I should bound it by replacing the actual value.

Garyfallidis:
Nope do not replace the original value. Thank you.

dipy/reconst/tests/test_ivim.py
```diff
@@ -65,7 +65,7 @@
                         500., 600., 700., 800., 900., 1000.])
 
 bvecs_no_b0 = generate_bvecs(N)
-gtab_no_b0 = gradient_table(bvals_no_b0, bvecs.T)
+gtab_no_b0 = gradient_table(bvals_no_b0, bvecs.T, b0_threshold=0)
```
arokem: The problem here is that it would break backwards compatibility for people who are already using IVIM with a gradient table initialized with the default parameters. I think that we would need to check and provide a better error for people who are using their code as it was.

ShreyasFadnavis: @arokem Thank you for this! Makes sense to me.. Should I set up a warning?

skoudoro: I think this PR is almost ready to go. the only point is I have no clue to solve this backward compatibility problem.  I do not think a warning message will solve this problem and we have already [this check](https://github.com/nipy/dipy/blob/master/dipy/reconst/ivim.py#L216). 

Any suggestion?

arokem: I think that we need to do this within the IVIM implementation. One way to deal with this would be to check whether the gtab provided has a non-zero b0_threshold, and raise an error when that's the case, unless the user gives some other indication that that's intended. It's a bit kludgy, I'll admit it. Open to other suggestions.

ShreyasFadnavis: @arokem Will setting the default value of `b0_threshold` to `50` have repercussions in the functioning of IVIM for current users? I'm not sure I understand. Please could you help me with  understanding this?

arokem: The problem is that for IVIM, a user might very reasonably have a measurement with a b-value of 40, that should not be considered a b0 measurement. In previous versions of the software, when they ran the default code, it wasn't. But with this change introduced that measurement will now be considered to have a b-value of 0, and they wouldn't have any way of knowing that this change was introduced.

dipy/reconst/ivim.py
```diff
@@ -215,6 +215,9 @@ def __init__(self, gtab, split_b_D=400.0, split_b_S0=200., bounds=None,
             e_s += "The IVIM model requires signal measured at 0 bvalue"
             raise ValueError(e_s)
 
+        if gtab.b0_threshold > 0:
+            warnings.warn('The default b0_threshold > 0 and is now set to 50')
```
arokem: I don't understand this warning. 

As I user, I would infer that this means that you have changed my b0_threshold to 50. 

I would propose to raise an Error (instead of the warning) with the following string:

"The IVIM model requires a measurement at b==0. As of version 0.15, the default `b0_threshold` for the `GradientTable` object is set to 50, so if you used the default settings to initialize the `gtab` input to the IVIM model, you may have provided a `gtab` with `b0_threshold` larger than 0. Please initialize the `gtab` input with `b0_threshold=0`"

ShreyasFadnavis: @arokem Thank you so much for this! I shall refactor accordingly!

skoudoro:
Good! I think this PR is ready to be merged. I will do it this evening if there is no objection

dipy/reconst/ivim.py
```diff
@@ -215,6 +215,16 @@ def __init__(self, gtab, split_b_D=400.0, split_b_S0=200., bounds=None,
             e_s += "The IVIM model requires signal measured at 0 bvalue"
             raise ValueError(e_s)
 
+        if gtab.b0_threshold > 0:
```
arokem: Any chance to get a test for this behavior in the IVIM tests?

dipy/reconst/ivim.py
```diff
@@ -381,7 +391,7 @@ def estimate_f_D_star(self, params_f_D_star, data, S0, D):
                 warningMsg += " as initial guess for leastsq. Parameters are"
                 warningMsg += " returned only from the linear fit."
                 warnings.warn(warningMsg, UserWarning)
-                f, D_star = params_f_D
+                f, D_star = params_f_D_star
```
arokem: What's this change? Was this just a bug that was hiding in there? Looks like it.

ShreyasFadnavis: @arokem  yes! It was just a minor bug in the code.. I found it while reading it.

arokem:
I wonder whether it would be worth adding a test that exercises this. I
think we can do that on a separate PR, but if you could open an issue so
that we remember to do that, it would help.

dipy/reconst/ivim.py
ShreyasFadnavis: @arokem Thank you for this! I will create a new issue and PR with the test for the same 👍

skoudoro:
ok, thank you @ShreyasFadnavis

PR #1669, opened 2018-11-26
Flow csd sh order

arokem:
In using the CSD workflow, I discovered that changing the `sh_order` input parameter doesn't seem to do anything. Tracking this down [in the code](https://github.com/nipy/dipy/blob/master/dipy/workflows/reconst.py#L523), I see that it was hard-coded as `sh_order = 8`. This PR changes that and also adds some testing around that. It also more flexibly checks the inputs, so that we are not making assumptions about the input data.

### DIFF
--- a/dipy/workflows/tests/test_reconst_csa_csd.py
+++ b/dipy/workflows/tests/test_reconst_csa_csd.py
@@ -13,6 +13,7 @@ from nibabel.tmpdirs import TemporaryDirectory
 
 from dipy.data import get_data
 from dipy.workflows.reconst import ReconstCSDFlow, ReconstCSAFlow
+from dipy.reconst.shm import sph_harm_ind_list
 logging.getLogger().setLevel(logging.INFO)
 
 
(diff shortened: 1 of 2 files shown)


dipy/workflows/tests/test_reconst_csa_csd.py
```diff
@@ -35,77 +36,82 @@ def reconst_flow_core(flow):
...
+            peaks_vals_data = nib.load(peaks_vals_path).get_data()
+            assert_equal(peaks_vals_data.shape[-1], 5)
+            assert_equal(peaks_vals_data.shape[:-1], volume.shape[:-1])
+
+            shm_path = reconst_flow.last_generated_outputs['out_shm']
+            shm_data = nib.load(shm_path).get_data()
+            # Test that the number of coefficients is what you would expect
+            # given the order of the sh basis:
+            assert_equal(shm_data.shape[-1],
+                            sph_harm_ind_list(sh_order)[0].shape[0])
```
arokem: This is where I test that the shape of the output conforms with the input `sh_order` provided.

PR #1675, opened 2018-11-29
WIP: Integer indices

arokem:
I ran into these two indexing operations that otherwise throw errors of this kind: 

```
  File "/Users/arokem/.virtualenvs/dipy/bin/dipy_labelsbundles", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/Users/arokem/source/dipy/bin/dipy_labelsbundles", line 9, in <module>
    run_flow(LabelsBundlesFlow())
  File "/Users/arokem/source/dipy/dipy/workflows/flow_runner.py", line 87, in run_flow
    return flow.run(**args)
  File "/Users/arokem/source/dipy/dipy/workflows/segment.py", line 317, in run
    save_trk(out_bundle, streamlines[location], np.eye(4))
  File "/Users/arokem/.virtualenvs/dipy/lib/python3.7/site-packages/nibabel/streamlines/array_sequence.py", line 315, in __getitem__
    seq._offsets = self._offsets[off_idx]
```

I have a couple of questions (I guess for @Garyfallidis?): The first is whether these fixes make sense to you? Do we need to use a rounding operation instead of the cast to integer in `bundles.py`? For the workflow indexing error, can we prevent this by saving the npy files as integer dtype? Any reason not to do that? The other is how we would go about testing for these errors. These cases don't seem to be covered right now.

### DIFF
--- a/dipy/segment/bundles.py
+++ b/dipy/segment/bundles.py
@@ -612,7 +612,8 @@ class RecoBundles(object):
         pruned_indices = [rtransf_cluster_map[i].indices
                           for i in np.where(mins != np.inf)[0]]
         pruned_indices = list(chain(*pruned_indices))
-        pruned_streamlines = transf_streamlines[np.array(pruned_indices)]
+        pruned_streamlines = transf_streamlines[
+                                  np.array(pruned_indices).astype(int)]
 
         initial_indices = list(chain(*neighb_indices))
         final_indices = [initial_indices[i] for i in pruned_indices]
--- a/dipy/workflows/segment.py
+++ b/dipy/workflows/segment.py
@@ -312,7 +312,7 @@ class LabelsBundlesFlow(Workflow):
             logging.info(sf)
             streamlines, header = load_trk(sf)
             logging.info(lb)
-            location = np.load(lb)
+            location = np.load(lb).astype(int)
             logging.info('Saving output files ...')
             save_trk(out_bundle, streamlines[location], np.eye(4))
             logging.info(out_bundle)


Garyfallidis:
Those should be always integers upon loading. These are indices to the original tractograms so they do not need some new operation (rounding or otherwise). I am impressed you got this error. Can you explain how you called dipy_labelsbundles?

Now about the npy files. We can use .txt files for these.  We will have to change np.load/save to np.loadtxt/savetxt. Be happy to update this PR with this change.

arokem:
I used this script: https://github.com/nipy/dipy_qa/blob/master/code/simple_flow_series.sh :smile:

I am trying to debug this, but it's turned out to be a bit tricky. It seems to be resolved by the fixes I propose, but also might be to do with cases in which the arrays are empty? Not sure, and will report back when I learn more.

I don't think that we'd want to move to text files, though. Wouldn't those be larger in size?

arokem:
One thing that I already know is that the script crashes if the index file for one of the bundles is empty (I guess that means no streamlines detected for this bundle?)

dipy/segment/bundles.py
```diff
@@ -612,7 +612,8 @@ def _prune_what_not_in_model(self, model_centroids,
         pruned_indices = [rtransf_cluster_map[i].indices
                           for i in np.where(mins != np.inf)[0]]
         pruned_indices = list(chain(*pruned_indices))
-        pruned_streamlines = transf_streamlines[np.array(pruned_indices)]
```
arokem: OK - more feedback. This definitely raises an error when the pruned_indices are empty, because in that case, the array does not have integer dtype, so you get "indexing array must be of integer dtype". Should we raise another error in this case, or just check and return an empty list in that case?

arokem:
Need to:

- [ ] Add some tests

skoudoro:
What is the status of this?  Could you create a test that generates this error?

Can you look at this @BramshQamar and give feedback to @arokem?

Thank you both

dipy/segment/bundles.py
BramshQamar: This we fixed by returning an empty object and printing this message: "You have removed all streamlines"

https://github.com/dipy/dipy/blob/7eca4013bd3a407df2fac533ba6cec3b310eb90f/dipy/segment/bundles.py#L829

BramshQamar:
> One thing that I already know is that the script crashes if the index file for one of the bundles is empty (I guess that means no streamlines detected for this bundle?)

This is also fixed. We first check the length of labels, if 0 then we save an empty .trk file.

https://github.com/dipy/dipy/blob/7eca4013bd3a407df2fac533ba6cec3b310eb90f/dipy/workflows/segment.py#L328

BramshQamar:
To sum it up:

- Labels.npy file has an index number of streamline in the original tractogram. An index would be always integer, we don't need to round it off (not a float/double). Label.npy file is numpy file that looks something like this: 

`array([ 64382, 278939, 913640, 386841, 634669, 311381, 686057, 365665,
       727446, 494229, 858977, 861820, 521244,  71859, 304173, 550622,
       728690, 703552, 176213, 105112, 679325, 693097, 540822, 524968,
       232786, 330169, 787551, 738648, 790197, 561420,  27656, 368316,
       550086, 839272, 994910, 335221, 422306, 777345, 881720, 404528,
       916488, 206925, 561916,  80920, 959659, 174820, 593094, 153357,
       193745,  41372, 231992])`

       Here, each integer is a streamline index in the tractogram. 

- If pruned streamlines has no streamlines left, we return an empty streamline object.
- If the labels.npy file is empty, we save the empty .trk bundle file.

I think, all the issues are already fixed. Let me know if I missed anything @arokem @skoudoro

arokem:
Yes. I think we can close this.

PR #1676, opened 2018-11-30
[FIX] Example Update

skoudoro:
It seems that some examples were broken after the migration to FURY package. So this is just an update.


arokem:
Thanks for doing this. Looks good! Since this is a documentation fix, I'm going to go ahead and merge it immediately. If there is something we need to change here, we can follow up with another PR.

skoudoro:
Sounds good, Thanks!

PR #1677, opened 2018-11-30
[FIX] workflow help msg

### LINKED ISSUE #1674: Workflows should print out help per default
For example, when you type: 

   dipy_fit_dti 

It would be much better if you got the help documentation than the current: 

```
(dipy) code (master)$dipy_fit_dti 
Too few arguments: Program dipy_fit_dti expects arguments.

Type dipy_fit_dti -h for help.

  File "/Users/arokem/.virtualenvs/dipy/bin/dipy_fit_dti", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/Users/arokem/source/dipy/bin/dipy_fit_dti", line 9, in <module>
    run_flow(ReconstDtiFlow())
  File "/Users/arokem/source/dipy/dipy/workflows/flow_runner.py", line 57, in run_flow
    args = parser.get_flow_args()
  File "/Users/arokem/source/dipy/dipy/workflows/base.py", line 275, in get_flow_args
    ns_args = self.parse_args(args, namespace)
  File "/Users/arokem/source/dipy/dipy/fixes/argparse.py", line 1667, in parse_args
    args, argv = self.parse_known_args(args, namespace)
  File "/Users/arokem/source/dipy/dipy/fixes/argparse.py", line 1699, in parse_known_args
    return self._parse_known_args(args, namespace)
  File "/Users/arokem/source/dipy/dipy/fixes/argparse.py", line 1917, in _parse_known_args
    self.error(_('too few arguments'))
UnboundLocalError: local variable '_' referenced before assignment
```

Which does include the instruction to use the `-h` flag, but is buried inside of a stack that is not super informative. Does anyone know how to make `-h` the default behavior in the absence of any other inputs?

skoudoro:
The goal of this PR is to fix #1674. Now, you get:

```
(dipy_dev_3) 149-160-137-169:dipy koudoro$   ~/.local/bin/dipy_mask
usage: dipy_mask [-h] [--ub float] [--out_dir str] [--out_mask str] [--force]
                 [--version] [--out_strat string] [--mix_names]
                 [--log_level string] [--log_file string]
                 input_files lb
dipy_mask: error: the following arguments are required: input_files, lb
(dipy_dev_3) 149-160-137-169:dipy koudoro$
```

I use the opportunity to:

- remove our `argparse` module copy. It was there for the backward compatibility with Python 2.6 so not needed anymore
- Standardize a bit more workflows input/output
- Check and update Workflows docstring


arokem:
This is the best thing I've seen all day: 

screen shot 2018-11-30 at 4 04 22 pm

dipy/workflows/base.py
```diff
@@ -153,7 +155,7 @@ def add_workflow(self, workflow):
                 _kwargs['type'] = str
 
             if isnarg:
-                _kwargs['nargs'] = '*'
+                _kwargs['nargs'] = '*' if is_optionnal else '+'
```
arokem: Could you please change `is_optionnal` to `is_optional` while you're at it?

skoudoro:
> This is the best thing I've seen all day:

ahahah 😄

> Is this still WIP?

Finally no, I wanted to fix another issue on the same PR. Finally, I should just create a new PR. 

> What's the crucial change that makes this produce the help message instead of the error?

[here](https://github.com/nipy/dipy/pull/1677/files#diff-589e4e365dd5ae544b73601cc9c2a04bR4) is the crucial change. we were using an old version of `argparse` to keep the compatibility with python 2.6. We do not need this anymore and Python dev improved the default behavior of `argparse` module. You can see [here the doc](https://docs.python.org/3/library/argparse.html)

The other changes are just code cleaning.

arokem:
Cool. This all looks good to me. If you just make a quick change to fix that typo (`optionnal`) then I will merge this tomorrow night, unless someone else wants to chime in.

arokem:
Thanks!

PR #1680, opened 2018-12-04
[Fix] workflow variable string

### LINKED ISSUE #1338: Variable string input does not work with self.get_io_iterator() in workflows
Hi @matthieudumont,

variable string currently works only as an optional parameter.
Can you look at this and see if we can integrated as an accepted input?

Let me know what you think.

skoudoro:
The aim of this PR is to fix #1338.

This allows `variable string` type for `self.get_io_iterator()` function


Garyfallidis:
Thanks for the fix Serge.

PR #1681, opened 2018-12-05
fixed issue with cst orientation in bundle_extraction example

BramshQamar:


doc/examples/bundle_extraction.py
```diff
@@ -153,10 +153,12 @@
 
 ren = window.Renderer()
 ren.SetBackground(1, 1, 1)
-ren.add(actor.line(model_bundle, colors=(1,0,1)))
-ren.add(actor.line(recognized_bundle, colors=(1,1,0)))
+ren.add(actor.line(model_bundle, colors=(.1,.7,.26)))
+ren.add(actor.line(recognized_bundle, colors=(.1,.1,6)))
 window.record(ren, out_path='CST_L_recognized_bundle.png',
-              size=(600, 600))
```
arokem: I believe these lines should be: 

```
ren.set_camera(focal_point=(-18.17281532, -19.55606842, 6.92485857),
               position=(-360.11, -340.46, -40.44),
               view_up=(-0.03, 0.028, 0.89))

window.record(ren, out_path='CST_L_recognized_bundle.png', size=(600, 600))
```

doc/examples/bundle_extraction.py
```diff
@@ -155,10 +155,11 @@
...
+ren.set_camera(focal_point=(-18.17281532, -19.55606842, 6.92485857),
+               position=(-360.11, -340.46, -40.44),
+               view_up=(-0.03, 0.028, 0.89))
 window.record(ren, out_path='CST_L_recognized_bundle.png',
-              ren.set_camera(focal_point=(-18.17281532, -19.55606842, 6.92485857),
-                             position=(-360.11, -340.46, -40.44),
-                             view_up=(-0.03, 0.028, 0.89)), size=(600, 600))
+              size=(600, 600))
 if interactive:
     window.show(ren)
```
Garyfallidis: I believe in show and record you have to say reset_camera=False

Garyfallidis: @BramshQamar let me know if this is not necessary.

doc/examples/bundle_extraction.py
BramshQamar: Hello Ariel,

Thanks for point out the typo. I have fixed it and also camera view for AF_L bundle.

here's how they look now:
af_l_recognized_bundle

cst_l_recognized_bundle

doc/examples/bundle_extraction.py
BramshQamar: It works fine without setting reset_camera=False.
if I set it to False then our streamlines appear to be very far and zoomed out.

doc/examples/bundle_extraction.py
BramshQamar: @Garyfallidis

doc/examples/bundle_extraction.py
Garyfallidis: I believe still fury updates the camera automatically when reset_camera=True. Only for scaling though.

arokem:
Looks great! One more tiny thing: I think that we want to set leave `interactive=False` per default (on L118 and on L154).

doc/examples/bundle_extraction.py
BramshQamar: if I do set_camera=False

this is how it saves the .png

af_l_recognized_bundle

BramshQamar: we cannot see anything

arokem: I can't see anything in there? :smile:

Seems like maybe a bug in fury?

arokem:
LGTM. Merging. 

I also posted an issue on Fury, so we can resolve the `reset_camera` conundrum there.

arokem:
Thanks for doing this!

Garyfallidis:
You need to call clipping_range too and make sure that the camera is looking at the object correctly.

Garyfallidis:
There is nothing wrong. I am explaining a bit what is the standard procedure when you do not want to have fury generating automatically camera parameters. It might be useful for other bundles.

Garyfallidis:
No issue for this being merged.

arokem:
Gotcha! Should we add more details about how to do this correctly in the [bundle visualization example](http://nipy.org/dipy/examples_built/viz_bundles.html#example-viz-bundles)? Maybe make an issue about this? Should be a relatively low-hanging fruit for relatively new contributors.

Garyfallidis:
It's fine. Let's move on. That clarification should go to fury and you already reported an issue there.

PR #1684, opened 2018-12-08
[FIX] testing line-based target function

### LINKED ISSUE #1679: Intermittent issue in testing line-based target function
On current `master`, I intermittently get this test error: 

```
FAIL: dipy.tracking.tests.test_utils.test_target_lb
  File "/Users/arokem/.virtualenvs/dipy/lib/python3.7/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/Users/arokem/source/dipy/dipy/tracking/tests/test_utils.py", line 237, in test_target_lb
    _target(target_line_based, streamlines, (1, 1, 1), (2, 1, 1), False)
  File "/Users/arokem/source/dipy/dipy/tracking/tests/test_utils.py", line 276, in _target
    assert_equal(len(new), 1)
AssertionError: 2 != 1

```

This probably has something to do with the values in the affine that is randomly set on L272, but I couldn't figure out what values cause this to happen.

skoudoro:
The aim of this PR is to fix #1679. 

line-based target function was failing for some random case so I replaced `np.random`by a static array.

### DIFF
--- a/dipy/tracking/tests/test_utils.py
+++ b/dipy/tracking/tests/test_utils.py
@@ -269,7 +269,11 @@ def _target(target_f, streamlines, voxel_both_true, voxel_one_true,
         assert_raises(ValueError, list, new)
 
     # Test smaller voxels
-    affine = np.random.random((4, 4)) - .5
+    random_array = np.array([[0.2862315, 0.44142904, 0.19837613, 0.422711],
+                             [0.7434331, 0.55335599, 0.18633461, 0.427429],
+                             [0.1465992, 0.84593179, 0.79033433, 0.576709],
+                             [0.6525159, 0.96735703, 0.69833409, 0.117800]])
+    affine = random_array - .5
     affine[3] = [0, 0, 0, 1]
     streamlines = list(move_streamlines(streamlines, affine))
     new = list(target_f(streamlines, mask, affine=affine))
--- a/dipy/tracking/tests/test_utils.py
+++ b/dipy/tracking/tests/test_utils.py
@@ -269,12 +269,10 @@ def _target(target_f, streamlines, voxel_both_true, voxel_one_true,
         assert_raises(ValueError, list, new)
 
     # Test smaller voxels
-    random_array = np.array([[0.2862315, 0.44142904, 0.19837613, 0.422711],
-                             [0.7434331, 0.55335599, 0.18633461, 0.427429],
-                             [0.1465992, 0.84593179, 0.79033433, 0.576709],
-                             [0.6525159, 0.96735703, 0.69833409, 0.117800]])
-    affine = random_array - .5
-    affine[3] = [0, 0, 0, 1]
+    affine = np.array([[.3, 0, 0, 0],
+                       [0, .2, 0, 0],
+                       [0, 0, .4, 0],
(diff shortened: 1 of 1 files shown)


arokem:
If the conditions under which this test fails are realistic, then it is a bug that needs to be fixed. If they are unrealistic, it should be detected and an informative error should be raised.

skoudoro:
They are realistic. However, I do not think it give us a wrong result or there is a bug.  

The input streamlines are close to each other and there is some crossing so I am not surprised that some transformations/deformations put 2 streamlines in the mask instead of 1.

I might miss something but I am surprised that this test was passing until now.

arokem:
I thought that passing the affine as input to the function should undo any transformation/deformation induced by the affine.

dipy/tracking/tests/test_utils.py
```diff
@@ -269,7 +269,11 @@ def _target(target_f, streamlines, voxel_both_true, voxel_one_true,
         assert_raises(ValueError, list, new)
 
     # Test smaller voxels
-    affine = np.random.random((4, 4)) - .5
+    random_array = np.array([[0.2862315, 0.44142904, 0.19837613, 0.422711],
+                             [0.7434331, 0.55335599, 0.18633461, 0.427429],
+                             [0.1465992, 0.84593179, 0.79033433, 0.576709],
+                             [0.6525159, 0.96735703, 0.69833409, 0.117800]])
```
arokem: If you're going to hard code this, you can set the last row to [0, 0, 0, 1] and subtract 0.5 upfront?

skoudoro:
ok,  so I will play more with this function to make sure:

- I do not misunderstand it
- There is no bug

Thank you for the feedback @arokem

Garyfallidis:
Hi @skoudoro, @arokem and @jchoude. The affine used in the test should not be a random matrix which is what it was before. A random matrix is not necessarily an affine matrix even with the 0001 correction of the last row. Serge you can go ahead and use  a more reasonable matrix i.e. 

```
2 0 0 0
0 1 0 0 
0 0 1 0
0 0 0 1
```
and then @jchoude can provide a more advanced version of that at his convenience. Cheers!

arokem:
That makes sense. What are the conditions for a matrix to be affine? Needs to be invertible, what else? 

I believe that the matrix you provided would not do what is intended here, which IIUC was to check what happens with smaller voxels. IIUC, you want diagonal elements that are <1.

Garyfallidis:
Ariel, that was only a suggestion. Be happy to change this to a new version such as
```
.5 0 0 0
0 .5 0 0 
0 0 .4 0
0 0 0 1
```

arokem:
Yup. That would probably work for the intended purpose of that test.

If want to generate random affines, we can also try to follow this (does it
make sense?):
https://math.stackexchange.com/questions/836069/randomly-generating-special-affine-transformations

Garyfallidis:
+1 link looks solid!

skoudoro:
> If want to generate random affines...

A nice utility function, it deserves its PR.

skoudoro:
I just created #1687 as a reminder.

I think this PR is ready to go!

arokem:
Yup. Off we go.

PR #1686, opened 2018-12-10
Standardize workflow

skoudoro:
A quick PR to clean a little bit all workflows

- Standardize some positional arguments: It helps with parsing.
- Docstring correction, it was missing some `optional` keywords
- fix some pep8 issues


Garyfallidis:
Release time! Bring it Serge!

PR #1695, opened 2018-12-17
A few tractometry functions 

arokem:
For tractometry, we want to be able to extract the profiles along the length of a bundle 

This PR introduces a couple of useful functions for that: 

- `bundle_profile` takes a bundle and returns the values along the length of the bundle, resampled to 100 (or some other number of) nodes. 
- `gaussian_weights` allows you to weight the streamlines within a bundle, depending on how far they are from the core of the bundle.
- Finally, `orient_by_streamline` will make sure that a bundle is all oriented in the same direction (using the other functions without doing this will result in erroneous results!), based on the trajectory of a "standard" streamline. 

Thanks to @jyeatman for guidance in implementing these in the prototype implementation we are using in pyAFQ!


dipy/tracking/streamline.py
```diff
@@ -649,7 +650,82 @@ def orient_by_rois(streamlines, roi1, roi2, in_place=False,
...
+        Whether to make the change in-place in the original input
+        (and return a reference), or to make a copy of the list
+        and return this copy, with the relevant streamlines reoriented.
+        Default: False.
+    as_generator : bool
+        Whether to return a generator as output. Default: False
+    affine : ndarray
+        Affine transformation from voxels to streamlines. Default: identity.
+
+    Returns
```
skoudoro: Can you complete this returns

dipy/tracking/streamline.py
```diff
@@ -649,7 +650,82 @@ def orient_by_rois(streamlines, roi1, roi2, in_place=False,
...
+
+    """
+    # Start by resampling into an array, so that distance calculation is easy:
+    fgarray = np.array(set_number_of_points(streamlines,  n_points))
+    std_array = np.array(set_number_of_points([standard],  n_points))
```
skoudoro: I recommend removing this `np.array`

arokem:
I am trying to create an example, which would follow the bundle extraction example, using these bundles to do tractometry. For that, I will also need parameter maps for this subject. 

@Garyfallidis: Which of the HCP subjects was used to create the `fetch_target_tractogram_hcp` tractogram?

arokem:
Sorry - @BramshQamar : I see that you made that tutorial. Do you know which subject `fetch_target_tractogram_hcp` belongs to?

BramshQamar:
> Sorry - @BramshQamar : I see that you made that tutorial. Do you know which subject `fetch_target_tractogram_hcp` belongs to?

Hello @arokem, we used subject 100206 from HCP data to create target tractogram for bundle extraction example.

dipy/tracking/streamline.py
```diff
@@ -802,5 +878,114 @@ def values_from_volume(data, streamlines, affine=None):
...
+        Weights for each node in each streamline, calculated as its relative
+        inverse of the Mahalanobis distance, relative to the distribution of
+        coordinates at that node position across streamlines.
+    """
+    if isinstance(bundle, np.ndarray):
+        # It's an array, go with it:
+        n_points = bundle.shape[1]
+    else:
+        # It's something else, assume that it needs to be resampled:
+        bundle = np.array(set_number_of_points(bundle, n_points))
```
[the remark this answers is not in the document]
arokem: I believe this makes some of the operations below impossible. See L925, for example.

dipy/tracking/streamline.py
arokem: Resolved in d18c9e84

dipy/tracking/streamline.py
```diff
@@ -802,5 +881,119 @@ def values_from_volume(data, streamlines, affine=None):
...
+        if return_mahalnobis:
+            return np.array([np.nan])
+        else:
+            return np.array([1])
+
+    for node in range(bundle.shape[1]):
+        # This should come back as a 3D covariance matrix with the spatial
+        # variance covariance of this node across the different streamlines
+        # This is a 3-by-3 array:
+        node_coords = bundle[:, node]
```
arokem: Sorry - it moved now. This is the line I referred to in my previous comment (https://github.com/nipy/dipy/pull/1695/files#r245489936). I don't think that I would be able to do this with something that isn't an array. But I'd be happy to learn that I am wrong.

dipy/tracking/streamline.py
```diff
@@ -802,5 +878,114 @@ def values_from_volume(data, streamlines, affine=None):
...
+        for fn in range(bundle.shape[0]):
+            # calculate Mahalanobis for node on fiber[fn]
+            w[fn, node] = mahalanobis(node_coords[fn], m, np.linalg.inv(c))
+    if return_mahalnobis:
+        return w
+    # weighting is inverse to the distance (the further you are, the less you
+    # should be weighted)
+    w = 1 / w
+    # Normalize before returning, so that the weights in each node sum to 1:
+    return w / np.sum(w, 0)
```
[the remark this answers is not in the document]
arokem: Do you have something particular in mind? 

I think that Mahalnobis is strictly non-negative, and so I think that all the mahalnobis distances on a node would only be 0 if all the streamlines in the bundle are identical in that node. 

I am adding a test for that situation, and that throws a linalg error earlier in the function earlier on, because the covariance matrix is all zeros. I think that I should catch it there instead.

The question is what to do in this case. I think that the right thing would be to give each streamline and identical weight that is 1/n_streamlines, which would mean that they would all contribute equally and sum to the value in that voxel. So that's what I will implement for now. Let me know if you think that's wrong.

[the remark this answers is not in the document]
arokem: Now implemented in b7f61a5e

dipy/tracking/streamline.py
```diff
@@ -802,5 +878,114 @@ def values_from_volume(data, streamlines, affine=None):
...
+
+def bundle_profile(data, bundle, affine=None, n_points=100,
+                   weights=None):
+    """
+    Calculates a summarized profile of data for a bundle along its length
+
+    Parameters
+    ----------
+    data : 3D volume
+    bundle : StreamLines class instance, list of arrays, or array
```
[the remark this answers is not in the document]
arokem: OK. Changed here and in the docstring above.

dipy/tracking/streamline.py
```diff
@@ -802,5 +878,114 @@ def values_from_volume(data, streamlines, affine=None):
...
+    weights : 1D array or 2D array (optional)
+        Weight each streamline (1D) or each node (2D) when calculating the
+        tract-profiles. Must sum to 1 across streamlines (in each node if
+        relevant).
+
```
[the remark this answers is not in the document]
arokem: Added in d18c9e84

dipy/tracking/streamline.py
```diff
@@ -802,5 +878,114 @@ def values_from_volume(data, streamlines, affine=None):
...
+    # It's already an array
+    if isinstance(bundle, np.ndarray):
+        fgarray = bundle
+    else:
+        # It's some other kind of thing (list, Streamlines, etc.).
+        # Resample each streamline to the same number of points
+        # list => np.array
+        fgarray = np.array(set_number_of_points(bundle, n_points))
+
+    values = values_from_volume(data, fgarray, affine=affine)
```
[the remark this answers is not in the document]
arokem: The purpose of the resampling is to use all of the streamlines in the bundle to sample from the volume. But to do so, we need to make sure that all the streamlines have the same number of nodes, so that they roughly (!) correspond to each other. 

The approach is described in some detail in this paper: https://journals.plos.org/plosone/article?id=10.1371/journal.pone.0049790, which I will add as a reference.

dipy/tracking/streamline.py
```diff
@@ -802,5 +878,114 @@ def values_from_volume(data, streamlines, affine=None):
...
+
+    if weights is None:
+        weights = np.ones(values.shape) / values.shape[0]
```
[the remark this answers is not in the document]
arokem: I believe that this would only happen if there are no streamlines in the bundle. Should we catch that and raise an error at the top?

arokem:
I added an example usage in f727bde2. I had to introduce some changes in the bundle extraction example, because I am reusing the bundles generated in that example.

dipy/tracking/streamline.py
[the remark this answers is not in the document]
skoudoro: > which would mean that they would all contribute equally and sum to the value in that voxel

Nice! I did not think about all of that but it all makes sense to me and I agree.  Thank you for this update!

dipy/tracking/streamline.py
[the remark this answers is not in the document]
skoudoro: Gotcha! it is clear to me for `fgarray` 

I still think that you do not need to cast `set_number_of_points(bundle, n_points)` with  `np.array`.

dipy/tracking/streamline.py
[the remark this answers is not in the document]
skoudoro: yes, it would be great, even if I think it will almost never happen...

doc/examples/bundle_extraction.py
```diff
@@ -92,23 +102,23 @@
...
-                                             slr=True,
-                                             slr_metric='asymmetric',
-                                             pruning_distance='mam')
+recognized_af_l, af_l_labels = rb.recognize(model_bundle=model_af_l,
+                                           model_clust_thr=5.,
+                                           reduction_thr=10,
+                                           reduction_distance='mam',
+                                           slr=True,
+                                           slr_metric='asymmetric',
+                                           pruning_distance='mam')
```
skoudoro: pep8: alignment

doc/examples/bundle_extraction.py
```diff
@@ -136,15 +146,25 @@
...
-                                             slr=True,
-                                             slr_metric='asymmetric',
-                                             pruning_distance='mam')
+recognized_cst_l, cst_l_labels = rb.recognize(model_bundle=model_cst_l,
+                                            model_clust_thr=5.,
+                                            reduction_thr=10,
+                                            reduction_distance='mam',
+                                            slr=True,
+                                            slr_metric='asymmetric',
+                                            pruning_distance='mam')
```
skoudoro: pep8: alignment

dipy/tracking/streamline.py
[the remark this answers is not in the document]
skoudoro: No, looks good! Personally, I prefer to get an error message than a silent empty result.

dipy/tracking/streamline.py
```diff
@@ -802,5 +880,133 @@ def values_from_volume(data, streamlines, affine=None):
...
+        if return_mahalnobis:
+            return np.array([np.nan])
+        else:
+            return np.array([1])
+
+    for node in range(n_points):
+        # This should come back as a 3D covariance matrix with the spatial
+        # variance covariance of this node across the different streamlines
+        # This is a 3-by-3 array:
+        node_coords = bundle.data[node::n_points]
```
skoudoro: `[node::n_points]` Nice! 👍

skoudoro:
After a rebase, this PR is ready to go, it will be good if someone could have a look on this one

arokem:
Now rebased!

arokem:
Any more thoughts here?

arokem:
Oh - sorry about that - should be fixed now.

arokem:
Does anyone else want to take a look? Any reason not to merge this?

skoudoro:
ping @Garyfallidis

arokem:
Looks like a new release of cvxpy is breaking MAPMRI.

skoudoro:
I just did a quick check on `cvxpy` changelog and there are very few changes. Mapmri tests seems really sensitive... I will have a deeper look later

skoudoro:
Hi @arokem, I just created this issue https://github.com/cvxgrp/cvxpy/issues/672 for our `cvxpy` problem. Concerning Travis,  I do not know what should be our strategy to make Travis happy until their fix

arokem:
Pin the cvxpy version to the previous release for now?

arokem:
Implemented in https://github.com/nipy/dipy/pull/1695/commits/cbf9275bf68bf8879416b7bbfa218eb48f2c9212, I think.

skoudoro:
Great! Thanks! we have to put a reminder somewhere for coming back to normal later

arokem:
OK. Now rebased to incorporate #1740

arokem:
@Garyfallidis : do you intend to give this a look through?

dipy/tracking/streamline.py
```diff
@@ -801,5 +879,133 @@ def values_from_volume(data, streamlines, affine=None):
...
+    if return_mahalnobis:
+        return w
+    # weighting is inverse to the distance (the further you are, the less you
+    # should be weighted)
+    w = 1 / w
+    # Normalize before returning, so that the weights in each node sum to 1:
+    return w / np.sum(w, 0)
+
+
+def bundle_profile(data, bundle, affine=None, n_points=100,
```
arokem: From @Garyfallidis : Consider renaming to something that doesn't have `profile` in it (to avoid confusing with profilometry).

doc/examples/bundle_profiles.py
```diff
@@ -0,0 +1,175 @@
+"""
+========================================================
+Tractometry: extracting profiles from segmeneted bundles
```
Garyfallidis: typo

dipy/tracking/streamline.py
arokem: Upon re-reading the [profilometry paper](https://www.ncbi.nlm.nih.gov/pubmed/26667008), I am not convinced that it's all that different. The first step in that analysis is also the extraction of bundle profiles, just like what we're doing here. Then, they do a whole lot of multivariate analysis on top of it, but you could do that with the profiles we generate here as well. 

WDYT, @Garyfallidis?

dipy/tracking/tests/test_streamline.py
```diff
@@ -1184,5 +1261,118 @@ def test_cluster_confidence():
...
+    npt.assert_equal(w, np.ones(w.shape))
+
+    bundle_len_1 = Streamlines([bundle[0]])
+    w = gaussian_weights(bundle_len_1, n_points=10, return_mahalnobis=True)
+    npt.assert_equal(w, np.ones(w.shape) * np.nan)
+
+
+def test_bundle_profile():
+    data = np.ones((10, 10, 10))
+    bundle = Streamlines([np.array([[0, 0., 0],
```
Garyfallidis: Try this
bundle = Streamlines()
bundle.extend(A)
bundle.extend(B)

avoid using lists of numpy arrays.

dipy/tracking/tests/test_streamline.py
```diff
@@ -1184,5 +1261,118 @@ def test_cluster_confidence():
...
+    # Disallow setting weights that don't sum to 1 across fibers/nodes:
+    assert_raises(ValueError, bundle_profile,
+                  data, bundle, affine=None,
+                  n_points=10, weights=np.ones((2, 10)) * 0.6)
+
+    # Test using an affine:
+    affine = np.eye(4)
+    affine[:, 3] = [-1, 100, -20, 1]
+    # Transform the streamlines:
+    x_bundle = Streamlines([sl + affine[:3, 3] for sl in bundle])
```
Garyfallidis: Same here. Use extend. Or even better transform the data directly using
x_bundle._data

doc/examples/bundle_profiles.py
```diff
@@ -0,0 +1,175 @@
...
+"""
+
+References
+----------
+
+.. [Yeatman2012] Yeatman, Jason D., Robert F. Dougherty, Nathaniel J. Myall,
+    Brian A. Wandell, and Heidi M. Feldman. 2012. "Tract Profiles of White
+    Matter Properties: Automating Fiber-Tract Quantification" PloS One 7
+    (11): e49790.
+
```
Garyfallidis: Add missing references. For example reference for QB.

Garyfallidis:
Ping @arokem :)

arokem:
Sorry. I am in grant writing prison until Monday. Will get back to this then

Garyfallidis:
Sure okay. Good luck!

arokem:
@Garyfallidis : I think that I addressed all your comments in https://github.com/nipy/dipy/pull/1695/commits/6f65734e13e9b55d98e8cdc0b73b2ebc5e816afe

Let me know what you think!

doc/examples/bundle_profiles.py
```diff
@@ -0,0 +1,183 @@
+"""
+========================================================
+Tractometry: extracting profiles from segmented bundles
```
Garyfallidis: Suggesting alternative title (this is too generic).

Extracting AFQ-style profiles from segmented bundles.

doc/examples/bundle_profiles.py
```diff
@@ -0,0 +1,183 @@
...
+========================================================
+
+In this example, we will we will extract the values of a statistic from a
+volume, using the coordinates along the length of a bundle. This is called
+tractometry.
```
Garyfallidis: This is a type of tractometry. 
Tractometry does not always measure along the the length of bundles.

doc/examples/bundle_profiles.py
```diff
@@ -0,0 +1,183 @@
...
+
+from dipy.io.streamline import load_trk
+cst_l, hdr = load_trk("CST_L.trk")
+af_l, hdr = load_trk("AF_L.trk")
+
+transform = np.load("slr_transform.npy")
+
+"""
+
+In the next step, we need to make sure that all the streamlines in each bundles
```
Garyfallidis: each bundle

doc/examples/bundle_profiles.py
```diff
@@ -0,0 +1,183 @@
...
+from dipy.segment.clustering import QuickBundles
+
+feature = ResampleFeature(nb_points=100)
+metric = AveragePointwiseEuclideanMetric(feature)
+
+"""
+
+Since we are going to include all of the streamlines in the single cluster
+from the streamlines, we set the threshold to `np.inf`. We pull out the centroid
+as the standard
```
Garyfallidis: full stop missing

doc/examples/bundle_profiles.py
```diff
@@ -0,0 +1,183 @@
...
+cluster_af_l = qb.cluster(model_af_l)
+standard_af_l = cluster_af_l.centroids[0]
+
+"""
+We use the centroid streamline for each atlas bundle as the standard to orient
+all of the streamlines in each bundle from the individual subject. Here, the
+affine used is the one from the transform between the atlas and individual
+tractogram. This is so that the orienting is done relative to the space of the
+individual, and not relative to the atlas space.
+
```
Garyfallidis: remove space

doc/examples/bundle_profiles.py
```diff
@@ -0,0 +1,183 @@
...
+import dipy.tracking.streamline as dts
+
+oriented_cst_l = dts.orient_by_streamline(cst_l, standard_cst_l,
+                                          affine=transform)
+oriented_af_l = dts.orient_by_streamline(af_l, standard_af_l,
+                                         affine=transform)
+
+
+"""
+
```
Garyfallidis: remove space

doc/examples/bundle_profiles.py
```diff
@@ -0,0 +1,183 @@
...
+w_cst_l = dts.gaussian_weights(oriented_cst_l)
+w_af_l = dts.gaussian_weights(oriented_af_l)
+
+
+"""
+And then use the weights to calculate the bundle profiles for each bundle
+"""
+
+profile_cst_l = dts.bundle_profile(fa, oriented_cst_l, affine=img.affine)#,
```
Garyfallidis: rename bundle_profile to afq_profile

Garyfallidis: Please move afq_profile to dipy.segment.bundle module.

Garyfallidis: The afq_profile could have done both the orientation and weights calculation. Not sure why you need independent functions for these.

dipy/align/streamlinear.py
```diff
@@ -290,14 +290,14 @@ def __init__(self, metric=None, x0="rigid", method='L-BFGS-B',
         References
         ----------
         .. [Garyfallidis15] Garyfallidis et al. "Robust and efficient linear
-            registration of white-matter fascicles in the space of
-            streamlines", NeuroImage, 117, 124--140, 2015
+                            registration of white-matter fascicles in the space
```
Garyfallidis: Not sure about the tab shift here. Is it really needed?

dipy/tracking/streamline.py
Garyfallidis: See my other comment below. Suggest renaming to afq_profile and move to dipy.segment.bundles

Garyfallidis: This function could also automatically calculate orient and weights. You could add flags inside this function for that. if orient==True etc.

dipy/tracking/streamline.py
```diff
@@ -801,5 +879,133 @@ def values_from_volume(data, streamlines, affine=None):
...
+
+    Parameters
+    ----------
+    data : 3D volume
+        The statistic to sample with the streamlines.
+    bundle : StreamLines class instance
+        The collection of streamlines (possibly already resampled into an array
+         for each to have the same length) with which we are resampling. See
+         Note below about orienting the streamlines.
+    weights : 1D
```
Garyfallidis: 1D what?

Garyfallidis:
Okay see my last comments. And please correct all last Pep8 errors. You have quite a few. And thanks for all the cool work.

doc/examples/bundle_profiles.py
```diff
@@ -0,0 +1,183 @@
...
+#                                   weights=w_cst_l)
+
+profile_af_l = dts.bundle_profile(fa, oriented_af_l, affine=img.affine)#,
```
Garyfallidis: Comments need deleting

doc/examples/bundle_profiles.py
```diff
@@ -0,0 +1,183 @@
...
+#                                  weights=w_af_l)
+
+
+fig, (ax1, ax2) = plt.subplots(1, 2)
+
+ax1.plot(profile_cst_l)
+ax1.set_ylabel("Fractional anisotropy")
+ax1.set_xlabel("Node")
+ax2.plot(profile_af_l)
+ax2.set_xlabel("Node")
```
Garyfallidis: Both labels say Node!

Garyfallidis: You will probably want to plot the std too as in the paper?

doc/examples/bundle_profiles.py
arokem: What else does tractometry do?

I think the term comes from this ISMRM presentation: https://cds.ismrm.org/protected/11MProceedings/files/678.pdf 

That's also the way it's used in this paper: https://www.frontiersin.org/articles/10.3389/fninf.2017.00042/full

dipy/align/streamlinear.py
arokem: I think you're right!

doc/examples/bundle_profiles.py
arokem: Standard deviation is usually calculated across subjects, so it's zero here.

dipy/tracking/streamline.py
arokem: It makes sense to orient separately, because you might want to use the same bundle to extract different metrics (FA, MD and so forth) and you don't want to have to orient them each time. Also, if you want to save the oriented tracks, you'll need to have them outside of the function. 

That said, you're right that these can be options. Adding as key-word arguments!

doc/examples/bundle_profiles.py
arokem: I've changed this, but I still think that tractometry is the right term here. Let me know what you think.

arokem:
Thanks for the review. I think that I have addressed all the comments. 

I renamed the function `afq_tract_profiles`, which is how they are called in the original Yeatman paper. 

Let me know what you think!

arokem:
Now also rebased.

Garyfallidis:
@arokem ready for final check?

arokem:
I believe so!

dipy/segment/bundles.py
```diff
@@ -632,3 +635,153 @@ def _prune_what_not_in_model(self, model_centroids,
...
+    if return_mahalnobis:
+        return w
+    # weighting is inverse to the distance (the further you are, the less you
+    # should be weighted)
+    w = 1 / w
+    # Normalize before returning, so that the weights in each node sum to 1:
+    return w / np.sum(w, 0)
+
+
+def afq_tract_profile(data, bundle, affine=None, n_points=100,
```
Garyfallidis: You are already in bundles.py no need for tract. I would just say afq_profile.

Garyfallidis:
I added a last comment. Please correct and ping me to merge! :)

Garyfallidis:
In general I use bundles to mean tracts, fascicles, muscle fibers etc. The tools are quite generic so I think bundles is good name to use. I don't think we need to specify tracts anymore. However, in the tutorial it is good to have some anatomical naming standards. Which you do :+1:

arokem:
Yeah - good call. Renamed!

Issue #1698, opened 2018-12-19
test failure in travis


skoudoro:
## Description

Something to investigate. I suppose it is due to a new package release (Numpy or scipy). Below the error:

```
======================================================================
ERROR: dipy.core.tests.test_sphere.test_interp_rbf
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/travis/build/nipy/dipy/venv/lib/python3.4/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/home/travis/build/nipy/dipy/venv/lib/python3.4/site-packages/dipy/core/tests/test_sphere.py", line 375, in test_interp_rbf
    interp_data_a = interp_rbf(data, s0, s1, norm="angle")
  File "/home/travis/build/nipy/dipy/venv/lib/python3.4/site-packages/dipy/core/sphere.py", line 572, in interp_rbf
    **kwargs)
  File "/home/travis/build/nipy/dipy/venv/lib/python3.4/site-packages/scipy/interpolate/rbf.py", line 241, in __init__
    self.nodes = linalg.solve(self.A, self.di)
  File "/home/travis/build/nipy/dipy/venv/lib/python3.4/site-packages/scipy/interpolate/rbf.py", line 247, in A
    r = squareform(pdist(self.xi.T, self.norm))  # Pairwise norm
  File "/home/travis/build/nipy/dipy/venv/lib/python3.4/site-packages/scipy/spatial/distance.py", line 2030, in pdist
    dm[k] = metric(X[i], X[j], **kwargs)
  File "/home/travis/build/nipy/dipy/venv/lib/python3.4/site-packages/dipy/core/sphere.py", line 544, in angle
    xx[np.isnan(xx)] = 0
TypeError: 'numpy.float64' object does not support item assignment
----------------------------------------------------------------------
```

arokem:
@jhlegarreta : I just submitted a PR with a potential fix in #1708, so you'll probably want to rebase all of your PRs (what a 🎄gift!) once that's resolved.

jhlegarreta:
@arokem Sure. Will do once it gets merged. Thanks for the fix and the note.

PR #1700, opened 2018-12-24
DOC: Use triple double-quoted strings for docstrings.

jhlegarreta:
Fix
```
Triple double-quoted strings should be used for docstrings.
```

PEP8 warning.

Indeed, according to
https://www.python.org/dev/peps/pep-0008/

*For triple-quoted strings, always use double quote characters to be
consistent with the docstring convention in PEP 257.*

### DIFF
--- a/dipy/align/imaffine.py
+++ b/dipy/align/imaffine.py
@@ -955,7 +955,7 @@ class AffineRegistration(object):
     def optimize(self, static, moving, transform, params0,
                  static_grid2world=None, moving_grid2world=None,
                  starting_affine=None):
-        r''' Starts the optimization process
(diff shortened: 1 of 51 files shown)


arokem:
Thanks for doing this @jhlegarreta! As you can see, this triggered a more thorough PEP8 checking on every file you touched. Do you want to go ahead and fix these issues as well, while we're here? If you don't, I can also just merge this one as soon as the CI finishes. It all looks good to me.

jhlegarreta:
Let's see how 0487ec2 behaves.

jhlegarreta:
The reported test failure is related to #1698.

skoudoro:

Issue #583, opened 2015-03-02
Make a cython style guide


arokem:
As proposed by @matthew-brett here: 

https://github.com/nipy/dipy/pull/533#issuecomment-76256571

Garyfallidis:
What about this? Pep8 everywhere except from the limitation for the length of lines.
Which is now recommendation only. With nested for loops the length restriction becomes difficult to sustain.

matthew-brett:
I was thinking more in terms of things like:

```
cimport numpy as cnp
import numpy a np
```

Use of npy_intp for anything holding an index or shape value, when to use the ndarray Cython syntax compared to the memoryviews compared to simple C arrays, how to use nogil to root out Python-side stuff and optimize speed, that kind of thing.

arokem:
I, for one, could use all the help I could get on how to write cython code.

@matthew-brett - do you think you could draft up something as a PR?

jhlegarreta:
@skoudoro following what was discussed in today's dev meeting, I would close this issue. IMHO, it is unlikely that it gets more updates. We do have a minimal style guide, and it is a starting point. If at some point we find that it needs more content, we may re-open the issue then or just open a PR with the new content.

skoudoro:
ok, I agree, let's close it.

PR #1706, opened 2018-12-25
BUG: Fix Cython one-liner non-trivial type declaration warnings.

jhlegarreta:
Fix Cython one-liner non-trivial type declaration warnings. Fixes
```
dipy/tracking/distances.pyx:246:14: Non-trivial type declarators in shared
declaration (e.g. mix of pointers and values). Each pointer declaration
should be on its own line.
```

and similars.

### DIFF
--- a/dipy/align/bundlemin.pyx
+++ b/dipy/align/bundlemin.pyx
@@ -377,7 +377,9 @@ def distance_matrix_mdf(streamlines_a, streamlines_b):
         tracksB64[i] = np.ascontiguousarray(streamlines_b[i], dtype=f64_dt)
     # preallocate buffer array for track distance calculations
     cdef:
-        cnp.float64_t *t1_ptr, *t2_ptr, *min_buffer
+        cnp.float64_t *t1_ptr
+        cnp.float64_t *t2_ptr
+        cnp.float64_t *min_buffer
     # cycle over tracks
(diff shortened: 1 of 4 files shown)


jhlegarreta:
The reported test failure is related to #1698.

jhlegarreta:
Ready to be merged then !

arokem:
Sorry for the delay taking a look here. The code itself actually looks more readable like this to me, so I am +1 for this change. Let's see if anyone else has opinions on this. Otherwise, we can merge it at the end of this week.

skoudoro:
+1 for merging. This PR reminds me #583. We should create this document and add this rule

arokem:
OK. Since we have two 👍 for merging, I will go ahead with this. 

Thanks @jhlegarreta!

jhlegarreta:
Thanks both.

I would not mind to contribute with a Cython style guide.

To be honest, I have not written a single Cython line (yet), so I'd need help from you. When opening the DIPY Cython files, PEP8 points at *a few* lines. So having a look at those could be useful when writing the guide if we adopt PEP8 as Elef suggested in #583. We could start with an initial version, and thus, a baseline over which make improvements would exist as we come along with related issues/questions.

So we can discuss further in #583 if you wish.

PR #1710, opened 2018-12-29
DOC: fix typo in instruction below sample snippet 

snapfinger:
Variable names in instruction and sample snippet in homepage are not consistent.


arokem:
Good catch. Thank you!

PR #1711, opened 2018-12-30
WIP: ENH: Improve the point distribution algorithm over the sphere.

### LINKED ISSUE #184: RF/OPT: Use scipy.optimize instead of dipy.core.sphere.disperse_charges
Most methods for optimal single shell and multishell design are based on electrostatic-like energy minimization [1-5]. To that extent, we already have in Dipy two very useful methods: dipy.core.sphere._get_forces and dipy.core.sphere.disperse_charges. I think to begin with, it would be cool to make these functions a bit more versatile so that they can be reused for the implementations of methods in [1-5]. Two major features would be:
1. to generalize repulsion energy from 1/r to 1/r^k, as suggested in [3]
2. to add some weighting matrix, which is the basis of methods in [4, 5]

So the general problem is a constrained optimization problem: the cost function is the potential, the constraints are imposed by the fact that points lie on the surface of the unit sphere. Both the cost function and the constraint have analytical gradient expressions, so I'd suggest using a constrained optimization routine in scipy.optimize. In particular, scipy.optimize.fmin_slsqp has proved useful in my tests.

Backward compatibility with _get_forces is not a problem. Question: Do we need backward compatibility with disperse_charges?

Suggestions welcome! 
1. Jones, D. K., M. A. Horsfield, and A. Simmons. "Optimal strategies for measuring diffusion in anisotropic systems by magnetic resonance imaging." Magn Reson Med 42 (1999).
2. Jansons, Kalvis M., and Daniel C. Alexander. "Persistent angular structure: new insights from diffusion magnetic resonance imaging data." Inverse problems 19.5 (2003): 1031.
3. Papadakis, Nikolaos G., et al. "Minimal gradient encoding for robust estimation of diffusion anisotropy☆." Magnetic resonance imaging 18.6 (2000): 671-679.
4. Cook, Philip A., et al. "Optimal acquisition orders of diffusion‐weighted MRI measurements." Journal of magnetic resonance imaging 25.5 (2007): 1051-1058.
5. Caruyer, Emmanuel, et al. "Design of multishell sampling schemes with uniform coverage in diffusion MRI." Magnetic Resonance in Medicine (2013).

jhlegarreta:
Improve the point distribution algorithm over the sphere according to
@ecaruyer's proposal.

Resolves #184.


jhlegarreta:
@ecaruyer @MrBago could you please review the PR and comment as appropriate? Thanks !

dipy/core/sphere.py
```diff
@@ -498,6 +500,155 @@ def disperse_charges(hemi, iters, const=.2):
...
+    vects = (vects.T / np.sqrt((vects ** 2).sum(1))).T
+    grad = np.zeros((N, N * 3))
+    for i in range(3):
+        grad[:, i * N:(i + 1) * N] = np.diag(vects[:, i])
+    return grad
+
+
+def _get_forces_alt(vects, alpha=2.0, **kwargs):
+    """Electrostatic-repulsion objective function. The alpha parameter
+    controls the power repulsion (energy varies as $1 / r^\alpha$).
```
ecaruyer: What about saying something like "For `alpha = 1.0`, this corresponds to electrostatic interaction energy."?

dipy/core/sphere.py
```diff
@@ -498,6 +500,155 @@ def disperse_charges(hemi, iters, const=.2):
...
+    vects : array-like shape (N * 3)
+        Points on the sphere.
+    Returns
+    -------
+    array-like (N,)
+        Difference between squared vector norms and 1.
+
+    """
+
+    N = vects.shape[0] / 3
```
ecaruyer: This may cause issues with Python 3, you may replace the division operator with // (here and in the remaining of the code).

dipy/core/sphere.py
```diff
@@ -498,6 +500,155 @@ def disperse_charges(hemi, iters, const=.2):
...
+    vects : array-like (N * 3)
+        Points on the sphere.
+    Returns
+    -------
+    array-like (N, N * 3)
+        grad[i, j] contains :math:`\partial f_i / \partial x_j`.
+
+    """
+
+    N = vects.shape[0] / 3
```
ecaruyer: Division operator: //

dipy/core/sphere.py
```diff
@@ -498,6 +500,155 @@ def disperse_charges(hemi, iters, const=.2):
...
+        Weight values to the electrostatic energy.
+
+    Returns
+    -------
+    energy : float
+        Sum of all interactions between any two vectors.
+
+    """
+
+    nb_points = vects.shape[0] / 3
```
ecaruyer: Division operator: //

dipy/core/sphere.py
```diff
@@ -498,6 +500,155 @@ def disperse_charges(hemi, iters, const=.2):
...
+    # weights = kwargs.get('weights', np.ones((nb_points, nb_points)) / nb_points**2)
```
ecaruyer: You can remove this commented line.

dipy/core/sphere.py
```diff
@@ -498,6 +500,155 @@ def disperse_charges(hemi, iters, const=.2):
...
+        Controls the power of the repulsion. Default is 1.0.
+
+    Returns
+    -------
+    grad : array-like (N * 3,)
+        Gradient of the objective function.
+
+    """
+
+    nb_points = vects.shape[0] / 3
```
ecaruyer: Division operator: //

dipy/core/sphere.py
```diff
@@ -498,6 +500,155 @@ def disperse_charges(hemi, iters, const=.2):
...
+    return forces.reshape((nb_points * 3))
+
+
+def disperse_charges_alt(init_pointset, iters, tol=1.0e-3):
+    """Reimplementation of disperse_charges making use of
+    scipy.optimize.fmin_slsqp.
+
+    Parameters
+    ----------
+    init_pointset : HemiSphere
```
ecaruyer: Technically here I think we are expecting a numpy array, right? Can you update the docstring?

dipy/core/sphere.py
```diff
@@ -498,6 +500,155 @@ def disperse_charges(hemi, iters, const=.2):
...
+        Distributed points on a unit sphere.
+
+    """
+
+    K = init_pointset.shape[0]
+    vects = optimize.fmin_slsqp(_get_forces_alt, init_pointset.reshape(K * 3),
+                                f_eqcons=_equality_constraints,
+                                fprime=_get_grad_forces_alt, iter=iters, acc=tol,
+                                args=[], iprint=0)
+    return vects
```
ecaruyer: Here you may want to return vects.reshape((K, 3))

dipy/core/sphere.py
```diff
@@ -498,6 +500,155 @@ def disperse_charges(hemi, iters, const=.2):
...
+    array-like (N * 3,)
+        Distributed points on a unit sphere.
+
+    """
+
+    K = init_pointset.shape[0]
+    vects = optimize.fmin_slsqp(_get_forces_alt, init_pointset.reshape(K * 3),
+                                f_eqcons=_equality_constraints,
+                                fprime=_get_grad_forces_alt, iter=iters, acc=tol,
+                                args=[], iprint=0)
```
ecaruyer: I think in newer versions of `scipy.optimize`, `args` is a `tuple`, not a `list`. Giving a `list` as input causes a `TypeError: can only concatenate tuple (not "list") to tuple`

ecaruyer:
Thanks @jhlegarreta for this PR. I started reviewing the code and it already looks great! This would be nice to have feedback from a Dipy core developer at some point, to. Maybe @arokem?

Did you consider including some profiling, as mentioned by @arokem (https://github.com/nipy/dipy/issues/184#issuecomment-450020514)?

dipy/core/sphere.py
```diff
@@ -498,6 +500,155 @@ def disperse_charges(hemi, iters, const=.2):
...
 
 
+def _equality_constraints(vects, *args):
+    """Spherical equality constraint. Returns 0 if vects lies on the unit
+    sphere.
+
+    Parameters
+    ----------
+    vects : array-like shape (N * 3)
+        Points on the sphere.
```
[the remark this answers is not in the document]
jhlegarreta: Addressed in b693c60. Did not add it in every method added. If you think I should, please let me know.

dipy/core/sphere.py
jhlegarreta: Addressed in b693c60.

dipy/core/sphere.py
jhlegarreta: Nice catch. Thanks. Fixed in f208032.

dipy/core/sphere.py
jhlegarreta: Nice catch. Thanks. Fixed in f208032.

dipy/core/sphere.py
jhlegarreta: Nice catch. Thanks. Fixed in f208032.

dipy/core/sphere.py
jhlegarreta: Nice catch. Thanks. Fixed in f208032.

dipy/core/sphere.py
jhlegarreta: Removed in 11dcd9c.

dipy/core/sphere.py
jhlegarreta: True. Thanks. Fixed in b693c60.

dipy/core/sphere.py
jhlegarreta: Not sure. I guess `optimize.fmin_slsqp` returns an array that has the same shape as
its `x0` parameter (`init_pointset.reshape(K * 3)` in this case)? Am I wrong?

dipy/core/sphere.py
jhlegarreta: Nice catch ! Yes, I checked that it is the case for the version required by DIPY as well. Thanks Emmanuel. Fixed in 059a9b8.

jhlegarreta:
> Did you consider including some profiling, as mentioned by @arokem (#184 (comment))?

I had not considered adding some profiling. I have not used profiling tools so far, but I would be willing to. I think it is a valuable addition. So if you have pointers, I'd be grateful.

Also, I guess we should have these methods tested/exercised, right? Do you have any proposal for this?

jhlegarreta:
Rebased on `master` to fix merge conflicts.

dipy/core/sphere.py
ecaruyer: > Not sure. I guess `optimize.fmin_slsqp` returns an array that has the same shape as its `x0` parameter (`init_pointset.reshape(K * 3)` in this case)? Am I wrong?

Yes, I think we reshape `x0` to a 1D array because `scipy.optimize` requires this. Don't you think it would be convenient to return here a reshaped array, since we are dealing with 3d points in space?


> I had not considered adding some profiling. I have not used profiling tools so far, but I would be willing to. I think it is a valuable addition. So if you have pointers, I'd be grateful.

I am not an expert in profiling, either... let's wait for suggestions by @arokem


> Also, I guess we should have these methods tested/exercised, right? Do you have any proposal for this?

A simple test I can think of is:
1. generate a random set of points, 
2. compute the associated electrostatic potential
3. disperse charges
4. verify that the potential of the optimal configuration is smaller than that of the original configuration

jhlegarreta: Thanks for the follow-up @ecaruyer !

> Yes, I think we reshape x0 to a 1D array because scipy.optimize requires this. Don't you think it would be convenient to return here a reshaped array, since we are dealing with 3d points in space?

Agreed. Done in 5e3bd7d.

>  Also, I guess we should have these methods tested/exercised, right? Do you have any proposal for this?
A simple test I can think of is:
...

Please check if 989b512 is according to expected.

jhlegarreta:
2625b19 rebased on `master` to fix merge conflicts.

@ecaruyer would you take a look to see what is missing?

If the method is to be used in the code/eventually substitute the current `disperse_charges` method, @arokem what do you propose to measure? You mentioned profiling. Any other aspects? Would you elaborate on that, please?

jhlegarreta:
a14777d should fix CI builds.

@ecaruyer anything else missing?

As for the profiling, suggestions and comments are welcome: should we reuse the [profiling script](https://github.com/ecaruyer/dipy/blob/profile-constrained-optimization/scratch/profile_generalized_electrostatic/compare.py) in @ecaruyer's repository and add it somewhere? If yes, do you happen to have a suggested location for this in the DIPY source tree @arokem?

dipy/core/tests/test_sphere.py
```diff
@@ -357,5 +358,28 @@ def test_disperse_charges():
     nt.assert_array_almost_equal(norms, 1)
 
 
+def test_disperse_charges_alt():
+    # Create a random set of points
+    charges = np.random.random((3, 3))
```
ecaruyer: Here you may want to use `dipy.core.sphere_stats.random_uniform_on_sphere`, which does sample from a uniform distribution on the sphere. 

What about having a variable `nb_points`?

dipy/core/tests/test_sphere.py
```diff
@@ -357,5 +358,28 @@ def test_disperse_charges():
...
+    charges /= vector_norm(charges, keepdims=True)
+    init_pointset = HemiSphere(xyz=charges).vertices
```
ecaruyer: Is this needed?

dipy/core/tests/test_sphere.py
jhlegarreta: :thumbsup: Thanks @ecaruyer ! Makes the test easier to read. changed in ebf9d83.

dipy/core/tests/test_sphere.py
jhlegarreta: Indeed, it was not. It was just making all vectors dwell in the `>0` octant actually. Thanks again.

dipy/core/benchmarks/bench_sphere.py
```diff
@@ -0,0 +1,118 @@
+""" Benchmarks for sphere
+
+Run all benchmarks with::
+
+    import dipy.core as dipycore
+    dipycore.bench()
+
+With Pytest, Run this benchmark with:
+
+    pytest -svv -c bench.ini /path/to/bench_sphere.py
```
jhlegarreta: Not sure the the command works correctly. May be it's just a local issue I'm having.

jhlegarreta:
Added the benchmarking to a `dipy.core.bnechmarks` package using @ecaruyer's script, and produced the following plots:

Electrostatic_repulsion_methods_benchmark


I kept the number of points he used. Tested also with as much as 100, but it looks like his is consistently better.

If I am not missing anything, not sure then what @MrBago concluded is accurate:
> It seems like your method works really well for small point-sets

@ecaruyer's statement seems to hold, though
> This compares convergence versus execution time, and it seems like the gradient descent converges faster for the very first few iterations, but then the convergence becomes very slow. While scipy.optimize routine converges faster to the optimum.

So there it is. If we are satisfied with the `_alt` suffix of the methods, and unless it is decided that
- we need to refactor the methods so that they are called the same and then use either depending on the value of a method e.g. `slsqp` vs. `gradient`, defaulting to one of them.
and/or
- the new proposal should be used for performance issues in the light of results,

The PR is ready to be merged.

MrBago:
Just some context here, when this first came up we had a use case that was very sensitive to execution time. We needed to produce the best possible point sets in a very sort amount of time.

This use case doesn't exist anymore for us and I agree that given any reasonable runtime using scipy optimize produces better point sets.

jhlegarreta:
Thanks @MrBago. Reading your answer and re-examining the results, the statement I made

> (...),  but it looks like his is consistently better.

was may be a bit of an oversimplification: I meant that the new implementation reaches a lower value systematically after a number of computations, but I agree that if the aim is to reach a lower value even at early stages of the computation, the current DIPY implementation performs better.

For a number of points beyond 60 (e.g. 100 that I mentioned) that still holds: the current implementation converges faster at early stages, but the new one achieves a lower value after sufficient time is given.

Thus, even if the use case that you described is no longer an issue for you, it could still be for other groups (?).

So yes, may depend on the number of points, and we may be willing to keep both.

jhlegarreta:
@ecaruyer @arokem 

Would you review the plots and comment on whether we are satisfied with the implementation, please?

Based on the plots, we should decide whether we keep both implementations, and in that case whether we need to refactor the methods so that they are called the same and then use either depending on the value of a method e.g. `slsqp` vs. `gradient`, defaulting to one of them.

The new methods have an `_alt` (for "alternative) suffix, which does not seem to be meaningful.

jhlegarreta:
@ecaruyer please have a looks at 25c15ea . Duplicating references is not something I enjoy but felt like both methods require them. I'd be glad to hear suggestions to keep them just in one part.

jhlegarreta:
@ecaruyer @arokem can the comments be reviewed or shall we have the PR merged?

arokem:
Yes: I think this is ready to merge.

PR #1715, opened 2019-01-04
DOC: Avoid the bullet points being interpreted as quoted blocks.

jhlegarreta:
Avoid the bullet points being interpreted as quoted blocks: the
indentation introduced in some previous commit made the bullet points be
interpreted as quoted blocks.


jhlegarreta:
Not sure why autosection cross-references are not working, though.

skoudoro:
Thanks for this @jhlegarreta

PR #1714, opened 2019-01-04
DOC: Add Cython style guideline for DIPY.

jhlegarreta:
Add Cython style guideline for DIPY.

Fixes #583.

PR #1716, opened 2019-01-04
DOC: Add `Imports` section on package shorthands recommendations.

jhlegarreta:
Add an `Imports` section to the style guideline on package shorthands
recommendations.

### DIFF
--- a/doc/devel/coding_style_guideline.rst
+++ b/doc/devel/coding_style_guideline.rst
@@ -35,6 +35,20 @@ compliance of your code with PEP8. Beyond the aspects checked, as a
 contributor to DIPY, you should try to ensure that your code, including
 comments, conform to the above principles.
 
+Imports
+-------
+DIPY recommends using the following package shorthands to increase consistency
+and readability across the library::
+
+  import nibabel as nib
+  import numpy as np
+  import numpy.testing as npt
+  import scipy as sp
+
+No alias should be used for `h5py`::
+
+  import h5py
+
 -------------
 Documentation
 -------------


jhlegarreta:
Other suggestions are welcome.

doc/devel/coding_style_guideline.rst
```diff
@@ -35,6 +35,15 @@ compliance of your code with PEP8. Beyond the aspects checked, as a
 contributor to DIPY, you should try to ensure that your code, including
 comments, conform to the above principles.
 
+Imports
+-------
+DIPY recommends using the following package shorthands to increase consistency
+and readability across the library::
+
+  import nibabel as nib
```
jhlegarreta: If we prefer using `nb`, or rather than using a per-package rule we'd like the shorthands to be composed by two characters, I'm open to suggestions.

arokem:
On Sat, Jan 5, 2019 at 8:20 AM Jon Haitz Legarreta Gorroño <

> *@jhlegarreta* commented on this pull request.
> ------------------------------
>
> In doc/devel/coding_style_guideline.rst
> <https://github.com/nipy/dipy/pull/1716#discussion_r245481671>:
>
> > @@ -35,6 +35,15 @@ compliance of your code with PEP8. Beyond the aspects checked, as a
>  contributor to DIPY, you should try to ensure that your code, including
>  comments, conform to the above principles.
>
> +Imports
> +-------
> +DIPY recommends using the following package shorthands to increase consistency
> +and readability across the library::
> +
> +  import nibabel as nib
>
> If we prefer using nb, or rather than using a per-package rule we'd like
> the shorthands to be composed by two characters, I'm open to suggestions.
>

The three-letter `nib` is commonly used in several other projects, so I
would rather stick to that.

jhlegarreta:
I do share that, far more clear than `nb`. If other strong preferences exist throughout DIPY, please go ahead and comment.

skoudoro:
I think you can add all default requirements `scipy as sp` and no alias for `h5py`

jhlegarreta:
`cython` is never imported from a Python file, right? For the Cython cases, PR #1714 adds the appropriate section.

Other cases I found and may potentially be added (with modifications if needed):
```
import os.path as op
import scipy.sparse as sps
import scipy.linalg as la
```

skoudoro:
> cython is never imported from a Python file, right?

Yes, you are right

> Other cases I found and may potentially be added

IMHO, I do not think you should add those

jhlegarreta:
Then we're done with this and it can be merged.

doc/devel/coding_style_guideline.rst
```diff
@@ -35,6 +35,20 @@ compliance of your code with PEP8. Beyond the aspects checked, as a
...
+  import numpy as np
+  import numpy testing as npt
```
skoudoro: Can you replace the space by a dot  --> `numpy.testing`

jhlegarreta: Thanks for the catch Sege. Done in 3130522.

PR #1719, opened 2019-01-11
BF: Syntax fix example images not rendering

kesshijordan:
Images aren't rendering in the website gallery (http://nipy.org/dipy/examples_built/path_length_map.html#example-path-length-map). I think this is because of a syntax error (there's an extra space in the examples_index.rst). I also fixed a random typo I found.


skoudoro:
Thank you @kesshijordan

PR #1726, opened 2019-01-24
IVIM MIX Model (MicroLearn)

ShreyasFadnavis:
First Model from the MicroLearn Framework in DIPY -- IVIM Fitting using Variable Projections

Fadnavis S, M. Reisert, H. Farooq, M. Afzali, C. Hu, B. Amirbekian, E. Garyfallidis, MicroLearn: Framework for machine learning, reconstruction, optimization and microstructure modeling, Proceedings of: International Society of Magnetic Resonance in Medicine (ISMRM), Montreal, Canada, 2019.

- [x] Improve the documentation for VarPro Framework


ShreyasFadnavis:
@arokem @skoudoro @Garyfallidis Can you please review this PR and let me know the changes?

Thanks!

dipy/reconst/ivim.py
```diff
@@ -241,11 +271,11 @@ def __init__(self, gtab, split_b_D=400.0, split_b_S0=200., bounds=None,
         elif self.bounds is None:
             self.bounds = ((0., 0., 0., 0.), (np.inf, .3, 1., 1.))
         else:
-            self.bounds = bounds
+            self.bounds = kwargs.get('bounds', None)
 
     @multi_voxel_fit
-    def fit(self, data, mask=None):
-        """ Fit method of the Ivim model class.
+    def fit(self, data):
```
arokem: What happened to the mask?

dipy/reconst/ivim.py
```diff
@@ -509,6 +536,360 @@ def _leastsq(self, data, x0):
...
+
+        self.maxiter = kwargs.get('maxiter', 10)
+        self.xtol = kwargs.get('xtol', 1e-8)
+        self.bvals = gtab.bvals
+        self.yhat_perfusion = np.zeros(self.bvals.shape[0])
+        self.yhat_diffusion = np.zeros(self.bvals.shape[0])
+        self.exp_phi1 = np.zeros((self.bvals.shape[0], 2))
+
+    @multi_voxel_fit
+    def fit(self, data):
```
arokem: We probably want an option to pass a mask here too.

dipy/reconst/ivim.py
```diff
@@ -509,6 +536,360 @@ def _leastsq(self, data, x0):
...
+        x_f = np.zeros(3)
+        x_f[0] = f[0]
+        x_f[1:3] = x
+        return x_f
+
+    def Phi(self, x):
+        """
+        Creates a structure for the combining the diffusion and pseudo-
+        diffusion by multipling with the bvals and then exponentiating each of
+        the twho components for fitting as per the IVIM- two compartment model.
```
arokem: Typo: "twho" => "two"

dipy/reconst/ivim.py
```diff
@@ -509,6 +536,360 @@ def _leastsq(self, data, x0):
...
+        x_f : array
+            Combined array of parameters 'x' and 'f' parameters.
+
+        """
+        x_f = np.zeros(3)
+        x_f[0] = f[0]
+        x_f[1:3] = x
+        return x_f
+
+    def Phi(self, x):
```
arokem: Could we use lower case here? "phi" instead of "Phi"?

doc/examples/reconst_ivim.py
```diff
@@ -128,13 +128,15 @@
 $\mathbf{f}$).
 """
 
-ivimmodel = IvimModel(gtab)
+ivimmodel = IvimModel(gtab, fit_method='LM')
```
arokem: It would be good to also add an example of using the MIX method into this example and maybe compare the two methods to each other.

dipy/reconst/ivim.py
```diff
@@ -509,6 +536,360 @@ def _leastsq(self, data, x0):
                 return x0
 
 
+class IvimModelVP(ReconstModel):
+
+    def __init__(self, gtab, fit_method='VarPro', *args, **kwargs):
+        r""" Initialize an IvimModelVP class.
+
+        The IVIM model assumes that biological tissue includes a volume
```
arokem: Is there some way to avoid the duplication of the parts of the docstring that are identical between the two models?

dipy/reconst/ivim.py
```diff
@@ -509,6 +536,360 @@ def _leastsq(self, data, x0):
...
+        """
+        if SCIPY_LESS_0_15:
+            raise ValueError("Using the Variable Projection Method for " +
+                             "fitting needs SciPy >= 0.15.1")
+        else:
+            from scipy.optimize import differential_evolution
+
+        data_max = data.max()
+        data = data / data_max
+        data[data > 1] = 1
```
arokem: Does this actually ever happen? I thought you just normalized by dividing by the maximum

dipy/reconst/ivim.py
```diff
@@ -509,6 +536,360 @@ def _leastsq(self, data, x0):
...
+        b = self.bvals
+
+        bounds = np.array([(0.005, 0.01), (10**-4, 0.001)])
```
arokem: Should these be configurable? What exactly do these values bound?

ShreyasFadnavis:
> Very well done! I had a few small comments and questions here and there. My main comment is that I would like to see this incorporated into the example. I think that a comparison between the two models (e.g., in terms of model fit accuracy) would be really interesting and should be easy to incorporate into the existing example.

Thank you so much for reviewing my PR!

Regarding the example, I have already created one, but need more time to write the documentation for it since the method is a little different. Here is the link to the example notebook for the MIX method of IVIM that I have created: [https://github.com/ShreyasFadnavis/dipy-tests/blob/master/IVIM_VarPro_DIPY_WS.ipynb](url)

I want this PR to go into the release branch `0.16` and felt that merging the example would get complicated. I will open up a PR for it shortly!

dipy/reconst/ivim.py
ShreyasFadnavis: Thanks, will fix!

dipy/reconst/ivim.py
ShreyasFadnavis: Yes, will do!

dipy/reconst/ivim.py
ShreyasFadnavis: @skoudoro @Garyfallidis ?

dipy/reconst/ivim.py
ShreyasFadnavis: Sorry, I forgot to add documentation for this!

I will do so immediately. And, I do not want them to be configurable because the behavior of `differential evolution` would become intractable! I have set the parameters after a lot of trial and error! 

In the coming days, I plan to Cythonize some parts of the code and then will play around with the parameters again, but for now, due to speed issues and stochastic nature of the optimization method, I want to keep them fixed!

dipy/reconst/ivim.py
ShreyasFadnavis: Will take it down! I didn't realize that it was still there :+1: 
Really good catch!

dipy/reconst/ivim.py
ShreyasFadnavis: Done!

dipy/reconst/ivim.py
ShreyasFadnavis: Done!

dipy/reconst/ivim.py
```diff
@@ -226,13 +252,17 @@ def __init__(self, gtab, split_b_D=400.0, split_b_S0=200., bounds=None,
...
 
         ReconstModel.__init__(self, gtab)
-        self.split_b_D = split_b_D
-        self.split_b_S0 = split_b_S0
-        self.bounds = bounds
-        self.two_stage = two_stage
-        self.tol = tol
-        self.options = options
-        self.x_scale = x_scale
+        self.args = args
```
skoudoro: What is the goal of `self.args` ? It is never used

dipy/reconst/ivim.py
```diff
@@ -226,13 +252,17 @@ def __init__(self, gtab, split_b_D=400.0, split_b_S0=200., bounds=None,
...
+        self.kwargs = kwargs
+        self.fit_method = fit_method
```
skoudoro: `self.fit_method`is never used. Do you really need this variable on this class?

dipy/reconst/ivim.py
```diff
@@ -126,15 +133,34 @@ def f_D_star_error(params, gtab, signal, S0, D):
...
+        return IvimModelVP(gtab, fit_method='LM', *args, **kwargs)
+
+
+IvimModel = ivim_model_selector
+
+
+class IvimModelLM(ReconstModel):
+    """Ivim model
+    """
+    def __init__(self, gtab, fit_method='LM', *args, **kwargs):
```
skoudoro: What is the goal of `fit_method` in this class?  What is the goal of `*args` in this class? 

Furthermore, I think it will be better to be explicit and put all keywords arguments  instead of `**kwargs`.

dipy/reconst/ivim.py
```diff
@@ -226,13 +252,17 @@ def __init__(self, gtab, split_b_D=400.0, split_b_S0=200., bounds=None,
...
-        self.x_scale = x_scale
+        self.args = args
+        self.kwargs = kwargs
+        self.fit_method = fit_method
+        self.split_b_D = kwargs.get('split_b_D', 400.0)
+        self.split_b_S0 = kwargs.get('split_b_S0', 200.)
+        self.bounds = kwargs.get('bounds', None)
+        self.two_stage = kwargs.get('two_stage', True)
+        self.tol = kwargs.get('tol', 1e-15)
+        self.options = kwargs.get('options', {'gtol': 1e-15, 'ftol': 1e-15,
```
skoudoro: What happens if I give this parameter: `options={'eps': 1e15}` ?

The user does not know what keyword should be in options. This is why I think it will be better to be explicit on the `__init__`and then, build your `options` variable.

dipy/reconst/ivim.py
```diff
@@ -509,6 +536,362 @@ def _leastsq(self, data, x0):
                 return x0
 
 
+class IvimModelVP(ReconstModel):
+
+    def __init__(self, gtab, fit_method='VarPro', *args, **kwargs):
```
skoudoro: same as before for  `fit_method` and `*args`

dipy/reconst/ivim.py
```diff
@@ -509,6 +536,362 @@ def _leastsq(self, data, x0):
...
+               (2016).
+
+        """
+        if SCIPY_LESS_0_15:
+            raise ValueError("Using the Variable Projection Method for " +
+                             "fitting needs SciPy >= 0.15.1")
+        else:
+            from scipy.optimize import differential_evolution
+
+        data_max = data.max()
```
skoudoro: I think it will be good to apply the mask here, even if 'multi_voxel_fit`will do it for you later

dipy/reconst/ivim.py
ShreyasFadnavis: The `@multi_voxel` decorator I believe adds the mask and does not need to be included in the fit I guess.. Does that make sense?

-- This is the reason I had taken it down..

dipy/reconst/ivim.py
```diff
@@ -126,15 +133,34 @@ def f_D_star_error(params, gtab, signal, S0, D):
...
-    def __init__(self, gtab, split_b_D=400.0, split_b_S0=200., bounds=None,
-                 two_stage=True, tol=1e-15,
-                 x_scale=[1000., 0.1, 0.001, 0.0001],
-                 options={'gtol': 1e-15, 'ftol': 1e-15,
-                          'eps': 1e-15, 'maxiter': 1000}):
+    if fit_method == 'LM':
+        return IvimModelLM(gtab, fit_method='LM', *args, **kwargs)
+
+    elif fit_method == 'VarPro':
+        return IvimModelVP(gtab, fit_method='LM', *args, **kwargs)
```
skoudoro: it would be good to have an `else` in case there is an error on the `fit_method`. 

You can call again `IvimModelLM(gtab, fit_method='LM', *args, **kwargs)`and add a warning for the wrong choice, selecting the default method

dipy/reconst/ivim.py
```diff
@@ -8,17 +8,22 @@
 import warnings
 from dipy.reconst.base import ReconstModel
 from dipy.reconst.multi_voxel import multi_voxel_fit
+from dipy.utils.optpkg import optional_package
+cvxpy, have_cvxpy, _ = optional_package("cvxpy")
 
 SCIPY_LESS_0_17 = (LooseVersion(scipy.version.short_version) <
                    LooseVersion('0.17'))
 
+SCIPY_LESS_0_15 = (LooseVersion(scipy.version.short_version) <
```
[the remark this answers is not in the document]
skoudoro: I am ok with that

dipy/reconst/ivim.py
ShreyasFadnavis: Will take it down :+1: Thank you for the catch!

dipy/reconst/ivim.py
Garyfallidis: Personally, not aware of a clean way. Future work?

dipy/reconst/ivim.py
```diff
@@ -509,6 +536,362 @@ def _leastsq(self, data, x0):
...
+
+        self.maxiter = kwargs.get('maxiter', 10)
+        self.xtol = kwargs.get('xtol', 1e-8)
+        self.bvals = gtab.bvals
+        self.yhat_perfusion = np.zeros(self.bvals.shape[0])
+        self.yhat_diffusion = np.zeros(self.bvals.shape[0])
+        self.exp_phi1 = np.zeros((self.bvals.shape[0], 2))
+
+    @multi_voxel_fit
+    def fit(self, data, mask=None):
```
Garyfallidis: mask not needed the decorator can deal with that.

dipy/reconst/ivim.py
Garyfallidis: Just add a parameter bounds_de = None and then set the default to these values. Should be no much work.

Garyfallidis: For now ...

dipy/reconst/ivim.py
ShreyasFadnavis: Cool!

dipy/reconst/ivim.py
ShreyasFadnavis: Done!

dipy/reconst/ivim.py
```diff
@@ -126,15 +133,43 @@ def f_D_star_error(params, gtab, signal, S0, D):
...
     """
 
+    if fit_method == 'LM':
+        return IvimModelLM(gtab, **kwargs)
+
+    elif fit_method == 'VarPro':
+        return IvimModelVP(gtab, **kwargs)
+
+    else:
+        print('The fit_method option chosen was not correct. \
```
skoudoro: Can you use `warn` instead of  `print`  like [here](https://github.com/nipy/dipy/blob/ebc9c59e280799dde8ba661aa3fe4cb6387e270c/dipy/reconst/ivim.py#L303)

dipy/reconst/ivim.py
```diff
@@ -509,6 +552,346 @@ def _leastsq(self, data, x0):
                 return x0
 
 
+class IvimModelVP(ReconstModel):
+
+    def __init__(self, gtab):
```
skoudoro: Do not forget `tol` and  `maxiter` with their defaults values.

dipy/reconst/ivim.py
ShreyasFadnavis: Okay!

ShreyasFadnavis: Done!

ShreyasFadnavis:
@arokem I have incorporated all the changes suggested by you, @skoudoro and @Garyfallidis .. Please can you take a look at it and merge if possible?

Let me know if you need any further information on my end!

skoudoro:
Thanks @ShreyasFadnavis for addressing all the issues. merging

PR #1730, opened 2019-02-04
[Fix] Cython syntax error

skoudoro:
This PR should fix the "PRE" Travis build error as you can see [here](https://travis-ci.org/nipy/dipy/jobs/487740657#L1346)

### DIFF
--- a/dipy/denoise/enhancement_kernel.pyx
+++ b/dipy/denoise/enhancement_kernel.pyx
@@ -102,13 +102,13 @@ cdef class EnhancementKernel:
         # if LUT exists, load
         if not force_recompute and os.path.isfile(kernellutpath):
             if verbose:
-                print "The kernel already exists. Loading from " + kernellutpath
+                print("The kernel already exists. Loading from " + kernellutpath)
             self.lookuptable = np.load(kernellutpath)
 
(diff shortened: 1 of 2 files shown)


arokem:
Looks good, of course. Do you understand why we missed this until now?

arokem:
Gotcha. That makes sense! Thanks for fixing this!

arokem:
I'll wait for Travis to finish, just in case anything comes up and then merge, since this is fairly uncontroversial. We're going to run into more and more of this kind of stuff, now that the community is moving away from Python 2. We should probably make a plan to transition away from Python 2 support too. I'll open an issue to discuss.

skoudoro:
It seems that this PR fix the syntax error, but a lot of new Cython errors appears on PRE. Theses errors deserve theirs owns PR's, I think.

arokem:
Hi Serge - sorry for the slow reply. Do you want me to merge this, or wait for the fix on PRE? I am happy either way.

skoudoro:
No worries Ariel, I am not in a hurry. Yes, you can go ahead to merge this, I will create a new PR to understand why the new Cython version does not find all *.pxd files

PR #1733, opened 2019-02-06
Minor Doc Fix

ShreyasFadnavis:
@arokem Let me know if I misunderstood!


arokem:
Yeah. This change is fine. 

One way to think about this is that it is pseudo-diffusion, but it's also pseudo-perfusion,  because it doesn't measure either of these perfectly. 

But you're right that the term "pseudo-diffusion" is more commonly used, so let's go with that!

arokem:
Thanks!

PR #1604, opened 2018-07-31
Cleaned PR for Affine Registration Workflow with Supporting Quality Metrics

parichit:
This work was done as part of the **Google Summer of Code (GSOC-2018) DIPY project titled 'DIPY Workflow(s) and Quality Assurance'**

This work pertains to developing and testing the Affine registration workflow for overlaying bio-medical images in different modes (center of mass, translation, rigid body and affine registration). The modes of registration can be controlled easily via the command line parameters and the relative impact on the results can be assessed quantitatively via the generated output metrics. More information in the metrics can be seen in the code documentation as well as by trying the '-h' or help of the workflow.

This PR contains the stable and clean code for affine registration workflow. This should be merged in place of PR 1581. **PR 1581 must not be merged now**. 

The original PR 1581 with the conversation history, all the original commits, commits for improvement based on community feedback can be seen below:
https://github.com/nipy/dipy/pull/1581

PR #1735, opened 2019-02-07
[Rebase] Affine Registration Workflow with Supporting Quality Metrics 

skoudoro:
The goal of this PR is to clean and rebase #1604. Look at #1604 for more information


dipy/io/image.py
```diff
@@ -24,3 +25,38 @@ def load_nifti(fname, return_img=False, return_voxsize=False,
 def save_nifti(fname, data, affine, hdr=None):
     result_img = nib.Nifti1Image(data, affine, header=hdr)
     result_img.to_filename(fname)
+
+
+def save_affine_matrix(fname, affine):
+    """Save Affive matrix.
```
arokem: Typo: Affine

dipy/align/imaffine.py
```diff
@@ -993,11 +993,20 @@ def optimize(self, static, moving, transform, params0,
...
+            similarity between the images (default 'False').
+            The metric containing optimal parameters and
+            the distance between the images.
 
         Returns
         -------
         affine_map : instance of AffineMap
             the affine resulting affine transformation
+        xopt : optimal parameters
+            the optimal parameters (translation, rotation shear etc.)
```
arokem: How many parameters are there here? In exactly what order are they returned?

dipy/io/image.py
```diff
@@ -24,3 +25,38 @@ def load_nifti(fname, return_img=False, return_voxsize=False,
...
+
+    Parameters
+    ----------
+    fname : string
+        File name to save the affine matrix.
+    affine : numpy array
+        The object containing the affine matrix.
+
+    """
+    np.savetxt(fname, affine)
```
arokem: Why do we need a wrapper for `np.savetxt`? Couldn't we just use that directly?

dipy/workflows/align.py
```diff
@@ -190,3 +197,460 @@ def run(self, static_files, moving_files,
...
+
+    def perform_transformation(self, static, static_grid2world, moving,
+                               moving_grid2world,
+                               affreg, params0, transform, affine):
+
+        """ Function to apply the transformation.
+
+        Parameters
+        ----------
+        static : array, shape (S, R, C) or (R, C)
```
arokem: You might want to make the notation here consistent with the changes here: https://github.com/nipy/dipy/pull/1735/files#diff-74c995598e7ee365915e537084234a8bR962

dipy/workflows/align.py
```diff
@@ -190,3 +197,460 @@ def run(self, static_files, moving_files,
...
+        Check the input metric type.
+
+        Parameters
+        ----------
+        metric: string
+            The similarity metric.
+            (default 'MutualInformation' metric)
+
+        """
+        if metric != 'mi':
```
arokem: What's the purpose of this check?

arokem: I mean: why is this a separate method?

dipy/io/image.py
skoudoro: Even if I have the same opinion, it is always good to have an explicit function name for the end-user. Currently, they have no "explicit official" way to save an affine matrix. I do not have a strict opinion about that, I can remove it. What do you think?

dipy/workflows/align.py
skoudoro: Thanks

dipy/align/imaffine.py
skoudoro: The number of parameters depends on the transformation.
eg. @2D translation -> 2 parameters, 2D Rotation -> 1 parameter, etc........

skoudoro: this flat vector of parameter build the affine_map so maybe we could just add a method in the AffineMap class to get this parameter instead of returning them, but it deserves its own PR

arokem:
This looks good to me. Anyone else want to take a look? If I don't hear anything, I'll go ahead and merge this on Friday.

PR #1738, opened 2019-02-12
[FIX] Python 3 compatibility for some tools script

skoudoro:
This PR is just an update of some useful script because they were only compatible with python 2...

I used the opportunity to fix some pep8.


arokem:
I see that you are still retaining Python 2 compatibility. Should we just go ahead and convert this to drop Python 2 support? That way, we can get rid of the `six` dependency.

skoudoro:
sounds good, I am ok with that

arokem:
LGTM. If no one else has any issues, I can merge this in a couple of days.

PR #1740, opened 2019-02-14
[FIX] mapmri  with cvxpy 1.0.15

skoudoro:
This PR is a workaround to fix our problem with `cvxpy 1.0.15` on Mapmri. More details on this issue https://github.com/cvxgrp/cvxpy/issues/672.

I would like to know by the MapMRI expert (@rutgerfick @RafaelNH @arokem @Garyfallidis) if this approximation is ok or not.

Thank you for your feedback

### DIFF
--- a/dipy/reconst/mapmri.py
+++ b/dipy/reconst/mapmri.py
@@ -395,6 +395,8 @@ class MapmriModel(ReconstModel, Cache):
 
             data_norm = np.asarray(data / data[self.gtab.b0s_mask].mean())
             c = cvxpy.Variable(M.shape[1])
+            # workaround for the bug on cvxpy 1.0.15 when lopt = 0
+            lopt = lopt or 10e-20
             design_matrix = cvxpy.Constant(M)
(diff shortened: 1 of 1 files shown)


arokem:
Would we still need this if we skip this version of cvxpy and pin to 1.0.14 (as in https://github.com/nipy/dipy/pull/1695)?

skoudoro:
I think so, even if we make Travis happy, any new users or if someone works with master and does an update of `cvxpy`. It will crash.

rutgerfick:
Strange bug indeed.

Instead of arbitrarily setting the minimum lopt to 1e-20, perhaps it's cleaner to solve the problem without the regularization term if lopt == 0, and avoid the bug altogether.

skoudoro:
Ok, thank you for your feedback @rutgerfick. implemented on 014b57f

arokem:
+1. Will merge as soon as the CI comes back

arokem:
Thanks for implementing this!

PR #1605, opened 2018-07-31
Cleaned PR for Apply Transform Workflow to quickly register a set of moving NIFTI images using a given affine matrix.

parichit:
This work was done as part of the **Google Summer of Code (GSOC-2018) DIPY project titled 'DIPY Workflow(s) and Quality Assurance'**

This work pertains to extending the work done in PR 1604 by using the affine matrix generated by image registration for transforming a set of images without having to do the image registration again. Evidently, this workflow (ApplyAffineFlow) creates the registered images in a fraction of seconds when compared to the Image registration workflow.

Together the Image Registration Workflow and the Apply Affine Workflow provides the ability to registering a set of images and applying the generated transform to a set of moving images.

This is a clean and stable version of the PR 1598 that got mixed up due to code merge and rebase. **The PR1598 must not be merged now**. Instead, this PR should be merged with the codebase.
 
The link to the older PR (with all the conversations, commit history and improvements made based on community feedback can be seen below):
https://github.com/nipy/dipy/pull/1598

PR #1742, opened 2019-02-15
[REBASED] Apply Transform Workflow

skoudoro:
The goal of this PR is to clean and rebase #1605. Look at #1605 for more information


dipy/io/image.py
```diff
@@ -27,6 +27,16 @@ def save_nifti(fname, data, affine, hdr=None):
     result_img.to_filename(fname)
 
 
+def load_affine_matrix(fname):
```
arokem: This is just a wrapper around `np.loadtxt`. Let's use that one.

skoudoro: thank you! addressed in d90ea0d

dipy/io/image.py
```diff
@@ -27,16 +27,6 @@ def save_nifti(fname, data, affine, hdr=None):
...
-    """
-        Parameters
-        ---------
-        fname : str
-            The file containing the saved affine matrix.
-    """
-    return np.loadtxt(fname)
-
-
 def save_affine_matrix(fname, affine):
```
arokem: While we're here, could you please get rid of `save_affine_matrix`? That looks like a wrapper around `np.savetxt`.

skoudoro: ok, done in 4306c36

arokem:
LGTM. Will merge in a couple of days, unless someone has something to add.

skoudoro:
This PR is ready to go! Does anyone else want to take a look?

PR #1603, opened 2018-07-30
[WIP] - Free water elimination algorithm for single-shell DTI

mvgolub:
This is my attempt at implementing a  free water elimination algorithm for single-shell DTI based on a Beltrami framework described in the work of Pasternak et al. 2009 (https://onlinelibrary.wiley.com/doi/full/10.1002/mrm.22055), it is still a work in progress and not ready to be merged. I have not yet created classes to integrate this with the dipy architecture. I have put this as a separate reconst module, but eventually we could integrate this in the existing fwdti.py module, I look forward on some suggestions on the best way to integrate this code.

This work is part of my thesis in biomedical engineering, one of my supervisors is Rafael Neto Henriques @RafaelNH , who worked  on the multi-shell fwDTI already implemented in dipy.

I apologize for the comments in my code not being very detailed, but I will be responding to the original issue #827 where this feature was requested and provide some documents where I explain the algorithm in more detail and show some results. I appreciate any suggestions to make the code more efficient and solve some numerical problems (overflows, underflows...).

Marc Golub

PR #1744, opened 2019-02-21
[WIP] [NF] Free Water Elimination for single-shell DTI (updated version)

mvgolub:
I previously made a pull request (#1603 ) to add this feature. However, at the time I was still in the middle of working on my thesis and substantial changes to the code were made since then.

Committing all the changes now would make it cumbersome for the reviewer (@RafaelNH ) to follow and understand the workflow behind this feature. Due to this, I am creating this pull request, where I will be updating the newest version of the code by successive follow-up commits, and explain each commit. This time, the code will also be more carefully documented than my last PR. Later, I will close the old PR and delete the branch.

In this initial PR, I added the beltrami.py module, which will contain the standalone functions necessary to perform Beltrami regularization. I added functions that convert the diffusion tensor components to the Iwasawa coordinate system, which is the initial step of the Beltrami routine. Also added the respective test functions in the "test_beltrami.py" file.

I will be adding more functions in the near future.

Regards,

Marc

### DIFF
--- a/dipy/reconst/beltrami.py
+++ b/dipy/reconst/beltrami.py
@@ -183,3 +183,211 @@ def backward_diff(Xi, dim, vox_size):
         dbX = Xi - (Xi[..., shift]) / vox_size
 
     return dbX
+
+
+def beltrami_euclidean(D, mask, zooms, beta):
+    """
+    Computes the Beltrami increments to update the diffusion tensor
+    components, using the euclidean metric [1].
+
+    Parameters
+    ----------
+    D : (..., 6) ndarray
+        The lower triangular components of the diffusion tensor,
+        in the following order: Dxx, Dxy, Dyy, Dxz, Dyz, Dzz.
+    mask : boolean array
+        Boolean mask that marks indices of the data that should be updated.
+    zooms : (3,) tuple of floats
+        Tuple containing the normalized voxel sizes (dx, dy, dz)-
+    beta : float
+        The ratio between the the spatial feature manifold and
+        the image domain metrics, controls how isottropic is the smoothing of the
+        D manifold, see [2] for more details.
+
+    Returns
+    -------
+    bad_g : boolean array
+        Boolean mask that marks the indices of the data where the induced metric
+        tensor "g" is unstable (i.e. the determinant of g is negative or increases to
+        extreme values), and where the update should be null.
+    gsqr : ndarray
+        The square root of the determinant of the metric tensor g.
+    dB : (..., 6) ndarray
+        The increments to be applied to each component of D, computed by applying
+        the Beltrami operator.
+
+    Notes
+    -----
+    In the Beltrami framework for DTI, a 3 by 3 symmetric tensor $g$ is used
+    to measure distances on the 3D image domain, defined (in Einstein summation) by:
+
+        $g_{ij} = \partial_{i} x^{m} \partial_{j} x^{n} h_{mn}$,
+
+    where the $$x^{m}$ denotes the "mth" component of a given manifold
+    (e.g. Dxx of D), the $\partial_{i}$ denotes the derivative along the
+    "ith" axis and $h_{mn}$ is the spatial feature metric tensor.
+    However, if this metric is chosen to be euclidean, then there is no need
+    for the Iwasawa conversion and $h$ becomes a 9 by 9 diagonal matrix, with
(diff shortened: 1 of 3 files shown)


dipy/reconst/beltrami.py
```diff
@@ -0,0 +1,95 @@
...
+    Dxy = D[..., 3]
+    Dxz = D[..., 4]
+    Dyz = D[..., 5]
+
+    out[mask, 0] = Dxx[mask]  # X1
+    out[mask, 3] = Dxy[mask] / Dxx[mask]  # X4
+    out[mask, 4] = Dxz[mask] / Dxx[mask]  # X5
+    out[mask, 1] = Dyy[mask] - out[mask, 0] * out[mask, 3]**2  # X2
+    out[mask, 5] = (Dyz[mask] - out[mask, 0] * out[mask, 3] * out[mask, 4]) / out[mask, 1]  # X6
+    out[mask, 2] = Dzz[mask] - out[mask, 0] * out[mask, 4]**2 - out[mask, 1] * out[mask, 5]**2  # X3
```
RafaelNH: Double checked your iwasawa conversion - all seems correct - well done!

dipy/reconst/beltrami.py
```diff
@@ -0,0 +1,95 @@
...
+    X4 = X[..., 3]
+    X5 = X[..., 4]
+    X6 = X[..., 5]
+
+    out[mask, 0] = X1[mask]
+    out[mask, 1] = X2[mask] + X1[mask] * X4[mask]**2
+    out[mask, 2] = X3[mask] + X1[mask] * X5[mask]**2 + X2[mask] * X6[mask]**2
+    out[mask, 3] = X1[mask] * X4[mask]
+    out[mask, 4] = X1[mask] * X5[mask]
+    out[mask, 5] = X1[mask] * X4[mask] * X5[mask] + X2[mask] * X6[mask]
```
RafaelNH: Here calculations are correct but you have to add then in the "right" order of the diffusion tensor: Dxx, Dxy, Dyy, Dxz, Dyz, Dzz

dipy/reconst/beltrami.py
```diff
@@ -0,0 +1,95 @@
...
+def x_manifold(D, mask, out):
+    """
+    Converts the diffusion tensor components
+    into Iwasawa coordinates, according to [1], [2]
+
+    Parameters
+    ----------
+    D : (..., 6) ndarray
+        The diffusion tensor components, in the following order:
+        Dxx, Dyy, Dzz, Dxy, Dxz, Dyz
```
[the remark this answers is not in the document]
mvgolub: Done.

dipy/reconst/beltrami.py
```diff
@@ -0,0 +1,393 @@
...
+    Dxx = D[..., 0]
+    Dxy = D[..., 1]
+    Dyy = D[..., 2]
+    Dxz = D[..., 3]
+    Dyz = D[..., 4]
+    Dzz = D[..., 5]
+
+    # computing all spatial derivatives of D,
+    # Xi denotes the "ith" components of D (e.g. X1x = d(Dxx) / dx)
+    X1x = forward_diff(Dxx, 0, dx) * mask  # not sure if it is efficient to constantly multiply by this mask,
```
RafaelNH: To avoid multiplying constantly the variables by mask, can just mask once the inputs of this function?

RafaelNH: Will the border derivatives effect a lot the performance of the algorithm?

mvgolub: If I mask the Iwasawa array, it becomes linearized and I loose the spatial information to make the derivatives, unless I use some equivalent function of Matlab's "ind2sub/sub2ind" to store which indices in the linearized array correspond to the shifted indices (n+1) along all dimensions of the 3D array, but those extra operations might end being more costly.

Also, I thought masking was efficient because it was creating a "view" of the original data, like slicing with (:), bu recently I discovered that masking actually makes copies, which is less efficient than creating a view. So multiplying by this mask is not so bad as I thought, at least for the array size that I'm working with.

If the border derivatives are not dealt with, it might lead to unstable "g" metrics in those voxels, that can propagate to neighboring voxels with successive iteartions, however I already have a guard in the code to avoid updating voxels with unstable metrics, so this masking might not be absolutely necessary, but it is always good to have an extra precaution.

skoudoro:
superseded by #2087, so closing

PR #1745, opened 2019-02-21
Adjust number of threads for SLR in Recobundles

frheault:
Pass down the _num_thread_ parameters and fix verbose
No change in behaviour, the default is still all CPU available

Without this option, using recobundles in a pipeline is sub-optimal

### DIFF
--- a/dipy/segment/tests/test_refine_rb.py
+++ b/dipy/segment/tests/test_refine_rb.py
@@ -1,6 +1,8 @@
 import numpy as np
 import nibabel as nib
-from numpy.testing import assert_equal, run_module_suite
+from numpy.testing import (assert_equal,
+                           assert_almost_equal,
+                           run_module_suite)
 from dipy.data import get_fnames
 from dipy.segment.bundles import RecoBundles
 from dipy.tracking.distances import bundles_distances_mam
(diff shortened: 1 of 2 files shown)


arokem:
Looks great. Would you mind adding a test that exercises this functionality? 

Thanks!

frheault:
> Looks great. Would you mind adding a test that exercises this functionality?
> 
> Thanks!

I think I made one that make senses, it run RB twice (one single-thread and the other multi-thread) and the output is expect to be almost equal. The optimizer does not give identical results due to the multi-threading, but the streamline are almost equal (to 5 decimals)

dipy/segment/tests/test_refine_rb.py
```diff
@@ -79,6 +81,33 @@ def test_rb_disable_slr():
...
+                                              model_clust_thr=5.,
+                                              reduction_thr=10,
+                                              slr=True,
+                                              slr_num_threads=1)
+
+    D = bundles_distances_mam(rec_trans_multi_threads, rec_trans_single_thread)
+
+    # check if the bundle is recognized correctly
+    # multi-threading prevent an exact match
+    for row in D:
```
arokem: What type is D at this point? It's not an array? 

Do you understand why multi-threading prevents an exact match? Is there any additional randomness introduced within the algorithm, beyond the control you have via `rng`?

arokem: Hey @frheault : I'd love to be able to merge this in. Any thoughts on these questions?

frheault: D is an array, I used the same test "format" a every other recobundle test. I didn't want to diverge too much from the existing code. That's why I used this verification style.

The randomness is introduced from the optimizer ('L_BFGS_B' or 'Powell') from dipy.core.optimize import Optimizer which ultimately call scipy.optimize import fmin_l_bfgs_b, fmin_powell does not accept fixed initialization or seed.

rng only controls the shuffle and clustering of QBx, so the input to the optimizer is identical. The optimizer reach the same local minima, however the slight difference comes from numerical error due to different initialization for each thread.

arokem: Gotcha. That all makes sense to me. The fact that scipy.optimize has randomization in it is very unfortunate, but out of our control here.

This all looks fine to me. I'm +1 for the merge here. 

Does anyone else want to take a look? If so, please do so in the next couple of days. I'll merge this mid next week, if no one complains before that.

jchoude: LGTM. I can merge if you want, or you can go ahead.

Thanks @frheault

arokem:
Yup. Do it!

jchoude:
It's a go. Thanks @frheault and @arokem

PR #1616, opened 2018-08-13
Implementing the Diffeomorphic registration 

parichit:
This work was done as part of the **Google Summer of Code (GSOC-2018) DIPY project titled 'DIPY Workflow(s) and Quality Assurance'**

This PR applies the diffeomorphic transformation for registering images that differ in size and shape by applying the displacement or diffeomorphic field to the static and moving images. 

The backward displacement field is applied to the static image to warp in the backward direction towards the moving image and the forward diffeomorphic field is applied to the moving image to warp in the forward direction towards the static image.

This PR should be merged only after **PR 1604 and 1605** since it uses common code from them.

Together with the Affine registration workflow the diffeomorphic registration workflow significantly improves the capability of DIPY to register medical images in an automated and quality-controlled manner.

The test case for this PR is still under development and will be committed soon.

Thanks.

PR #1748, opened 2019-03-04
[REBASED ] Symmetric Diffeomorphic registration workflow

skoudoro:
The goal of this PR is to clean and rebase #1616. Look at #1616 for more information. 

I use the opportunity to add more metric from the original PR. It seems that there are no tests so I will add them too on a future commit.


dipy/workflows/align.py
```diff
@@ -698,3 +699,188 @@ def run(self, static_image_file, moving_image_files, affine_matrix_file,
...
+            the displacement field inversion algorithm will stop iterating
+             when the inversion error falls below this threshold.
+
+        out_dir : string, optional
+            Directory to save the transformed files (default '').
+
+        out_warped : string, optional
+            Name of the warped file. If no name is given then a
+             suffix 'transformed' will be appended to the name of the
+             original input file (default 'warped_moved.nii.gz').
```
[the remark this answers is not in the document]
arokem: I think it's probably just the documentation that needs clarification here?

dipy/workflows/align.py
```diff
@@ -698,3 +699,188 @@ def run(self, static_image_file, moving_image_files, affine_matrix_file,
...
+
+        for static_file, moving_file, in_affine, \
+                warped_file, inv_static_file, displ_file in io_it:
+
+            # Loading the image data from the input files into object.
+            static_image, static_grid2world = load_nifti(static_file)
+            moving_image, moving_grid2world = load_nifti(moving_file)
+
+            # Sanity check for the input image dimensions.
+            ImageRegistrationFlow.check_dimensions(static_image, moving_image)
```
arokem: Maybe make the `check_dimensions` a module-wide utility function, rather than a method of the `ImageRegistrationFlow`?

dipy/workflows/align.py
```diff
@@ -698,3 +699,188 @@ def run(self, static_image_file, moving_image_files, affine_matrix_file,
...
+                                                     level_iters)
+
+            mapping = sdr.optimize(static_image, moving_image,
+                                   static_grid2world, moving_grid2world,
+                                   affine_matrix)
+
+            warped_moving = mapping.transform(moving_image)
+
+            # Saving the warped moving file and the alignment matrix.
+            save_nifti(warped_file, warped_moving, static_grid2world)
```
arokem: Should we also save out the transformation? That is, the X by Y by Z by 3 values of the diffeomorphism? We could then apply this transformation to other files, which may be useful. 

We have an implementation of something like that here: 

https://github.com/yeatmanlab/pyAFQ/blob/master/AFQ/registration.py#L138
https://github.com/yeatmanlab/pyAFQ/blob/master/AFQ/registration.py#L153

skoudoro: sounds good, will do

arokem: Great! We'll want to eventually follow up with an API to apply these transforms to input files. But that's probably for another release cycle...

skoudoro: Yes, I think of changing [ApplyAffineFlow](https://github.com/nipy/dipy/blob/master/dipy/workflows/align.py#L645) to make it more generic. I want to call it `ApplyTransformFlow` with an option `--affine` or `--diffeormorphic`. 

I will see if I have time to do that today. Otherwise, next release....

Garyfallidis:
Thanks good job @skoudoro and @parichit !

PR #1753, opened 2019-03-06
BF: Fixes #1751, by adding a `scale` key-word arg to `decfa`

### LINKED ISSUE #1751: Scale values in `decfa`
## Description

The function [dipy.io.utils.decfa](https://github.com/nipy/dipy/blob/def59ac67a0fc4de778eb3398fbaf353dea19237/dipy/io/utils.py#L49) currently expects as input a color FA image that has been scaled, so that it can be cast into the uint8-based dtype that is used in the function itself. 

If users provide the output of our color FA function, the result is most likely to be a volume full of (0, 0, 0) tuples, which is useless. 

We need to either document this behavior or (maybe even better?) scale the data within the function.

arokem:
See #1751

### DIFF
--- a/dipy/io/tests/test_utils.py
+++ b/dipy/io/tests/test_utils.py
@@ -16,3 +16,16 @@ def test_decfa():
     assert data_new.dtype == np.dtype([('R', 'uint8'),
                                        ('G', 'uint8'),
                                        ('B', 'uint8')])
+
+    data_orig = np.zeros((4, 4, 4, 3))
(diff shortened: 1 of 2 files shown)


arokem:
I think that could potentially go wrong in data where the maximal number is smaller than 1.0 

I think that in that case, you'd want the maximal output to be smaller than 255.

skoudoro:
Thanks @arokem, merging

PR #1760, opened 2019-03-11
Implements the inverse of decfa

arokem:
Just for you, @frheault.

### DIFF
--- a/dipy/io/utils.py
+++ b/dipy/io/utils.py
@@ -95,7 +95,8 @@ def decfa(img_orig, scale=False):
 
 def decfa_to_float(img_orig):
     """
-    Create a nifti-compliant directional-encoded color FA file.
+    Convert a nifti-compliant directional-encoded color FA image into a
+    nifti image with RGB encoded in floating point resolution.
 
     Parameters
     ----------
@@ -105,8 +106,7 @@ def decfa_to_float(img_orig):
 
     Returns
     -------
-    img : Nifti1Image class instance with uint8 dtype.
-
+    img : Nifti1Image class instance with float dtype.
 
     Notes
     -----
--- a/dipy/io/utils.py
+++ b/dipy/io/utils.py
@@ -48,11 +48,11 @@ def make5d(input):
 
 def decfa(img_orig, scale=False):
     """
-    Create a nifti-compliant directional-encoded color FA file.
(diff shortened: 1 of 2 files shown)


dipy/io/utils.py
```diff
@@ -91,3 +91,42 @@ def decfa(img_orig, scale=False):
     new_hdr.set_data_dtype(dest_dtype)
 
     return Nifti1Image(out_data, affine=img_orig.affine, header=new_hdr)
+
+
+def decfa_to_float(img_orig):
+    """
+    Create a nifti-compliant directional-encoded color FA file.
```
jchoude: `file` -> `image`

dipy/io/utils.py
```diff
@@ -91,3 +91,42 @@ def decfa(img_orig, scale=False):
...
+
+    Parameters
+    ----------
+    img_orig : Nifti1Image class instance.
+        Contains encoding of the DEC FA image with a 3D volume of data, where
+        each element is a (R, G, B) tuple in uint8.
+
+    Returns
+    -------
+    img : Nifti1Image class instance with uint8 dtype.
```
jchoude: Written that it's going to be in `uint8`, but you convert to `np.float` in the header and the name of the function says so.

dipy/io/utils.py
```diff
@@ -91,3 +91,42 @@ def decfa(img_orig, scale=False):
...
+
+
+    Notes
+    -----
+    For a description of this format, see:
+
+    https://nifti.nimh.nih.gov/nifti-1/documentation/nifti1fields/nifti1fields_pages/datatype.html
```
jchoude: Is it really there? This format is the classical format used by most software, but is not in the standard.

arokem: I am not sure what you mean by "the standard". 

The triplet format is on this wiki page as "DT_RGB".

jchoude: True. Went too fast.

jchoude:
To me this is good to go. What's up with Travis acting up?

arokem:
It ran into both https://github.com/nipy/dipy/issues/1752 and https://github.com/nipy/dipy/issues/1747 in one run. Doesn't look like anything to do with this PR.

jchoude:
Perfect. Let's go then :rocket:

PR #1769, opened 2019-03-13
Change "is" check for 'GCV'

mattcieslak:
This is causing an error in python 3.7


arokem:
Hey @mattcieslak - thanks for fixing this. Under what circumstances do you see this raising an error in Python 3.7? Is this something not currently covered by tests?

mattcieslak:
It works in tests, but the string that ends up being compared in my code comes from a `traits.Str` object (https://github.com/PennBBL/qsiprep/blob/master/qsiprep/interfaces/dipy.py#L216), so checking the address using `is` doesn't work. Using `==` will convert it to a string and works.

arokem:
OK. I figure it's a small change and since it sounds like testing it would be more than we need, I'm just going to go ahead and merge this.

PR #1772, opened 2019-03-14
BF: changes min_signal defaults from 1 to 1e-5

scott-trinkle:
Addresses https://github.com/nipy/dipy/issues/1763

The default value for `min_signal` in all Q-ball models in reconst/shm.py is set to 1. This causes all ODFs to be reconstructed the same if the data happens to be between [0,1] prior to reconstruction. 

I checked the history, and at some point in the past, the default was set to 1e-5. I'm assuming there was a reason it was changed up to 1.0, but as far as I know, changing it back to 1e-5 shouldn't affect the stability.

### DIFF
--- a/dipy/reconst/shm.py
+++ b/dipy/reconst/shm.py
@@ -507,7 +507,7 @@ class SphHarmModel(OdfModel, Cache):
 
 class QballBaseModel(SphHarmModel):
     """To be subclassed by Qball type models."""
-    def __init__(self, gtab, sh_order, smooth=0.006, min_signal=1.,
+    def __init__(self, gtab, sh_order, smooth=0.006, min_signal=1e-5,
                  assume_normed=False):
         """Creates a model that can be used to fit or sample diffusion data
 
@@ -733,7 +733,7 @@ class QballModel(QballBaseModel):
         return dot(data[..., self._where_dwi], self._fit_matrix.T)
 
 
-def normalize_data(data, where_b0, min_signal=1., out=None):
+def normalize_data(data, where_b0, min_signal=1e-5, out=None):
     """Normalizes the data with respect to the mean b0
     """
     if out is None:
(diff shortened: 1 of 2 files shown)


arokem:
Thanks for doing this! Could you please point to the commit where this was changed? It might tell us why this was set to the value.

Also, would you mind adding into the tests ([here](https://github.com/nipy/dipy/blob/master/dipy/reconst/tests/test_shm.py#L197)) a case that fails with the code as it is right now (i.e., with `min_signal = 1`) and is fixed with your suggested fix? 

Thanks again!

scott-trinkle:
No problem! 

As of https://github.com/nipy/dipy/commit/a731dbfb00e17f59b55d0b36a14f69c65e9a849a, `min_signal` was initialized to 1e-5 in `normalize_data`, and it was not a parameter for the other functions. Looks it started being initialized to 1 in https://github.com/nipy/dipy/commit/69429a7670aa8797f756bc24629cbc7c1e7ec0e5.

I'll start working on the tests!

ShreyasFadnavis:
Hi @scott-trinkle!  
Thank you for this... This PR LGTM! @arokem @skoudoro any further suggestions?

skoudoro:
No further suggestions, I am ok with this PR, and you @arokem?

Garyfallidis:
@skoudoro what git blame says for the line that changed? Can we first contact the person that made the change?

skoudoro:
@scott-trinkle said this:

> As of a731dbf, min_signal was initialized to 1e-5 in normalize_data, and it was not a parameter for the other functions. Looks it started being initialized to 1 in 69429a7.

So we need to ask @MrBago for advice and if he is ok with this change.

Garyfallidis:
That is the correct action. Yes. So @MrBago what do you think?

MrBago:
I picked 1e-5 for the same reason, it allows these functions to work with data that's normalized to [0, 1]. I haven't seen there be an issue with the min_signal being too low, 1e-5 seems to work fine with data that's in the int16 range, which how many scanners save their data.

MrBago:
I looked through the commit to see if I could remember why it got changed to 1, I can't find the reason :(

Garyfallidis:
Maybe inheritance? Class used somewhere else?

skoudoro:
What is the conclusion of this discussion @Garyfallidis and @MrBago? Can we go ahead and merge this PR? it seems that this change makes sense.

Garyfallidis:
Yes, go ahead.

skoudoro:
Thank @scott-trinkle for this!

PR #1778, opened 2019-03-15
jupyter notebooks from examples

arokem:
This continues the work in #1022 and supersedes it.

### DIFF
--- a/tools/make_notebooks.py
+++ b/tools/make_notebooks.py
@@ -1,17 +1,33 @@
-def clean_string(teststr):
-    if teststr[:2] == "\n\n":
-        teststr = teststr[2:]
-    if teststr[-2:] == "\n\n":
-        teststr = teststr[:-2]
-    return teststr
+import os
+from nbformat.v4 import new_notebook, new_code_cell, new_markdown_cell
+import nbformat.v4 as nbf
+import codecs
 
 
-def make_notebook(fname):
-    file_path = os.path.join(directory, fname)
-    f = open(file_path, "r")
-    fdata = f.read()
-    f.close()
-    allcells = fdata.split("\"\"\"")
+def clean_string(test_str):
+    """Take a string and remove the newline characters"""
+    if test_str[:2] == "\n\n":
+        test_str = test_str[2:]
+    if test_str[-2:] == "\n\n":
+        test_str = test_str[:-2]
+    return test_str
+
+
+def make_notebook(example):
+    """Generate Ipython notebook of the given example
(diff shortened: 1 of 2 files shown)


arokem:
Still needs some more cleaning up. The rendered notebooks will include (for example) figure blocks from the original rst and other crufy, so we need to process the cells even further. Will follow up with that, before marking this as "ready for review"

tools/make_notebooks.py
```diff
@@ -0,0 +1,166 @@
...
+    textcells = [new_markdown_cell(source=textcells[i])
+                 for i in range(len(textcells))]
+
+    cells = []
+    for i in range(0, len(allcells)):
+        try:
+            cells.append(textcells[i])
+            cells.append(codecells[i])
+        except:
+            pass
```
Borda: just `pass` is not best practice... :)

tools/make_notebooks.py
```diff
@@ -0,0 +1,166 @@
...
+                call the make_notebook function
+
+                Default to ../doc/examples/
+    Returns
+    -------
+    None
+    """
+
+    file_path = os.path.join(directory, fname)
+    f = open(file_path, "r")
```
Borda: rather use `with` statement
```
with open(file_path, 'r') as fp:
    fdata = fp.read()
```

tools/make_notebooks.py
```diff
@@ -0,0 +1,166 @@
...
+
+    nbf.write(notebook, nbname, 4)
+    nbname.close()
+
+
+def valid_examples():
+    """Get the valid examples to be converted"""
+    flist_name = pjoin(os.path.dirname(os.getcwd()), 'doc', 'examples',
+                       'valid_examples.txt')
+    flist = open(flist_name, "r")
```
Borda: again, rather use `with` statement

tools/make_notebooks.py
```diff
@@ -0,0 +1,166 @@
...
+    validated_examples = [line.split("#", 1)[0] for line in validated_examples]
+    # Remove leading and trailing white space from example names
+    validated_examples = [line.strip() for line in validated_examples]
+    # Remove blank lines
+    validated_examples = filter(None, validated_examples)
+    out_examples = []
+    for example in validated_examples:
+        fullpath = pjoin(EG_SRC_DIR, example)
+        if not example.endswith(".py"):
+            print("%s not a python file, skipping." % example)
```
Borda: why not using `logging` module?

PR #1788, opened 2019-03-16
Add test for different dtypes

arokem:
This tests whether https://github.com/nipy/dipy/issues/1783 is still an issue, which I believe is not the case anymore. 🤷‍♂️


skoudoro:
Thanks @arokem! Merging

Issue #1791, opened 2019-03-19
ENH: Save seeds in TRK/TCK


AntoineTheb:
## Description
It has been brought up in meetings that it would be really useful to know where a tract started, for "debugging", analysis and other purposes. For example, to see if an underepresented region in a tractogram is the result of less seeds, or if a streamline going through an "hard" region is the result of a streamline powering through or two separate streamlines.

Using the property `data_per_streamline` of the `Tractogram` class, each seed could be saved with its corresponding streamline using `seeds` as the key. 

Here's an example of its usage, where white dots are a seed:
![image](https://user-images.githubusercontent.com/10272382/54615495-9a5c3600-4a34-11e9-94e4-97c130bc4ca2.png)

I already have much of the work done, **a PR will be opened soon**

GuillaumeTh:
I know @frheault must refactor trk saving. Maybe something like the seed saving could be a important to support.

@frheault do you want to support that in your function ?

skoudoro:
fixed by #1792, closing

PR #1792, opened 2019-03-19
[NF]: Add seeds to TRK

AntoineTheb:
Seeds can now be saved alongside their respective streamlines using the `data_per_streamline` dictionary of the Tractogram class, using `seeds` as a key. Seed return from the `LocalTracker` class is optional so as to prevent breaking changes to pre-existing scripts.

This solves #1791.

--- 

Example usage:
```
from dipy.io.streamline import TrkFile
from dipy.viz import window, actor

# Load files and data
trk = TrkFile.load(args.tractogram)
tractogram = trk.tractogram
streamlines = tractogram._streamlines
seeds = tractogram.data_per_streamline['seeds']

# Make display objects
streamlines_actor = actor.line(streamlines)
points = actor.dots(seeds, color=(1., 1., 1.))

# Add display objects to canvas
s = window.Scene()
s.add(streamlines_actor)
s.add(points)

window.show(s)
```
Which should output something like
image


AntoineTheb:
@skoudoro @Garyfallidis Seems like the test that failed has nothing to do with the current PR (the test is about DKI modeling). Should I raise an issue ?

skoudoro:
I think there is already an issue. I have just restarted the failing Travis matrix.

AntoineTheb:
@skoudoro Any update on this ? What do you think ?

skoudoro:
Hi @AntoineTheb,

This is definitively a nice feature! I need to play a bit with it before reviewing.

it will be good to have other people opinion too like @arokem @guaje @gabknight, ...

When you save the seeds, is the TRK file working correctly with other software like Trackvis?
It would be good to add an option on `save_trk` and `load_trk` function or to create a user-friendly  load/save function

skoudoro:
To clarify my last sentence, we should be able to load/save the seeds without running a tracking algorithm

AntoineTheb:
Thanks ! I just checked it and everything seems normal with Trackvis. Also, it does seem like a good idea to be able to save only the seeds/save them separately. I'll get on it right way.

arokem:
This looks like a good addition to me. +1 to adding I/O options, but that can happen on a follow-up PR.

arokem:
Looks like there are a few conflicts introduced when https://github.com/nipy/dipy/pull/1766 was merged. Could you please resolve these, @AntoineTheb ? Thanks!

dipy/tracking/utils.py
```diff
@@ -921,13 +921,16 @@ def unique_rows(in_array, dtype='f4'):
...
+                     return_seeds=False):
     """Applies a linear transformation, given by affine, to streamlines.
 
     Parameters
     ----------
     streamlines : sequence
         A set of streamlines to be transformed.
+        If return_seeds is True, `streamlines` will be a tuple
+        of (streamlines, seeds)
     output_space : array (4, 4)
```
arokem: We need an additional entry here for `return_seeds`. In particular, it needs to be spelled out that it's optional and that the default is False.

dipy/tracking/utils.py
```diff
@@ -954,8 +958,13 @@ def move_streamlines(streamlines, output_space, input_space=None):
     yield
     # End of initialization
 
-    for sl in streamlines:
-        yield np.dot(sl, lin_T) + offset
+    if return_seeds:
+        # Seeds are already moved
```
arokem: This comment is confusing: are they already moved? Why are you applying the transformation to them?

dipy/tracking/utils.py
AntoineTheb: This is indeed weird, not sure why I left it that way. Will change asap !

dipy/tracking/utils.py
AntoineTheb: Woops, yes at first I thought they would already be in the same space, but when I rechecked with another script they weren't. The transformation does need to be applied so that seeds and streamlines are in the same space. I'll change the comment

AntoineTheb: Actually I'll add a test just to make sure they're in the same space, too.

AntoineTheb: Should be fixed in 07e9d79 !

dipy/tracking/utils.py
AntoineTheb: Changed it so seeds have their own field, should be less awkward now !

AntoineTheb:
@arokem @skoudoro Looks like a single build has failed during the CI run. Could it be possible to retry it ? Thanks !

arokem:
Hey @AntoineTheb : Is this ready to go from your point of view? Would you mind rebasing on the master branch once again?

AntoineTheb:
@arokem Sure, no problem ! And yes I think this is ready to go, extra features like I/O options can be added at a later time

AntoineTheb:
@arokem @skoudoro Any updates on this ? I think the CI errors are unrelated to this PR

skoudoro:
As soon as there is new PR for I/O options, I am ok with this PR

arokem:
Hi @AntoineTheb : sorry for the slowness here. OHBM has been holding us back. 

I think that this is ready to go. I'll post an issue about I/O and assign that to you.

arokem:
Hi @AntoineTheb : sorry for the slowness here. OHBM has been holding us back. 

I think that this is ready to go. I'll post an issue about I/O and assign that to you.

Issue #1793, opened 2019-03-23
ENH: Improving Command Line and WF Docs 


ShreyasFadnavis:
## Description
Discussion issue to collect everyone's opinion and pain points for Workflows in DIPY to for a common consensus.

## Add your ideas/ suggestions in the list below:
- [ ] Add examples with Datasets
- [ ] Explore Sphinx for documentation like `Fury`

jchoude:
As discussed with @skoudoro during the workshop, it would be helpful to have a way to not strip the ` ` (space) character at the end of the description of arguments for workflows. Else, when displayed, some words are together because the line break strips the space.

arokem:
I think that we need to revamp the [command line utility reference](http://nipy.org/dipy/reference_cmd/index.html) so that it shows the name of the command line interface, rather than the WorkFlow object. As it is currently, I think that it's not particularly useful to users, because it doesn't tell them how they would refer to these CLIs. I think that auto-generating these with the CLI names, and possibly also with the help documentation for each CLI would be good.

skoudoro:
citing @Garyfallidis from #1401
> The workflows documentation is very minimal and needs update for example:
We need to explain the other coders that the output parameters should always be after out_dir and the docstrings should have exactly the same order as the input parameters. Consistency, is critical for our workflow system.
Also we need to explain the difference between int and variable int.

skoudoro:
Closing this issue. Many problems have been fixed and this issue starts to be too generic.

PR #1795, opened 2019-03-25
[Fix] workflows description

skoudoro:
This PR fix the problem with the description disappearing on the workflow help. 
Reference: #1793

Before:
image

After:
image


arokem:
Nice. Thanks!

PR #1796, opened 2019-03-27
[Fix] stripping in workflow documentation

skoudoro:
this PR should fix @jchoude [comment](https://github.com/nipy/dipy/issues/1793#issuecomment-476027822) concerning the line break strips the space.
Reference: #1793

Before (look at positional args documentation):

image

After (look at positional args documentation):

image


jchoude:
LGTM! I'll wait until Travis finishes.
Thanks @skoudoro

Issue #1794, opened 2019-03-25
Removed/renamed DetTrackPAMFlow?


mgxd:
## Description
This issue is popping up in [nipype builds](https://circleci.com/gh/nipy/nipype/8384?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

https://github.com/nipy/nipype/blob/232ac23bfd48240906e7e629e98eec854bfdcb1c/nipype/interfaces/dipy/tracks.py#L23

Looks like version `0.16.0` removed `dipy.workflows.tracking.DetTrackPAMFlow` - has this been refactored to https://github.com/nipy/dipy/blob/193a8d95d9b5505a94c14b101c1aab49bb0ec43b/dipy/workflows/tracking.py#L21? cc @skoudoro

skoudoro:
Hi @mgxd,

Indeed, I need to make this smarter. I will create a PR this week to fix that.

thank you for pinging me

PR #1800, opened 2019-03-28
Update command line documentation generation

skoudoro:
# Description

this PR should fix @arokem  [comment](https://github.com/nipy/dipy/issues/1793#issuecomment-476239226) concerning the command line documentation.
Reference: #1793 

- Maybe we need to "cherry-pick" on maint/0.16.x branch and regenerate the documentation after merging.

- We need a new PR to reformat this docstring. It is not beautiful for the moment.

# Result Example:

image

image

### DIFF
--- a/doc/tools/docgen_cmd.py
+++ b/doc/tools/docgen_cmd.py
@@ -2,8 +2,8 @@
 """
 Script to generate documentation for command line utilities
 """
+import os
 from os.path import join as pjoin
-from os import listdir
 import re
 from subprocess import Popen, PIPE, CalledProcessError
 import sys
@@ -60,18 +60,16 @@ def abort(error):
     exit()
 
 
-def get_rst_string(module_name, help_string):
-    """
-    Generate rst text for module
-    """
-    dashes = "========================\n"
+def get_help_string(class_obj):
+    # return inspect.getdoc(class_obj.run)
+    try:
+        ia_module = importlib.import_module("dipy.workflows.base")
+        parser = ia_module.IntrospectiveArgumentParser()
+        parser.add_workflow(class_obj())
+    except Exception as e:
+        abort("Error on {0}: {1}".format(class_obj.__name__, e))
 
-    rst_text = ""
-    rst_text += dashes
-    rst_text += module_name + "\n"
-    rst_text += dashes + "\n"
-    rst_text += help_string
-    return rst_text
+    return parser.format_help()
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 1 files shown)


doc/tools/docgen_cmd.py
```diff
@@ -108,43 +106,62 @@ def get_rst_string(module_name, help_string):
...
+        if module_name is None:
+            continue
+
+        module = importlib.import_module("dipy.workflows." + module_name)
+        members = inspect.getmembers(module)
+        d_wkflw = {name: {"module": obj, "help": get_help_string(obj)}
+                   for name, obj in members
+                   if inspect.isclass(obj) and
+                   issubclass(obj, workflow_module.Workflow) and
+                   name not in ['Workflow', 'CombinedWorkflow']
```
arokem: Maybe put this list somewhere more prominent as a module-level constant? In case we need to find it and add more items in the future?

skoudoro:
> One thing I noticed is that the first word in "positional arguments" and "optional arguments" should be capitalized. Where are these?

They are on this [line](https://github.com/nipy/dipy/pull/1800/files#diff-e06dbcf04a87973e91c474c039d8a9a6R72). I really think that we should reformat the output but I will do it on a new PR

> Regarding what to do now: I suggest that we fix up #1794,

It seems to be already fix and on nipype master. they used a alias (`from ... import ... as ...`)

> and then put out a quick 0.16.1 with this included in the documentation

I do not think we need a quick 0.16.1 since we will start to promote the command line only after the 1.0.0. We still have to improve and fix some workflows

doc/tools/docgen_cmd.py
skoudoro: Good idea, will do, Thanks!

doc/tools/docgen_cmd.py
```diff
@@ -13,6 +13,8 @@
 # version comparison
 from distutils.version import LooseVersion as V
 
+# List of workflows to ignore
+SKIP_WORKFLOWS_LIST = ['Workflow', 'CombinedWorkflow']
```
Borda: use rather tuple as this list should no change
`SKIP_WORKFLOWS_LIST = ('Workflow', 'CombinedWorkflow')`

doc/tools/docgen_cmd.py
```diff
@@ -60,18 +62,16 @@ def abort(error):
...
 
-def get_rst_string(module_name, help_string):
-    """
-    Generate rst text for module
-    """
-    dashes = "========================\n"
+def get_help_string(class_obj):
+    # return inspect.getdoc(class_obj.run)
+    try:
+        ia_module = importlib.import_module("dipy.workflows.base")
```
Borda: why you cannot use `ia_module = dipy.workflows.base` ?

doc/tools/docgen_cmd.py
```diff
@@ -108,43 +108,62 @@ def get_rst_string(module_name, help_string):
...
+    workflow_module = importlib.import_module("dipy.workflows.workflow")
+
+    workflow_flist = [os.path.abspath(pjoin(workflows_folder, f))
+                      for f in os.listdir(workflows_folder)
+                      if os.path.isfile(pjoin(workflows_folder, f)) and
+                      f.lower().startswith("dipy_")]
+
+    workflow_desc = {}
+    # We get all workflows class obj in a dictionnary
+    for f in os.listdir(pjoin('..', 'dipy', 'workflows')):
```
Borda: rather use a longer name which is more describtive `f` -> `path_file`

doc/tools/docgen_cmd.py
```diff
@@ -108,43 +108,62 @@ def get_rst_string(module_name, help_string):
...
+                   issubclass(obj, workflow_module.Workflow) and
+                   name not in SKIP_WORKFLOWS_LIST
+                   }
+
+        workflow_desc.update(d_wkflw)
+
+    cmd_list = []
+    for fpath in workflow_flist:
+        fname = os.path.basename(fpath)
+        with open(fpath) as f:
```
Borda: rather use at least characters `f` -> `fp` as file pointer

doc/tools/docgen_cmd.py
```diff
@@ -108,43 +108,62 @@ def get_rst_string(module_name, help_string):
...
+            flow_name = set(re.findall(r"[A-Z]\w+Flow", f.read(), re.X | re.M))
+
+        if not flow_name or len(flow_name) != 1:
+            continue
+
+        flow_name = list(flow_name)[-1]
+        print("Generating docs for: {0} ({1})".format(fname, flow_name))
+        out_f = fname + ".rst"
```
Borda: rather `out_fname`

doc/tools/docgen_cmd.py
skoudoro: Hi @Borda, Thank you for your review.

I do not understand your point here and how it will work without importing the module. Can you give me more detail?

Borda: the question is why are you using `ia_module = importlib.import_module("dipy.workflows.base")` instead of `ia_module = dipy.workflows.base` ?

skoudoro: I need to import the module. if I do `ia_module = dipy.workflows.base` the module is not imported right or I miss something?

skoudoro:
What is the final decision concerning this PR @arokem?

arokem:
Looks like some of @Borda's comments are still not addressed?

skoudoro:
they are addressed on https://github.com/nipy/dipy/pull/1800/commits/adee03ff7fa0dafc3c124c48b3c4d212c5ee270a . Any opinion @Borda ?

doc/tools/docgen_cmd.py
Borda: what do you mean by right imported? `import dipy.workflows.base` works firn for me...

Borda:
Hi, I made it as comments, not as requirements :) most of my comments were to the PEP8 and clarity of code except the `import dipy.workflows.base`

arokem:
Oh sorry - I didn't notice that these comments were marked "outdated". Merging.

PR #1804, opened 2019-04-12
BF: added check to avoid infinite loop on consecutive coordinates.

jchoude:
In the `_streamline_in_mask` function, if two or more consecutive points shared exactly the same coordinates, the loop would never finish. Added a quick check on the distance to avoid such case.

This is a test that is already present in our in house code and has been tested extensively.

### DIFF
--- a/dipy/tracking/vox2track.pyx
+++ b/dipy/tracking/vox2track.pyx
@@ -267,6 +267,10 @@ cdef cnp.npy_intp _streamline_in_mask(
         direction_norm = norm(direction[0], direction[1], direction[2])
         remaining_distance = direction_norm
 
+        # If consecutive coordinates are the same, skip one.
+        if direction_norm == 0:
+            continue
+
         # Check if it's already a real edge. If not, find the closest edge.
         if floor(current_edge[0]) != current_edge[0] and \
            floor(current_edge[1]) != current_edge[1] and \


jchoude:
@skoudoro Is there something I could do to make this go through? I would really like to have this in master (for my own needs ;) ). The CI failures are unrelated to what I changed and seem to be in the line of the Numpy / Cython issues. 

I know I could merge it, but I don't think that's good style to merge my own changes without any other review (even for 2 lines, like here).

skoudoro:
Hi @jchoude,

Do not worry about the CI error. But,  Can you add a small test for this case? (see @MarcCote comment [here](https://github.com/nipy/dipy/issues/1805#issuecomment-482678785) during your issue #1805 )

And then, I will go ahead and merge it. Thank you

jchoude:
Ok. Just FYI, this is not the fix for #1805, but is still related.

Garyfallidis:
Okay this looks good to me too. Small test all is needed.

skoudoro:
Hi @jchoude, Any chances to have a small test?

jchoude:
Quick question @skoudoro : how can I test an infinite loop condition? If I just call my test on master, it will simply run forever. I'm not aware of facilities to test such a use case...

skoudoro:
Good question 😄! I propose you to just create a small wrapper of your function where you compute the time and, after a certain amount of time, you raise an Exception. In your test, check if this exception is raised or not.  Tell me if I am not clear...

jchoude:
thanks @skoudoro for the suggestion.

However, I feel like this would probably add a lot of complexity to the testing framework. In order for this to work. I would need to spawn the method call in another thread, wait for the timer in the current thread, and after that kill the spawned thread.

It can be done, but don't you fell it makes the test a bit heavyweight?

arokem:
Yeah - I agree. This sounds dangerous and brittle. If we can't come up with
a test for this (I can't), maybe lets go ahead with this.

skoudoro:
> It can be done, but don't you fell it makes the test a bit heavyweight?

ok, you are right, just a bit heavyweight 😄 

> If we can't come up with a test for this (I can't), maybe lets go ahead with this.

Sounds good. Thank you for this PR @jchoude! merging!

PR #1812, opened 2019-04-26
ENH: Statefull tractogram, robust spatial handling and IO

frheault:
The aim of this PR is to provide a easy-to-use and robust way to handle spatial transformation for tractogram.

- The new class use other (existing) IO code, but once created an object remains in a consistent state which make manipulation (transformation, saving, etc.) easier.
- Operations that would lead to a non-respect of conventions are prohibited (invalid state)
- For any atypical operations or operations outside of the conventions, using existing IO (such as Nibabel) is recommended

A future PR will be done later to modify tutorials, remove depreciated code, adapt functions to reduce risky behaviour by promoting the use of this new class.


skoudoro:
wow, Thank you for this @frheault! That's a huge PR! I feel like a part of this code should be directly on Nibabel but I am maybe wrong, I need to look deeper into it.

frheault:
@skoudoro I realize that most python2 build fail for Travis, this is due to a use of Enum in my code.
(Issue explained here: https://stackoverflow.com/questions/26828206/importerror-no-module-named-enum)

What is the 'vision' for Dipy 1.0, will you still support python2 ? If yes, a new requirement is needed (enum34).

skoudoro:
> What is the 'vision' for Dipy 1.0, will you still support python2 ? If yes, a new requirement is needed (enum34).

No, Dipy 1.0 will not support python2. This problem should be fixed when we merge PR #1775. After that, you can rebase. So you can keep your code as it is

jchoude:
Hi @skoudoro ,

if I can bring my point of view WRT the placement of some parts in Nibabel versus here: this is exactly one of the points we wondered about when discussing this PR with @frheault . Our arguments for moving some things to Nibabel were that, since this relates to streamlines, it would make sense to integrate them in Nibabel, to keep them in sync. For example, utilities functions like those used to get the reference info or validate that the `data_per_point` elements fit with the number of points per streamline could be moved over there. 

However, most of the functions and the new class are wrappers over the pure streamlines API, meant to ease use for Dipy users and developers and to reduce possible mistakes. They might be useful directly in the streamlines API, and it would centralize things, but I don't know if they would add too much overhead for users that just want to directly use the Streamlines API. For example, in @frheault code, most of the time, a reference anatomy / tractogram is needed, to correctly manage the reference frame. In `nibabel.streamlines`, the user can directly use the affine and manage it as he sees fit. If we integrate all of this to Nibabel, it needs to be correctly engineered to not remove some freedom from end users.

Part of the reason why this PR is still marked as WIP is because we felt we might need to discuss such questions with the team, since this is a PR that can break things (but since 1.0 might be backwards incompatible, that might not be a huge issue). I think that it would be very helpful to get the opinion of @MarcCote (since he's Mr. Streamlines API), @arokem (since he was interested in this) and @effigies (since he maintains Nibabel). We are absolutely open to discussion. This is an important issue that should be fixed once and for all.

Cheers!

dipy/io/statefull_tractogram.py
```diff
@@ -0,0 +1,879 @@
...
+                                 data_per_point=None,
+                                 data_per_streamline=None):
+        """ Modify streamlines AND data at the same time. Creating a new object
+        would be less risky.
+
+        Setting one of both or both data at the same time is possible.
+        Using overwrite_data without data will erase data (with {})
+
+        Subsampling streamlines from ArraySequence should ALWAYS be done with a
+        deepcopy and pass as a list. ArraySequence views are broken !
```
MarcCote: What do you mean by "ArraySequence views are broken!" Is it something that should be fixed in NiBabel?

frheault: For the array sequence view. If you slices an ArraySequence (something like sub_streamlines = streamlines[0:10]) sub_streamlines.data contains ALL data from streamlines. I called it broken, maybe that's not the right word, it is more like "surprising behaviour" for a lot of people.

I think that this class is needed for Dipy 1.0 in July. Do you think this new class would be acceptable in its current state for Nibabel ?

I am not trying to push anything too fast, but when (if) this is merged there is a lot of work (that I am willing to do) to fix the current handling of spatial transformation and IO.
3D_screenshot
All of these tractograms were computed from the example (and workflows) and they should be aligned (out of 13, 3 display at the right place in MI-Brain, only 2 out of these 3 have a adequate header)

jchoude: The main thing that is "limiting" right now is the support for lazy-loading and lazy-saving. From a "design" POV, it will be hard to maintain a state (like with the current PR) while also supporting lazy operations.

For example, say that we are lazy loading some streamlines. Then, for some reason, in the middle of iterating on the streamlines, some operation is done, like a call to `to_vox`. Then, what should be the behavior? Invalidate the iterator, I guess. This would make sense since we probably don't expect people to call a state changing function during iteration.

I think it would be doable, but would take longer. In the meantime, as @frheault just showed, most of the tutorials / functions are not consistent in Dipy. I would suggest the following plan, if you agree:

1- Integrate everything that is relevant to Nibabel based on the current design (after reviews and refactors if needed)
2- Fix tutorials, functions and examples in Dipy to be consistent and clear on what they expect.
3- [optional] Refactor StatefullTractogram to support lazy operations, if we can reliably handle it.

What do you think, @MarcCote ?

dipy/io/statefull_tractogram.py
```diff
@@ -0,0 +1,879 @@
...
+        if not self._shifted_origin:
+            logging.info('Origin moved to the center of voxel')
+        else:
+            logging.info('Origin moved to the corner of voxel')
+
+        self._shifted_origin = not self._shifted_origin
+
+
+def save_tractogram(sft, filename, bbox_valid_check=True):
+    """ Save the statefull tractogram in any format (trk, tck, fib, dpy)
```
arokem: "statefull" => "stateful"

dipy/io/statefull_tractogram.py
```diff
@@ -0,0 +1,879 @@
...
+
+    Parameters
+    ----------
+    sft : StateFullTractogram
+        The tractogram to save (must have been generated properly)
```
arokem: How do you know if it was properly generated?

dipy/io/statefull_tractogram.py
```diff
@@ -0,0 +1,879 @@
...
+    if old_shift:
+        sft.to_corner()
+
+    return True
+
+
+def load_tractogram(filename, reference, to_space=Space.RASMM,
+                    shifted_origin=False, bbox_valid_check=True,
+                    trk_header_check=True):
+    """ Applies median filter multiple times on input data.
```
arokem: It does what now? I think this first line of the docstring must be a copy-paste error (?)

dipy/io/statefull_tractogram.py
```diff
@@ -0,0 +1,879 @@
...
+        sft.to_voxmm()
+
+    if bbox_valid_check and not sft.is_bbox_in_vox_valid():
+        raise ValueError('Bounding box is not valid in voxel space, cannot ' +
+                         'load a valid file if some coordinates are invalid')
+
+    return sft
+
+
+def robust_split_name(filename):
```
arokem: This function is more broadly useful and should be in the dipy.io base namespace!

dipy/io/vtk.py
```diff
@@ -88,3 +91,115 @@ def save_polydata(polydata, file_name, binary=False, color_array_name=None):
...
+    while current_idx < len(lines_idx):
+        line_len = lines_idx[current_idx]
+
+        next_idx = current_idx + line_len + 1
+        line_range = lines_idx[current_idx + 1: next_idx]
+
+        lines += [lines_vertices[line_range]]
+        current_idx = next_idx
+    if to_lps:
+        to_ras = np.eye(4)
```
arokem: This variable name is confusing. You are transforming this into LPS, right?

dipy/io/tests/test_streamline.py
```diff
@@ -242,6 +246,18 @@ def test_io_dpy():
                                       decimal=4)
 
 
+def test_io_vtk():
+    if have_vtk:
```
arokem: Better to decorate the function with a `@npt.dec.skipif(not have_vtk)`

dipy/io/statefull_tractogram.py
arokem: That's a sobering picture!

Regarding the brokeness in nibabel, this is what I was mentioning here? https://github.com/nipy/nibabel/issues/729

I still owe nibabel a tutorial :-(

MarcCote: @frheault I'm all for making it easier for the users. Those are great wrappers :+1: .

@jchoude I agree with that plan. I think it also goes well with what @arokem proposed below (i.e. test everything out in DIPY 1.0 then move what needs to be moved into Nibabel).

@arokem yes you are right about the issue you mentioned. I wasn't expecting people to call `.data` directly (I don't remember why we didn't make it private in the first place, though) but I guess it is useful after all. That should be addressed in Nibabel, though.

effigies:
Hi all, I haven't read through the code here, but some general thoughts: With regard to nibabel, I'm pretty much in accord with everything @jchoude said. The current streamlines API should continue to be usable, but I don't see any reason that an API for coordinating a reference file would be inappropriate in nibabel. Possibly there are some things that are really dipy-specific, but if you have data structures and algorithms that are usable without the overall dipy framework, nibabel seems a good place for them.

I think probably the biggest question is how you picture the back-and-forth between your development process and nibabel. You'll lose some flexibility by pushing these APIs to an upstream, as you'll be more dependent on our release processes. If you're willing to work on our release schedule (nipy/nibabel#734), then that will work well, but if you need faster iteration, that could grow frustrating for you. So I'd give some consideration to @arokem's suggestion of going a few versions refining the API here before moving upstream. If/when you're confident that it's an API you can use with new features coming only quarterly and bug fixes ~monthly, then that's a good time to move to nibabel.

dipy/io/statefull_tractogram.py
```diff
@@ -0,0 +1,879 @@
...
+            raise TypeError('Streamlines MUST be a list or an ArraySequence')
+        self._streamlines = ArraySequence(streamlines)
+
+        if not _is_data_per_point_valid(streamlines, data_per_point):
+            raise ValueError('Invalid data per point, does not match')
+        if not _is_data_per_streamline_valid(streamlines, data_per_streamline):
+            raise ValueError('Invalid data per streamlines, does not match')
+
+        self._data_per_point = dict(data_per_point)
+        self._data_per_streamline = dict(data_per_streamline)
```
MarcCote: I'm wondering why aren't you using a `Tractogram` object internally? It would manage the `streamlines`, `data_per_point` and `data_per_streamline` for you and handle the space transformations given that you provide the correct affine in the constructor (i.e. which brings the streamlines in "RAS+ MM" space).

dipy/io/statefull_tractogram.py
```diff
@@ -0,0 +1,879 @@
...
+        - dimensions list (3), int, volume shape for each axis
+        - voxel_sizes  list (3), float, size of voxel for each axis
+        - voxel_order, string, Typically 'RAS' or 'LPS'
+    """
+
+    is_nifti = False
+    is_trk = False
+    if isinstance(reference, six.string_types):
+        _, ext = robust_split_name(reference)
+        if ext == '.nii' or ext == '.nii.gz':
```
MarcCote: Personally, I would avoid checking for specific extensions and delegate that to Nibabel. I'd try to do a `nib.load` (or if it fails a `nib.streamlines.load`) and check the type of object that is returned (like you do anyway). I think that makes it more generic since you could provide a BytesIO object and makes it more robust to people messing with extension.

dipy/io/statefull_tractogram.py
MarcCote: I'd delegate this to Nibabel. See my other comment about dealing with filename extensions.

dipy/io/statefull_tractogram.py
skoudoro: I was wondering the same thing, Why your class does not inherit from `Tractogram`?

dipy/io/statefull_tractogram.py
frheault: And then you would continue with the check for instance of Nifti1Image, TrkFile, Nifti1Header, etc. ?

dipy/io/statefull_tractogram.py
```diff
@@ -0,0 +1,879 @@
...
+            Dictionary in which each key has X items
+            X being the number of streamlines
+
+        Notes
+        -----
+        Very important to respect the convention, verify that streamlines
+        match the reference and are effectively in the right space.
+
+        Any change to the number of streamlines, data_per_points or
+        data_per_streamlines requires particular verification.
```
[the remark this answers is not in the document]
frheault: Well I meant to say that changes to these attribute (using the function of the class) will fail if invalid (failing to respect the length mentioned in the description).
What is the optimal way to 'warn' the user that these attributes need to remain consistent with each other ?
(I ask, because I am a bit too much close to the code to know what is clear)

dipy/io/statefull_tractogram.py
jchoude: I'll let @frheault answer this, but a quick answer is that using a Tractogram object internally would allow people to access methods such as `Tractogram.affine_to_rasmm()` and `Tractogram.to_world()`, which would leave the `StatefulTractogram` in an inconsistent state.

I guess that's a limitation of Python, since we can't really make anything 100% private.

Inheriting might be a solution, however.

jchoude:
I agree with @arokem and @effigies proposal of integrating to Dipy, and then migrating to Nibabel when it has been test ran thoroughly.

Additionally, about the stability vs release cycle of Nibabel: the functions in here are meant to cover all cases of space changes. Nothing is purely proper to Dipy: any project interacting with streamlines might want to use them in any of the spaces defined here. Normally, once everything is tested, this should remain extremely stable going forward.

Also, just a note that @frheault 's work is based on work we have been doing in the lab for the last 6,7 years. François and I have broken our teeth many times on funky cases. I'm pretty confident in what is here, and it is based on code we have been using internally since the Streamlines API has been released. Of course, this doesn't replace getting it in the wild and seeing if issues arise. I just post this here to ease some concerns people might have.

dipy/io/statefull_tractogram.py
```diff
@@ -0,0 +1,879 @@
...
+            if self._streamlines.data.size > 0:
+                self._streamlines._data = apply_affine(self._inv_affine,
+                                                       self._streamlines.data)
+                self._space = Space.VOX
+                logging.info('Moved streamlines from rasmm to vox')
+        else:
+            logging.warning('Wrong initial space for this function')
+            return
+
+    def _voxmm_to_rasmm(self):
```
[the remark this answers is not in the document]
frheault: I think that it was clearer in term of code to have every private function on the same pattern.

dipy/io/statefull_tractogram.py
frheault: I could use tractogram to help with the consistency of streamlines and its data_per_streamlines and data_per_points. However for the spatial handling I find the attribute and function related to tractogram to be potentially confusing if access (what JC said)

After it get in Dipy 1.0 (and some real-life testing), if this get moved to Nibabel. It would be interesting to inherit from Tractogram and potentially LazyTractogram. That redesign/refactor is probably a good idea, but maybe later ?

But for now I will look into Tractogram for data/streamlines handling and keep the internal consistency and spatial transformation handling for the stateful_tractogram

arokem:
> After it get in Dipy 1.0 (and some real-life testing), if this get moved to Nibabel. It would be interesting to inherit from Tractogram and potentially LazyTractogram. That redesign/refactor is probably a good idea, but maybe later ?

+1 from me for that plan

dipy/io/statefull_tractogram.py
MarcCote: Yes.

dipy/io/stateful_tractogram.py
```diff
@@ -0,0 +1,746 @@
...
+            self.to_voxmm()
+
+        if not old_shift:
+            self.to_center()
+
+        return is_valid
+
+    def _get_streamline_count(self):
+        """ Safe getter for the number of streamlines """
+        return len(self._tractogram.streamlines._offsets)
```
MarcCote:
```suggestion
        return len(self._tractogram)
```

dipy/io/stateful_tractogram.py
```diff
@@ -0,0 +1,746 @@
...
+
+    def _get_point_count(self):
+        """ Safe getter for the number of streamlines """
+        return len(self._tractogram.streamlines.data)
```
MarcCote: If the `streamlines` object (i.e. an `ArraySequence`) is the result of a slicing or advanced indexing operation (and no `.copy()` was called on it) taking the `len` of `.data)` would be wrong.
```suggestion
        return self._tractogram.streamlines.total_nb_rows
```
Under the hood (see https://github.com/nipy/nibabel/blob/master/nibabel/streamlines/array_sequence.py#L115), it sums the lengths of the remaining sliced/indexed streamlines.

dipy/io/stateful_tractogram.py
```diff
@@ -0,0 +1,746 @@
...
+
+    sft.to_rasmm()
+    sft.to_center()
+
+    timer = time.time()
+    if extension in ['.trk', '.tck']:
+        tractogram_type = detect_format(filename)
+        header = create_tractogram_header(tractogram_type,
+                                          *sft.get_space_attribute())
+        new_tractogram = Tractogram(sft.get_streamlines(),
```
MarcCote: Can't you reuse `self._tractogram`? Why do you need to create a new one?

frheault: I could, I wanted to minimize the use (and visibility) of _tractogram outside of the class.
This function is not part of the class, so it would not be a self and I preferred not to access the private tractogram variable.

Also I decided not to include load/save in the class so they can replace the older function (that have the same name) elsewhere. I think it would be a smoother transition.

frheault: tractogram.affine_to_rasmm can be set properly, but then if someone call .to_world(), tractogram.affine_to_rasmm becomes identity (which would be problematic for the stateful_tractogram). And since I need the voxel size for some transformations anyway, I needed to store variables in the stateful_tractogram.

For technical clarity, I think that stateful_tractogram should store its own "spatial_attribute" and be free of the possibility of someone calling .to_word() that would modifying it.

dipy/io/stateful_tractogram.py
```diff
@@ -0,0 +1,746 @@
...
+
+    if bbox_valid_check and not sft.is_bbox_in_vox_valid():
+        raise ValueError('Bounding box is not valid in voxel space, cannot ' +
+                         'save a valid file if some coordinates are invalid')
+
+    old_space = deepcopy(sft.get_current_space())
+    old_shift = deepcopy(sft.get_current_shift())
+
+    sft.to_rasmm()
+    sft.to_center()
```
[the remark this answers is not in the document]
frheault: Yes exactly, to revert to a state that nibabel expect and thus creating a valid file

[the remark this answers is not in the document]
frheault: Yes exactly, to revert to a state that nibabel expect and thus creating a valid file

dipy/io/stateful_tractogram.py
```diff
@@ -0,0 +1,798 @@
...
+            self.to_rasmm()
+        elif old_space == Space.VOXMM:
+            self.to_voxmm()
+
+        if not old_shift:
+            self.to_center()
+
+        return is_valid
+
+    def remove_invalid_streamlines(self):
```
Garyfallidis: @frheault what is the exact purpose of this method?

dipy/io/stateful_tractogram.py
```diff
@@ -0,0 +1,798 @@
...
+
+        indices_to_save = np.setdiff1d(np.arange(len(self._tractogram)),
+                                       np.array(ic_indices)).astype(int)
+
+        tmp_streamlines = \
+            itemgetter(*indices_to_save)(self.get_streamlines_copy())
+        tmp_data_per_point = {}
+        tmp_data_per_streamline = {}
+
+        for key in self._tractogram.data_per_point:
```
Garyfallidis: What are you trying to do here? What is the purpose?

dipy/io/stateful_tractogram.py
frheault: Sometime directly after generation, or after a transformation, or after cropping a volume, or even numerical error, streamlines end up having invalid coordinates (above the volume dimension or below 0 in voxel space). 
This will cause crash in function such as `target`, `density_map`, `connectivity_map`.

Removing such streamlines yourself can be surprisingly difficult, and since it is technically an invalid state for any tractogram. Having this function in the class will help user to save/load valid object.

PS : Loading and Saving will raise an error (which you can disable with an optional parameters) if trying to load/save an invalid stateful_tractogram.

dipy/io/stateful_tractogram.py
frheault: Considering this function removes streamlines that are invalid, I remove their data (per_point, per_streamline) so the new object is still consistent and valid for the `nibabel.tractogram` object

arokem:
Hey @frheault : I think that something went wrong with your recent rebase. I am now seeing a lot of unrelated commits in this PR.

frheault:
@arokem I rebased on origin instead of upstream... This simple typo is making the whole PR very confusing for me now. I don't know how to revert this action...

arokem:
I'm afraid that you might need to do an interactive rebase. I'd be happy to walk you through this on a video chat if this is your first time doing this.

frheault:
@arokem Sorry I deleted a message almost immediately after posting it...
I think I fixed it using cherry-picking and then force the push. I am up-to-date with master and I see only my 12 commits (and only 6 files changed)

arokem:
Yes. Looks like you are back on track now. Thanks!

frheault:
Failling tests are in the _test_linear_mixed_models_flow_, only for two build. 
Is that considered normal ? (I'm up-to-date with the master)

arokem:
Hey @frheault and others: where does this currently stand? It looks like discussion on several points was adressed. I am wondering what still needs to happen here for us to be able to merge this and start using it.

Also, @frheault : would you mind rebasing here?

frheault:
I have been asking the same questions for 1 month. I think everyone is on vacation. Since there was little comment on the code itself and no external testing I was wondering if it was ready to merge or not. I am currently on vacation, next week I could work on this and finalize it if there is new comments. (and I will rebase as soon as I turn on my laptop, so Monday morning)

skoudoro:
Indeed, with the conferences and vacation, This PR slow down, but it will be **HIGH** priority as soon as you come back. so next week

skoudoro:
Hi @frheault,

Thank you for the rebase! Travis is failing because of this error on many tests:
`cannot import name 'save_trk'`

I advise you to keep these functions `save_trk`, `load_trk`, `load_tck`, etc... They are just small interface to `save/load_tractograms` and they will need a reference too. Even if they all run `save/load_tractogram`, they are convenient for the users.

frheault:
Hi @skoudoro, so I fixed everything that was failing involving loading and saving tractogram. The only thing left I see on my computer is test_peak_dg.py and test_nlmeans.py which does not concern me.

Also, everywhere in the test I am using load_tractogram and save_tractogram (new version) as best as possible considering the context. Right now I updated the tracks300.trk so the header is "valid". BUT it is still an invalid file as the value are actually in no specific space (So I have to turn off the check for validity in voxel_space, but technically the file can be loaded normally anyway)

Then in the workflow such as align (SLR) and segment (RB) I am not using the new stateful_tractogram and instead directly use nibabel as before they were basically doing exactly that and I didn't want to add a new parameters to the flow (as discussed in the video call). The dipy_horizon workflow is the same, but there is a if to support dpy like before.

Finally I added the load_trk, save_trk, etc. version, BUT I added a check so they can load and save only the file format they expect (tiny wrapper instead of an assignation). This is to avoid what was happening very often that people use load_trk in a script and give a tck instead and everything load normally but crash later. Now it crash and tell you that load_trk can only load trk, for more generability use load_tractogram instead.

skoudoro:
Awesome!  Really good job @frheault! Just a quick comment, Can you replace your `assert` by `assert_equal`. This is more explicit and respect the convention. I will review completely this PR tomorrow morning. Thank you again

Garyfallidis:
@frheault I have removed the WIP from the title of your PR as this is getting really close to be merged.

dipy/io/utils.py
```diff
@@ -1,7 +1,9 @@
 """ Utility functions for file formats """
 from __future__ import division, print_function, absolute_import
+import six
```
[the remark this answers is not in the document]
skoudoro: and you can remove `from __future__ import division, print_function, absolute_import`

dipy/io/utils.py
```diff
@@ -130,3 +132,140 @@ def decfa_to_float(img_orig):
...
+    """
+
+    affine_1, dimensions_1, voxel_sizes_1, voxel_order_1 = get_reference_info(
+        reference_1)
+    affine_2, dimensions_2, voxel_sizes_2, voxel_order_2 = get_reference_info(
+        reference_2)
+
+    identical_header = True
+    if not np.allclose(affine_1, affine_2):
+        logging.error('Affine not equal')
```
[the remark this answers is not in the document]
frheault: The function is made to test if the header are identical, it returns a boolean and the testing is usually done outside.
It is useful to give some liberty to users when making verification themselves.
Also the logging is useful to see which (and all) difference. Then the user can do what they want with the true/false return.

dipy/io/utils.py
```diff
@@ -130,3 +132,140 @@ def decfa_to_float(img_orig):
...
+    elif isinstance(reference, nib.streamlines.trk.TrkFile):
+        header = reference.header
+        is_trk = True
+    elif isinstance(reference, nib.nifti1.Nifti1Header):
+        header = reference
+        is_nifti = True
+    elif isinstance(reference, dict) and 'magic_number' in reference:
+        header = reference
+        is_trk = True
+
```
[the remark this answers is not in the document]
frheault: But that's not enough, an affine does not contains de the shape/dimensions information which is important for function such as density_map computation.

dipy/io/utils.py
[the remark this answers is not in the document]
skoudoro: Yeah, sorry, Got it! so that's good

dipy/io/utils.py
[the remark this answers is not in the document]
skoudoro: ok, Got it! I went through again and it makes total sense

skoudoro:
Thank you @frheault for this Huge work! merging

MarcCote:
Great job @frheault

PR #1813, opened 2019-04-30
[Fix] Add Readme for doc generation

### LINKED ISSUE #436: Doc won't build without cvxopt
As an optional dependency, the doc refuses to build if you don't have it installed with an import error. There should be a check in the doc build system to not build that part if the user does not have it, as it won't let you build the rest of the doc and just crash.

skoudoro:
This PR adds a basic readme for doc generation. It is always useful and should fix #436 (really old issue)

### DIFF
--- a/doc/README.MD
+++ b/doc/README.MD
@@ -1,0 +1,37 @@
+# Documentation Generation
+
+## Index
+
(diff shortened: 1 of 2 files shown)


arokem:
Wonderful. Thanks!

PR #1815, opened 2019-04-30
[FIX] median_otsu deprecated parameter

### LINKED ISSUE #1761: Deprecation warning when running median_otsu
C:\ProgramData\Anaconda3\lib\site-packages\dipy\core\histeq.py:22: VisibleDeprecationWarning: Passing `normed=True` on non-uniform bins has always been broken, and computes neither the probability density function nor the probability mass function. The result is only correct if the bins are uniform, when density=True will produce the same result anyway. The argument will be removed in a future version of numpy.

skoudoro:
This PR should fix #1761. As you can see on [numpy documentation](https://docs.scipy.org/doc/numpy/reference/generated/numpy.histogram.html), `normed` parameter has been deprecated and superseded by `density` since a long time.

### DIFF
--- a/dipy/core/histeq.py
+++ b/dipy/core/histeq.py
@@ -19,7 +19,7 @@ def histeq(arr, num_bins=256):
         Histogram equalized image.
     """
     # get image histogram
-    histo, bins = np.histogram(arr.flatten(), num_bins, normed=True)
+    histo, bins = np.histogram(arr.flatten(), num_bins, density=True)
     cdf = histo.cumsum()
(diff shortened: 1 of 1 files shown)


arokem:
Yeah. This looks right to me. But let's wait to see that we don't get anything funky from the CI.

arokem:
I don't think any of the errors on the CI are related to this PR.  I am going to restart those and see whether something else comes up.

arokem:
Ugh. Let's just get rid of Python 2 already.

PR #1819, opened 2019-05-03
[Fix] Diffeormorphic + CCMetric on small image

### LINKED ISSUE #1048: divide by zero error in DiffeomorphicRegistration of small image volumes
I came across the following traceback while using SyN `DiffeomorphicRegistration` registration today.  The error occured because the input image size was too small for the default `level_iters`, but one would be very hard pressed to determine that from the following ZeroDivisionError traceback:

``` python

  File "/home/lee8rx/anaconda/envs/py35mkl/lib/python3.5/site-packages/dipy/align/imwarp.py", line 1461, in optimize
    self._optimize()

  File "/home/lee8rx/anaconda/envs/py35mkl/lib/python3.5/site-packages/dipy/align/imwarp.py", line 1387, in _optimize
    derivative = self._iterate()

  File "/home/lee8rx/anaconda/envs/py35mkl/lib/python3.5/site-packages/dipy/align/imwarp.py", line 1258, in _iterate
    der = self._get_energy_derivative()

  File "/home/lee8rx/anaconda/envs/py35mkl/lib/python3.5/site-packages/dipy/align/imwarp.py", line 1350, in _get_energy_derivative
    y = [v / ss for v in y]

  File "/home/lee8rx/anaconda/envs/py35mkl/lib/python3.5/site-packages/dipy/align/imwarp.py", line 1350, in <listcomp>
    y = [v / ss for v in y]

ZeroDivisionError: float division by zero
```

It turns out that this occurs because the cross-correlation metric can't compute anything meaningful if any of the image dimensions are < `2*CCMetric.radius + 1`.  The current default CCMetric radius is 4 and by default iterations start at an image downsampling factor of 4, so any inputs that have a dimension < 36 will cause this error to occur unless fewer downsampling levels or a smaller radius are manually specified.

The underlying issue is that `compute_cc_forward_step_3d` or  `compute_cc_forward_step_2d` will return only zeros if any of the image dimensions are < 2*radius + 1.
This is to be expected from the description in the docstring:  "The returned vector field will be zero along a boundary of width radius voxels"

A relatively straightforward solution would be to add a dimension check either when calling the CCMetric or when initializing the DiffeoMorphicRegistration (or AffineRegistration) object to raise a more informative error.  If I find time I may make a PR, but it could be a while.

skoudoro:
This PR is just an implementation of [this comment](https://github.com/nipy/dipy/issues/1048#issue-154572349). It should fix #1048.

### DIFF
--- a/dipy/align/metrics.py
+++ b/dipy/align/metrics.py
@@ -244,6 +244,21 @@ class CCMetric(SimilarityMetric):
         re-orienting the gradients in the voxel space using the corresponding
         affine transformations.
         """
+        min_size = self.radius * 2 + 1
+
+        def invalid_image_size(image):
+            valid_shape = [size for size in image.shape if
+                           size >= min_size]
+            return True if len(valid_shape) != image.ndim else False
+
+        msg = ("Each image dimensions should be superior to 2 * radius + 1."
+               "Decrease CCMetric radius or Increase your image size")
+
+        if invalid_image_size(self.static_image):
+            raise IOError("Static image size is too small. " + msg)
+        if invalid_image_size(self.moving_image):
+            raise IOError("Moving image size is too small. " + msg)
+
         self.factors = self.precompute_factors(self.static_image,
                                                self.moving_image,
                                                self.radius)
--- a/dipy/align/metrics.py
+++ b/dipy/align/metrics.py
@@ -244,20 +244,18 @@ class CCMetric(SimilarityMetric):
         re-orienting the gradients in the voxel space using the corresponding
         affine transformations.
         """
-        min_size = self.radius * 2 + 1
 
         def invalid_image_size(image):
-            valid_shape = [size for size in image.shape if
-                           size >= min_size]
-            return True if len(valid_shape) != image.ndim else False
+            min_size = self.radius * 2 + 1
+            return any([size < min_size for size in image.shape])
 
-        msg = ("Each image dimensions should be superior to 2 * radius + 1."
-               "Decrease CCMetric radius or Increase your image size")
+        msg = ("Each image dimension should be superior to 2 * radius + 1."
+               "Decrease CCMetric radius or increase your image size")
 
         if invalid_image_size(self.static_image):
-            raise IOError("Static image size is too small. " + msg)
+            raise ValueError("Static image size is too small. " + msg)
         if invalid_image_size(self.moving_image):
-            raise IOError("Moving image size is too small. " + msg)
+            raise ValueError("Moving image size is too small. " + msg)
 
         self.factors = self.precompute_factors(self.static_image,
                                                self.moving_image,
--- a/dipy/align/tests/test_metrics.py
+++ b/dipy/align/tests/test_metrics.py
@@ -1,3 +1,4 @@
+import itertools
 import numpy as np
 from scipy import ndimage
 from dipy.align import floating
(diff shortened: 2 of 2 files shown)


skoudoro:
@grlee77, Can you have a look at this PR when you find the time? 

Thank you! 😄

dipy/align/metrics.py
```diff
@@ -244,6 +244,21 @@ def initialize_iteration(self):
...
         """
+        min_size = self.radius * 2 + 1
+
+        def invalid_image_size(image):
+            valid_shape = [size for size in image.shape if
+                           size >= min_size]
+            return True if len(valid_shape) != image.ndim else False
+
+        msg = ("Each image dimensions should be superior to 2 * radius + 1."
+               "Decrease CCMetric radius or Increase your image size")
```
grlee77: dimensions -> dimension
Increase -> increase

dipy/align/metrics.py
```diff
@@ -244,6 +244,21 @@ def initialize_iteration(self):
         re-orienting the gradients in the voxel space using the corresponding
         affine transformations.
         """
+        min_size = self.radius * 2 + 1
+
+        def invalid_image_size(image):
```
grlee77: I would simplify this function using something like:
e.g.
```Python
def invalid_image_size(image):
    return any([size < min_size for size in image.shape])
```
Given that this is now only a single line, it may not be worth making a separate function for it, but that is just a stylistic choice.

dipy/align/metrics.py
```diff
@@ -244,6 +244,21 @@ def initialize_iteration(self):
...
+                           size >= min_size]
+            return True if len(valid_shape) != image.ndim else False
+
+        msg = ("Each image dimensions should be superior to 2 * radius + 1."
+               "Decrease CCMetric radius or Increase your image size")
+
+        if invalid_image_size(self.static_image):
+            raise IOError("Static image size is too small. " + msg)
+        if invalid_image_size(self.moving_image):
+            raise IOError("Moving image size is too small. " + msg)
```
grlee77: I would have expected `ValueError` or `RuntimeError` rather than `IOError`, but I'm not familiar with what convention DiPy tends to favor.

dipy/align/tests/test_metrics.py
```diff
@@ -234,7 +250,7 @@ def test_em_demons_step_3d():
 
 
 if __name__ == '__main__':
-    test_em_demons_step_2d()
-    test_em_demons_step_3d()
+    # test_em_demons_step_2d()
+    # test_em_demons_step_3d()
     test_exceptions()
-    test_EMMetric_image_dynamics()
+    # test_EMMetric_image_dynamics()
```
grlee77: Why have these other tests been disabled? Is this left over from when you were debugging?

dipy/align/tests/test_metrics.py
```diff
@@ -15,6 +16,21 @@ def test_exceptions():
...
 
+    def check_image_size_error(metric):
+        metric.set_static_image(np.arange(np.prod(shape)).reshape(shape),
+                                np.eye(4), np.ones(3), np.eye(3))
+        metric.set_moving_image(np.arange(np.prod(shape)).reshape(shape),
+                                np.eye(4), np.ones(3), np.eye(3))
+        assert_raises(IOError, metric.initialize_iteration)
+
+    metric = CCMetric(2)
+    for shape in itertools.product((5, 8), (8, 5)):
```
grlee77: it will be clearer to others in the future if you add a small comment:
```
# expected to fail for any dimension < 2*radius + 1.
```
and then explicitly set the `radius` argument to `CCMetric`.

For completeness, you could also check that it does not raise when `size == 2*radius + 1`.

dipy/align/metrics.py
skoudoro: wow, Nice! I should definitively use more `any` and `all`

dipy/align/tests/test_metrics.py
skoudoro: Agree, done in my last commit

dipy/align/tests/test_metrics.py
skoudoro: left over, I will correct that...

skoudoro:
Thank you for your Nice review @grlee77. I just addressed all your comment

grlee77:
Thanks for addressing the comments. This looks good to me (assuming the CI passes)

arokem:
Looks good to me. CI failure is unrelated, so I will go ahead and merge.

PR #1827, opened 2019-05-06
Remove deprecated module dipy.segment.quickbundes

skoudoro:
This PR removes the deprecated module `dipy.segment.quickbundles` in favor of  `dipy.segment.clustering`.


dipy/segment/tests/test_qb.py
```diff
@@ -1,15 +0,0 @@
-import nibabel as nib
-
-from numpy.testing import assert_equal
-from dipy.data import get_fnames
-
-from dipy.segment.quickbundles import QuickBundles
-
-
-def test_qbundles():
```
arokem: Is there an equivalent of this test somewhere else?

skoudoro: yes, More complete on this file: [test_quickbundles.py](https://github.com/nipy/dipy/blob/master/dipy/segment/tests/test_quickbundles.py)

arokem: Sweet. Lets' go!

PR #1830, opened 2019-05-10
Delete six module

skoudoro:
Deleting `six` module. This module was used to maintain compatibility with Python 2. Not needed anymore


arokem:
I find it fascinating that CodeCov thinks that this resulted in decreased coverage.  At any rate, this LGTM, and unless someone objects strongly, I can merge this in a day or two.

PR #1835, opened 2019-05-16
[Fix]  Workflow mask documentation

### LINKED ISSUE #1764: DTI metrics workflow: mask is optional, but crashes when no mask provided
## Description
In the `dipy_fit_dti` workflow help message, the `mask` argument is said to be optional (`default: no mask used`). However, if no mask is provided, the script crashes and says that an argument is missing.

What should be the real behavior? 
- Make the argument really optional?
- Leave it mandatory and change the help message?

I would prefer to have it optional, but I want your opinions.

@skoudoro @Garyfallidis

skoudoro:
As discussed with @jchoude, the goal of this PR is to fix #1764. The docstring was not coherent with the current behavior. Indeed, `mask` is mandatory for all reconstruction workflow.

### DIFF
--- a/dipy/workflows/reconst.py
+++ b/dipy/workflows/reconst.py
@@ -242,7 +242,7 @@ class ReconstDtiFlow(Workflow):
             multiple bvectors files at once.
         mask_files : string
             Path to the input masks. This path may contain wildcards to use
-            multiple masks at once. 
+            multiple masks at once.
         b0_threshold : float, optional
             Threshold used to find b=0 directions (default 0.0)
         bvecs_tol : float, optional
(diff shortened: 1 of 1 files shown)


arokem:
Hmm. This fixes the documentation, but doesn't implement [your suggestion](https://github.com/nipy/dipy/issues/1764#issuecomment-472582016), so I don't think it really resolves the issue.

skoudoro:
I misunderstood the original issue. The original issue was just a misalignment between the workflow behavior and the documentation.

More generally, `mask_files` is mandatory for many reconstruction workflows. Indeed, it does not make a sense to fit the background on many reconstructions. To have a design coherence between all reconstruction workflow, I think it is a really bad idea to turn `mask_files` optional for DTI. I believe we should educate our users to provide a mask and not run a reconstruction without it. So my previous suggestion/question does not make any more sense and the current behavior is good.

arokem:
Yep yep. Mea culpa for the long wait.

PR #1821, opened 2019-05-03
Fixes 238, by requiring vol_idx input with 4D images.

arokem:
See #238

PR #1836, opened 2019-05-16
Corrected median_otsu function declaration that was breaking tutorials

clintg6:
The prior function declaration breaks all these tutorials that use median_otsu:
brain_extraction_dwi.py
reconst_msdki.py
reconst_dti.py
reconst_csa_parallel.py
reconst_fwdti.py
reconst_csa.py
reconst_csd_parallel.py
reconst_dki.py
syn_registration_2d.py
syn_registration_3d.py
tracking_quick_start.py

You can test this by running any of these scripts in dipy/doc/examples or here http://nipy.org/dipy/examples_built/reconst_csa.html. My modification corrects this so that all the tutorials still function normally.

### DIFF
--- a/doc/examples/reconst_csa.py
+++ b/doc/examples/reconst_csa.py
@@ -41,8 +41,8 @@ Remove most of the background using DIPY's mask module.
 from dipy.segment.mask import median_otsu
 
 
-maskdata, mask = median_otsu(data, 3, 1, True,
-                             vol_idx=range(10, 50), dilate=2)
+maskdata, mask = median_otsu(data, vol_idx=range(10, 50), median_radius=3, numpass=1,
+                             autocrop=True, dilate=2)
 
 """
 We instantiate our CSA model with spherical harmonic order of 4
--- a/doc/examples/reconst_csa.py
+++ b/doc/examples/reconst_csa.py
@@ -41,8 +41,8 @@ Remove most of the background using DIPY's mask module.
 from dipy.segment.mask import median_otsu
 
 
-maskdata, mask = median_otsu(data, vol_idx=range(10, 50), median_radius=3, numpass=1,
-                             autocrop=True, dilate=2)
+maskdata, mask = median_otsu(data, vol_idx=range(10, 50), median_radius=3,
+                             numpass=1, autocrop=True, dilate=2)
 
 """
 We instantiate our CSA model with spherical harmonic order of 4
--- a/doc/examples/syn_registration_3d.py
+++ b/doc/examples/syn_registration_3d.py
@@ -42,9 +42,9 @@ We first remove the skull from the b0's
 """
 
 from dipy.segment.mask import median_otsu
-stanford_b0_masked, stanford_b0_mask = median_otsu(data, median_radius=4,
+stanford_b0_masked, stanford_b0_mask = median_otsu(stanford_b0, median_radius=4,
                                                    numpass=4)
-syn_b0_masked, syn_b0_mask = median_otsu(data, median_radius=4, numpass=4)
+syn_b0_masked, syn_b0_mask = median_otsu(syn_b0, median_radius=4, numpass=4)
 
 static = stanford_b0_masked
 static_affine = nib_stanford.affine
--- a/doc/examples/syn_registration_3d.py
+++ b/doc/examples/syn_registration_3d.py
@@ -42,7 +42,8 @@ We first remove the skull from the b0's
 """
 
 from dipy.segment.mask import median_otsu
-stanford_b0_masked, stanford_b0_mask = median_otsu(data, median_radius=4, numpass=4)
+stanford_b0_masked, stanford_b0_mask = median_otsu(data, median_radius=4,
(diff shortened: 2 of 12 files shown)


skoudoro:
Good catch! Thank you for this! 

LGTM.  I see that @arokem introduce this change [here](https://github.com/nipy/dipy/pull/1821/files#diff-4fed05202cdf6fb524314714d4e1a210R122) (PR #1821) so is that ok for you or you prefer to keep you previous change and update all tutorials?

clintg6:
@skoudoroI I looked over the [PR 1821](https://github.com/nipy/dipy/pull/1821) and ran some further tests. The modification to the older declaration passes the following workflow and segment tests. 
1. dipy/workflows/tests/test_segment.py
2. dipy/workflows/tests/test_workflow.py  
3. dipy/segment/tests/test_mask.py

I also looked further into [dipy/workflows/segment.py](https://github.com/nipy/dipy/blob/8379bbde5bfe07ed2d6fa48f4f08953b298f62f5/dipy/workflows/segment.py#L68). The ordering is changed at line 68 in PR1821 but because the variables are specified the new ordering works fine with the older ordering in my correction.

arokem:
Hi @clintg6 : thanks for cleaning up after me! The truth is that I moved that key-word argument intentionally, because I wanted to give it a more prominent location, as it will often be required. This is really just a stylistic choice... Do you want to have a go at fixing the tutorials that were broken, instead of this fix?

clintg6:
Hi @arokem I can fix the tutorials. Do you want me to issue a new PR when I'm done?

arokem:
Yes. That would be great. Thanks!

doc/examples/syn_registration_3d.py
```diff
@@ -42,7 +42,8 @@
 """
 
 from dipy.segment.mask import median_otsu
-stanford_b0_masked, stanford_b0_mask = median_otsu(data, median_radius=4, numpass=4)
+stanford_b0_masked, stanford_b0_mask = median_otsu(data, median_radius=4,
```
arokem: This doesn't require a `vol_idx` input?

doc/examples/reconst_csa.py
```diff
@@ -41,8 +41,8 @@
 from dipy.segment.mask import median_otsu
 
 
-maskdata, mask = median_otsu(data, vol_idx=range(10, 50), median_radius=3, numpass=1,
-                             autocrop=True, dilate=2)
+maskdata, mask = median_otsu(data, vol_idx=range(10, 50), median_radius=3,
```
arokem: This is weird. I believe that the `vol_idx` here  (and elsewhere using this data) should be just the values where the `gtab.b0s_index` is `True`. Or am I missing  something?

doc/examples/syn_registration_3d.py
clintg6: No because the b0 itself is 3D. I accidentally introduced a typo. It should have been
stanford_b0_masked, stanford_b0_mask = median_otsu(stanford_b0, median_radius=4,
                                                   numpass=4)
syn_b0_masked, syn_b0_mask = median_otsu(syn_b0, median_radius=4, numpass=4)
I have corrected it.

arokem: Gotcha. That makes sense!

doc/examples/reconst_csa.py
clintg6: Yes, technically it should be where `gtab.b0s_index` is `True` if you want to perform skull stripping based upon the b0s. The skull is still present in the DWIs so I guess it still works with the `range(10, 50)`.

arokem:
Let's change it all to use the b0s. I think that's a better general example
of how to use this and will lead to less errors when users adapt this to
use their own data.

Garyfallidis:
Sorry, what is the issue here? Is vol_idx the default value to a range?

arokem:
The issue is that, if I understand correctly, it would usually be safer to
use the b=0 volumes for skull-stripping,  but here we are using a (random?)
selection of diffusion-weighted images  for this (b0s are in `data[...,
:10]`). I would like the example to demonstrate the most generic ("safest")
way to  use this.

Garyfallidis:
Nope this is the correct behavior here. Using a collection of non b0 volumes helps a lot the quality of segmentation. If you only use the b0 here this specific algorithm does not behave well. You would have to use a different dataset with a better b0 if you want to showcase vol_idx=0.

arokem:
Gotcha. Leave it as-is then.

We might want to better document somewhere how to use this algorithm. I
didn't know until now that the safest use pattern is diffusion-weighted
volumes and I still don't understand under what conditions you should use
b=0 (that's what you mean by `vol_idx=0`, right? It's not always the first
volume).

Garyfallidis:
Agreed.

arokem:
In that case, I think that  this is ready to be merged, right?

Garyfallidis:
Are we good with pep8 everywhere?

arokem:
Yes. Let's go!

PR #1843, opened 2019-05-23
[NF] new workflow: FetchFlow

skoudoro:
This PR adds a new workflow: `FetchFlow`. This command line is an easy way to grab our tutorial data via a command line.

Example:
----------
```
$ dipy_fetch standford_hardi
$ dipy_fetch all
$ dipy_fetch all --out_dir=~/my_folder
$ dipy_fetch standford_hardi bundle_fa_hcp
$ dipy_fetch standford_hardi bundle_fa_hcp --out_dir=~/my_folder
```

### DIFF
--- a/dipy/workflows/io.py
+++ b/dipy/workflows/io.py
@@ -147,35 +147,41 @@ class FetchFlow(Workflow):
                                if name.lower().startswith("fetch_")
                                if not len(getfullargspec(func).args)])
 
-        skipped_names = []
-        for data_name in data_names:
-            data_name = data_name.lower()
-            if data_name == 'all':
-                for name, fetcher_function in available_data.items():
-                    logging.info('------------------------------------------')
-                    logging.info('Fetching at {0}'.format(name))
-                    logging.info('------------------------------------------')
-                    fetcher_function()
-                break
+        data_names = [name.lower() for name in data_names]
 
-            if data_name not in available_data.keys():
-                skipped_names.append(data_name)
-                continue
+        if 'all' in data_names:
+            for name, fetcher_function in available_data.items():
+                logging.info('------------------------------------------')
+                logging.info('Fetching at {0}'.format(name))
+                logging.info('------------------------------------------')
+                fetcher_function()
 
-            logging.info('------------------------------------------')
-            logging.info('Fetching at {0}'.format(data_name))
-            logging.info('------------------------------------------')
-            available_data[data_name]()
-
-        nb_success = len(data_names) - len(skipped_names)
-        print('\n')
-        logging.info('Fetched {0} / {1} Files '.format(nb_success,
-                                                       len(data_names)))
-        if skipped_names:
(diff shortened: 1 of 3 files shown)


dipy/workflows/io.py
```diff
@@ -97,3 +98,95 @@ def run(self, input_files,
...
+        module: module object
+
+        """
+        if module_path in sys.modules:
+            return importlib.reload(sys.modules[module_path])
+        else:
+            return importlib.import_module(module_path)
+
+    def run(self, data_names, out_dir=''):
+        """Download files to folder and checks their md5 checksums.
```
arokem: Either "Downloads files to folder and checks their md5 checksums" or "Download files to folder and check their md5 checksums"

dipy/workflows/io.py
```diff
@@ -97,3 +98,95 @@ def run(self, input_files,
...
+                             ' names: {0}'.format(
+                                 ', '.join(available_data.keys())))
+
+        if out_dir:
+            if dipy_home:
+                os.environ['DIPY_HOME'] = dipy_home
+            else:
+                os.environ.pop('DIPY_HOME', None)
+
+            self.load_module('dipy.data.fetcher')
```
arokem: Why are you loading the module here again?

dipy/workflows/tests/test_io.py
```diff
@@ -27,13 +30,27 @@ def test_io_info():
...
+        fetch_flow.run(['bundle_fa_hcp'], out_dir=out_dir)
+        npt.assert_equal(os.path.isdir(os.path.join(out_dir,
+                                                    'bundle_fa_hcp')),
+                         True)
+
+
 if __name__ == '__main__':
-    test_io_info()
+    test_io_fetch()
+    # test_io_info()
```
arokem: Why is this commented out?

skoudoro: I forgot to put it back after my test

dipy/workflows/io.py
skoudoro: Since we can combine Workflow, I update the module as it was originally otherwise, it will be wrong [here](https://github.com/nipy/dipy/blob/master/dipy/data/fetcher.py#L26)

skoudoro:
Thank you for the review, just updated

dipy/workflows/io.py
arokem: How about adding a comment, like this?

dipy/workflows/io.py
```diff
@@ -97,3 +98,95 @@ def run(self, input_files,
...
```
arokem:
```suggestion
            # We load the module again so that if we run another one of these in 
            # the same process, we don't have the env variable pointing to the 
            # wrong place
            self.load_module('dipy.data.fetcher')
```

arokem:
OK. LGTM. Merging.

PR #1849, opened 2019-05-29
Adds control for number of iterations in CSD recon

scott-trinkle:
This allows the user to pass the maximum number of iterations as an argument when setting up a CSD model. This functionality already existed in the `csdeconv` function, but was not accessible with a standard workflow of setting up a CSD model object, then calling the `fit` method.

### DIFF
--- a/dipy/reconst/tests/test_csdeconv.py
+++ b/dipy/reconst/tests/test_csdeconv.py
@@ -577,6 +577,18 @@ def test_csd_superres():
     assert_(all(cos_sim > .99))
 
 
+def test_csd_convergence():
+    """ Check existence of `convergence` keyword in CSD model """
+    _, fbvals, fbvecs = get_fnames('small_64D')
+    bvals, bvecs = read_bvals_bvecs(fbvals, fbvecs)
+    gtab = gradient_table(bvals, bvecs)
+
+    evals = np.array([[1.5, .3, .3]]) * [[1.], [1.]] / 1000.
+
+    model = ConstrainedSphericalDeconvModel(gtab, (evals[0], 3.),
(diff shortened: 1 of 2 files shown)


arokem:
Looks good to me. Any chance you could add a test that exercises this? At the very least a smoke-test would be good to have. Thanks!

dipy/reconst/tests/test_csdeconv.py
```diff
@@ -577,6 +577,18 @@ def test_csd_superres():
...
+    """ Check existence of `convergence` keyword in CSD model """
+    _, fbvals, fbvecs = get_fnames('small_64D')
+    bvals, bvecs = read_bvals_bvecs(fbvals, fbvecs)
+    gtab = gradient_table(bvals, bvecs)
+
+    evals = np.array([[1.5, .3, .3]]) * [[1.], [1.]] / 1000.
+
+    model = ConstrainedSphericalDeconvModel(gtab, (evals[0], 3.),
+                                            sh_order=8, convergence=50)
+
```
arokem: Sorry - could you add a `fit` call in here? Just to check that it propagates through properly.

scott-trinkle: ah, good point - will do!

arokem:
Great. BTW: the CI failures are unrelated to this PR. Already recorded in
https://github.com/nipy/dipy/issues/1844

dipy/reconst/tests/test_csdeconv.py
skoudoro: It would be good to have an `assert` somewhere on this test.

skoudoro:
Thank you @scott-trinkle!

PR #1855, opened 2019-06-12
Pin scipy version for bots that need statsmodels.

arokem:
This is due to breaking changes in the scipy API that will be fixed on an eventual release of statsmodels. For now, let's pin this and unpin when a new release of scipy comes along.

Should help deal with this breakage: https://travis-ci.org/nipy/dipy/jobs/542426220#L6092

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ addons:
 
 env:
     global:
-        - DEPENDS="cython numpy scipy matplotlib h5py nibabel cvxpy"
+        - DEPENDS="cython numpy matplotlib h5py nibabel cvxpy"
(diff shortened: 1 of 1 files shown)


arokem:
It looks like the bots are happy, but is there a more elegant way to implement this?

skoudoro:
No Idea yet but I propose to merge this and we can create an issue to improve our Travis design.

What do you think?

arokem:
Sounds good. I also want to experiment a bit with CircleCI, so I can do that on a separate PR. But yes: feel free to merge this one.

skoudoro:
Thank you @arokem for this!

PR #1858, opened 2019-06-14
NF: MSMT - CSD 

ShreyasFadnavis:
- [x] Replace QP from CVXOPT with CVXPY
- [x] Fix Example
- [x] Add tests
- [x] Add docstrings


dipy/reconst/mcsd.py
```diff
@@ -0,0 +1,347 @@
...
+    B[np.ix_(gtab.b0s_mask, n > 0)] = 0.
+
+    iso = np.empty([B.shape[0], iso_comp])
+    iso[:] = SH_CONST
+
+    B = np.concatenate([iso, B], axis=1)
+    return B, m, n
+
+
+class MultiShellResponse(object):
```
arokem: Hmm. Is there a compelling reason to use an object here? An object with one method is basically a function, as eloquently laid out in [this pycon talk](https://www.youtube.com/watch?v=o9pEzgHorH0).

dipy/reconst/mcsd.py
```diff
@@ -0,0 +1,347 @@
...
+        self.shells = shells
+        if self.iso < 1:
+            raise ValueError("sh_order and shape of response do not agree")
+
+    @property
+    def iso(self):
+        return self.response.shape[1] - (self.sh_order // 2) - 1
+
+
+def closest(haystack, needle):
```
arokem: This function is: 

1. two lines implementing a fairly straightforward use of numpy, 
2. not documented or directly tested, and 
3. used exactly once. 

Could you just move this code into the calling function and eliminate this?

dipy/reconst/mcsd.py
```diff
@@ -0,0 +1,347 @@
...
+    kernal = response.response / delta
+
+    return kernal[np.ix_(b_idx, n_idx)]
+
+
+def _basic_delta(iso, m, n, theta, phi):
+    """Simple delta function
+    Parameters
+    ----------
+    iso: int (optional)
```
arokem: Says it's optional, but it doesn't seem to be optional in the function signature.

arokem: That said, it would make sense to add error handling to make sure that it's at least 2.

dipy/reconst/mcsd.py
```diff
@@ -0,0 +1,347 @@
...
+
+def _pos_constrained_delta(iso, m, n, theta, phi, reg_sphere=default_sphere):
+    """
+    Delta function optimized to avoid negative lobes. Implements a Linear
+    Programming solver from `CVXPY` to impose this positivity constraint. The
+    default solver used is GLPK.
+
+    Parameters
+    ----------
+    iso: int (optional)
```
arokem: Same comment here.

dipy/reconst/mcsd.py
```diff
@@ -0,0 +1,347 @@
...
+
+    # Realign reg_sphere so that the first vertex is aligned with delta
+    # orientation (theta, phi).
+    M = geo.vec2vec_rotmat(reg_sphere.vertices[0], [x, y, z])
+    new_vertices = np.dot(reg_sphere.vertices, M.T)
+    _, t, p = geo.cart2sphere(*new_vertices.T)
+
+    B = shm.real_sph_harm(m, n, t[:, None], p[:, None])
+    G_temp = np.ascontiguousarray(B[:, n != 0])
+    # c_ samples the delta function at the delta orientation.
```
arokem: This comment seems to refer to a variable that no longer exists (?).

dipy/reconst/mcsd.py
```diff
@@ -0,0 +1,347 @@
...
+    n : int ``>= 0``
+        The degree of the harmonic.
+    theta : array_like
+       inclination or polar angle
+    phi : array_like
+       azimuth angle
+    reg_sphere : Sphere (optional)
+        sphere used to build the regularization B matrix.
+        Default: 'symmetric362'.
+        weight given to the constrained-positivity regularization part of
```
arokem: This seems to document a parameter that no longer exists(?)

dipy/reconst/mcsd.py
arokem: Or maybe this is `c_temp`?

dipy/reconst/mcsd.py
```diff
@@ -0,0 +1,347 @@
...
+    # Realign reg_sphere so that the first vertex is aligned with delta
+    # orientation (theta, phi).
+    M = geo.vec2vec_rotmat(reg_sphere.vertices[0], [x, y, z])
+    new_vertices = np.dot(reg_sphere.vertices, M.T)
+    _, t, p = geo.cart2sphere(*new_vertices.T)
+
+    B = shm.real_sph_harm(m, n, t[:, None], p[:, None])
+    G_temp = np.ascontiguousarray(B[:, n != 0])
+    # c_ samples the delta function at the delta orientation.
+    c_temp = G_temp[0][:, None]
```
arokem: Why are we pulling out the first element of this array?

dipy/reconst/mcsd.py
```diff
@@ -0,0 +1,347 @@
...
+    return np.concatenate([iso_d, out])
+
+
+delta_functions = {"basic": _basic_delta,
+                   "positivity_constrained": _pos_constrained_delta}
+
+
+class MultiShellDeconvModel(shm.SphHarmModel):
+    def __init__(self, gtab, response, reg_sphere=default_sphere, iso=2,
+                 delta_form='basic'):
```
arokem: Maybe instead of `delta_form`, we can call this `positive_delta` and have it be a boolean that chooses `_pos_constrained_delta` for the delta function, when set to `True`?

dipy/reconst/mcsd.py
```diff
@@ -0,0 +1,347 @@
...
+        response : tuple or AxSymShResponse object
+            A tuple with two elements. The first is the eigen-values as an (3,)
+            ndarray and the second is the signal value for the response
+            function without diffusion weighting.  This is to be able to
+            generate a single fiber synthetic signal. The response function
+            will be used as deconvolution kernel ([1]_)
+        reg_sphere : Sphere (optional)
+            sphere used to build the regularization B matrix.
+            Default: 'symmetric362'.
+            weight given to the constrained-positivity regularization part of
```
arokem: Again this text about the weight. Does it really belong together with `reg_sphere`?

dipy/reconst/mcsd.py
```diff
@@ -0,0 +1,347 @@
...
+            the deconvolution equation (see [1]_). Default: 1
+        iso: int (optional)
```
arokem: Here it really is optional, with default set to 2. This is also the level at which I would implement error handling to enforce that it's >= 2

dipy/reconst/mcsd.py
```diff
@@ -0,0 +1,347 @@
...
+            X = B * multiplier_matrix
+        return np.dot(params, X.T)
+
+    @multi_voxel_fit
+    def fit(self, data):
+        coeff = self.fitter(data)
+        return MSDeconvFit(self, coeff, None)
+
+
+class MSDeconvFit(shm.SphHarmFit):
```
arokem: This one needs some documentation

arokem: Also, is the ODF inherited directly from `SphHarmFit` correct?

dipy/reconst/mcsd.py
```diff
@@ -0,0 +1,347 @@
...
+    def shm_coeff(self):
+        return self._shm_coef[..., self.model.response.iso:]
+
+    @property
+    def volume_fractions(self):
+        tissue_classes = self.model.response.iso + 1
+        return self._shm_coef[..., :tissue_classes] / SH_CONST
+
+
+def _rank(A, tol=1e-8):
```
arokem: Maybe we should use this one instead?

https://docs.scipy.org/doc/numpy-1.14.2/reference/generated/numpy.linalg.matrix_rank.html

dipy/reconst/mcsd.py
```diff
@@ -0,0 +1,347 @@
...
+    s = la.svd(A, False, False)
+    threshold = (s[0] * tol)
+    rnk = (s > threshold).sum()
+    return rnk
+
+
+def quadprog(P, Q, G, H):
+    r"""
+    Helper funstion to set up the Quadratic Program (QP) solver in CVXPY.
```
arokem: Typo: `funstion` => `function`

arokem: This documentation is a bit misleading. This doesn't just set up the problem. It sets it up and then solves it.

dipy/reconst/mcsd.py
```diff
@@ -0,0 +1,347 @@
...
+    A QP problem has the following form:
+    minimize      1/2 x' P x + Q' x
+    subject to    G x <= H
+
+    Here the QP solver is based on CVXPY and uses OSQP by default.
```
arokem: Both here and above, the solver is not a default that can be changed. This should be something like "Here the QP solver is based on CVXPY and uses OSQP." without that last phrase.

dipy/reconst/tests/test_mcsd.py
```diff
@@ -0,0 +1,116 @@
...
+        expected = np.dot(response.response[i, iso:], B.T)
+        npt.assert_array_almost_equal(signal, expected)
+
+    signal = model.predict(wm_delta, gtab)
+    fit = model.fit(signal)
+    m = model.m
+    npt.assert_array_almost_equal(fit.shm_coeff[m != 0], 0., 2)
+
+
+test_mcsd_model_delta()
```
arokem: Why are you calling this here?

dipy/reconst/tests/test_mcsd.py
```diff
@@ -0,0 +1,116 @@
...
+    angles = [(0, 0), (60, 0)]
+
+    S_wm, sticks = multi_tensor(gtab, mevals, S0, angles=angles,
+                                fractions=[30., 70.], snr=None)
+    S_gm = np.exp(-gtab.bvals * gm_md)
+    S_csf = np.exp(-gtab.bvals * csf_md)
+
+    sh_order = 8
+    response = sim_response(sh_order, [0, 1000, 2000, 3500])
+    model = MultiShellDeconvModel(gtab, response,
```
arokem: Do you want to also test the version without positivity constraint?

ShreyasFadnavis:
> Strong work! I had some questions/comments. I realize that maybe some of these are for @MrBago.

Thank you so much @arokem :) Will address all the changes mentioned above ASAP!

dipy/reconst/mcsd.py
ShreyasFadnavis: > Also, is the ODF inherited directly from `SphHarmFit` correct?

Yes :+1:

dipy/reconst/mcsd.py
ShreyasFadnavis: Not required anymore!

dipy/reconst/mcsd.py
ShreyasFadnavis: I agree that the `object` is not necessary, but, the class I guess has been designed to make the API look similar to `csdeconv`..

dipy/reconst/tests/test_mcsd.py
ShreyasFadnavis: Done!

dipy/reconst/mcsd.py
ShreyasFadnavis: That would  be a question for @MrBago ... it works as expected,  nevertheless :)

dipy/reconst/mcsd.py
```diff
@@ -0,0 +1,347 @@
...
+    G = cvx.Parameter((a_temp, b_temp))
+    G.value = -G_temp
+    h = cvx.Parameter((a_temp, b_temp))
+    h_temp = np.full((a_temp, b_temp), SH_CONST ** 2)
+    h.value = h_temp
+
+    # n == 0 is set to sh_const to ensure a normalized delta function.
+    # n > 0 values are optimized so that delta > 0 on all points of the sphere
+    # and delta(theta, phi) is maximized.
+    lp_prob = cvx.Problem(cvx.Maximize(cvx.sum(c_temp)), [G*x <= h])
```
MrBago: This doesn't seem right, is this `x` the same as the x defined in on 132? I thing `x` should be a variable & should be referenced in the `Maximize`.

I'm not sure how useful the `_pos_constrained_delta` is, consider removing it for this PR and adding it later with better tests.

ShreyasFadnavis: Will do.. I was about to ask the same..! Will take this function down for now :)

ShreyasFadnavis:
@arokem @MrBago Ttravis seems happy now! Any other comments, refactoring or suggestions? Also, I'm testing this with some HCP dataset.. Any other data that you have in mind for the tutorial? Stanford HARDi is an option I guess! 

Will follow up with another PR for the tutorial and workflow once this is merged :)

MrBago:
I think standford HARDI is 1 shell, you'll want to use a multi shell
dataset, preferably at least 3 shells with 30 directions each.

arokem:
The Stanford data is single shell. The CENIR or CFIN datasets should work
for this.

dipy/reconst/mcsd.py
```diff
@@ -0,0 +1,268 @@
...
+               spherical deconvolution for improved analysisof multi-shell
+               diffusion MRI data
+        .. [2] Tournier, J.D., et al. NeuroImage 2007. Robust determination of
+               the fibre orientation distribution in diffusion MRI:
+               Non-negativity constrained super-resolved spherical
+               deconvolution
+        .. [3] Tournier, J.D, et al. Imaging Systems and Technology
+               2012. MRtrix: Diffusion Tractography in Crossing Fiber Regions
+        """
+        if not iso >= 2:
```
arokem: Could you please add a test for this? That is, verify that the error is raised for number smaller than 2.

dipy/reconst/mcsd.py
arokem: Yeah. That's fine.

dipy/reconst/mcsd.py
```diff
@@ -0,0 +1,268 @@
...
+        multiple shells.
+
+        Parameters
+        ----------
+        response : tuple or AxSymShResponse object
+            A tuple with two elements. The first is the eigen-values as an (3,)
+            ndarray and the second is the signal value for the response
+            function without diffusion weighting.  This is to be able to
+            generate a single fiber synthetic signal.
+        sh_order : ndarray
```
arokem: Is this really an array? Why?

dipy/reconst/mcsd.py
arokem: Great. It still needs some documentation.

dipy/reconst/mcsd.py
```diff
@@ -0,0 +1,268 @@
...
+        coeff = self.fitter(data)
+        return MSDeconvFit(self, coeff, None)
+
+
+class MSDeconvFit(shm.SphHarmFit):
+
+    def __init__(self, model, coeff, mask):
+        """
+        Inherits the SphHarmFit which fits the diffusion data to a spherical
+        harmic model.
```
arokem: Typo: "harmic" => "harmonic"

dipy/reconst/mcsd.py
```diff
@@ -0,0 +1,268 @@
...
+    return opt
+
+
+class QpFitter(object):
+
+    def __init__(self, X, reg):
+        r"""
+        Makes use of the quadratic programming solver `quadprog` to fit the
+        model. The initialization for the model is done using the warm-start by
+        default in `CVXPY`.
```
arokem: Would be good to document the parameters.

arokem:
Oh - just saw your comment - tutorial on separate PR sounds good. So, there's just a couple of small things to resolve here and it will be ready to merge.

ShreyasFadnavis:
> A couple more comments here. Getting closer! Are you planning to add an example on this PR?

Hi @arokem ! Thank you for the suggestions! I will incorporate all the changes  :+1: 
I will follow up with another PR for the example and workflow :)

The example has multiple steps and I feel that it will make this PR too big.

dipy/reconst/mcsd.py
ShreyasFadnavis: > Is this really an array? Why?

My bad!

dipy/reconst/mcsd.py
ShreyasFadnavis: Looks okay?

ShreyasFadnavis:
@arokem is this good now? I would like to create one more PR for the tutorial for this model for the 1.0 release... Let me know in case of any more changes :)

dipy/reconst/mcsd.py
```diff
@@ -0,0 +1,294 @@
...
+
+from dipy.utils.optpkg import optional_package
+cvx, have_cvxpy, _ = optional_package("cvxpy")
+
+SH_CONST = .5 / np.sqrt(np.pi)
+
+
+def multi_tissue_basis(gtab, sh_order, iso_comp):
+    """
+    Builds a basis for multi-shell CSD model.
```
arokem: Documentation for this function is still not complete.

dipy/reconst/mcsd.py
arokem: Still no documentation here. What is `X`? What is `reg`? Are they expected to have a particular type/shape?

dipy/reconst/mcsd.py
ShreyasFadnavis: > Almost there. Just need to document these.

Sorry for that.. On it!

dipy/reconst/mcsd.py
```diff
@@ -0,0 +1,334 @@
...
+        Makes use of the quadratic programming solver `solve_qp` to fit the
+        model. The initialization for the model is done using the warm-start by
+        default in `CVXPY`.
+
+        Parameters
+        ----------
+        X : ndarray
+            Matrix to be fit by the QP solver calculated in
+            `MultiShellDeconvModel`
+        B : ndarray
```
arokem:
```suggestion
        reg : ndarray
```

arokem: I believe this is called `reg` in the signature of this function.

arokem:
Thanks @ShreyasFadnavis. Merging! 

Please do follow up with an example.

ShreyasFadnavis:
@arokem any idea why appveyor CI is failing?

PR #1862, opened 2019-06-19
Use logging instead of print statements

arokem:
Also removing "from future" imports along the way.

### DIFF
--- a/dipy/data/fetcher.py
+++ b/dipy/data/fetcher.py
@@ -1,8 +1,8 @@
 # -*- coding: utf-8 -*-
-from __future__ import division, print_function, absolute_import
 import os
 import sys
 import contextlib
+import logging
 
 from os.path import join as pjoin
 from hashlib import md5
@@ -38,7 +38,8 @@ def _log(msg):
     """Helper function to keep track of things.
     For now, just prints the message
     """
-    print(msg)
+    logger = logging.getLogger(__name__)
+    logger.info(msg)
 
 
 def update_progressbar(progress, total_length):
--- a/dipy/data/fetcher.py
+++ b/dipy/data/fetcher.py
@@ -35,7 +35,7 @@ class FetcherError(Exception):
 
 
 def _log(msg):
-    """Helper function to keep track of things.
+    """Helper function used as short hand for logging.
     """
     logger = logging.getLogger(__name__)
     logger.info(msg)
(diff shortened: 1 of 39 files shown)


skoudoro:
Overall this PR looks good to me. I see a WIP on your title, are there other files that you want to update?

dipy/data/fetcher.py
```diff
@@ -39,9 +39,9 @@ class FetcherError(Exception):
 
 def _log(msg):
```
skoudoro: Do we really want to keep this function? Since you are using `logging` module, you should delete it.

arokem: It's useful as shorthand. I just pushed a clarification of the docstring.

arokem:
Now also rebased.

arokem:
I'm also taking the opportunity to remove `verbose` settings, because I believe this should be set by the logging level. Would this break something else?

skoudoro:
I think we should keep `verbose`. We just need to improve it and do like the registration module.  Look at `dipy.align.imwarp`

skoudoro:
Hi @arokem, What is your status on this PR? it looks close to being done!

arokem:
Just rebased. IIUC, the main thing that needs to be done is reinstating `verbose` setting?

skoudoro:
Yes, I think so

arokem:
OK. I believe that I have restored the `verbose` flag everywhere we previously had it. I still think that this is something that is best handled by having the users set their logging levels, but we can also do that later if y'all end up agreeing with me.

arokem:
I believe that this ready for a re-review. Codecov is just being dumb.

arokem:
@arokem will look into nicer printing options for logging

skoudoro:
Is there anything you want to add? I forgot what we said during the last online meeting concerning this PR but it looks ready to be merged for me.  Maybe, the logging "_design/string format_" can be done on a new PR. What do you think?

arokem:
Ready on my end. @Garyfallidis : have you had a chance to think about this more?

skoudoro:
can you rebase this one or resolve the conflict @arokem?

arokem:
Rebased!

skoudoro:
Thank you @arokem! Any other comments? will merge it by the end of the week

Garyfallidis:
Hello @arokem. Sorry can you rebase this one more time? Will look into merging the following days.

arokem:
All set.

PR #1874, opened 2019-06-26
DOC: Fix documentation oversights.

jhlegarreta:
Fix documentation oversights: a typo and the default value given to a
parameter.


arokem:
Good catch! 🚀

PR #1877, opened 2019-07-01
Clean up: Remove all SCIPY_LESS_0.x.x

skoudoro:
This PR should fix #1863. It removes all SCIPY_LESS_0.x.x

PR #1878, opened 2019-07-01
Clean up: Remove NUMPY_LESS_0.8.x

skoudoro:
As a follow up of #1877, this PR remove NUMPY_LESS_0.8.x and `eigh` function


arokem:
Good stuff.

Issue #1779, opened 2019-03-15
Probabilistic Direction Getter gallery example


araikes:
## Description
The examples given in the example for getting probabilistic streamlines (http://nipy.org/dipy/examples_built/probabilistic_fiber_tracking.html#example-probabilistic-fiber-tracking) demonstrate saving the streamline_generator, not the actual streamlines. The code on the website is: 

```
from dipy.data import default_sphere

prob_dg = ProbabilisticDirectionGetter.from_shcoeff(csd_fit.shm_coeff,
                                                    max_angle=30.,
                                                    sphere=default_sphere)
streamlines_generator = LocalTracking(prob_dg, classifier, seeds, affine, step_size=.5)

save_trk("probabilistic_shm_coeff.trk", streamlines_generator, affine, labels.shape)
```

This should be (as in the Quick Start Tracking):
```
from dipy.data import default_sphere
from dipy.tracking.streamline import Streamlines

prob_dg = ProbabilisticDirectionGetter.from_shcoeff(csd_fit.shm_coeff,
                                                    max_angle=30.,
                                                    sphere=default_sphere)
streamlines_generator = LocalTracking(prob_dg, classifier, seeds, affine, step_size=.5)
streamlines = Streamlines(streamlines_generator)
save_trk("probabilistic_shm_coeff.trk", streamlines, affine, labels.shape)
```

## Way to reproduce
[If reporting a bug, please include the following important information:]
- [x ] Relevant images (if any)
Streamline generator export (open the image and zoom and you will see several small streamlines): 
![3D_View2](https://user-images.githubusercontent.com/18663624/54443297-93b08480-46fd-11e9-8686-e548abc05cf1.png)

Streamlines export as in the revised code:
![3D_View2](https://user-images.githubusercontent.com/18663624/54443320-9dd28300-46fd-11e9-8328-953b34df63ba.png)

skoudoro:
Thanks for pointing that out @araikes! It should normally work with the `streamlines_generator` too so I will check `save_trk` function before updating the tutorial.

gabknight:
Thanks @araikes. Saving the streamlines_generator works fine on my system. For clarity, we should always add the `Streamline(.)` in examples. I will go ahead and prepare the PR. @skoudoro do know what could have caused the issue?

skoudoro:
No idea @gabknight. Maybe a memory error, or just a visualization problem.

I tried on Windows and Mac and I can not reproduce the problem, so hard to help...

PR #1879, opened 2019-07-02
DOC - explicitly run the streamline generators before saving the trk

gabknight:
This addresses #1779 and clarifies the examples.

Any idea on the root of the problem in #1779 ?


skoudoro:
Nice! Thank you @gabknight! 

> Any idea on the root of the problem in #1779 ?

No idea at the moment, I will look deeper into it this week.

arokem:
An idea: how about if `save_trk` always tried to call `Streamlines` on its input? That way, we could pass it a generator and it would do the right thing.

gabknight:
@arokem `save_trk` accept generators, normally without issues (I cannot reproduce the problem in #1779).  It is actually a good idea to provide a generator as the streamlines will be generated as they are saved, limiting memory issue for a high number of seeds. 

There seems the be corner cases where this fails though.

skoudoro:
Agree with @gabknight.  Thank you for this update! merging

Issue #592, opened 2015-03-14
Shouldn't TRACKPOINT be renamed to NODIRECTION?


Garyfallidis:
Hi @gabknight and @MrBago, 

Is the that the TRACKPOINT stopping criterion is used now only when there is no available direction. I think it would make the docs much more clear if we use NODIRECTION rather than the current name.

Cheers!

MrBago:
The tissue classifier is independent of the direction getter and cannot know whether a direction exists or not. TRACKPOINT is used when the tissue classifier believes the point should be part of a tract (and most of the time it is), but for some points that are marked as TRACKPOINT the direction getter is unable to generate a valid tracking direction.

If you point us to where you believe the docs are unclear, we can brainstorm some ways to improve them.

Garyfallidis:
Yeah, what you say here afaik doesn't come out from the current tutorials. Please add this information in tutorial  doc/examples/tracking_tissue_classifier.py or in the other ones if you think it is necessary.

skoudoro:
closing, fixed on #1882 by @gabknight

PR #1882, opened 2019-07-02
DOC - clarified the state of the tracking process once stopped

gabknight:
This addresses #592.


arokem:
LGTM. CI failure is unrelated as far as I can tell.

PR #1884, opened 2019-07-05
Clean up: Remove streamlines memory patch

skoudoro:
Since our Nibabel minimum version is 2.4, we do not need this patch anymore. This patch was fixing a memory issue on streamlines API and has been fixed on Nibabel 2.3

### DIFF
--- a/dipy/tracking/streamline.py
+++ b/dipy/tracking/streamline.py
@@ -7,7 +7,8 @@ from scipy.spatial.distance import cdist
 import numpy as np
 from nibabel.affines import apply_affine
 from nibabel.streamlines import ArraySequence as Streamlines
-from dipy.tracking.streamlinespeed import set_number_of_points, length
+from dipy.tracking.streamlinespeed import (set_number_of_points, length,
+                                           compress_streamlines)
 from dipy.tracking.distances import bundles_distances_mdf
 import dipy.tracking.utils as ut
 from dipy.core.geometry import dist_to_corner
--- a/dipy/tracking/streamline.py
+++ b/dipy/tracking/streamline.py
@@ -2,163 +2,16 @@ from copy import deepcopy
 from warnings import warn
 import types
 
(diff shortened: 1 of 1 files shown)


dipy/tracking/streamline.py
```diff
@@ -2,164 +2,18 @@
...
 from nibabel.affines import apply_affine
-from dipy.tracking.streamlinespeed import set_number_of_points
-from dipy.tracking.streamlinespeed import length
+from nibabel.streamlines import ArraySequence as Streamlines
+from dipy.tracking.streamlinespeed import (set_number_of_points, length,
+                                           compress_streamlines)
 from dipy.tracking.distances import bundles_distances_mdf
-from dipy.tracking.streamlinespeed import compress_streamlines
 import dipy.tracking.utils as ut
-from dipy.tracking.utils import streamline_near_roi
```
gabknight: What is the preferred convention here?

-  import as ut,  or 
- import : move_streamlines, streamline_near_roi, reduce_rois

gabknight:
LGTM. thx @skoudoro

dipy/tracking/streamline.py
skoudoro: No strong opinion here, but I will change it to `import : move_streamlines, streamline_near_roi, reduce_rois`

arokem: I think that either is fine. Some (me included) prefer the `import utils as ut` and then calling everything with a `ut.this` and `ut.that`, because it's more explicit. It tells you each time exactly where that function comes from.

skoudoro:
Personally, I find `Streamlines` more explicit than `ArraySequence` so I think it will be better to let it as it is.

> I think that either is fine.

If it is ok for both of you, I will not make any change. So, this PR is ready to be merged

arokem:
Yeah. Fine by me. Let's go!

PR #1666, opened 2018-11-19
Remove the points added oustide of a mask. Fix the related tests.

czotti:
This PR is to address the issue #1661. 
I recreate a PR because I made a mistake during the rebase.

- It change the way points are added to a streamline if the point is outside the mask.
- I changed all the needed test.

PR #1885, opened 2019-07-05
BF - remove single pts streamline

gabknight:
Since #1666, it was possible to return "valid" streamlines with a single point. This could happen in corner cases with a sequence of voxels: gm - wm - gm, wherein a single tracking step the tracking stops in the gm on both sides, resulting in a correctly ending streamline of a single point.

This PR fixes this.

### DIFF
--- a/dipy/tracking/local/localtracking.py
+++ b/dipy/tracking/local/localtracking.py
@@ -162,10 +162,13 @@ class LocalTracking(object):
                     parts = (B[stepsB - 1:0:-1], F[:stepsF])
                     streamline = np.concatenate(parts, axis=0)
 
-                if self.save_seeds:
-                    yield streamline, s
-                else:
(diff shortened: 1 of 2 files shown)


gabknight:
@skoudoro Do you know what is failing here?

arokem:
Unrelated to this PR. Something to do with byte ordering on Windows in the bowels of some Cython code...

skoudoro:
In General, we should start to track cython and python warning on Travis. I think it will helps. I will create an issue for that

skoudoro:
Can you rebase this one too to make Appveyor happy? This error should be fixed. 

This PR is ready to be merged, Does someone has a suggestion? it will be good to merge this in the middle of this coming week

skoudoro:
thank @gabknight for this bug fix, merging

Issue #1501, opened 2018-04-29
Refactoring tracking and checking tutorials and workflows - high priority for next release.


Garyfallidis:
We need to look carefully to our tracking framework and refactor some of the cython parts as well as improve tutorials and add workflows. I suggest this being high priority for release 0.15. Let me know if you agree and have suggestions for improvements.

As discussed in #1786 (DIPY LAB) we need to:

- [ ] Change local tracking with initial directions. This change will change current behavior of LocalTracker.
- [ ] Test EuDX, PFT and Bootstrap trackers (build workflows) with data and quality control them (dipy_qa). 
- [ ] All should be ready for July release.
- [x] Discuss in the next meeting
- [x] Remove legacy code and examples 
- [ ] #1648 Mesh seeding to be rebased after changes, as well as #1635.

gabknight:
yes! With 2 large PR + a cython refactoring, that would a good to do some tweaks now. We have to push the workflows for all the various tracking algorithm for 0.15!

jchoude:
Might this be a good opportunity to uniformize the streamlines handling, with @frheault and what was discussed at the Workshop?

PR #1887, opened 2019-07-05
DOC - updated streamline_tools example with the LocalTracking Framework

gabknight:
Changed the call from EuDX to LocalTracking. EuDX is still the algorithm use here, but in the new framework.

### DIFF
--- a/doc/examples/streamline_tools.py
+++ b/doc/examples/streamline_tools.py
@@ -18,19 +18,20 @@ EuDX along with the CsaOdfModel to make some streamlines. Let's import the
 modules and download the data we'll be using.
 """
 
-from dipy.tracking.eudx import EuDX
-from dipy.reconst import shm
-from dipy.direction import peaks
-from dipy.tracking import utils
-from dipy.tracking.streamline import Streamlines
+import numpy as np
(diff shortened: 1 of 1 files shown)


skoudoro:
Thanks @gabknight. Should we keep the `Eudx` class or more generally `eudx.py` (old framework)?

gabknight:
@skoudoro, no, no need to keep it as it is essentially the same as LocalTracking with `PeaksAndMetricDG`. I want to refactor this: remove eudx.py, rename `PeaksAndMetricDG` to EudxDG, and move the eudx tests.

This will be in the another PR addressing #1501.

skoudoro:
ok, sounds good! Do you need help? Today, I can do this part of removing `eudx.py` and `renaming PeaksAndMetricsDG`

gabknight:
@skoudoro sure give it a go if you have time. The tricky part will adjust the tests. It will be very good to do it, as the current `PeakAndMetricDG` is not properly tested.
If you don't, I will do it at the time RF the tracking.local package. I'll be away from now until 19.07, I'm not sure I can make it for the next release.

skoudoro:
I just run the documentation and this example, everything looks good. merging

Thanks @gabknight!

Issue #325, opened 2014-02-11
streamline extraction from eudx is failing - but perhaps eudx is failing


satra:
running into a weird error.

model = CSD

i suspect something is wrong in my usage as eudx finishes very quickly relative to doing the same thing from a tensor model.

Code

```
peaks = peaks_from_model(model=model,
                         data=data,
                         sphere=sphere,
                         mask=mask,
                         return_sh=True,
                         return_odf=False,
                         normalize_peaks=True,
                         relative_peak_threshold=.5,
                         min_separation_angle=25,
                         parallel=True)

from dipy.tracking.eudx import EuDX
eu = EuDX(peaks.gfa, peaks.peak_indices, odf_vertices = sphere.vertices, 
          a_low=0.2, seeds=10**6, ang_thr=35)
```

Error

```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-9-f2fb6b1e8d00> in <module>()
      1 import dipy.tracking.metrics as dmetrics
----> 2 streamlines = [streamline for streamline in eu] # if dmetrics.length(streamline) > 15]
      3 
      4 hdr = nib.trackvis.empty_header()
      5 hdr['voxel_size'] = new_zooms

/software/anaconda/envs/devpype/lib/python2.7/site-packages/dipy/tracking/_utils.pyc in move_streamlines(streamlines, output_space, input_space)
    536     # End of initialization
    537 
--> 538     for sl in streamlines:
    539         yield np.dot(sl, lin_T) + offset
    540 

/software/anaconda/envs/devpype/lib/python2.7/site-packages/dipy/tracking/eudx.pyc in _voxel_tracks(self, seed_voxels)
    200                                              self.step_sz,
    201                                              self.total_weight,
--> 202                                              self.max_points)
    203                 if track is not None and track.shape[0] > 1:
    204                     yield track

/software/anaconda/envs/devpype/lib/python2.7/site-packages/dipy/tracking/propspeed.so in dipy.tracking.propspeed.eudx_both_directions (dipy/tracking/propspeed.c:3189)()

ValueError: Buffer has wrong number of dimensions (expected 4, got 5)
```

skoudoro:
Due to the new tracking API, this is not anymore relevant. Closing

Issue #1807, opened 2019-04-15
tracking fails when npeaks=1 for peaks_from_model with tensor model


scott-trinkle:
## Description
If you use `peaks_from_model` as the direction getter with a tensor model, setting `npeaks=1`, no streamlines appear when you attempt to visualize. If you set `npeaks=2` (or anything > 1), it visualizes fine. I'm not sure if the problem is in `LocalTracking` or `actor.line`, but this should probably be fixed, this is probably the most intuitive way to run deterministic tractography with a tensor model. 

## Way to reproduce
I first noticed this on my own data, and reproduced with the Stanford data:

```import numpy as np
from dipy.data import read_stanford_labels, read_stanford_pve_maps

print('Loading data')
hardi_img, gtab, labels_img = read_stanford_labels()
data = hardi_img.get_fdata()
labels = labels_img.get_data()
affine = hardi_img.affine
pve_csf, pve_gm, pve_wm = read_stanford_pve_maps()

from dipy.tracking.local import BinaryTissueClassifier
classifier = BinaryTissueClassifier(pve_wm.get_data() > 0)

from dipy.tracking.utils import seeds_from_mask
seed_mask = labels == 2
seeds = seeds_from_mask(seed_mask, density=1, affine=hardi_img.affine)

from dipy.reconst.dti import TensorModel
tensor_model = TensorModel(gtab)

from dipy.data import get_sphere
from dipy.reconst.peaks import peaks_from_model

print('Getting peaks, npeaks=1')
dti_dg_1 = peaks_from_model(model=tensor_model,
                            data=data,
                            sphere=get_sphere('repulsion724'),
                            relative_peak_threshold=.2,
                            min_separation_angle=25,
                            mask=(labels == 2) | (labels == 1),
                            npeaks=1)

print('Tracking')
from dipy.tracking.local import LocalTracking
from dipy.tracking.streamline import Streamlines

streamline_generator_1 = LocalTracking(dti_dg_1,
                                       classifier,
                                       seeds,
                                       hardi_img.affine,
                                       step_size=0.5)
streamlines_1 = Streamlines(streamline_generator_1)

print('Visualizing')
from dipy.viz import window, actor, colormap
ren = window.Renderer()
ren.add(actor.line(streamlines_1, colormap.line_colors(streamlines_1)))
window.record(ren, out_path='dti_from_peaks_npeaks1.png', size=(600, 600))

print('Getting peaks, npeaks=2')
[...]

PR #1888, opened 2019-07-05
BF - fix eudx tracking for npeaks=1

gabknight:
This addresses #1807 

- The last peak was always incorrectly missing. This was especially visible when there was a single peak...
- Added a test for it.


dipy/reconst/peak_direction_getter.pyx
```diff
@@ -75,8 +75,9 @@ cdef class PeaksAndMetricsDirectionGetter(DirectionGetter):
                 raise IndexError("point outside data")
 
         # Check to see how many peaks were found in the voxel
-        for numpeaks in range(self._ind.shape[3]):
-            if self._ind[ijk[0], ijk[1], ijk[2], numpeaks] < 0:
+        for numpeaks in range(self._ind.shape[3] + 1):
+            if (numpeaks >= self._ind.shape[3] or
```
arokem: I am not sure that I follow this. You add one more item into the iteration (the highest value of the  shape on  the last dimension), but then you seem to ignore this value? Is that really supposed to be a `>=`, not just `>`?

dipy/reconst/tests/test_peak_dg.py
```diff
@@ -67,6 +67,17 @@ def get_direction(dg, point, dir):
...
                 npt.assert_array_almost_equal(id, [expected_dir])
 
+    peaks1 = peaks_from_model(SillyModel(), data, default_sphere,
+                              relative_peak_threshold=.5,
+                              min_separation_angle=25,
+                              npeaks=1)
+    peaks1._initialize()
+    point = np.array([1, 1, 1], dtype=float)
+
+    # id should have one direction
```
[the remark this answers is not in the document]
gabknight: fixed. thanks.

dipy/reconst/peak_direction_getter.pyx
gabknight: The problem was when numpeaks==self._ind.shape[3]. 
E.g. if the true number of peaks is 3 and self._ind.shape[3]==3, the loop was exiting with numpeaks==2 (never entering the if), later on, this was used in range(numpeaks), which then miss the last existing peak.

My fix didn't make the code clear. I re-worked it following your comment, making the logic more clear.

Thanks @arokem

skoudoro:
Since you are looking at `Eudx`, is the following issue still relevant: https://github.com/nipy/dipy/issues/325?

dipy/reconst/peak_direction_getter.pyx
```diff
@@ -75,8 +75,12 @@ cdef class PeaksAndMetricsDirectionGetter(DirectionGetter):
...
 
         # Check to see how many peaks were found in the voxel
-        for numpeaks in range(self._ind.shape[3]):
-            if self._ind[ijk[0], ijk[1], ijk[2], numpeaks] < 0:
+        numpeaks=0
+        for i in range(self.peak_values.shape[3]):
+            # Test if the value associated to the peak is > 0
+            if self._qa[ijk[0], ijk[1], ijk[2], i] > 0:
+                numpeaks = numpeaks + 1
+            else:
```
arokem: Thanks! This is certainly clearer. 

Would:

```
numpeaks = sum(self._qa[ijk[0], ijk[1], ijk[2], i] > 0) 
```

work here?

gabknight:
> Since you are looking at `Eudx`, is the following issue still relevant: #325?

@skoudoro `EuDX(.)` will now be `LocalTracking(.)` and `PeaksAndMetricDG` will get the needed information directly from the `PeaksAndMetric` object. So the specific error should not happen. 

However, this does not clarify the cython ValueError from  `dipy.tracking.propspeed.eudx_both_directions(.)` that @Garyfallidis mentioned (which is also used by `PeaksAndMetricDG`).

So might want to close this issue, and make a new general one on better handling of cython ValueError.

dipy/reconst/peak_direction_getter.pyx
gabknight: No, `>` doesn't work for arrays. Am I missing something?
```
Error compiling Cython file:
------------------------------------------------------------
...
         numpeaks = sum(self._qa[ijk[0], ijk[1], ijk[2], :] > 0)
                                                          ^
------------------------------------------------------------

dipy/reconst/peak_direction_getter.pyx:85:59: Invalid types for '>' (double[::1], long)

```

Would have been more clear :)

arokem: Oh well. Thanks for checking.

arokem:
I think this is ready to go? @skoudoro : do you agree? Or is your [comment](https://github.com/nipy/dipy/pull/1888#issuecomment-509683315) still something that needs too be addressed here?

skoudoro:
Agree, you can merge it! I just need to create a new issue and close #325 as gab suggested. will do later

Issue #1889, opened 2019-07-05
`segment.bundles` package not being tested


jhlegarreta:
## Description
The [`bundles`](https://github.com/nipy/dipy/blob/master/dipy/segment/bundles.py) package's [unit test](https://github.com/nipy/dipy/blob/master/dipy/segment/tests/test_bundles.py) is empty.

The test file's history shows that there has been some refactoring at some point, but it looks like some testing code was either misplaced or the `test_bundles.py` needs to be fed with the appropriate testing code.

codecov [reports](https://codecov.io/gh/nipy/dipy/src/master/dipy/segment/bundles.py) a 91.47% coverage for that package, but it may well be because it is being called as an intermediate step from other tests or code fragments.

arokem:
Looks like prior to [this](https://github.com/nipy/dipy/commit/b3a6d0eb376ca2211ff514e17cc71fbd812804fc#diff-ca53b3e9e0ca17697e9c4822526980a8) commit there wasn't a test file for this module. Then, I moved the tests that I had created in [this](https://github.com/nipy/dipy/commit/e255facb10c28096016746649d2dba378230126c#diff-ca53b3e9e0ca17697e9c4822526980a8) commit. 

Maybe we should simply remove this file altogether? Looks like most of that stuff is tested in  `test_rb.py`. Another option would be to rename that to `test_bundles.py`  for clarity. WDYT @Garyfallidis?

skoudoro:
fixed by #1891, closing

PR #1891, opened 2019-07-08
Move the tests from test_refine_rb to test_bundles.

arokem:
This addresses #1889 and also fixes up a few things in dipy.segment.bundles: 

1. Removed an undocumented two-liner in favor of inline code. 
2. Addresses an issue where empty indices throw an error. 

Testing in this module is still pretty spotty, as far as I can tell.


dipy/segment/bundles.py
```diff
@@ -615,7 +608,12 @@ def _prune_what_not_in_model(self, model_centroids,
         pruned_indices = [rtransf_cluster_map[i].indices
                           for i in np.where(mins != np.inf)[0]]
         pruned_indices = list(chain(*pruned_indices))
-        pruned_streamlines = transf_streamlines[np.array(pruned_indices)]
+        idx = np.array(pruned_indices)
+        if len(idx) == 0:
+            print(' You have removed all streamlines')
```
skoudoro: `logging` instead of `print` ? or it should be merged before #1862

arokem: If this is good, let's merge this one first, and then I can rebase #1862 and fix this up.

skoudoro: sounds like a plan! merging

Issue #1890, opened 2019-07-08
Tensor I/O in dipy_fit_dti


ecaruyer:
## Short description
When exporting a tensor field using dipy_fit_dti, the order in not the standard row-order lower-triangular. There is a NIFTI standard for storing symmetric matrices (see below), I would suggest to stick to the standard, although there is a good reason not to do so. Another argument is that in `dipy.reconst.dti`, there is a `from_lower_triangular()` method which is expecting row-order.

## In the code
In the source code of `ReconstDtiFlow`, a `correct_order = [0, 1, 3, 2, 4, 5]` is introduced, which basically changes the row-order lower triangular into a column-order lower triangular. Could you guide me to the rational behind this reordering?

## Reference to NIFTI standard
```
 /*! To store an NxN symmetric matrix at each voxel:
       - dataset must have a 5th dimension
       - intent_code must be NIFTI_INTENT_SYMMATRIX
       - dim[5] must be N*(N+1)/2
       - intent_p1 must be N (in float format)
       - the matrix values A[i][[j] are stored in row-order:
         - A[0][0]
         - A[1][0] A[1][1]
         - A[2][0] A[2][1] A[2][2]
         - etc.: row-by-row                           */

#define NIFTI_INTENT_SYMMATRIX 1005
```
(source [Official definition of the nifti1 header](https://nifti.nimh.nih.gov/pub/dist/src/niftilib/nifti1.h))

Garyfallidis:
@arokem and @MrBago please respond to @ecaruyer and we can look if it is possible to export to the current or nifti standard way.

arokem:
Maybe @matthieudumont knows? He wrote that line: https://github.com/nipy/dipy/commit/63dc1be1b31be569353316e3123258ce3cf8eaf1 

Does he still hang out here? I'd be +1 to changing this to the "correct" order.

arokem:
See #1892

arokem:
Hey @ecaruyer : just to make sure. You wrote "... I would suggest to stick to the standard, although there is a good reason not to do so...". Did you mean " I would suggest to stick to the standard,*unless* there is a good reason not to do so"?

ecaruyer:
Hi @arokem; indeed :-). Thanks for the discussion and for your responsiveness!

jchoude:
Regarding @matthieudumont I explained the reasoning in #1892

MrBago:
We have a function that'll build a properly formed nifti symmetric matrix from an array:

[...]

PR #1892, opened 2019-07-09
Use the correct (row) order of the tensor components

arokem:


arokem:
Addresses #1890

ecaruyer:
Thanks @arokem for this PR; if we wanted to be fully compliant with the Nifti1 standard, there would still be 2 things to modify:

- providing the right intent code in the header (using something like `hdr.set_intent("symmetric matrix")` in nibabel)
- adding one more dimension to the image data (using something like `data = data[..., np.newaxis, :]`) -- as far as I understand the 4th dimension is dedicated to store temporal data, e.g. for fMRI.

This would be cherry on the cake, but as it stands this PR solves #1890

arokem:
Thanks for the feedback. I’ll do those things later today. I also want to
add a little test to be sure we get this right.

jchoude:
Just a quick mention: I fully agree that we should follow the Nifti standard (using the 5 dimension for the elements, since the 4th is for temporal), and I even often argued that this should be the case. However, I was most often told that "other software expect the values to be in the 4th dimension". Same goes for the SH coefficients, etc.

If we change it to the 5th dimension, I expect to see a few issues coming about external software not loading that correctly anymore. Just something to take into account.

Also, @matthieudumont is probably not seeing those, even if he's sitting at the desk besides mine. I'm pretty sure that the reason the reordering was done was to conform to what the Fibernavigator expected at the time. I personally would say that this is not a valid reason anymore. Just legacy.

dipy/workflows/reconst.py
```diff
@@ -280,6 +287,8 @@ def run(self, input_files, bvalues_files, bvectors_files, mask_files,
             (default 'evecs.nii.gz')
         out_eval : string, optional
             Name of the eigenvalues to be saved (default 'evals.nii.gz')
+        fsl_format : bool, optional
+            Whether the tensor
```
jhlegarreta: The parameter explanation here seems to be unfinished :book:.

arokem:
Good catch! See last commit.

jhlegarreta:
Nice. Thanks @arokem.

arokem:
Hey @jchoude : does this allay your concerns? The default might break other software, but there's a flag. 

@MrBago : I understand the concern about back compatibility, but with 1.0 as our coming release, I think this is the time to break back compatibility if there ever was one. I am not sure what issues would come up, but we'd help the Nipype-ers refactor as needed.

MrBago:
Sounds good.

skoudoro:
Can you rebase this PR? Thank you!

arokem:
Rebased!

skoudoro:
Thank you @arokem. Any other comments?  I will wait until tomorrow before merging this PR.

dipy/workflows/reconst.py
```diff
@@ -222,7 +223,7 @@ def run(self, input_files, bvalues_files, bvectors_files, mask_files,
             out_dir='', out_tensor='tensors.nii.gz', out_fa='fa.nii.gz',
             out_ga='ga.nii.gz', out_rgb='rgb.nii.gz', out_md='md.nii.gz',
             out_ad='ad.nii.gz', out_rd='rd.nii.gz', out_mode='mode.nii.gz',
-            out_evec='evecs.nii.gz', out_eval='evals.nii.gz'):
+            out_evec='evecs.nii.gz', out_eval='evals.nii.gz', fsl_tensor=False):
```
Garyfallidis: The name of the parameter is misleading. Should we say nifti_standard_tensor or nifti_tensor?

arokem: And this would be set to True per default? The problem with that is that it's not so clear what the alternative means. Here, the idea is that this tells the users that this would give them a file that interoperates with FSL. Does that make sense?

Garyfallidis: Trying to use software names in the past did not go well. Why fsl_tensor and not trackvis_tensor or other? I think having the ```nifti_standard``` there is better and you can explain more in the docstring. And in the docstring we should have the convention explained for the choice of lower vs upper triangular and row vs column order. @ecaruyer do you agree?

arokem: How's this? See recent https://github.com/nipy/dipy/pull/1892/commits/24dc0eaac65b9020a931aa5171a854d647234fee

ecaruyer: Yes, I agree with the nifti_standard suggestion. 

Just one last comment from me, I am quite sure trackvis does follow the row-order lower triangular convention, which is I think in contradiction with the docstring as it stands.

Thanks again for this PR!

arokem: Oh yeah. I think the other software that was mentioned was Fibernavigator, but I think even that may no longer be true. Fixed in https://github.com/nipy/dipy/pull/1892/commits/162de19e392cda86a1bec4138c54c644062179ec

Garyfallidis:
Great, thank you all. Merging...

PR #1894, opened 2019-07-10
Default sphere: From symmetric724 to repulsion724

### LINKED ISSUE #590: Change default sphere
We still have the symmetric724 as the default sphere. However the repulsive724 is even better. 
I am suggesting to replace the default sphere with the repulsive724.

skoudoro:
This PR should fix #590. We use repulsion724 as a default sphere instead of symmetric724


arokem:
Hey @skoudoro : thanks for doing this.  Maybe better to change everything to use `default_sphere` *except* in `data/__init__.py`?

skoudoro:
Yes, you are right, I hesitate to do it this way for no reason, I will update this PR this afternoon

jhlegarreta:
Thanks for having addressed this @skoudoro !

dipy/reconst/tests/test_csdeconv.py
```diff
@@ -564,8 +565,9 @@ def test_csd_superres():
                                               sh_order=16)
     fit16 = model16.fit(S)
 
+    sphere = HemiSphere.from_sphere(get_sphere('symmetric724'))
```
arokem: Why are we still using the symmetric sphere here?

dipy/reconst/tests/test_shore_metrics.py
```diff
@@ -74,8 +74,7 @@ def test_shore_metrics():
 
     # compare the shore pdf with the ground truth multi_tensor pdf
 
-    sphere = get_sphere('symmetric724')
-    v = sphere.vertices
+    v = default_sphere.vertices
```
arokem: I wonder if the change here caused the CI error on Python 3.5. I just restarted that build, to seehow intermittent that is.

dipy/tracking/local/tests/test_tracking.py
```diff
@@ -567,7 +567,9 @@ def test_bootstap_peak_tracker():
     seeds = [np.array([0., 1., 0.]), np.array([2., 4., 0.])]
 
     tc = BinaryTissueClassifier((simple_image > 0).astype(float))
-    boot_dg = BootDirectionGetter.from_data(data, csd_model, 60)
+    sphere = HemiSphere.from_sphere(get_sphere('symmetric724'))
```
arokem: Here as well: why are we still using symmetric?

arokem:
LGTM. Let's see how persistent that Travis failure is.

dipy/reconst/tests/test_csdeconv.py
skoudoro: To avoid to update the whole tests 😄. It seems that when we use repulsion, there is 3 peaks instead of 2 and they do not line up with the sticks

dipy/tracking/local/tests/test_tracking.py
skoudoro: almost the same comment as above. we can create an issue to spend more time on these 2 specifics tests and update them

arokem:
Yeah. Let's do this separately. It's not great if this depends strongly on
the choice of sphere, but maybe there's a prosaic explanation.

arokem:
Looks like that failure is rather persistent. Seems like something about the SHORE metrics is sensitive to the choice of sphere.

arokem:
The remaining CI error here seems to have been a glitch on Travis side. I just kicked off that build again. If that comes back green, should I go ahead and merge it? Are things all set on your side @skoudoro?

skoudoro:
Thank you for catching that. Everything good from my side. You can go ahead if it is good for you

arokem:
OK. Looks like the CI is all passing here. 

Is this ready to go, or do you want to do anything more here, @skoudoro? 

Anyone else have any additional comments?

skoudoro:
Nothing more, Ready to go

arokem:
Let's go!

arokem:
Oh no - looks like I can't merge this anymore. One more rebase, please?

PR #1896, opened 2019-07-11
New module: dipy.core.interpolation

### LINKED ISSUE #728: Many interpolation functions in different places can they all go to same module?
Hi all,
we have different interpolation functions in different places. Should they all or some of them go to one place? Maybe dipy.core.interpolation?
I can see for example that we have trilinear interpolation in viz, tracking, denoising and align. Many of these do different things. So I would like to hear some opinions here. @MrBago, @omarocegueda ?
Otherwise, maybe I can make a start and move my trilinear interpolation function to dipy.core.interpolation and then we see how it goes. By the way this is a non-critical issue. So put it in lower priority. But at the same it is good to try not to replicate code when it makes sense and stick to a good practice.

skoudoro:
The goal of this PR is to bring all our interpolation functions in one place. By doing that, It seems that some functions are doing exactly the same thing. This part needs a clean up on a new PR.

This PR should fix #728.

Let me know if there is an interpolation function that I forgot to move in this module


arokem:
Do you want to include this one? https://github.com/nipy/dipy/blob/a23239177cc8b779adfc7bf5b66275078203e7a2/dipy/core/sphere.py#L499

skoudoro:
good catch! let's add this function too!

arokem:
This looks overall good to me.

@skoudoro : could you please rebase and resolve conflicts?

arokem:
Once you rebase, I think that the Appveyor failer should go away, as #1899 is now merged.

skoudoro:
rebased and no Appveyor error anymore. this PR is ready. 

Comments are welcomed!

Garyfallidis:
Great, finally we have one place for all interpolation functions. In the future we may want to separate interpolation.pyx to interpolation.py and interpolation_speed.pyx but no need to this now.

Garyfallidis:
Thank you @skoudoro. Please everyone rebase!

PR #1898, opened 2019-07-12
Replace SingleTensor by single_tensor

### LINKED ISSUE #844: Refactor behavior of dipy.sims.voxel.single_tensor vs SingleTensor
Currently, SingleTensor is aliased to single_tensor. Following discussions on Gitter, the behavior should be aligned with what is done in dipy.reconst and with gradient_table versus GradientTable.

single_tensor should stay a function to operate on small data, whereas SingleTensor should become a class that is able to deal with large datasets.

The following are the important parts of the discussion: (can't seem to find a way to link to the gitter discussion directly):

jchoude

> We still have 2 functions (dipy.sims.voxel.SingleTensor and dipy.sims.voxel.single_tensor), where SingleTensor just points to single_tensor? From the code, I see that it's to maintain backwards compatibility. Should we replace the last remaining calls to SingleTensor and move it to the Deprecation path?
> Right now it might be confusing to a new user, since there are 2 functions to do the same thin gin the Doc.

arokem

> +1 from me

Garyfallidis

> Although the comment says that this is for backwards compatibility there is also another point. These simulations are still clearly expressed as functions. The proper object oriented approach is not available and it is something to think better if we want efficient simulations.
> This is not the only place we have two ways of accessing different algorithms.

jchoude

> Hum ok... What do you suggest in this case? I just leave it like that for now?

Garyfallidis

> Another place where something like that happened is with gradient_table and GradientTable . Those are a bit different in behaviour.

Garyfallidis

> When Oscar from nipype used the sims module to do some full data simulations he noticed that with the current design we repeat a lot of things when in dipy.reconst we have a better design for dealing with big data.

Garyfallidis

> Yes, I think we need to think a bit about this.
> We may keep the functions for single voxels and change the classes to work with full data.

skoudoro:
This PR removes the backward compatibilities with `SingleTensor`, `MultiTensor` and StickAndBall`.

`dipy.sims.voxel.SingleTensor` has been replaced by `dipy.sims.voxel.single_tensor`
`dipy.sims.voxel.MultiTensor` has been replaced by `dipy.sims.voxel.multi_tensor`
`dipy.sims.voxel.SticksAndBall` has been replaced by `dipy.sims.voxel.sticks_and_ball`

This should fix #844

### DIFF
--- a/dipy/data/__init__.py
+++ b/dipy/data/__init__.py
@@ -14,7 +14,7 @@ import gzip
 import numpy as np
 from dipy.core.gradients import GradientTable, gradient_table
 from dipy.core.sphere import Sphere, HemiSphere
-from dipy.sims.voxel import SticksAndBall
+from dipy.sims.voxel import sticks_and_ball
 from dipy.data.fetcher import (fetch_scil_b0,
                                read_scil_b0,
                                fetch_stanford_hardi,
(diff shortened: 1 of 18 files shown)


arokem:
Looks good to me too. I'll go ahead and merge

PR #1899, opened 2019-07-12
Fix: Byte ordering error on Python 3.5

skoudoro:
I believe this PR should fix #1642 and more specifically, the following error we have seen intermittently:

`ValueError: cannot expose native-only dtype 'g' in non-native byte order '<' via buffer interface`

Let's see

### DIFF
--- a/dipy/reconst/peak_direction_getter.pyx
+++ b/dipy/reconst/peak_direction_getter.pyx
@@ -8,17 +8,6 @@ from dipy.tracking.local.direction_getter cimport DirectionGetter
 cdef extern from "dpy_math.h" nogil:
     double dpy_rint(double x)
 
-def make_nd(array, N):
-    """Makes an array that's less than Nd - Nd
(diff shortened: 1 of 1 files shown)


arokem:
Big if true

skoudoro:
Hi @arokem, it seems to go well :crossed_fingers: , but I think we should try to rerun multiple times python3.5 builds. Can you do it for Appveyor when it is done? It seems that I can't...

arokem:
Currently rebuilding on Appveyor. Two things: 

1. I am not sure why you don't have permissions to kick off a rebuild on appveyor. You can always trigger a rebuild by doing a `git commit --amend`, changing your last commit message and then force pushing. 

2. This change looks good whether or not it fixes that particular issue, so I am +1 for merging it either way :-)

arokem:
OK. Since no one complained here, I am going to go ahead and merge.

Issue #1718, opened 2019-01-08
cannot import name window


muyinglin:
I installed dipy,when I import dipy.viz,it's right,but when I "from dipy.viz import window, actor"，the error warning is "cannot import name window".

skoudoro:
Hi @muyinglin,

Did you install Fury ? this is an optional dependency for visualisation. 

`pip install fury`

skoudoro:
closing, If you still have this problem, Feel free to reopen it @muyinglin.

muyinglin:
I'm sorry, but I really don't understand your meaning,could you please explain it in detail,thank you very much.

skoudoro:
You need to install a python package named FURY.

To install this package, you need to run this command on a terminal :  `pip install fury` 

After installing this package, your problem should disappear. Let me know if it works

muyinglin:
Ok,this worked,thank you very much.

PR #1902, opened 2019-07-16
Warn users if they don't have FURY installed 

arokem:
And provide a link to information about install

### DIFF
--- a/dipy/viz/__init__.py
+++ b/dipy/viz/__init__.py
@@ -1,10 +1,13 @@
 # Init file for visualization package
 from __future__ import division, print_function, absolute_import
 
-
 from dipy.utils.optpkg import optional_package
 # Allow import, but disable doctests if we don't have fury
-fury, have_fury, _ = optional_package('fury')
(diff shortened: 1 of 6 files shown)


arokem:
This better addresses such as #1718

dipy/viz/__init__.py
```diff
@@ -1,10 +1,13 @@
...
 # Allow import, but disable doctests if we don't have fury
-fury, have_fury, _ = optional_package('fury')
+fury, has_fury, _ = optional_package(
+    'fury',
+    "You do not have FURY installed. Some visualization functions"
+    "might not work for you. For installation instructions, please visit: "
+    "https://fury.gl/")
 
 
 if have_fury:
```
skoudoro: Can you update this variable `have_fury` to `has_fury`

arokem: Updated here and elsewhere to `has_fury` to be consistent with `has_mpl` below.

skoudoro:
Thank you for this @arokem

PR #1903, opened 2019-07-16
Skip some tests on big endian architecture (like s390x)

### LINKED ISSUE #1587: Could tests for functionality not supported on big endians just skip?
## Description
in debian on s390x we have a number of tests failing thus failing the entire build process:
https://buildd.debian.org/status/fetch.php?pkg=dipy&arch=s390x&ver=0.14.0-1&stamp=1526401155&raw=0
I wondered if you could
1. Move from `ValueError` to some dedicated `ArchitectureNotSupportedError` derived from `NotImplementedError` exception?
2. decorate such tests (or may be possible to adjust general runner?) so they would catch such exception and skip the test?

Cheers!

skoudoro:
this PR should resolve #1587.

@arokem and @yarikoptic, Can you have a look? 

I do not know how we can add a test. Any idea?

### DIFF
--- a/dipy/segment/tests/test_bundles.py
+++ b/dipy/segment/tests/test_bundles.py
@@ -11,7 +11,7 @@ from dipy.tracking.distances import bundles_distances_mam
 from dipy.tracking.streamline import Streamlines
 from dipy.segment.clustering import qbx_and_merge
 
-is_big_endian = 'big' in sys.byteorder
+is_big_endian = 'big' in sys.byteorder.lower()
 
 
 def setup_module():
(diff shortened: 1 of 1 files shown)


arokem:
This looks good to me.

Anyone have access to a big 🔚-ian machine?

skoudoro:
It seems that no one has access to this kind of machine. What should we do @arokem with this PR?

arokem:
I say let’s merge it, to be on the safe side

Garyfallidis:
This looks safe to merge because it only skips a few tests on big endian. So merging...

Issue #1870, opened 2019-06-24
Documentation example  of MSMT CSD


arokem:
Follow up on #1858

skoudoro:
Hi @ShreyasFadnavis, 

What is the status of this? Any chance to have it ready for the release next week?

ShreyasFadnavis:
@skoudoro Yes, I will follow up with a PR shortly!
Thanks :+1:

ShreyasFadnavis:
Done!
Thanks!! @arokem and @skoudoro 👍

PR #1909, opened 2019-07-18
[WIP] - MTMS-CSD Tutorial

ShreyasFadnavis:
This tutorial is also intended to delineate DIPY's ability to create pipelines.
Steps involved:
- Denoising
- Registration
- Tissue Classifier
- MCSD

Addresses #1870

### DIFF
--- a/doc/examples/reconst_mcsd.py
+++ b/doc/examples/reconst_mcsd.py
@@ -1,0 +1,136 @@
+import numpy as np
+from dipy.reconst.mcsd import MultiShellResponse
+from dipy.reconst.csdeconv import auto_response
+from dipy.io import read_bvals_bvecs
+from dipy.core.gradients import gradient_table
+from dipy.segment.tissue import TissueClassifierHMRF
+from dipy.io.image import load_nifti, save_nifti
+from dipy.data import fetch_cfin_multib, read_cfin_dwi, read_cfin_t1
+from dipy.sims.voxel import (single_tensor)
+from dipy.reconst import shm
+from dipy.data import default_sphere
+from dipy.core.gradients import GradientTable
+from dipy.denoise.nlmeans import nlmeans
+from dipy.denoise.noise_estimate import estimate_sigma
+import dipy.reconst.dti as dti
+from dipy.reconst.mcsd import MultiShellDeconvModel
+from dipy.viz import window, actor
+from dipy.data import get_sphere
+sphere = get_sphere('symmetric724')
+
+fetch_cfin_multib()
+img, gtab = read_cfin_dwi()
+
+data = img.get_data()
+data_small = data[40:65, 50:51]
+
+t1 = read_cfin_t1()
+t1_data = t1.get_data()
+t1_data_small = np.array(t1_data[40:65, 50:51, :19], dtype='float64')
+
+"""
+Now we will define the other two parameters for the segmentation algorithm.
+We will segment three classes, namely corticospinal fluid (CSF), white matter
+(WM) and gray matter (GM).
+"""
+nclass = 3
+"""
+Then, the smoothness factor of the segmentation. Good performance is achieved
+with values between 0 and 0.5.
+"""
+beta = 0.1
+
+# denoising
+#sigma = estimate_sigma(t1_data_small, True, N=6)
+
+#t1[mask == 0] = 0
+
+#t1_den = nlmeans(t1_data_small, sigma=sigma)
+
+bvals, bvecs = gtab.bvals, gtab.bvecs
+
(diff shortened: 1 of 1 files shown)


ShreyasFadnavis:
Hi all! I am facing some issues with the data for this example.. The only viable option that I see from the fetcher is CFIN, but needs some brain extraction on the T1.
I have a working example with the HCP data, but we do not have a fetcher for it! any suggestions would be super helpful..

jhlegarreta:
The HCP data requires a username/pwd to be able to download it. @arokem mentioned in PR #1900 that the terms and conditions of the HCP data prevent to redistribute them, and this is one of the reasons for having the CENIR dataset in DIPY. CENIR does not have T1-weighted images, though.

I had a quick look at the [UW-Minn HCP data terms](https://www.humanconnectome.org/study/hcp-young-adult/document/wu-minn-hcp-consortium-open-access-data-use-terms), and according to point 4, data and derivatives may be re-distributed under the same data use terms. So for that part of the data may be DIPY could comply with those restrictions? May be confirmation should be asked for to UW-Minn HCP team?

If that is feasible and the data of interest for DIPY is part of the UW-Minn HCP data, then may be it could be hosted in UW's digital library, as it is the case of the CENIR data, and the corresponding fetcher could then be built?

ShreyasFadnavis:
> The HCP data requires a username/pwd to be able to download it. @arokem mentioned in PR #1900 that the terms and conditions of the HCP data prevent to redistribute them, and this is one of the reasons for having the CENIR dataset in DIPY. CENIR does not have T1-weighted images, though.
> 
> I had a quick look at the [UW-Minn HCP data terms](https://www.humanconnectome.org/study/hcp-young-adult/document/wu-minn-hcp-consortium-open-access-data-use-terms), and according to point 4, data and derivatives may be re-distributed under the same data use terms. So for that part of the data may be DIPY could comply with those restrictions? May be confirmation should be asked for to UW-Minn HCP team?
> 
> If that is feasible and the data of interest for DIPY is part of the UW-Minn HCP data, then may be it could be hosted in UW's digital library, as it is the case of the CENIR data, and the corresponding fetcher could then be built?

Hi @jhlegarreta ! I completely agree with you. I have reached out to the HCP team and am awaiting a response. In the meantime, let me also take a look at the clauses of UW-Minn HCP ✔

arokem:
Sorry - I believe that the HCP data is a no go for us. If I understand correctly, you can redistribute the data, but you have to put in place a mechanism for downstream users to also agree to the original terms of conditions, and we can't do that in our software. Please use the CENIR data, or some other data. Both here and in #1900.

ShreyasFadnavis:
Hi @arokem , the issue is that CENIR does not have a T1. The next option that we have using CFIN, but is not extracted, will need something like BET + registration to do this.. Can do this, but I dont want to go outside DIPY tools for the tutorial. Does this make sense?

jhlegarreta:
@skoudoro Sorry to chime in the middle of the release process, but I am wondering why the file added in this example is not shown in the `master` branch (either in GitHub or when fetching the branch locally). Am I missing something?

BTW, was the example left out of the [examples_index.rst](https://github.com/nipy/dipy/blob/master/doc/examples_index.rst) on purpose (e.g. needs polishing)?

skoudoro:
Strange... 

- First, This tutorial should not be merged it is not ready at all...
- Second, I do not remember that I merged this PR, it is not even on my mailbox, 
- Third, I do not see it on master too...

Something weird happens.... Github bug?  need to investigate....

Thank you for pinging me @jhlegarreta !

skoudoro:
I really do not know..... Somehow, this PR is linked to #1931 which I merged yesterday.  I suppose #1931 was build on the top of this branch but that's still weird, I really do not know what happens here! @arokem @Garyfallidis, @ShreyasFadnavis any idea?

arokem:
Looks like it wasn't actually merged at all, right? So no big deal (though mysterious...). I think that @ShreyasFadnavis would need to create a new PR. Either from this branch or from a new branch.

ShreyasFadnavis:
@jhlegarreta Thank your for pointing this out! This is scary! @arokem @skoudoro are right that this was never merged is probably a bug! Will open up a new PR for this. Nevertheless, any idea on what data set to use for this example? This tutorial needs T1 (extracted) + DWI...

arokem:
Would a "pseudo-T1" work? You can create one from DWI data with an Anisotropic Power Map (see  e.g., https://gist.github.com/arokem/5d7441902669a4a82389)

ShreyasFadnavis:
Thanks @arokem !! Let me this a shot... it should work I guess ✔

Issue #1906, opened 2019-07-17
Add python_requires=">=3.5"


effigies:
## Description

I noticed that you've dropped tests, and presumably support, for Python 2.7 and 3.4. If you add the `python_requires=">= 3.5"` option to your `setuptools.setup()` call, then Python versions lower than 3.5 will not attempt to install versions built with that option.

It looks like the main effect in version 0.16.0 is that 2.7 and 3.4 manylinux wheels aren't getting built, so it just takes longer to install, but future versions will eventually break, so avoiding the installation would be nice.

skoudoro:
>  If you add the python_requires=">= 3.5" ....

Thank you @effigies, that's a good idea, will do!

skoudoro:
fixed by #1911, closing

PR #1911, opened 2019-07-18
Add python_requires

skoudoro:
As suggested by @effigies, this PR make sure that we use python3.5 as minimal version during the installation step. 

address #1906


arokem:
Thanks!

PR #1914, opened 2019-07-19
[Fix] vol_idx missing in snr_in_cc Tutorial

skoudoro:
This PR fix `snr_in_cc.py` tutorial. Since we are using 4d data for this tutorial, `vol_idx` was missing on `median_otsu` function. I use the opportunity to fix some pep8


ShreyasFadnavis:
+ 1 I was getting the same issue! Thanks @skoudoro 👍

arokem:
Yep. Makes sense. And thanks for all the other fixes as well.

Issue #283, opened 2013-12-10
Spherical deconvolution model can be constructed without specifying a response


stefanv:
This is not mentioned in the docstring

skoudoro:
fixed by #1916, closing

PR #1916, opened 2019-07-21
Spherical deconvolution model CANNOT be constructed without specifying a response

RafaelNH:
I've created this PR basically to address issue #283. 

After analysing the issue, I've decided that the best solution is to remove the response=None option. If one wants to use a predefined response function, it can just defined it on its own risk by calling:
response=(np.array([AD, RD, RD]), S0)

Using an None option is not ideal also because it does not allows the adjustment of S0. At the current version of the CSD implementation, data is normalised based on this inputed S0 estimate, so having it properly estimated might be crucial for the estimation of the fODF shape. 

After revising  the CSD example, I've noticed that it might be hard to ready for non diffusion MRI experts. For instance, it might not be clear that the time consuming response function calibration procedure is optional. To make the example easier to read, I've added some sub-headings. Please let me know your comments and suggestions on this issue.

### DIFF
--- a/dipy/direction/tests/test_bootstrap_direction_getter.py
+++ b/dipy/direction/tests/test_bootstrap_direction_getter.py
@@ -14,6 +14,7 @@ from dipy.sims.voxel import single_tensor, multi_tensor
 
 
 DEFAULT_SH = 4
+response = (np.array([1.5e3, 0.3e3, 0.3e3]), 1)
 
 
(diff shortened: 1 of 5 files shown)


skoudoro:
Thank you @RafaelNH  for fixing this old issue! 

Can you have a look at this one @Garyfallidis?

Garyfallidis:
Thanks @RafaelNH

PR #1917, opened 2019-07-21
A functional implementation of Random matrix local pca.

RafaelNH:
Hi all! I am working on a functional implementation of the Random matrix local PCA for Dipy's next release. I've already added the classifier based on the moments of the Marchenko–Pastur probability distribution according to [Veraart et al., 2016](https://www.ncbi.nlm.nih.gov/pubmed/27523449) -  which seems to pass my added tests. Now, I just need to:

- [x] Add a test of the main localpca function.

- [x] Update the example (If you want to start playing with the new procedure, you can run it in the current example by skipping the sigma estimation part and setting the sigma parameter to None when calling the localpca function - results are looking good and example runs in a couple of minutes). 

In the meantime, I need our opinion on the follow aspect - The new feature does not require the estimation of noise sigma, so what should I do with the previous implemented noise estimate procedure? @arokem, @skoudoro did you mentioned that this procedure is not working well? The current version of the code is still compatible with this previous procedure; however if it is not working I don't mind removing it to simplified a bit the code. 

PS: Also I hope this work helps the future integration of the code in PR#1781 in a future release.

### DIFF
--- a/dipy/denoise/localpca.py
+++ b/dipy/denoise/localpca.py
@@ -18,7 +18,7 @@ def _pca_classifier(L, nvoxels):
     ----------
     L : array (n,)
         Array containing the PCA eigenvalues in ascending order.
-    wdim : int
+    nvoxels : int
         Number of voxels used to compute L
 
     Returns
@@ -50,8 +50,8 @@ def _pca_classifier(L, nvoxels):
     return var, ncomps
 
 
-def localpca(arr, sigma=None, mask=None, pca_method='eig', patch_radius=2,
-             tau_factor=None, return_sigma=False, out_dtype=None):
+def genpca(arr, sigma=None, mask=None, patch_radius=2, pca_method='eig',
+           tau_factor=None, return_sigma=False, out_dtype=None):
     r"""General function to perform PCA-based denoising of diffusion datasets.
 
     Parameters
@@ -67,14 +67,14 @@ def localpca(arr, sigma=None, mask=None, pca_method='eig', patch_radius=2,
         A mask with voxels that are true inside the brain and false outside of
         it. The function denoises within the true part and returns zeros
         outside of those voxels.
+    patch_radius : int (optional)
+        The radius of the local patch to be taken around each voxel (in
+        voxels). Default: 2 (denoise in blocks of 5x5x5 voxels).
     pca_method : 'eig' or 'svd' (optional)
         Use either eigenvalue decomposition (eig) or singular value
         decomposition (svd) for principal component analysis. The default
         method is 'eig' which is faster. However, occasionally 'svd' might be
         more accurate.
-    patch_radius : int (optional)
-        The radius of the local patch to be taken around each voxel (in
-        voxels). Default: 2 (denoise in blocks of 5x5x5 voxels).
     tau_factor : float (optional)
         Thresholding of PCA eigenvalues is done by nulling out eigenvalues that
         are smaller than:
@@ -102,13 +102,13 @@ def localpca(arr, sigma=None, mask=None, pca_method='eig', patch_radius=2,
 
     References
     ----------
-    .. [1] Veraart J, Fieremans E, Novikov DS. 2016. Diffusion MRI noise
-           mapping using random matrix theory. Magnetic Resonance in Medicine.
-           doi: 10.1002/mrm.26059.
-    .. [2] Veraart J, Novikov DS, Christiaens D, Ades-aron B, Sijbers,
+    .. [1] Veraart J, Novikov DS, Christiaens D, Ades-aron B, Sijbers,
            Fieremans E, 2016. Denoising of Diffusion MRI using random matrix
            theory. Neuroimage 142:394-406.
            doi: 10.1016/j.neuroimage.2016.08.016
+    .. [2] Veraart J, Fieremans E, Novikov DS. 2016. Diffusion MRI noise
+           mapping using random matrix theory. Magnetic Resonance in Medicine.
+           doi: 10.1002/mrm.26059.
     .. [3] Manjon JV, Coupe P, Concha L, Buades A, Collins DL (2013)
            Diffusion Weighted Image Denoising Using Overcomplete Local
            PCA. PLoS ONE 8(9): e73021.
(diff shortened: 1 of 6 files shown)


skoudoro:
Hi @RafaelNH, Thank you for this!

>  The new feature does not require the estimation of noise sigma so what should I do with the previous implemented noise estimate procedure?

If it does not need it, I do not think so

ShreyasFadnavis:
@RafaelNH Thank you for this wonderful PR! Can we do this.. can we add the `noise_classifier` to the `noise_estimate.py` ? with an option to switch between `lpca` and `mppca` ?  Does this make sense?

Garyfallidis:
@RafaelNH not sure if @ShreyasFadnavis  was clear. We need to keep the existing lpca implementation and add the mppca as a new function. Do NOT replace lpca.

RafaelNH:
@Garyfallidis and @ShreyasFadnavis, I've tried to address your comment of having a separate function for mppca; however, when doing this I was just replicating code. 

As I mentioned in our personal chat, the mppca can be seen as a general version of the localpca algorithm proposed by Manjon et al. In general, the core of pca denoising algorithm is thresholding the eigenvalues of principal components using the following expression:
 
th = (tau * sigma) ** 2

where th is the threshold, sigma is the noise std estimate and tau is a scaling factor of the relationship between sigma and eigenvalue threshold.  While Manjon et al. used a prior estimation for sigma and an empirical value for tau, Veraart et al. showed that optimal values for tau and sigma can be computed from random matrix theory. For instance, tau can be given as (1+sqrt(N/M)) where N is the number of diffusion-weighted images and M the number of voxels of the sliding windows. As you can see in that expression, having a fixed empirical values is far to be optimal when running this algorithm for different diffusion datasets or when changing the sliding window patch radius. 

In this PR, I basically generalize the localpca function and changed its defaults setting for mppca. However, if you are still want to use Manjom suggestions to can run the algorithm as:

den = localpca(data, sigma=sigma_estimate, tau_factor=2.3)

Please let me know your comments.

dipy/denoise/localpca.py
```diff
@@ -10,39 +10,86 @@
 from scipy.linalg import eigh
 
 
-def localpca(arr, sigma, mask=None, pca_method='eig', patch_radius=2,
-             tau_factor=2.3, out_dtype=None):
+def _pca_classifier(L, m):
+    """ Classifies which PCA eigenvalues are related to noise and estimates the
+    noise variance
+
+    Parameters
```
ShreyasFadnavis: I guess `m` needs to be documented here...

dipy/denoise/localpca.py
```diff
@@ -10,39 +10,86 @@
 from scipy.linalg import eigh
 
 
-def localpca(arr, sigma, mask=None, pca_method='eig', patch_radius=2,
-             tau_factor=2.3, out_dtype=None):
+def _pca_classifier(L, m):
```
ShreyasFadnavis: @RafaelNH would it be better to relocate this function to `noise_estimate.py` ? This will help improve the design from a user standpoint I guess..

dipy/denoise/localpca.py
```diff
@@ -10,39 +10,86 @@
...
+    r = L[c] - L[0] - 4 * np.sqrt((c + 1.0) / m) * var
+    while r > 0:
+        var = np.mean(L[:c])
+        c = c - 1
+        r = L[c] - L[0] - 4 * np.sqrt((c + 1.0) / m) * var
+    ncomps = c + 1
+    return var, ncomps
+
+
+def localpca(arr, sigma=None, mask=None, pca_method='eig', patch_radius=2,
```
ShreyasFadnavis: If we move the `_noise_classifier` function to `noise_estimate.py` then, we can have an option here that says `method='lpca'` or `method='mppca'` ? This will also help if we need to add any other noise estimation methods in the future.. Does this make sense? Your current design is good and should be able to support this capability!

Garyfallidis:
Is this still a WIP?

RafaelNH:
@Garyfallidis - yes, still WIP, but I am almost there - I will finish it tomorrow morning!

dipy/denoise/localpca.py
RafaelNH: Done

dipy/denoise/localpca.py
RafaelNH: That would make sense. However, the _pca_classifier is just an auxiliar function for mppca. The noise estimate from mppca requires the full localpca body, so having the noise estimation from these method on the noise_estimate.py file will imply having to duplicate the full localpca code there.

RafaelNH:
Hi all - examples updated! This should be a nice first version of the mp-pca for release 1.0. Let me know if there are further comment to address before merging. 

@ShreyasFadnavis - I agree that it would be nice to have this method in the noise_estimate.py file. However, as I've mentioned above this might not be a straightforward thing to do. We really need time to think and discuss how to do this without requiring code duplicates. I will suggest to do this in a following PR after this one being merged. At the moment, I am explaining how to obtain the different noise std estimates in the two PCA example files.

dipy/denoise/localpca.py
ShreyasFadnavis: I understand! Sounds good! Resolving this 👍

dipy/denoise/localpca.py
```diff
@@ -145,19 +209,42 @@ def localpca(arr, sigma, mask=None, pca_method='eig', patch_radius=2,
...
+                this_theta = 1.0 / (1.0 + dim - ncomps)
                 theta[ix1:ix2, jx1:jx2, kx1:kx2] += this_theta
                 thetax[ix1:ix2, jx1:jx2, kx1:kx2] += Xest * this_theta
+                if return_sigma is True and sigma is None:
+                    var[ix1:ix2, jx1:jx2, kx1:kx2] += this_var * this_theta
+                    thetavar[ix1:ix2, jx1:jx2, kx1:kx2] += this_theta
 
     denoised_arr = thetax / theta
     denoised_arr.clip(min=0, out=denoised_arr)
     denoised_arr[~mask] = 0
```
ShreyasFadnavis: @RafaelNH this is not your fault, but just a previous bug I guess: 
Can you change this to..
`denoised_arr[mask==0] = 0`

doc/examples/denoise_localpca.py
```diff
@@ -1,14 +1,15 @@
 """
-===============================
-Denoise images using Local PCA
-===============================
+=======================================================
+Denoise images using Local PCA and empirical thresholds
```
ShreyasFadnavis: Denoise images using Local PCA **via** empirical thresholds

doc/examples/denoise_mppca.py
```diff
@@ -0,0 +1,282 @@
...
+Denoise images using the Marcenko-Pastur PCA algorithm
+======================================================
+
+The PCA-based denoising algorithm exploits the redundancy across the
+diffusion-weighted images [Manjon2013]_, [Veraart2016a]_. This algorithm has
+been shown to provide an optimal compromise between noise suppression and loss
+of anatomical information for different techniques such as DTI [Manjon2013]_,
+spherical deconvolution [Veraart2016a] and DKI [Henri2018]_.
+
+The basic idea behind the PCA-based denoising algorithm is to remove the
```
ShreyasFadnavis: The basic idea behind the PCA-based denoising algorithms is to remove the
Principal Components of the data that are mostly classified as noise.

doc/examples/denoise_mppca.py
```diff
@@ -0,0 +1,282 @@
...
+data's principal components classified being mostly related to noise.
+The principal components classification can be performed based on prior noise
+variance estimates [Manjon2013]_ (see :ref:`denoise_localpca`)or automatically
+based on the Marcenko-Pastur distribution [Veraa2016a]_. In addition to noise
+suppression, the PCA algorithm can be used to estimate the noise standard
```
ShreyasFadnavis: In addition to the noise
suppression, the PCA algorithm can be used to get the standard deviation of the estimated noise [Veraa2016b].

doc/examples/denoise_mppca.py
```diff
@@ -0,0 +1,282 @@
...
+Below we show how the denoised data can be saved.
+"""
+
+nib.save(nib.Nifti1Image(denoised_arr,
+                         affine), 'denoised_mppca.nii.gz')
+
+print("Entire denoised data saved in denoised_mppca.nii.gz")
+
+"""
+Additionally, we show on this example how the PCA denoising algorithm effects
```
ShreyasFadnavis: Additionally, we show **in** this example how the PCA denoising algorithm **affects**
different diffusion measurements.

Garyfallidis:
I like the design. But agree also with Serge that we need an mppca utility function that calls localpca. Will help with the tutorial too. Also on the local pca function have Manjon cited first. And on the mppca have Veraart cited first.

Garyfallidis:
Apart from these few comments that we have here. This is getting super close to be merged and be included to this release. Rafael please act quickly! Thank you in advance.

RafaelNH:
@Garyfallidis, @skoudoro - Done! Many thanks for your comments, I've created a general pca function (genpca) containing the main body for both pca algorithms (localpca and mppca). I've only maintained the essential input parameters for both individual localpca and mppca functions. I thinks this indeed improves a lot the functions design.

15thai:
Hi, I just saw that you have a great work on Random matrix local pca already. I just have one stupid question. 
If your 3D patch data from `patch_radius` to `arr.shape - patch_radius`. Does the border problem after reconstructing denoised image still exist? 
Thank you.

skoudoro:
Yes, I agree, @15thai, I think there is a small border problem but it is easy to fix with `np.pad` with the patch radius. It will be for the release 1.1

PR #1923, opened 2019-07-26
[Fix] removing minmax_norm parameter from peak_direction

### LINKED ISSUE #389: minmax_norm in peaks_directions does nothing
In the function dipy.recont.peaks.peak_directions, the option minmax_norm does not seem used. A proper normalization is done in peaks from model, but the old, inner function still accepts the arguments without doing anything.

The doc doesn't even speak about it anymore. So should it be removed, or should it be re-enabled so that one can use the function on an odf (while using peaks_from model on an odf doesn't really seem intuitive)?

skoudoro:
The goal of this PR is to fix #389.  I use the opportunity to fix one test not related

### DIFF
--- a/dipy/tracking/tests/test_life.py
+++ b/dipy/tracking/tests/test_life.py
@@ -162,7 +162,7 @@ def test_fit_data():
     gtab = grad.gradient_table(fbval, fbvec)
     ni_data = nib.load(fdata)
     data = ni_data.get_data()
-    tensor_streamlines, _ = load_tractogram(fstreamlines)
+    tensor_streamlines = nib.streamlines.load(fstreamlines).streamlines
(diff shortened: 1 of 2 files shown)


dipy/tracking/tests/test_life.py
```diff
@@ -162,7 +162,7 @@ def test_fit_data():
     gtab = grad.gradient_table(fbval, fbvec)
     ni_data = nib.load(fdata)
     data = ni_data.get_data()
-    tensor_streamlines, _ = load_tractogram(fstreamlines)
```
gabknight: Could you also remove the import for load_tractogram

skoudoro: Indeed, Thank you! Done in db26854

Garyfallidis:
Fixes!! Yep! :)

PR #1925, opened 2019-07-26
FIX: Stateful tractogram examples

frheault:
After the new merge of PR #1812, most examples were not running.

Considering that before most trk saved and loaded by the tutorials were not valid, this PR is not only updating the function call, but also fixing the resulting TRK files involved.

PR #1926, opened 2019-07-27
RF - dipy.tracking.local

gabknight:
Addresses in part #1501 

- move the module `dipy.tracking.local` into `dipy.tracking` 
- renamed `dipy.tracking.localtracking.py` to `dipy.tracking.local_tracking.py`
- renamed `dipy.tracking.tissue_classifier.TissueClass` to `dipy.tracking.tissue_classifier.StreamStatus` 
- renamed `TissueClassifier` class to `StoppingCriterion`
  - all sub classes
  - all related variable names
  - all related file names
- remove temporary fix, while requirements for cython weren't >=0.21


gabknight:
I'm not sure what is the standard here for file names.

- **tissue_classifier.py** or tissueclassifier.py
- **localtracking.py** or local_tracking.py

Which are better? the bold are the current ones.

gabknight:
Waiting on #1925 to fix the examples with this RF.

Do we want to change more here? We briefly discussed the ``TissueClassifier`` renaming at some point to disentangle it with for instance ``TissueClassifierHMRF``, which are not the same kind of object. That could be for instance ``MaskingStrategy``, ``BinaryMaskingStratery``. ``ActMaskingStrategy``.  or ``StoppingCriterion``, ``ThresholdStoppingCriterion``, ``ActStoppingCriterion``. What do you think  @Garyfallidis @skoudoro @arokem ?

skoudoro:
Thank you for this work @gabknight! Concerning the naming of `TissueClassifier`, I like the `StoppingCriterion`, but I need to discuss with @Garyfallidis.

skoudoro:
Concerning the filenames, I am not sure we have a standard rule. But since we are following the pep8, this is what it said: 

> Modules should have short, all-lowercase names. Underscores can be used in the module name if it improves readability. Python packages should also have short, all-lowercase names, although the use of underscores is discouraged.

So I will prefer `local_tracking.py and `tissue_classifier.py`

skoudoro:
#1925 is in, you can rebase @gabknight 😄

Garyfallidis:
What about something more generic? ``BinaryStrategy``, ``ThresholdedStrategy``, ``AnatomicallyConstrainedStrategy`` etc. We could say StoppingStrategy but too many SSs.

Garyfallidis:
Other ideas?

gabknight:
Thanks @skoudoro @Garyfallidis. I'm fine with both *StoppingStrategy or *StoppingCriterion. *Strategy is not descriptive enough in my opinion and *TissueClassifier is also long.

gabknight:
With the exception of `tissue_classifier` renaming. This PR is ready. I updated the description with the main changes. This PR ended up being bigger than I expected, but most changes are simple variable/import renaming.

dipy/tracking/tissue_classifier.pxd
```diff
@@ -1,5 +1,5 @@
 
-cdef enum TissueClass:
+cpdef enum StreamlineStatus:
```
skoudoro: Why the `cdef` became a `cpdef`?

doc/api_changes.rst
```diff
@@ -54,8 +54,19 @@ The `voxel_size` parameter has been removed from the following function:
...
+The `LocalTracking` and `ParticleFilteringTracking` functions were moved from
+``dipy.tracking.local.localtracking`` to ``dipy.tracking.local_tracking``.
+They now need to be imported from ``dipy.tracking.local_tracking``.
+
+The `TissueClassifiers` were moved from ``dipy.tracking.local.tissue_classifier``
+to ``dipy.tracking.tissue_classifier``.
+They now need to be imported from ``dipy.tracking.tissue_classifier``.
+
+The ``dipy.tracking.tissue_classifier.TissueClass`` was renamed
+``dipy.tracking.tissue_classifier.StreamStatus`.
```
skoudoro: typo? `StreamStatus` -> `StreamlineStatus`

dipy/tracking/tissue_classifier.pxd
gabknight: It needs to be accessed in python files. Enum could not be accessed in python file before cython 0.21. We had it redefined in tissue_classifier.py before. I removed it, and set the cpdef, following the new cython minimum requirements.

doc/api_changes.rst
gabknight: yes! good catch. thanks.

dipy/tracking/tissue_classifier.pxd
Garyfallidis: That should not have any performance changes right? Let's make sure.

Garyfallidis:
Fine with StoppingCriterion. Go ahead @gabknight!

doc/examples/tracking_introduction_eudx.py
```diff
@@ -15,8 +15,8 @@
 
 In order to perform local fiber tracking, three things are needed: 1) A method
 for getting directions from a diffusion data set. 2) A method for identifying
-different tissue types within the data set. 3) A set of seeds from which to
-begin tracking.  This example shows how to combine the 3 parts described above
+when the tracking must stops. 3) A set of seeds from which to
```
Garyfallidis: stop

dipy/tracking/tissue_classifier.pxd
gabknight: Thanks for the comment. Tested, there is no difference in speed.

gabknight:
OK. The renaming is done. Quite a few files modify from this change. Let's see if I missed something.

dipy/tracking/tests/test_stopping_criterion.py
```diff
@@ -41,11 +41,13 @@ def test_binary_tissue_classifier():
...
-                npt.assert_equal(state_float64, TissueTypes.TRACKPOINT)
+                npt.assert_equal(state_boolean,
+                                 int(StreamlineStatus.TRACKPOINT))
+                npt.assert_equal(state_float64,
+                                 int(StreamlineStatus.TRACKPOINT))
             else:
-                npt.assert_equal(state_boolean, TissueTypes.ENDPOINT)
-                npt.assert_equal(state_float64, TissueTypes.ENDPOINT)
+                npt.assert_equal(state_boolean, int(StreamlineStatus.ENDPOINT))
+                npt.assert_equal(state_float64, int(StreamlineStatus.ENDPOINT))
```
Garyfallidis: @gabknight why this needs casting now?

gabknight: I don't know why, but on a single build of Travis, with python 3.5, the test was failing because it was comparing a ``int`` with a ``StreamlineStatus`` Enum type.  Somehow, with that build setup, assert_equal wasn't comparing the int value of the Enum but the actual type ``Enum <StreamlineStatus.ENDPOINT>`` with the ``int`` value on the other side.

Since all other tests passed and all other build worked fine, I just fixed the issue this way. Should I add a comment here?

gabknight: This was the failing build: Python: 3.5
DEPENDS="cython==0.29 numpy==1.9.0 scipy==1.0 nibabel==2.4.0 h5py==2.4.0"

Maybe the numpy version it too old here to correctly handle this.

Garyfallidis: Makes sense. Thanks

PR #1927, opened 2019-07-27
BF - move import to top level

gabknight:
I'm not sure why I getting this now, but it crashes at line 68, `warning` not imported.


skoudoro:
I suppose because of the new release of both packages. Thank you for that. Merging

PR #1933, opened 2019-07-29
Remove deprecated voxel_size from seed_from_mask

skoudoro:
everything in the title 😄!

### DIFF
--- a/dipy/tracking/utils.py
+++ b/dipy/tracking/utils.py
@@ -319,8 +319,8 @@ def subsegment(streamlines, max_segment_length):
         yield output_sl
 
 
-def seeds_from_mask(mask, density=[1, 1, 1], voxel_size=None, affine=None):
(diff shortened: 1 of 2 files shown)


arokem:
Was there no test to catch that?

skoudoro:
> Was there no test to catch that?

No, All tests and tutorials were using `affine` parameter and not the deprecated `voxel_size`

frheault:
There will be a conflict, I have an incoming PR that remaster the whole dipy.tracking.utils to force the affine in most functions. It also change the docstring so the affine description is the same everywhere, delete the functions we agreed upon (confusing and/or not use anywhere)

skoudoro:
ok @frheault, I will just close this PR as soon as you create your PR.

I want to create a PR for removing `save_trk` from `dipy.io.trackvis`. Are you working on this too?

frheault:
I'm not, you can delete the whole file 😉

PR #1934, opened 2019-07-29
Random lpca denoise

15thai:
In this pull request, we submitted a different pre-processing technique to reduce the noise in DWI Random matrix theory. We provide 2 different versions with python and cython. 

[Ref] Veraart J, Fieremans E, Novikov DS (2016) Diffusion MRI noise mapping using random matrix theory. Magnetic resonance in Medicine 76(5), p1582-1593. https://doi.org/10.1002/mrm.26059


setup.py
```diff
@@ -50,14 +50,13 @@
     # filenames to .c filenames, and we probably don't have the .c files.
     sys.path.insert(0, pjoin(dirname(__file__), 'fake_pyrex'))
     # Set setuptools extra arguments
-    # We need nosetests for numpy.testing<=1.15
     extra_setuptools_args = dict(
-        tests_require=['pytest', 'nose'],
```
skoudoro: Why you remove this ?

setup.py
```diff
@@ -66,7 +65,6 @@
 ext_kwargs = {'include_dirs': ['src']}  # We add np.get_include() later
 
 for modulename, other_sources, language in (
-        ('dipy.core.interpolation', [], 'c'),
```
skoudoro: Same comment as above. You need to keep this

15thai: Probably I got it from the old version. 
I will update now

ShreyasFadnavis:
@15thai This PR looks super good now! 2 quick suggestion that I guess you need to make before we proceed with this: 
- Remove the signatures at the beginning of the file, as we do not have them for any other model
- The PR is too big to review at once. It would be a good idea to just remove the Python files and keep only the Cython ones in this PR? 

@RafaelNH @arokem @Garyfallidis @skoudoro does this sound ok?

skoudoro:
Hi @15thai,

Thanks for this implementation and all your updates. I tried to run your code and compare the result with @RafaelNH implementation. Below, the result and some observations:

mppca

- As you can see, @RafaelNH implementation (top row) **gives a slightly better result**
- Your implementation is almost **6 times faster**.

```
MP-PCA rafael:  616.4245405197144 seconds on a 96x96x19x67 volume
MP-PCA your version: 135.27022767066956 seconds on a 96x96x19x67 volume
```
-  After running some tests. I obtain sometimes a negative noise which is weird.... I think there is still a small bug in your implementation.
 
- I rebased and refactored your code for better integration in DIPY. You can see it on https://github.com/15thai/dipy/pull/1. Feel free to merge it on your branch or tell me if you prefer to close this PR and I will create a new clean one.

Sorry again for the late review

15thai:
Hi, @skoudoro  thank you for getting back to me. By any chance, can I see which part that giving  the negative noise? If you can give me the testing that's would be good too. 
I will take a look on how to merge it in my branch also.

skoudoro:
> can I see which part that giving the negative noise?

[here, the sigma give sometimes negative value](https://github.com/15thai/dipy/pull/1/files#diff-274705ac3292cf8a5506c6b475625ee8R206)

> If you can give me the testing that's would be good too

I was running [the test on this line](https://github.com/15thai/dipy/pull/1/files#diff-6ce0def49e0edf7f31a3f0ab363f1ce7R308)

Let me know if you need help.

skoudoro:
Hi @15thai,

- Can you give us an update about this PR?
- Can you rebase this PR, please?
- Did you find the small issue with `sigma`?
- How can we help? this PR seems to be really close to done! let's make it happens in 2020 😄 !!!

15thai:
> Hi @15thai,
> 
> * Can you give us an update about this PR?
> * Can you rebase this PR, please?
> * Did you find the small issue with `sigma`?
> * How can we help? this PR seems to be really close to done! let's make it happens in 2020 smile !!!

Hi @skoudoro. Sorry for the late reply. I have been not working on this one. I will take a look again and let you know

skoudoro:
After `patch2self`, I am closing this PR. Feel free to reopen it @15thai if you find time to finalize it. Let me know if you need any help.

dipy/denoise/denspeed_backup.pyx
```diff
@@ -0,0 +1,261 @@
...
+    cdef double [:, :, :, :] denoised_arr_view = denoised_arr
+    cdef double [:, :, :] noise_arr_view = noise_arr
+    cdef int [:, :, :] mask_arr_view = mask.astype(np.int32)
+
+    cdef int e = 0
+    cdef double tamp = 0.
+    cdef int m2 = 0
+
+    # OPENMP loop over slices
+    for k in prange(0, sizes[2], nogil=True, schedule=static):
```
samcoveney, line 145: but isn't a loop over slices going to reduce the number of samples? Treating each slice separately, the voxels must be drawn from a slice (i.e. 2D), thus reducing the number of samples = voxels. Either it is required to go 'further out' to get voxels from further away in order to satisfy number_of_samples > number_of_features (in which case, the noise level is more likely to change, and the redundancy in the feature dimension is likely to be worse), of number_of_samples < number_of_features (which isn't ideally what we want). Therefore, I would not be surprised that this performs worse than the original version

samcoveney: Sorry but that doesn't generally hold. If you use a 3x3x3 spatial neighborhood, this is 27 voxels, so if N is above 27 then clearly you would have more features than samples... This might explain the slightly worse performance? Would you not tune the spatial neighborhood to guarantee sample > features?

I think I understand your point about the par over slices now though.

PR #1937, opened 2019-07-30
Add a warning about future changes that will happen in dipy.stats.

arokem:
This addresses comment made here: https://github.com/nipy/dipy/issues/1765#issuecomment-515537354


skoudoro:
Thank you @arokem, Even if it is not really necessary,  I will wait for Travis before merging this PR.

PR #1939, opened 2019-07-30
Increase affine consistency in dipy.tracking.utils

frheault:
To increase consistency within Dipy and increase cautious behaviors when coding the affine is now required when it was previously optional. 
Older/unused functions that had complex behavior related to spatial transform and trackvis file format were deleted to prevent confusion.
Docstring is now uniform when describing the affine.

Tests were executed and examples tested.


dipy/tracking/local/localtracking.py
```diff
@@ -111,12 +111,9 @@ def _tracker(self, seed, first_step, streamline):
...
         # Make tracks, move them to point space and return
         track = self._generate_streamlines()
-        seeds = None
-        if self.save_seeds:
-            track, seeds = zip(*track)
-        return utils.move_streamlines(track,
-                                      self.affine,
-                                      seeds=seeds)
+
+        return utils.transform_tracking_output(track, self.affine,
```
skoudoro: What is this function? Why not `transform_streamlines` from `dipy.tracking.streamlines`?

dipy/tracking/tests/test_streamline.py
```diff
@@ -354,7 +353,6 @@ def test_set_number_of_points_memory_leaks():
         streamlines.append(rng.randn(rng.randint(10, 100), 3).astype(dtype))
 
     list_refcount_before = get_type_refcount()["list"]
-    rstreamlines = set_number_of_points(streamlines, nb_points=2)
```
skoudoro: Why do you remove this ?

dipy/tracking/tests/test_streamline.py
```diff
@@ -760,7 +758,6 @@ def test_compress_streamlines_memory_leaks():
         streamlines.append(rng.randn(rng.randint(10, 100), 3).astype(dtype))
 
     list_refcount_before = get_type_refcount()["list"]
-    cstreamlines = compress_streamlines(streamlines)
```
skoudoro: same question here

dipy/tracking/utils.py
```diff
@@ -397,21 +379,17 @@ def seeds_from_mask(mask, density=[1, 1, 1], voxel_size=None, affine=None):
     seeds = seeds.reshape((-1, 3))
 
     # Apply the spatial transform
-    if affine is not None:
+    if seeds.any():
```
skoudoro: I do no understand this Condition

dipy/tracking/utils.py
```diff
@@ -510,7 +490,7 @@ def random_seeds_from_mask(mask, seeds_count=1, seed_count_per_voxel=True,
         seeds = seeds[:seeds_count]
 
     # Apply the spatial transform
-    if affine is not None:
+    if seeds.any():
```
skoudoro: same as above

dipy/tracking/utils.py
```diff
@@ -919,37 +789,29 @@ def unique_rows(in_array, dtype='f4'):
 
 
 @_with_initialize
-def move_streamlines(streamlines, output_space, input_space=None,
-                     seeds=None):
+def transform_tracking_output(tracking_output, affine, save_seeds=False):
```
skoudoro: I think this function should be deleted in favor of `transform_streamlines`. You can add the seed management inside `transform_streamlines`

dipy/tracking/local/localtracking.py
frheault: My reasoning was that transform_streamlines should do one thing and one thing only, to stay as simple as possible.

This function takes a generator as a input that can be a tuple or not, and in 99% of the case where people want to transform_streamlines it is just that.
This function with the generator, the tuple and the saving_seeds option is used only once in all dipy and is specifically for the tracking.

I think merging the two function would add to the confusion. And complexify a function that should remain as simple as simple (transform_streamlines)

dipy/tracking/tests/test_streamline.py
frheault: Unused variable, declared variable without details and useless afterward

dipy/tracking/utils.py
frheault: Before the affine was optional, so it had to be tested if it existed.
Now it is not need (since it is mandatory)

However I realize that a test supposed to check if empty seed mask were crashing was not actully getting tested because there was no affine provided and so was not getting transform.

So to apply a transform to seed, you need at least one (thats why any() is used, any non zero value in the array)

dipy/tracking/utils.py
frheault: As mentioned in the other comment, transforming streamlines is confusing enough, I think this would make the `transform_streamlines` function less clear by making it doing too many thing at once

dipy/tracking/tests/test_streamline.py
```diff
@@ -1110,15 +1109,15 @@ def test_values_from_volume():
...
+            x_sl1 = transform_streamlines(sl1, affine)
+            x_sl2 = transform_streamlines(sl1, affine)
 
             vv = values_from_volume(data, x_sl1, affine=affine)
             npt.assert_almost_equal(vv, ans1, decimal=decimal)
 
             # The generator has already been consumed so needs to be
             # regenerated:
-            x_sl1 = list(ut.move_streamlines(sl1, affine))
+            x_sl1 = list(transform_streamlines(sl1, affine))
```
Garyfallidis: this is already returning a list

dipy/tracking/local/localtracking.py
skoudoro: ok, make sense!

dipy/tracking/tests/test_utils.py
```diff
@@ -258,46 +207,46 @@ def _target(target_f, streamlines, voxel_both_true, voxel_one_true,
...
         npt.assert_raises(ValueError, list, new)
 
     # Test smaller voxels
     affine = np.array([[.3, 0, 0, 0],
                        [0, .2, 0, 0],
                        [0, 0, .4, 0],
                        [0, 0, 0, 1]])
-    streamlines = list(move_streamlines(streamlines, affine))
-    new = list(target_f(streamlines, mask, affine=affine))
+    streamlines = list(transform_streamlines(streamlines, affine))
```
skoudoro: you should remove `list` since `transform_streamlines` return a list and not an iterator like `move_streamlines`

dipy/tracking/utils.py
skoudoro: Thank you for the explanation

skoudoro:
Can you address my small comment above @frheault, and then this PR is ready to go!

skoudoro:
Hi @frheault, I just merge #1926(local tracking) so this PR needs a rebase. Can you do it?  Thank you!

skoudoro:
Thank you @frheault

PR #1943, opened 2019-07-31
Increase affine consistency in dipy.tracking.life and dipy.stats.analysis

frheault:
To increase consistency within Dipy and increase cautious behaviors when coding the affine is now required when it was previously optional.
Older/unused functions that had complex behavior related to spatial transform and trackvis file format were deleted to prevent confusion.
Docstring is now uniform when describing the affine.

Tests were executed and examples tested.

### DIFF
--- a/dipy/tracking/life.py
+++ b/dipy/tracking/life.py
@@ -262,8 +262,7 @@ class LifeSignalMaker(object):
         return sig_out
 
 
-def voxel2streamline(streamline, affine, transformed=False,
-                     unique_idx=None):
+def voxel2streamline(streamline, affine, unique_idx=None):
     """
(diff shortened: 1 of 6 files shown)


dipy/tracking/life.py
```diff
@@ -261,7 +262,7 @@ def streamline_signal(self, streamline):
         return sig_out
 
 
-def voxel2streamline(streamline, transformed=False, affine=None,
+def voxel2streamline(streamline, affine, transformed=False,
```
skoudoro: Here, you forgot to remove `transformed=False` parameter

Garyfallidis:
:+1:

PR #1946, opened 2019-08-01
Fix empty tractogram loading saving

frheault:
As of right now loading and saving empty tractogram is crashing because of the bbox in voxel space being checked.

Addition of a simple condition in the verification.


skoudoro:
Thank you @frheault, Can you add a small test for this? Thank you

frheault:
@skoudoro I added the tests. The is_bbox_in_vox_valid check is technically done in the init, so everything is covered.

skoudoro:
Thank you, waiting for Travis

PR #1947, opened 2019-08-02
DOC: fixing examples links

Garyfallidis:
A strange commit generating invalid example links was added in 04cc19c. Fixing ...


skoudoro:
Awesome! Thank you @Garyfallidis! merging

PR #1952, opened 2019-08-13
Adding lpca, mppca and gibbs ringing workflows

skoudoro:
This PR adds new workflows:

- `LPCAFlow`
- `MPPCAFlow`
- `GibbsRingingFlow`


dipy/workflows/denoise.py
```diff
@@ -49,7 +69,217 @@ def run(self, input_files, sigma=0, out_dir='',
...
+                                           smooth=3)
+                logging.debug('Found sigma {0}'.format(sigma))
+
+            denoised_data = localpca(data, sigma=sigma,
+                                     patch_radius=patch_radius,
+                                     pca_method=pca_method,
+                                     tau_factor=tau_factor)
+            save_nifti(odenoised, denoised_data, affine, image.header)
+
+            logging.info('Denoised volume saved as {0}'.format(odenoised))
```
Borda: you may use `logging.info('Denoised volume saved as %s', odenoised)`
see: https://reinout.vanrees.org/weblog/2015/06/05/logging-formatting.html
"Don’t use the `'{}'.format()`, as it formats the string before it gets passed to the logger. So even if the log message doesn’t need to be actually logged somewhere, the full string gets created.
The proposed example only gets passed a string and some extra arguments and only turns it into a real full string for in your logfile if it actually gets logged. So if you only display WARN level and higher, your DEBUG messages don’t need to be calculated."

Borda: but there is no time difference
```
$ python -mtimeit -s"import logging; logging.info('sample %s', 'text')"
100000000 loops, best of 3: 0.0121 usec per loop
$ python -mtimeit -s"import logging; logging.info('sample {}'.format('text'))"
100000000 loops, best of 3: 0.0127 usec per loop
```

dipy/workflows/tests/test_denoise.py
```diff
@@ -0,0 +1,114 @@
+import os
+import numpy as np
+import numpy.testing as npt
+from dipy.testing import (assert_true, assert_false, assert_greater,
```
Borda: why not use standard `unitests` package?

skoudoro: we are using `numpy.testing`  as you can see above this line, but these functions are missing so we added them to `dipy.testing` to extend `numpy.testing`

Concerning `unitests` package,  I do not see any `assert_greater`, `assert_greater_equal`, `assert_less`, etc....

dipy/workflows/denoise.py
skoudoro: Thank you for this information. I believe that `%` was the **old style** and we should avoid it. I do not know what is the official rule and your reference start to be quite old (2015). I will have a look atthis before updating this code

dipy/workflows/tests/test_denoise.py
Borda: have look at https://docs.python.org/3.6/library/unittest.html
`unittest.TestCase` has almost all asserts you can imagine... :-D

dipy/workflows/denoise.py
Borda: I would say that in case you have lots of logging and not all are always needed - levels info/debug, then the formating output at the time it is really needed (just-in-time) can be useful...

dipy/workflows/tests/test_denoise.py
skoudoro: ok, I see, thanks @Borda. I will fix that and create a new PR to update all DIPY later.

skoudoro: After a quick test, it is really not convenient. I will keep it as it is. As you can see below, I find this syntax a bit heavy and I do want to have a global `TestCase` class instantiation. 

```
>>> from unittest import TestCase
>>> TestCase().assertTrue(1)
```

dipy/workflows/denoise.py
skoudoro: After a quick check on the Python doc, your reference is still up to date. They keep this behavior for backward compatibility and it seems they will keep it for a long time.... there is a workaround but I prefer to update the syntax. Thanks @borda, good to know

dipy/workflows/tests/test_denoise.py
Borda: I think that the good way of using `TestCase` is inheriting test class and call asserts as `self.`

skoudoro:
Can you have a look at this PR @RafaelNH?

skoudoro:
Any other comments? it is ready to go!

arokem:
Looks good! I can merge this in a couple of days if no one has any objections.

PR #1956, opened 2019-08-19
stateful_tractogram_post_1.0_fixes

frheault:
Fix mssing docstring
Added `to_space` function
Useful private attribute are now accessible
Fetcher now uses checksum instead of re-downloading

### DIFF
--- a/dipy/io/stateful_tractogram.py
+++ b/dipy/io/stateful_tractogram.py
@@ -75,6 +75,8 @@ class StatefulTractogram(object):
         if data_per_streamline is None:
             data_per_streamline = {}
 
+        if isinstance(streamlines, Streamlines):
+            streamlines = streamlines.copy()
         self._tractogram = Tractogram(streamlines,
                                       data_per_point=data_per_point,
                                       data_per_streamline=data_per_streamline)
@@ -177,6 +179,8 @@ class StatefulTractogram(object):
         streamlines : list or ArraySequence (list and deepcopy recommanded)
             Streamlines of the tractogram
         """
+        if isinstance(streamlines, Streamlines):
+            streamlines = streamlines.copy()
         self._tractogram._streamlines = Streamlines(streamlines)
         self.data_per_point = self.data_per_point
         self.data_per_streamline = self.data_per_streamline
--- a/dipy/io/stateful_tractogram.py
+++ b/dipy/io/stateful_tractogram.py
@@ -359,7 +359,8 @@ class StatefulTractogram(object):
 
         tmp_streamlines = self.streamlines[indices_to_keep]
         tmp_data_per_point = self._tractogram.data_per_point[indices_to_keep]
-        tmp_data_per_streamline = self._tractogram.data_per_streamline[indices_to_keep]
+        tmp_data_per_streamline =\
+         self._tractogram.data_per_streamline[indices_to_keep]
 
         self._tractogram = Tractogram(list(tmp_streamlines),
                                       data_per_point=tmp_data_per_point,
--- a/dipy/io/stateful_tractogram.py
+++ b/dipy/io/stateful_tractogram.py
@@ -359,7 +359,7 @@ class StatefulTractogram(object):
         tmp_data_per_point = self._tractogram.data_per_point[indices_to_keep]
         tmp_data_per_streamline = self._tractogram.data_per_streamline[indices_to_keep]
 
-        self._tractogram = Tractogram(list(tmp_streamlines),
+        self._tractogram = Tractogram(tmp_streamlines,
                                       data_per_point=tmp_data_per_point,
                                       data_per_streamline=tmp_data_per_streamline,
                                       affine_to_rasmm=np.eye(4))
--- a/dipy/io/stateful_tractogram.py
+++ b/dipy/io/stateful_tractogram.py
@@ -48,8 +48,8 @@ class StatefulTractogram(object):
             the space is RASMM
         shifted_origin : bool
             Information on the position of the origin,
-            False is Trackvis standard, default (corner of the voxel)
-            True is NIFTI standard (center of the voxel)
+            True is Trackvis standard, default (corner of the voxel)
+            False is NIFTI standard (center of the voxel)
         data_per_point : dict
             Dictionary in which each key has X items, each items has Y_i items
             X being the number of streamlines
--- a/dipy/io/stateful_tractogram.py
+++ b/dipy/io/stateful_tractogram.py
@@ -48,8 +48,8 @@ class StatefulTractogram(object):
             the space is RASMM
         shifted_origin : bool
             Information on the position of the origin,
-            True is Trackvis standard, default (corner of the voxel)
-            False is NIFTI standard (center of the voxel)
+            False is Trackvis standard, default (center of the voxel)
+            True is NIFTI standard (corner of the voxel)
         data_per_point : dict
             Dictionary in which each key has X items, each items has Y_i items
             X being the number of streamlines
--- a/dipy/io/stateful_tractogram.py
+++ b/dipy/io/stateful_tractogram.py
@@ -180,7 +180,7 @@ class StatefulTractogram(object):
         self._tractogram._streamlines = Streamlines(streamlines)
         self.data_per_point = self.data_per_point
         self.data_per_streamline = self.data_per_streamline
-        logging.warning('Streamlines has been modified, caution')
+        logging.warning('Streamlines has been modified')
 
     @property
     def data_per_point(self):
@@ -241,6 +241,8 @@ class StatefulTractogram(object):
             self._voxmm_to_rasmm()
 
     def to_space(self, target_space):
+        """ Safe function to transform streamlines to a particular space using
+        an enum and update state """
         if target_space == Space.VOX:
             self.to_vox()
         elif target_space == Space.VOXMM:
--- a/dipy/io/stateful_tractogram.py
+++ b/dipy/io/stateful_tractogram.py
@@ -359,7 +359,7 @@ class StatefulTractogram(object):
         tmp_data_per_point = self._tractogram.data_per_point[indices_to_keep]
         tmp_data_per_streamline = self._tractogram.data_per_streamline[indices_to_keep]
 
-        self._tractogram = Tractogram(tmp_streamlines,
+        self._tractogram = Tractogram(list(tmp_streamlines),
                                       data_per_point=tmp_data_per_point,
                                       data_per_streamline=tmp_data_per_streamline,
                                       affine_to_rasmm=np.eye(4))
--- a/dipy/io/stateful_tractogram.py
+++ b/dipy/io/stateful_tractogram.py
@@ -85,9 +85,9 @@ class StatefulTractogram(object):
                             'Nifti or Trk filename, Nifti1Image or TrkFile, ' +
                             'Nifti1Header or trk.header (dict)')
 
-        (self.affine, self.dimensions,
-         self.voxel_sizes, self.voxel_order) = space_attributes
-        self._inv_affine = np.linalg.inv(self.affine)
+        (self._affine, self._dimensions,
+         self._voxel_sizes, self._voxel_order) = space_attributes
+        self._inv_affine = np.linalg.inv(self._affine)
 
         if space not in Space:
             raise ValueError('Space MUST be from Space enum, e.g Space.VOX')
@@ -101,14 +101,14 @@ class StatefulTractogram(object):
     def __str__(self):
         """ Generate the string for printing """
         text = 'Affine: \n{}'.format(
-            np.array2string(self.affine,
+            np.array2string(self._affine,
                             formatter={'float_kind': lambda x: "%.6f" % x}))
         text += '\ndimensions: {}'.format(
-            np.array2string(self.dimensions))
+            np.array2string(self._dimensions))
         text += '\nvoxel_sizes: {}'.format(
-            np.array2string(self.voxel_sizes,
+            np.array2string(self._voxel_sizes,
                             formatter={'float_kind': lambda x: "%.2f" % x}))
-        text += '\nvoxel_order: {}'.format(self.voxel_order)
+        text += '\nvoxel_order: {}'.format(self._voxel_order)
 
         text += '\nstreamline_count: {}'.format(self._get_streamline_count())
         text += '\npoint_count: {}'.format(self._get_point_count())
@@ -126,13 +126,33 @@ class StatefulTractogram(object):
     @property
     def space_attributes(self):
         """ Getter for spatial attribute """
-        return self.affine, self.dimensions, self.voxel_sizes, \
-            self.voxel_order
+        return self._affine, self._dimensions, self._voxel_sizes, \
+            self._voxel_order
 
     @property
     def space(self):
         """ Getter for the current space """
         return self._space
+
+    @property
+    def affine(self):
+        """ Getter for the reference affine """
+        return self._affine
+
+    @property
+    def dimensions(self):
+        """ Getter for the reference dimensions """
+        return self._dimensions
+
+    @property
+    def voxel_sizes(self):
+        """ Getter for the reference voxel sizes """
+        return self._voxel_sizes
+
+    @property
+    def voxel_order(self):
+        """ Getter for the reference voxel_order """
+        return self._voxel_order
 
     @property
     def shifted_origin(self):
@@ -160,7 +180,7 @@ class StatefulTractogram(object):
         self._tractogram._streamlines = Streamlines(streamlines)
         self.data_per_point = self.data_per_point
         self.data_per_streamline = self.data_per_streamline
-        logging.warning('Streamlines has been modified')
+        logging.warning('Streamlines has been modified, caution')
 
     @property
     def data_per_point(self):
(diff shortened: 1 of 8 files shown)


doc/examples/streamline_formats.py
```diff
@@ -163,6 +163,11 @@
 Now lets move them all to voxel space, subsample them to 100 streamlines,
 compute a density map and save everything for visualisation in another
 software such as Trackvis or MI-Brain.
+
+To access volume information in a grid, the corner of the voxel must be
+considered the origin in order to prevent negative values.
+Any operation doing interpolation or accession in a grid must use the
```
ppoulin91: Accession feels weird... Maybe: "accession in a grid" -> "accessing a grid"?

skoudoro:
Hi @frheault, 

Thank you for this update! is it ready for a review? Do not forget to add a small test for `to_space`.

frheault:
@skoudoro I just added the test, and yes it is ready for review

dipy/io/streamline.py
```diff
@@ -167,158 +176,49 @@ def load_tractogram(filename, reference, to_space=Space.RASMM,
...
+                              trk_header_check=trk_header_check)
+        return sft
+
+    f_gen.__doc__ = load_tractogram.__doc__.replace(
+        'from any format (trk, tck, vtk, fib, dpy)',
+        'of the {} format'.format(ttype))
+    return f_gen
+
+
+def save_generator(ttype):
```
arokem: Docstring

dipy/io/streamline.py
```diff
@@ -167,158 +176,49 @@ def load_tractogram(filename, reference, to_space=Space.RASMM,
...
+load_trk = load_generator('.trk')
+load_tck = load_generator('.tck')
+load_vtk = load_generator('.vtk')
+load_fib = load_generator('.fib')
+load_dpy = load_generator('.dpy')
+save_trk = save_generator('.trk')
+save_tck = save_generator('.tck')
+save_vtk = save_generator('.vtk')
+save_fib = save_generator('.fib')
+save_dpy = save_generator('.dpy')
```
arokem: Nice!

doc/examples/streamline_formats.py
```diff
@@ -74,8 +74,8 @@
                           bbox_valid_check=False)
 
 """
-The function ``load_tractogram`` requires a reference, any of the following input
-is considered valid (as long as they are in the same share space)
+The function ``load_tractogram`` requires a reference, any of the following
+input is considered valid (as long as they are in the same share space)
```
arokem:
```suggestion
inputs is considered valid (as long as they are in the same share space)
```

dipy/io/stateful_tractogram.py
```diff
@@ -161,7 +180,7 @@ def streamlines(self, streamlines):
         self._tractogram._streamlines = Streamlines(streamlines)
         self.data_per_point = self.data_per_point
         self.data_per_streamline = self.data_per_streamline
-        logging.warning('Streamlines has been modified')
+        logging.warning('Streamlines has been modified, caution')
```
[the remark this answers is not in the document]
frheault: Well, it could be that the use shuffled them, resampled them, smoothed them or even took a subset (if there was not data_per_point or data_per_streamlines)

There is no way of knowing what changed exactly, that's why it is a "dangerous" operation and creating a new statefulTractogram is recommanded.

Any suggestion ? Personally I don't know what could be display to the user

frheault:
@GuillaumeTh or @jchoude Any chance for a review, before this get cold ?

dipy/io/stateful_tractogram.py
```diff
@@ -135,6 +134,26 @@ def space(self):
...
+        return self._dimensions
+
+    @property
+    def voxel_sizes(self):
+        """ Getter for the reference voxel sizes """
+        return self._voxel_sizes
+
+    @property
+    def voxel_order(self):
+        """ Getter for the reference voxel_order """
```
GuillaumeTh: Please remove _ in voxel_order

dipy/io/streamline.py
```diff
@@ -167,158 +176,73 @@ def load_tractogram(filename, reference, to_space=Space.RASMM,
...
-                         ' generability use save_tractogram instead.')
-
-    save_tractogram(sft, filename, bbox_valid_check=bbox_valid_check)
+def load_generator(ttype):
+    """ Generate a loading function that performs a file extension 
+    check to restrict the user to a single file format.
 
+    Parameters
+    ----------
+    sft : string
```
GuillaumeTh: sft -> ttype

dipy/io/streamline.py
```diff
@@ -167,158 +176,73 @@ def load_tractogram(filename, reference, to_space=Space.RASMM,
...
-    'vtk format')
-save_fib.__doc__ = save_tractogram.__doc__.replace(
-    'any format (trk, tck, vtk, fib, dpy)',
-    'fib format')
-save_dpy.__doc__ = save_tractogram.__doc__.replace(
-    'any format (trk, tck, vtk, fib, dpy)',
-    'dpy format')
+    Parameters
+    ----------
+    sft : string
```
GuillaumeTh: sft -> ttype

skoudoro:
Hi @frheault,  

LGTM to me too. Can you try to make [pep8speaks](https://github.com/nipy/dipy/pull/1956#issuecomment-522704630) and Codacy happy? it will be good to increase a bit the coverage.

Otherwise, I have a quick question: Most of the time, I encounter this error:

```
ValueError: Bounding box is not valid in voxel space, cannot load a valid file if some coordinates are invalid
```

It is quite annoying... Do we really need to check this bbox by default? I know that I can switch it off by putting `bbox_valid_check=False`, but It seems not intuitive for most of the user since it is the most frequent question I get since the release. So, can we set it up to `False` by default?

dipy/io/stateful_tractogram.py
```diff
@@ -326,25 +357,16 @@ def remove_invalid_streamlines(self):
...
-        for key in self._tractogram.data_per_streamline:
-            tmp_data_per_streamline[key] = \
-                self._tractogram.data_per_streamline[key][indices_to_keep]
+        tmp_streamlines = self.streamlines[indices_to_keep]
+        tmp_data_per_point = self._tractogram.data_per_point[indices_to_keep]
+        tmp_data_per_streamline =\
+         self._tractogram.data_per_streamline[indices_to_keep]
 
-        self._tractogram = Tractogram(tmp_streamlines,
+        self._tractogram = Tractogram(list(tmp_streamlines),
```
skoudoro: list -> Streamlines to avoid potential memory issue

frheault: Actually I encounter a problem recently were the using the ArraySequence was copy the view while keeping the whole underlying ArraySequence data, which is problematic.

Casting to a list was dropping all attribute and just keep streamlines to recreate a clean ArraySequence not using a view, with the right data.

I lost the data, I don't know where it is. But giving only tmp_streamlines was not working. Have you ever encounter this problem and do you think Streamlines() constructor would solve that ?

frheault:
@skoudoro about the bbox_valid_check=False, my honest response would be no. I understand the problem, but it is important to consider that negative value or above the volume are actually invalid.

They should not be there and will make anything in voxel space crash (memory error) and even some external software. I set it to True so people can acknowledge the problem before it happen.

Also it happen more than once that (in the lab) that people created a StatefulTractogram in the wrong space and mixed input and that warning saved them from saving invalid file.

That's my honest answer, however it can be set to False by default and I would understand. But just like forcing a reference and making the affine mandatory everywhere in the code, I think we need to be strict and enforce validity.

Compromise, I suggest a clearer error message with the two potential solution in it to help people solve in a instant

arokem:
> Compromise, I suggest a clearer error message with the two potential solution in it to help people solve in a instant

+100

arokem:
Whoops. Sorry about that.

skoudoro:
> Compromise, I suggest a clearer error message with the two potential solution in it to help people solve in a instant

👍 , let's go for the compromise and a more explicit message with 2 potentials solutions

dipy/io/stateful_tractogram.py
skoudoro: Indeed, really strange.... 

if we go to the compromise again 😄, you can use `tmp_streamlines.data` to get only the data and not the whole `ArraySequence`. 

To answer your question, I think using `Streamlines()` constructor will solve that but I am not 100% sure.

skoudoro:
Thank you for all explanation!

dipy/io/stateful_tractogram.py
frheault: Don't know if that's the expected behaviour, but that was my problem (it was counter intuitive for me at first)
The cleanest solution is to force a copy after the slicing.

`from nibabel.streamlines.array_sequence import ArraySequence
from dipy.tracking.streamline import Streamlines
import numpy as np

# 50 streamlines, 6 points per streamlines
streamlines_a = ArraySequence(np.arange(900).reshape((50,6,3)))
streamlines_b = streamlines_a[0:10]
streamlines_c = ArraySequence(streamlines_a[0:10])
streamlines_d = Streamlines(streamlines_a[0:10])
streamlines_e = ArraySequence(list(streamlines_a[0:10]))
streamlines_f = ArraySequence(streamlines_a[0:10].copy())

# 10 streamlines, 6 points per streamlines
print(len(streamlines_a), len(streamlines_b), len(streamlines_c), len(streamlines_d), len(streamlines_e))

# Should have 300 points for original, 60 points for other
print(len(streamlines_a.data), len(streamlines_b.data), len(streamlines_c.data), len(streamlines_d.data), len(streamlines_e.data), len(streamlines_f.data))`

frheault:
@skoudoro Any suggestion on codecov for coverage ? Do I really have to "Fake test" load/save trk/tck/fib/vtk/dpy ? That are identical to load_tractogram ?

That seems pretty useless considering they are automatically generated and have by design the same behaviour

skoudoro:
I get your point but it is never useless! 

For this case, I will just make sure that your `ValueError` popup as expected. Furthermore, not sure that `load_trk` will work with `*.TRK` so you can add a test for checking upper/smaller case. (Thanks weird files 😄)

dipy/io/stateful_tractogram.py
skoudoro: interesting... When I look at the `ArraySequence` code , your demo make totally sense and we can improve that but is it the desired behavior? 

I will need the opinion of @MarcCote, @Garyfallidis, and @arokem.

Thanks for pointing that out! Let's see if we need a new issue

MarcCote: Your issue seems similar to what was discussed here: https://github.com/nipy/nibabel/issues/729#issuecomment-467959221

I'm still open to suggestion, to make it more intuitive for the user.

For convenience, here's my original answer
"""
That was the intended behavior but I agree it might be unexpected when looking under the hood. The advanced indexing on `ArraySequence` is a lazy operation to avoid creating more memory if we can. If you want to prune the data, you should call [`.copy()`](https://github.com/nipy/nibabel/blob/master/nibabel/streamlines/array_sequence.py#L255-L268) on the `ArraySequence`, like so
```
In [5]: tgram[idx].streamlines.copy().data
Out[5]: 
array([[2, 2, 2],
       [3, 3, 3],
       [4, 4, 4],
       [5, 5, 5],
       [6, 6, 6]])
```
I think we want with that approach (i.e. explicitly calling `.copy()`) so users are aware that operation will be creating additional memory. I'm open to suggestions to make it clearer to the user?
"""

ppoulin91: If I can add to this discussion: I feel like having the public `data` getter return data that wasn't part of the tractogram view is pretty dangerous (and I'm not even talking about dipy functions that use `_data` directly).

I like the idea of having a view to save memory, but I would expect public methods to return what the object "says" it contains, not more than that because it is sharing memory with a larger ArraySequence.

MarcCote: I agree with you, `.data` should be returning the data of the view and nothing more. Unfortunately, I can't see any way of doing that without implicitly creating a copy of the data :-(. Initially, `.data` was private but we made it public so it could be used "legitely" in Dipy's streamlines manipulation (compress, cluster, etc.).

I just discover https://github.com/scikit-hep/awkward-array, maybe ArraySequence should be replaced with that.

frheault:
@skoudoro I just added the tests. It test the load_trk and save_trk against all the other file format.
Since the functions are automatically generated there is no point in testing the other way around. So the generator function is tested by tested on of its output (I chose trk since it is the "default format" everywhere in Dipy.

dipy/io/stateful_tractogram.py
ppoulin91: Interesting, seems like a good replacement for fast streamline operations

MarcCote: @frheault two questions for you. 
1) Why do you need `.data` to be not referring to the original data?
2) Why do you need to modify directly `.data` in your code?

I'm thinking of making `.data` "private" because I agree with @ppoulin91 and it was not intended to be used by the users. The proper way for modifying `.data` is by using `.offsets` and `.lengths` (like in https://github.com/nipy/dipy/blob/master/dipy/tracking/streamlinespeed.pyx#L106).

Users can still modify individual streamline if they want, like so
```python
...
streamline = streamlines[4]
streamline += 10
```

frheault: I think a good old getter and setter function would be needed.
1) When doing slicing, using a view, it is not intuitive that .data is the whole thing. I know it is for memory reason, but like mentioned earlier this error is very easy to do. But it is easily solve by doing copy every times I am getting a view.
2) For example for warping streamlines it is extremely efficient to just get all points at once, throw them at scipy.interpolate and add a something to each point. Or when getting value from a map, so again interpolation.

If you make .data private, but create a getter/setter function for it that deal with it safely and prevent modification of view. I have no problem with that.
I think trying to solve all problem only using attribute is difficult, you can make it private and create a safe function on the side that is meant for specific use case.

skoudoro:
Fair enough.  I still would prefer that this function accept uppercase extension instead of an exception but that's ok. 

+1 for merging now. if it is ok for your @arokem, you can go for it!

dipy/io/stateful_tractogram.py
skoudoro: > Unfortunately, I can't see any way of doing that without implicitly creating a copy of the data :-(

Maybe by using a masked array. Mask is None at the beginning and each call to `__getitem__` just create/update the mask. (ok, we still increase the memory)

In any case, slicing implies jump in memory. Not sure that all streamlines manipulations on DIPY will like this kind of jump. Currently, even if we improve `data`, it is quite dangerous. So, it is not a bad idea to force a copy when we prune the data. Otherwise, we have to review all streamlines operations in DIPY and make sure they align memory before starting any operations.

ppoulin91: There are a few specific examples where `._data` is accessed in a dangerous way (without checking if `is_view==True`):
- In dipy, e.g. `dipy.tracking.streamline.transform_streamlines()` with `in_place=True`, or `Dpy.write_tracks()`
- In nibabel, `Tractogram` can get an `ArraySequence` view when creating a new instance, but doesn't do any check when calling `apply_affine()` or `to_world()`.

In the end, both users and developers need to be very careful when playing with the internals of `ArraySequence`. I know I personally had trouble with it before I understood what was happening with the view concept.

dipy/io/stateful_tractogram.py
```diff
@@ -76,18 +75,20 @@ def __init__(self, streamlines, reference, space,
         if data_per_streamline is None:
             data_per_streamline = {}
 
+        if isinstance(streamlines, Streamlines):
+            streamlines = streamlines.copy()
```
MarcCote: Just to make sure we agree that this will be duplicating all the streamlines. Meaning when you create a `StateFullTractogram` from a `Streamlines` object, you might get twice the memory consumption!

frheault: Yes, I don't want anything outside modifying the streamlines. Even if it is very rare, the goal of the StatefulTractogram is to be super strict and robust. So to me this necessary to guarantee that the state is always valid and can only be change by the object.

For example, if somebody load a Tractogram using nibabel, create a StatefulTractogram and later call to_world() or save the Tractogram using nibabel the StatefulTractogram will be corrupted.

If someone has file with 10M streamlines, do multiprocessing and loads multiple big tractograms at once, and has 8GB of RAM, well they should use Nibabel and be careful.
I am personally on the side of safety and keeping the state valid over RAM limitation (even if it is a rare occurance)

dipy/io/stateful_tractogram.py
MarcCote: @skoudoro not sure about `MaskedArray`, the mask has to be the same size as the data, so it would always take ~twice the memory. :( 

@ppoulin91 yes I saw how it is used in Tractogram.apply_affine. That's clearly a bug. The transformation should only affect the "viewed" data (e.g. if we only want to apply a transformation on a specific cluster of streamlines) and not the entire referenced data! I'm trying to come up with an efficient fix for it, but I don't think I can :(. My only option is to loop through the "viewed" streamlines one by one and modify them in-place, or make a copy of the streamlines, first.

ppoulin91: @MarcCote Yeah, I don't think there is any other way apart from looping, because you never know if the view is contiguous in memory or not... :(

frheault:
@skoudoro @arokem Any updates ?

skoudoro:
+1 for merging! @arokem?

arokem:
Yep. Let's go!

PR #1963, opened 2019-09-02
DOC: Fix parameter docstring in LiFE tracking script

jhlegarreta:
Fix parameter docstring in LiFE tracking script.


arokem:
Good catch!

Issue #1895, opened 2019-07-11
Cython warnings


skoudoro:
## Description

Just a reminder. It will be good to track and clean all Cython warnings. (for release 1.0 or 1.1)

PR #1966, opened 2019-09-05
ENH: Speed up Cython, remove warnings

larsoner:
Gets rid of a slew of warnings of the form:
```
    warning: dipy/core/interpolation.pyx:749:25: Use boundscheck(False) for faster access
```
when building from source. Also should make things a tiny bit faster.

Alternatively this can be enabled using more fine-grained control via individual decorators like:
```
@cython.boundscheck(False)
```
From a quick look it seems like these are reasonable options for this file, but I did not look super closely. This is thus really more of a "suggestion PR" so feel free to close if it does not make sense.

### DIFF
--- a/dipy/core/interpolation.pyx
+++ b/dipy/core/interpolation.pyx
@@ -1,3 +1,5 @@
+# cython: boundscheck=False, wraparound=False, cdivision=True
+
 cimport cython
 cimport numpy as np
 cimport numpy as cnp


skoudoro:
I think it does make sense and thank you really much for that! 

I will still have a quick look at the `interpolation` module before merging. Any other opinion?

skoudoro:
Thanks @larsoner! merging

fix a part of #1895.

PR #1972, opened 2019-09-25
Small fix in warning message

arokem:
Since this uses bval.min() it's not the first, but rather the smallest b-value


skoudoro:
Make sense, Thanks!  merging

PR #1975, opened 2019-10-01
Azure CI

skoudoro:
As discussed in a recent lab meeting and via [DIPY mailing list](https://mail.python.org/archives/list/dipy@python.org/), I am looking into whether we'd like to switch over to Azure pipeline. For now, I've activated Azure pipeline for my own fork, and builds are appearing here: https://github.com/skoudoro/dipy/pull/3

Todo:
------

- [x] Add viz and XVFB
- [x] improve DEPENDS variable management
- [x] ~~Check optional build and activate USE-PRE~~ (will move to NEW PR)
- [x] ~~Check Doc generation~~ (will move to NEW PR)
- [x] ~~Check deploy doc~~ (will move to NEW PR)
- [x] check rebuild only one matrix if it failed for no reason ([see this comment](https://github.com/nipy/dipy/pull/1975#issuecomment-537530837))
- [x] check coverage
- [x] check Windows build
- [x] check OSX build
- [x] check Linux build
- [x] remove Nose dependency

supersede #785


skoudoro:
It does not seem possible to restart only failing matrix following [this post](https://developercommunity.visualstudio.com/content/problem/393053/restart-single-build-matrix-entry.html). It seems we will have to restart all the build if one failed. Maybe this feature is coming soon or did not catch it.

Any opinion on that @arokem, @Garyfallidis, @jhlegarreta

skoudoro:
Allow specific build to fail seems really hacky as you can see [here](https://github.com/microsoft/azure-pipelines-tasks/issues/9302), So I wonder what should I do with "USE_PRE" matrix. 

I will deactivate the Linux pipeline for the moment until we find a reasonable solution for these 2 problems.

If someone has any other options/workaround/ideas to fix this point, I will be happy to implement ans update this PR.

jhlegarreta:
Thanks for giving this a try @skoudoro.

Not sure whether all the following are accurate answers to the question, but if I understand correctly the issue and what [this section of the docs](https://docs.microsoft.com/en-us/azure/devops/release-notes/2018/sep-10-azure-devops-launch#manage-build-pipelines-using-the-new-builds-page) say, that would be possible?

I am not used to/haven't seen in other repositories the "new builds page" they talk about in that section, and I ignore how one can have it in one's builds. The builds page that you actually have in your fork seems the old (?) one, which is the one I am used to/I know @skoudoro.

Also, other people seem to [have asked about this](https://developercommunity.visualstudio.com/idea/365697/rerun-failed-build-taskstep.html) proposed [some workarounds](https://marketplace.visualstudio.com/items?itemName=BrunoMartinsPro.ping-pong-build-trigger-task).

IMHO, if this is really an issue for our builds, I would go with the hacking solution, open an issue in the repository and pin it so that whenever a solution becomes available from Azure DevOps we can remove the hack.

Also, slightly related to this may be the caching of the builds, i.e. if installing other packages or DIPY's `setup.py` step takes a long time and we want to speed the builds, may be caching could be used?
https://github.com/microsoft/azure-pipelines-yaml/pull/113

Not sure if this fits only for other `packages` using `DIPY` or, as said, whether it could be used to skip the `setup.py` step.

Not sure if that is the purpose of the `USE_PRE` flag that you are mentioning/using.

skoudoro:
As you can see [here](https://dev.azure.com/skab12/dipy/_build/results?buildId=118), builds are passing on Azure with Windows - Linux + OSX, with and without visualization using conda or pip.

I did not activate `USE_PRE` since I do not know yet how to allow builds to fail.

The only inconvenient part with Azure is you can not restart one matrix/job when it failed. You have to restart the whole PR.

Concerning the doc generation, I will create a new PR for that. Currently, this PR is big enough and some tutorials are currently failing.

So I will fix now Travis error but overall, this PR is ready to be reviewed and to go. We are ready to use Azure! Most of the PR will need a rebase when we merge this.

> The builds page that you actually have in your fork seems the old (?) one,

I checked and it seems to be the new one. This allows you to restart a PR/branch but not just a single matrix

> Also, other people seem to have asked about this proposed some workarounds.

These workarounds are too geeky for me and not enough reliable, It seems I will have to be patient that Azure fix that important feature.

jhlegarreta:
@skoudoro amazing job !

Nice and clean. Having them all in one page makes things easier IMHO.

The use cases I have seen so for just one configuration per platform (Linux, macOS, Windows), but do have separate pipelines for each, so I guess they can be started independently. In our case, having one pipeline per config is may be too lengthy, and would no longer have a single dashboard with the config matrix as you do now, but may allow to restart jobs independently. Not sure how this is managed from the point of view of the scripts/necessary files.

In any case, I'm seeing the configs be running now and they seem to run in parallel, right? I'd dare to say that for our needs, even if multiple PRs are queued, the waiting times would be that bad in terms of productivity. As long as the service is free for open-source projects, and until individual config triggers are offered natively (I agree that some workarounds may be geeky or unstable), I think it is fine.

Note that the current name for Apple's OS is `macOS` since a couple of years now.

jhlegarreta:
You checked the `coverage` checkbox: where can we see in the results of the corresponding build in Azure?

skoudoro:
> You checked the coverage checkbox: where can we see in the results of the corresponding build in Azure?

Strange and good catch, it disappears... Need to look at what is going wrong. it should appear on Linux-python3.7 + VIZ and the one above. I will add a tag on the title too.
Thanks

skoudoro:
ok, Done for coverage. Thanks @jhlegarreta for catching that.  `Conda`  builds seems to fail since yesterday because of this [issue](https://github.com/conda/conda/issues/9337).

Apart from that, it is ready to go!

jhlegarreta:
:thumbsup:  I see the coverage being reported, thanks @skoudoro ! Looks OK to me.

skoudoro:
- Conda issue Fixed with the proposed workaround.
- MacOS is now failing with python 3.5-3.6 because of `scs` package (https://github.com/cvxgrp/scs/issues/127).

Garyfallidis:
Thank you @skoudoro.  Looks good. Merging.

jhlegarreta:
May be the Azure CI badges should be added to the `README` file?

Issue #1976, opened 2019-10-02
Better warning when dipy is installed without fury


arokem:
## Description

If dipy is installed without fury, and the following line (from our examples) is run: 

`from dipy.viz import actor, window, ui`

the user will likely see an uninformative error: 

```
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-1-5eadb5980917> in <module>
----> 1 from dipy.viz import actor, window, ui

ImportError: cannot import name 'actor' from 'dipy.viz' (/Users/arokem/.virtualenvs/test_dipy_no_fury/lib/python3.7/site-packages/dipy/viz/__init__.py)`
```

This is because the current implementation of `optional_pkg` doesn't handle `from foo import bar` imports elegantly. 

I think that we should provide more useful information [here](https://github.com/nipy/dipy/blob/master/dipy/viz/__init__.py#L6-L10)

skoudoro:
Fixed by #1977 , closing

PR #1977, opened 2019-10-02
Add a warning on attempted import.

arokem:
Would this properly address #1976 ? 

The cryptic error is still raised, but is preceded by a warning: 
Screen Shot 2019-10-02 at 5 38 01 AM

### DIFF
--- a/dipy/viz/__init__.py
+++ b/dipy/viz/__init__.py
@@ -1,5 +1,6 @@
 # Init file for visualization package
 from __future__ import division, print_function, absolute_import
+import warnings
 
 from dipy.utils.optpkg import optional_package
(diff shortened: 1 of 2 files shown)


skoudoro:
Any chance to have a test for this PR @arokem?

arokem:
Sorry - yes - I have this for now. But it seems to be failing... 

I don't understand why the warning is not being captured by catch_warnings. It's definitely being emitted!

skoudoro:
Thanks! Maybe because of a [warning filter rules](https://docs.python.org/3/library/warnings.html). I do not know which one but I got this problem once.

skoudoro:
Thank you for this @arokem! Merging

PR #1983, opened 2019-10-14
[Fix] one example

skoudoro:
This PR fix `streamlines_registration.py` tutorial and some pep8. The documentation generation was failing because of this tutorial.


arokem:
Yeah.  Looks good!

PR #1984, opened 2019-10-17
Enabling TensorFlow 2 as an optional dependency

Garyfallidis:
Starting with a test on the basic dependency. 
And added simple perceptron in dipy.nn.model
The idea here build a model with fit as we do in reconstruction models. 
Use tensorflow.keras if you can to reduce size of code. Otherwise code the lower level of tensorflow API in a model such as SingleLayerPerceptron and provide fit, evaluate, predict etc.


skoudoro:
This PR seems ready to go! 

Is someone wants to have a look?  I will wait until this Friday to merge it.

skoudoro:
Ok, I'm going ahead and merge this PR. Thanks @Garyfallidis

Issue #1985, opened 2019-10-17
Something is wonky with the AFQ  tracts profile example


arokem:
This figure looks quite wrong: 

![image](https://user-images.githubusercontent.com/118582/67047177-a1404980-f0e6-11e9-8063-7d18c20f7fc9.png)

arokem:
I'm getting this when running locally on my laptop
![tract_profiles](https://user-images.githubusercontent.com/118582/67047245-cc2a9d80-f0e6-11e9-9481-981fb5a65c26.png)

Which seems about right.

skoudoro:
here what I'm getting on my macOS (python3.7). Different results from both of you...

![tract_profiles](https://user-images.githubusercontent.com/23106443/67048368-62d17b80-f134-11e9-8af8-2a44a211948c.png)

arokem:
Wow. That's a head-scratcher. I'm also on MacOS, with python 3.7. I have a hunch that it might have to do with the state of our examples directory when we start running this, because these results depend on tracking and then on bundle recognition, so there are a bunch of moving pieces upstream from this.

arokem:
Deleting my entire examples directory, getting a fresh checkout of all these files, and then rerunning resulted in the same figure I got before.

captainnova:
My figure looks sort of like skoudoro's for the CST, but more like the tutorial page's for the AF (it is a bit different, though): 
![tract_profiles](https://user-images.githubusercontent.com/15852777/67051816-d4440680-f101-11e9-8973-f75f3729bba8.png)

arokem:
Maybe we should fix the RNG seeds in bundle_extraction?

here:
https://github.com/nipy/dipy/blob/master/doc/examples/bundle_extraction.py#L74

and here:
https://github.com/nipy/dipy/blob/master/doc/examples/bundle_extraction.py#L123

It would be strange if the random element in that script resulted in such
large discrepancies, but it might be useful to check that.

Incoming with a PR.

arokem:
I fixed the random seeds in #1986. Still getting a figure very similar to the one shown above.  

Could you try fetching this down or similarly fixing the seeds and rerunning?

Please make sure to delete the files CST_L.trk, AF_L.trk and slr_transform.npy before running afq_tract_profiles.

skoudoro:
here, what I obtain with your branch (still macOS): 

[...]

PR #1986, opened 2019-10-17
Fix the random state for SLR and Recobundles in bundle_extraction.

arokem:
Attempts to address #1985


skoudoro:
Thank you for this update @arokem. Merging

PR #1989, opened 2019-10-24
DOC: Fix parameter docstring in `dipy.io.streamline.load_tractogram`

jhlegarreta:
Fix parameter docstring in `dipy.io.streamline.load_tractogram`: fix
name and type.


arokem:
LGTM! Merging

Issue #1828, opened 2019-05-07
IVIM VarPro implementation throws infeasible 'x0' 


ShreyasFadnavis:
## Description
This bug occurs rarely when the number of iterations for Differential Evolution is low.

Will follow-up with a PR with: 
- Default
- `Warning` when user reduces the `maxiter` below a certain value. 

#1826 will also help fix some of these issues.

## Way to reproduce
-- Cannot be reproduced with a fixed recipe.

ShreyasFadnavis:
@Garyfallidis @arokem Please feel free to let me know any other approach you have in mind. Another alternative I have is to switch to `LM` implementation is the solution is infeasible (may result in a linear fit though).  I think, after `Cythonizing` the implementation, this bug would go away anyways!

ShreyasFadnavis:
#1839 fixes this. Closing!

xxhdd:
@ShreyasFadnavis Hello, I'm trying out ivim varpro reconstruction using dipy, and getting error message " `x0` is infeasible." while not setting the bounds, and "Inconsistent shapes between bounds and `x0`." after setting bounds. the 'trr' method won't give this error. May I ask how should I solve this? Thanks!

ShreyasFadnavis:
> @ShreyasFadnavis Hello, I'm trying out ivim varpro reconstruction using dipy, and getting error message " `x0` is infeasible." while not setting the bounds, and "Inconsistent shapes between bounds and `x0`." after setting bounds. the 'trr' method won't give this error. May I ask how should I solve this? Thanks!

Hi @xxhdd ! Thank you for reaching out and reporting this. As for the bounds, there are certain things to keep in mind when using `VarPro` as it's a slightly different optimizer as compared to `trr`. Can I take a quick look at your code? You can mail it to me on shreyasfadnavis@gmail.com

On a side note, are you using a mask?

arokem:
@xxhdd : You can post the code here, so that people other than @ShreyasFadnavis can see it, and so that we don't have to wonder what happened a year down the line, when we're dealing with any other issues.

ShreyasFadnavis:
PR #1993 should fix this. The issue was that @xxhdd is working with animal data that is not the brain. the `x0` is not feasible error has been handled by reusing the same procedure used in `trr`.

skoudoro:
Hi @ShreyasFadnavis,

What is the status of this issue?  I do not find the related PR, Can you point it to me? 

Thank you

[...]

PR #1993, opened 2019-10-29
[ENH][BF] Error handling for IVIM VarPro NLLS 

ShreyasFadnavis:
- [x] Fixes the issue #1828

### DIFF
--- a/dipy/reconst/ivim.py
+++ b/dipy/reconst/ivim.py
@@ -499,7 +499,7 @@ class IvimModelTRR(ReconstModel):
             return ivim_params
         except ValueError:
             warningMsg = "x0 is unfeasible for leastsq fitting."
-            warningMsg += " Returning x0 values from the trr fit."
+            warningMsg += " Returning x0 values from the linear fit."
             warnings.warn(warningMsg, UserWarning)
             return x0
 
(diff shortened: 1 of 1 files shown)


ShreyasFadnavis:
@arokem can you take a look at this PR? I think its a minor fix to something that was missing in the VarPro implementation..

dipy/reconst/ivim.py
```diff
@@ -622,6 +619,49 @@ def fit(self, data, bounds_de=None):
...
+                                x_f,
+                                bounds=bounds,
+                                args=(data,))
+            ivim_params = res.x
+            if np.all(np.isnan(ivim_params)):
+                return np.array([-1, -1, -1, -1])
+            return ivim_params
+
+        except ValueError:
+            ivimmodel = IvimModel(gtab, fit_method='trr')
```
arokem: This quietly falls back to the `trr` method? I am not sure that's what we want here. Imagine a user comparing varpro to trr and getting exactly the same for both? Is there some way to instead throw a sufficiently informative error here?

ShreyasFadnavis: I understand what you mean, I could alternatively return the values from `x_f` obtained from `convex optimization + global optimization` of `VarPro` step too. 

But, this `x0` error typically occurs when the mask is not applied and is working in the background and not the data. 
So I feel that this should be okay.. Does that make sense?  

Nevertheless, I you feel that returning values from the previous step is better, I can do that too.. Just need to think of an appropriate way to do that.

ShreyasFadnavis: @Garyfallidis which one of the 2 alternatives do you suggest?

ShreyasFadnavis:
PR on wrong branch! Will create a new one! Sorry..

PR #1997, opened 2019-11-05
[FIX] Avoid copy of  Streamlines data

skoudoro:
As you can see on this PR (https://github.com/nipy/nibabel/pull/811), `ArraySequence.data` will now return a copy of the data. This way, altering the data of the sliced/indexed tractogram will not affect the original data.

So, the goal of this PR is to prevent this copy since the future release of Nibabel is quite soon.

### DIFF
--- a/dipy/io/stateful_tractogram.py
+++ b/dipy/io/stateful_tractogram.py
@@ -277,9 +277,9 @@ class StatefulTractogram(object):
         output : ndarray
             8 corners of the XYZ aligned box, all zeros if no streamlines
         """
-        if self._tractogram.streamlines.data.size > 0:
-            bbox_min = np.min(self._tractogram.streamlines.data, axis=0)
-            bbox_max = np.max(self._tractogram.streamlines.data, axis=0)
+        if self._tractogram.streamlines._data.size > 0:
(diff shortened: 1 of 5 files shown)


arokem:
Do you want to wait until nipy/nibabel#811 lands in nibabel before merging this. Looks like it doesn't matter, right?

skoudoro:
> Looks like it doesn't matter, right?

Yes, it does not really matter, I just anticipate the future Nibabel release (plan next week I think with nipy/nibabel#811 in)

skoudoro:
Any other opinions about this PR?

skoudoro:
rebased. let's see how `USE_PRE` is doing with Nibabel 3.0.0rc1

skoudoro:
Ok, everything green, it is ready to go!

skoudoro:
Any other comment for this PR?

arokem:
Looks like Appveyor is sick. But that's unrelated, right?

skoudoro:
Yes, I disconnected Appveyor but I do not why it is still triggered. I will add Azuce CI when the repo will move under DIPY org

arokem:
OK - I also just turned off the Appveyor builds through their control panel. Let's see if it stops building.

PR #1990, opened 2019-10-24
[WIP] Sphinx-Gallery integration

skoudoro:
The goal is to replace our example doc generation system by Sphinx-gallery.

TODO
====
- [ ] Update all examples syntax
- [ ] Change workflow tutorials, sphinx-gallery fails when I add them
- [ ] check design with our new website
- [ ] update doc/readme instruction
- [ ] update make for Linux and window
- [ ] Remove our system? Should we keep it or I can go ahead and delete it? 
- [ ] Check links and references

Below, the result of a quick test : 

image

image

PR #1998, opened 2019-11-07
[DOC] Workflows Tutorial

skoudoro:
This PR is the first basic tutorial about DIPY Workflows. More to come on a new PR...

### DIFF
--- a/doc/documentation.rst
+++ b/doc/documentation.rst
@@ -12,7 +12,7 @@ Contents:
    mission
    installation
    examples_index
-   pipelines/index
+   interfaces/index
    faq
    developers
    cite
--- a/doc/documentation.rst
+++ b/doc/documentation.rst
@@ -12,6 +12,7 @@ Contents:
    mission
    installation
    examples_index
+   pipelines/index
    faq
    developers
    cite
--- a/doc/interfaces/basic_flow.rst
+++ b/doc/interfaces/basic_flow.rst
@@ -1,8 +1,8 @@
 .. _basic_flow:
 
-=======================================
-Get Started with a basic DIPY pipelines
-=======================================
+====================================================
+Get Started with a basic DIPY Workflow command lines
+====================================================
 
 In a terminal, Let's get some data. This is multi-shell dataset which was
 kindly provided by Hansen and Jespersen (more details about the data are
--- a/doc/interfaces/index.rst
+++ b/doc/interfaces/index.rst
@@ -1,0 +1,9 @@
+.. _workflow_interfaces:
+
+=========================
+DIPY Workflows Interfaces
+=========================
+
+.. toctree::
+
+    basic_flow.rst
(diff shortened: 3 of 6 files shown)


arokem:
Small initial comment: I would reserve the word "pipeline" for future developments of more complex processing that automates the subsequent execution of several workflows together. In  other words, I would only use the word "workflow" here.

skoudoro:
Ok, Thank you for feedback @arokem!  I Will update that now.

Garyfallidis:
Looks good thanks.

doc/interfaces/basic_flow.rst
```diff
@@ -0,0 +1,42 @@
+.. _basic_flow:
+
+====================================================
+Get Started with a basic DIPY Workflow command lines
+====================================================
+
+In a terminal, Let's get some data. This is multi-shell dataset which was
```
arokem: "Let's" => "let's"

doc/interfaces/basic_flow.rst
```diff
@@ -0,0 +1,42 @@
...
+
+Create stopping criteria::
+
+    dipy_fit_dti dwi.nii.gz dwi.bval dwi.bvec out_work/brain_mask.nii.gz --out_dir out_work/
+
+Create peaks::
+
+    dipy_fit_csd dwi.nii.gz dwi.bval dwi.bvec out_work/brain_mask.nii.gz --out_dir out_work/
+
+Create seeding mask (this creates seed_mask in new folder)::
```
arokem: "in new folder" => "in a new folder"

doc/interfaces/basic_flow.rst
```diff
@@ -0,0 +1,42 @@
...
+
+    dipy_mask out_work/fa.nii.gz 0.4 --out_dir out_work/ --out_mask seed_mask.nii.gz
+
+To avoid that we ommit the --out_dir option::
```
arokem: "ommit" => "omit"

arokem: I would add backticks around '--out_dir'

doc/documentation.rst
```diff
@@ -12,6 +12,7 @@ Contents:
    mission
    installation
    examples_index
+   interfaces/index
```
arokem: "interfaces" is also not a great word here, in my opinion. Why not "workflows"?

arokem:
BTW - where does #1990 stand? This would presumably need to be integrated there.

skoudoro:
@Garyfallidis will update the typo and rephrase this basic tutorial on a new PR indeed. Thank you for all the corrections.

Concerning #1990, it will be easily integrated. I just have a small concern about the thumbnail that we can not generate. I do not think it is a good idea to run workflows script on real data to get a thumbnail (can be heavy a long computation). This part needs some thinking about the way we want to report workflow results.

PR #2000, opened 2019-11-07
NF: SplitFlow 

ShreyasFadnavis:
New workflow to split the input files from the command line to ease the creation of new interfaces such as registration. Currently aimed at extracting 3D volume from 4D data

### DIFF
--- a/bin/dipy_split
+++ b/bin/dipy_split
@@ -1,0 +1,10 @@
+#!python
+
+from __future__ import division, print_function
+
+from dipy.workflows.flow_runner import run_flow
+from dipy.workflows.io import SplitFlow
+
+
+if __name__ == "__main__":
+    run_flow(SplitFlow())
--- a/bin/dipy_split
+++ b/bin/dipy_split
@@ -1,7 +1,5 @@
 #!python
 
-from __future__ import division, print_function
-
 from dipy.workflows.flow_runner import run_flow
 from dipy.workflows.io import SplitFlow
 
--- a/dipy/workflows/io.py
+++ b/dipy/workflows/io.py
@@ -224,6 +224,9 @@ class SplitFlow(Workflow):
             (default 0)
         out_dir : string, optional
             Output directory. Default: dipy home folder (~/.dipy)
+        out_split : string, optional
+            Name of the resulting split volume (default: split.nii.gz)
+
         """
         io_it = self.get_io_iterator()
         for fpath, osplit in io_it:
--- a/dipy/workflows/io.py
+++ b/dipy/workflows/io.py
@@ -2,11 +2,13 @@ from __future__ import division, print_function, absolute_import
 
 import os
 import sys
+import shutil
+
 import numpy as np
 import logging
 import importlib
 from inspect import getmembers, isfunction, getfullargspec
-from dipy.io.image import load_nifti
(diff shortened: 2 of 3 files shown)


dipy/workflows/io.py
```diff
@@ -205,3 +207,39 @@ def run(self, data_names, out_dir=''):
...
             # wrong place
             self.load_module('dipy.data.fetcher')
+
+
+class SplitFlow(Workflow):
+    @classmethod
+    def get_short_name(cls):
+        return 'split'
+
+    def run(self, input_files, grad_dir=0, out_dir='',
```
arokem: I find `grad_dir` to be confusing, because I'd expect something like a 3D vector pointing in some direction, rather than an integer. How about `vol_idx` instead?

bin/dipy_split
```diff
@@ -0,0 +1,10 @@
+#!python
+
+from __future__ import division, print_function
```
skoudoro: To remove, since DIPY is compatible with python3 only

dipy/workflows/io.py
```diff
@@ -205,3 +207,39 @@ def run(self, data_names, out_dir=''):
...
+
+        Parameters
+        ----------
+        input_files : variable string
+            Any number of Nifti1 files
+        grad_dir : int, optional
+            (default 0)
+        out_dir : string, optional
+            Output directory. Default: dipy home folder (~/.dipy)
+        """
```
skoudoro: missing `out_split: string, optional`

dipy/workflows/io.py
```diff
@@ -205,3 +207,39 @@ def run(self, data_names, out_dir=''):
...
+        io_it = self.get_io_iterator()
+        for fpath, osplit in io_it:
+            if self._skip:
```
skoudoro: What is `self._skip` ?

dipy/workflows/io.py
```diff
@@ -205,3 +207,39 @@ def run(self, data_names, out_dir=''):
...
+                logging.info('Splitting {0}'.format(fpath))
+                data, affine, image = load_nifti(fpath, return_img=True)
+
+                if grad_dir == 0:
+                    logging.info('Splitting and extracting 1st b0')
+
+                split_vol = data[..., grad_dir]
+                save_nifti(osplit, split_vol, affine, image.header)
+
+                logging.info('Denoised volume saved as {0}'.format(osplit))
```
skoudoro: `denoised` ??

dipy/workflows/io.py
skoudoro: Agree, +1

Garyfallidis: +1 we can use the same script to split streamline files too.

ShreyasFadnavis:
@arokem @skoudoro @Garyfallidis Fixed! Thanks a lot :+1: 

#2000 !!

dipy/workflows/io.py
ShreyasFadnavis: done!

skoudoro:
Looks good to me, will merge it by the end of the day. Thank you @ShreyasFadnavis.

skoudoro:
Thanks @ShreyasFadnavis, merging !

PR #2002, opened 2019-11-11
BUG: Fix workflows IO `FetchFlow` `all` dataset fetching

jhlegarreta:
Fix workflows IO `FetchFlow` `all` dataset fetching: the `fetch_data` is
contained in the `available_data` dictionary because as the actual data fetcher
function it starts with  `fetch_`. However, the `fetch_data` function is what
the actual fetchers call with the necessary arguments. Thus, the
`dipy.workflows.tests.test_io.test_io_fetch` test fails if we specify
`fetch_flow.run(['all'])` because the parameters of `fetch_data` are empty.

The fix is to avoid adding the `fetch_data` function when the dictionary is
created.

Fixes
```
  File
"/src/dipy/dipy/workflows/tests/test_io.py",
line 80, in <module>
    test_io_fetch()
  File
"src/dipy/dipy/workflows/tests/test_io.py",
line 54, in test_io_fetch
    fetch_flow.run(['all'])
  File
"/src/dipy/dipy/workflows/io.py",
line 172, in run
    fetcher_function()
TypeError: fetch_data() missing 2 required positional arguments: 'files' and
'folder'
```

Add a test for the `all` case and compare to the expected number of dataset
folders (18).

### DIFF
--- a/dipy/workflows/tests/test_io.py
+++ b/dipy/workflows/tests/test_io.py
@@ -7,6 +7,7 @@ from dipy.testing import assert_true
 from dipy.data.fetcher import dipy_home
 from dipy.workflows.io import IoInfoFlow, FetchFlow, SplitFlow
 from nibabel.tmpdirs import TemporaryDirectory
+from os.path import join as pjoin
 from tempfile import mkstemp
 fname_log = mkstemp()[1]
 
@@ -50,6 +51,11 @@ def test_io_fetch():
         npt.assert_equal(os.path.isdir(os.path.join(out_dir,
                                                     'bundle_fa_hcp')),
                          True)
+
+        fetch_flow.run(['all'])
+        npt.assert_equal(
+            len(list(filter(lambda x: os.path.isdir(pjoin(dipy_home, x)),
+                            os.listdir(dipy_home)))), 18)
 
 
 def test_split_flow():
--- a/dipy/workflows/tests/test_io.py
+++ b/dipy/workflows/tests/test_io.py
@@ -52,10 +52,23 @@ def test_io_fetch():
                                                     'bundle_fa_hcp')),
                          True)
 
-        fetch_flow.run(['all'])
-        npt.assert_equal(
-            len(list(filter(lambda x: os.path.isdir(pjoin(dipy_home, x)),
-                            os.listdir(dipy_home)))), 18)
+
+def test_io_fetch_fetcher_datanames():
+    available_data = FetchFlow.get_fetcher_datanames()
+
+    dataset_names = ['bundle_atlas_hcp842', 'bundle_fa_hcp',
+                     'bundles_2_subjects', 'cenir_multib', 'cfin_multib',
+                     'file_formats', 'gold_standard_io', 'isbi2013_2shell',
+                     'ivim', 'mni_template', 'qtdMRI_test_retest_2subjects',
+                     'scil_b0', 'sherbrooke_3shell', 'stanford_hardi',
+                     'stanford_labels', 'stanford_pve_maps', 'stanford_t1',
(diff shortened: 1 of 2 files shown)


skoudoro:
Hi @jhlegarreta,

Good catch! Thank you for fixing that.

I think I introduce this issue during my last PR when I removed [this line](https://github.com/nipy/dipy/pull/1999/files#diff-a92d52b04539a319a163d0f6ce9beaeeL160) to get `multi_cfinb`

jhlegarreta:
You're welcome.

If adding back that line instead of the solution I propose is more robust and passes the test, go ahead. Not sure that is equivalent, though: a quick local test with that replacement:

```
available_data = dict([(name.replace('fetch_', ''), func)
                       for name, func in getmembers(fetcher_module,
                                                    isfunction)
                       if name.lower().startswith("fetch_") and
                       not len(getfullargspec(func).args)])
```

gives a different result in the `available_data` dictionary (the `cenir_multib` is missing). So that may require some other part.

Also, note that the number of elements of the dictionary (21 in the version I pushed; 20 in the above version) does not match the number of folders expected in the test (18): I guess this is because may be the `stanford_*` elements are stored in the same folder (have not checked).

jhlegarreta:
None of the builds succeeded, which is weird given that I started from an up-to-date `master`. A few of the builds failed because were unable to download the `bundle_fa_hcp` dataset:
```
Downloading "hcp_bundle_fa.nii.gz" to /home/travis/.dipy/bundle_fa_hcp
FAILED
```

In other case there is no failure notice. Weird.

@skoudoro can we re-trigger builds, please?

skoudoro:
Some datas are big, we will always encounter this `TimeOut` problem with this test, so we need to figure out another way to test this feature without downloading these datas.

> If adding back that line instead of the solution I propose is more robust and passes the test, go ahead

Totally fine with your solution. I just pointed the reason as an info.

jhlegarreta:
Thanks for the explanation @skoudoro . Then we are set to merge the patch set?

dipy/workflows/tests/test_io.py
```diff
@@ -51,6 +52,11 @@ def test_io_fetch():
                                                     'bundle_fa_hcp')),
                          True)
 
+        fetch_flow.run(['all'])
```
skoudoro: Just remove this block, I do not think we can test it without a `TimeOut` error

jhlegarreta: If we remove it, the error is likely to appear at some point in the future. I do acknowledged that I have no workaround for the `TimeOut` at the moment. But the culprit dataset is `bundle_fa_hcp`, which is fetched in the two tests before, so this is striking.

skoudoro: I do not think this is the only one "culprit" dataset, specially when you look at each CI's. Sometimes multi_cenir, other times  multi_cfin, etc...
In general, I do not think it is a good idea to download heavy data during unittest
Unfortunately, `fetch_all` can not be tested  (fetching part) but if you really want a workaround, you can create a static method in this class named get_fetcher_datanames and compare the results of this function what we expect. This function will replace the line 158.

jhlegarreta: That looks reasonable. @arokem @Garyfallidis any other suggestion?

arokem: Yeah. I think that's reasonable. If we figure out a way to continuously build/deploy the docs, the fetching operation itself will presumably get tested there.

jhlegarreta: @skoudoro @arokem Please have a look at commit 082c1c0.

skoudoro:
ok, no comments, merging! Thanks @jhlegarreta!

PR #1965, opened 2019-09-03
Horizon updates b1

Garyfallidis:
Testing mesa and xvfb issues

PR #2008, opened 2019-12-01
Fix origin nomenclature sft (center/corner)

frheault:
Fix the inversion in the documentation about the standard (trackvis vs nifti).
Replace the unclear variable name (`shifted_origin` -> `origin_at_corner`)
Added a small catch for all-zero affine that was crashing in `nib.aff2axcodes`

### DIFF
--- a/dipy/io/utils.py
+++ b/dipy/io/utils.py
@@ -193,9 +193,9 @@ def get_reference_info(reference):
         dimensions = header['dim'][1:4]
         voxel_sizes = header['pixdim'][1:4]
 
-        if not affine.any():
-            raise TypeError('Invalid affine, contains only zeros.'
-                            'Cannot determine voxel order from transformation')
+        if not affine[0:3, 0:3].any():
+            raise ValueError('Invalid affine, contains only zeros.'
+                             'Cannot determine voxel order from transformation')
         voxel_order = ''.join(nib.aff2axcodes(affine))
     elif is_trk:
         affine = header['voxel_to_rasmm']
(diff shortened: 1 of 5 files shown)


frheault:
@Garyfallidis I hope it does not impact your ongoing visualisation projects, the variable has the same behavior as before (a find&replace should do the trick).

dipy/io/utils.py
```diff
@@ -192,6 +192,10 @@ def get_reference_info(reference):
         affine[2, 0:4] = header['srow_z']
         dimensions = header['dim'][1:4]
         voxel_sizes = header['pixdim'][1:4]
+
+        if not affine.any():
```
arokem: Could you please add a test for this behavior?

skoudoro:
Thank you for this update @frheault. I just want to merge #1965 before your PR otherwise it looks good.

arokem:
The appveyor error is unrelated. I think this is ready to go -- any objections?

GuillaumeTh:
@arokem Code looks good. This PR could be merged now to allow @frheault to continue #2013. It could be nice to have #2013 in next release.

skoudoro:
No objection, you can go ahead @arokem

arokem:
Off we go. Thanks @frheault !

arokem:
We should really figure out how to turn off that pesky appveyor...

skoudoro:
agree

arokem:
I think maybe I did that now. For now, I just tried checking as many items in the settings not to build, but it would also be possible to delete the project from appveyor altogether. I am a bit reluctant to do that, because it would mean that we would lose all the history of builds.

skoudoro:
I tried just tried to remove the Appveyor webhook and service on Github and it seems that a new one is created... strange...

Issue #2003, opened 2019-11-15
Artefacts in DKI reconstruction


mdesco:
I wrote a personal email to @RafaelNH about problems I was having using DKI reconstruction on multi-shell data. I move the discussion here because it may help other dipy users.

> Hey Rafael,
> 
> I've turned your Dipy example into a script for our multi-shell data to extract kurtosis maps (ak, rk, mk). Unfortunately, I always have some very bad degenerate voxels, especially in voxels of very high anisotropy (CC, CST voxels). I get this no matter how I denoise the DWI (lpca, nlm, gaussian)... I was wondering if you had any insights on the problem?

Answer from @RafaelNH :

> Yes! I am quite familiarized with this kurtosis artefact! This is induced by the low radial diffusivities of aligned white matter (CC, CST voxels). Since it is very hard to capture non-Gaussian information due to the low decays in radial direction, its kurtosis estimates have very low robustness. Noisy kurtosis estimates tend to be negative and its absolute values can have order of magnitudes higher than the typical kurtosis values. Consequently, these negative kurtosis values will heavily propagate to the mean and radial kurtosis metrics. I've reported this artefact in my master thesis back in 2012 (https://repositorio.ul.pt/bitstream/10451/8511/1/ulfc104137_tm_Rafael_Henriques.pdf, see chapter 3).
> 
> One way that I've come up to overcome this issue is to compute the kurtosis values from powder-averaged diffusion-weighted signals. On powder-averaged signal decays, you don't have this low diffusivity issue and your kurtosis estimates have much higher precision (additionally they are independent to the fODF). To compute the kurtosis of powder-average signals in dipy follow the tutorial:
> 
> https://dipy.org/documentation/1.0.0./examples_built/reconst_msdki/#example-reconst-msdki
> 

This indeed helped and the MSDKI model is much more robust in these artefacted voxels.

Thanks!
(issue closed)

arokem:
Thanks so much for reporting here! Since this seems already resolved, I will go ahead and close it.

Garyfallidis:
Thanks for reporting @mdesco !

Garyfallidis:
I suggest this information also be added in the tutorial as advice @RafaelNH if not already there.

skoudoro:
I didn't check the tutorial but I think it would be good to add this information in the tutorial. So we should keep this issue open for informing the users. 

Thank you for reporting!

arokem:
Yes. Good call! Reopening.

[...]

PR #2009, opened 2019-12-02
Upgrade on DKI implementations

RafaelNH:
Hi all! Four years have passed since we've implemented the DKI module, so I've decided to do a review on its code and I've decided to give it an upgrade. 

Here are some features added:

1) Non-linear and Restore DKI fits - Basically I've made the DTI's non-linear and RESTORE fits compatible with DKI, so now dki.py can call these fitting procedures directly from dti.py.

2) Added some numerical solutions to the estimation of axial, radial and mean kurtosis. Theoretically, these should give the same results than the analytical solutions that we previously implemented. However, numerical solutions can be more stable for some voxels with parameters near to the analytical solution singularities.

3) Our previous implementations did not have a quantification of kurtosis anisotropy. Therefore, I've decided to add the Kurtosis Fractional Anisotropy index according to [Glenn et al. (2015)](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4378654/). Also added the mean kurtosis tensor as proposed by [Hansen & Jespersen (2013)](https://onlinelibrary.wiley.com/doi/full/10.1002/mrm.24743)

4) New metrics are already added in DKI example.


dipy/reconst/dti.py
```diff
@@ -1680,43 +1691,49 @@ def nlls_fit_tensor(design_matrix, data, weighting=None,
...
-                from_lower_triangular(this_tensor[:6]))
-            dti_params[vox, :3] = evals
-            dti_params[vox, 3:] = evecs.ravel()
+                from_lower_triangular(this_param[:6]))
+            params[vox, :3] = evals
+            params[vox, 3:12] = evecs.ravel()
             if return_S0_hat:
-                model_S0[vox] = np.exp(-this_tensor[6])
+                model_S0[vox] = np.exp(-this_param[-1])
+            if not dti:
```
arokem: This is basically just DKI, right? The `not dti` seems to suggest there are other models as well, which may be confusing.

arokem: Also, does this really need to be inside the `try` block? You seem to do this again below in the `except` block.

dipy/reconst/dti.py
```diff
@@ -1825,37 +1855,51 @@ def restore_fit_tensor(design_matrix, data, sigma=None, jac=True,
...
+
+        # Convert diffusion tensor parameters to the evals and the evecs:
+        try:
+            evals, evecs = decompose_tensor(
+                from_lower_triangular(this_param[:6]))
+            params[vox, :3] = evals
+            params[vox, 3:12] = evecs.ravel()
+            if return_S0_hat:
+                model_S0[vox] = np.exp(-this_param[-1])
+            if not dti:
```
arokem: Same comments as above.

dipy/reconst/tests/test_dki.py
```diff
@@ -425,14 +450,16 @@ def test_compare_MK_method():
...
+    MK_as = dkiF.mk(None, None, analytical=True)
 
     # MK numerical method
-    sph = Sphere(xyz=gtab.bvecs[gtab.bvals > 0])
-    MK_nm = np.mean(dki.apparent_kurtosis_coef(dkiF.model_params, sph),
-                    axis=-1)
+    MK_nm = dkiF.mk(None, None, analytical=False)
 
-    assert_array_almost_equal(MK_as, MK_nm, decimal=1)
+    # sph = Sphere(xyz=gtab.bvecs[gtab.bvals > 0])
```
arokem: What's this all about?

RafaelNH:
@arokem - answering your question. In theory, numerical and analytical solutions should give the exact same solution; however, numerical solution avoids some instabilities for voxels corresponding to the singularities of the analytical solution.

RafaelNH:
Example: If we run the simulations in MSDKI tutorial, you can see some discontinuities in MK curves for F=0.3. This are caused by instabilities of the estimates near the analytical solution singularities. 

Screen Shot 2019-12-03 at 10 37 05

Reruning the maps using MK numerical solution:

Screen Shot 2019-12-03 at 10 36 26

dipy/reconst/dki.py
```diff
@@ -1259,7 +1516,7 @@ def dki_prediction(dki_params, gtab, S0=1.):
         else:
             this_S0 = S0_vol
         X = np.concatenate((dt, fkt[v] * MD * MD,
-                            np.array([np.log(this_S0)])),
+                            np.array([-np.log(this_S0)])),
```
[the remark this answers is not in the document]
RafaelNH: @arokem! Don´t worry, this was not a bug! I had to change the sign here because I just changed the sign of DKI's design matrix so that we can use DTI's non-linear least squares fits.

dipy/reconst/dti.py
RafaelNH: I have in mind to use the same trick for other models based on cumulant expansion (e.g. correlation tensor imaging, future work), that is why I decided to call this variable ´dti´.

dipy/reconst/tests/test_dki.py
RafaelNH: Removing these commented lines of code - these lines were being used to evaluate MK analytical solution before I´ve implemented MK's numerical solution.

dipy/reconst/dti.py
RafaelNH: It was inside try blook because in expect blook another variable is called. Anyway, I will arrange a way to take it out to simplify code.

RafaelNH:
Just updated the documentation to address issue #2003 . 

All done at my side. Let me know if there is something else that you want me to address.

arokem:
Looks great! Could you please rebase this one more time? I think that you might have to make some adjustments to the changes you introduced in dipy.data and similarly to the examples to be consistent with changes in #2020 and #2021.

skoudoro:
I can manage this part and update this example on #2021 as soon as you merge this PR @arokem. No worries 😄

arokem:
Off we go then!

PR #2012, opened 2019-12-09
Add ability to analyze entire streamline in connectivity_matrix

Lawreros:
Hi, this is Ross Lawrence (a research assistant in Jovo's lab). This is my first dipy PR, so forgive any formatting issues. In this PR I added the option for connectivity_matrix to use the entire streamline in its analysis (instead of just the start and end points). This method is essentially what is used for our [ndmg ](https://github.com/neurodata/ndmg/tree/staging) connectome pipeline. This allows for connectivity matrices to use label volumes of white matter structures and record the connections between them.

- Works by looking at all the rois the streamline goes through and recording the connection between each roi. For example, if the streamlines path is through rois 3->2->4->5, then a connection will be recorded for (3,2), (3,4), (3,5), (2,4), (2,5), and (4,5).
- This is added functionality and works with the existing arguments (symmetric, return_mapping, etc.)
- Maintains directionality of streamlines unless symmetric=true

### DIFF
--- a/dipy/tracking/utils.py
+++ b/dipy/tracking/utils.py
@@ -170,29 +170,29 @@ def connectivity_matrix(streamlines, affine, label_volume, inclusive=False,
         #Create ndarray to store streamline connections
         edges = np.ndarray(shape=(3,0),dtype=int)
         lin_T, offset = _mapping_to_voxel(affine)
-        for s in range(len(streamlines)):
+        for sl in range(len(streamlines)):
             # Convert streamline to voxel coordinates
-            entire = _to_voxel_coordinates(streamlines[s], lin_T, offset)
+            entire = _to_voxel_coordinates(streamlines[sl], lin_T, offset)
             i, j, k = entire.T
             
             if symmetric:
                 # Create list of all labels streamline passes through
                 entirelabels = list(OrderedDict.fromkeys(label_volume[i,j,k]))
                 # Append all connection combinations along with streamline number
-                for e in combinations(entirelabels,2):
-                    edges=np.append(edges,[[e[0]],[e[1]],[s]], axis=1)
+                for comb in combinations(entirelabels,2):
+                    edges=np.append(edges,[[comb[0]],[comb[1]],[sl]], axis=1)
             else:
                 # Create list of all labels streamline passes through, preserving order
                 # and whether a label was entered multiple times
                 entirelabels = list(groupby(label_volume[i,j,k]))
                 # Append all connection combinations along with streamline number, removing
                 # duplicates and connections from a label to itself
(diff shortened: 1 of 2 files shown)


skoudoro:
That's Great! Thank you for doing this @Lawreros! 

We will review your PR during the day or tomorrow but as a first comment, I see that your modify the test which means there is a change of behavior. So we need to start a deprecation cycle and keep the old behavior for at least 1 or 2 releases. Let's see how we can integrate that smoothly 😃

dipy/tracking/utils.py
```diff
@@ -160,37 +165,87 @@ def connectivity_matrix(streamlines, affine, label_volume,
...
+            mapping = defaultdict(list)
+            for i, (a, b, c) in enumerate(edges.T):
+                mapping[a, b].append(c)
+            # Replace each list of indices with the streamlines they index
+            if mapping_as_streamlines:
+                for key in mapping:
+                    mapping[key] = [streamlines[i] for i in mapping[key]]
+
+            return matrix, mapping
+        else:
```
skoudoro: this last `else` is not needed. You can just return

dipy/tracking/utils.py
```diff
@@ -160,37 +165,87 @@ def connectivity_matrix(streamlines, affine, label_volume,
...
-            for key in mapping:
-                mapping[key] = [streamlines[i] for i in mapping[key]]
-
-        # Return the mapping matrix and the mapping
-        return matrix, mapping
+    if inclusive:
+        # Create ndarray to store streamline connections
+        edges = np.ndarray(shape=(3, 0), dtype=int)
+        lin_T, offset = _mapping_to_voxel(affine)
+        for sl in range(len(streamlines)):
```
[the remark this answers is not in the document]
Lawreros: I only really need the number, not the actual streamline. Added enumerate. https://github.com/nipy/dipy/pull/2012/commits/369eb44d5b416175dbccebf6f3c88f1d6ba25084

dipy/tracking/utils.py
Lawreros: Done. https://github.com/nipy/dipy/pull/2012/commits/369eb44d5b416175dbccebf6f3c88f1d6ba25084

skoudoro:
Thank you for the update @Lawreros!

Anyone for a quick look at this PR?  @gabknight, @guaje, @StongeEtienne, @arokem?

arokem:
This looks good to me: :+1: for the merge. @skoudoro : do you still want to give this another look/try?

skoudoro:
Great! Thank you for this work @Lawreros, merging!

PR #2013, opened 2019-12-11
SFT constructor from another SFT

frheault:
Instead of creating a `StatefulTractogram` from scratch providing the the reference, the space, the `voxel_at_origin` is cumbersome if the user wanted to have the exact same state as another SFT.
Colleagues were trying to do that and sometime forgot to pass along  `voxel_at_origin` messing up the new SFT state.

So to simplifying the operation (or at least provide a clear function for that), a constructor was added for that situation. Quite specific, but useful and clear functionality.

Added function to improve usability.
The reference for `StatefulTractogram` can now be (on top of the 6 other ways) a tuple similar to what `get_reference_info` was returning. The function `is_reference_info_valid` now check the basic type and value of the element to decrease chance of mistake.

This will have to be merged after #2008 , in time I will rebase and make sure the history is fine. Only the last 2 commits are truly for that PR (in case someone wanted to give comment, otherwise this can wait after the merge)

For focus `dipy.io.utils.is_reference_info_valid` and the static method `dipy.io.stateful_tractogram.from_sft` and `dipy.io.stateful_tractogram.change_origin` are the only one new. If anyone has suggestions for improvement.

_(Does not break anything, is not even a real new features, simple adjustment from observing colleagues and from our day to day usage.)_

### DIFF
--- a/dipy/io/stateful_tractogram.py
+++ b/dipy/io/stateful_tractogram.py
@@ -396,7 +396,7 @@ class StatefulTractogram(object):
             is_valid = False
 
         self.to_space(old_space)
-        self.change_origin(old_shift)
+        self.change_origin(old_origin)
 
         return is_valid
 
@@ -445,7 +445,7 @@ class StatefulTractogram(object):
                                       affine_to_rasmm=np.eye(4))
 
         self.to_space(old_space)
-        self.change_origin(old_shift)
+        self.change_origin(old_origin)
 
         return indices_to_remove, indices_to_keep
 
--- a/dipy/io/stateful_tractogram.py
+++ b/dipy/io/stateful_tractogram.py
@@ -10,6 +10,7 @@ from nibabel.streamlines.tractogram import (Tractogram,
                                             PerArrayDict)
 import numpy as np
 
+import dipy
 from dipy.io.dpy import Streamlines
 from dipy.io.utils import get_reference_info
 
@@ -96,6 +97,12 @@ class StatefulTractogram(object):
                                       data_per_point=data_per_point,
                                       data_per_streamline=data_per_streamline)
 
+        if isinstance(reference, dipy.io.stateful_tractogram.StatefulTractogram):
+            logging.warning('Using a StatefulTractogram as reference, this will' +
+                            ' copy only the space_attributes, not the state.' +
+                            ' Variables space shifted_origin must be specified' +
+                            ' separately.')
+
         space_attributes = get_reference_info(reference)
         if space_attributes is None:
             raise TypeError('Reference MUST be one of the following:\n' +
@@ -114,6 +121,36 @@ class StatefulTractogram(object):
             raise TypeError('origin_at_corner MUST be a boolean')
         self._origin_at_corner = origin_at_corner
         logger.debug(self)
+
+
+    @staticmethod
+    def from_sft(streamlines, sft,
+                 data_per_point=None,
+                 data_per_streamline=None):
+        """ Re-create a strict, state-aware, robust tractogram from another
+
+        Parameters
+        ----------
+        streamlines : list or ArraySequence
+            Streamlines of the tractogram
+        sft : StatefulTractgram,
+            The other StatefulTractgram to copy the space_attribute AND
+            state from.
+        data_per_point : dict
+            Dictionary in which each key has X items, each items has Y_i items
+            X being the number of streamlines
+            Y_i being the number of points on streamlines #i
+        data_per_streamline : dict
+            Dictionary in which each key has X items
+            X being the number of streamlines
+        -----
+        """
+        new_sft = StatefulTractogram(streamlines, sft.space_attributes, sft.space,
+                                     shifted_origin=sft.shifted_origin,
+                                     data_per_point=data_per_point,
+                                     data_per_streamline=data_per_streamline)
+        return new_sft
+
 
     def __str__(self):
         """ Generate the string for printing """
--- a/dipy/io/stateful_tractogram.py
+++ b/dipy/io/stateful_tractogram.py
@@ -36,6 +36,11 @@ class Space(enum.Enum):
     VOXMM = 'voxmm'
     RASMM = 'rasmm'
 
+class Origin(enum.Enum):
+    """ Enum to simplify future change to convention """
+    NIFTI = False
+    TRACKVIS = True
+
 
 class StatefulTractogram(object):
     """ Class for stateful representation of collections of streamlines
@@ -317,6 +322,18 @@ class StatefulTractogram(object):
             logger.error('Unsupported target space, please use Enum in '
                          'dipy.io.stateful_tractogram')
 
+    def change_origin(self, target_origin):
+        """ Safe function to change streamlines to a particular origin standard 
+        False means NIFTI (center) and True means TrackVis (corner) """
+        print(Origin.NIFTI)
+        if target_origin == Origin.NIFTI.value:
+            self.to_center()
+        elif target_origin == Origin.TRACKVIS.value:
+            self.to_corner()
+        else:
+            logger.error('Unsupported origin standard, please use Enum in '
+                         'dipy.io.stateful_tractogram')
+
     def to_center(self):
         """ Safe function to shift streamlines so the center of voxel is
         the origin """
@@ -378,13 +395,8 @@ class StatefulTractogram(object):
             logger.debug(bbox_corners)
             is_valid = False
 
-        if old_space == Space.RASMM:
-            self.to_rasmm()
-        elif old_space == Space.VOXMM:
-            self.to_voxmm()
-
-        if not old_origin:
-            self.to_center()
+        self.to_space(old_space)
+        self.change_origin(old_shift)
 
         return is_valid
 
@@ -432,13 +444,8 @@ class StatefulTractogram(object):
                                       data_per_streamline=tmp_data_per_streamline,
                                       affine_to_rasmm=np.eye(4))
 
-        if old_space == Space.RASMM:
-            self.to_rasmm()
-        elif old_space == Space.VOXMM:
-            self.to_voxmm()
-
-        if not old_origin:
-            self.to_center()
+        self.to_space(old_space)
+        self.change_origin(old_shift)
 
         return indices_to_remove, indices_to_keep
 
--- a/dipy/io/stateful_tractogram.py
+++ b/dipy/io/stateful_tractogram.py
@@ -10,7 +10,6 @@ from nibabel.streamlines.tractogram import (Tractogram,
                                             PerArrayDict)
 import numpy as np
 
-import dipy
 from dipy.io.dpy import Streamlines
 from dipy.io.utils import get_reference_info, is_reference_info_valid
 
@@ -103,7 +102,7 @@ class StatefulTractogram(object):
                                       data_per_point=data_per_point,
                                       data_per_streamline=data_per_streamline)
 
-        if isinstance(reference, dipy.io.stateful_tractogram.StatefulTractogram):
+        if isinstance(reference, type(self)):
             logging.warning('Using a StatefulTractogram as reference, this '
                             'will copy only the space_attributes, not '
                             'the state. The variables space and origin_at_corner '
--- a/dipy/io/stateful_tractogram.py
+++ b/dipy/io/stateful_tractogram.py
@@ -68,15 +68,15 @@ class StatefulTractogram(object):
             Current space in which the streamlines are (vox, voxmm or rasmm)
             Typically after tracking the space is VOX, after nibabel loading
             the space is RASMM
-        origin_at_corner : bool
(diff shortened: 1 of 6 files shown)


frheault:
@skoudoro After this one, I promise there will be no new features to the SFT.
It just that it was merge without beta testing for 1.0, so I learn on the spot what people needed.

ppoulin91:
This would be super useful, could it be part of the 1.1 release?

arokem:
Agreed. Would be great to have this in ASAP. @frheault : would you mind rebasing? My schedule today is a bit of a train-wreck, but I'll do my best to review this tonight.

skoudoro:
> @skoudoro After this one, I promise there will be no new features to the SFT.
It just that it was merge without beta testing for 1.0, so I learn on the spot what people needed.

No problem, By curiosity, does it copy the data/streamlines when you create a new one from an old one? I did not check it in depth yet

frheault:
@skoudoro from the SFT init if the streamlines are an ArraySequence they will be copied.

dipy/io/stateful_tractogram.py
```diff
@@ -36,6 +37,12 @@ class Space(enum.Enum):
     RASMM = 'rasmm'
 
 
+class Origin(enum.Enum):
+    """ Enum to simplify future change to convention """
+    NIFTI = False
+    TRACKVIS = True
+
```
skoudoro: `True`, `False`, What does it mean? Why there are not just integers? Does it imply we can not add any other origin type?

frheault: Well the Enum was there just in case someone forget which standard is which.
I could remove it altogether and continue with a direct use of true and false. But it seemed a bit more explicit to use this instead of boolean
But there is indeed only two type (which is why they it was a boolean value in the first place)

dipy/io/stateful_tractogram.py
```diff
@@ -10,8 +10,9 @@
                                             PerArrayDict)
 import numpy as np
 
+import dipy
```
arokem: Do you really need this import? I don't see this being used further down.

dipy/io/utils.py
```diff
@@ -136,6 +138,80 @@ def decfa_to_float(img_orig):
...
+    Parameters
+    ----------
+        affine: ndarray (4,4)
+            Tranformation of VOX to RASMM
+        dimensions: ndarray (3,), int16
+            Volume shape for each axis
+        voxel_sizes:  ndarray (3,), float32
+            Size of voxel for each axis
+        voxel_order: string
+            Typically 'RAS' or 'LPS'
```
arokem: Please de-indent the parameters block.

dipy/io/utils.py
```diff
@@ -136,6 +138,80 @@ def decfa_to_float(img_orig):
     return Nifti1Image(out_data, affine=img_orig.affine, header=new_hdr)
 
 
+def is_reference_info_valid(affine, dimensions, voxel_sizes, voxel_order):
+    """ Will validate basic data type and value of spatial attribute.
+    Does not ensure that voxel_sizes and voxel_order are self-coherent with
+    affine.
```
arokem: This is a bit confusing. Maybe best to list out what exactly this checks? Or provide a few examples in the docstring?

dipy/io/stateful_tractogram.py
```diff
@@ -96,11 +103,25 @@ def __init__(self, streamlines, reference, space,
...
+            logging.warning('Using a StatefulTractogram as reference, this '
+                            'will copy only the space_attributes, not '
+                            'the state. The variables space and origin_at_corner '
+                            'must be specified separately.')
+
+        if isinstance(reference, tuple) and len(reference) == 4:
+            if is_reference_info_valid(*reference):
+                space_attributes = reference
+            else:
+                raise TypeError('The provided space attributes are not '
```
arokem: I guess you can be a bit more specific here? If `is_reference_info_valid` were to return something more about why it's returning false?

dipy/io/stateful_tractogram.py
```diff
@@ -110,11 +131,45 @@ def __init__(self, streamlines, reference, space,
...
+            raise TypeError('origin_at_corner MUST be a boolean or from '
+                            'Origin enum, e.g Origin.NIFTI')
         self._origin_at_corner = origin_at_corner
         logger.debug(self)
 
+    @staticmethod
+    def from_sft(streamlines, sft,
+                 data_per_point=None,
+                 data_per_streamline=None):
+        """ Re-create a strict, state-aware, robust tractogram from another
```
arokem:
```suggestion
        """ Create an instance of `StatefulTractogram` from another instance of `StatefulTractogram`.
```

Unless I am missing something: what does "re-create" mean in this context?

dipy/io/stateful_tractogram.py
```diff
@@ -110,11 +131,45 @@ def __init__(self, streamlines, reference, space,
...
+
+        Parameters
+        ----------
+        streamlines : list or ArraySequence
+            Streamlines of the tractogram
+        sft : StatefulTractgram,
+            The other StatefulTractgram to copy the space_attribute AND
+            state from.
+        data_per_point : dict
```
arokem:
```suggestion
        data_per_point : dict, optional
```

Also: would be good to explain what the default behavior would be (presumably inherits from the source SFT?)

dipy/io/stateful_tractogram.py
```diff
@@ -110,11 +131,45 @@ def __init__(self, streamlines, reference, space,
...
+            Dictionary in which each key has X items, each items has Y_i items
+            X being the number of streamlines
+            Y_i being the number of points on streamlines #i
+        data_per_streamline : dict
```
arokem: Same comment here: specify that it's optional and what the default behavior is.

dipy/io/stateful_tractogram.py
arokem: Oh - I see. If it's not specified it's set to None. Of course.

dipy/io/tests/test_utils.py
```diff
@@ -51,6 +52,73 @@ def test_decfa():
...
+                                   'RAS')
+
+
+def is_voxel_order_valid(voxel_order):
+    return is_reference_info_valid(np.eye(4), [1, 1, 1], [1.0, 1.0, 1.0],
+                                   voxel_order)
+
+
+def test_reference_info_validity():
+    if is_affine_valid(np.eye(3)):
```
arokem: How about this instead?

```
assert not is_affine_valid(np.eye(3))
```

And so on?

dipy/io/utils.py
```diff
@@ -1,6 +1,8 @@
 """ Utility functions for file formats """
 import logging
+import numbers
 import os
+import six
```
arokem: Why use six? We no longer support Python 2. See also comment below.

dipy/io/utils.py
```diff
@@ -136,6 +138,80 @@ def decfa_to_float(img_orig):
...
+            logging.warning('Voxel size must be int/float.')
+        if i <= 0:
+            all_valid = False
+            logging.warning('Voxel size must be above 0.')
+
+    if not len(voxel_order) >= 3:
+        all_valid = False
+        only_3d_warning = True
+    for i in voxel_order:
+        if not isinstance(i, six.string_types):
```
arokem: In Python 3, I believe that this could be simply: 

```suggestion
        if not isinstance(i, str):
```

dipy/io/stateful_tractogram.py
frheault: Maybe I am limited by my knowledge of such a thing in python, but I needed for this condition:
`if isinstance(reference, dipy.io.stateful_tractogram.StatefulTractogram):`
Since it is within the class I could use a direct reference from within it own declaration and I could not import the class from the file declaring the class, so `import dipy` was needed just for that.

dipy/io/stateful_tractogram.py
frheault: The function does provide all the `logging.warning`, which would likely be displayed (unless the user turn off logging entirely). An user would see every failed verification before the crash as logging.

dipy/io/tests/test_utils.py
frheault: @skoudoro Would you agree with that ? If I remember correctly I had a ton of these in test_stateful_tractogram.py, but change them all for a reason I can't remember ?

Was it you or is it all in my head ?

dipy/io/stateful_tractogram.py
arokem: Unless I am missing something, I think that `isinstance(reference, type(self))` should work.

dipy/io/stateful_tractogram.py
```diff
@@ -96,11 +103,25 @@ def __init__(self, streamlines, reference, space,
                                       data_per_point=data_per_point,
                                       data_per_streamline=data_per_streamline)
 
-        space_attributes = get_reference_info(reference)
-        if space_attributes is None:
-            raise TypeError('Reference MUST be one of the following:\n' +
-                            'Nifti or Trk filename, Nifti1Image or TrkFile, ' +
-                            'Nifti1Header or trk.header (dict)')
+        if isinstance(reference, dipy.io.stateful_tractogram.StatefulTractogram):
```
arokem: So this would be:

```suggestion
        if isinstance(reference, type(self)):
```

dipy/io/stateful_tractogram.py
frheault: I feel like I should have known that, thx !

arokem:
OK - this looks good to me. @skoudoro : I believe the remaining questions are for you, so as far as I am concerned, when you are satisfied with this, you can go ahead and merge.

dipy/io/tests/test_utils.py
skoudoro: No, you are right. We said that we should avoid `assert` because they do not give an informative error message. So, in your case, you should just do

`np.assert_equal( is_affine_valid(np.eye(3)), False)`

dipy/io/stateful_tractogram.py
Garyfallidis: This enum is misleading @frheault. Please suggest alternative direction.

frheault: Well I could put `NIFTI = 'nifti'` and `TRACKVIS = 'trackvis'` or `NIFTI = 'center'` and `TRACKVIS = 'corner'`

But either way the goal remains the same, to have a value to check simply to switch between one state to the other. Which one do you prefer ?

frheault: Well I could put NIFTI = 'nifti' and TRACKVIS = 'trackvis' or NIFTI = 'center' and TRACKVIS = 'corner'

But either way the goal remains the same, to have a value to check simply to switch between one state to the other. Which one do you prefer ?

skoudoro: I understand your point @frheault, but I think `Enum` is not the correct structure for this boolean case. I try to find an alternative to propose to you.

frheault: I can remove it all together. The function can accept either `'center'` or `'corner'`. This was there just in case someone forgot which one was which one.

People can just read the docstring and see which are the acceptable option. I can also write a proper logging error so people know the two options if they fail to provide an acceptable string

skoudoro: I mean, `Enum` is ok, it is just the way you are using it. Just try to avoid the boolean `True` or `False`. it should work without this boolean.

frheault: @skoudoro what about using `NIFTI = 'center'` and `TRACKVIS = 'corner` ?

Would not that be what you are saying ?

skoudoro: Let's go for that, they are much more meaningful

dipy/io/tests/test_utils.py
arokem: @skoudoro :  I believe you meant `npt.assert_equal`? Probably even better to use `npt.assert_`?

    npt.assert_(not is_affine_valid(np.eye(3)))

If you need an informative message, you could also do: 
    
    assert not is_affine_valid(np.eye(3)), "This affine is not valid"

But, IMO, `npt.assert_` is the best. WDYT @skoudoro ?

skoudoro: > I believe you meant npt.assert_equal?

yes

> But, IMO, npt.assert_ is the best...

Agree, and still with a message if possible 😄

frheault:
@skoudoro @arokem I change the ENUM et modified the code, docstring and test accordingly !
And I replace all the AssertionError to npt.assert_ with a (simple) message for each

arokem:
Excellent! 👍 from for the merge. @skoudoro : as soon as you're ready.

skoudoro:
I confirm, perfect! 👍  Thank you really much @frheault. I will merge it as soon as Travis is Happy.

arokem:
Looks all green. Merging.

PR #2014, opened 2019-12-16
Removes the appveyor configuration file.

arokem:
Following up on the conversation on https://github.com/nipy/dipy/pull/2008#issuecomment-566191117, I want to check whether appveyor will build this branch, considering it doesn't have an appveyor config file.


arokem:
Blergh. Looks like appveyor is trying to build this. Don't have time now, but I'll give it another shot late.

skoudoro:
if you look at #2013, it seems to be fixed.  I just deactivate more magic triggers from our Github settings. So I close this PR. Thanks @arokem!

arokem:
That's great! Shouldn't we remove the appveyor config file, though?

skoudoro:
As you said, let's keep it just in case we want to come back to Appveyor and match/reproduce our build history.

arokem:
OK!

PR #2015, opened 2019-12-23
Update nibabel minimum version (3.0.0)

### LINKED ISSUE #2006: get_data deprecated
Nibabel deprecate `get_data()`. We should start slowly to replace `get_data()` by either

- `get_fdata()` from Nibabel
- `load_nifti`.  this is our small interface

I suggest updating our tutorials with `load_nifti`. I have no strong opinion concerning the rest of our codebase.

skoudoro:
The new Nibabel release introduces important changes that we should take into consideration (nipy/nibabel#833, nipy/nibabel#844  and nipy/nibabel#811). Furthermore, Nibabel 3.0.0 is python3 only, which is coherent with DIPY 1.0.0+. This is why I think, this PR should in before the next release.

This implies to update all our deprecated functions `get_data` to `get_fdata`.

Fix #2006 

PS: Since there is already a lot of change, I split this PR in 2 so  I will update our doc examples on another PR.


arokem:
+1. LGTM! 

I think this is OK, but it sets a pretty strict threshold for users. If (for some reason), someone wants/needs to use an earlier nibabel, they are forever stuck in DIPY 1.0 or earlier.  So, let's give others a chance to chime in here.

skoudoro:
Completely agree with you and sorry for the ones who needs an earlier Nibabel :sweat_smile:

Garyfallidis:
On my side this is fine to merge. Serge will follow up with an additional PR for docs and some strategy for follow up for backwards compatibility. One thing that will help a lot is calling nibabel from a central place in DIPY which we are already doing in most places.

arokem:
Cool. That should make it easier for users to adjust. Merging.

PR #2016, opened 2019-12-31
replace get_fdata by load_nifti_data

skoudoro:
Introduce a small interface for `get_fdata` function. The function name is more explicit and makes easier the maintenance.


Garyfallidis:
okay thanks

arokem:
Is there some way to special-case (perhaps a kwarg) the cases that require direct access to the `dataobj` attribute? Does that make sense?

PR #2017, opened 2020-01-02
Update CONTRIBUTING.md

arokem:
I don't think we need this section anymore.


skoudoro:
Great! Can you remove this line too : https://github.com/nipy/dipy/blob/master/dipy/__init__.py#L36

skoudoro:
Thanks!

PR #2020, opened 2020-01-02
Update get_fnames

### LINKED ISSUE #1777: Maybe `read_*` should return full paths?
Currently, a function like `read_stanford_hardi` will return an image object and a gradient table object. That makes it easy for us to use when we write examples, but as I've learned this week, it makes it hard for users when they try to adapt the examples to their own data, because file i/o is the first thing they'll need to do and it's not obvious how to go about it. 

What if we had `read_stanford_hardi` return file-names/full paths instead of returning the img and gtab and then we explicitly did the file i/o in the example? Maybe it would make using the examples easier?

skoudoro:
The goal of this PR is to provide a more explicit way to handle our I/O in the examples..

I will create a follow-up PR to update our examples today or tomorrow. Our examples will become more transparent for our users.

This PR should fix #1777.


dipy/data/__init__.py
```diff
@@ -204,7 +197,7 @@ def get_sphere(name='symmetric362'):
 
 
 def get_fnames(name='small_64D'):
-    """ provides filenames of some test datasets or other useful parametrisations
+    """Provide filenames of some datasets or other useful parametrisations.
```
arokem:
```suggestion
    """Provides full paths to example datasets.
```

skoudoro:
> The one worry I have is that this is one more place we need to remember to update when we add new fetchers...

Agree, We should minimize our maintenance work. I porpose 2 steps:

For now, I will move this function in `fetcher.py` and make sure that all `read_*` functions call `get_fnames`. 

Then, we can talk about it during the workshop and redesign the loading/fetching code (Release 1.3)

What do you think?

dipy/data/fetcher.py
```diff
@@ -593,143 +760,121 @@ def read_qtdMRI_test_retest_2subjects():
...
 
-    return nib.load(file)
+    """
+    fnames = get_fnames('scil_b0')
+    return nib.load(fnames[0])
 
 
 def read_siemens_scil_b0():
-    """ Load Siemens 1.5T b0 image form the scil b0 dataset.
+    """Load Siemens 1.5T b0 image form the scil b0 dataset.
```
arokem:
```suggestion
    """Load Siemens 1.5T b0 image from the scil b0 dataset.
```

arokem: This is just a small typo unrelated to the PR.

Garyfallidis:
Looks good. Merging.

PR #2021, opened 2020-01-03
Examples update 

skoudoro:
Follow-up of #2016 and #2020. All examples will use `get_fnames`, `load_nifti` and `save_nifti` to handle the I/O. This is more explicit for the basic user.

PR #2022, opened 2020-01-03
Update pep8speaks configuration

arokem:
Ignore the examples folder, because many of these files will not comply with PEP8. For example, see all the messages that popped up in #2021


skoudoro:
👍 , a good idea. Thanks, merging

PR #2025, opened 2020-01-03
Adds an align.api module, which provides simplified API to registration functions

arokem:
As discussed in our last call, I am adding here some functions that provide a simplified functional API to registration functionality. 

Questions: 

1. Useful? 
2. Should we add/change examples to use this API? 
3. Should this be incorporated into use in the respective workflows? Would that simplify things there?


arokem:
No need to look at this right now! This is definitely not for 1.1. I just added the 1.2 label -- I think we can wrap this up by then.

dipy/align/__init__.py
```diff
@@ -1,15 +1,18 @@
+from .api import *  # noqa
```
skoudoro: We should avoid `import *` and just import the desired function

arokem: That's why the module has an `__all__` attribute.

arokem:
OK - I have made another pretty thorough pass through here, adding documentation, making sure tests work (at least locally) and adding a couple of documentation examples (as part of the existing affine registration example). 

Any more thoughts about how to name this module? If not "api", what should we call this?

dipy/align/__init__.py
arokem: OK OK. Changed in 274248fa4

arokem:
OK - more thoughts about naming this: I think that it would be neat to have an `api.py` module within each of our top-level modules: `reconst`, `segment`, `denoise`, etc.. This sub-module would expose this kind of functional API. Then, in each module `__init__.py` we would import the names from `api` into the module name-space. That means, that, like here, users would never write something like`import dipy.align.api`, but would instead import things from the top-level name-space. For example, `from dipy.align import affine_registration`. The benefit of having it all in one place, clearly marked as 'api' is that it removes the ambiguity of  "where do I put this?".  If you are writing a functional interface that uses objects and other names from a specific module, this would go into the relevant modules `api.py`. That said, I am not particularly attached to 'api' and happy to hear suggestions for other names that this sub-module should take, so long as this is the same name across the different top-level modules.

skoudoro:
Completely agree with your thought @arokem and still thinking about another name.

Just as a reminder, all functions in these "api" modules should NOT do any file management. This is the Workflows role.

arokem:
> Just as a reminder, all functions in these "api" modules should NOT do
any file management. This is the Workflows role.

In principle, I agree, but it was really easy to make these functions
accept arrays, nibabel image objects, or full paths :-)

dipy/align/api.py
```diff
@@ -0,0 +1,854 @@
...
+from dipy.tracking.utils import transform_tracking_output
+from dipy.io.streamline import load_trk
+
+
+syn_metric_dict = {'CC': CCMetric,
+                   'EM': EMMetric,
+                   'SSD': SSDMetric}
+
+
+def _input_as_img_arr_or_path(data, affine=None):
```
skoudoro: I do not think this function should be in this module. it is really useful and can be used everywhere. 

At the same time, I feel this is the role of the workflows to do and we should not start to mix it and be rigorous with this point.

So I have this double feeling and we need to make a decision about this point.

dipy/align/api.py
```diff
@@ -0,0 +1,854 @@
...
+    mapping = sdr.optimize(static, moving,
+                           static_grid2world=static_affine,
+                           moving_grid2world=moving_affine,
+                           prealign=prealign)
+
+    warped_moving = mapping.transform(moving)
+    return warped_moving, mapping
+
+
+def dwi_to_template(dwi, gtab, dwi_affine=None, template=None,
```
skoudoro: This function name should be more explicit. you can add `register` in front of this function name

dipy/align/api.py
```diff
@@ -0,0 +1,854 @@
...
+                                starting_affine=between_affine)
+    affine_map = AffineMap(between_affine,
+                           static.shape, static_affine,
+                           moving.shape, moving_affine)
+    resampled = affine_map.transform(moving)
+    return nib.Nifti1Image(resampled, static_affine)
+
+
+# Affine registration pipeline:
+affine_metric_dict = {'MI': MutualInformationMetric}
```
skoudoro: Can you bring this variable up with `syn_metric_dict`? It is a bit hidden here

dipy/align/api.py
arokem: I'd be happy to move it elsewhere. In `dipy.io` somewhere?

skoudoro: good idea, `dipy.io.utils` seems a good place!

dipy/align/__init__.py
```diff
@@ -21,7 +23,8 @@ def __init__(self, **kwds):
...
 
-
-
-
-
+from dipy.align._public import (syn_registration, register_dwi_to_template,
+                                write_mapping, read_mapping, resample,
+                                c_of_mass, translation, rigid, affine,
+                                affine_registration, register_series,
+                                register_dwi_series, streamline_registration)
```
skoudoro, line 30: It would be good if you could add `__all__ = ['syn_registration', 'register_dwi_to_template', etc..]`

dipy/align/_public.py
```diff
@@ -0,0 +1,833 @@
...
+    static_affine, moving_affine: 4D arrays
+
+    starting_affine : in case this is needed.
+    """
+    static, static_affine = read_img_arr_or_path(static,
+                                                      affine=static_affine)
+    moving, moving_affine = read_img_arr_or_path(moving,
+                                                      affine=moving_affine)
+    if starting_affine is None:
+        starting_affine = np.eye(4)
```
skoudoro, line 61: `starting_affine = starting_affine or np.eye(4)` : The `if` is  not needed and this expression is always evaluated

dipy/align/_public.py
```diff
@@ -0,0 +1,833 @@
...
+    Helper function to prepare inputs for pipeline functions
+
+    Parameters
+    ----------
+    moving, static: Either as a 3D/4D array or as a nifti image object, or as
+        a string containing the full path to a nifti file.
+
+    static_affine, moving_affine: 4D arrays
+
+    starting_affine : in case this is needed.
```
skoudoro, line 54: Strange documentation 😉

dipy/align/_public.py
```diff
@@ -0,0 +1,833 @@
...
+        The data in `moving`, warped towards the `static` data.
+    forward : ndarray (..., 3)
+        The vector field describing the forward warping from the source to the
+        target.
+    backward : ndarray (..., 3)
+        The vector field describing the backward warping from the target to the
+        source.
+    """
+    if level_iters is None:
+        level_iters = [10, 10, 5]
```
skoudoro, line 112: `level_iters = level_iters or [10, 10, 5]`, same comment as above

dipy/align/_public.py
```diff
@@ -0,0 +1,833 @@
...
+
+    static, static_affine, moving, moving_affine, _ = \
+        _handle_pipeline_inputs(moving, static,
+                                moving_affine=moving_affine,
+                                static_affine=static_affine,
+                                starting_affine=None)
+
+    use_metric = syn_metric_dict[metric](dim, sigma_diff=sigma_diff,
```
skoudoro, line 120: `metric.upper()` I just made this mistake 😄

dipy/align/_public.py
```diff
@@ -0,0 +1,833 @@
...
+                                         radius=radius)
```
skoudoro, line 121: Here, the EM metric will crash since the parameters are different. There are 3 different initialization:
- `CCMetric(dim, sigma_diff=2.0, radius=4)`
- `EMMetric(dim, smooth=1.0, inner_iter=5, q_levels=256, double_gradient=True, step_type='gauss_newton')`
- `SSDMetric(dim, smooth=4, inner_iter=10, step_type='demons')`

dipy/align/_public.py
```diff
@@ -0,0 +1,833 @@
...
+    Notes
+    -----
+    This function assumes that the DWI data is already internally registered.
+    See :func:`register_dwi_series`.
+
+    """
+    dwi_data, dwi_affine = read_img_arr_or_path(dwi, affine=dwi_affine)
+
+    if template is None:
+        template = dpd.read_mni_template()
```
skoudoro, line 183: same as above

dipy/align/_public.py
```diff
@@ -0,0 +1,833 @@
...
+
+    template_data, template_affine = read_img_arr_or_path(
+                                       template,
+                                       affine=template_affine)
+    if isinstance(dwi, str):
+        dwi = nib.load(dwi)
```
skoudoro, line 189: Can you use DIPY function: `load_nifti`

dipy/align/_public.py
```diff
@@ -0,0 +1,833 @@
...
+    if reg_method == "syn":
+        warped_b0, mapping = syn_registration(mean_b0, template_data,
+                                              moving_affine=dwi_affine,
+                                              static_affine=template_affine,
+                                              **reg_kwargs)
+    elif reg_method == "aff":
+        warped_b0, mapping = affine_registration(mean_b0, template_data,
+                                                 moving_affine=dwi_affine,
+                                                 static_affine=template_affine,
+                                                 **reg_kwargs)
```
skoudoro, line 204: `else: raise an error` To avoid any strange behavior, just in case...

dipy/align/_public.py
```diff
@@ -0,0 +1,833 @@
...
+                                       template,
+                                       affine=template_affine)
+    if isinstance(dwi, str):
+        dwi = nib.load(dwi)
+
+    if not isinstance(gtab, dpg.GradientTable):
+        gtab = dpg.gradient_table(*gtab)
+
+    mean_b0 = np.mean(dwi_data[..., gtab.b0s_mask], -1)
+    if reg_method == "syn":
```
skoudoro, line 195: `reg_method.lower()` to avoid `Syn`

dipy/align/_public.py
```diff
@@ -0,0 +1,833 @@
...
+        Full path to the nifti file storing the mapping
+
+    Notes
+    -----
+    The data in the file is organized with shape (X, Y, Z, 2, 3, 3), such
+    that the forward mapping in each voxel is in `data[i, j, k, 0, :, :]` and
+    the backward mapping in each voxel is in `data[i, j, k, 0, :, :]`.
+    """
+    mapping_data = np.array([mapping.forward.T, mapping.backward.T]).T
+    nib.save(nib.Nifti1Image(mapping_data, mapping.codomain_world2grid), fname)
```
skoudoro, line 225: Could you use `save_nifti`? Thank you

dipy/align/_public.py
```diff
@@ -0,0 +1,833 @@
...
+    See :func:`write_mapping` for the data format expected.
+    """
+    if isinstance(disp, str):
+        disp = nib.load(disp)
+
+    if isinstance(domain_img, str):
+        domain_img = nib.load(domain_img)
+
+    if isinstance(codomain_img, str):
+        codomain_img = nib.load(codomain_img)
```
skoudoro, line 257: same comment as above for the load/save

dipy/align/_public.py
```diff
@@ -0,0 +1,833 @@
...
+
+    Notes
+    -----
+    See :func:`write_mapping` for the data format expected.
+    """
+    if isinstance(disp, str):
+        disp = nib.load(disp)
+
+    if isinstance(domain_img, str):
+        domain_img = nib.load(domain_img)
```
skoudoro, line 254: same comment as above for the load/save

dipy/align/_public.py
```diff
@@ -0,0 +1,833 @@
...
+    Returns
+    -------
+    A :class:`DiffeomorphicMap` object.
+
+    Notes
+    -----
+    See :func:`write_mapping` for the data format expected.
+    """
+    if isinstance(disp, str):
+        disp = nib.load(disp)
```
skoudoro, line 251: same comment as above for the load/save

dipy/align/_public.py
```diff
@@ -0,0 +1,833 @@
...
+                                static_affine=static_affine,
+                                starting_affine=between_affine)
+    affine_map = AffineMap(between_affine,
+                           static.shape, static_affine,
+                           moving.shape, moving_affine)
+    resampled = affine_map.transform(moving)
+    return nib.Nifti1Image(resampled, static_affine)
+
+
+def c_of_mass(moving, static, static_affine=None, moving_affine=None,
```
skoudoro, line 321: `center_of_mass` instead

dipy/align/_public.py
```diff
@@ -0,0 +1,833 @@
...
+
+    static_affine : 4x4 array, optional
+        An affine transformation associated with the static object. Required if
+        data is provided as an array. If provided together with nifti/path,
+        will over-ride the affine that is in the nifti.
+
+    starting_affine: 4x4 array, optional
+        Initial guess for the transformation between the spaces.
+
+    reg : not needed here. Use None
```
skoudoro, line 349: So why do we keep this unused parameter?

dipy/align/_public.py
```diff
@@ -0,0 +1,833 @@
...
+    degrees of freedom adjusted.
+    """
+    if pipeline is None:
+        pipeline = [c_of_mass, translation, rigid, affine]
+    if level_iters is None:
+        level_iters = [10000, 1000, 100]
+    if sigmas is None:
+        sigmas = [3, 1, 0.0]
+    if factors is None:
+        factors = [4, 2, 1]
```
skoudoro, line 613: same comment as above for all this block: `factors = factors or [4, 2, 1]`

dipy/align/_public.py
```diff
@@ -0,0 +1,833 @@
...
+    static, static_affine, moving, moving_affine, starting_affine = \
+        _handle_pipeline_inputs(moving, static,
+                                moving_affine=moving_affine,
+                                static_affine=static_affine,
+                                starting_affine=starting_affine)
+
+    # Define the Affine registration object we'll use with the chosen metric:
+    use_metric = affine_metric_dict[metric](
+                        nbins=nbins,
+                        sampling_proportion=sampling_proportion)
```
skoudoro, line 624: Currently, it is working because we have only one metric. This needs more thinking in case we have other kind of metric. I do not know yet how to manage it.

dipy/align/_public.py
```diff
@@ -0,0 +1,833 @@
...
+        together with the nifti img or file.
+
+    Returns
+    -------
+    xformed, affines : 4D array with transformed data and a (4,4,n) array
+    with 4x4 matrices associated with each of the volumes of the input moving
+    data that was used to transform it into register with the static data.
+    """
+    if pipeline is None:
+        [c_of_mass, translation, rigid, affine]
```
skoudoro, line 680: same as above

dipy/align/_public.py
```diff
@@ -0,0 +1,833 @@
...
+
+    series, series_affine = read_img_arr_or_path(series,
+                                                      affine=series_affine)
+    if isinstance(ref, numbers.Number):
```
skoudoro, line 684: nice! I did not know this 👏

skoudoro:
Hi @arokem,

Any chance to rebase and address my comment above? Thank you.

arokem:
I just rebased, but no chance to address any comments before mid May. I
have been slammed with other things that can't wait and my bandwidth is
severely narrow, because of no child care. Sorry.

skoudoro:
ok, thank you for the rebase and your feedback!

dipy/align/api.py
arokem: Done in 25f24dc8b49f0fbd2405

dipy/align/__init__.py
arokem: Done in 78d835550

dipy/align/_public.py
arokem: But sometimes `starting_affine` has a value that we would like to pass on as output.

dipy/align/_public.py
arokem: Resolved in 7b393b296

dipy/align/_public.py
arokem: Oh  gotcha - you just want a oneliner here. OK!

dipy/align/_public.py
arokem: Done in 173b2d25c

dipy/align/_public.py
arokem: Done in 173b2d25c

dipy/align/_public.py
arokem: Done in 5ee908ffc

dipy/align/_public.py
arokem: So that we can use the same call to different pipeline components.

dipy/align/_public.py
arokem: Done in 2cad7da1

dipy/align/_public.py
arokem: Maybe we can refactor when that becomes relevant? Should I add some kind of comment here?

dipy/align/_public.py
arokem: I just also noticed that this was not even being assigned 🤣 

Done in 833e508c2

dipy/align/_public.py
arokem: Yep - the standard lib is awesome.

arokem:
OK - almost there - I think. 

I still need to add a test on the  io/utils function. Hopefully later today.

dipy/align/_public.py
arokem: Looks like this doesn't work with arrays. See: https://travis-ci.org/github/dipy/dipy/jobs/685933762#L4454

arokem:
I believe this is ready for another review. CI failures look unrelated.

dipy/align/_public.py
skoudoro: Fair enough, I'm fine and +1 for a small comment

dipy/align/_public.py
```diff
@@ -0,0 +1,830 @@
...
+    """
+    level_iters = level_iters or [10, 10, 5]
+
+    static, static_affine, moving, moving_affine, _ = \
+        _handle_pipeline_inputs(moving, static,
+                                moving_affine=moving_affine,
+                                static_affine=static_affine,
+                                starting_affine=None)
+
+    use_metric = syn_metric_dict[metric.upper()](dim, sigma_diff=sigma_diff,
```
skoudoro, line 122: Here, the EM metric will crash since the parameters are different. There are 3 different initialization:
- `CCMetric(dim, sigma_diff=2.0, radius=4)`
- `EMMetric(dim, smooth=1.0, inner_iter=5, q_levels=256, double_gradient=True, step_type='gauss_newton')`
- `SSDMetric(dim, smooth=4, inner_iter=10, step_type='demons')`

dipy/align/_public.py
arokem: The problem here (and below) is that we actually don't need the data read from file here. Just the shape and affine, so I think that for these two it would make more sense to use nibabel directly.

dipy/align/_public.py
arokem: See above comment.

arokem:
OK - thanks for taking another look (and sorry for missing those comments!). I think that I've now addressed these.

arokem:
I just rebased this. I think everything _should be_ good, but we kept hitting a time-out downloading some data.

skoudoro:
Hi @arokem, 

What is the status of this PR? I think this first version should go in, it looks good to me. Then, we can always iterate to improve it. 

I just saw that there are a few issues not addressed. Can you fix them? Also, can you rebase?

Let me know what you think and if you have time to make it happens. Thanks!

dipy/align/_public.py
arokem: Addressed below.

dipy/align/_public.py
arokem: Is this still an issue?

dipy/align/_public.py
arokem: Same here.

arokem:
Sorry: I don't see unaddressed comments, but I might be missing something. 

What still remains to do here?

skoudoro:
Sorry, I was talking about the pep8 comments [here](https://github.com/dipy/dipy/pull/2025#issuecomment-570732755). 

Our comments are addressed indeed so this PR can go in.

arokem:
Whoops! Now rebased and with fixed flake8.

skoudoro:
Hi @arokem,

Do you have any idea of what could generate [this issue on Azure](https://dev.azure.com/dipy/dipy/_build/results?buildId=612&view=logs&j=06cb7910-729e-54cb-52d8-a8328392b8cd&t=34ad7fbf-e7dd-5c5b-e7c8-56ecb0d4306e&l=14387)?

arokem:
No. I have no idea. Looks like the main difference between this and other machine is that it uses conda? I am not sure why that would affect this test so severely. Just to see if this is a glitch of some sort, I'll try to retrigger this run.

arokem:
Hmm. This failure seems unrelated: https://dev.azure.com/dipy/dipy/_build/results?buildId=694&view=logs&jobId=9c5ef928-2cd6-52e5-dbe6-9d173a7d951b&j=9c5ef928-2cd6-52e5-dbe6-9d173a7d951b&t=1fd94682-a265-5c56-2b0e-ca2ae59556ed

Let's see if the other failures are resolved with https://github.com/dipy/dipy/pull/2025/commits/f5b713f3aa28ee331d438507b723dce8ac4b4dda

skoudoro:
Yes, those macOS failures are due to Azure pipeline side. See [here](https://developercommunity.visualstudio.com/content/problem/975083/pip-installation-problem-on-hosted-mac-agent-appea.html).

I am hesitating to disable them until they fix it.

arokem:
Looks like all the failures are now in OSX installs and tests pass on other platforms. Is this ready to go?

skoudoro:
This first version looks good to me. It is ready to go. In new PRs, We need to update the workflows and add the same `_public` API to other modules.  

Thank you for this work. Let's way until tomorrow to see if there are any additional comments.  If not, we can merge it.

skoudoro:
ok, no other comment, so merging! Thank you @arokem!

PR #2027, opened 2020-01-06
QuickBundles: fix test on order invariance property

koendc:
This allows order variant metrics to work properly

### DIFF
--- a/dipy/segment/clusteringspeed.pyx
+++ b/dipy/segment/clusteringspeed.pyx
@@ -582,7 +582,7 @@ cdef class QuickBundles(object):
         nearest_cluster = self.find_nearest_cluster(self.features)
 
         # Find nearest cluster to s_i_flip if metric is not order invariant
-        if not self.metric.feature.is_order_invariant:
+        if self.metric.feature.is_order_invariant != 0:
             self.metric.feature.c_extract(datum[::-1], self.features_flip)
             nearest_cluster_flip = self.find_nearest_cluster(self.features_flip)
 
--- a/dipy/segment/tests/test_quickbundles.py
+++ b/dipy/segment/tests/test_quickbundles.py
@@ -194,6 +194,19 @@ def test_quickbundles_with_not_order_invariant_metric():
     assert_array_equal(clusters[0].centroid, streamline)
 
 
+def test_quickbundles_with_not_order_invariant_metric_multiple_clusters():
+    metric = dipymetric.AveragePointwiseEuclideanMetric()
+    qb = QuickBundles(threshold=0.1, metric=metric)
+
+    streamline = np.arange(10*3, dtype=dtype).reshape((-1, 3))
+    streamlines = [streamline, streamline[::-1]]
+
+    clusters = qb.cluster(streamlines)
+    assert_equal(len(clusters), 2)
+    assert_array_equal(clusters[0].centroid, streamlines[0])
+    assert_array_equal(clusters[1].centroid, streamlines[1])
+
+
 def test_quickbundles_memory_leaks():
     qb = QuickBundles(threshold=2*threshold)
 


skoudoro:
Thank you for this @koendc!  

Can you add a small test to reproduce this problem or point it on our current test? Thank you!

dipy/segment/clusteringspeed.pyx
```diff
@@ -582,7 +582,7 @@ cdef class QuickBundles(object):
         nearest_cluster = self.find_nearest_cluster(self.features)
 
         # Find nearest cluster to s_i_flip if metric is not order invariant
-        if not self.metric.feature.is_order_invariant:
+        if self.metric.feature.is_order_invariant != 0:
```
skoudoro: I think the test was good before when I look at the docstring just above. Can you explain to us what is wrong?

koendc: This code in the if-block is not executed in case of a non-order_invariant metric like AveragePointwiseEuclideanMetric.

I have added a test that hopefully proves the point :)

skoudoro:
Can you have a look at this PR @MarcCote? this is about `order_invriant` and `quickbundles`.

I would love your opinion 😄!!! Others' opinions are welcomed too!

dipy/segment/clusteringspeed.pyx
MarcCote: @koendc your change is inverting the test condition, i.e. only compute the flip distance if the metric **is** order-invariant. This is not what it should be doing, we want to compute the flip when it **is not** order-invariant. Otherwise, it is just wasting computation as the results between flip and non-flip should be the same.

dipy/segment/tests/test_quickbundles.py
```diff
@@ -194,6 +194,19 @@ def test_quickbundles_with_not_order_invariant_metric():
...
 
+def test_quickbundles_with_not_order_invariant_metric_multiple_clusters():
+    metric = dipymetric.AveragePointwiseEuclideanMetric()
+    qb = QuickBundles(threshold=0.1, metric=metric)
+
+    streamline = np.arange(10*3, dtype=dtype).reshape((-1, 3))
+    streamlines = [streamline, streamline[::-1]]
+
+    clusters = qb.cluster(streamlines)
+    assert_equal(len(clusters), 2)
```
MarcCote: It is not clear to me why you expect two clusters here. Both streamlines are the same, they just start from different ends. If you truly want to differentiate these two streamlines, you don't want to use QuickBundles because QB relies on the flip to merge streamlines correctly (i.e. to form the centroids).

Let me know if I didn't understand what you are trying to accomplish.

skoudoro:
Thank you for the feedback @MarcCote!

By curiosity, how do you setup this variable `is_order_invariant`? is it only by creating a features class?
something like (missing some mandatory function): 

```
class MyFeature(Feature):
    def __init__(self):
        super(MyFeature, self).__init__(is_order_invariant=True)

metric = dipymetric.AveragePointwiseEuclideanMetric(feature=MyFeature())
qb = QuickBundles(threshold=0.1, metric=metric)
```

MarcCote:
Yes, you are correct.

In QuickBundles, we are dealing with clustering of sequences and to build the centroid we need to merge/aggregate data points (i.e. features). Since the data points are sequences their ordering is very important, otherwise you'll end up with centroid that doesn't make sense. For instance, merging `[1, 2, 3]` and `[3, 2, 1]` would result in `[2, 2, 2]` instead of `[1, 2, 3]` or `[3, 2, 1]` (which are equivalent).

I think one shortcoming of the current implementation of QuickBundles is we only support one aggregation method (i.e. point-wise averaging).

While on the subject, a common pitfall with this implementation is using QuickBundles with [MinimumAverageDirectFlipMetric](https://github.com/nipy/dipy/blob/master/dipy/segment/metricspeed.pyx#L289) will not work as expected. This is because MDF is order-invariant  (i.e. the flip is done within the metric calculation), thus QuickBundles won't do the flip check thus resulting in erroneous centroids (i.e. point-wise averaging will use the original ordering of the sequences).

skoudoro:
Thank you for this nice explanation @MarcCote! make sense!

I will close this PR but before that, Do you have any other question @koendc? What is your specific problem? How can we help you?

koendc:
Thank you @MarcCote and @skoudoro to handle this. You can close this PR as it is pointless indeed.

For me, the original ordering of the sequences is important. This can be achieved by simply setting is_order_invariant on the feature to True:

```
feature = ResampleFeature(nb_points=resample_points)
feature.is_order_invariant = True
metric = AveragePointwiseEuclideanMetric(feature=feature)
```

Garyfallidis:
@koendc for which kind of data are you using QB?

koendc:
To cluster GPS data point sequences - basically following the idea described here: https://medium.com/isiway-tech/gps-trajectories-clustering-in-python-2f5874204a53

It works remarkably well. The centroids match the roads even though the signal can be quite noisy.

Garyfallidis:
This is wonderful to hear. QuickBundles for GPS tracking! Thank you guys for sharing, reporting possible issues and of course writing the blog.

PR #2029, opened 2020-01-07
[MAINT] Use the array proxy instead of get_fdata()

skoudoro:
In Nibabel, `get_data` was using the array proxy by default which seems to not be the case for `get_fdata`. So now, we use `np.asarray(img.dataobj)` to keep the default behavior.

this is a PR for the maintenance branch, I will have to cut a new release which will be DIPY 1.1.1 and announce this one by default instead of DIPY 1.1.0.

### DIFF
--- a/dipy/io/image.py
+++ b/dipy/io/image.py
@@ -3,15 +3,15 @@ import nibabel as nib
 import numpy as np
 
 
-def load_nifti_data(fname):
+def load_nifti_data(fname, as_ndarray=True):
     img = nib.load(fname)
-    return np.asarray(img.dataobj)
+    return np.asanyarray(img.dataobj) if as_ndarray else img.dataobj
 
 
 def load_nifti(fname, return_img=False, return_voxsize=False,
-               return_coords=False):
+               return_coords=False, as_ndarray=True):
     img = nib.load(fname)
-    data = np.asarray(img.dataobj)
+    data = np.asanyarray(img.dataobj) if as_ndarray else img.dataobj
     vox_size = img.header.get_zooms()[:3]
 
     ret_val = [data, img.affine]
--- a/dipy/io/image.py
+++ b/dipy/io/image.py
@@ -3,15 +3,15 @@ import nibabel as nib
 import numpy as np
 
 
-def load_nifti_data(fname, dtype=np.float64):
+def load_nifti_data(fname):
     img = nib.load(fname)
-    return img.get_fdata(dtype=dtype)
+    return np.asarray(img.dataobj)
 
 
 def load_nifti(fname, return_img=False, return_voxsize=False,
-               return_coords=False, dtype=np.float64):
+               return_coords=False):
     img = nib.load(fname)
-    data = img.get_fdata(dtype=dtype)
+    data = np.asarray(img.dataobj)
     vox_size = img.header.get_zooms()[:3]
 
     ret_val = [data, img.affine]
--- a/dipy/tracking/tests/test_life.py
+++ b/dipy/tracking/tests/test_life.py
@@ -158,7 +158,7 @@ def test_fit_data():
     fstreamlines = dpd.get_fnames('small_25_streamlines')
     gtab = grad.gradient_table(fbval, fbvec)
     ni_data = nib.load(fdata)
-    data = ni_data.get_fdata()
+    data = np.asarray(ni_data.dataobj)
 
     tensor_streamlines = nib.streamlines.load(fstreamlines).streamlines
     sft = StatefulTractogram(tensor_streamlines, ni_data, Space.RASMM)
--- a/dipy/viz/app.py
+++ b/dipy/viz/app.py
@@ -205,7 +205,7 @@ class Horizon(object):
             if self.random_colors:
                 colors = next(color_gen)
             else:
-                colors = "RGB"
+                colors = None
 
             if not self.world_coords:
                 # TODO we need to read the affine of a tractogram
--- a/dipy/viz/app.py
+++ b/dipy/viz/app.py
@@ -205,7 +205,7 @@ class Horizon(object):
             if self.random_colors:
                 colors = next(color_gen)
             else:
-                colors = None
+                colors = "RGB"
 
             if not self.world_coords:
                 # TODO we need to read the affine of a tractogram
--- a/doc/api_changes.rst
+++ b/doc/api_changes.rst
@@ -5,7 +5,7 @@ API changes
 Here we provide information about functions or classes that have been removed,
 renamed or are deprecated (not recommended) during different release circles.
 
-DIPY 1.1.0 changes
+DIPY 1.1.x changes
 ----------------
 
 **Tractogram**
--- a/doc/examples/denoise_gibbs.py
+++ b/doc/examples/denoise_gibbs.py
@@ -155,7 +155,7 @@ bvals = [200, 400, 1000, 2000]
 
 img, gtab = read_cenir_multib(bvals)
 
-data = img.get_fdata()
+data = np.asarray(img.dataobj)
 
 """
 For illustration proposes, we select two slices of this dataset
--- a/doc/examples/reconst_forecast.py
+++ b/doc/examples/reconst_forecast.py
@@ -10,6 +10,7 @@ axially symmetric tensor and the fODF using the FORECAST model from
 First import the necessary modules:
 """
 
+import numpy as np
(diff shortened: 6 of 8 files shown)


matthew-brett:
I'm not sure what you mean by `get_fdata` not using the array proxy - could you clarify?

If it were me - I'd use `get_fdata` - for the reason we switched from `get_data` in the first place - the data type is nastily difficult to predict from `get_data` / `np.array(img.dataobj)` - and that can have unpleasant consequences that are only apparent when you try stuff with different data, that has made different, also arbitrary choices for the header scaling.

skoudoro:
> I'm not sure what you mean by get_fdata not using the array proxy - could you clarify?

it seems that `get_fdata()` keeps the array in its internal cache by default which, I suppose, was not the case before with `get_data`. After this change, we started to have memory issue with big datasets

> If it were me - I'd use get_fdata - for the reason we switched from get_data in the first place

Do you mean you recommend us to use `get_fdata(caching='unchanged')` instead when we load our data? 

Thank you for your feedback @matthew-brett!

matthew-brett:
No - `get_fdata` and `get_data` both cache their arrays.  I'm sure the difference you're observing is that `get_fdata` loads as float64 by default, whereas `get_data` loads with God-knows-what datatype.   Hence the problem - it was easy for an apparently trivial change to the image (change in scale factors in the header) to switch an image from coming back as int16 data to float64 or the other way round.  For example students would quickly hit integer overlflow errors if they got the wrong image, or linear algebra would slow to a halt (because it's not well optimized for - say - int16).

I think your choices are - guarantee that DiPy will always work in the same way with any returned data type (int8, int16, float64, ...) - which I suspect will be a lot of work - or load the data as an explicit data type - perhaps as float32 - so you can predict the subsequent behavior.

Garyfallidis:
Hi @matthew-brett and thank you for helping with this. This is not what we are observing. The issue is not with data types. The data before using get_fdata were staying in a memory map and not loaded directly in memory. Maybe this is different than caching. For example, we care a lot for doing something like vol = arr[..., 1] where arr still stays on disk. Before the change we could trivially visualize brain data of 15GBytes but not anymore. That is why we go back to the previous strategy. Please do suggest alternatives.

matthew-brett:
Yes, that's awkward.  You only get a memory map from `np.array(img.dataobj)` when the output data type is the same as the data type on disk.    So, if the header had scaling, you would not have got a memory map.

Do your images all have float data type?   Or are you also dealing with int images (with no scaling)?

Garyfallidis:
For sure we have images that do not use floats and for sure we have images with no scaling we often have zooms of (1., 1., 1) (which I suppose this is what you mean no scaling). It would be nice to have a way to always get data as memory maps (or similar tech) I think the data today are becoming just way too big to load everything at once. At least for DWIs. Possibly for fMRI should be of a similar problem too. We can send you some example data if you want to play with.

Garyfallidis:
On a side note this fix is necessary and currently blocking the release. Any suggestions? @effigies, @matthew-brett? Is it okay if we go back to the dataobj solution?

matthew-brett:
No - by scaling I mean - having a scalefactor in the header that is not 1.  This signals that the values in the image should be mutiplied by that value after loading from disk.

To understand more - have a look at https://github.com/nipy/nibabel/wiki/BIAP8 . 

We can't easily make the returned array a memory map, if the image has scalefactors.  The problem is that it's essentially arbitrary whether the image does have scalefactors, so the previous solution, which was just to ignore that and hope for the best, is putting the user at risk of wrong and unexpected results, that you can easily miss using a selected set of test images.

effigies:
> scaling we often have zooms of (1., 1., 1) (which I suppose this is what you mean no scaling)

No, the scaling refers to the scale factors `scl_slope`, `scl_inter`. If these are not `1` and `0`, respectively, then `get_data()` will cast to float64, the same as `get_fdata()`. `get_fdata()` has the advantage that you an also specifically request `float32` with `img.get_fdata(dtype=np.float32)`.

> It would be nice to have a way to always get data as memory maps

You can, if your image is uncompressed and you use `img.dataobj.get_unscaled()`. Note, however, that you will get incorrect values if your header indicates you have non-identity scale factors.

> Is it okay if we go back to the dataobj solution?

It will restore `get_data()` behavior, yes. `np.asanyarray(img.dataobj)` guarantees that the values will be correct (scaled if the header indicates they should be), but nothing about type. `get_fdata()` allows you to be sure about the type.

----

One thing I haven't seen mentioned is that, if you're willing to work with the `dataobj` directly, you can slice it:

```
img.dataobj[..., 1]
```

This is equivalent to `img.get_data()[..., 1]`, as the scaling is applied during the call, but only the necessary slices of data are read into memory. This also has the advantage that it works with compressed images, and if you have [indexed_gzip](https://pypi.org/project/indexed-gzip/) installed, it can be done with very little penalty for discontiguous slices.

----

For the best long-term solution, a relevant question is what minimum nibabel you're willing to set. nibabel 3.0.0 has some improvements that seem like they might dramatically improve your use-case.

The biggest one from the top of my head is that, `img.get_fdata(dtype=np.float32)` will return a `memmmap` if the on-disk data type is `float32`, and will not cast up to `float64` internally, if scale factors need application (nibabel <3 did...).

Garyfallidis:
Thank you @matthew-brett  and @effigies this is all very useful information. Here is a link to a sample  dataset that we are using to test nibabel 
https://figshare.com/articles/Superresolved_HYDI_dataset/10266194 you will see there a 15GB .nii dataset that you try.
Right now only the np.asanyarray(img.dataobj) gives something that is memory efficient for most cases. Otherwise we cannot load these data in our machines. 

So for now to move ahead with our release, the only viable solution that we see is to use get_fdata when data is float32 and use np.asanyarray(img.dataobj) for all other cases. 

You may want to create a flag at some point that always memmaps the data in all cases even if they are coming from gz files. If it already exists and I do not know about it - my apologies in advance.

arokem:
On Wed, Jan 8, 2020 at 4:09 PM Eleftherios Garyfallidis <

> Thank you @matthew-brett <https://github.com/matthew-brett> and @effigies
> <https://github.com/effigies> this is all very useful information. Here
> is a link to a sample dataset that we are using to test nibabel
> https://figshare.com/articles/Superresolved_HYDI_dataset/10266194 you
> will see there a 15GB .nii dataset that you try.
> Right now only the np.asanyarray(img.dataobj) gives something that is
> memory efficient for most cases. Otherwise we cannot load these data in our
> machines.
>
> So for now to move ahead with our release, the only viable solution that
> we see is to use get_fdata when data is float32 and use
> np.asanyarray(img.dataobj) for all other cases.
>
> I am not sure that I understand this solution. Would you require that
users always specify the dtype of the data to read it from disk?

Another (probably annoying to implement?) solution is maybe to read the
data as memmap and apply the scaling whenever the data is loaded into
memory.

You may want to create a flag at some point that always memmaps the data in

matthew-brett:
Another (probably annoying to implement?) solution is maybe to read the
> data as memmap and apply the scaling whenever the data is loaded into
> memory.
>

If I understand you right, that's what Chris was pointing to - when you
slice into the dataobj.

img.dataobj[..., 1]

That's the best solution - and should work reasonably for .gz files as
well, as Chris points out - is there a reason you can't use that?

effigies:
Right, I think using the `dataobj` directly as much as possible is probably the goal to aim at, so delaying the `np.asanyarray()` until you really need a numpy array would be the right strategy. A lot of operations will treat `dataobj` just like a numpy array, and if you can slice it before doing anything that needs an array, you're going to reduce I/O and memory usage.

This might be a tough thing to do before a hot-fix release, so to restore the old behavior, you definitely can use `np.asanyarray(dataobj)`. If the caching is important, nilearn did add a [hack](https://github.com/nilearn/nilearn/blob/fcc22e65a2d7f07e95d470f1c3d41f269b0e6fb2/nilearn/_utils/niimg.py#L17-L25) that you could copy. I would strongly recommend against making that a long-term solution, though. There are memory profile issues that will likely not get fixed in the `get_data()` path: https://github.com/nilearn/nilearn/issues/2209

One thing to think about is building up `ArrayProxy` to support more of the `ndarray` protocol while keeping things on disk. Right now, for instance, we can reshape an `ArrayProxy`, but another thing that jumps to mind is having `dataobj.astype(dtype)` return a new `ArrayProxy` that will change its default target dtype when slicing. That should probably be a new discussion, though...

PR #2030, opened 2020-01-10
DOC: Documents I/O functions.

arokem:
I noticed on #2029 that these functions are not documented. This adds documentation in `master`. You may also want to apply a similar change in `1.1.x`

### DIFF
--- a/dipy/io/image.py
+++ b/dipy/io/image.py
@@ -3,13 +3,53 @@ import nibabel as nib
 import numpy as np
 
 
-def load_nifti_data(fname, as_ndarray=True):
+def load_nifti_data(fname, dtype=np.float64):
+    """
+    Loads only the data array from a nifti file.
+
+    Paramters
+    ---------
+    fname : str
+        Full path to the file.
+    dtype : numpy dtype object, optional
+        What dtype to read the data as. Default: np.float64.
+
+    Returns
+    -------
+    np.ndarray
+
+    See also
+    --------
+    ``load_nifti``
+    """
     img = nib.load(fname)
     return np.asanyarray(img.dataobj) if as_ndarray else img.dataobj
 
 
 def load_nifti(fname, return_img=False, return_voxsize=False,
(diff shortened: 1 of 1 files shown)


skoudoro:
can you do this pull request for branch `maint/1.1.x` instead of master? then I merge both PR on master after the release.

skoudoro:
ps: my previous PR was on this branch (`maint/1.1.x`) and not `master`

arokem:
Rebased!

dipy/io/image.py
```diff
@@ -3,13 +3,57 @@
 import numpy as np
 
 
-def load_nifti_data(fname, as_ndarray=True):
+def load_nifti_data(fname, dtype=np.float64):
```
skoudoro: Travis will fail here because of `as_ndarray`. This variable is needed, it was the reason for the conflict.

dipy/io/image.py
```diff
@@ -3,13 +3,57 @@
...
+    --------
+    ``load_nifti``
+    """
     img = nib.load(fname)
     return np.asanyarray(img.dataobj) if as_ndarray else img.dataobj
 
 
 def load_nifti(fname, return_img=False, return_voxsize=False,
-               return_coords=False, as_ndarray=True):
+               return_coords=False, dtype=np.float64):
```
skoudoro: same comment as above

dipy/io/image.py
arokem: Could you please tell me what the docstring should say for this variable?

arokem: Or feel free to add a docstring in #2031, superseding this PR.

skoudoro: Too late it seems, already merged. so you can add: 

```python
as_ndarray: bool
    convert nibabel ArrayProxy to a numpy.ndarray.
    If you want to save memory and delay this operation, just turn this option to False.
    (default: (True))
```

skoudoro:
superseded by #2032, closing

PR #2033, opened 2020-01-11
RF - Direction getters naming

gabknight:
Change direction getter naming in preparation of new direction getters based on peaks object (PR in prep).

- class BaseDirectionGetter -> BasePmfDirectionGetter
- file peak_direction_getter.pyx -> eudx_direction_getter.pyx
- file test_peak_dg.py -> test_eudx_dg.py


skoudoro:
Hi @gabknight,

Thank you for this update. If this naming conventions change is really necessary, we will need to set up a deprecation cycle for at least 2 releases and keep the old system (release 1.0.0 was an exception for breaking without any deprecation cycle)
So we need:
- to keep the old file and create a new one
- In the old file, add a deprecation warning and call the new one.
 
Let me know if you need help with that and I will create a PR on local branch/dipy repo.

gabknight:
Hi @skoudoro, thanks for your response. I wasn't sure how to do the deprecation for a class. Let me know if you would do it differently. 

The warning is in peak_direction_getter **.py** rather than **.pyx**. Do you think this is an issue?

skoudoro:
Hi @gabknight, 

I think it is good the way you do it!

It is just missing a small test to make sure that the warning popup.

Furthermore,  I need to test it quickly.  Thank you !

gabknight:
@skoudoro, Thanks I added the test. 

Do you know why those build failed?

skoudoro:
Not yet, need to look into it.

- it seems that something change in Nibabel and their Numpy minimum version is different from ours.
- Or maybe, they just dropped python3.5

Anyway, I need to look.

dipy/direction/closest_peak_direction_getter.pyx
```diff
@@ -56,7 +56,7 @@ cdef int closest_peak(np.ndarray[np.float_t, ndim=2] peak_dirs,
             return 0
     return 1
 
-cdef class BaseDirectionGetter(DirectionGetter):
+cdef class BasePmfDirectionGetter(DirectionGetter):
```
skoudoro: `BaseDirectionGetter` needs a deprecation cycle as you did for `EuDXDirectionGetter`. 

Can you add this deprecation and a small test for it? 

Thank you

skoudoro:
Sorry for the delay but thank you @gabknight! Everything looks good and you can continue with #2036. merging

PR #2035, opened 2020-01-11
DOC: Update DKI documentation according to the new get data functions

RafaelNH:
Just a small correction in DKI's example. We forgot to update the documentation when changing the read data functions.


arokem:
Oh yeah - good catch!

skoudoro:
Thank you for that @RafaelNH !

PR #2041, opened 2020-01-16
BF: SFM prediction with mask

arokem:
This fixes a bug in the way that SFM does signal prediction when fit with a mask. This is something I just stumbled upon earlier today while I was working with @dPys on https://github.com/nipreps/dmriprep/pull/62. This should fix the issues we were seeing there.

### DIFF
--- a/dipy/reconst/sfm.py
+++ b/dipy/reconst/sfm.py
@@ -53,7 +53,11 @@ def _to_fit_iso(data, gtab, mask=None):
     if mask is None:
         mask = np.ones(data.shape[:-1], dtype=bool)
     # Turn it into a 2D thing:
-    data = data[mask]
+    if len(mask.shape) > 0:
+        data = data[mask]
+    else:
+        # This handles the corner case of fitting a single voxel:
+        data = data.reshape((-1, data.shape[0]))
     data_no_b0 = data[:, ~gtab.b0s_mask]
     nzb0 = data_no_b0 > 0
     nzb0_idx = np.where(nzb0)
--- a/dipy/reconst/sfm.py
+++ b/dipy/reconst/sfm.py
@@ -25,11 +25,9 @@ except ImportError:
     from scipy.stats import nanmean
 
 from dipy.utils.optpkg import optional_package
-import dipy.core.geometry as geo
 import dipy.core.gradients as grad
 import dipy.core.optimize as opt
 import dipy.sims.voxel as sims
-import dipy.reconst.dti as dti
 import dipy.data as dpd
 from dipy.reconst.base import ReconstModel, ReconstFit
 from dipy.reconst.cache import Cache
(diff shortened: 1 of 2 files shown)


dPys:
@arokem - this is great! are there any further optimizations that can be made in runtime for the predict method of isotropic specifically? (Or maybe a better question would be what particular lines take the longest to run at the moment?)

arokem:
@dPys : without a doubt, the most time is spent in this line: https://github.com/dipy/dipy/blob/master/dipy/reconst/sfm.py#L433, which is where sklearn's ElasticNet fit is being called. Unfortunately, I don't think there's much we can do about it.

dPys:
@arokem - 

A few more questions:

1) what aspect(s) of signal prediction would be lost when reducing the ratio of L1 penalty (i.e. closer to if not equal to 0)?

2) is the estimator fit at every voxel in the ROI? what other data dimensionality reduction strategies might improve fit and/or increase the speed of prediction?

3) maybe related to (2) — could a T1w image, registered to the B0 template, be incorporated into this somehow?

Curious to hear from anyone else who wants to chime in on this.

arokem:
> what aspect(s) of signal prediction would be lost when reducing the ratio of L1 penalty (i.e. closer to if not equal to 0)?

That depends on the data. In my experience, signal prediction can be almost equally accurate with various ratios, leading to quite different fODFs. 

> is the estimator fit at every voxel in the ROI? what other data dimensionality reduction strategies might improve fit and/or increase the speed of prediction? Currently the prediction with this model is probably much too slow to lend to predicting from every vector coordinate in the case of head motion correction.

Yes. As suggested  by @mattcieslak, one way to make this go much faster is  to provide an initialized sklearn [Ridge](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.Ridge.html) object as the input to the `solver` keyword argument when initializing the SparseFascicleModel object.

> maybe related to (2) — could a T1w image, registered to the B0 template, be incorporated into this somehow? 

I am not sure. There might be different levels of regularization that work better in different tissue types, but it would require quite a bit of empirical work to sort this out. I would be inclined to first try one ridge regularization parameter everywhere and see how that works.

dipy/reconst/sfm.py
```diff
@@ -51,13 +49,21 @@
 # extensible, by inheriting from IsotropicModel/IsotropicFit below:
 
 # First, a helper function to derive the fit signal for these models:
-def _to_fit_iso(data, gtab):
-    data_no_b0 = data[..., ~gtab.b0s_mask]
+def _to_fit_iso(data, gtab, mask=None):
+    if mask is None:
+        mask = np.ones(data.shape[:-1], dtype=bool)
+    # Turn it into a 2D thing:
+    if len(mask.shape) > 0:
```
skoudoro: I do not understand this part. Can you tell us more about this part?

arokem: Yes: this is just to deal with the (rather unusual) case that your data is a one-dimensional vector representing one voxel's worth of data. In that case, `data.shape[:-1]` is  `()` and you might run into trouble here.

skoudoro: ok, I see, thanks. I wonder if there is a better way to do that but It is fine like this too

skoudoro:
Anyone to review this PR? Do you want to have a last look @dPys?

I will wait until Sunday to merge it.

skoudoro:
ok, Thanks @arokem, merging

dPys:
Hi @skoudoro -- sorry to chime back in late this convo, but thanks for merging. I took another look at all of @arokem 's edits and they indeed looked good to go.

Cheers,
@dPys

PR #2042, opened 2020-01-16
Add tests to rng module

### LINKED ISSUE #1864: Add tests to dipy.core.rng
Unless I am missing something thing, classes implemented in this module don't seem to be used anywhere else.

skoudoro:
This PR adds tests for `WichmannHill2006`, `WichmannHill1982`, `LEcuyer`.

fix #1864

### DIFF
--- a/dipy/core/rng.py
+++ b/dipy/core/rng.py
@@ -1,11 +1,13 @@
-""" Random number generation utilities """
+"""Random number generation utilities."""
 
 from math import floor
 from platform import architecture
+import numpy as np
 
 
 def WichmannHill2006():
-    """
+    """Wichmann Hill (2006) random number generator.
+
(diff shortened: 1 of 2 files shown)


arokem:
Any idea where that `global s1` variable is supposed to be defined?

skoudoro:
It seems to be an old function, I think, I will make `s1` a parameter of this function. This is just a randomState. I do not know why it was global before but it does not seem adapted

arokem:
👍 , which is why it's great that you are adding tests!

skoudoro:
Thank you for the feedback 😃, update done and code rebased.

arokem:
Why codecov thinks this is a *reduction* in coverage just baffles me. But off we go!

skoudoro:
Agree, weird sometimes... I need to spend time to understand more Codecov one day...

PR #2043, opened 2020-01-22
[NF] Add a Deprecation system

skoudoro:
The goal of this PR is to bring a standard way to deprecate functions, modules, ... 

This is heavily based on the Nibabel deprecation system and adds a dependency to the small package named `packaging`. Usually, you do not need to install this package because `setuptools` depends on it and already did the installation for you. But it can happen that `setuptools` is not installed in some environments.


skoudoro:
Any opinion concerning this PR?

dipy/testing/__init__.py
```diff
@@ -43,8 +43,76 @@ def assert_arrays_equal(arrays1, arrays2):
...
+    Examples
+    --------
+    >>> import warnings
+    >>> with clear_and_catch_warnings(modules=[np.core.fromnumeric]):
+    ...     warnings.simplefilter('always')
+    ...     # do something that raises a warning in np.core.fromnumeric
+
+    Note
+    ----
+    this class is copied (with minor modifications) from the Nibabel.
```
Garyfallidis: This class is copied (with minor modifications) from Nibabel

dipy/utils/tests/test_deprecator.py
```diff
@@ -0,0 +1,169 @@
+"""Module to test ``dipy.utils.deprecator`` module.
+
+Notes
+-----
+this file is copied (with minor modifications) from the Nibabel.
```
Garyfallidis: Same as before

Garyfallidis:
So ``package`` will be installed if not installed in the system?

skoudoro:
> So package will be installed if not installed in the system?

Yes, for all future release.

Concerning the current master, you may have to install it. But I suppose it is already done since you need `setuptools` to install DIPY and `setuptools` depends on `packaging`

Garyfallidis:
@arokem can you check this PR too?

arokem:
Yeah. LGTM!

PR #2045, opened 2020-01-30
[Fix] workflow variable string

skoudoro:
A quick fix to make sure that we unpack variable string when we check file existence. Also, some docstring + pep8 improvement


Garyfallidis:
:)

Issue #2050, opened 2020-02-18
BUG: Polyfit in IVIM TRR returns Negative Values


ShreyasFadnavis:
Following up with this issue with a PR!

Potential fixes:
- Use `scipy.optimize.lsq_linear` and set the bounds manually
or 
- Use `scipy.optimize.nnls`

I assume `nnls` will be faster as it is running FORTAN under the hood.

Any other suggestions?

ShreyasFadnavis:
Any other suggestions anyone? @arokem @skoudoro @RafaelNH ?

skoudoro:
Good catch and +1 for `scipy.optimize.nnls`

arokem:
Yes. NNLS should be the way to go. I'm not sure about performance, but it will probably be more correct than trying to implement the bounds by ourselves.

PR #2051, opened 2020-02-19
BF: Non-negative Least Squares for IVIM 

ShreyasFadnavis:
- Working version to fix: #2050 
- Needs more rigorous testing on quality

- [ ] Test against: `lsq_linear` and `curve_fit` from `scipy`

### DIFF
--- a/dipy/reconst/ivim.py
+++ b/dipy/reconst/ivim.py
@@ -1,7 +1,7 @@
 """ Classes and functions for fitting ivim model """
 
 import numpy as np
-from scipy.optimize import least_squares, differential_evolution
+from scipy.optimize import least_squares, differential_evolution, nnls
 import warnings
 from dipy.reconst.base import ReconstModel
 from dipy.reconst.multi_voxel import multi_voxel_fit
@@ -10,7 +10,7 @@ cvxpy, have_cvxpy, _ = optional_package("cvxpy")
 
 
 # global variable for bounding least_squares in both models
-BOUNDS = ([0., 0., 0., 0.], [np.inf, .2, 1., 1.])
+BOUNDS = ([0., 0., 0., 0.], [np.inf, .9, .9, .9])
 
 
 def ivim_prediction(params, gtab):
@@ -344,7 +344,11 @@ class IvimModelTRR(ReconstModel):
             return IvimFit(self, params_linear)
 
     def estimate_linear_fit(self, data, split_b, less_than=True):
-        """Estimate a linear fit by taking log of data.
+        """Estimate a linear fit by taking log of data. We make use of a non-
+        negative least squares fitting to constrain the diffusion coefficient
+        to be positive. It solves the KKT (Karush-Kuhn-Tucker) conditions for
+        the non-negative least squares problem:
+        ``argmin_x || Ax - b ||_2`` for ``x>=0``
 
         Parameters
         ----------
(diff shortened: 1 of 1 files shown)


dipy/reconst/ivim.py
```diff
@@ -364,20 +368,28 @@ def estimate_linear_fit(self, data, split_b, less_than=True):
...
+            _, neg_log_S0 = np.polyfit(bvals_split, b, 1)
+
         else:
             bvals_split = self.gtab.bvals[self.gtab.bvals >= split_b]
-            D, neg_log_S0 = np.polyfit(bvals_split,
-                                       -np.log(data[self.gtab.bvals >=
-                                                    split_b]), 1)
+            _b = -np.log(data[self.gtab.bvals >= split_b])
+
+            D = nnls(np.matrix(bvals_split).T, _b)[0]
```
arokem: Do you need the `[0]` both here and below on line 392?

ShreyasFadnavis: Yes, so... D returned by `nnls` is a single element `array`. with `D[0]` I am picking out that element to ease the `IvimFit`. Does that make sense?

dipy/reconst/ivim.py
```diff
@@ -364,20 +368,28 @@ def estimate_linear_fit(self, data, split_b, less_than=True):
...
+
         """
         if less_than:
             bvals_split = self.gtab.bvals[self.gtab.bvals <= split_b]
-            D, neg_log_S0 = np.polyfit(bvals_split,
-                                       -np.log(data[self.gtab.bvals <=
-                                                    split_b]), 1)
+            b = -np.log(data[self.gtab.bvals <= split_b])
+
+            D = nnls(np.matrix(bvals_split).T, b)[0]
```
arokem: The variable name `b` is a bit confusing here. Maybe `neg_log_S`? 

Also, why the negative sign?

Finally, I might be confused: to fit diffusivity, shouldn't we be using the data from b-values larger than `split_b`?

ShreyasFadnavis: For the >= `split_b` see line `385`

and I am using the negative sign as it was in the original implementation, therefore did not touch it...

I will fix the variable names :+1:

arokem: Does `nnls` require the input to be a numpy `matrix` object? If not, I would avoid using that class -- it's deprecated (see https://stackoverflow.com/a/53254739/3532933 for a lot of detail about this).

ShreyasFadnavis: Yes, its compulsory! Throws an error otherwise... I tried it.

arokem: Also, one more thought: do we need to also include an intercept term here? Maybe add a column of 1's to the design matrix?

arokem: I think that I see why. If you do this:

```
X = np.random.randn(100)                                               
y = np.random.randn(100)                                               
coef = nnls(X, y)   
```

it raises:

```
<ipython-input-14-1a1020100ad2> in <module>
----> 1 coef = nnls(X, y)

~/.virtualenvs/dipy/lib/python3.7/site-packages/scipy/optimize/nnls.py in nnls(A, b, maxiter)
     44 
     45     if len(A.shape) != 2:
---> 46         raise ValueError("expected matrix")
     47     if len(b.shape) != 1:
     48         raise ValueError("expected vector")

ValueError: expected matrix

```

But it's not a great error message, because it suggests that you need a `np.matrix` object, where really what is required is an array with more two dimensions (see line 45 in that traceback), so this works: 

```
coef = nnls(X[:, np.newaxis], y)
```

without using a `np.matrix` object at all. This will also work when you add ones to account for the intercept. Something like: 

```
coef = nnls(np.vstack([X, np.ones(X.shape[0])]).T, y)    
```

should work for that.

arokem: See https://github.com/scipy/scipy/pull/11574

ShreyasFadnavis: Thank you for this @arokem ! Let me try this and get back ASAP :+1:

ShreyasFadnavis: image
This is what the `D` map looks like. The black region is mostly zero. I tried both of the above suggestions made by you :+1: still don't see any difference...

Even the D* looks similar:
image

Maybe the results aren't wrong, but I haven't seen exactly `0` values for `D` and `D*`.

ShreyasFadnavis: @arokem, one more thing I forgot to mention is that `NNLS` seems to become sensitive to the bounds. For instance, when I change the bounds to `([0, 0, 0, 0], [np.inf, .9, .9, .9])` the result for D looks like this:
image

The places where you see the black spots above are perfect `0` or something like `2.9e-28`

arokem: That's odd. As far as I can tell, NNLS doesn't know anything about these bounds. Unless I am missing something?

ShreyasFadnavis: Nope.. I think you are right! But I also think that the results that we are seeing are not wrong either. Think of a case where there is a lot of noise, if you solve a QP, it is likely that we see a horizontal line right?

I was looking into what others are doing, and seems like they do a `polyfit` and discard+interpolate the values where D is negative. For instance: https://www.mathworks.com/matlabcentral/fileexchange/65579-ivim-model-fitting -> `IVIM_seg`

Does this: 
image

arokem: Hey @ShreyasFadnavis : could you try changing this line to  `coef = nnls(np.vstack([X, np.ones(X.shape[0])]).T, y)` as suggested above?

ShreyasFadnavis: Hi @arokem I did try doing this.. but did not give anything better! I just did not update this PR with that commit...

skoudoro:
What is the status of this PR @ShreyasFadnavis? All CI's Are failing. Can you give us an overview? Should I expect this to be ready for the release this week?

Thank you for your feedback

ShreyasFadnavis:
@skoudoro this needs a lot more testing and also does not have any standard literature so far. I think let's aim for the next release cycle for this PR!

Garyfallidis:
Any plans to progress with this PR @ShreyasFadnavis ? Or close it?

ShreyasFadnavis:
> Any plans to progress with this PR @ShreyasFadnavis ? Or close it?

Closing this for now!

Issue #1991, opened 2019-10-27
[DISCUSSION] b-tensor encoding gradient table format 


mdesco:
## Description
Moving an email thread to this DIPY issue. 
We need to discuss plan of action for b-tensor encoding gtab support in DIPY. 
@RafaelNH should take the lead or initiate a Doodle.

### Previous email discussion

Mail 1:
Hey Rafael and Dipy-ers,

I missed the lab meeting yesterday, sorry. Etienne talked to me today.

I’m cc-ing 2 students from my lab involved in b-tensor encoding and md-dmri dev in Sherbrooke. Both code in python. Philippe has worked on phase correction, complex data processing and now, b-tensor encoding spherical deconvolution. Alex has worked on DIAMOND and Magic-DIAMOND with my previous post-doc and Benoit Scherrer.

So, if you need:
1) Data, we have and we can share (linear, planar, spherical) on a clinical 3T Philips
  (Filip Scz. Has published a data paper recently also. Data is public)

2) Advise on bval/bvec encoding. We are used to deal with it. And we could be involved.

3) other…

Let us know how we can help.
Max, Phil, Alex

Response 1: 
Hi Maxime et al., 

Many thanks for your e-mail. Indeed, the data that Filip published might be a good starting point for our developments. At the moment, I was thinking to start implementing some techniques assuming that microstructural components are Gaussian and time-independent, so having the information of the full b-tensor in dipy's gradient table might be enough to start incorporating some reconstruction models into dipy. What do you think? 

Best,
Rafael

Response 2:
Hi Rafael,

Maybe we need a separate call for this. I'm not sure what you mean by "full b-tensor". For us, the gradient table is one thing and the "trajectory" another, which is coded as an analytical function when you know you are in the linear, planar or spherical case. So, aside from indicating that the bvals/bvecs are for a planar, spherical or linear b-tensor encoding, there is nothing else to be saved in the gradient table...

Maybe we are saying the samething with different terms. We could have a quick call between you, Phil, Alex and I if you want.

Max

RafaelNH:
Thanks @mdesco! Yes, I will take the lead on this! I will go back to it ASAP!

arokem:
Thanks for moving the conversation here! Worth also keeping an eye on
related discussions going on in the BIDS standard repo:
https://github.com/bids-standard/bids-specification/issues/349 and
https://github.com/bids-standard/bids-specification/pull/352.

For completeness, and for future reference, the link to the data is:
[...]

PR #2055, opened 2020-02-20
Multidimensional gradient table

kerkelae:
I suggest extending the GradientTable class to contain information about b-tensor shapes.

This preliminary code calculates the b-tensors for each acquisition and saves the information as an attribute of the GradientTable. The user can explicitly define the used tensor shapes as linear, planar, or spherical. Next step is to add support for arbitrary tensor shapes in the form of a 3 x 3 matrix or an array of such matrices. If the user does not specify the tensor shape, the code assumes the acquisition to have been performed using linear tensors (traditional single diffusion encoding), so the gradient_table function and GradientTable class can be used as before.

Before finishing the implementation, I would like to ask you for your opinion on the organization of the proposed code.

After this is implemented, the next steps can be implementing multidimensional simulations (Topgaard. "Multidimensional diffusion MRI". https://doi.org/10.1016/j.jmr.2016.12.007) and simple methods for estimating microscopic diffusion anisotropy from data acquired with multidimensional diffusion encoding (Yang, et al. "Double diffusion encoding MRI for the clinic."  https://doi.org/10.1002/mrm.27043, Nery, et al. "In vivo demonstration of microscopic anisotropy in the human kidney using multidimensional diffusion MRI." https://doi.org/10.1002/mrm.27869).

### DIFF
--- a/dipy/core/gradients.py
+++ b/dipy/core/gradients.py
@@ -69,20 +69,24 @@ class GradientTable(object):
         self.big_delta = big_delta
         self.small_delta = small_delta
         self.b0_threshold = b0_threshold
+        if btens is not None:
+            linear_tensor = np.array([[1, 0, 0],
+                                      [0, 0, 0],
+                                      [0, 0, 0]])
+            planar_tensor = np.array([[1, 0, 0],
+                                      [0, 1, 0],
+                                      [0, 0, 0]]) / 2
+            spherical_tensor = np.array([[1, 0, 0],
+                                         [0, 1, 0],
+                                         [0, 0, 1]]) / 3
         if isinstance(btens, str):
             b_tensors = np.zeros((len(self.bvals), 3, 3))
             if btens == 'LTE':
-                b_tensor = np.array([[1, 0, 0],
-                                     [0, 0, 0],
-                                     [0, 0, 0]])
+                b_tensor = linear_tensor
             elif btens == 'PTE':
-                b_tensor = np.array([[0, 0, 0],
-                                     [0, 1, 0],
-                                     [0, 0, 1]]) / 2
+                b_tensor = planar_tensor
             elif btens == 'STE':
-                b_tensor = np.array([[1, 0, 0],
-                                     [0, 1, 0],
-                                     [0, 0, 1]]) / 3
+                b_tensor = spherical_tensor
             else:
                 raise ValueError("%s is an invalid value for btens. "%btens
                                  + "Please provide one of the following: "
@@ -105,22 +109,13 @@ class GradientTable(object):
             for i, (bvec, bval) in enumerate(zip(self.bvecs, self.bvals)):
                 R = vec2vec_rotmat(np.array([1, 0, 0]), bvec)
                 if btens[i] == 'LTE':
-                    b_tensor = np.array([[1, 0, 0],
-                                         [0, 0, 0],
-                                         [0, 0, 0]])
-                    b_tensors[i] = (np.matmul(np.matmul(R, b_tensor), R.T)
+                    b_tensors[i] = (np.matmul(np.matmul(R, linear_tensor), R.T)
                                     * bval)
                 elif btens[i] == 'PTE':
-                    b_tensor = np.array([[0, 0, 0],
-                                         [0, 1, 0],
-                                         [0, 0, 1]]) / 2
-                    b_tensors[i] = (np.matmul(np.matmul(R, b_tensor), R.T)
+                    b_tensors[i] = (np.matmul(np.matmul(R, planar_tensor), R.T)
                                     * bval)
                 elif btens[i] == 'STE':
-                    b_tensor = np.array([[1, 0, 0],
-                                         [0, 1, 0],
-                                         [0, 0, 1]]) / 3
-                    b_tensors[i] = b_tensor * bval
+                    b_tensors[i] = spherical_tensor * bval
                 else:
                     raise ValueError(
                             "%s is an invalid value in btens array. "%btens[i]
@@ -210,8 +205,8 @@ def gradient_table_from_bvals_bvecs(bvals, bvecs, b0_threshold=50, atol=1e-2,
            with the corresponding gradient direction and the planar tensor's
            normal is aligned with the corresponding gradient direction.
            Magnitude is scaled to match the b-value.
-        3. an arry of shape (N,3,3) specifying the b-tensor of each volume
-           exactly. N corresponds to the number volumes in data No rotation of
+        3. an array of shape (N,3,3) specifying the b-tensor of each volume
+           exactly. N corresponds to the number volumes in data. No rotation of
            scaling is performed.
 
     Other Parameters
@@ -483,7 +478,7 @@ def gradient_table(bvals, bvecs=None, big_delta=None, small_delta=None,
            normal is aligned with the corresponding gradient direction.
            Magnitude is scaled to match the b-value.
         3. an arry of shape (N,3,3) specifying the b-tensor of each volume
-           exactly. N corresponds to the number volumes in data No rotation of
+           exactly. N corresponds to the number volumes in data No rotation or
            scaling is performed.
 
     Returns
--- a/dipy/core/gradients.py
+++ b/dipy/core/gradients.py
@@ -97,11 +97,11 @@ class GradientTable(object):
             self.btens = b_tensors
         elif (isinstance(btens, np.ndarray) and (btens.shape ==
                 (gradients.shape[0],) or (btens.shape ==
-                (gradients.shape[0],1)) or (btens.shape == (1,
+                (gradients.shape[0], 1)) or (btens.shape == (1,
                 gradients.shape[0])))):
             b_tensors = np.zeros((len(self.bvals), 3, 3))
-            if (btens.shape == (1,gradients.shape[0],1)):
-                btens = btens.reshape((gradients.shape[0],1))
+            if btens.shape == (1,gradients.shape[0], 1):
+                btens = btens.reshape((gradients.shape[0], 1))
             for i, (bvec, bval) in enumerate(zip(self.bvecs, self.bvals)):
                 R = vec2vec_rotmat(np.array([1, 0, 0]), bvec)
                 if btens[i] == 'LTE':
@@ -451,7 +451,7 @@ def gradient_table(bvals, bvecs=None, big_delta=None, small_delta=None,
 
     atol : float
         All b-vectors need to be unit vectors up to a tolerance.
-        
+
     btens : can be any of two options
 
         1. a string specifying the shape of the encoding tensor shape for all
--- a/dipy/core/gradients.py
+++ b/dipy/core/gradients.py
@@ -44,7 +44,7 @@ class GradientTable(object):
         Gradients with b-value less than or equal to `b0_threshold` are
         considered to not have diffusion weighting.
     btens : (N,3,3) ndarray
-        The b-tensors of each gradient direction. 
+        The b-tensor of each gradient direction.
 
     See Also
     --------
@@ -58,7 +58,7 @@ class GradientTable(object):
 
     """
     def __init__(self, gradients, big_delta=None, small_delta=None,
-                 b0_threshold=50, btens='LTE'):
+                 b0_threshold=50, btens=None):
         """Constructor for GradientTable class"""
         gradients = np.asarray(gradients)
         if gradients.ndim != 2 or gradients.shape[1] != 3:
@@ -69,8 +69,8 @@ class GradientTable(object):
         self.big_delta = big_delta
         self.small_delta = small_delta
         self.b0_threshold = b0_threshold
-        b_tensors = np.zeros((len(self.bvals), 3, 3))
         if isinstance(btens, str):
+            b_tensors = np.zeros((len(self.bvals), 3, 3))
             if btens == 'LTE':
                 b_tensor = np.array([[1, 0, 0],
                                      [0, 0, 0],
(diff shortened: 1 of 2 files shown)


dipy/core/tests/test_gradients.py
```diff
@@ -81,6 +81,69 @@ def test_GradientTable():
...
+        npt.assert_almost_equal(np.trace(bten), bval)        
+        # Check that the b tensor orientation is correct
+        if btens[i] == 'LTE':
+            if bval != 0:
+                evals, evecs = np.linalg.eig(bten)
+                dot_prod = np.dot(np.real(evecs[:,np.argmax(evals)]), 
+                                  gt.bvecs[i])
+                npt.assert_almost_equal(np.abs(dot_prod), 1)
+                
+    # Check invalide
```
arokem:
```suggestion
    # Check invalid
```

dipy/core/gradients.py
```diff
@@ -67,7 +69,61 @@ def __init__(self, gradients, big_delta=None, small_delta=None,
...
+                                    * bval)
+                elif btens[i] == 'STE':
+                    b_tensor = np.array([[1, 0, 0],
+                                         [0, 1, 0],
+                                         [0, 0, 1]]) / 3
+                    b_tensors[i] = b_tensor * bval
+                else:
+                    raise ValueError("Invalid value in btens array")     
+        else:
+                raise ValueError("Invalid value for btens")
```
arokem: Again: probably want better error handling here. Tell the users what they gave you (they might not know, because there's another layer between them and this code!) and what they should give you.

dipy/core/gradients.py
```diff
@@ -43,6 +43,8 @@ class GradientTable(object):
     b0_threshold : float
         Gradients with b-value less than or equal to `b0_threshold` are
         considered to not have diffusion weighting.
+    btens : (N,3,3) ndarray
```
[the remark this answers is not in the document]
kerkelae: The btens attribute of GradientTable contains the b-tensors for each volume in the form of a 3 by 3 matrix. The constructor argument btens is currently a string or a list of strings, based on which the b-tensors are calculated. I am currently extending the constructor to also support 3 by 3 matrices or arrays of such matrices.

dipy/core/gradients.py
```diff
@@ -67,6 +69,68 @@ def __init__(self, gradients, big_delta=None, small_delta=None,
...
+                    R = vec2vec_rotmat(np.array([1, 0, 0]), bvec)
+                    b_tensors[i] = (np.matmul(np.matmul(R, b_tensor), R.T)
+                                    * bval)
+            self.btens = b_tensors
+        elif (isinstance(btens, np.ndarray) and (btens.shape ==
+                (gradients.shape[0],) or (btens.shape ==
+                (gradients.shape[0], 1)) or (btens.shape == (1,
+                gradients.shape[0])))):
+            b_tensors = np.zeros((len(self.bvals), 3, 3))
+            if btens.shape == (1,gradients.shape[0], 1):
```
RafaelNH: PEP8 - space between '1,' and  'gradient.shape'

RafaelNH:
Hi @kerkelae! Great job! Just want to ask you to support also the option of giving the b-matrix as an input, i.e. array (N, 3, 3) - btensors can be assumed to be already rotated. Apart from that this is ready to be merged at my side.

kerkelae:
Hi, the most recent commit addresses the issues raised by @arokem and @RafaelNH.

arokem:
Yes. I think that this is good to go. The CI failure looks like it's a time-out?

Garyfallidis:
Give me a day to look at this. Thank you in advance.

areymbaut:
Hey guys! I was a post-doc with Maxime Descoteaux in Sherbrooke, Canada, and I'm currently doing a post-doc with Daniel Topgaard in Lund, Sweden. I've been following the discussion here and it looks pretty neat. From what I understand, it will either be possible to 1) work with linear, planar and spherical b-tensors (which is quite practical for most of what's been done so far in tensor-valued diffusion encoding), or 2) input an arbitrary b-tensor (which is great for current research in the field). 

I would just have one suggestion: would it be possible to also include a "cigar" b-shape in the accepted default axisymmetric shapes? To be clear, if you consider that planar tensors have a normalized anisotropy b_\Delta=-0.5 (within the Haeberlen convention for symmetric positive-definite tensors), spherical tensors have b_\Delta=0, and linear tensors have b_\Delta=1, then cigar tensors have b_\Delta=+0.5. If implemented, pretty much any work regarding axisymmetric b-tensors will only have to rely on the 4 default shapes.

Thanks!

arokem:
Hey @areymbaut : thanks for pitching in here! 

Just to be clear, you want the following? 

```
b_tensor = np.array([[1, 0, 0],
                                     [0, 0.5, 0],
                                     [0, 0, 0.5]]) / 2
```

to be a "canned" option?

Possibly a minor point here, but what would be a good three-letter acronym for the cigar-shaped tensor? Would ETE for ellipsoid tensor encoding work for you? PTE (for "prolate tensor encoding") wouldn't work, because it's already taken by the planar tensor encoding.

dipy/core/gradients.py
```diff
@@ -67,6 +69,71 @@ def __init__(self, gradients, big_delta=None, small_delta=None,
         self.big_delta = big_delta
         self.small_delta = small_delta
         self.b0_threshold = b0_threshold
+        if isinstance(btens, str):
+            b_tensors = np.zeros((len(self.bvals), 3, 3))
+            if btens == 'LTE':
```
jhlegarreta: We may gain some consistency and it may be easier to maintain if the supported tensor types are put to an enum, e.g.

```
class TensorEncoding(enum.Enum):
    """Tensor encoding types."""
    LTE = 'linear_tensor_encoding'
    PTE = 'planar_tensor_encoding'
    STE = 'spherical_tensor_encoding'
```

and then check as e.g. `if btens == TensorEncoding.LTE`. The `ValueError` exception below will need to be modified to use the above enum instead of the strings. Similarly, the test will need to be updated if the enum is adopted.

Given that the `LTE`, `PTE`, `STE` acronyms have been adopted in literature, may be having `TensorEncoding.LTE` is a bit verbose, so the enum may need another name.

dipy/core/gradients.py
```diff
@@ -67,6 +69,71 @@ def __init__(self, gradients, big_delta=None, small_delta=None,
...
+                b_tensor = np.array([[1, 0, 0],
+                                     [0, 0, 0],
+                                     [0, 0, 0]])
```
jhlegarreta: These three matrice are used at least twice here. Can we think of defining them only once and re-using the object?

dipy/core/gradients.py
```diff
@@ -67,6 +69,71 @@ def __init__(self, gradients, big_delta=None, small_delta=None,
...
+        elif (isinstance(btens, np.ndarray) and (btens.shape ==
+                (gradients.shape[0],) or (btens.shape ==
+                (gradients.shape[0], 1)) or (btens.shape == (1,
+                gradients.shape[0])))):
+            b_tensors = np.zeros((len(self.bvals), 3, 3))
+            if btens.shape == (1, gradients.shape[0]):
+                btens = btens.reshape((gradients.shape[0], 1))
+            for i, (bvec, bval) in enumerate(zip(self.bvecs, self.bvals)):
+                R = vec2vec_rotmat(np.array([1, 0, 0]), bvec)
+                if btens[i] == 'LTE':
```
jhlegarreta: Same comment as above concerning the supported tensor types and the exception.

dipy/core/gradients.py
```diff
@@ -127,6 +194,25 @@ def gradient_table_from_bvals_bvecs(bvals, bvecs, b0_threshold=50, atol=1e-2,
     atol : float
         Each vector in `bvecs` must be a unit vectors up to a tolerance of
         `atol`.
+    btens : can be any of three options
+        1. a string specifying the shape of the encoding tensor for all volumes
+           in data. Options: 'LTE', 'PTE', 'STE' corresponding to linear,
```
jhlegarreta: Use the enum instead of the strings?

dipy/core/gradients.py
```diff
@@ -127,6 +194,25 @@ def gradient_table_from_bvals_bvecs(bvals, bvecs, b0_threshold=50, atol=1e-2,
...
+           b-value.
+        2. an array of strings of shape (N,), (N, 1), or (1, N) specifying
+           encoding tensor shape for each volume separately. N corresponds to
+           the number volumes in data. Options for elements in array: 'LTE',
+           'PTE', 'STE' corresponding to linear, planar, and spherical tensor
+           encoding. Tensors are rotated so that the linear tensor is aligned
+           with the corresponding gradient direction and the planar tensor's
+           normal is aligned with the corresponding gradient direction.
+           Magnitude is scaled to match the b-value.
+        3. an arry of shape (N,3,3) specifying the b-tensor of each volume
```
jhlegarreta: arry -> array

dipy/core/gradients.py
```diff
@@ -127,6 +194,25 @@ def gradient_table_from_bvals_bvecs(bvals, bvecs, b0_threshold=50, atol=1e-2,
...
+           exactly. N corresponds to the number volumes in data No rotation of
```
jhlegarreta: (...) in data No rotation (...) -> (...) in data. No rotation (...)

dipy/core/gradients.py
```diff
@@ -127,6 +194,25 @@ def gradient_table_from_bvals_bvecs(bvals, bvecs, b0_threshold=50, atol=1e-2,
...
+           scaling is performed.
```
jhlegarreta: _No rotation of scaling is performed_ : can this be re-worded or clarified, e.g.
_No rotation or scaling is performed_
or
_Neither rotation nor scaling is performed_
or
_The result of the scaling is not rotated_
?

dipy/core/gradients.py
```diff
@@ -127,6 +194,25 @@ def gradient_table_from_bvals_bvecs(bvals, bvecs, b0_threshold=50, atol=1e-2,
     atol : float
         Each vector in `bvecs` must be a unit vectors up to a tolerance of
         `atol`.
+    btens : can be any of three options
```
jhlegarreta: I think the supported types still need to be summarized next to the variable name.

areymbaut:
Hey @arokem : if you consider b*Diag(1,0,0) for a linear b-tensor (LTE), b/2*Diag(0,1,1) for a planar b-tensor (PTE) and b/3*Diag(1,1,1) for a spherical b-tensor (STE), then a cigar b-tensor is given by b/3*Diag(2, 0.5, 0.5). These expressions ensure that the trace of the b-tensor is always the b-value b.

As for the acronym for cigar encoding, I usually go with CTE, even though it matches the acronym of Chronic Traumatic Encephalopathy from the field of concussions, because "cigar" is commonly used to describe the shape of such tensors in the literature. ETE sounds good, but also applies to b_delta values between -0.5 and 0, as "ellipsoid" is not very specific. So I let you choose what you think is best. Thanks!

Garyfallidis:
Hello @kerkelae and @RafaelNH, is this still WIP or not?

dipy/core/gradients.py
kerkelae: I followed the conventions that had been used in docstrings of other functions like gradient_table_from_qvals_bvecs.

dipy/core/gradients.py
kerkelae: Typo fixed. Thanks!

dipy/core/gradients.py
kerkelae: Typo fixed. Thanks!

dipy/core/gradients.py
kerkelae: Typo fixed. Thanks!

dipy/core/gradients.py
jhlegarreta: I see. Thanks. I'd then be for changing them in another PR, but I admit the difficulty of proposing a compact notation for all cases.

dipy/core/gradients.py
kerkelae: Done.

dipy/core/gradients.py
kerkelae: Let's leave this for another pull request after we have agreed on an optimal way of doing it.

kerkelae:
@Garyfallidis I think the minimum requirements are met. The functionality implemented here enables us to implement functions for calculating micro FA. The most recent commit addresses most of the issues raised by @jhlegarreta

dipy/core/gradients.py
jhlegarreta: May be a separate function to generate them depending on the encoding type would be more appropriate/easier to re-use, maintain and expand, but I am fine with the current solution for now. Thanks @kerkelae.

arokem:
@kerkelae : Should we also add the cigar-shaped tensor here, before merging? See https://github.com/dipy/dipy/pull/2055#issuecomment-593515911 and subsequent discussion.

arokem:
Also, please check out the automated review by Codacy. This currently doesn't pass the automated review there. Do you understand the issues raised there (see: https://app.codacy.com/manual/dipy/dipy/pullRequest?prid=5043062)?

kerkelae:
I added the cigar-shaped tensors in the most recent commit. The current implementation has too many conditionals so I will have to refactor to reduce complexity. I will try to do this in the next few days.

arokem:
This looks fine to me. @skoudoro and @Garyfallidis : how seriously should we take Codacy's advice? It seems to be pretty strict 😄

skoudoro:
> how seriously should we take Codacy's advice?

I think we should be strict about pep8. But we should be flexible with the `duplication` section or `complexity` section. I will relax both rules.  So in this PR, we are good with Codacy.

Otherwise, +1 for merging but I think @Garyfallidis want to have a last look. Let's wait for his feedback

kerkelae:
@Garyfallidis What is the status of this? Me and @fnery have Python code to fit micro FA and V_iso to MDE data in Python and would be keen to make it a part of DiPy once b-tensor info is available in gtab.

Garyfallidis:
It is in my TODO to review this PR today and possibly merge if everything is fine.

Garyfallidis:
Looks good. Merging... @kerkelae and all please do move ahead to the next stage and apologies for any delays.

skoudoro:
tag #1991

PR #2056, opened 2020-02-20
MAINT: Up Numpy version to 1.12.

arokem:
This should help with this: https://github.com/dipy/dipy/pull/2033#issuecomment-587558735. It seems that nibabel has upped their minimal requirement for 3.0
 and we can follow along.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,7 +38,7 @@ matrix:
     - python: 3.5
       env:
         # Check these values against requirements.txt and dipy/info.py
-        - DEPENDS="cython==0.29 numpy==1.12.0 scipy==1.0 nibabel==3.0.0 h5py==2.4.0 nose"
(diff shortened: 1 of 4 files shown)


skoudoro:
Thank you @arokem! Let see what is the CI's result

arokem:
It looks like the issue is with building h5py. I increased the minimal version of that as well.

arokem:
Looks like this might resolve  the issue (works on travis, but has not been tested on azure yet). 

What are your thoughts? Any compelling reason to stay with h5py version 2.4? Version 2.5 was released in 2015.

skoudoro:
Cool, Travis is passing! No reason to sta with h5py 2.4. We use the basic functionalities of h5py

skoudoro:
I think this is ready to go. I just restarted an Azure pipeline Matrix to see everything green. When it is done I will merge it.

skoudoro:
Thank you @arokem! merging

PR #1957, opened 2019-08-21
niftis2pam, pam2niftis Worflow

skoudoro:
This PR adds 2 utility command lines to convert nifti to pam5 and inverse.

**Todo:**
- [x] Add tests
- [x] Check `sphere`  management
- [x] Update doc

PR #2060, opened 2020-02-24
Replace old function with cython version

frheault:
Remove all references to `downsample()` and replaced it with `set_number_of_points()`

Downsample is giving identical results to set_number_of_points, only much slower.


dipy/tracking/metrics.py
```diff
@@ -758,71 +758,6 @@ def _extrap(xyz, cumlen, distance):
     return Lambda*xyz[ind]+(1-Lambda)*xyz[ind-1]
 
 
-def downsample(xyz, n_pols=3):
```
arokem: To not break existing code, Instead of deleting this entirely, could we do something like the following? 

    downsample = set_number_of_points

skoudoro: You can not delete this function, we need to respect a deprecation cycle for backward compatibility.
Here, how to do it:
- add the deprecation decorator like here: https://github.com/dipy/dipy/pull/1957/files#diff-096658fc22dd6806d39993cc4f1db5d8R30
- keep the function signature but call `set_number_points` underneath.

dipy/tracking/tests/test_metrics.py
```diff
@@ -8,140 +8,6 @@
 from dipy.tracking import distances as pf
 
 
-def test_downsample():
```
skoudoro: same as above. but you need to add a test to check if the deprecation warning appears. Look the following link to see an example : https://github.com/dipy/dipy/pull/1957/files#diff-1f8b2bf0bddd7571f2d196ac846a37c0R106

dipy/tracking/metrics.py
skoudoro: I just see your review, sorry, I duplicated your comment 😄

dipy/tracking/tests/test_metrics.py
```diff
@@ -1,145 +1,23 @@
...
-    # print td
-    assert_equal(len(td), pts)
-
-    res = []
-    t = np.array([[0, 0, 0], [1, 1, 1], [2, 2, 2]], 'f4')
-    for pts in range(3, 200):
-        td = tm.downsample(t, pts)
-        res.append(pts-len(td))
-    assert_equal(np.sum(res), 0)
+def test_deprecated():
```
arokem:
```suggestion
def test_downsample_deprecated():
```

arokem:
One small suggestion, to reduce ambiguity, but other than that, looks good!

skoudoro:
Looks good to me too! I just need to try quickly on my laptop

skoudoro:
I do not understand the issue with Python3.8 on OSX with Azure pipelines. For sure it is not related to your PR.

skoudoro:
Thank you @frheault!

PR #2071, opened 2020-03-09
Some change backs

kunakl07:
I have done some changes again, could you please suggest to me whether they are correct?
Thank you


skoudoro:
thank you @kunakl07, merging

PR #2090, opened 2020-03-19
Small fix reorient_bvecs

weningerleon:
reorient_gradients creates a new gradient table, but it was not possible to manually define the b0_threshold. So, when a b0 image has a bval of 60, the function crashes.

### DIFF
--- a/dipy/core/gradients.py
+++ b/dipy/core/gradients.py
@@ -444,7 +444,7 @@ def gradient_table(bvals, bvecs=None, big_delta=None, small_delta=None,
                                            atol=atol)
 
 
-def reorient_bvecs(gtab, affines, b0_threshold=50, atol=1e-2):
+def reorient_bvecs(gtab, affines, atol=1e-2):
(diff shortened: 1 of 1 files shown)


arokem:
Good catch! I would set `b0_threshold` to `None` per default and then inherit it from the input gtab. Does that make sense?

Garyfallidis:
Please update test

weningerleon:
Ok, we can just inherit b0_threshold (as well as big_delta, small_delta) from gtab. Those should not change when rotating the bvecs, so I think there is actually no need to pass b0_threshold manually to the function.

What do I need to do to update test?

skoudoro:
I think after your last commit (1c2f326), no need to add tests. It makes much more sense and it is simpler, I like it!

GTG for me. Let's wait for another +1 and we can merge it. 

Thank you @weningerleon!

arokem:
Yep. Looks good. Merging.

arokem:
Thanks again!

PR #2093, opened 2020-03-22
Placed all imports in the beginning

ShrishtiHore:
I think if all the modules are imported in the top to avoid added import calls in the middle of the code plus it seems more systematic.


ShrishtiHore:
Thank you very much @skoudoro .

PR #2094, opened 2020-03-24
Detailed Beginner Friendly Tutorials

ShrishtiHore:
I think DIPY tutorials are great but there are certain things that can just make it more beginner friendly and common to use for anyone.

1. Updates and improvements made in Github source codes are not updates immediately in real time so people are deprived of the advancements this amazing community contributes regularly.

2. Many beginners find it difficult to understand the neuroscience terms and dipy structure as well as it's utility.

3. The examples in dipy if uploaded in the form of notebooks with all visualizations and term explanations would be much helpful than the plain python source codes.

The given File is just an example of how the tutorials can be given in this Github repo such that more people can easily use it.

Please go through the **tutorials** folder. I have uploaded one sample of the notebook tutorial and I would like to contribute more such notebooks into this repo.

**Patch 2 PR link : https://github.com/ShrishtiHore/dipy/tree/ShrishtiHore-patch-2**


arokem:
Thanks @ShrishtiHore. I think that the notebook issue would be covered by #1990. 

After that one is merged, it should be possible to contribute more tutorials in the sphinx gallery format, which would be welcome.

ShrishtiHore:
Okay @arokem  Let's see what @skoudoro would say. I would love to contribute though.

skoudoro:
I agree with @arokem, This should be covered by #1990.  We do not want to maintain 2 versions of the same tutorial. DIPY has 80 tutorials so we can not create 80 notebooks and make sure that they are the same each time. It is a lot of work. Sphinx-gallery will generate those notebooks automatically and we will have to maintain "only" the 80+ tutorials and not 160+... 

I will close this PR but Thank you for this nice suggestion

PR #2097, opened 2020-03-27
Replace manual sform values with get_best_affine

frheault:
A simple switch to a much more robust Nibabel `get_best_affine()`

Manual getting the sform was often giving full zeros affine when they came from ITK/MITK/ANTs

### DIFF
--- a/dipy/io/tests/test_streamline.py
+++ b/dipy/io/tests/test_streamline.py
@@ -162,7 +162,6 @@ def io_tractogram(extension):
         nii_header = create_nifti_header(in_affine, in_dimensions,
                                          in_voxel_sizes)
         sft = StatefulTractogram(streamlines, nii_header, space=Space.RASMM)
-        print(nii_header.get_best_affine())
         save_tractogram(sft, fname, bbox_valid_check=False)
 
         if extension == 'trk':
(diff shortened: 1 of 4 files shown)


arokem:
Looks good to me. I don't know this function. Is it a relatively recent addition to nibabel? I don't think that would be a major concern as we are already requiring nibabel >3.0

dipy/io/utils.py
```diff
@@ -268,9 +268,7 @@ def get_reference_info(reference):
 
     if is_nifti:
         affine = np.eye(4).astype(np.float32)
-        affine[0, 0:4] = header['srow_x']
-        affine[1, 0:4] = header['srow_y']
-        affine[2, 0:4] = header['srow_z']
+        affine = header.get_best_affine()
```
arokem: Should we remove L270 as well?

arokem:
This all looks good to me. Anyone else want to take a look?

dipy/io/tests/test_streamline.py
```diff
@@ -162,7 +162,7 @@ def io_tractogram(extension):
         nii_header = create_nifti_header(in_affine, in_dimensions,
                                          in_voxel_sizes)
         sft = StatefulTractogram(streamlines, nii_header, space=Space.RASMM)
-
+        print(nii_header.get_best_affine())
```
skoudoro: Do we really need this print?

arokem:
Thanks!

PR #2100, opened 2020-03-29
Fixed typos, grammatical errors and time import 

ShrishtiHore:
Fixed some documentation and grammatical typos and import time in the beginning to avoid calls in between.

### DIFF
--- a/doc/examples/tissue_classification.py
+++ b/doc/examples/tissue_classification.py
@@ -4,10 +4,10 @@
 Tissue Classification of a T1-weighted Structural Image
 =======================================================
 
-This example explains how to segment a T1-weighted structural image by using a
+This example explains how to segment a T1-weighted structural image by using 
 Bayesian formulation. The observation model (likelihood term) is defined as a
-Gaussian distribution and a Markov Random Field (MRF) is used to model the a
-priori probability of the context-dependent patterns of the different tissue
+Gaussian distribution and a Markov Random Field (MRF) is used to model the 
+apriori probability of context-dependent patterns of different tissue
 types of the brain. Expectation Maximization and Iterated Conditional
 Modes are used to find the optimal solution. Similar algorithms have been
 proposed by Zhang et al. [Zhang2001]_ and Avants et al. [Avants2011]_ available
@@ -22,6 +22,7 @@ import matplotlib.pyplot as plt
 from dipy.data import get_fnames
 from dipy.io.image import load_nifti_data
 from dipy.segment.tissue import TissueClassifierHMRF
+import time
 
 """
 First we fetch the T1 volume from the Syn dataset and determine its shape.
@@ -32,7 +33,7 @@ t1 = load_nifti_data(t1_fname)
 print('t1.shape (%d, %d, %d)' % t1.shape)
 
 """
-We have fetched the T1 volume. Now we will look at the axial and the coronal
+We have fetched T1 volume. Now we will look at the axial and coronal
 slices of the image.
 """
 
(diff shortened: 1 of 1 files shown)


doc/examples/tissue_classification.py
```diff
@@ -4,10 +4,10 @@
 Tissue Classification of a T1-weighted Structural Image
 =======================================================
 
-This example explains how to segment a T1-weighted structural image by using a
+This example explains how to segment a T1-weighted structural image by using 
 Bayesian formulation. The observation model (likelihood term) is defined as a
-Gaussian distribution and a Markov Random Field (MRF) is used to model the a
-priori probability of the context-dependent patterns of the different tissue
+Gaussian distribution and a Markov Random Field (MRF) is used to model the 
+apriori probability of context-dependent patterns of different tissue
```
jhlegarreta: _a priori_ is spelled as two words, and although I am not a native English speaker I do not think the articles are necessary:

doc/examples/tissue_classification.py
```diff
@@ -32,7 +33,7 @@
 print('t1.shape (%d, %d, %d)' % t1.shape)
 
 """
-We have fetched the T1 volume. Now we will look at the axial and the coronal
+We have fetched T1 volume. Now we will look at the axial and coronal
```
jhlegarreta: I'd dare to say that the article is necessary/correct.

doc/examples/tissue_classification.py
```diff
@@ -64,21 +65,19 @@
 
 """
 Then, the smoothness factor of the segmentation. Good performance is achieved
-with values between 0 and 0.5.
+with values between 0 and 0.5. Here we are considering it as 0.1
```
jhlegarreta: If the value changes in the code, then adding this makes it more difficult to maintain.

doc/examples/tissue_classification.py
```diff
@@ -64,21 +65,19 @@
...
 """
 
 beta = 0.1
 
 """
 We could also set the number of iterations. By default this parameter is set to
-100 iterations, but most of the times the the ICM (Iterated Conditional Modes)
+100 iterations, but most of the times the ICM (Iterated Conditional Modes)
```
jhlegarreta: 👍

doc/examples/tissue_classification.py
ShrishtiHore: But for the given example the value is same and it should be mentioned for better understandability. At the end of day it is an example tutorial and understandability is more important than flexibility for future changes.

ShrishtiHore: Plus it is only a comment mentioning the value used in between 0-0.5 so when value changes in the code the comment can also be changed and this does not make it difficult to maintain.

doc/examples/tissue_classification.py
ShrishtiHore: Thanks

doc/examples/tissue_classification.py
ShrishtiHore: Appreciate for reviewing @jhlegarreta but I still think the change is correct and I will for any maintainer to review it.

doc/examples/tissue_classification.py
ShrishtiHore: Corrected it. Thank you @jhlegarreta . But other changes are valid .

doc/examples/tissue_classification.py
jchoude: The second `the` is not absolutely necessary, but the first one (before `T1`) should be kept.

doc/examples/tissue_classification.py
jchoude: I tend to agree with @jhlegarreta . This does not make it "hard to maintain", as you say, but someone modifying this example might very easily forget to change it. I am not 100% sure that we need to specify the value, since it set directly in the line after, which is quite self explanatory.

doc/examples/tissue_classification.py
skoudoro: Can you update @ShrishtiHore? Thank you

doc/examples/tissue_classification.py
skoudoro: +1 too, should be removed. By experience, it increases easy mistakes during an update. 
`Code should be self-explanatory because documentation can be easily a lie`.

Can you update this? thank you

PR #2101, opened 2020-03-29
Minor typos and imports

ShrishtiHore:
Fixed some documentation and grammatical typos and import time in the beginning to avoid calls in between.

### DIFF
--- a/doc/examples/restore_dti.py
+++ b/doc/examples/restore_dti.py
@@ -5,7 +5,7 @@ Using the RESTORE algorithm for robust tensor fitting
 
 The diffusion tensor model takes into account certain kinds of noise (thermal),
 but not other kinds, such as "physiological" noise. For example, if a subject
-moves during the acquisition of one of the diffusion-weighted samples, this
+moves during acquisition of one of the diffusion-weighted samples, this
 might have a substantial effect on the parameters of the tensor fit calculated
 in all voxels in the brain for that subject. One of the pernicious consequences
 of this is that it can lead to wrong interpretation of group differences. For
@@ -58,6 +58,8 @@ visualizations:
 
 from dipy.viz import window, actor
 import matplotlib.pyplot as plt
+
+import dipy.denoise.noise_estimate as ne
 
 # Enables/disables interactive visualization
 interactive = False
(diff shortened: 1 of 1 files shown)


doc/examples/restore_dti.py
```diff
@@ -230,7 +230,7 @@
 closely resembles the baseline distribution of the noiseless signal, and
 demonstrates the utility of the method to data with intermittent
 noise. Importantly, this method assumes that the tensor is a good
-representation of the diffusion signal in the data. If you have reason to
+representation of diffusion signal in data. If you have reason to
```
jhlegarreta: I'd dare to say that the article is necessary/correct.

doc/examples/restore_dti.py
```diff
@@ -5,7 +5,7 @@
 
 The diffusion tensor model takes into account certain kinds of noise (thermal),
 but not other kinds, such as "physiological" noise. For example, if a subject
-moves during the acquisition of one of the diffusion-weighted samples, this
+moves during acquisition of one of the diffusion-weighted samples, this
```
jhlegarreta: I'd dare to say that the article is necessary/correct.

doc/examples/restore_dti.py
ShrishtiHore: Let the maintainers review it, If wrong then I will change it.

doc/examples/restore_dti.py
ShrishtiHore: Let the maintainers review it, If it's wrong then I will change it.

jchoude: Both versions are ok. When both cases are acceptable, we normally prefer not changing it, so that the review is quicker for the maintainers.

doc/examples/restore_dti.py
jchoude: The article is necessary.

skoudoro: Please, can you update this change @ShrishtiHore?

Otherwise, Both @jhlegarreta and @jchoude are part of the maintainers' team

doc/examples/restore_dti.py
skoudoro: +1 with @jchoude comment. Both are ok.

doc/examples/restore_dti.py
ShrishtiHore: Done. @skoudoro

PR #2102, opened 2020-03-29
Modules imported in the beginning

ShrishtiHore:
Fixed some documentation and grammatical typos and import time in the beginning to avoid calls in between. Same as previous PRs.


skoudoro:
I'm ok with this one too since there is few imports. Merging.

Thank you @ShrishtiHore

ShrishtiHore:
Thank you @skoudoro

PR #2103, opened 2020-03-29
Documentation typos, grammar corrections & imports

ShrishtiHore:
Corrected some minor typos and placed modules in the beginning to avoid calls in between code.

### DIFF
--- a/doc/examples/segment_extending_clustering_framework.py
+++ b/doc/examples/segment_extending_clustering_framework.py
@@ -4,7 +4,7 @@ Enhancing QuickBundles with different metrics and features
 ==========================================================
 
 QuickBundles [Garyfallidis12]_ is a flexible algorithm that requires only a
-distance metric and an adjacenct threshold to perform clustering. There is a
+distance metric and an adjacency threshold to perform clustering. There is a
 wide variety of metrics that could be used to cluster streamlines.
 
 The purpose of this tutorial is to show how to easily create new ``Feature``
--- a/doc/examples/segment_extending_clustering_framework.py
+++ b/doc/examples/segment_extending_clustering_framework.py
@@ -4,7 +4,7 @@ Enhancing QuickBundles with different metrics and features
 ==========================================================
 
 QuickBundles [Garyfallidis12]_ is a flexible algorithm that requires only a
-distance metric and an adjacency threshold to perform clustering. There is a
+distance metric and an adjacenct threshold to perform clustering. There is a
 wide variety of metrics that could be used to cluster streamlines.
 
 The purpose of this tutorial is to show how to easily create new ``Feature``
@@ -61,6 +61,15 @@ Let's start by importing the necessary modules.
 
 from dipy.segment.metric import Feature
 from dipy.tracking.streamline import length
+import numpy as np
+from dipy.data import get_fnames
+from dipy.io.streamline import load_tractogram
+from dipy.tracking.streamline import Streamlines
+from dipy.viz import window, actor
+from dipy.segment.clustering import QuickBundles
+from dipy.segment.metric import SumPointwiseEuclideanMetric
+from dipy.segment.metric import Metric
+from dipy.segment.metric import VectorOfEndpointsFeature
 
 """
 We now define the class ``ArcLengthFeature`` that will perform the desired
@@ -72,7 +81,6 @@ its process is invariant to the order of the points within a streamline.
 This is needed as there is no way one can tell which extremity of a
 streamline is the beginning and which one is the end.
 """
-
 
 class ArcLengthFeature(Feature):
     """ Computes the arc length of a streamline. """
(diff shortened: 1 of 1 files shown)


doc/examples/segment_extending_clustering_framework.py
```diff
@@ -4,7 +4,7 @@
 ==========================================================
 
 QuickBundles [Garyfallidis12]_ is a flexible algorithm that requires only a
-distance metric and an adjacency threshold to perform clustering. There is a
+distance metric and an adjacenct threshold to perform clustering. There is a
```
jhlegarreta: The term is accurate and matches the term used in the original paper and cross-referenced above:
https://www.frontiersin.org/articles/10.3389/fnins.2012.00175/full

ShrishtiHore: Firstly, Thank you @jhlegarreta for reviewing the PR and correcting me about "adjacency" I didn't really know the term but now I have changed it.

Secondly, I had made a PR https://github.com/dipy/dipy/pull/2094 regarding making the examples more simple and beginner friendly in a notebook manner but then @skoudoro is working on a sphinx format for the tutorials. Since these are simple python scripts and people run them directly it is good to place all imports in the beginning, plus it is just more systematic. The use of these modules is explained in the comments right where the classes are so it should not be an issue.

And lastly please stop using "I'd dare to" in every review you make cause we are all here as a community to contribute and not to challenge each other.

skoudoro:
Thank you @jhlegarreta, for your feedback. 

> I'd dare to say that the rationale of the imports being placed just before the necessary classes and methods are invoked in the examples is to show where those constructs dwell so that people have a better picture of the toolkit.

I agree but I do not have a strong opinion for that. Both are ok and I do not know the history behind this choice.

> Additionally, I'd say that the review process is eased if all grammar changes can be proposed in a single PR, as long as this is possible.

Completely agree.


it would be good to have other developer's opinions.

ShubhamShaswat:
@ShrishtiHore I personally found all imports at beginning much helpful, but have you considered adding comments to the imports so that we can know the why it has been imported at the first place,maybe one line.Just a suggestion.

doc/examples/segment_extending_clustering_framework.py
ShreyasFadnavis: Hi @ShrishtiHore ! Thank you for this PR 👍 I don't think @jhlegarreta is trying to be aggressive whatsoever. In fact, the phrase `I'd dare to say` is a rather polite way of giving constructive criticism on your PR, given that its is not his research or contribution to the code-base. 

However, I do see how this could be misconstrued as impolite and appreciate you sharing the way it made you feel. We are more than happy to have you on board and I believe we are looking at your PR's very actively! Looking forward to more contributions and engagement with the community!

ShrishtiHore: Hi@ShreyasFadnavis Thank you I totally understand. And I don't mind about it now. I would surely love to engage more with the community.

jchoude: Just a side note, because just from reading the emails threads of the notifications for some of those PRs, I have felt some friction about language and the way to address each other.

I just want everyone to remember that we all have our native language, and more often than not, it is not english. If someone sometimes feel that another member is being rude, most of the time it simply is a misunderstanding due to varying language patterns.

We're all in this together, and especially in those weird times where we are now, we should strive to be caring for each other.

jchoude:
@skoudoro I also generally agree that when changes are only related to grammar and typos, they should mostly be bunched in a single PR. For changes like here, touching examples, I think that changing more than just typos is ok.

ShrishtiHore:
Hii @jchoude and @skoudoro Thank you for your feedback. I understand now. Actually I am still learning about contributions in public repositories and didn't really know about it. I will make sure that all documentation contributions are made in a single PR hence forth.

ShrishtiHore:
@ShubhamShaswat even I feel its helpful. I will surely try adding comments on these imports then.

skoudoro:
Before going ahead with this PR, I would like more opinions about moving `import` at the top of all the examples.

Please, Team, tell me what do you think?

arokem:
It brings us closer to PEP8 compliance, so on the one hand it's a good thing. On the other hand, I believe that it will make your work in #1990 a little harder, because of new potential conflicts in rebasing. So, I would opt to finish that one first, if at all possible.

doc/examples/segment_extending_clustering_framework.py
ShrishtiHore: Hi @jchoude Thank you and I really appreciate that you took time to reflect.

ShrishtiHore:
Thank you @arokem and @skoudoro for considering this PR !!! Should I add comments on before the imports as suggested by @ShubhamShaswat  ???

skoudoro:
Thank you @ShrishtiHore

PR #2108, opened 2020-04-03
[Horizon] Update clipping range on slicer

skoudoro:
When the slicer goes up or down we need to update the scene clipping range because some of the slices are not immediately visible.

I use this opportunity to add 2 new optionq in Horizon: 
- `--bg_color 0.5 0.5 0.5` or `--bg_color 0.5`(change background color of the scene)
- `--disable_order_transparency`

### DIFF
--- a/dipy/workflows/viz.py
+++ b/dipy/workflows/viz.py
@@ -17,8 +17,8 @@ class HorizonFlow(Workflow):
     def run(self, input_files, cluster=False, cluster_thr=15.,
             random_colors=False, length_gt=0, length_lt=1000,
             clusters_gt=0, clusters_lt=10**8, native_coords=False,
-            stealth=False, emergency_header='icbm_2009a', out_dir='',
-            out_stealth_png='tmp.png'):
+            stealth=False, emergency_header='icbm_2009a', bg_color=(0, 0, 0),
+            out_dir='', out_stealth_png='tmp.png'):
         """ Interactive medical visualization - Invert the Horizon!
 
         Interact with any number of .trk, .tck or .dpy tractograms and anatomy
@@ -54,6 +54,9 @@ class HorizonFlow(Workflow):
         emergency_header : str
             If no anatomy reference is provided an emergency header is
             provided. Current options 'icbm_2009a' and 'icbm_2009c'.
+        bg_color : variable float
+            Define the background color of the scene. Value should be
+            between [0-1]. Default is black (0, 0, 0)
         out_dir : string
             Output directory. Default current directory.
         out_stealth_png : string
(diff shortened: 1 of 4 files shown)


azure-pipelines.yml
```diff
@@ -7,62 +7,62 @@ pr:
...
-        python.version: '3.7'
-        EXTRA_DEPENDS: "scikit-learn pandas statsmodels==0.9.0 pytables scipy==1.1"
-        INSTALL_TYPE: "conda"
-      CONDA Python37-64bit:
-        python.version: '3.7'
-        INSTALL_TYPE: "conda"
-      CONDA Python36-64bit:
-        python.version: '3.6'
-        INSTALL_TYPE: "conda"
+# - template: ci/azure/linux.yml
```
Garyfallidis, line 10: Why all these comments here?

Garyfallidis:
Can I ask a small favor can you add also a flag for order_transparency horizon in some cloud systems textures are not visible when order_transparency is True. We can provide the option to disable that feature.

azure-pipelines.yml
```diff
@@ -71,57 +71,57 @@ jobs:
...
+      #   python.version: '3.6'
+      #   INSTALL_TYPE: "conda"
+      # Python37-64bit + VIZ:
+      #   TEST_WITH_XVFB: "1"
+      #   python.version: '3.7'
+      #   MESA_GL_VERSION_OVERRIDE: '3.3'
+      #   LIBGL_ALWAYS_INDIRECT: 'y'
+      #   EXTRA_DEPENDS: "scikit_learn vtk fury scipy xvfbwrapper"
+
+# - template: ci/azure/windows.yml
```
Garyfallidis, line 96: Many comments added here. Is that what you want to do. Disable some bots? Please enlighten us.

azure-pipelines.yml
skoudoro: I use this PR to understand @BramshQamar issue with CI's. All this comments are just to accelerate the build process so they are temporary. I will remove them soon.

azure-pipelines.yml
skoudoro: I use this PR to understand @BramshQamar issue with CI's. All these comments are just to accelerate the build process so they are temporary. I will remove them soon.

dipy/direction/tests/test_peaks.py
```diff
@@ -531,31 +531,34 @@ def test_peaksFromModelParallel():
     data, _ = multi_tensor(gtab, mevals, S0, angles=[(0, 0), (60, 0)],
                            fractions=[50, 50], snr=SNR)
 
-    for sphere in [_sphere, default_sphere]:
+    for sphere in [_sphere, get_sphere('symmetric642')]:
 
         # test equality with/without multiprocessing
+        print("DEBUG -- 1")
```
Garyfallidis, line 537: Please remove these prints. It seems overall in this PR you are trying to solve two problems.

skoudoro:
ok for `order_transparency`. will do

dipy/workflows/viz.py
```diff
@@ -142,9 +145,15 @@ def run(self, input_files, cluster=False, cluster_thr=15.,
                     print('Peak_dirs shape')
                     print(pam.peak_dirs.shape)
 
+        if len(bg_color) == 1:
```
Garyfallidis, line 148: This will work only if bg_color is a tuple (0.5,) are you really supporting this option? Not explained in the docstring above.

skoudoro: Good catch! I will explain it in the docstring. This option is for lazy people. Instead of doing `--bg_color 0.5 0.5 0.5` you can do `--bg_color 0.5`

skoudoro: it will work, because `variable float` return always a list

skoudoro:
Hi All, 

- @arokem: I can not restart the failing matrix and I do not why. Can you restart it for me? Thank you
- @Garyfallidis: This PR is ready to go, do you any other comments?

PR #2111, opened 2020-04-06
Added an epsilon to bounding_box check

frheault:
To fix the very rare occurrence of streamlines with point closer than epsilon to the edge of the volume.
Using a parameter in case people are doing small animal. 
`epsilon=1e-6` seems to be a safe, yet effective, value for human data.


arokem:
Looks good! I think... could you please add a test that demonstrates how this works?

frheault:
@arokem it does not change the current behavior of the function. It should have been an epsilon in the first place. 
I don't think there is a 'way' to use it, its a floating-point check with an epsilon.

arokem:
How about using the key-word argument here: https://github.com/dipy/dipy/blob/master/dipy/io/tests/test_stateful_tractogram.py#L622 ? 

Presumably, there is an epsilon large enough such that the number of streamlines would change?

jchoude:
Just to make sure, you don't want to keep points directly on the edge, right?

frheault:
@jchoude Yes, I saw one or two dataset with an affine with a strong rotation and point that were sometime falling out of the bbox and some other time it was not.

@arokem I can do a test where I put a 1mm epsilon and I would loose 8 streamlines (corner) of my gold standard. That would be the only thing I can do with my gold standard

skoudoro:
Hi @frheault, Can you rebase? just to make all tests green since it has been fixed. 
Otherwise, +1 for merging

skoudoro:
Thank you @frheault!

PR #2116, opened 2020-04-08
fixed code tags and minor changes

ShubhamShaswat:


skoudoro:
LGTM, merging since it is only docstring. Thank @ShubhamShaswat

PR #2121, opened 2020-04-13
BF: ensure `btens` attribute of `GradientTable` is initialised

fnery:
Create an instance of the [GradientTable](https://github.com/dipy/dipy/blob/85738ae82223aae9af9893220665f1203ea3dd8a/dipy/core/gradients.py#L17) class without specifying the `btens` attribute:

    >>> gt = GradientTable(gradients)

Where `gradients` is a gradient table without specifying b-tensors (e.g. see [here](https://github.com/dipy/dipy/blob/85738ae82223aae9af9893220665f1203ea3dd8a/dipy/core/tests/test_gradients.py#L87)).
Query the `btens` attribute on that instance:

    >>> print(gt.btens)
    AttributeError: 'GradientTable' object has no attribute 'btens'

This error being triggered does seem to be the expected behaviour, as judging by [dipy/core/tests/test_gradients.py](https://github.com/dipy/dipy/blob/master/dipy/core/tests/test_gradients.py), specifically [this](https://github.com/dipy/dipy/blob/85738ae82223aae9af9893220665f1203ea3dd8a/dipy/core/tests/test_gradients.py#L96) line.

However, this behaviour is not consistent with other [GradientTable](https://github.com/dipy/dipy/blob/85738ae82223aae9af9893220665f1203ea3dd8a/dipy/core/gradients.py#L17) attributes. For example, making a similar query for `big_delta` does return its default value:

    >>> print(gt.big_delta)
    None

This pull requests aims to implement this behaviour. I also updated the corresponding test function accordingly.

### DIFF
--- a/dipy/core/gradients.py
+++ b/dipy/core/gradients.py
@@ -139,7 +139,7 @@ class GradientTable(object):
                                  + "strings, or an array of exact b-tensors. "
                                  + "String options: 'LTE', 'PTE', 'STE', 'CTE'")
         else:
-            self.btens = btens
+            self.btens = None
 
     @auto_attr
     def bvals(self):
--- a/dipy/core/gradients.py
+++ b/dipy/core/gradients.py
@@ -138,6 +138,8 @@ class GradientTable(object):
                                  + "Please provide a string, an array of "
                                  + "strings, or an array of exact b-tensors. "
                                  + "String options: 'LTE', 'PTE', 'STE', 'CTE'")
+        else:
+            self.btens = btens
 
     @auto_attr
     def bvals(self):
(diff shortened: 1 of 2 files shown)


dipy/core/gradients.py
```diff
@@ -138,6 +138,8 @@ def __init__(self, gradients, big_delta=None, small_delta=None,
                                  + "Please provide a string, an array of "
                                  + "strings, or an array of exact b-tensors. "
                                  + "String options: 'LTE', 'PTE', 'STE', 'CTE'")
+        else:
+            self.btens = btens
```
arokem, line 142: Maybe be explicit here?
```suggestion
            self.btens = None
```

fnery: Thanks @arokem, updated accordingly.

skoudoro:
thank you @fnery!

PR #2134, opened 2020-04-20
Slicing, adding function of StatefulTractogram

frheault:
Thanks to new code from @MarcCote in Nibabel, interaction with StatefulTractogram are now much safer for slicing and/or adding.

Code such as this is now possible, it makes sure the spatial attributes are the same, that the sft share the same space, origin and have the same keys for data_per_*. Then it basically concatenates everything creating a new sft.
`new_sft = sft_cc[0:2000] + sft_pt[0:2000] + sft_af[0:2000]`

These operations are strict, in line with everything else in the StatefulTractogram.
For example an operation like this is not possible (inconsistent space):
```
sft_1.to_vox()
sft_2.to_rasmm()
sft_3 = sft_1 + sft_2 
```

Also important to know, to avoid side effects, like the rest of StatefulTractogram, nothing is done inplace. For example:
```
sft.to_rasmm()
sft_subset = sft[0:25]
sft.to_vox()
```
sft_subset is still in rasmm and has not be modified by the operation on sft.

Bonus:  `if sft_1 == sft_2:` now works
Implementing this made the test a lot easier. But it is a bit useless in a real-life scenario.
Everything has to be identical (header, space, origin) or all_close (streamlines coordinates, data_per_point, data_per_streamline)

### DIFF
--- a/dipy/io/stateful_tractogram.py
+++ b/dipy/io/stateful_tractogram.py
@@ -184,15 +184,65 @@ class StatefulTractogram(object):
         text += '\nstreamline_count: {}'.format(self._get_streamline_count())
         text += '\npoint_count: {}'.format(self._get_point_count())
         text += '\ndata_per_streamline keys: {}'.format(
-            self.data_per_point.keys())
+            self.get_data_per_streamline_keys())
         text += '\ndata_per_point keys: {}'.format(
-            self.data_per_streamline.keys())
+            self.get_data_per_point_keys())
 
         return text
 
     def __len__(self):
         """ Define the length of the object """
         return self._get_streamline_count()
+
+    def __getitem__(self, key):
+        """ Slice all data in a consistent way """
+        if isinstance(key, int):
+            key = [key]
+
+        return self.from_sft(self.streamlines[key], self,
+                             data_per_point=self.data_per_point[key],
+                             data_per_streamline=self.data_per_streamline[key])
+
+    def __add__(self, other_sft):
+        """ Addition of two StatefulTractogram in a way that is consistent
+        with space, origin, data_per_point and data_per_streamline. """
+        if self.space != other_sft.space:
+            raise ValueError('Inconsistent space between both sft.\n'
+                             'Switch space of one or both sft before addition.')
+
+        if self.origin != other_sft.origin:
+            raise ValueError('Inconsistent origin between both sft.\n'
+                             'Switch origin of one or both sft before addition.')
+
+        if self.get_data_per_point_keys() != other_sft.get_data_per_point_keys():
+            raise ValueError('Inconsistent data_per_point between both sft.\n'
+                             'Either delete it from one or generate it for the '
+                             'other.')
+        if self.get_data_per_streamline_keys() != other_sft.get_data_per_streamline_keys():
+            raise ValueError('Inconsistent data_per_streamline between both '
+                             'sft.\nEither delete it from one or generate it '
+                             'for the other.')
+
+        data_per_point = deepcopy(self.data_per_point)
+        for key in other_sft.data_per_point:
+            data_per_point._extend_entry(key,
+                                         other_sft.data_per_point[key])
+
+        data_per_streamline = deepcopy(self.data_per_streamline)
+        for key in other_sft.data_per_streamline:
+            data_per_streamline._extend_entry(key,
+                                              other_sft.data_per_streamline[key])
+
+        streamlines = deepcopy(self.streamlines)
+        streamlines.extend(other_sft.streamlines)
+
+        return self.from_sft(streamlines, self,
+                             data_per_point=data_per_point,
+                             data_per_streamline=data_per_streamline)
+
+    def __iadd__(self, other):
+        self.value = self + other
+        return self.value
 
     @property
     def space_attributes(self):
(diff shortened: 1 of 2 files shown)


frheault:
@jchoude if you want to look for inconsistency, typo or have a special request.

skoudoro:
Hi @frheault,

Thank you for this! Do you plan to add the other's operations on this PR? ("__sub__", "__mul__", "__mod__", "__pow__", "__floordiv__", "__truediv__", "__lshift__", "__rshift__", "__or__", "__and__", "__xor__", "__lt__", "__le__", "__gt__", "__ge__").

Maybe you should just keep slicing on this PR and create a new PR for the operators.

I will try your PR later in the day. Thank you

frheault:
Considering the complexity of the objects (streamlines, dictionary of attributes per point and per streamlines) I believe none of these functions could be implemented (even if we wanted to).

E.G: What is `sft ** 2` and what does it represent? or what does `sft_1 / sft_2` mean?
I think the addition is the only meaningful operator (`==` makes some sense, but as mentioned it is quite useless in real-life).

skoudoro:
Globally, it makes sense. Than you for the explanation.

Before adding the whole operator list, I was wondering why having `+ (add)` and not `- (sub)` since they have the exact same meaning. 
```
>>> new_sft = sft_cc[0:2000] + sft_pt[0:2000]
>>> my_custom_preprocess  # random operations
>>> new_sft = new_sft - sft_pt[500:1000]) 
```
I think you can add `sub` but I agree that the others are not enough meaningful.

frheault:
Well, adding is easy since it is a concatenation (append at the end or `extend`)
Subtraction implies finding the identical streamlines first, which is `O(n²)` unless we add a good 1000 lines of code for a hashing approach (which would be `O(n*log(n))` at best)

mdesco:
Good and very useful stuff. Thanks @frheault and @MarcCote. This is important stuff.

dipy/io/stateful_tractogram.py
```diff
@@ -138,6 +140,33 @@ def __init__(self, streamlines, reference, space,
         self._origin = origin
         logger.debug(self)
 
+    @staticmethod
+    def is_compatible_sft(sft_1, sft_2):
+        """ Compatibility verication of two StatefulTractogram to ensure space,
```
skoudoro, line 145: typo:  verification.

dipy/io/stateful_tractogram.py
```diff
@@ -184,16 +213,77 @@ def __str__(self):
...
+        if not self.is_compatible_sft(self, other):
+            return False
+
+        streamlines_equal = np.allclose(self.streamlines.get_data(),
+                                        other.streamlines.get_data())
+        dpp_equal = True
+        for key in self.data_per_point:
+            dpp_equal = dpp_equal and np.allclose(
+                self.data_per_point[key].get_data(),
+
```
skoudoro, line 246: disturbing empty line

dipy/io/stateful_tractogram.py
```diff
@@ -184,16 +213,77 @@ def __str__(self):
...
+                other.data_per_point[key].get_data())
+        dps_equal = True
+        for key in self.data_per_streamline:
+            dps_equal = dps_equal and np.allclose(
+                self.data_per_streamline[key],
+                other.data_per_streamline[key])
+
+        return streamlines_equal and dpp_equal and dps_equal
```
skoudoro, line 254: Nice! I like this! However, I wonder why should we continue if `streamlines_equal` is False. it avoids us to go through the 2 others loop.  Same comment if `dpp_equal` is False.

So I think, you skip or quit the function as soon as one test is False

dipy/io/stateful_tractogram.py
```diff
@@ -184,16 +213,77 @@ def __str__(self):
...
+        """ Addition of two sft with attributes consistency checks """
+        if not self.is_compatible_sft(self, other_sft):
+            logger.debug(self)
+            logger.debug(other_sft)
+            raise ValueError('Inconsistent StatefulTractogram.\n'
+                             'Make sure Space, Origin are the same and that '
+                             'data_per_point and data_per_streamline keys are '
+                             'the same.')
+
+        streamlines = deepcopy(self.streamlines)
```
skoudoro, line 270: self.streamlines.copy() instead ?

skoudoro:
Thank you for the update @frheault. good to go for me

Any opinion @arokem, @arokem, @Garyfallidis, or others before merging this? waiting for another +1 or review.

dipy/io/stateful_tractogram.py
```diff
@@ -138,6 +140,35 @@ def __init__(self, streamlines, reference, space,
         self._origin = origin
         logger.debug(self)
 
+    @staticmethod
```
MarcCote, line 143: Any particular reason why this is a staticmethod instead of method? I see later you are calling it like this `self.is_compatible_sft` anyway. Also, if it stays a staticmethod, I think `are_compatible` would be a more accurate name for (or `is_compatible` if you make it a normal method).

frheault: I have to call it self.is_compatible_sft since it is defined in that file. But elsewhere it is going to be:
```
from dipy.io.stateful_tractogram import StatefulTractogram
StatefulTractogram.are_compatible(sft_1, sft_2)
```
(I changed the name)

skoudoro:
Thank you @frheault! nice job!

PR #2137, opened 2020-04-21
Deep-learning-based registration example

sarathknv:
A simple example of deep-learning-based non-rigid image registration with the MNIST dataset, taking ideas from

- [An Unsupervised Learning Model for Deformable Medical Image Registration](http://openaccess.thecvf.com/content_cvpr_2018/papers/Balakrishnan_An_Unsupervised_Learning_CVPR_2018_paper.pdf)
- [Non-rigid image registration using fully convolutional networks with deep self-supervision](https://arxiv.org/pdf/1709.00799.pdf)

#### Framework
framework

An encoder-decoder network is trained to output a dense displacement field that is used to warp the moving image to align with the fixed image. I've used MNIST images of one class, same as that of the fixed image, for training and testing.

#### Colab notebook
https://colab.research.google.com/drive/1fTzz1aT2sb8oAXRO1-dr6O_IR6dof36e

Training takes about 5 minutes on GPU.

### DIFF
--- a/dipy/nn/registration.py
+++ b/dipy/nn/registration.py
@@ -491,7 +491,7 @@ def main():
                                      'with MNIST dataset')
     parser.add_argument('--batch-size', type=int, default=8,
                         help='batch size (default: 8)')
-    parser.add_argument('--epochs', type=int, default=1,
+    parser.add_argument('--epochs', type=int, default=50,
                         help='number of epochs to train (default: 50)')
     parser.add_argument('--lr', type=float, default=0.004,
                         help='learning rate (default: 0.004)')
--- a/dipy/nn/registration.py
+++ b/dipy/nn/registration.py
@@ -1,0 +1,561 @@
+"""A simple example for deep-learning-based non-rigid image registration
+with the MNIST dataset.
+"""
+import tensorflow as tf
+import tensorflow.keras.layers as layers
+import numpy as np
+import matplotlib.pyplot as plt
+import argparse
+
+
+@tf.function
+def mse_loss(static, moving):
+    """Computes the mean squared error (MSE) loss.
+
+    Currently, only 4-D inputs are supported.
+
+    Parameters
+    ----------
+    static : tf.Tensor, shape (N, H, W, C)
+        The static image to which the moving image is aligned.
+    moving : tf.Tensor, shape (N, H, W, C)
+        The moving image, the same shape as the static image.
+
+    Returns
+    -------
+    loss : tf.Tensor, shape ()
+        Mean squared error between the static and the moving images,
+        averaged over the batch.
+    """
+    loss = tf.reduce_mean(tf.square(moving - static))  # shape ()
+    return loss
+
+
+@tf.function
+def ncc_loss(static, moving):
+    """Computes the normalized cross-correlation (NCC) loss.
+
+    Currently, only 4-D inputs are supported.
+
+    Parameters
+    ----------
+    static : tf.Tensor, shape (N, H, W, C)
+        The static image to which the moving image is aligned.
+    moving : tf.Tensor, shape (N, H, W, C)
+        The moving image, the same shape as the static image.
+
+    Returns
+    -------
+    loss : tf.Tensor, shape ()
+        Normalized cross-correlation loss between the static and the
+        moving images, averaged over the batch. Range is [-1.0, 1.0].
+        The best value is -1 (perfect match) and the worst is 1.
+
+    References
+    ----------
+    .. [1] `Wikipedia entry for the Cross-correlation
+           <https://en.wikipedia.org/wiki/Cross-correlation>`_
+    """
+    eps = tf.constant(1e-9, 'float32')
+
+    static_mean = tf.reduce_mean(static, axis=[1, 2], keepdims=True)
+    moving_mean = tf.reduce_mean(moving, axis=[1, 2], keepdims=True)
+    # shape (N, 1, 1, C)
+
+    static_std = tf.math.reduce_std(static, axis=[1, 2], keepdims=True)
+    moving_std = tf.math.reduce_std(moving, axis=[1, 2], keepdims=True)
+    # shape (N, 1, 1, C)
+
+    static_hat = (static - static_mean)/(static_std + eps)
+    moving_hat = (moving - moving_mean)/(moving_std + eps)
+    # shape (N, H, W, C)
+
+    ncc = tf.reduce_mean(static_hat * moving_hat)  # shape ()
+    loss = -ncc
+    return loss
+
+
+def simple_cnn(input_shape=(32, 32, 2)):
+    """Creates a 2-D convolutional encoder-decoder network.
+
+    Parameters
+    ----------
+    input_shape : sequence of ints, optional
+        Input data shape of the form (H, W, C). Default is (32, 32, 2).
+
+    Returns
+    -------
+    model
+        An instance of Keras' Model class.
+
+    Notes
+    -----
+    Given a concatenated pair of static and moving images as input, the
+    CNN computes a dense displacement field that is used to warp the
+    moving image to match with the static image.
+
+    The number of channels in the output (displacement field) is equal
+    to the dimensionality of the input data. For 3-D volumes, it is 3,
+    and for 2-D images, it is 2. The first channel comprises
+    displacement in the x-direction and the second comprises
+    displacement in the y-direction.
+    """
+    out_channels = 2
+    inputs = layers.Input(shape=input_shape)
+
+    # encoder
+    x = layers.Conv2D(32, kernel_size=3, strides=2, padding='same',
+                      activation='relu')(inputs)            # 32 --> 16
+    x = layers.BatchNormalization()(x)                      # 16
+    x = layers.Conv2D(32, kernel_size=3, strides=1, padding='same',
+                      activation='relu')(x)                 # 16
+    x = layers.BatchNormalization()(x)                      # 16
+    x = layers.MaxPool2D(pool_size=2)(x)                    # 16 --> 8
+    x = layers.Conv2D(64, kernel_size=3, strides=1, padding='same',
+                      activation='relu')(x)                 # 8
+    x = layers.BatchNormalization()(x)                      # 8
+    x = layers.Conv2D(64, kernel_size=3, strides=1, padding='same',
+                      activation='relu')(x)                 # 8
+    x = layers.BatchNormalization()(x)                      # 8
+    x = layers.MaxPool2D(pool_size=2)(x)                    # 8 --> 4
+    x = layers.Conv2D(128, kernel_size=3, strides=1, padding='same',
+                      activation='relu')(x)                 # 4
+    x = layers.BatchNormalization()(x)                      # 4
+
+    # decoder
+    x = layers.Conv2DTranspose(64, kernel_size=2, strides=2,
+                               padding='same')(x)           # 4 --> 8
+    x = layers.Conv2D(64, kernel_size=3, strides=1, padding='same',
+                      activation='relu')(x)                 # 8
+    x = layers.BatchNormalization()(x)                      # 8
+    x = layers.Conv2DTranspose(32, kernel_size=2, strides=2,
+                               padding='same')(x)           # 8 --> 16
+    x = layers.Conv2D(32, kernel_size=3, strides=1, padding='same',
+                      activation='relu')(x)                 # 16
+    x = layers.BatchNormalization()(x)                      # 16
+    x = layers.Conv2DTranspose(16, kernel_size=2, strides=2,
+                               padding='same')(x)           # 16 --> 32
+    x = layers.Conv2D(16, kernel_size=3, strides=1, padding='same',
+                      activation='relu')(x)                 # 32
+    x = layers.BatchNormalization()(x)                      # 32
+    x = layers.Conv2D(out_channels, kernel_size=1, strides=1,
+                      padding='same')(x)                    # 32
+
+    # Create the model.
+    model = tf.keras.Model(inputs, x, name='simple_cnn')
+    return model
+
+
+@tf.function
+def grid_sample(moving, grid):
+    """Given a moving image and a sampling grid as input, computes the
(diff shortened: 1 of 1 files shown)


sarathknv:
I've refactored my code a little. @BramshQamar @skoudoro @ShreyasFadnavis @Garyfallidis your thoughts on this. 

Usage:
```
from dipy.nn.registration import SimpleFCN2d
from dipy.nn.metrics import NormalizedCrossCorrelation, MeanSquaredError


ncc = NormalizedCrossCorrelation()

model = SimpleFCN2d()
model.compile(optimizer='adam', loss=ncc)

# train
model.fit(x_train, y_train, validation_split=0.2, batch_size=8, epochs=10)

# evaluate on test set
model.evaluate(x_test, y_test, batch_size=8)

# save weights
model.save_weights('./weights_fcn2d.h5')

# load_weights
model.load_weights('weights_fcn2d.h5')

# for single/batch input
moved = model.predict_on_batch({'moving': moving, 'static': static})
```
If this API is okay, I'll go ahead and clean it and add the test file.

ShreyasFadnavis:
Hi @sarathknv ... I quite like this API design. Would like some other people's opinions on this!

@skoudoro did you get a chance to see if we can unify the metrics that we already have in DIPY with TF?

@finalelement can you take a look too? Does this abide by your PR? Any other stuff you'd like the API to have/ we missed out on?

skoudoro:
Hi @sarathknv,

Thank you for this update! 

1. Personally, I prefer when the `compile` method is in the `__init__` of your class like `MultipleLayersPerceptron` class. I do not think the user needs this extra work. However, I like that your `loss` parameter is a callable. I think we should keep this point.

2. Concerning `load_weights/save_weights`, I think it is good, and you should keep these methods.

3. I am not sure what is the difference between `predict` and `predict_on_batch`. Can you tell me more? Can it just be an option of `predict`? 

4. Concerning the metric. Do you think you will add more methods to these classes? If not, we do not need a class and you can keep them as a function with a decorator. A class with only `__init__` and `__call__` methods is a sign that it can be simpler.

5. There a lot of code.... too much code... it will be hard to review. Please, consider 1 PR = 1 model. You should avoid adding 3-4 models on the same PR. Let's focus on 1 model and make sure that we all agree with the API.

Thank you

sarathknv:
Thanks for the feedback @ShreyasFadnavis and @skoudoro :relaxed:

> 1. Personally, I prefer when the compile method is in the __init__ of your class like MultipleLayersPerceptron class. I do not think the user needs this extra work. However, I like that your loss parameter is a callable. I think we should keep this point.

I agree with the idea of keeping the API simple. But I also think that it's good to have the `compile` method accessible to the user. If we want to change the parameters of loss or optimizer, like the learning rate, or maybe the optimizer itself, we can just call the `compile` method to update them in the model, without affecting the model weights. I generally try a few learning rates on sample data before running the full-scale experiment, to get a feel of how long it would take to train. So I thought it would be nice to have it as a method. :relaxed:

We could do something like this

```
class CNN(object):
    def __init__(self, loss=None, optimizer='adam')
        self.model.compile(loss=loss, optimizer=optimizer)

    def compile(self, optimizer='adam', loss=None):
        self.model.compile(optimizer=optimizer, loss=loss)
```
  


> 2. Concerning load_weights/save_weights, I think it is good, and you should keep these methods.

:+1: 


> 3. I am not sure what is the difference between predict and predict_on_batch. Can you tell me more? Can it just be an option of predict?

`predict` is for inputs with many samples. It internally processes the data batch by batch. For inputs with few samples `predict` has overhead and is slower than `predict_on_batch`. This [StackOverflow thread](https://stackoverflow.com/a/61287324) clears up the differences and also has some benchmarks.

From the [TensorFlow docs](https://www.tensorflow.org/api_docs/python/tf/keras/Model?version=nightly#predict)

> Computation is done in batches. This method is designed for performance in large scale inputs. For small amount of inputs that fit in one batch, directly using __call__ is recommended for faster execution, e.g., model(x), or model(x, training=False)

If you think `predict` and `predict_on_batch` would be confusing to the user, we can have a `__call__` method calling the `predict_on_batch` method. Then inference usage would be simple
```
model = CNN()
moved = model({'moving': moving, 'static': static})
```


> 4. Concerning the metric. Do you think you will add more methods to these classes? If not, we do not need a class and you can keep them as a function with a decorator. A class with only __init__ and __call__ methods is a sign that it can be simpler.

Changed the metrics from classes to functions with decorators. :+1: 


> 5. There a lot of code.... too much code... it will be hard to review. Please, consider 1 PR = 1 model. You should avoid adding 3-4 models on the same PR. Let's focus on 1 model and make sure that we all agree with the API.

Removed VoxelMorph models. :+1:

skoudoro:
> We could do something like this
   ```python
class CNN(object):
    def __init__(self, loss=None, optimizer='adam')
        self.model.compile(loss=loss, optimizer=optimizer)

    def compile(self, optimizer='adam', loss=None):
        self.model.compile(optimizer=optimizer, loss=loss)
```

ok, a great hybrid solution, I am fine with that 👍.

> predict is for inputs with many samples. It internally processes the data batch by batch. For inputs with few samples predict has overhead and is slower than predict_on_batch. This StackOverflow thread clears up the differences and also has some benchmarks.

Ok, thanks for the reference, I will read more about it.

> If you think predict and predict_on_batch would be confusing to the user, we can have a __call__ method calling the predict_on_batch method. Then inference usage would be simple

Yes, I think they are confusing. It needs more thinking and we need to be explicit so I would not recommend using `__call__` here.

> Removed VoxelMorph models. 👍

Thank you for the cleaning!

File: dipy/nn/registration/_utils.py [hunk 1]
```diff
@@ -0,0 +1,82 @@
...
+
+tf, have_tf, _ = optional_package("tensorflow")
+layers, _, _ = optional_package("tensorflow.keras.layers")
+
+if have_tf:
+    if LooseVersion(tf.__version__) < LooseVersion('2.0.0'):
+        raise ImportError('Please upgrade to TensorFlow 2+')
+
+
+def regular_grid_2d(height, width):
```
BramshQamar, line 12: Missing docstring

dipy/nn/registration/_utils.py
```diff
@@ -0,0 +1,82 @@
...
+    y = tf.linspace(-1.0, 1.0, height)  # shape (H, )
+
+    X, Y = tf.meshgrid(x, y)  # shape (H, W), both X and Y
+
+    grid = tf.stack([X, Y], axis=-1)
+
+    return grid
+
+
+def grid_sample_2d(moving, grid):
```
BramshQamar, line 23: Missing docstring

dipy/nn/registration/fcn.py
```diff
@@ -0,0 +1,147 @@
...
+
+if have_tf:
+    if LooseVersion(tf.__version__) < LooseVersion('2.0.0'):
+        raise ImportError('Please upgrade to TensorFlow 2+')
+
+
+__all__ = ["SimpleFCN2d"]
+
+
+class SimpleFCN2d(object):
```
BramshQamar, line 16: All functions in this class missing docstring

BramshQamar:
> I've refactored my code a little. @BramshQamar @skoudoro @ShreyasFadnavis @Garyfallidis your thoughts on this.
> 
> Usage:
> 
> ```
> from dipy.nn.registration import SimpleFCN2d
> from dipy.nn.metrics import NormalizedCrossCorrelation, MeanSquaredError
> 
> 
> ncc = NormalizedCrossCorrelation()
> 
> model = SimpleFCN2d()
> model.compile(optimizer='adam', loss=ncc)
> 
> # train
> model.fit(x_train, y_train, validation_split=0.2, batch_size=8, epochs=10)
> 
> # evaluate on test set
> model.evaluate(x_test, y_test, batch_size=8)
> 
> # save weights
> model.save_weights('./weights_fcn2d.h5')
> 
> # load_weights
> model.load_weights('weights_fcn2d.h5')
> 
> # for single/batch input
> moved = model.predict_on_batch({'moving': moving, 'static': static})
> ```
> 
> If this API is okay, I'll go ahead and clean it and add the test file.

It would be useful to have documentation/tutorial on how to use the API. Something similar to the usage example you have provided here.

Here are the existing tutorials in [ DIPY](https://github.com/dipy/dipy/tree/master/doc/examples).

Tutorial can be added later in a different PR.

finalelement:
@ShreyasFadnavis Thanks for the tag. Will check out the code.

doc/examples/nn_deformable_registration_2d.py
```diff
@@ -0,0 +1,272 @@
...
+for i in range(nb):
+    for j in range(3):
+        ax = fig.add_subplot(nb, 3, i * 3 + j + 1)
+        if i == 0:
+            ax.set_title(titles_list[j], fontsize=20)
+        ax.set_axis_off()
+        ax.imshow(images_list[j][i], cmap='gray')
+plt.tight_layout()
+plt.show()
+plt.savefig('sample_results.png')
```
arokem, line 227: This figure is empty for me when I run this.

arokem: Maybe use `fig.savefig` instead?

doc/examples/nn_deformable_registration_2d.py
```diff
@@ -0,0 +1,272 @@
...
+        if self.shuffle:
+            np.random.shuffle(self.x)
+
+
+"""
+Let's define some hyperparameters used for training the network.
+"""
+
+batch_size = 32
+epochs = 100
```
arokem, line 116: It looks like performance converges after much fewer epochs. Maybe set this to 25 for the purpose of the example and then add a comment that this could be increased to a larger number for more accurate results?

Garyfallidis:
@sarathknv please enable automatic trailing whitespace removal in your IDE. It will help you resolve pep8 issues faster.

Garyfallidis:
@sarathknv, @BramshQamar, @ShreyasFadnavis, @skoudoro what is the progress here? Does this need a rebase?

dipy/nn/registration/fcn.py
```diff
@@ -10,19 +10,18 @@
...
 
 
-__all__ = ["SimpleFCN2d"]
+__all__ = ["FCN2d"]
 
 
-class SimpleFCN2d(object):
-    def __init__(self, input_shape=(32, 32)):
-        in_channels = 1
+class FCN2d(object):
```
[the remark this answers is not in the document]
sarathknv: Really sorry for the late reply. I can be active from now on.

This was some custom architecture that I wrote for the POC. I’ve added a standard model ([U-Net](https://arxiv.org/abs/1505.04597)) and will be removing fcn.py.

[dipy/nn/registration/unet.py](https://github.com/dipy/dipy/pull/2137/files#diff-b90df81deaadaf5daab47f409163c79c13ac6e03cd04a7f7622b030a89f2947c)

doc/examples/nn_deformable_registration_2d.py
```diff
@@ -0,0 +1,272 @@
...
+                        size=num_samples)
+x_sample = x_test[idxs].copy()  # shape (num_samples, 32, 32)
+
+"""
+Implement a data loader that fetches and preprocesses batches of images for 
+real-time data feeding to our model. 
+"""
+
+
+class DataLoader(tf.keras.utils.Sequence):
```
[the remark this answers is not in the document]
sarathknv: It’s definitely useful, but we need to handle several cases if we are planning to have this in the library. 
Current location of the class [dipy/nn/registration/dataloader.py
](https://github.com/dipy/dipy/pull/2137/files#diff-71543d23c03c8f0a98683495f26868a66696e78cb0a61b8b4b922028bb8ca91c)
The function of the dataloader object is to feed batches of processed data to the model. The user needs to pass the entire dataset as input either as arrays (shape (num_images, H, W, …, num_channels))  or as a list of file paths, and the dataloader object performs some preprocessing (data transformations, adding dimensions, changing data types) on batches of data and feeds it to the model. Currently, this code only supports NumPy arrays as input, which will work for small datasets that can be loaded into the memory, but a general use case would be file paths as input.

Some cases that we need to handle:
1. NumPy arrays as input
    1. Single static image and multiple moving images - All the moving images will be registered to this static image (moving shape (N, H, W, ..., C), static shape (H, W, ..., C))
    2. Multiple static and moving images - every moving image has a different static image (moving shape (N, H, W, C), static shape (N, H, W, C))
2. Lists of file paths as input
    1. Single static image and multiple moving images
    2. Multiple static and moving images
    3. Different file extensions .npy, .h5, .nii, …

doc/examples/nn_deformable_registration_2d.py
```diff
@@ -0,0 +1,272 @@
...
+
+moved = model.predict(sample_loader)  # shape (num_samples, 32, 32, 1)
+# Note: 'predict' returns outputs as numpy arrays.
+
+"""
+Finally, let's visualize the transformed image together with the static and 
+the moving images.
+"""
+
+moved = moved.squeeze(axis=-1)  # Remove the channel dim.
```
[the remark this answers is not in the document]
sarathknv: This is only needed for single-channel images. Since most images are single-channel I also think it is useful to implement as library code. Do you mean something like this? Add the squeeze option in the predict method

```python
moved = model.predict(data, squeeze=True)
```

Paluck1Arora2:
Can anyone please provide the dataset of this code "https://github.com/sarathknv/voxelmorph"
data = np.load('drive/My Drive/DIPY/brain_affine_128.npy')
static = np.load('drive/My Drive/DIPY/static.npy')
These lines are mention in voxelmorph.ipynb file

data = np.load('data/t1_moving_128.npy')
static = np.load('data/t1_static_128.npy')
These lines are mention in main.py

jhlegarreta:
https://github.com/dipy/dipy/pull/2137#issuecomment-1374726556 @Paluck1Arora2 Please post questions in DIPY's gitter room, mailing list or the DIPY wiki:
https://dipy.org/support/

Also, please avoid asking the same question as a comment and opening an issue at the same time:
https://github.com/dipy/dipy/issues/2708

Note that issues should primarily be opened to either raise attention about bug or to propose/ask about new features.

Thanks.

PR #2140, opened 2020-04-21
BF: Fix NumPy warning when creating arrays from ragged sequences

jhlegarreta:
Fix NumPy warning when creating arrays from ragged sequences in streamline test.

Fixes

```
(which is a list-or-tuple of lists-or-tuples-or ndarrays with different
lengths or shapes) is deprecated. If you meant to do this, you must specify
'dtype=object' when creating the ndarray
orig_streamlines_world = list(np.subtract(new_streamlines_world, disps))
```

triggered in
https://travis-ci.org/github/dipy/dipy/jobs/676101427#L5569
and
https://travis-ci.org/github/dipy/dipy/jobs/676764173#L5561

for example.

Rationale:
https://numpy.org/neps/nep-0034-infer-dtype-is-object.html


jhlegarreta:
I left this behind in PR #2133. I am sorry. I think this PR fixes the remaining warnings of that kind.

dipy/tracking/tests/test_streamline.py
```diff
@@ -550,7 +550,8 @@ def test_deform_streamlines():
                                                   stream2world)
 
     # Subtract disps from new_streamlines in world space
-    orig_streamlines_world = list(np.subtract(new_streamlines_world, disps))
+    orig_streamlines_world = list(np.subtract(new_streamlines_world, disps,
+                                              dtype=list))
```
skoudoro, line 554: I do not understand why `dtype=list` and not `dtype=object`?  Can you tell me more?

jhlegarreta: If I read well, both operands to `subtract` are lists in this case, and `subtract` takes array-like arguments, which includes lists.

skoudoro: - `new_streamlines_world` is a `Streamlines` object / class.  it can be a list of `ndarray` from the old API but not recommended
- `np.subtract` returns a `ndarray`
- `disps` is a `ndarray` or a list

it is quite generic so my preference is for `object`. Any other opinions are welcome!

jhlegarreta: OK. Agreed Serge. Push forced in 58c0caa.

skoudoro:
LGTM, Thank you @jhlegarreta, merging!

PR #2143, opened 2020-04-24
Build template CLI

noobiecoder1942:


arokem:
Just a comment here that a lot of this looks like a duplication of what I am trying to do in https://github.com/dipy/dipy/pull/2025. In general, I think that workflow code should be as 'thin' as possible, with as much functionality deferred to library code.

skoudoro:
Could you have a look at this @drombas and @BramshQamar? It seems related to your current work. 

Please, let us know what you think about this PR and what is the next step.

Thank you!

skoudoro:
- This is an old work and I am not sure if it is still relevant.
- No feedback and I do not think someone will continue this work.

Closing this PR for now. Feel free to reopen it !

PR #2146, opened 2020-04-24
Cluster threshold parameter in dipy_buan_shapes workflow

BramshQamar:
Added cluster threshold (clust_thr) as an optional argument in dipy_buan_shapes workflow. 
Updated default clust_thr values as well.

This small update makes the workflow faster.

### DIFF
--- a/dipy/segment/bundles.py
+++ b/dipy/segment/bundles.py
@@ -79,7 +79,7 @@ def bundle_adjacency(dtracks0, dtracks1, threshold):
     return res
 
 
-def ba_analysis(recognized_bundle, expert_bundle, nb_pts=20, threshold=5.):
+def ba_analysis(recognized_bundle, expert_bundle, nb_pts=20, threshold=6.):
     """ Calculates bundle adjacency score between two given bundles
 
     Parameters
@@ -91,7 +91,7 @@ def ba_analysis(recognized_bundle, expert_bundle, nb_pts=20, threshold=5.):
         from inout tractogram
     nb_pts : integer (default 20)
         Discretizing streamlines to have nb_pts number of points
-    threshold : float (default 5)
+    threshold : float (default 6)
         Threshold used for in computing bundle adjacency. Threshold controls
         how much strictness user wants while calculating bundle adjacency
         between two bundles. Smaller threshold means bundles should be strictly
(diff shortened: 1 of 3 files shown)


dipy/segment/bundles.py
```diff
@@ -162,7 +163,9 @@ def bundle_shape_similarity(bundle1, bundle2, rng, threshold=5):
     bundle2 : Streamlines
         White matter tract from another subject (eg: AF_L)
     rng : RandomState
-    threshold : float (default 5)
+    clust_thr : float (default [5, 3, 1.5])
```
skoudoro, line 166: list of float instead of float

dipy/workflows/stats.py
```diff
@@ -527,9 +528,13 @@ def run(self, subject_folder, threshold=6, out_dir=''):
             Path to the input subject folder. This path may contain
             wildcards to process multiple inputs at once.
 
+        clust_thr : string (default [5,3,1.5]), optional
```
skoudoro, line 531: `variable float` instead of `string`

dipy/workflows/stats.py
```diff
@@ -543,6 +548,11 @@ def run(self, subject_folder, threshold=6, out_dir=''):
 
         """
         rng = np.random.RandomState()
+
+        clust_thr = list(map(float, clust_thr.strip('[]').split(',')))
+        #clust_thr=[5,3,1.5]
+        print(clust_thr)
```
skoudoro, line 554: remove the print

dipy/workflows/stats.py
```diff
@@ -543,6 +548,11 @@ def run(self, subject_folder, threshold=6, out_dir=''):
 
         """
         rng = np.random.RandomState()
+
+        clust_thr = list(map(float, clust_thr.strip('[]').split(',')))
```
skoudoro, line 552: you do not need that if you use `variable float`

BramshQamar: but it is a list
if I just write float
it gives error

BramshQamar: let me try one more time

skoudoro: `variable float` not just `float` and you will get your list.

dipy/workflows/stats.py
BramshQamar: should I call workflow like this:
`dipy_buan_shapes subjects_small/ --clust_thr [5,3,1.25] --out_dir "sm_test"`

or

`dipy_buan_shapes subjects_small/ --clust_thr "[5,3,1.25]" --out_dir "sm_test"`

both are giving me an error

BramshQamar: got it, like this:

`dipy_buan_shapes subjects_small/ --clust_thr 5 3 1.25 --out_dir "sm_test"`

Thank you @skoudoro

Garyfallidis:
Thank you @BramshQamar !

PR #2148, opened 2020-04-26
BF: Fix `distutils` Python version requirement option warning

jhlegarreta:
Fix `distutils` Python version requirement option warning.

Fixes
```
/opt/python/3.7.1/lib/python3.7/distutils/dist.py:274: UserWarning: Unknown
distribution option: 'python_requires'

  warnings.warn(msg)
```

raised for example in
https://dev.azure.com/dipy/dipy/_build/results?buildId=343&view=logs&j=d9aefae6-c332-51b5-d933-a85009f5a2ac&t=10ba5eab-67e4-5f58-2aec-60b4dbb37e45&l=1948

or
https://travis-ci.org/github/dipy/dipy/jobs/679159532
(warning in raw log: https://api.travis-ci.org/v3/job/679159532/log.txt)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,7 +55,8 @@ if using_setuptools:
         zip_safe=False,
         extras_require=dict(
             doc=['Sphinx>=1.0'],
-            test=['pytest']))
+            test=['pytest']),
+        python_requires=">= 3.5")
 
 # Define extensions
 EXTS = []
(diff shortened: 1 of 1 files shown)


jhlegarreta:
Not sure if using `distutils` instead of `setuptools` is intended in some CI builds to test it or if it is a(n undesired) consequence of [this import](https://github.com/dipy/dipy/blob/master/setup.py#L31); same applies to the `Extension` import immediately after).

jhlegarreta:
As a side note, in order to test this properly, an attempt to install DIPY with Python 3.4 or lower could be made.

jhlegarreta:
CIs are passing (the codecov one seems an FP?), but this may require additional testing for the reasons stated in the above comments.

skoudoro:

PR #2150, opened 2020-04-26
[Fix] some warning on clustering test

### LINKED ISSUE #2130: Checking empty Cluster objects generates NumPy warning
## Description
When running the [test_clustering](https://github.com/dipy/dipy/blob/master/dipy/segment/tests/test_clustering.py) suite, a number of warnings are generated due to NumPy checking the truth value of an empty array:
```
DeprecationWarning: The truth value of an empty array is ambiguous.
Returning False, but in future this will result in an error. Use `array.size > 0` to check
that an array is not empty.
```
triggered at
https://github.com/numpy/numpy/blob/master/numpy/testing/_private/utils.py#L395
and stemming from
https://github.com/numpy/numpy/blob/master/numpy/core/src/multiarray/number.c#L842

## Way to reproduce
Test the `test_clustering` suite. No input data required.

See for example
https://dev.azure.com/dipy/46625eaf-7255-4a2c-9e92-befcdbdca2a8/_apis/build/builds/293/logs/263

A method of the test set where this is seen is `test_cluster_attributes_and_constructor`, where empty clusters are created.

On a side note, but still related to creating empty clusters, in `test_cluster_map_comparison_with_object` the first [cluster that is created](https://github.com/dipy/dipy/blob/master/dipy/segment/tests/test_clustering.py#L537) is empty when the index is 0 because there is no range to account for. Not sure if that is an intended behavior.

Not sure about the specific dependency versions or python versions that trigger the warning. Although the deprecation note in NumPy is 3 years old, reproducing it locally is still elusive to me.

Using [assert_arrays_equal](https://github.com/dipy/dipy/blob/master/dipy/testing/__init__.py#L41) avoids comparing empty arrays (even when one of the two is empty) but returns `None`, so no effective comparison takes place (not sure if that case should be covered in `assert_arrays_equal` regardless of whether its use is appropriate for the current issue). Also, not sure how that would avoid the case of the mentioned side note, where one of the clusters is an empty list.

Suggestions are welcome.

skoudoro:
Remove/fixing the following warning:

```python
segment/tests/test_clustering.py: 18 tests with warnings
  /home/travis/build/dipy/dipy/venv/lib/python3.8/site-packages/numpy/testing/_private/utils.py:402: DeprecationWarning: The truth value of an empty array is ambiguous. Returning False, but in future this will result in an error. Use `array.size > 0` to check that an array is not empty.
    if isdesnan and isactnan:

reconst/tests/test_multi_voxel.py::test_squash
reconst/tests/test_multi_voxel.py::test_squash
  /home/travis/build/dipy/dipy/venv/lib/python3.8/site-packages/numpy/testing/_private/utils.py:794: DeprecationWarning: elementwise comparison failed; this will raise an error in the future.
    val = comparison(x, y)
```

fix #2130

### DIFF
--- a/dipy/reconst/tests/test_multi_voxel.py
+++ b/dipy/reconst/tests/test_multi_voxel.py
@@ -83,9 +83,6 @@ def test_squash():
     msk = np.array([1, 0, 1], dtype=np.bool_)  # explicit mask
     npt.assert_array_equal(_squash(obj_masked, mask=msk, fill=99),
                            arr_masked)
-    msk[1] = True  # unmask None, object array back
-    npt.assert_array_equal(_squash(obj_masked, mask=msk, fill=99),
-                           obj_masked)
 
 
 def test_CallableArray():
--- a/dipy/segment/tests/test_clustering.py
+++ b/dipy/segment/tests/test_clustering.py
@@ -34,9 +34,9 @@ def test_cluster_attributes_and_constructor():
     assert_equal(len(cluster), 0)
 
     # Duplicate
-    assert_equal(cluster, Cluster(cluster.id,
-                                  cluster.indices,
-                                  cluster.refdata))
+    assert_true(cluster == Cluster(cluster.id,
+                                   cluster.indices,
+                                   cluster.refdata))
     assert_false(cluster != Cluster(cluster.id,
                                     cluster.indices,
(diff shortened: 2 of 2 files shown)


Garyfallidis:
Good call.

PR #2152, opened 2020-04-27
changed buan_flow.rst to BUAN_flow.rst

BramshQamar:
This change is in support of the submission to the paper.


skoudoro:
Hi @BramshQamar,

Why do you rename it to uppercase ? @Garyfallidis rename it to lower case in a previous commit. I think lower case for a file name is the best but you both have to agree because this is the back and forward

Garyfallidis:
In the paper we linked the capitalized version of the filename. We can change this when we hear back from the reviewers.

Garyfallidis:
@BramshQamar please add an issue as a reminder.

BramshQamar:
Hi @Garyfallidis and @skoudoro,

Issue created #2153. Will fix the issue once we get the review.

PR #2161, opened 2020-05-01
Fix a memory overlap bug in multi_median (and median_otsu)

grlee77:
This PR works around a bug in SciPy. input/output arrays cannot overlap in memory during calls to `scipy.ndimage.median` (see scipy/scipy#9553).

The issue will be fixed upstream by scipy/scipy#11986 (See the figure in the first comment there, which illustrates the severity of the issue!)

The solution proposed here works with current SciPy and should be kept even once the SciPy minimum version is updated, as it avoids repeatedly creating a new temporary array within `scipy.ndimage.median` in each iteration.

I have plotted the smoothed volumes from the following small script pre and post-PR below:
```Python
import numpy as np
from dipy.io.image import load_nifti
from dipy.segment.mask import multi_median

hardi_fname, hardi_bval_fname, hardi_bvec_fname = get_fnames('stanford_hardi')

stanford_b0, stanford_b0_affine = load_nifti(hardi_fname)
stanford_b0 = np.squeeze(stanford_b0)[..., 0]

filtered = multi_median(stanford_b0, median_radius=4, numpass=1)
```

**Current Master (erroneous, overly smooth)**
01_current_master_1iter

**With this PR (looks as expected)**
02_fixed_1iter

**One other issue to consider:** Does the change here imply that the default values for `median_radius` and `numpass` should be adjusted? I am not sure if they were initially set empirically, relying on the faulty filtering, or if the values come from the literature. I checked back to SciPy 1.1 and it still had the same bug, but I didn't go back to much older versions..

### DIFF
--- a/dipy/segment/mask.py
+++ b/dipy/segment/mask.py
@@ -200,7 +200,7 @@ def median_otsu(input_volume, vol_idx=None, median_radius=4, numpass=4,
         else:
             raise ValueError("For 4D images, must provide vol_idx input")
     else:
-        b0vol = input_volume.copy()
+        b0vol = input_volume
     # Make a mask using a multiple pass median filter and histogram
     # thresholding.
     mask = multi_median(b0vol, median_radius, numpass)
--- a/dipy/segment/mask.py
+++ b/dipy/segment/mask.py
@@ -35,8 +35,10 @@ def multi_median(input, median_radius, numpass):
     medarr = np.ones_like(input.shape) * ((median_radius * 2) + 1)
 
     # Multi pass
+    output = np.empty_like(input)
     for i in range(0, numpass):
-        median_filter(input, medarr, output=input)
+        median_filter(input, medarr, output=output)
+        input, output = output, input
     return input
 
 
--- a/dipy/segment/mask.py
+++ b/dipy/segment/mask.py
@@ -34,6 +34,10 @@ def multi_median(input, median_radius, numpass):
     # Array representing the size of the median window in each dimension.
     medarr = np.ones_like(input.shape) * ((median_radius * 2) + 1)
 
+    if numpass > 1:
+        # ensure the input array is not modified
+        input = input.copy()
+
     # Multi pass
     output = np.empty_like(input)
     for i in range(0, numpass):
--- a/dipy/segment/tests/test_mask.py
+++ b/dipy/segment/tests/test_mask.py
@@ -37,13 +37,14 @@ def test_mask():
     assert_equal(final, initial)
 
     # Test multi_median.
-    median_test = np.arange(25).reshape(5, 5)
-    median_control = median_test.copy()
+    img = np.arange(25).reshape(5, 5)
+    img_copy = img.copy()
     medianradius = 2
-    median_test = multi_median(median_test, medianradius, 3)
+    median_test = multi_median(img, medianradius, 3)
+    assert_equal(img, img_copy)
 
-    medarr = np.ones_like(median_control.shape) * ((medianradius * 2) + 1)
-    median_control = median_filter(median_control, medarr)
+    medarr = np.ones_like(img.shape) * ((medianradius * 2) + 1)
+    median_control = median_filter(img, medarr)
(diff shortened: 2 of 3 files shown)


skoudoro:
Thank you for doing this @grlee77 and really good catch!

> Does the change here imply that the default values for median_radius and numpass should be adjusted? I am not sure if they were initially set empirically, relying on the faulty filtering, or if the values come from the literature

I can not answer this question. When I look at the git history, it seems that @Garyfallidis or @mdesco will have this answer. From what I see, it seems that these values do not come from the literature.

Otherwise, this change looks good to me.

skoudoro:
I just find it strange that [this test](https://github.com/dipy/dipy/blob/master/dipy/segment/tests/test_mask.py#L48) is passing after your change and especially your result. Can you check why? Thank you.

grlee77:
> I just find it strange that this test is passing after your change and especially your result. Can you check why? Thank you.

Yeah, that is a little odd that it happens to match in that case. It does fail if I reduce `medianradius` from 3 to 1 or 2. I will update the test to use a smaller radius and validate against a version not relying on the output argument.

grlee77:
Also, I think it `multi_median` should make an internal copy of `input`, or at least warn the user in the docstring that calling this function may modify its input. What do you think?

There is an internal copy made in `multi_otsu`, so the same problem doesn't occur there.

skoudoro:
> if I reduce medianradius from 3 to 1 or 2

it makes much more sense, I love this kind of tricky coincidence 😄 

> Also, I think it multi_median should make an internal copy of input, or at least warn the user in the docstring that calling this function may modify its input. What do you think?

I agree, we should avoid modifying the user data without warning them. Let me know if you want to do into this PR or I will do it on a new one after this merge. 

Thank you for the update!

grlee77:
> I agree, we should avoid modifying the user data without warning them. Let me know if you want to do it into this PR or I will do it on a new one after this merge.

I went ahead and added it here (see the last two small commits)

skoudoro:
Thank you @grlee77.  It looks good to me and it is ready to go. 

Is someone else wants to have a look at this PR? I will wait until before merging it.

grlee77:
Is there a good place to put a release note to warn users that this fix may cause a change in the brain mask in user workflows?

I checked a few of the demos (`brain_extraction_dwi.py`,  `syn_registration_3d.py`) and they still look okay without changing the parameters in the scripts.

One case that is subtly different is `doc/demos/reconst_fwdti.py`. I get the following image after this PR:
In_vivo_free_water_DTI_and_standard_DTI_measures_after_PR

whereas the image currently in the docs on dipy.org has a slightly different mask which includes some more non-brain tissue:

In_vivo_free_water_DTI_and_standard_DTI_measures

skoudoro:
You can update the [API_change document](https://github.com/dipy/dipy/blob/master/doc/api_changes.rst). Then, your change will appears on the release notes.

skoudoro:
Anyone else needs to take a look?

skoudoro:
Hi @grlee77, 

Can you just add a short note in the [API_change document](https://github.com/dipy/dipy/blob/master/doc/api_changes.rst). I will go ahead and merge your PR after that. 
Thank you !

grlee77:
> Can you just add a short note in the API_change document. I will go ahead and merge your PR after that.

I wouldn't consider this an API change, but a bug fix (all arguments remain the same and nothing was deprecated). I don't see a better document to list it in though, so can go ahead and add it there if you like. I suppose that documentation section could just be titled "API changes and Bug Fixes", although that may also be a bit misleading if other past bugs have not been listed there.

skoudoro:
Indeed, this is a more a bug fix but there is a big change of behavior that will impact many codes.  After looking at the documentation, `api_changes` seems to be the best place since there is no other section/document for this. Maybe @arokem has a better idea?
In any case, Thank you for the update.

arokem:
This is fine. One other option is to add a note section in the docstring that explains this.

skoudoro:
Thank you for this fix @grlee77! merging

PR #2173, opened 2020-05-18
[WIP] Install ssl certificate for azure pipeline windows

skoudoro:
It seems we have an SSL certificate issue on Azure Windows.

Let's see if everything is going well when we force the installation.


skoudoro:
OK, this change fixes our SSL problem on Azure Windows which is a good news.

However, it seems that VTK 9 is just released. So the `viz` module is failing on Windows and MacOS.

I would recommend merging this PR and I will start to focus on VTK to fix the new issue.

Any opinions?

arokem:
That seems like a very good idea to me. You don't want to have to work to figure out one bug with another (unrelated, no less!) bug going on at the same time. Merging!

PR #2174, opened 2020-05-19
Improve performance of tissue classification

grlee77:
This PR improves the performance of tissue classification by modifying the Cython code in the `seg_stats` method of `ConstantObservationModel` to avoid high overhead. There should be no change in behavior with this PR.

This initial call to `seg_stats` in the `classify` method of `TissueClassifierHMRF` for the example data in `doc/demo/tissue_classification.py` takes > 60 seconds on my system prior to this PR and ~60 ms afterwards. So, the setup prior to iteration is now nearly instantaneous. The iterated conditional modes loop itself still takes the same time as before (~2 minutes), giving an overall speedup of around 30% for the demo.

I would recommend reviewing the first commit in isolation as it has all of the important changes.
The remaining commits are just minor maintenance/style fixes.

### DIFF
--- a/dipy/segment/mrf.pyx
+++ b/dipy/segment/mrf.pyx
@@ -66,7 +66,7 @@ class ConstantObservationModel(object):
         return np.array(mu), np.array(sigma)
 
 
-    def seg_stats(self, input_image, seg_image, nclass):
+    def seg_stats(self, input_image, seg_image, int nclass):
         r""" Mean and standard variation for N desired  tissue classes
 
         Parameters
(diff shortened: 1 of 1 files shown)


dipy/segment/mrf.pyx
```diff
@@ -217,48 +239,7 @@ class ConstantObservationModel(object):
         return mu_upd, var_upd
 
 
-    def update_param_new(self, image, P_L_Y, mu, nclasses):
```
Garyfallidis, line 220: It seems this was not used anywhere @grlee77 ?

grlee77: Yes, it seems to be some leftover dead code. I did a quick grep of the code base just to be sure and don't see any other files that refer to it.

dipy/segment/mrf.pyx
```diff
@@ -66,47 +66,67 @@ class ConstantObservationModel(object):
...
+            double v
+            cnp.npy_intp size = input_image.size
+            double [::1] input_view
+            short [::1] seg_view
+            double [::1] mu_view
+            double [::1] std_view
+            cnp.npy_intp [::1] num_vox_view
+
+        # ravel C-contiguous array so we can use 1D indexing in the loop below
+        input_view = np.ascontiguousarray(input_image, dtype=np.double).ravel()
```
Garyfallidis, line 100: Okay so here is the trick that speeds up the code. Have a look @villalonreina

Garyfallidis:
+1 on my side. AFAIK this is ready to be merged. @villalonreina do have a look. If @villalonreina does not get back to us during the next 2 days then we can move ahead. This loos pretty safe. We still may need to wait for the CLI issues. Thank you @grlee77.

skoudoro:
Hi @grlee77, 

CI's are back to green !  So can you just rebase this PR to make sure that everything is ok? and then, the code can be merged. 

Thank you!

grlee77:
Hi @skoudoro, I rebased this on master.  The two Azure failures seem like unrelated errors due to a network error/timeout when downloading packages. Can you try restarting those two and see if it was just a fluke.

skoudoro:
Thanks ! I just restarted them

skoudoro:
merging, Thank you @grlee77!

PR #2175, opened 2020-05-19
Save streamlines as vtk polydata to a supported format file updated t…

svabhishek29:
…oVTK, FIB and everything supported by save_polydata method

### DIFF
--- a/dipy/io/tests/test_streamline.py
+++ b/dipy/io/tests/test_streamline.py
@@ -209,6 +209,15 @@ def test_low_io_vtk():
         npt.assert_array_almost_equal(tracks[1], streamline, decimal=4)
 
 
+@pytest.mark.skipif(not have_fury, reason="Requires FURY")
+def test_ply_io_vtk():
+    with InTemporaryDirectory():
+        fname = 'test.ply'
+
+        # Test save
+        save_vtk_streamlines(streamlines, fname)
+
+
 def trk_loader(filename):
     try:
         with InTemporaryDirectory():
--- a/dipy/io/vtk.py
+++ b/dipy/io/vtk.py
@@ -94,14 +94,14 @@ def save_vtk_streamlines(streamlines, filename,
                          to_lps=True, binary=False):
     """Save streamlines as vtk polydata to a supported format file.
 
-    File formats can be VTK, FIB
+    File formats can be VTK, FIB and everything supported by save_polydata method
 
     Parameters
     ----------
     streamlines : list
         list of 2D arrays or ArraySequence
     filename : string
-        output filename (.vtk or .fib)
+        output filename (.vtk, .fib or everything supported by save_polydata method)
     to_lps : bool
         Default to True, will follow the vtk file convention for streamlines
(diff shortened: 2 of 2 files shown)


dipy/io/vtk.py
```diff
@@ -146,15 +146,19 @@ def save_vtk_streamlines(streamlines, filename,
     polydata.SetPoints(vtk_points)
     polydata.SetLines(vtk_lines)
 
-    writer = vtk.vtkPolyDataWriter()
-    writer.SetFileName(filename)
-    writer = utils.set_input(writer, polydata)
+    file_extension = filename.split(".")[-1].lower()
```
skoudoro, line 149: I think, it would be better to use `os.path.splitext()`. it is more explicit

dipy/io/vtk.py
```diff
@@ -146,15 +146,19 @@ def save_vtk_streamlines(streamlines, filename,
...
+    if file_extension != ".vtk" or file_extension != ".fib":
```
skoudoro, line 150: let's be more pythonic!  `if file_extension not in [".vtk", ".fib"]` instead

svabhishek29: i have made the corresponding change

dipy/io/vtk.py
svabhishek29: i have made the corresponding change

dipy/io/tests/test_streamline.py
```diff
@@ -209,6 +209,15 @@ def test_low_io_vtk():
...
 
 
+@pytest.mark.skipif(not have_fury, reason="Requires FURY")
+def test_ply_io_vtk():
+    with InTemporaryDirectory():
+        fname = 'test.ply'
+
+        # Test save
+        save_vtk_streamlines(streamlines, fname)
+
```
guaje, line 219: Can you complete the test? This can be made in a similar way as the other test cases in this module. Read the saved file and compare its length to the length of the streamlines object i.e., `tracks = load_vtk_streamlines(fname)` and `npt.assert_equal(len(tracks), len(streamlines))`. You could compare the content of both objects as follows: `npt.assert_array_almost_equal(tracks[1], streamline, decimal=4)`.

dipy/io/vtk.py
```diff
@@ -146,15 +147,19 @@ def save_vtk_streamlines(streamlines, filename,
     polydata.SetPoints(vtk_points)
     polydata.SetLines(vtk_lines)
 
-    writer = vtk.vtkPolyDataWriter()
-    writer.SetFileName(filename)
-    writer = utils.set_input(writer, polydata)
+    file_extension = os.path.splitext(filename)
+    if file_extension not in [".vtk", ".fib"]:
```
guaje, line 151: You could add all the file extensions supported in the `save_polydata` function.

skoudoro:
Closing this PR in favor of #2196. Thanks @svabhishek29  for your changes, I use them on the new PR. Thanks @guaje for the review.  closing

Issue #1969, opened 2019-09-13
imaffine mask support


d-v-b:
## Description
Affine registration doesn't support masked regions. The mutual information metric does appear to have some mask support, but it's not exposed by `AffineRegistration`. Is this on the roadmap?

skoudoro:
Hi @d-v-b, 

Thank you for pointing that, we will try to look into it. Indeed, I saw some mask support and test for the `ParzenJointHistogram` but we need to go deeper and play a bit with it. But feel free to create a PR if you see exactly what needs to be done!

d-v-b:
Hi @skoudoro 
I'm playing around with the code at the moment to try and understand things properly... If I feel like I can get the mask stuff working I will absolutely submit a PR.

d-v-b:
@skoudoro I got the masking working in the `mask` branch of my fork of `dipy`, e.g.: https://github.com/d-v-b/dipy/blob/mask/dipy/align/imaffine.py

Here are the results: the middle blob is moved while the outer 8 blobs remain in place. Aligning without a mask doesn't find the translation of the middle blob, but using a mask to focus the optimization on the central region (colored square) does find that translation. 
![image](https://user-images.githubusercontent.com/3805136/64926185-93866400-d7c8-11e9-9ba2-26d0368e87f9.png)


I next need to put together a proper test for it (maybe the example above can be the test case).

I was pretty aggressive about moving code around in `AffineRegistration`, I'm not sure you will like all my changes. But hopefully we can separate that from the changes related to getting the masking working right.

I'm also interested in benchmarking the timings -- i'm not sure how much (if any) speed benefit will come from masking.

skoudoro:
Good job @d-v-b ! The result look nice! Will look into your code tonight or tomorrow.

Concerning the test, I agree that your example above is a nice test case. Look forward to see your benchmarking results!

PR #2179, opened 2020-05-25
NF: Complete masking implementation in affine registration with MI

bpinsard:
I noticed that the ParzenJointHistogram can use binary mask, so I changed the mutual-info and registration code to be able to use it, as masking can sometime improve the registration.

For the moment, there is no testing nor updated doc, nor pep8 (will do if there is interest in merging).

Comments and advices are welcomed. 

Thanks.

### DIFF
--- a/dipy/align/imaffine.py
+++ b/dipy/align/imaffine.py
@@ -498,7 +498,8 @@ class MutualInformationMetric(object):
         self.metric_grad = None
 
     def setup(self, transform, static, moving, static_grid2world=None,
-              moving_grid2world=None, starting_affine=None):
+              moving_grid2world=None, starting_affine=None,
+              static_mask=None, moving_mask=None):
         r"""Prepare the metric to compute intensity densities and gradients.
 
         The histograms will be setup to compute probability densities of
@@ -541,6 +542,8 @@ class MutualInformationMetric(object):
         self.transform = transform
         self.static = np.array(static).astype(np.float64)
         self.moving = np.array(moving).astype(np.float64)
+        self.static_mask = static_mask
+        self.moving_mask = moving_mask
         self.static_grid2world = static_grid2world
         self.static_world2grid = npl.inv(static_grid2world)
         self.moving_grid2world = moving_grid2world
@@ -584,7 +587,7 @@ class MutualInformationMetric(object):
             static_p = static_p[..., :self.dim]
             self.static_vals, inside = self.interp_method(static, static_p)
             self.static_vals = np.array(self.static_vals, dtype=np.float64)
-        self.histogram.setup(self.static, self.moving)
+        self.histogram.setup(self.static, self.moving, self.static_mask, self.moving_mask)
 
     def _update_histogram(self):
         r"""Update the histogram according to the current affine transform.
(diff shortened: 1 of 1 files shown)


skoudoro:
Thank you for doing this @bpinsard! 

I confirm, there is an interest in merging. It looks like an implementation of issue #1969. It would be good if you could look at it and show similar results on this PR. It gives an idea of what could be a nice test.

Can you have a look at this PR @d-v-b? you might be interested.

We will try your branch asap.

Thanks!

bpinsard:
Thanks for the quick reply, I will read that issue carefully. 

Does it make sense to mask the image for center-of-mass init? I think so, as the masking really helps when hyper-intense areas mess up the registration in MRI images (eg. neck fat for instance).

dipy/align/imaffine.py
```diff
@@ -967,23 +979,47 @@ def _init_optimizer(self, static, moving, transform, params0,
                                                  self.sigmas,
                                                  moving_grid2world,
                                                  moving_spacing, False)
+            if moving_mask:
+                self.moving_mask_ss = IsotropicScaleSpace(
+                    moving_mask, self.factors,
+                    [0]*len(self.factors),
```
bpinsard, line 985: Here I set the smoothing factors to zero, as it is a mask. But it might be better to smooth, interpolate, and threshold. What do you think?

d-v-b:
This looks good! Sorry I haven't revisited #1969 . I eventually found that masking didn't have the performance impact I expected, and I moved on to other stuff...

BramshQamar:
I think it would be useful to have an option to mask out non-brain tissues before registration.

Not sure if it improves the performance of registration. We will need to check if it improves registration quality and speed.

If the mask is provided, the output would be masked out registred moving brain to static?

dipy/align/imaffine.py
```diff
@@ -498,7 +498,8 @@ def __init__(self, nbins=32, sampling_proportion=None):
         self.metric_grad = None
 
     def setup(self, transform, static, moving, static_grid2world=None,
```
BramshQamar, line 500: docstring needs to be updated with information about new parameters, `static_mask` and `moving_mask`

dipy/align/imaffine.py
```diff
@@ -880,7 +891,8 @@ def __init__(self,
 
     def _init_optimizer(self, static, moving, transform, params0,
```
BramshQamar, line 892: docstring needs to be updated with information about new parameters, `static_mask` and `moving_mask`

dipy/align/imaffine.py
```diff
@@ -921,16 +933,22 @@ def _init_optimizer(self, static, moving, transform, params0,
         n = transform.get_number_of_parameters()
         self.nparams = n
 
+        static_masked, moving_masked = static, moving
+        if not static_mask is None:
```
BramshQamar, line 937: and if it's `None`?
`static_masked` will be not defined?
It will give an error in line number 949

dipy/align/imaffine.py
```diff
@@ -921,16 +933,22 @@ def _init_optimizer(self, static, moving, transform, params0,
         n = transform.get_number_of_parameters()
         self.nparams = n
 
+        static_masked, moving_masked = static, moving
```
BramshQamar, line 936: by doing so `static_masked` and `static` will be both same? and the same goes for `moving` and `moving_masked`

dipy/align/imaffine.py
BramshQamar, line 937: here `static_mask` object will not be `None` as we asigned static object to it earlier

dipy/align/imaffine.py
BramshQamar: okay, I see, if the mask is empty we use original data and if the mask is available we mask the data and use that for registration.

BramshQamar: So the registration parameters will be optimized for masked data (say mask is provided) and if someone applies transformation to original data (not masked) will it give good output?

I guess we need to check it by testing it.

bpinsard: What do you mean by good output?
As it is affine registration, the transformation of out-of-mask data should not go completely crazy as it could with warping field. Of course registering masked brain would not register well other non-rigidly attached body-parts such as neck and spine.

BramshQamar: I am saying we should test different senarios and compare the results.

Garyfallidis:
Any more thoughts about this @bpinsard?

skoudoro:
Hi @bpinsard,

Can you rebase and address the comments? Let us know if you need help. Thank you

PR #2181, opened 2020-06-11
BUAN bundle highlight option in horizon

BramshQamar:
Added functionality in `dipy_horizon` to show highlighted areas/segments on a bundle where pvalues (negative log of pvalues) are greater than a certain threshold (threshold is a parameter).

Usage:

`dipy_horizon IFOF_L.trk IFOF_L_fa_pvalues_log.npy --buan --buan_thr 2`

`buan_thr 2` for `-log(pvalues)` is equal to `0.01` pvalues.

Resulting visualization looks something like this:

Screen Shot 2020-06-10 at 10 34 19 PM

Where, red color indicates -log(pvalues) greater than a buan threshold of 2.

We can also provide `pvalues` instead of `-log(pvalues)`.

`dipy_horizon IFOF_L.trk IFOF_L_fa_pvalues.npy --buan --buan_thr 0.01`

Results will be the same but with inverted colors. 
Screen Shot 2020-06-10 at 10 33 47 PM
Now, green is the highlighted area.

### DIFF
--- a/dipy/viz/app.py
+++ b/dipy/viz/app.py
@@ -80,8 +80,8 @@ HELP_MESSAGE = """
 
 class Horizon(object):
 
-    def __init__(self, tractograms=None, images=None, pams=None, np_files=None,
-                 cluster=False, cluster_thr=15.0,
+    def __init__(self, tractograms=None, images=None, pams=None,
+                 npy_files=None, cluster=False, cluster_thr=15.0,
                  random_colors=False, length_gt=0, length_lt=1000,
                  clusters_gt=0, clusters_lt=10000,
                  world_coords=True, interactive=True,
@@ -100,7 +100,7 @@ class Horizon(object):
             Each tuple contains data and affine
         pams : sequence of PeakAndMetrics
             Contains peak directions and spherical harmonic coefficients
-        np_files : sequence of arrays
+        npy_files : sequence of arrays
             Contains numpy array with pvalues along the length of bundles
             extracted from BUAN framework.
         cluster : bool
@@ -774,7 +774,7 @@ class Horizon(object):
                           reset_camera=False)
 
 
-def horizon(tractograms=None, images=None, pams=None, np_files=None,
+def horizon(tractograms=None, images=None, pams=None, npy_files=None,
             cluster=False, cluster_thr=15.0,
             random_colors=False, bg_color=(0, 0, 0), order_transparent=True,
             length_gt=0, length_lt=1000, clusters_gt=0, clusters_lt=10000,
@@ -792,7 +792,7 @@ def horizon(tractograms=None, images=None, pams=None, np_files=None,
         Each tuple contains data and affine
     pams : sequence of PeakAndMetrics
         Contains peak directions and spherical harmonic coefficients
-    np_files : sequence of arrays
+    npy_files : sequence of arrays
         Contains numpy array with pvalues along the length of bundles
         extracted from BUAN framework.
     cluster : bool
@@ -847,7 +847,7 @@ def horizon(tractograms=None, images=None, pams=None, np_files=None,
         adaptive visualization, Proceedings of: International Society of
         Magnetic Resonance in Medicine (ISMRM), Montreal, Canada, 2019.
     """
-    hz = Horizon(tractograms, images, pams, np_files, cluster, cluster_thr,
+    hz = Horizon(tractograms, images, pams, npy_files, cluster, cluster_thr,
                  random_colors, length_gt, length_lt,
                  clusters_gt, clusters_lt,
                  world_coords, interactive,
--- a/dipy/viz/app.py
+++ b/dipy/viz/app.py
@@ -83,7 +83,7 @@ class Horizon(object):
 
 
     def __init__(self, tractograms=None, images=None, pams=None,
-                 npy_files=None, cluster=False, cluster_thr=15.0,
+                 cluster=False, cluster_thr=15.0,
                  random_colors=False, length_gt=0, length_lt=1000,
                  clusters_gt=0, clusters_lt=10000,
                  world_coords=True, interactive=True,
@@ -102,10 +102,6 @@ class Horizon(object):
             Each tuple contains data and affine
         pams : sequence of PeakAndMetrics
             Contains peak directions and spherical harmonic coefficients
-        npy_files : sequence of arrays
-
-            Contains numpy array with pvalues along the length of bundles
-            extracted from BUAN framework.
         cluster : bool
             Enable QuickBundlesX clustering
         cluster_thr : float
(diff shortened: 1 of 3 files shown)


BramshQamar:
Hello @Garyfallidis and @skoudoro,

This PR is ready and needs review.

The only test failing on all machines is `FAILED workflows/tests/test_reconst_mapmri.py::test_reconst_mmri_both `. Which is not related to this PR.

dipy/viz/app.py
```diff
@@ -775,6 +792,9 @@ def horizon(tractograms=None, images=None, pams=None,
         Each tuple contains data and affine
     pams : sequence of PeakAndMetrics
         Contains peak directions and spherical harmonic coefficients
+    np_files : sequence of arrays
```
Garyfallidis, line 795: ``np_files`` seems too generic to me if it is going to be used only for BUAN.

Garyfallidis: But I think you kept that name because it can be used for other visualizations too.
Maybe ``npy_files``?

dipy/workflows/viz.py
```diff
@@ -150,6 +158,36 @@ def run(self, input_files, cluster=False, cluster_thr=15.,
                     print('Peak_dirs shape')
                     print(pam.peak_dirs.shape)
 
+            if ends(".npy"):
+
+                data = np.load(fname)
+                numpy_files.append(data)
```
Garyfallidis, line 164: I am not sure how safe is appending numpy arrays into lists. I think you should use ``ArraySequence`` if you can. That is much preferred method. Remember that we had memory issues when adding arrays to lists. You can check if this problem persists by trying to delete this memory and see if the memory (list of arrays) is actually being removed from the RAM while the object is running.

dipy/workflows/viz.py
```diff
@@ -150,6 +158,36 @@ def run(self, input_files, cluster=False, cluster_thr=15.,
...
+                pvalues = numpy_files[i]
+                bundle = tractograms[i].streamlines
+
+                indx = assignment_map(bundle, bundle, n)
+                ind = np.array(indx)
+
+                colors = []
+                for i in range(len(ind)):
+
+                    if pvalues[ind[i]] < buan_thr:
```
Garyfallidis, line 184: I would prefer if the highlight color was available as an option.
Also I would like if we could keep the original color of the bundle if we want.

skoudoro:
Hi @BramshQamar,

Please, can you rebase your PR? Also, can you address @Garyfallidis comments and add some tests. Thank you

BramshQamar:
Updates this PR color feature. 

- [ ] - Bundles now have an original color with bundle highlight area color as optional input parameter (default red).

- [ ] - Giving both p-values and log(p-values) was redundant. It only takes p-values now. 

Some examples:
Screen Shot 2020-10-15 at 4 08 17 PM

Screen Shot 2020-10-15 at 4 10 39 PM

- [ ] Needs tests

dipy/workflows/viz.py
BramshQamar: These files are very small in size. Eg: 100 integers only in one file. Mostly user will send one .npy file with pvalues in it (100 integers in the file). This should not cause any memory issues.

dipy/workflows/viz.py
BramshQamar: Done. Added original color for bundles with bundle highlight area color as an option (default red).

BramshQamar:
> Hi @BramshQamar,
> 
> Please, can you rebase your PR? Also, can you address @Garyfallidis comments and add some tests. Thank you

Hi @skoudoro, I have addressed all @Garyfallidis comments and added tests as well.
Let me know if I missed anything. Thank you!

dipy/viz/app.py
```diff
@@ -282,15 +291,21 @@ def add_cluster_actors(self, scene, tractograms,
                     apply_shader(self, centroid_actor)
 
             else:
+                if self.buan:
+                    streamline_actor = actor.line(streamlines,
+                                                  self.buan_colors[color_ind])
+                else:
+                    streamline_actor = actor.line(streamlines, colors=colors)
```
skoudoro, line 298: Just a small suggestion: you could replace this block with this 2 lines:

```python
s_color = self.buan_colors[color_ind] if self.buan else colors
streamline_actor = actor.line(streamlines, colors=s_colors)
```

dipy/viz/app.py
```diff
@@ -807,6 +823,10 @@ def horizon(tractograms=None, images=None, pams=None,
     interactive : bool
         Allow user interaction. If False then Horizon goes on stealth mode
         and just saves pictures.
+    buan : bool
```
skoudoro, line 826: `buan : bool, optional` instead

dipy/viz/app.py
```diff
@@ -807,6 +823,10 @@ def horizon(tractograms=None, images=None, pams=None,
...
+        Enables BUAN framework visualization.
```
skoudoro, line 827: `Default is False`, at the end

dipy/workflows/tests/test_viz.py
```diff
@@ -67,6 +67,13 @@ def test_horizon_flow():
             clusters_lt=np.inf, clusters_gt=0,
             world_coords=True, interactive=False)
 
+    buan_colors = np.ones(streamlines.get_data().shape)
+
+    print(buan_colors)
```
skoudoro, line 72: `print` to remove

dipy/workflows/viz.py
```diff
@@ -6,7 +6,15 @@
...
 from dipy.io.utils import create_nifti_header
+from dipy.stats.analysis import assignment_map
+from vtk.util import numpy_support
+from dipy.utils.optpkg import optional_package
 
+fury, has_fury, setup_module = optional_package('fury')
+
+if has_fury:
+    from fury.colormap import line_colors
+from fury.utils import numpy_to_vtk_colors
```
skoudoro, line 17: indentation issue

dipy/workflows/viz.py
```diff
@@ -62,6 +70,15 @@ def run(self, input_files, cluster=False, cluster_thr=15.,
         disable_order_transparency : bool, optional
             Default False. Use depth peeling to sort transparent objects.
             If True also enables anti-aliasing.
+        buan : bool
```
skoudoro, line 73: `buan : bool, optional` instead

dipy/workflows/viz.py
```diff
@@ -62,6 +70,15 @@ def run(self, input_files, cluster=False, cluster_thr=15.,
...
+            Enables BUAN framework visualization.
```
skoudoro, line 74: same as before `default...`

skoudoro (commented):
Hi @BramshQamar,

Thank you for the update!  Overall, it looks good. There are just some quick details to fix. I will try your code a bit later in the day

skoudoro:
Also, Can you fix all pep8 listed [here](https://github.com/dipy/dipy/pull/2181#issuecomment-642366335)? 
Thank you

skoudoro:
Thank you @BramshQamar! merging

Issue #2182, opened 2020-06-11
`target_line_based` might read out of bounds 


takhs91:
## Description
`target_line_based` and specifically its underlying `_streamline_in_mask` don't have any check for out of bounds reads, this results in random fibers with points falling outside of the mask being returned as passing from the mask as it probably reads bytes out of bounds. It may even result in a segmentation fault

Having points falling out of bounds is a valid case since somebody might want to filter streamlines with a mask taken on a anatomical image with smaller bounding box than the DTI image.

## Code Example
```
import numpy as np
from dipy.tracking.vox2track import _streamlines_in_mask

streamline = np.array([[-8, 0, 0], [0, 0, 0]])
mask = np.zeros((2, 1, 1), dtype=np.uint8)
mask[1, 0, 0] = 1
_streamlines_in_mask([streamline], mask, np.eye(3), 0)
```
This results in `Out[8]: array([1], dtype=uint8)`

```
streamline = np.array([[0, 0, 0], [-8, 0, 0]])
```
This results in `Out[17]: array([49], dtype=uint8)` 

```
streamline = np.array([[-100000000000000000, 0, 0], [0, 0, 0]])
```
This results in a `Segmentation fault (core dumped)`

## dipy version
`1.1.1`

## Python version
`Python 3.5.2`

## Operating System and version
`Linux-4.4.0-179-generic-x86_64-with-LinuxMint-18-sarah`

takhs91:
I would like to discuss some more issues

[...]

PR #2183, opened 2020-06-12
_streamlines_in_mask bounds check

takhs91:
Closes https://github.com/dipy/dipy/issues/2182

Adds bound checking in _streamlines_in_mask when accessing the mask array.

I also changed it to explicitly return 1 or 0  for the final point to avoid returning unwanted values given in the mask array

### DIFF
--- a/dipy/tracking/tests/test_utils.py
+++ b/dipy/tracking/tests/test_utils.py
@@ -314,7 +314,7 @@ def _target(target_f, streamlines, voxel_both_true, voxel_one_true,
     assert_true(exclude[0] is streamlines[1])
 
 
-@pytest.mark.parametrize("streamline, no_matched", [
+@pytest.mark.parametrize("streamline, expected_matched", [
     (np.array([[-10, 0, 0], [0, 0, 0]]), 0),
     (np.array([[-10, 0, 0], [1, -10, 0]]), 0),
     (np.array([[0, 0, 0], [10, 10, 10]]), 0),
@@ -323,13 +323,13 @@ def _target(target_f, streamlines, voxel_both_true, voxel_one_true,
     (np.array([[-10, 0, 0], [10, 0, 0]]), 1),
     (np.array([[1, -10, 0], [1, 10, 0]]), 1),
 ])
-def test_target_line_based_out_of_bounds(streamline, no_matched):
+def test_target_line_based_out_of_bounds(streamline, expected_matched):
     # Ensures https://github.com/dipy/dipy/issues/2182 doesn't happen
     # and that target_line_based works with out of bounds points
     mask = np.zeros((2, 1, 1), dtype=np.uint8)
     mask[1, 0, 0] = 1
     matched = list(target_line_based([streamline], np.eye(4), mask))
-    assert len(matched) == no_matched
+    assert len(matched) == expected_matched
 
 
 def test_near_roi():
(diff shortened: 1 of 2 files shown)


arokem:
Looks legit to me. Could you maybe add a test case along the lines of what you had in #2182 ?

takhs91:
I have updated that based on my comment https://github.com/dipy/dipy/issues/2182#issuecomment-644857092

jchoude:
That's a good check to have as a general case.

However, when calling `target_line_based`, in #2182 you mention using a mask coming a different anatomical reference and bounding box than the DWI used to generate the streamline. I'm not sure if this should be handled here or not, but this can be quite risky if the affines are not handled correctly, since the basic assumption here is that the streamline points and intersections to the grid are mapped to a grid which is based on the dimensions of the reference anatomy. If the dimensions do not fit with the content of the streamlines file, there's a very probable chance of getting inconsistent results due to wraparound, which would not necessarily be caught here.

takhs91:
Hi @jchoude,
> you mention using a mask coming a different anatomical reference and bounding box than the DWI used to generate the streamline.

I mean using a mask taken on a anatomical e.g. T1 image of the same subject in the same scanner that might have a smaller bounding box than the DWI image. The affine provided in the `target_line_based` would have to be the appropriate one to bring the streamline points to the masks's coordinate system of course. You might have also cropped the mask array based on the masks bounding box to save space (its affine will had to be updated ).

>  If the dimensions do not fit with the content of the streamlines file, there's a very probable chance of getting inconsistent results due to wraparound, which would not necessarily be caught here.

This is what I try to fix with this PR. The points reaching `_streamline_in_mask` might be out of bounds of the mask's shape, the line based algorithm will work on them(since I believe they are handled correctly in the rest of the algorithm except from when accessing the mask array (and when were turned to x,y,z with truncation instead of floor))  but the mask array will never be accessed with out of bounds indices.

In that way you can still  pass streamlines that may completely or partially fall out of masks's shape to `target_line_based` and they will be still filtered correctly. Out of bound points will simply get ignored.

 > there's a very probable chance of getting inconsistent results due to wraparound, which would not necessarily be caught here.

I think that what was happening was worse than wraparound  because when out of bounds indices were passed [here](https://github.com/dipy/dipy/blob/5e4adff8b1d86210a683b84034c18e388914a09e/dipy/tracking/vox2track.pyx#L319) for example  it  was even returning random values other than  0 and 1 that were only included  in the mask array. I was even getting segmentation fault at some cases as I state in the issue. I think it was reading out of bounds as it is possible to do in C language and since it is Cython this is possible I guess.

Do you think after the changes in this PR, out of bounds points still create an issue for this function? If so can you please provide me with an example?

takhs91:
Hi! Any updates on that?

Garyfallidis:
@jchoude can you please suggest further steps ?

jchoude:
In fact, the fix works. I was just wary of people getting edge cases when misusing the internal function.

Code looks good to me.

Just not sure why CICD is not working.

skoudoro:
Hi @takhs91,

Can you rebase your PR to be sure that everything is ok? Thank you. 

After the rebase and If everything is ok, we will go ahead and merge it. Thank you

skoudoro:
Thank you very much @takhs91 for the rebase and the update. merging

PR #2195, opened 2020-07-20
[ENH] Validate streamlines pre-LiFE

dPys:
Maybe we could also add an optional brain mask argument to ensure we're only considering fibers inside the brain?

### DIFF
--- a/dipy/tracking/life.py
+++ b/dipy/tracking/life.py
@@ -467,6 +467,9 @@ class FiberModel(ReconstModel):
         """
         if affine is None:
             affine = np.eye(4)
+        streamline = [s for s in streamline if len(s) >= 1]
+        if len(streamline) == 0:
+            raise ValueError('Streamlines have no nodes.')
         life_matrix, vox_coords = \
             self.setup(streamline, affine, evals=evals, sphere=sphere)
         (to_fit, weighted_signal, b0_signal, relative_signal, mean_sig,
--- a/dipy/tracking/life.py
+++ b/dipy/tracking/life.py
@@ -467,9 +467,9 @@ class FiberModel(ReconstModel):
         """
         if affine is None:
             affine = np.eye(4)
-        streamline = [s for s in streamline if len(s) >= 1]
-        if len(streamline) == 0:
-            raise ValueError('Streamlines have no nodes.')
+        sl_len = np.array([len(s) for s in streamline])
+        if np.any(sl_len < 2):
+            raise ValueError('Input contains streamlines with only one node. The LiFE model cannot be fit with these streamlines included.')
         life_matrix, vox_coords = \
             self.setup(streamline, affine, evals=evals, sphere=sphere)
         (to_fit, weighted_signal, b0_signal, relative_signal, mean_sig,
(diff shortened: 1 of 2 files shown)


dipy/tracking/life.py
```diff
@@ -467,6 +467,9 @@ def fit(self, data, streamline, affine, evals=[0.001, 0, 0],
         """
         if affine is None:
             affine = np.eye(4)
+        streamline = [s for s in streamline if len(s) >= 1]
+        if len(streamline) == 0:
+            raise ValueError('Streamlines have no nodes.')
```
arokem, line 472: I'd rather check and let the user do their own filtering:

```suggestion
        sl_len = np.array([len(s) for s in streamline])
        if np.any(sl_len < 2):
            raise ValueError('Input contains streamlines with only one node. The LiFE model cannot be fit with these streamlines included.')
```

dPys: Sounds good

dPys:
Test added

arokem:
Looks like there's some problem with the fixture you added? https://travis-ci.com/github/dipy/dipy/jobs/362957473#L2288

dPys:
Hmm that's odd. I'm trying to test this module, but when I run pytest -vvv test_life.py, I get:
```
Hint: make sure your test modules/packages have valid Python names.
Traceback:
/usr/local/anaconda3/lib/python3.7/site-packages/_pytest/python.py:511: in _importtestmodule
    mod = self.fspath.pyimport(ensuresyspath=importmode)
/usr/local/anaconda3/lib/python3.7/site-packages/py/_path/local.py:701: in pyimport
    __import__(modname)
/usr/local/anaconda3/lib/python3.7/site-packages/_pytest/assertion/rewrite.py:152: in exec_module
    exec(co, module.__dict__)
test_life.py:3: in <module>
    import dipy.core.gradients as grad
../../core/gradients.py:10: in <module>
    from dipy.core.sphere import disperse_charges, HemiSphere
../../core/sphere.py:9: in <module>
    from dipy.reconst.recspeed import remove_similar_vertices
E   ModuleNotFoundError: No module named 'dipy.reconst.recspeed'
```
Have you encountered this error before? It's strange because no relative imports are used in the file and I can easily import dipy.reconst.recspeed in ipython so the install is good...

arokem:
Yo @dpys: are you sure you are running this in the same env as the import in python? This looks like what I get when I don't have the cython extensions compiled. Could you please try compiling them again (e.g., `python setup.py build_ext --inplace`)?

skoudoro:
> This looks like what I get when I don't have the cython extensions compiled

+1, I encounter this when there is an environment problem

Can you rebase your PR? Furthermore, I recommend to keep the test simple so to avoid the use of `pytest.parametrize`.  Can you update your test without this decorator ?
Thank you

dPys:
Thanks @arokem and @skoudoro , that was totally what it was!

Compiling, rebasing, and will update test per everyone's suggestions... Stay tuned

dPys:
Done

arokem:
LGTM. I think we can merge when the CI comes back green.

skoudoro:
CI's are green, Thank you @dPys! merging

PR #2197, opened 2020-07-21
(WIP) Estimate smt2 metrics from mean signal kurtosis

RafaelNH:
As showed in [(Henriques et. 2019)](https://onlinelibrary.wiley.com/doi/full/10.1002/mrm.27606) there is a link between the mean signal kurtosis and different SMT models.

Therefore, in this PR, I propose adding the estimates of the 2 compartment SMT model (SMT2, [Kaden et al., 2016](https://www.sciencedirect.com/science/article/pii/S1053811916302063)) parameters that can be obtained from the MSDKI model (equations 16 and 17 of [(Henriques et. 2019)](https://onlinelibrary.wiley.com/doi/full/10.1002/mrm.27606)).


dipy/reconst/msdki.py
```diff
@@ -56,6 +57,163 @@ def mean_signal_bvalue(data, gtab, bmag=None):
...
+            continue
+
+        if msk[v] > 2.4:
+            awf[v] = 1
+        elif msk[v] < 0:
+            awf[v] = 0
+        else:
+            mski = msk[v]
+            fini = mski / 2.4  # Initial guess based on linear assumption
+            awf[v] = opt.fsolve(_msk_from_awf_error, fini, args=(mski,),
```
arokem, line 211: Do you want to catch and raise a warning here if this does not converge properly? Users will see something like:

```
Users/arokem/.virtualenvs/dipy/lib/python3.7/site-packages/scipy/optimize/minpack.py:162: RuntimeWarning: The iteration is not making good progress, as measured by the 
  improvement from the last ten iterations.
```

Which is pretty cryptic.

skoudoro:
Hi @RafaelNH,

Do you think you find time to finish this PR before the release in 3 weeks or should I move it to release 1.4 in December? 
Thank you for your feedback.

RafaelNH:
Hi all! I am not done yet but almost! I've just updated the tutorial to incorporate these new metrics. The only thing that I want to add before merging is the microscopic anisotropy estimate from MSDKI-SMT2. 

@arokem - I never got that warning on my data - only if I add some NANs in MSK maps. Is that your case? Should I still address that issue?

@skoudoro - I think I can finish this PR next Monday - will I be still in time for the release 1.3?

skoudoro:
> @skoudoro - I think I can finish this PR next Monday - will I be still in time for the release 1.3?

Yes @RafaelNH, I think the release will be cut next Thursday.

RafaelNH:
I am done here! Also addressed @arokem comment. Please let me know if you want me to address something else before merging the work done here! Many thanks!!!

skoudoro:
Thank you for the update @RafaelNH. Can you address all pep8 [here](https://github.com/dipy/dipy/pull/2197#issuecomment-707087092)?

RafaelNH:
Ups! I forgot to check PEP8 on the latest updates. Done! Thanks for pointing this out @skoudoro!

PR #2199, opened 2020-07-25
BUG: Fix NumPy and Cython deprecation and initialization warnings

jhlegarreta:
Fix NumPy and Cython deprecation and initialization warnings.

Fixes:
```
dipy/core/interpolation.c:45594:3: warning: 'tp_print' is deprecated
[-Wdeprecated-declarations]
   0, /*tp_print*/
   ^
/Users/runner/hostedtoolcache/Python/3.8.3/x64/include/python3.8/cpython/object.h:260:5:
note: 'tp_print' has been explicitly marked deprecated here
    Py_DEPRECATED(3.8) int (*tp_print)(PyObject *, FILE *, int);
    ^
/Users/runner/hostedtoolcache/Python/3.8.3/x64/include/python3.8/pyport.h:515:54:
note: expanded from macro 'Py_DEPRECATED'
```

and
```
/Users/runner/work/1/s/venv/lib/python3.8/site-packages/numpy/core/include/numpy/npy_1_7_deprecated_api.h:17:2:
warning: "Using deprecated NumPy API, disable it with "          "#define
NPY_NO_DEPRECATED_API NPY_1_7_API_VERSION" [-W#warnings]
 ^
```

raised for example in:
https://dev.azure.com/dipy/46625eaf-7255-4a2c-9e92-befcdbdca2a8/_apis/build/builds/492/logs/85

The `tp_print` deprecation warning is only present when using Python 3.8 on
macOS. Related Cython references:
https://github.com/cython/cython/pull/3201
https://github.com/cython/cython/issues/3474

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,6 +8,14 @@ from copy import deepcopy
 from os.path import join as pjoin, dirname, exists
 from glob import glob
 
+
+def is_platform_mac():
+    return sys.platform == "darwin"
+
+
+def is_platform_windows():
+    return sys.platform == "win32" or sys.platform == "cygwin"
+
 # BEFORE importing distutils, remove MANIFEST. distutils doesn't properly
 # update it when the contents of directories change.
 if exists('MANIFEST'):
@@ -30,6 +38,7 @@ if force_setuptools:
 # removing MANIFEST
 from distutils.core import setup
 from distutils.extension import Extension
+from distutils.version import LooseVersion
 
 from cythexts import cyproc_exts, get_pyx_sdist
 from setup_helpers import (install_scripts_bat, add_flag_checking,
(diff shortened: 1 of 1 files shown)


jhlegarreta:
A few comments:
- The patch set is inspired by the solution provided by the pandas folks: https://github.com/pandas-dev/pandas/blob/master/setup.py
- Not an expert in these Cython+NumPy issues so reviews are much appreciated.
- Although silencing warnings can hide real bugs, maybe it is the only possibility here.
- Not sure which Cython statement triggers the `tp_print` initialization/deprecation warning.
- Not sure about the benefit/difference between using the `macros` and using the `extra_compiler_flags` or if they serve the same purpose.
- Not sure if the `is_platform_windows()` method should also include the `win64` case.

The NumPy API deprecation warnings and the `tp_print` deprecation warnings seem to stem from two separate issues; the later seems to stem from https://github.com/cython/cython/pull/3201 (see also https://github.com/cython/cython/issues/3474). But not sure if silencing the NumPy deprecation warnings (e.g. `extra_compile_args.append("-Wno-error=deprecated-declarations")`) would silence also the `tp_print` deprecation warnings.

Also, it looks like if Python 3.8 is detected, then the min Cython version should be set to 0.29.12 to avoid the `tp_print` deprecation warnings (see also https://github.com/Cantera/cantera/pull/765/files `SConstruct` vs. `SConscript`). We have `0.29`.

We may be willing to go step by step, and address the `tp_print` issues first, then silence NumPy API deprecation, and maybe split this PR into two different PRs to check the fixes.

So not sure what would be best in this case. Comments and reviews are appreciated.

jhlegarreta:
OK. I think it will be easier one thing at a time:
- Closing this PR and addressing the issues separately will be easier.
- Using `extra_compile_args = ["-Werror"]` marks all warnings as errors (and hence, all builds failed early in this PR). So maybe in the future we will be willing to use it as an early check to avoid warnings becoming into real errors/bugs. So that will be left for the future discussions.

PR #2200, opened 2020-07-30
BF: Fix `'tp_print' is deprecated` Cython warning

jhlegarreta:
Fix `'tp_print' is deprecated` Cython warning.

Fixes:
```
dipy/core/interpolation.c:45594:3: warning: 'tp_print' is deprecated
[-Wdeprecated-declarations]
   0, /*tp_print*/
   ^
/Users/runner/hostedtoolcache/Python/3.8.3/x64/include/python3.8/cpython/object.h:260:5:
note: 'tp_print' has been explicitly marked deprecated here
    Py_DEPRECATED(3.8) int (*tp_print)(PyObject *, FILE *, int);
    ^
/Users/runner/hostedtoolcache/Python/3.8.3/x64/include/python3.8/pyport.h:515:54:
note: expanded from macro 'Py_DEPRECATED'
```

raised for example in:
https://dev.azure.com/dipy/46625eaf-7255-4a2c-9e92-befcdbdca2a8/_apis/build/builds/492/logs/85

The warning is only present when using Python 3.8 on macOS. Related Cython references:
cython/cython#3201
cython/cython#3474

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -8,6 +8,11 @@ from copy import deepcopy
 from os.path import join as pjoin, dirname, exists
 from glob import glob
 
+
+def is_platform_mac():
+    return sys.platform == "darwin"
+
+
 # BEFORE importing distutils, remove MANIFEST. distutils doesn't properly
 # update it when the contents of directories change.
 if exists('MANIFEST'):
@@ -61,6 +66,12 @@ if using_setuptools:
 # Define extensions
 EXTS = []
 
+extra_compile_args = []
+
+# Silence Cython `tp_print` deprecation warnings
+if sys.version_info[:2] == (3, 8) and is_platform_mac:
+    extra_compile_args.append("-Wno-error=missing-field-initializers")
+
 # We use some defs from npymath, but we don't want to link against npymath lib
 ext_kwargs = {'include_dirs': ['src']}  # We add np.get_include() later
 
@@ -105,6 +116,7 @@ for modulename, other_sources, language in (
     pyx_src = pjoin(*modulename.split('.')) + '.pyx'
     EXTS.append(Extension(modulename, [pyx_src] + other_sources,
                           language=language,
+                          extra_compile_args=extra_compile_args,
                           **deepcopy(ext_kwargs)))  # deepcopy lists
 
 # Do our own build and install time dependency checking. setup.py gets called


jhlegarreta:
As said in PR #2199, the deprecation API warnings may be best addressed in a separate PR.

Having a closer look at the `tp_print` issue in the Cython issues and PRs it is less clear to me whether the warning was present in earlier versions of 0.29.x, and whether it has been fixed in 0.19.14 (since [according to this comment](https://github.com/cython/cython/issues/3474#issue-590574992) 0.29.13 was the last version that did not have the warning; [a fix was merged in 0.29.14 vs. 0.29.13](https://github.com/cython/cython/compare/0.29.13...0.29.14); and [the warning is still reported for 0.29.14](https://github.com/cython/cython/issues/3474)).

Also, fixing a min and a max version for Cython may require some refactoring in `setup_helpers::SetupDependency`, so I think the most reasonable thing to do is to silence all warnings of that type for now. An issue can be opened to keep track of it and to remove the workaround when the issue is fixed Cython setting that as the new minimum version.

skoudoro:
Thank you for that @jhlegarreta! I will look into it and keep you updated. Overall, I'm fine with your solution. Can you create an issue to keep it in mind? Thank you

jhlegarreta:
I have checked the build logs and although the [front page of the build of interest shows 0 warnings](https://github.com/dipy/dipy/pull/2200/checks?check_run_id=929793361), the [raw logs show that warning is still there](https://dev.azure.com/dipy/46625eaf-7255-4a2c-9e92-befcdbdca2a8/_apis/build/builds/519/logs/99), so looks like the flag is not doing the expected job. That is, the flag only tells the compiler not to mark the warning as an error, which was not either before, and hence warnings are still reported, which makes sense, and having a closer look at the flag doc, it is the expected behavior.

And I assume the same will happen if we try to silence all deprecation warnings: they will still be reported but not marked as errors.

So we will need to live with the warnings until the Cython folks come up with a version that fixes it and we set it as the min version.

@skoudoro so I can open the issue telling that when we detect that it has been fixed in a Cython release we may want to set that version as the min version, and close this PR without being merged.

arokem:
It's so weird that the CI indicator is stuck on yellow for this PR (at least that's what I am seeing right now). Looks like everything came up green, including [that one Travis run](https://travis-ci.com/github/dipy/dipy/builds/177884259). 

Nevertheless, looks like this should be closed? I'll close it based on the previous conversation, but @jhlegarreta should feel free and reopen if I missed something.

jhlegarreta:
@arokem sounds reasonable. Rather that CIs status, what is worrying about these is the multiple warnings being logged in the build logs. For now, warnings are not signaled as errors. However, fixing them seems not to be immediate or clean based on my findings and attempts, and maybe a deep discussion and wide agreement is required to opt for silencing the `NumPy` API deprecation warnings, so thanks for the clean-up Ariel.

PR #2203, opened 2020-08-17
End of line period in sft

frheault:
Every logging and Raise message now has a period at the end of the line. It was a mix before, which annoyed me.
One bounding box raise message also hed missing space.

After 2 months of testing "in the field" I now realized that the epsilon for the `remove_invalid_streamlines()` function is too low. At 1e-6 it does not remove enough streamlines WHEN the affine of the data has a mild rotational component and crash (often) at saving due to floating-point difference. I never have problems with 1e-3, which is what I expect after a call to `remove_invalid_streamlines()`

Any typo anyone saw in the last few months is welcome: @GuillaumeTh @ppoulin91 @AntoineTheb @fullbat

### DIFF
--- a/dipy/io/stateful_tractogram.py
+++ b/dipy/io/stateful_tractogram.py
@@ -110,7 +110,7 @@ class StatefulTractogram(object):
                            'must be specified separately.')
             logger.warning('To copy the state from another StatefulTractogram '
                            'you may want to use the function from_sft '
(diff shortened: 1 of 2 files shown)


AntoineTheb:
LGTM :100:

skoudoro:
Thank you @frheault

PR #2204, opened 2020-08-18
from Renderer to scene

skoudoro:
In FURY, `renderer` is deprecated since version 0.2. This is currently the version 0.6 and it will beak our current examples in the next update so this is a simple update

### DIFF
--- a/dipy/sims/phantom.py
+++ b/dipy/sims/phantom.py
@@ -246,7 +246,7 @@ if __name__ == "__main__":
     # """
 
     # from dipy.viz import window, actor
-    # ren = window.Renderer()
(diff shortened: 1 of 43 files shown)


arokem:
LGTM! 

It would be nice to add requirements files for various optional dependencies, so that we can use pip to install extras, a la: `pip install dipy[fury]` or `pip install dipy[all]`.

skoudoro:
> It would be nice to add requirements files for various optional dependencies, so that we can use pip to install extras, a la: pip install dipy[fury] or pip install dipy[all]

I agree 👍 , but do you know how to set it up?

arokem:
Yes. Maybe I can try to set that up on a separate PR, after merging this one.

PR #2205, opened 2020-08-18
Add optional installs for different settings.

arokem:
Follow up on https://github.com/dipy/dipy/pull/2204#issuecomment-675530197 iin #2204:

This would entail that installing with:

`pip install dipy[all]`

would now include all the optional dependencies specified in this dict. 

Question: what else do we need to include in this dict (if anything)?

### DIFF
--- a/dipy/info.py
+++ b/dipy/info.py
@@ -108,3 +108,23 @@ REQUIRES            = ["numpy (>=%s)" % NUMPY_MIN_VERSION,
                        "nibabel (>=%s)" % NIBABEL_MIN_VERSION,
                        "h5py (>=%s)" % H5PY_MIN_VERSION,
                        "packaging (>=%s)" % PACKAGING_MIN_VERSION]
+EXTRAS_REQUIRE = {
+    "doc": [
+        "fury>=0.6",
+        "sphinx",
+    ],
+    "viz": [
+        "fury>=0.6"
+    ],
+    "ml": [
+        "scikit_learn",
+        "pandas",
+        "statsmodels"
+        "tables",
+        "tensorflow"
+    ]
+
+}
+
+EXTRAS_REQUIRE["all"] = list(set([a[i] for a in list(EXTRAS_REQUIRE.values())
+                                  for i in range(len(a))]))
--- a/dipy/info.py
+++ b/dipy/info.py
@@ -109,12 +109,30 @@ REQUIRES            = ["numpy (>=%s)" % NUMPY_MIN_VERSION,
                        "h5py (>=%s)" % H5PY_MIN_VERSION,
                        "packaging (>=%s)" % PACKAGING_MIN_VERSION]
 EXTRAS_REQUIRE = {
+    "test":[
+        "pytest",
+        "coverage",
+        "coveralls",
+        "codecov"
+    ]
     "doc": [
-        "fury>=0.6",
-        "sphinx",
+        "cython",
+        "numpy",
+        "scipy",
+        "nibabel>=3.0.0",
+        "h5py",
+        "cvxpy",
+        "pandas",
+        "tables",
+        "matplotlib",
+        "fury>=0.6"
+        "scikit-learn",
+        "scikit-image",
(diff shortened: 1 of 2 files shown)


dipy/info.py
```diff
@@ -108,3 +108,23 @@
                        "nibabel (>=%s)" % NIBABEL_MIN_VERSION,
                        "h5py (>=%s)" % H5PY_MIN_VERSION,
                        "packaging (>=%s)" % PACKAGING_MIN_VERSION]
+EXTRAS_REQUIRE = {
+    "doc": [
+        "fury>=0.6",
+        "sphinx",
```
skoudoro, line 114: For the doc, there is more: you can look at: https://github.com/dipy/dipy/blob/master/doc/doc-requirements.txt

dipy/info.py
```diff
@@ -108,3 +108,23 @@
...
+        "pandas",
+        "statsmodels"
+        "tables",
+        "tensorflow"
+    ]
+
+}
+
+EXTRAS_REQUIRE["all"] = list(set([a[i] for a in list(EXTRAS_REQUIRE.values())
+                                  for i in range(len(a))]))
```
skoudoro, line 130: Nice for the maintenance!

dipy/info.py
```diff
@@ -108,3 +108,41 @@
                        "nibabel (>=%s)" % NIBABEL_MIN_VERSION,
                        "h5py (>=%s)" % H5PY_MIN_VERSION,
                        "packaging (>=%s)" % PACKAGING_MIN_VERSION]
+EXTRAS_REQUIRE = {
+    "test":[
+        "pytest",
+        "coverage",
+        "coveralls",
+        "codecov"
+    ]
```
skoudoro, line 117: it seems to miss a comma here. Can you fix the pep8 as well? Thank you @arokem

arokem:
Looks like extras_require is already being used [here](https://github.com/dipy/dipy/blob/master/setup.py#L56), but I need to take a closer look to understand what circumstances this is used and whether this is even relevant.

skoudoro:
Since `setuptools` is not on the standard library (but recommended), our `setup.py` manage `distutils`.  Not sure if `distutils` manage extra_requires.
I think, this is why it is in this conditional `if`. You can just move your change in this place. Most of the time `setuptools` is used.

arokem:
We could add a pyproject.toml to make the project PEP517/PEP518 compliant
(see
https://medium.com/@grassfedcode/pep-517-and-518-in-plain-english-47208ca8b7a6).
Then we can always use setuptools (I believe). Do you want me to try do
that here on this PR? I think that would significantly simplify our
setup.py file.

skoudoro:
> We could add a pyproject.toml to make the project PEP517/PEP518 compliant

Not sure we want that after Nibabel experience. It seems to make more complicated to use `pip install -e`.More information: https://github.com/nipy/nibabel/issues/859.

skoudoro:
Can I go ahead and merge this PR? Any other comments?

arokem:
>> We could add a pyproject.toml to make the project PEP517/PEP518 compliant

> Not sure we want that after Nibabel experience. It seems to make more complicated to use pip install -e.More information: nipy/nibabel#859.

Oh wow - I missed that. Let's hold off then. Otherwise, this is ready on my end.

skoudoro:
Thanks you @arokem for this feature! merging

effigies:
> > > We could add a pyproject.toml to make the project PEP517/PEP518 compliant
> 
> > Not sure we want that after Nibabel experience. It seems to make more complicated to use pip install -e.More information: [nipy/nibabel#859](https://github.com/nipy/nibabel/issues/859).
> 
> Oh wow - I missed that. Let's hold off then. Otherwise, this is ready on my end.

Just a note that it's specifically `--user --editable` that becomes a problem. Not sure how common this use case is.

skoudoro:
Thank you for the additional information @effigies. This is a common habit DIPY developer, we recommend it a lot before.

Issue #392, opened 2014-07-23
mrtrix 0.3 default basis is different from mrtrix 0.2


samuelstjean:
So according to this https://github.com/jdtournier/mrtrix3/wiki/Orthonormal-SH-basis, the new mrtrix 0.3 will have an orthornormal basis. That means that any function using the -basis option will either
1. Need to call it mrtrix 0.2 basis to distinguish the old from the new one.
2. Implement a mrtrix 0.2 and mrtrix 0.3 basis since they are not the same.

They now say it will be orthornormal, so if it fits with dipy basis all is good, else the new basis will have to be implemented when the 0.3 release is more mainstream.

jhlegarreta:
Partially addressed by #1657.

I guess explaining the MRtrix0.2 vs. MRtrix3 basis does no longer make sense, since DIPY uses MRtrix3, and stopped using MRtrix0.2.

ToDo's:
- [ ] Explain **why the sqrt(2) is relevant**.
- [ ] Add a **function to infer the SH basis used by the data**, as it has been suggested.

jhlegarreta:
True, a confusion on my side. Edited/removed the part. The factor is not actually present in the code as you said in #1657, and as shown in the new doc `sh_basis.rst` doc.

scott-trinkle:
Raised a separate issue (https://github.com/nipy/dipy/issues/1782), but it looks like I should move the discussion here. 

I think something is still off between the default DIPY and MRtrix3 bases. It doesn't look like the current DIPY `basis_type` implementations are accounting for the following quote from the [`dwi2fod` documentation](https://mrtrix.readthedocs.io/en/latest/reference/commands/dwi2fod.html) in MRtrix:

> Note that the spherical harmonics equations used here differ slightly from those conventionally used, in that the (-1)^m factor has been omitted. 

So currently, both software packages include the sqrt(2) scaling factor, but have opposite conventions for taking the Real vs. Imaginary components for m > 0 or m < 0 (which amounts to just reordering the way in which the SH coefficients are stored), and MRtrix does not have the (-1)^m factor included in `scipy.special.sph_harm`, which amounts to certain coefficients being off by a factor of -1. 

Currently, the `basis_type='tournier07'` argument calls to the `real_sym_sh_mrtrix` function, which correctly reorders the coefficients, but does not include the sqrt(2) factor and does not account for the (-1)^m factor. 

[...]

Issue #388, opened 2014-07-08
SDT: odf_deconv from spherical values or SH coefficients?


mdesco:
Need to investigate the stability of the deconvolution and ODF sharpening if we start from ODF values on the sphere or ODF sh coefficients. Especially important to check in the case of less data than coefficients to estimate.

skoudoro:
Hi @CHrlS98,

Since you are looking to the sh basis, Could you also look at this issue please?

You can ask @mdesco for more information. 

If it is not relevant anymore, feel free to close this issue. 

Thank you

CHrlS98:
Hi @mdesco,

I'm not sure I understand the full extent of the issue. We talked about not allowing the user to convert from SF coefficients to SH coefficients when there is less data than coefficients to estimate. I could add a check in function `sh_to_sf` but I'm not sure it would fully cover the scope of the current issue. Maybe this is material for a standalone PR other than #2206?

Thanks

CHrlS98:
Hi @skoudoro @mdesco ,
This issue was not solved by PR #2206.
It needs to be reopened.

Thanks

skoudoro:
I just reopen it, sorry about that.

PR #2206, opened 2020-08-26
NF: Update definitions of SH bases and documentation

CHrlS98:
Update of spherical harmonics and their associated documentation for better agreement with literature as well as MRtrix. This pull request mostly consists of:

1. A new `descoteaux07` basis corresponding exactly to the one from the cited paper, that is without an absolute value for negative values of `m`;
2. A new `tournier07` basis corresponding to the new MRtrix3 definition;
3. New functions `real_sh_descoteaux` and `real_sh_tournier` to replace `real_sym_sh_mrtrix` and `real_sym_sh_basis`. These function allow to chose between the new bases (by default) and the _legacy_ bases. Furthermore, they allow the user to use a full SH basis by setting `is_full_basis=True`;
4. Deprecation warnings for the _legacy_ bases and functions `real_sym_sh_mrtrix` and `real_sym_sh_basis`;
5. Utility methods for converting to and from _legacy_ bases;
6. An updated documentation on SH theory giving further explanation regarding the differences between implementations and definitions.

Looking forward to your feedback,
Have a great day!

### DIFF
--- a/dipy/reconst/shm.py
+++ b/dipy/reconst/shm.py
@@ -1141,7 +1141,7 @@ def sh_to_sf(sh, sphere, sh_order=4, basis_type=None,
         Default: False
     use_legacy_definition: bool, optional
         True to use a legacy basis definition for backward compatibility
-        with previous tournier07 and descoteaux07 implementations. 
+        with previous tournier07 and descoteaux07 implementations.
         Default: False
 
     Returns
--- a/dipy/reconst/shm.py
+++ b/dipy/reconst/shm.py
@@ -280,7 +280,7 @@ def real_sh_tournier_from_index(m, n, theta, phi, legacy=True):
            framework for medical image processing and visualisation.
            NeuroImage. 2019 Nov 15;202:116-137.
     """
-    # In the case where m < 0, Tournier basis consider |m|
+    # In the m < 0 case, Tournier basis considers |m|
     sh = spherical_harmonics(np.abs(m), n, phi, theta)
     real_sh = np.where(m < 0, sh.imag, sh.real)
 
--- a/dipy/reconst/shm.py
+++ b/dipy/reconst/shm.py
@@ -695,7 +695,6 @@ def _gfa_sh(coef, sh0_index=0):
         The coefficients, using a normalized sh basis, that represent each odf.
     sh0_index : int, optional
         The index of the coefficient associated with the 0th order sh harmonic.
-        Default: 0
 
     Returns
     -------
@@ -881,7 +880,6 @@ class SphHarmFit(OdfFit):
 
         S0 : float array
            The mean non-diffusion-weighted signal in each voxel.
-           Default: 1.0 in all voxels
         """
         if not hasattr(self.model, 'predict'):
             msg = "This model does not have prediction implemented yet"
@@ -1363,12 +1361,11 @@ def anisotropic_power(sh_coeffs, norm_factor=0.00001, power=2,
         A ndarray where the last dimension is the
         SH coefficients estimates for that voxel.
     norm_factor: float, optional
-        The value to normalize the ap values. Default is 10^-5.
+        The value to normalize the ap values.
     power : int, optional
-        The degree to which power maps are calculated. Default: 2.
+        The degree to which power maps are calculated.
     non_negative: bool, optional
         Whether to rectify the resulting map to be non-negative.
-        Default: True.
 
     Returns
     -------
@@ -1531,7 +1528,7 @@ def convert_sh_to_legacy(sh_coeffs, sh_basis, full_basis=False):
         ``descoteaux07`` for the Descoteaux 2007 [1]_ basis.
     full_basis: bool, optional
         True if the input SH basis includes both even and odd
-        order SH functions. Default: False
+        order SH functions.
 
     Returns
     -------
--- a/dipy/reconst/shm.py
+++ b/dipy/reconst/shm.py
@@ -115,12 +115,12 @@ def sh_to_rh(r_sh, m, n):
     return r_rh
 
 
-def gen_dirac(m, n, theta, phi):
+def gen_dirac(m, n, theta, phi, legacy=True):
     """ Generate Dirac delta function orientated in (theta, phi) on the sphere
 
     The spherical harmonics (SH) representation of this Dirac is returned as
-    coefficients to spherical harmonic functions produced by
-    `shm.real_sph_harm`.
+    coefficients to spherical harmonic functions produced from ``descoteaux07``
+    basis.
 
     Parameters
     ----------
@@ -137,7 +137,7 @@ def gen_dirac(m, n, theta, phi):
 
     See Also
     --------
-    shm.real_sph_harm, shm.real_sym_sh_basis
+    shm.real_sh_descoteaux_from_index, shm.real_sh_descoteaux
 
     Returns
     -------
@@ -146,7 +146,7 @@ def gen_dirac(m, n, theta, phi):
         `(m + 2) * (m + 1) / 2`.
 
     """
-    return real_sph_harm(m, n, theta, phi)
+    return real_sh_descoteaux_from_index(m, n, theta, phi, legacy=legacy)
 
 
 def spherical_harmonics(m, n, theta, phi, use_scipy=True):
@@ -202,7 +202,7 @@ def spherical_harmonics(m, n, theta, phi, use_scipy=True):
 @deprecate_with_version('dipy.reconst.shm.real_sph_harm is deprecated, '
                         'Please use '
                         'dipy.reconst.shm.real_sh_descoteaux_from_index '
-                        'instead')
+                        'instead', since='1.3', until='2.0')
 def real_sph_harm(m, n, theta, phi):
     """ Compute real spherical harmonics.
 
@@ -238,7 +238,7 @@ def real_sph_harm(m, n, theta, phi):
     return real_sh_descoteaux_from_index(m, n, theta, phi, legacy=True)
 
 
-def real_sh_tournier_from_index(m, n, theta, phi, legacy=False):
+def real_sh_tournier_from_index(m, n, theta, phi, legacy=True):
     """ Compute real spherical harmonics as initially defined in Tournier
     2007 [1]_ then updated in MRtrix3 [2]_, where the real harmonic $Y^m_n$
     is defined to be:
@@ -295,7 +295,7 @@ def real_sh_tournier_from_index(m, n, theta, phi, legacy=False):
     return real_sh
 
 
-def real_sh_descoteaux_from_index(m, n, theta, phi, legacy=False):
+def real_sh_descoteaux_from_index(m, n, theta, phi, legacy=True):
     """ Compute real spherical harmonics as in Descoteaux et al. 2007 [1]_,
     where the real harmonic $Y^m_n$ is defined to be:
 
@@ -350,7 +350,7 @@ def real_sh_descoteaux_from_index(m, n, theta, phi, legacy=False):
 
 def real_sh_tournier(sh_order, theta, phi,
                      full_basis=False,
-                     legacy=False):
+                     legacy=True):
     """ Compute real spherical harmonics as initially defined in Tournier
     2007 [1]_ then updated in MRtrix3 [2]_, where the real harmonic $Y^m_n$
     is defined to be:
@@ -409,7 +409,7 @@ def real_sh_tournier(sh_order, theta, phi,
 
 def real_sh_descoteaux(sh_order, theta, phi,
                        full_basis=False,
-                       legacy=False):
+                       legacy=True):
     """ Compute real spherical harmonics as in Descoteaux et al. 2007 [1]_,
     where the real harmonic $Y^m_n$ is defined to be:
 
@@ -463,7 +463,8 @@ def real_sh_descoteaux(sh_order, theta, phi,
 
 
 @deprecate_with_version('dipy.reconst.shm.real_sym_sh_mrtix is deprecated, '
-                        'Please use dipy.reconst.shm.real_sh_tournier instead')
+                        'Please use dipy.reconst.shm.real_sh_tournier instead',
+                        since='1.3', until='2.0')
 def real_sym_sh_mrtrix(sh_order, theta, phi):
     """
     Compute real symmetric spherical harmonics as in Tournier 2007 [2]_, where
@@ -513,7 +514,7 @@ def real_sym_sh_mrtrix(sh_order, theta, phi):
 
 @deprecate_with_version('dipy.reconst.shm.real_sym_sh_basis is deprecated, '
                         'Please use dipy.reconst.shm.real_sh_descoteaux '
-                        'instead')
+                        'instead', since='1.3', until='2.0')
 def real_sym_sh_basis(sh_order, theta, phi):
     """Samples a real symmetric spherical harmonic basis at point on the sphere
 
@@ -1104,7 +1105,7 @@ class ResidualBootstrapWrapper(object):
 
 
 def sf_to_sh(sf, sphere, sh_order=4, basis_type=None, full_basis=False,
-             legacy=False, smooth=0.0):
+             legacy=True, smooth=0.0):
     """Spherical function to spherical harmonics (SH).
 
     Parameters
@@ -1168,7 +1169,7 @@ def sf_to_sh(sf, sphere, sh_order=4, basis_type=None, full_basis=False,
 
 
 def sh_to_sf(sh, sphere, sh_order=4, basis_type=None,
-             full_basis=False, legacy=False):
+             full_basis=False, legacy=True):
     """Spherical harmonics (SH) to spherical function (SF).
 
     Parameters
@@ -1228,7 +1229,7 @@ def sh_to_sf(sh, sphere, sh_order=4, basis_type=None,
 
 
 def sh_to_sf_matrix(sphere, sh_order=4, basis_type=None, full_basis=False,
-                    legacy=False, return_inv=True, smooth=0):
+                    legacy=True, return_inv=True, smooth=0):
     """ Matrix that transforms Spherical harmonics (SH) to spherical
     function (SF).
 
--- a/dipy/reconst/shm.py
+++ b/dipy/reconst/shm.py
@@ -24,13 +24,13 @@ following form: Y.T = x.T B.T, or in python syntax data = np.dot(sh_coef, B.T)
 where data is Y.T and sh_coef is x.T.
 """
 
+from warnings import warn
 import numpy as np
 import scipy.special as sps
 
 from numpy.random import randint
 
 from dipy.utils.deprecator import deprecate_with_version
-
 from dipy.reconst.odf import OdfModel, OdfFit
 from dipy.core.geometry import cart2sphere
 from dipy.core.onetime import auto_attr
@@ -51,8 +51,8 @@ def forward_sdeconv_mat(r_rh, n):
     ----------
     r_rh : ndarray
         Rotational harmonics coefficients for the single fiber response
-        function. Each element `rh[i]` is associated with spherical harmonics
-        of degree `2*i`.
+        function. Each element ``rh[i]`` is associated with spherical harmonics
+        of degree ``2*i``.
     n : ndarray
         The order of spherical harmonic function associated with each row of
         the deconvolution matrix. Only even orders are allowed
@@ -73,7 +73,7 @@ def sh_to_rh(r_sh, m, n):
     """ Spherical harmonics (SH) to rotational harmonics (RH)
 
     Calculate the rotational harmonic decomposition up to
-    harmonic order `n`, degree `m` for an axially and antipodally
+    harmonic order ``n``, degree ``m`` for an axially and antipodally
     symmetric function. Note that all ``m != 0`` coefficients
     will be ignored as axial symmetry is assumed. Hence, there
     will be ``(sh_order/2 + 1)`` non-zero coefficients.
@@ -85,10 +85,10 @@ def sh_to_rh(r_sh, m, n):
         These coefficients must correspond to the real spherical harmonic
         functions produced by `shm.real_sph_harm`.
     m : ndarray (N,)
-        The order of the spherical harmonic function associated with each
+        The degree of the spherical harmonic function associated with each
         coefficient.
     n : ndarray (N,)
-        The degree of the spherical harmonic function associated with each
+        The order of the spherical harmonic function associated with each
         coefficient.
 
     Returns
@@ -130,10 +130,10 @@ def gen_dirac(m, n, theta, phi):
     n : ndarray (N,)
         The order of the spherical harmonic function associated with each
         coefficient.
-    theta : float [0, 2*pi]
-        The azimuthal (longitudinal) coordinate.
-    phi : float [0, pi]
+    theta : float [0, pi]
         The polar (colatitudinal) coordinate.
+    phi : float [0, 2*pi]
+        The azimuthal (longitudinal) coordinate.
 
     See Also
     --------
@@ -166,19 +166,27 @@ def spherical_harmonics(m, n, theta, phi, use_scipy=True):
     phi : float [0, pi]
         The polar (colatitudinal) coordinate.
     use_scipy : bool, optional
-        if True, use scipy implementation. Default: True
+        If True, use scipy implementation.
 
     Returns
     -------
     y_mn : complex float
-        The harmonic $Y^m_n$ sampled at `theta` and `phi`.
+        The harmonic $Y^m_n$ sampled at ``theta`` and ``phi``.
 
     Notes
     -----
     This is a faster implementation of scipy.special.sph_harm for
     scipy version < 0.15.0. For scipy 0.15 and onwards, we use the scipy
-    implementation of the function
+    implementation of the function.
 
+    The usual definitions for ``theta` and `phi`` used in DIPY are interchanged
+    in the method definition to agree with the definitions in
+    scipy.special.sph_harm, where `theta` represents the azimuthal coordinate
+    and `phi` represents the polar coordinate.
+
+    Altough scipy uses a naming convention where ``m`` is the order and ``n``
+    is the degree of the SH, the opposite of DIPY's, their definition for both
+    parameters is the same as ours, with ``n >= 0`` and ``|m| <= n``.
     """
     if use_scipy:
         return sps.sph_harm(m, n, theta, phi, dtype=complex)
@@ -190,9 +198,12 @@ def spherical_harmonics(m, n, theta, phi, use_scipy=True):
     val = val * np.exp(1j * m * theta)
     return val
 
-
+@deprecate_with_version('dipy.reconst.shm.real_sph_harm is deprecated, '
+                        'Please use '
+                        'dipy.reconst.shm.real_sh_descoteaux_from_index '
+                        'instead')
 def real_sph_harm(m, n, theta, phi):
-    r""" Compute real spherical harmonics.
+    """ Compute real spherical harmonics.
 
     Where the real harmonic $Y^m_n$ is defined to be:
 
@@ -209,10 +220,10 @@ def real_sph_harm(m, n, theta, phi):
         The degree of the harmonic.
     n : int ``>= 0``
         The order of the harmonic.
-    theta : float [0, 2*pi]
-        The azimuthal (longitudinal) coordinate.
-    phi : float [0, pi]
+    theta : float [0, pi]
         The polar (colatitudinal) coordinate.
+    phi : float [0, 2*pi]
+        The azimuthal (longitudinal) coordinate.
 
     Returns
     --------
@@ -223,20 +234,124 @@ def real_sph_harm(m, n, theta, phi):
     --------
     scipy.special.sph_harm
     """
-    # dipy uses a convention for theta and phi that is reversed with respect to
-    # function signature of scipy.special.sph_harm
+    return real_sh_descoteaux_from_index(m, n, theta, phi, legacy=True)
+
+
+def real_sh_tournier_from_index(m, n, theta, phi, legacy=False):
+    """ Compute real spherical harmonics as initially defined in Tournier
+    2007 [1]_ then updated in MRtrix3 [2]_, where the real harmonic $Y^m_n$
+    is defined to be:
+
+        Real($Y^m_n$) * sqrt(2)      if m > 0
+        $Y^0_n$                      if m = 0
+        Imag($Y^|m|_n$) * sqrt(2)    if m < 0
+
+    This may take scalar or array arguments. The inputs will be broadcasted
+    against each other.
+
+    Parameters
+    ----------
+    m : int ``|m| <= n``
+        The degree of the harmonic.
+    n : int ``>= 0``
+        The order of the harmonic.
+    theta : float [0, pi]
+        The polar (colatitudinal) coordinate.
+    phi : float [0, 2*pi]
+        The azimuthal (longitudinal) coordinate.
+    legacy: bool, optional
+        If true, uses MRtrix 0.2 SH basis definition, where the ``sqrt(2)``
+        factor is omitted. Else, uses the MRtrix 3 definition presented above.
+
+    Returns
+    -------
+    real_sh : real float
+        The real harmonics $Y^m_n$ sampled at ``theta`` and ``phi``.
+
+    References
+    ----------
+    .. [1] Tournier J.D., Calamante F. and Connelly A. Robust determination
+           of the fibre orientation distribution in diffusion MRI:
+           Non-negativity constrained super-resolved spherical deconvolution.
+           NeuroImage. 2007;35(4):1459-1472.
+    .. [2] Tournier J-D, Smith R, Raffelt D, Tabbara R, Dhollander T,
+           Pietsch M, et al. MRtrix3: A fast, flexible and open software
+           framework for medical image processing and visualisation.
+           NeuroImage. 2019 Nov 15;202:116-137.
+    """
+    # In the case where m < 0, Tournier basis consider |m|
     sh = spherical_harmonics(np.abs(m), n, phi, theta)
+    real_sh = np.where(m < 0, sh.imag, sh.real)
+
+    if not legacy:
+        # The Tournier basis from MRtrix3 is normalized
+        real_sh *= np.where(m == 0, 1., np.sqrt(2))
+    else:
+        warn('The legacy tournier07 basis is outdated and will be deprecated '
+             'in a future release of DIPY. Consider using the new tournier07 '
+             'basis.', category=PendingDeprecationWarning)
+
+    return real_sh
+
+
+def real_sh_descoteaux_from_index(m, n, theta, phi, legacy=False):
+    """ Compute real spherical harmonics as in Descoteaux et al. 2007 [1]_,
+    where the real harmonic $Y^m_n$ is defined to be:
+
+        Imag($Y^m_n$) * sqrt(2)      if m > 0
+        $Y^0_n$                      if m = 0
+        Real($Y^m_n$) * sqrt(2)      if m < 0
+
+    This may take scalar or array arguments. The inputs will be broadcasted
+    against each other.
+
+    Parameters
+    ----------
+    m : int ``|m| <= n``
+        The degree of the harmonic.
+    n : int ``>= 0``
+        The order of the harmonic.
+    theta : float [0, pi]
+        The polar (colatitudinal) coordinate.
+    phi : float [0, 2*pi]
+        The azimuthal (longitudinal) coordinate.
+    legacy: bool, optional
+        If true, uses DIPY's legacy descoteaux07 implementation (where |m|
+        is used for m < 0). Else, implements the basis as defined in
+        Descoteaux et al. 2007 (without the absolute value).
+
+    Returns
+    -------
+    real_sh : real float
+        The real harmonic $Y^m_n$ sampled at ``theta`` and ``phi``.
+
+    References
+    ----------
+     .. [1] Descoteaux, M., Angelino, E., Fitzgibbons, S. and Deriche, R.
+           Regularized, Fast, and Robust Analytical Q-ball Imaging.
+           Magn. Reson. Med. 2007;58:497-510.
+    """
+    if legacy:
+        # In the case where m < 0, legacy descoteaux basis considers |m|
+        warn('The legacy descoteaux07 SH basis is outdated and will be '
+             'deprecated in a future DIPY release. Consider using the new '
+             'descoteaux07 basis.', category=PendingDeprecationWarning)
+        sh = spherical_harmonics(np.abs(m), n, phi, theta)
+    else:
+        # In the cited paper, the basis is defined without the absolute value
+        sh = spherical_harmonics(m, n, phi, theta)
 
     real_sh = np.where(m > 0, sh.imag, sh.real)
     real_sh *= np.where(m == 0, 1., np.sqrt(2))
+
     return real_sh
 
 
 def real_sh_tournier(sh_order, theta, phi,
-                     is_full_basis=False,
-                     is_legacy=False):
+                     full_basis=False,
+                     legacy=False):
     """ Compute real spherical harmonics as initially defined in Tournier
-    2007 [1]_ then updated in MRtrix [2]_, where the real harmonic $Y^m_n$
+    2007 [1]_ then updated in MRtrix3 [2]_, where the real harmonic $Y^m_n$
     is defined to be:
 
         Real($Y^m_n$) * sqrt(2)      if m > 0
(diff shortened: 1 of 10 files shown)


CHrlS98:
Oups! It looks like I broke some tests by changing the default SH bases. It'll be fixed soon.

dipy/reconst/shm.py
```diff
@@ -63,15 +65,15 @@ def forward_sdeconv_mat(r_rh, n):
...
+        raise ValueError("n has odd orders, expecting only even orders")
     return np.diag(r_rh[n // 2])
 
 
 def sh_to_rh(r_sh, m, n):
     """ Spherical harmonics (SH) to rotational harmonics (RH)
 
     Calculate the rotational harmonic decomposition up to
-    harmonic order `m`, degree `n` for an axially and antipodally
+    harmonic order `n`, degree `m` for an axially and antipodally
```
skoudoro, line 76: Could you update the description of the parameters on this function

dipy/reconst/shm.py
```diff
@@ -156,15 +158,15 @@ def spherical_harmonics(m, n, theta, phi, use_scipy=True):
     Parameters
     ----------
     m : int ``|m| <= n``
-        The order of the harmonic.
-    n : int ``>= 0``
         The degree of the harmonic.
+    n : int ``>= 0``
+        The order of the harmonic.
```
skoudoro, line 163: Can you look at Scipy description? I think, we will start to confuse the user.

skoudoro (commented):
Hi @CHrlS98,

Thank you for this nice PR.

In Scipy, we have the [following description](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.sph_harm.html ):

 ```
m -> Order of the harmonic (int); must have |m| <= n.
n -> Degree of the harmonic (int); must have n >= 0. This is often denoted by l (lower case L) in descriptions of spherical harmonics.
```

However, your change (and the litterature) introduces the opposite. 

```
n -> Order of the harmonic (int)
m -> Degree of the harmonic (int)
```

This is really confusing and I am not sure anymore if the following call `scipy.sph_harm(m, n, theta, phi, dtype=complex)` [here](https://github.com/dipy/dipy/blob/master/dipy/reconst/shm.py#L182) is correctly called with your update.

it seems that Dipy was following Scipy convention instead of the literature. So the documentation was misleading but align with Scipy convention.

I would love to ear your opinions to clarify this point before going ahead in the review. Other opinions are welcomed too

Thank you

File: dipy/reconst/shm.py [hunk 3]
```diff
@@ -230,164 +232,205 @@ def real_sph_harm(m, n, theta, phi):
...
+           NeuroImage. 2007;35(4):1459-1472.
 
     """
-    m, n = sph_harm_ind_list(sh_order)
-    phi = np.reshape(phi, [-1, 1])
-    theta = np.reshape(theta, [-1, 1])
-
-    real_sh = real_sph_harm(m, n, theta, phi)
-    return real_sh, m, n
+    deprecate_with_version('Function real_sym_sh_mrtix has been replaced \
```
skoudoro, line 426: I recommend to : 
- use it as a decorator
- add from and until when

look [here](https://github.com/dipy/dipy/pull/2191/files#diff-3d2aafe93b8f41be9cf3521aa9d5d194R20-R24) as an example

dipy/reconst/shm.py
```diff
@@ -426,39 +469,35 @@ def real_full_sh_basis(sh_order, theta, phi):
            Magn. Reson. Med. 2007;58:497-510.
 
     """
-    m, n = sph_harm_full_ind_list(sh_order)
-    phi = np.reshape(phi, [-1, 1])
-    theta = np.reshape(theta, [-1, 1])
-
-    real_sh = real_sph_harm(m, n, theta, phi)
-    return real_sh, m, n
+    deprecate_with_version('Function real_sym_sh_basis has been replaced \
```
skoudoro, line 472: I recommend to : 
- use it as a decorator
- add from and until when

look [here](https://github.com/dipy/dipy/pull/2191/files#diff-3d2aafe93b8f41be9cf3521aa9d5d194R20-R24) as an example

dipy/reconst/shm.py
```diff
@@ -230,164 +232,205 @@ def real_sph_harm(m, n, theta, phi):
...
-    real_sh /= np.where(m == 0, 1., np.sqrt(2))
+    # In the case where m < 0, Tournier bases
+    # consider |m| for the SH degree
+    sh = spherical_harmonics(np.abs(m), n, phi, theta)
+
+    real_sh = np.where(m < 0, sh.imag, sh.real)
+    if not is_legacy:
+        real_sh *= np.where(m == 0, 1., np.sqrt(2))
+    else:
+        deprecate_with_version('The legacy tournier07 basis is outdated and \
```
skoudoro, line 301: This is a decorator's initialization. The call is incorrect. Can you add a test to check this? You can look at `test_decorator.py` to see how to call the decorator.

dipy/reconst/shm.py
```diff
@@ -230,164 +232,205 @@ def real_sph_harm(m, n, theta, phi):
...
     theta = np.reshape(theta, [-1, 1])
 
-    m = -m
-    real_sh = real_sph_harm(m, n, theta, phi)
-    real_sh /= np.where(m == 0, 1., np.sqrt(2))
-    return real_sh, m, n
+    if is_legacy:
+        # In the case where m < 0, legacy descoteaux basis
+        # consider |m| for the SH degree
+        deprecate_with_version('The legacy descoteaux07 SH basis is outdated \
```
skoudoro, line 365: same comment as above.

> This is a decorator's initialization. The call is incorrect. Can you add a test to check this? You can look at `test_decorator.py` to see how to call the decorator.

CHrlS98:
Hi @skoudoro ,
Thanks for the feedback. I'll fix the deprecation warnings in an upcoming commit. 

As for the `scipy` definition of `m` and `n`, my understanding is that where the literature labels `m` as the _degree_ and `n` the  _order_, `scipy` inverts these names, with `m` being the _order_ and `n` the _degree_. However, the definitions for `m` and `n` are the same for both `scipy` and `dipy`, with `n >= 0` and `-n <= m <= n`. That makes the call [here](https://github.com/dipy/dipy/blob/c6f6f3f6db27e8b91886e35dc71709ba991bfb43/dipy/reconst/shm.py#L182) correct. 

So, I guess we have to choose what is the least confusing for the user between naming `m` and `n` the same way they are in the literature and naming `m` and `n` the way they are in `scipy`.

I feel like it would be best to define `m` as the degree and `n` as the order since it agrees with the usual definition of spherical harmonics in the literature, making it more straightforward for the user. We could also add a comment saying `scipy` inverts those names just to remove any ambiguity.

If we decide to keep `scipy` convention, I think we should at least add a note to explain the choice of naming `m` the order and `n` the degree.

What do you think?
Thank you!

skoudoro:
> I feel like it would be best to define m as the degree and n as the order since it agrees with the usual definition of spherical harmonics in the literature, making it more straightforward for the user. We could also add a comment saying scipy inverts those names just to remove any ambiguity.

I agree, Let's respect the literature and add a note. 👍

skoudoro:
Could you look and review at this PR @scott-trinkle, @jhlegarreta, @jchoude, and @arokem? it is related to SH basis and follow up your discussion on #392.

Thank you!

dipy/reconst/shm.py
```diff
@@ -230,164 +232,205 @@ def real_sph_harm(m, n, theta, phi):
...
         The maximum degree or the spherical harmonic basis.
     theta : float [0, pi]
         The polar (colatitudinal) coordinate.
     phi : float [0, 2*pi]
         The azimuthal (longitudinal) coordinate.
+    is_full_basis: bool, optional
+        If true, returns a basis including odd order SH functions as well as
+        even order SH functions. Otherwise returns only even order SH
+        functions. Default: False
+    is_legacy: bool, optional
```
arokem, line 261: I prefer to be explicit here. How about calling this keyword argument `mrtrix3`, set to `True` per default?

dipy/reconst/shm.py
arokem: Yeah. I think this should be a warning instead, and no need to use `deprecate_with_version`. 

That said, when do we want to deprecate this? Is there anything else in our code-base that uses the mrtrix 0.2 definition?

dipy/reconst/shm.py
```diff
@@ -230,164 +232,205 @@ def real_sph_harm(m, n, theta, phi):
...
+
     return real_sh, m, n
 
 
-def real_full_sh_mrtrix(sh_order, theta, phi):
-    """
-    Compute real spherical harmonics as in Tournier 2007 [2]_, where the real
-    harmonic $Y^m_n$ is defined to be::
+def real_sh_descoteaux(sh_order, theta, phi,
+                       is_full_basis=False,
```
arokem, line 310:
```suggestion
                       full_basis=False,
```

dipy/reconst/shm.py
```diff
@@ -230,164 +232,205 @@ def real_sph_harm(m, n, theta, phi):
...
+                       is_legacy=False):
```
arokem, line 311:
```suggestion
                       legacy=False):
```

arokem: Small thing, but I don't think that you need the `is_` in these variable names.

dipy/reconst/shm.py
arokem: And again -- the same question: are there other functions in DIPY itself that need the legacy implementation of this function?

dipy/reconst/shm.py
```diff
@@ -1012,7 +1053,8 @@ def __getitem__(self, index):
         return signal
 
 
-def sf_to_sh(sf, sphere, sh_order=4, basis_type=None, smooth=0.0):
+def sf_to_sh(sf, sphere, sh_order=4, basis_type=None, use_full_basis=False,
+             use_legacy_definition=False, smooth=0.0):
```
arokem, line 1057: I would simplify these variable names a bit:

```suggestion
def sf_to_sh(sf, sphere, sh_order=4, basis_type=None, full_basis=False,
             legacy=False, smooth=0.0):
```

dipy/reconst/shm.py
```diff
@@ -1108,22 +1158,27 @@ def sh_to_sf(sh, sphere, sh_order=4, basis_type=None):
...
 
     sf = np.dot(sh, B.T)
 
     return sf
 
 
-def sh_to_sf_matrix(sphere, sh_order=4, basis_type=None, return_inv=True,
-                    smooth=0):
+def sh_to_sf_matrix(sphere, sh_order=4, basis_type=None, use_full_basis=False,
+                    use_legacy_definition=False, return_inv=True, smooth=0):
```
arokem, line 1181: Here as well:

```suggestion
def sh_to_sf_matrix(sphere, sh_order=4, basis_type=None, full_basis=False,
                    legacy=False, return_inv=True, smooth=0):
```

arokem (commented):
I hope this helps clarify this... once and for all. Thanks for putting in the work! 

I had just a couple of small comments. I am not too attached to any of these, so see if they make sense and take them or leave 'em!

doc/theory/sh_basis.rst
```diff
@@ -57,28 +68,42 @@ literature for the computation of the ODF. DIPY implements two of these in the
...
-     \Im(Y_l^{|m|}(\theta, \phi)) & -l \leq m < 0, \\
+     \Im(Y_l^{m}(\theta, \phi)) & -l \leq m < 0, \\
      Y_l^0(\theta, \phi) & m = 0, \\
      \Re(Y_{l}^m(\theta, \phi)) & 0 < m \leq l
      \end{cases}
 
 In both cases, $\Re$ denotes the real part of the spherical harmonic basis, and
-$\Im$ denotes the imaginary part.
+$\Im$ denotes the imaginary part. The SH bases are both orthogonal and real. Moreover,
+the `descoteaux07` basis is orthonormal.
```
jhlegarreta, line 78: Double backticks, right?

dipy/reconst/shm.py
```diff
@@ -52,8 +54,8 @@ def forward_sdeconv_mat(r_rh, n):
         function. Each element `rh[i]` is associated with spherical harmonics
         of degree `2*i`.
```
jhlegarreta, line 55: Double backticks or math mode.

dipy/reconst/shm.py
jhlegarreta, line 76: Double backticks, right?

doc/theory/sh_basis.rst
```diff
@@ -57,28 +68,42 @@ literature for the computation of the ODF. DIPY implements two of these in the
...
+     \sqrt{2}\Im(Y_l^{|m|}(\theta, \phi)) & -l \leq m < 0, \\
+     Y_l^0(\theta, \phi) & m = 0, \\
+     \sqrt{2}\Re(Y_{l}^m(\theta, \phi)) & 0 < m \leq l
+     \end{cases}
+
+As we can see, the differences include a $\sqrt{2}$ factor multiplying the SH functions
+when $m$ is different than 0 as well as an absolute value for the degree $m$ when $m$ is negative.
+The $\sqrt{2}$ factor has been added to the definition in MRtrix3 in order to make the Tournier basis
+orthonormal. Considering the absolute value of $m$ when $m < 0$ only affects the sign of the
+resulting SH functions, due to the relations $-p=|p|$ $\forall p < 0$ and
```
jhlegarreta, line 96: $-p=|p| \; \forall p < 0$ all within the same math command?

doc/theory/sh_basis.rst
jhlegarreta, line 78: `$\Im$ denotes the imaginary part. The SH bases are both orthogonal and real.` looks like a contradiction to me. Can this be explained (e.g. in DIPY/diffusion only the real par is used)?

doc/theory/sh_basis.rst
```diff
@@ -14,9 +14,17 @@ Spherical harmonics are ortho-normal functions defined by:
 
     Y_l^m(\theta, \phi) = \sqrt{\frac{2l + 1}{4 \pi} \frac{(l - m)!}{(l + m)!}} P_l^m( cos \theta) e^{i m \phi}
 
-where $l$ is the band index, $m$ is the order, $P_l^m$ is an associated
-$l$-th degree, $m$-th order Legendre polynomial, and $(\theta, \phi)$ is the
-representation of the direction vector in the spherical coordinate.
+where $l$ is the order, $m$ is the degree, $P_l^m$ is an associated
+$l$-th order, $m$-th degree Legendre polynomial, and $(\theta, \phi)$ is the
+representation of the direction vector in the spherical coordinate. The relation
```
jhlegarreta, line 19: `in the spherical coordinate.` : `in the spherical coordinate system.` or  `in spherical coordinates.`

doc/theory/sh_basis.rst
```diff
@@ -57,28 +68,42 @@ literature for the computation of the ODF. DIPY implements two of these in the
...
+     \end{cases}
+
+As we can see, the differences include a $\sqrt{2}$ factor multiplying the SH functions
+when $m$ is different than 0 as well as an absolute value for the degree $m$ when $m$ is negative.
+The $\sqrt{2}$ factor has been added to the definition in MRtrix3 in order to make the Tournier basis
+orthonormal. Considering the absolute value of $m$ when $m < 0$ only affects the sign of the
+resulting SH functions, due to the relations $-p=|p|$ $\forall p < 0$ and 
+$Y_l^{-m} = (-1)^m\overline{Y_l^{m}}$, and therefore still results in a valid basis.
+
+In practice, a maximum even order $k$ is used to truncated the SH series. By only
```
jhlegarreta, line 99: `is used to truncated the SH series.`: `is used to truncate the SH series.`

jhlegarreta (commented):
Have not gone into the details of it, so the following comments are more formal:
- Kudos to the monumental effort of making this more consistent, documenting and adding testing :100:. Having a single method having a full vs non-full boolean with a default value eases things for users, reduces maintenance and allows to keep track of things more cleanly.
- Can the same philosophy be applied to the `sph_harm_ind_list` vs. `sph_harm_full_ind_list` methods? i.e. Can we have a single method? Their parameters and return values being the same, and although it may involve some work to provide the former behavior with a single function, it would make things more consistent IMHO. Using an additional default `full` boolean parameter and a couple of internal functions may be an easy solution (but better ones may exist).
- If I recall correctly, there have been PRs in the past where some documentation leading to issues had to be fixed (e.g. having orphan triple double quotes). See [here](https://github.com/dipy/dipy/pull/2141/files#diff-01e42570a73c8036f360a36ad6f5e7aeL10).
- Also, according to the Sphinx documentation raw double quotes are only necessary if backslashes are included in a docstring (I guess LaTeX math is excluded from the rule). Maybe these can be done where docstrings have been added or modified in this PR.
- I have commented  about a few single vs. double backticks, but going through all appearances would be required.
- Ariel's [suggestion](https://github.com/dipy/dipy/pull/2206/files#r486011742) has not been made extensive to all appearances of `use_full_basis` and `use_legacy_definition`. But not sure if the `is_legacy` proposal clashes with this [other suggestion](https://github.com/dipy/dipy/pull/2206/files#r486010387). The docstrings are not consistent everywhere with the variable name changes at this point.
- Although this maybe considered as a choice, if an argument is optional, documenting its default value may lead to inconsistencies when it is decided, for some reason, that the default value needs to be changed but the docstring is forgot to be updated. If an argument is optional, its value in the method's signature is enough, avoids the mentioned situation and thus involves less maintenance IMHO.
- IMHO, it would be cleaner to put basis names into an `enum` instead of the current lookup dictionary (although I did perpetuate the dictionary in a previous PR and did not comment about it; sorry). I think the appropriate, essential documentation about the bases could also exist in the documentation to that `enum`, all in one place if it is currently spread throughout the code (beyond the excellent and necessary `rst`).
- Can the commits and PR incorporate the [prefixes that DIPY recommends](https://github.com/dipy/dipy/blob/master/doc/devel/commit_codes.rst) ?
- Other comments can be found in-line.

dipy/reconst/shm.py
```diff
@@ -426,39 +469,35 @@ def real_full_sh_basis(sh_order, theta, phi):
...
-                   "descoteaux07_full": real_full_sh_basis}
+sph_harm_lookup = {None: real_sh_descoteaux,
+                   "tournier07": real_sh_tournier,
+                   "descoteaux07": real_sh_descoteaux}
 
 
 def sph_harm_ind_list(sh_order):
     """
-    Returns the degree (n) and order (m) of all the symmetric spherical
+    Returns the degree (m) and order (n) of all the symmetric spherical
```
jhlegarreta, line 485: Double backticks for `m` and `n`?

CHrlS98:
Hi,
First of all, thanks to all for your useful comments. I'll address them in upcoming commits. 

To answer @jhlegarreta's [comments](https://github.com/dipy/dipy/pull/2206#pullrequestreview-486003361) :

> * Can the same philosophy be applied to the `sph_harm_ind_list` vs. `sph_harm_full_ind_list` methods? i.e. Can we have a single method? Their parameters and return values being the same, and although it may involve some work to provide the former behavior with a single function, it would make things more consistent IMHO. Using an additional default full boolean parameter and a couple of internal functions may be an easy solution (but better ones may exist).

I agree. It does sound like a clean and easy solution to me. I'll add a boolean parameter to `sph_harm_ind_list` and remove `sph_harm_full_ind_list`.

> * If I recall correctly, there have been PRs in the past where some documentation leading to issues had to be fixed (e.g. having orphan triple double quotes). See [here](https://github.com/dipy/dipy/pull/2141/files#diff-01e42570a73c8036f360a36ad6f5e7aeL10).

> * Also, according to the Sphinx documentation raw double quotes are only necessary if backslashes are included in a docstring (I guess LaTeX math is excluded from the rule). Maybe these can be done where docstrings have been added or modified in this PR.

I want to make sure I understand correctly these ones. Your suggestion is simply that we should take advantage of this PR to fix docstrings issues in the changed files?

> * Ariel's [suggestion](https://github.com/dipy/dipy/pull/2206/files#r486011742) has not been made extensive to all appearances of `use_full_basis` and `use_legacy_definition`. But not sure if the `is_legacy` proposal clashes with this [other suggestion](https://github.com/dipy/dipy/pull/2206/files#r486010387). The docstrings are not consistent everywhere with the variable name changes at this point.

I think we would have to decide if we rename `legacy` to `mrtrix3` in methods using `tournier` basis. The thing is methods such as `sh_to_sf` and `sh_to_sf_matrix` can be called with the `tournier` or `descoteaux` basis. For this reason, I think we should stick with the same parameter name as it prevents us from having to explain that `legacy=True` corresponds to `mrtrix3=False`. It would also feel more consistent through the whole file. What do you think? Maybe @arokem would like to weight in?

> * IMHO, it would be cleaner to put basis names into an `enum` instead of the current lookup dictionary (although I did perpetuate the dictionary in a previous PR and did not comment about it; sorry). I think the appropriate, essential documentation about the bases could also exist in the documentation to that `enum`, all in one place if it is currently spread throughout the code (beyond the excellent and necessary `rst`).

Is your suggestion we simply replace the `basis_type` string by an `enum`? If so, I agree it would be cleaner and easier to maintain as it removes the need to document the available bases in each method taking a `basis_type` as argument. Couldn't the lookup dictionary simply take the `enum` values as keys? How would you replace the dictionary by an `enum`?

> * Can the commits and PR incorporate the [prefixes that DIPY recommends](https://github.com/dipy/dipy/blob/master/doc/devel/commit_codes.rst) ?

Gladly, thanks for the link!

While we're at it, I'd like to have inputs on the following matter:
* In [this issue](https://github.com/dipy/dipy/issues/392), it is suggested the default basis should be the `legacy` ones. As it was back in 2014, opinions on this might have changed. I feel like setting `legacy=False` by default will force the user to move to the modern bases, making it easier to deprecate them in the future. Downside is all scripts using data in legacy bases will need to specify `legacy=True` for the data to make sense.

Does having `legacy=False` by default make sense to you or would it be best to make it `True` by default with a warning for the user to consider moving to modern bases?

Thanks again!

jhlegarreta:
> I want to make sure I understand correctly these ones. Your suggestion is simply that we should take advantage of this PR to fix docstrings issues in the changed files?

Given that this is a complete refactoring for a few methods and that you already needed to modify the docstrings in these, yes. But I am only implying it for the modified methods, not for the ones that you did not modify or the entire files.

> I think we would have to decide if we rename legacy to mrtrix3 in methods using tournier basis. The thing is methods such as sh_to_sf and sh_to_sf_matrix can be called with the tournier or descoteaux basis. For this reason, I think we should stick with the same parameter name as it prevents us from having to explain that legacy=True corresponds to mrtrix3=False. It would also feel more consistent through the whole file. What do you think? Maybe @arokem would like to weight in?

I personally dislike the idea using names from third-party software for such purposes for a few reasons, although I grant that in this case the issue was in part motivated between differences `MRtrix0.2` and `MRtrix3`. IMHO if the `sqrt` difference is clearly explained in `DIPY`'s doc, using `tournier` allows to clearly identify the basis, and credit the author. But I will not oppose if Ariel or others still think that it should be changed.

> Is your suggestion we simply replace the basis_type string by an enum? If so, I agree it would be cleaner and easier to maintain as it removes the need to document the available bases in each method taking a basis_type as argument. Couldn't the lookup dictionary simply take the enum values as keys? How would you replace the dictionary by an enum?

Yes to the first question. The `enum` would completely replace the dictionary.

My proposal would be something like:
```
(...)
import enum
(...)

class Basis(enum.Enum):
    """Necessary documentation for the bases.
    """
    DESCOTEAUX07 = 'descoteaux07'
    TOURNIER07 = 'tournier07'
```

But others may oppose to the proposal, or maybe suggest other names if necessary (e.g. removing the `07`, etc.), simply use integers instead of strings for the values, auto-numbering them, etc.
 
> Gladly, thanks for the link!

:+1: You did it for the PR, but not the commit messages, which seem as important :heavy_check_mark: . Let me know if you find trouble doing that, I would be glad to help.

> Does having legacy=False by default make sense to you or would it be best to make it True by default with a warning for the user to consider moving to modern bases?

Switching to the new definitions could be done in a mayor release I think, or after a deprecation cycle of the number of releases that `DIPY` uses. Changing all potential tests or examples affected in `DIPY` may involve some work, but is doable if the CI builds do flag such cases. The concern would more be about end users now getting different results: they may not find this change one that brings them a benefit. They would be forced to change all their scripts and pipelines to get the same results in downstream tasks. So IMHO an automatic warning will, yes, be necessary if the switch is made. As suggested in the mentioned issue, a warning could be added now to let users know that the behavior will change after the deprecation release cycle. And the effective change be made in a specific, separate commit, together with the potential test check changes, be merged to `master` but be excluded from the next few minor releases. This involves maintenance and being very careful when the releases are made.

CHrlS98:
> Given that this is a complete refactoring for a few methods and that you already needed to modify the docstrings in these, yes. But I am only implying it for the modified methods, not for the ones that you did not modify or the entire files.

Got it, will do!

> Yes to the first question. The `enum` would completely replace the dictionary.
> 
> My proposal would be something like:
> 
> ```
> (...)
> import enum
> (...)
> 
> class Basis(enum.Enum):
>     """Necessary documentation for the bases.
>     """
>     DESCOTEAUX07 = 'descoteaux07'
>     TOURNIER07 = 'tournier07'
> ```
> 
> But others may oppose to the proposal, or maybe suggest other names if necessary (e.g. removing the `07`, etc.), simply use integers instead of strings for the values, auto-numbering them, etc.

The dictionary associates the SH basis name with the right method to call (see [here](https://github.com/dipy/dipy/blob/2e0ce1af69a9b2844fbac7c6628f06ef2e31c170/dipy/reconst/shm.py#L478) or pasted below):

```
sph_harm_lookup = {None: real_sh_descoteaux,
                   "tournier07": real_sh_tournier,
                   "descoteaux07": real_sh_descoteaux}
```
Getting rid of the dictionary means having another way of linking the basis name with the function to call. I feel like the dictionary does a pretty good job for this purpose. What are you suggesting?

> :+1: You did it for the PR, but not the commit messages, which seem as important :heavy_check_mark:. Let me know if you find trouble doing that, I would be glad to help.

I found [this reference](https://gist.github.com/nepsilon/156387acf9e1e72d48fa35c4fabef0b4). Is there a preferred way to do this? I also want to make sure it will not mess up the ongoing discussions on the PR.

jhlegarreta:
> Getting rid of the dictionary means having another way of linking the basis name with the function to call. I feel like the dictionary does a pretty good job for this purpose. What are you suggesting?

True, sorry, my bad. We could have a method that gets the `enum` type as the argument and returning the corresponding callable. That way we would not need to do the check at every method, but we would need to call this new method in all instances where the callables are needed. Adds some work but looks nicer to me. Not sure if there would be another way to do it with the `enum`. As said, if there is opposition to this, I will be fine if folks think that it is better to stick to the dictionary.

> I found this reference. Is there a preferred way to do this? I also want to make sure it will not mess up the ongoing discussions on the PR.

That is the way to do it. We will need to force push; there is no choice in there. I will help you off-line.

dipy/reconst/shm.py
CHrlS98: I agree it would be less confusing, but it wouldn't work with the current dictionary called in `sh_to_sf` and `sh_to_sf_matrix` because the meaning of the legacy and mrtrix3 arguments are different. Maybe this could justify replacing the dictionary by a method as suggested by @jhlegarreta [here](https://github.com/dipy/dipy/pull/2206#issuecomment-691522783). I'm also starting to doubt the pertinence of the `legacy` argument itself. Maybe the legacy implementations should be methods of their own, with a unique basis name, that we could remove in the future when they are deprecated. This would involve less maintenance as we won't have to change `real_sh_tournier` and `real_sh_descoteaux` when the legacy bases get deprecated. Opinions on this?

skoudoro:
Hi @CHrlS98,

What is the status of this PR?  Let us know if you have any questions.

doc/theory/sh_basis.rst
CHrlS98: Looking at this [page](https://dipy.org/documentation/1.2.0./theory/sh_basis/) from the documentation for DIPY 1.2, `$\Im$` and `$\Re$` appear to be formatted the right way. Should I add  the double backticks around the `$` symbol?

jhlegarreta: True, my bad ! No, the rendered doc looks right an consistent with the LaTeX symbols used in the `cases`, so no changes are needed.

doc/theory/sh_basis.rst
```diff
@@ -27,26 +35,29 @@ computed using the expression:
     a_l^m = \int_S f(\theta, \phi) Y_l^m(\theta, \phi) ds
 
 Once the coefficients are computed, the function $f(\theta, \phi)$ can be
-approximately computed as:
+computed as:
 
 ..  math::
 
     f(\theta, \phi) = \sum_{l = 0}^{\inf} \sum_{m = -l}^{l} a^m_l Y_l^m(\theta, \phi)
```
jhlegarreta, line 42: Having a look at the rendered page, looks like `\inf` is a typo and should be `\infty`. @CHrlS98 given that the file is being modified in this PR, do you mind changing this, please?

CHrlS98:
Hi @skoudoro,
Hopefully I will add new commits by the end of the week. 

I have some subjects on which I'd like to have some input :
* Is the parameter `legacy` worth it ? Would it be simpler to just give different names to the legacy and modern bases? More detailed comment [here](https://github.com/dipy/dipy/pull/2206#discussion_r487566910) ;
* What would be the `from` and `until` parameters value for `deprecate_with_version`?
* To follow up on @arokem's comments on the use of `tournier07` ([here](https://github.com/dipy/dipy/pull/2206#discussion_r486010749)) and `descoteaux07` ([here](https://github.com/dipy/dipy/pull/2206#discussion_r486011239)) _legacy_ implementations in our current code base, `SphHarmModel` and `QballBaseModel` directly use `real_sym_sh_basis` (`descoteaux07` _legacy_ implementation). I also encountered methods in other modules using `real_sph_harm`, an implementation of `descoteaux07` _legacy_ basis. Therefore, calling those methods would result in a deprecation warning because of the used bases. Refactoring the problematic methods looks like a lot of work. Do we want to do it or do you prefer an other alternative?

Thanks

doc/theory/sh_basis.rst
CHrlS98: No problem :+1:

dipy/reconst/shm.py
CHrlS98: I agree, I'll remove them and update the documentation.

dipy/reconst/shm.py
CHrlS98: I agree, I'll remove them and update the documentation.

dipy/reconst/shm.py
CHrlS98: Hi, the docstrings should be fixed by commit [11beb24](https://github.com/dipy/dipy/pull/2206/commits/11beb2414689604f8c256a4dfb1e703c34a70ff4)!

jhlegarreta: Looks good at first sight. Thanks @CHrlS98 !

dipy/reconst/shm.py
jhlegarreta: Done in 11beb24.

dipy/reconst/shm.py
CHrlS98: Hi @skoudoro, I fixed it in order to use it as a decorator in [11beb24](https://github.com/dipy/dipy/pull/2206/commits/11beb2414689604f8c256a4dfb1e703c34a70ff4)! As for the from and until, what values should I use?

doc/theory/sh_basis.rst
CHrlS98: Hi, the comments regarding sh_basis.rst have been adressed in commits d928b33  and 3f9ea83 !

doc/theory/sh_basis.rst
jhlegarreta: Addressed in 3f3f9ea83.

doc/theory/sh_basis.rst
jhlegarreta: Addressed in d928b33.

dipy/reconst/shm.py
CHrlS98: Done, see 11beb24!

dipy/reconst/shm.py
CHrlS98: Done, see 11beb24!

doc/theory/sh_basis.rst
jhlegarreta: Done in d928b33.

doc/theory/sh_basis.rst
jhlegarreta: Re-worded in in d928b33.

dipy/reconst/shm.py
jhlegarreta: Done in 11beb24.

arokem (commented):
Overall +1 to merge this. 

I had a couple of comments/questions.

arokem:
We do still need to fix all the failing tests: https://travis-ci.com/github/dipy/dipy/jobs/387267835

I think that a lot of these are due to warnings that are no longer being raised, with the integration of the new code (which is a good thing!).

CHrlS98:
Hi @arokem,
 
> We do still need to fix all the failing tests: https://travis-ci.com/github/dipy/dipy/jobs/387267835
> 
> I think that a lot of these are due to warnings that are no longer being raised, with the integration of the new code (which is a good thing!).

Actually a lot of the failing tests are a result of deprecation warning being raised by methods calling now-deprecated SH bases.

Before fixing the tests, I'd like to have opinions on the following subjects (pasted from a [previous comment](https://github.com/dipy/dipy/pull/2206#issuecomment-694476048)):

* Is the parameter `legacy` worth it ? Would it be simpler to just give different names to the legacy and modern bases? More detailed comment [here](https://github.com/dipy/dipy/pull/2206#discussion_r487566910) ;

* What would be the `from` and `until` parameters value for `deprecate_with_version`?

* To follow up on @arokem's comments on the use of `tournier07` ([here](https://github.com/dipy/dipy/pull/2206#discussion_r486010749)) and `descoteaux07` ([here](https://github.com/dipy/dipy/pull/2206#discussion_r486011239)) _legacy_ implementations in our current code base, `SphHarmModel` and `QballBaseModel` directly use `real_sym_sh_basis` (`descoteaux07` _legacy_ implementation). I also encountered methods in other modules using `real_sph_harm`, an implementation of `descoteaux07` _legacy_ basis. Therefore, calling those methods would result in a deprecation warning because of the bases used. Refactoring the problematic methods looks like a lot of work. Do we want to do it or do you prefer an other alternative?

Thanks

dipy/reconst/shm.py
```diff
@@ -156,27 +158,35 @@ def spherical_harmonics(m, n, theta, phi, use_scipy=True):
...
+    n : int ``>= 0``
+        The order of the harmonic.
     theta : float [0, 2*pi]
         The azimuthal (longitudinal) coordinate.
     phi : float [0, pi]
         The polar (colatitudinal) coordinate.
-    use_scipy : bool
-        if True, use scipy implementation. default(True)
+    use_scipy : bool, optional
+        If True, use scipy implementation.
```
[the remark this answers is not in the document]
CHrlS98: I removed default value in method headers for optional arguments as suggested [here](https://github.com/dipy/dipy/pull/2206#pullrequestreview-486003361) by @jhlegarreta. Should I add them again?

doc/theory/sh_basis.rst
```diff
@@ -57,28 +69,95 @@ literature for the computation of the ODF. DIPY implements two of these in the
...
+sign flip for the SH functions of even degree $m < 0$. This has no effect on the
+mathematical properties of each basis.
+
+The ``tournier07`` SH basis defined above is the basis used in MRtrix 0.2 [3]_.
+However, the omission of the $\sqrt{2}$ factor seen in the basis from Descoteaux
+*et al.* [1]_ makes it non-orthonormal. For this reason, the MRtrix3 [4]_ SH
+basis uses a new basis including the normalization factor.
+
+Since DIPY 1.3, the ``descoteaux07`` and ``tournier07`` SH bases have been
+updated in order to agree with the litterature and the latest MRtrix3
```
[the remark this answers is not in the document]
CHrlS98: Fixed by 0e98149

doc/theory/sh_basis.rst
```diff
@@ -57,28 +69,95 @@ literature for the computation of the ODF. DIPY implements two of these in the
...
+These bases differ from the literature by the presence of an absolute value around
+$m$ when $m < 0$. Due to relations $-p = |p| ; \forall p < 0$ and
+$Y_l^{-m}(\theta, \phi) = (-1)^m \overline{Y_l^m}$, the effect of this change is a
+sign flip for the SH functions of even degree $m < 0$. This has no effect on the
+mathematical properties of each basis.
+
+The ``tournier07`` SH basis defined above is the basis used in MRtrix 0.2 [3]_.
+However, the omission of the $\sqrt{2}$ factor seen in the basis from Descoteaux
+*et al.* [1]_ makes it non-orthonormal. For this reason, the MRtrix3 [4]_ SH
+basis uses a new basis including the normalization factor.
```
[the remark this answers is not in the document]
CHrlS98: Added a note in 0e98149 :+1:

doc/theory/sh_basis.rst
```diff
@@ -27,26 +35,30 @@ computed using the expression:
...
+the ODF. Descoteaux *et al.* [1]_ use the Q-Ball Imaging (QBI) formalization
+to recover the ODF, while Tournier *et al.* [2]_ use the Spherical Deconvolution
+(SD) framework.
 
-Several Spherical Harmonics bases have been proposed in the diffusion imaging
-literature for the computation of the ODF. DIPY implements two of these in the
-:mod:`~dipy.reconst.shm` module tool set:
+Several modified SH bases have been proposed in the diffusion imaging literature
+for the computation of the ODF. DIPY implements two of these in the 
+:mod:`~dipy.reconst.shm` module set. Below are the formal definitions taken
```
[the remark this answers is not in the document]
CHrlS98: Solved by 0e98149 !

dipy/reconst/shm.py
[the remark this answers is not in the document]
arokem: No. Jon is right.

arokem:
Looks like everything here is good to go! Thanks for the quick turnaround here @CHrlS98! I think that the one remaining failing CI machine should be resolved in #2271, but I am not 100% sure. @skoudoro -- is that right?

skoudoro:
> I think that the one remaining failing CI machine should be resolved in #2271, but I am not 100% sure. @skoudoro -- is that right?

Yes, I confirm, 💯 % sure

doc/theory/sh_basis.rst
jhlegarreta: Fixed in d928b33. Thanks @CHrlS98 !

dipy/reconst/shm.py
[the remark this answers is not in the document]
jhlegarreta: :+1: Thanks both. @CHrlS98 There are two appearances that have been left behind: 
https://github.com/dipy/dipy/pull/2206/files#diff-cb49c37767759fd0bc11db233e96ac13a346e92cc696f01932f55b52f7c970f9R1534
https://github.com/dipy/dipy/pull/2206/files#diff-cb49c37767759fd0bc11db233e96ac13a346e92cc696f01932f55b52f7c970f9R698

dipy/reconst/shm.py
```diff
@@ -221,173 +235,288 @@ def real_sph_harm(m, n, theta, phi):
...
+    .. [1] Tournier J.D., Calamante F. and Connelly A. Robust determination
+           of the fibre orientation distribution in diffusion MRI:
+           Non-negativity constrained super-resolved spherical deconvolution.
+           NeuroImage. 2007;35(4):1459-1472.
+    .. [2] Tournier J-D, Smith R, Raffelt D, Tabbara R, Dhollander T,
+           Pietsch M, et al. MRtrix3: A fast, flexible and open software
+           framework for medical image processing and visualisation.
+           NeuroImage. 2019 Nov 15;202:116-137.
+    """
+    # In the case where m < 0, Tournier basis consider |m|
```
jhlegarreta, line 283: # In the m < 0 case, Tournier basis considers |m|

jhlegarreta: Done in 641c484. Thanks @CHrlS98 :+1:.

CHrlS98: Fixed in 641c484 :+1:

dipy/reconst/shm.py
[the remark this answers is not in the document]
jhlegarreta: @CHrlS98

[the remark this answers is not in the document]
CHrlS98: Oups I missed this comment, on it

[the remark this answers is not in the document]
CHrlS98: Done in 59d97f5, should be good for all methods in `shm.py`

jhlegarreta (approved):

skoudoro:
Thank you @CHrlS98, nice work!  I am going ahead and merge this PR. 

Can you have a look at #392  and #388, and reopen them if needed. They will get close as soon as I merge your PR. 

Thank you

PR #2210, opened 2020-09-07
Really release the memory used by arclength variable.

arokem:
This is a one-liner. I intermittently run into the following crash when running scripts that use particle filtering tractography: 

```
python(68504,0x11d1fcd40) malloc: Incorrect checksum for freed object 0x7fd1b24b2400: probably modified after being freed.
Corrupt value: 0x3fda8e5f3588a582
python(68504,0x11d1fcd40) malloc: *** set a breakpoint in malloc_error_break to debug
Abort trap: 6

```

I found this SO post: https://stackoverflow.com/a/51546263/3532933, and this change was the only one that made sense given the context.

### DIFF
--- a/dipy/tracking/streamlinespeed.pyx
+++ b/dipy/tracking/streamlinespeed.pyx
@@ -242,7 +242,7 @@ cdef void c_set_number_of_points(Streamline streamline, Streamline out) nogil:
         out[new_N-1, dim] = streamline[N-1, dim]
 
     free(arclengths)
-
+    arclengths = NULL
 
 cdef void c_set_number_of_points_from_arraysequence(Streamline points,
                                                     np.npy_intp[:] offsets,


skoudoro (commented):
LGTM, but I do not know why this change impacts the compilation Linux and macOS (it is failing on my macOS too).

it needs a deeper introspection.

arokem:
Oh weird. I managed to compile on my Mac laptop. This looks to be one of those tricky ones...

richford:
FWIW, I can confirm the issue that this is trying to correct. I get failures during PFT tracking. Locally, they are intermittent failures. On AWS Batch, they seem to be consistent. They generate no traceback but rarely they will include information like
```
munmap_chunk(): invalid pointer
```
or
```
The futex facility returned an unexpected error code.
```

arokem:
@gabknight : this is the issue I mentioned to you just now

skoudoro:
@arokem, can you rebase this PR? 

Thank you

arokem:
OK - rebased! I'm still not sure whether this solves the problem I was having, though.

arokem:
Seems like this is causing compilation errors on some platforms: https://dev.azure.com/dipy/dipy/_build/results?buildId=968&view=logs&jobId=8862b770-af0d-5931-a6ea-30fedf892d13&j=8862b770-af0d-5931-a6ea-30fedf892d13&t=0d3b08de-c1d8-5058-80ae-46393c154e54. 

So, I think that we can close this. It doesn't really seem to solve the issue I was having either.

PR #2212, opened 2020-09-07
[FIX] renamed Direct Streamline Normalization tutorial title

### LINKED ISSUE #2114: Tutorial title in Streamline-based registration section is misleading
If you look at the Streamline-based registration section there are two tutorials.

The first is called ``Direct registration of streamlines`` and therefore it is in the right section.
https://dipy.org/documentation/1.1.1./examples_built/bundle_registration/#example-bundle-registration

The second is called ``Direct Image-based Streamline Normalization`` but it is actually using image-based registration applied to streamlines.
https://dipy.org/documentation/1.1.1./examples_built/streamline_registration/#example-streamline-registration

We need to change the title to ``Image-based deformations applied into streamlines`` or similar and remove direct from the title. This is clearly indirect.

areeshatariq:
fixes #2114

### DIFF
--- a/doc/examples/streamline_registration.py
+++ b/doc/examples/streamline_registration.py
@@ -1,7 +1,7 @@
 """
-===============================
-Direct Streamline Normalization
-===============================
+================================================
+Applying image-based deformations to streamlines
+================================================
 
 This example shows how to register streamlines into a template space by
 applying non-rigid deformations.


doc/examples/streamline_registration.py
```diff
@@ -1,7 +1,7 @@
 """
-===============================
-Direct Streamline Normalization
-===============================
+=================================================
+Image-based deformations applied into streamlines
```
arokem, line 3:
```suggestion
Applying image-based deformations to streamlines
```

arokem:
Thanks for doing this. I offered an alternative wording, which I think works better here.

arokem:
:tada: congratulations @areeshatariq on your first merged contribution to DIPY!

skoudoro:
🥳 🎉 👍

PR #2218, opened 2020-09-16
DOC : inconsistent save_seeds documentation

areeshatariq:
Removed the 'streamlines' variable name from the 'Returns' list since the method returns a generator, and hence not the value of a variable named 'streamlines'.

### DIFF
--- a/dipy/tracking/utils.py
+++ b/dipy/tracking/utils.py
@@ -853,13 +853,14 @@ def transform_tracking_output(tracking_output, affine, save_seeds=False):
     affine : array (4, 4)
         The mapping between voxel indices and the point space for seeds.
         The voxel_to_rasmm matrix, typically from a NIFTI file.
-    seeding_activated : bool, optional
+    save_seeds : bool, optional
         If set, seeds associated to streamlines will be also moved and returned
     Returns
     -------
     streamlines : generator
-        A sequence of transformed streamlines.
-        If return_seeds is True, also return seeds
+        A generator for the sequence of transformed streamlines.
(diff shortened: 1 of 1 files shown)


jhlegarreta (approved):
:+1:

dipy/tracking/utils.py
```diff
@@ -853,13 +853,13 @@ def transform_tracking_output(tracking_output, affine, save_seeds=False):
     affine : array (4, 4)
         The mapping between voxel indices and the point space for seeds.
         The voxel_to_rasmm matrix, typically from a NIFTI file.
-    seeding_activated : bool, optional
+    save_seeds : bool, optional
         If set, seeds associated to streamlines will be also moved and returned
     Returns
     -------
-    streamlines : generator
```
skoudoro, line 860: I recommend to keep this line.

dipy/tracking/utils.py
```diff
@@ -853,13 +853,13 @@ def transform_tracking_output(tracking_output, affine, save_seeds=False):
...
-        A sequence of transformed streamlines.
-        If return_seeds is True, also return seeds
+        A generator for the sequence of transformed streamlines.
+        If save_seeds is True, also return a generator for the
+    transformed seeds.
```
skoudoro, line 862: I think this line should be indented at the same level of  `if save_seeds....`

skoudoro (commented):
Hi @areeshatariq,

Thank you for this correction. See below some of the comments.

skoudoro (approved):
Thanks @areeshatariq !

PR #2220, opened 2020-09-17
[DOC] move Denoising on its own section

areeshatariq:


skoudoro:
Thank you @areeshatariq! LGTM.  This is a small PR so I merge it.

PR #2180, opened 2020-05-26
[NF] add new linear transforms for Rigid+IsoScaling and Rigid+Scaling

bpinsard:
Added new affine transforms with 7 and 9 parameters.
Wrote the 3D one only, but will write the 2D one if there is interest. 

These are mainly copy-paste from existing transforms but it would be good to double-check that my linear algebra is correct. :D

PR #2223, opened 2020-09-22
[NF] new linear transforms for Rigid+IsoScaling and Rigid+Scaling

skoudoro:
This PR supersede and fix #2180. the goal of this PR is:

- Add the 2D versions of this transformation.
- Correct the Jacobian matrix of the previous 3D implementation.
- update the documentation of the previous PR.
- Add Tests for all transforms.
- rebase the previous PR.

Can you have a look @bpinsard?

Thank you.

### DIFF
--- a/dipy/align/tests/test_transforms.py
+++ b/dipy/align/tests/test_transforms.py
@@ -2,7 +2,6 @@ from dipy.align.transforms import regtransforms, Transform
 import numpy as np
 from numpy.testing import (assert_array_equal,
                            assert_array_almost_equal,
-                           assert_almost_equal,
                            assert_equal,
                            assert_raises)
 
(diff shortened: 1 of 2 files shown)


skoudoro:
@Borda @grlee77 @arokem, Can you have a look at PR, please?

Thank you very much

grlee77:
+1 on providing these transforms

A couple of basic suggestions prior to looking too closely at the details of the code:

1.) Perhaps provide a new example based on modification of the existing `affine_registration_3d.py`? One could choose the b0 volume and then introduce a second copy after a small rotation, translation and/or scaling (possibly via scipy.ndimage's `rotate`/`shift`/`zoom`). The proposed transforms should then be able to co-register these. (I suppose this could also be used as an integration test if you wanted to validate the accuracy of the estimated translations/rotations to some precision).

2.) consider adding additional methods for these to `workflows/align.py` (can always be done as a follow-up PR too)

The test cases proposed here do appear to verify that the classes produce the expected transformation matrix for a given set of parameters, but to increase confidence, it would be nice to demonstrate that they operate as expected when used in an actual registration example.

skoudoro:
Thank you for this feedback @grlee77!

Concerning the test cases, I agree with your point and I will add a check/demonstration with the actual registration framework. Thank you!

Concerning point 1, I planned to create 2 follow up PR:
- update the `worflows/align.py`.
- create a basic tutorial for all transformations. How to use them, with or without our registrations framework.

In general, I tried to avoid creating too big PR. It helps to have a smoother review, this is why it is not present.

PR #2224, opened 2020-09-22
Replace pytest.xfail by npt.assert_raises

skoudoro:
Quick PR to replace `pytest.xfail` by `npt.assert_raises` due to some confusion during the Pytest report.


arokem:
Yep! Looks good.

PR #2226, opened 2020-09-23
STYLE: Conform to `reStructuredText` syntax in examples sections

jhlegarreta:
Conform to `reStructuredText` syntax in examples sections.


jhlegarreta:
Only `codecov`is complaining; the potential failure is most probably unrelated as this is a `STYLE` change. I think we can assume it is safe to merge this.

skoudoro:
👍  Thank you for this @jhlegarreta. Indeed, it looks good! we just need to check the doc generation and it will be in today

skoudoro:
Thank you @jhlegarreta.

PR #2227, opened 2020-09-24
RF: Replaces our own custom progressbar with a tqdm progressbar.

### LINKED ISSUE #2219: Replace fetcher progress bar with tqdm
Does anyone mind if we replace the progress bar defined here:

https://github.com/dipy/dipy/blob/b0b89bcdd2fa349b8f7fde50287fac3cf35bc845/dipy/data/fetcher.py#L45

with a [tqdm](https://tqdm.github.io/) progress bar? 

The one we are currently using tends to create very long logs, because each iteration generates its own line. I think that tqdm would solve that problem. This would add another dependency, but I think it's pretty light-weight.

arokem:
Closes #2219.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ addons:
 
 env:
     global:
-        - DEPENDS="cython numpy matplotlib h5py nibabel cvxpy"
+        - DEPENDS="cython numpy matplotlib h5py nibabel cvxpy tqdm"
         - VENV_ARGS="--python=python"
         - INSTALL_TYPE="setup"
         - PRE_WHEELS="https://pypi.anaconda.org/scipy-wheels-nightly/simple"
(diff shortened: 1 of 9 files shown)


arokem:
I'm at a bit of a loss here. Why is Travis not picking up the tqdm dependency from [here](https://github.com/dipy/dipy/pull/2227/files#diff-2d2616242db54458268e23e05ae71666R111) on the Travis `DEPEND=""` build?

skoudoro:
I do not know the impact of [this line](https://github.com/dipy/dipy/blob/master/setup.py#L157) but it might be because of that. I release that we forgot it for `packaging` module too. However, `packaging` is always installed via Setuptools. I suppose this is why I had no error with this module.

skoudoro:
Hi @arokem,

This PR is close to being done, Can you try the proposition above?

arokem:
Thanks for the tip @skoudoro! Let's see if this does the trick.

arokem:
Looks like the failures are now all unrelated to this PR.

skoudoro:
Indeed, Looks good to me so merging! thanks @arokem

PR #2230, opened 2020-09-25
TST: Assert the shape of the output based on the docstring.

arokem:
This one is a question for @karanphil. According to the [docstring](https://github.com/dipy/dipy/blob/master/dipy/reconst/mcsd.py#L746), each of the outputs of this function should have the shape: (len(gtab.bvals)-1, 4). And yet, this test fails. Am I missing something?

### DIFF
--- a/dipy/reconst/tests/test_mcsd.py
+++ b/dipy/reconst/tests/test_mcsd.py
@@ -302,6 +302,8 @@ def test_auto_response_msmt():
         with bvalues lower than 1200, followed by response_from_mask_msmt
         to overcome this.""" in str(w[0].message))
 
+        npt.assert_equal(response_auto_wm.shape, (len(gtab.bvals)-1, 4))
+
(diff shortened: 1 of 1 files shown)


arokem:
In particular, I get: 
```
E           AssertionError: 
E           Items are not equal:
E           item=0
E           
E            ACTUAL: 3
E            DESIRED: 192
```
Which is to say, that I was expecting a (192, 3) array, but I got a (3, 3) array instead.

karanphil:
> This one is a question for @karanphil. According to the [docstring](https://github.com/dipy/dipy/blob/master/dipy/reconst/mcsd.py#L746), each of the outputs of this function should have the shape: (len(gtab.bvals)-1, 4). And yet, this test fails. Am I missing something?

It is because the response function is computed for each `unique_bvals`. So 3!  I agree that it is not so clear.

arokem:
That makes sense! I'll close this, but maybe we should clarify the docstring.

PR #2234, opened 2020-09-26
DOC: Improve some of the links in the `info.py` file

jhlegarreta:
Improve some of the links in the `info.py` file:
- Use https for the links.
- Also, change the `DOWNLOAD_URL` to the releases page, which contains
downloadable compressed files of the released versions: the previous
`DOWNLOAD_URL` string was pointing to a place in `GitHub` where no tarballs
existed.


jhlegarreta:
`PEP8` is right, but I sticked to the existing style:
- Should we add the file to the files that `PEP8` should ignore?
- Should we stick to `PEP8`?
- Should we simply ignore `PEP8 `?

skoudoro:
Yes, we should this file to be ignored by pep8.

Otherwise, Thank you for the update. Merging

jhlegarreta:
> Yes, we should this file to be ignored by pep8.

Note: I realize that since we are using the `pep8speaks` app and not doing this in the CI's, this may not be possible. I guess it would be possible if at some point the `PEP8` checks are transitioned to the CI configuration files.

skoudoro:
> Note: I realize that since we are using the pep8speaks app and not doing this in the CI's, this may not be possible. I guess it would be possible if at some point the PEP8 checks are transitioned to the CI configuration files.

it is possible, we need to update the `.pep8speaks.yml` file like we did [here](https://github.com/dipy/dipy/blob/master/.pep8speaks.yml#L23) for the example

jhlegarreta:
:grimacing: True, sorry for the confusion: #2241.

PR #2239, opened 2020-09-28
[ENH] Add inplace kwarg to gibbs_removal

### LINKED ISSUE #2237: gibbs_removal overwrites input data when inputting 3-d or 4-d data.
## Description
`gibbs_removal` overwrite input data when inputting 3-d or 4-d data that happens on this line. https://github.com/dipy/dipy/blob/c1c43b4ea71086682a5b408b539274495a9052cb/dipy/denoise/gibbs.py#L291 I doubt this is intentional, but is an easy fix by simply making a copy of the input data in the beginning or creating an empty array to write into.

## Way to reproduce
[If reporting a bug, please include the following important information:]
- [x] Code example

> import numpy as np
> from dipy.denoise.gibbs import gibbs_removal
> 
> X = np.random.normal(size=(50, 50, 25))
> Y = X.copy()
> _ = gibbs_removal(X)
> np.array_equal(X, Y)
> False

- [ ] Relevant images (if any)
- [x] Operating system and version (run `python -c "import platform; print(platform.platform())"`)
  - Darwin-19.6.0-x86_64-i386-64bit
- [x] Python version (run `python -c "import sys; print("Python", sys.version)"`)
  - Python 3.7.9
- [x] dipy version (run `python -c "import dipy; print(dipy.__version__)"`)
  - Dipy 1.2.0
- [x] dependency version (numpy, scipy, nibabel, h5py, cvxpy, fury)
  - NumPy 1.19.1
  - SciPy 1.5.2
  - Nibabel 3.1.1
  - H5py 2.10.0

j1c:
Added `inplace` kwarg to `gibbs_removal`. Set to `True` by default. The inplace behavior is consistent regardless of the dimensionality of input data. 

Note that I had to modify the original tests by adding `inplace=False` in some cases to keep the original behavior, which assumed 2d inputs always returned a copy.

Closes #2237

### DIFF
--- a/dipy/denoise/gibbs.py
+++ b/dipy/denoise/gibbs.py
@@ -220,7 +220,7 @@ def _gibbs_removal_2d(image, n_points=3, G0=None, G1=None):
     return imagec
 
 
-def gibbs_removal(vol, slice_axis=2, n_points=3):
+def gibbs_removal(vol, slice_axis=2, n_points=3, inplace=True):
     """Suppresses Gibbs ringing artefacts of images volumes.
 
     Parameters
@@ -233,6 +233,10 @@ def gibbs_removal(vol, slice_axis=2, n_points=3):
     n_points : int, optional
         Number of neighbour points to access local TV (see note).
         Default is set to 3.
+    inplace : bool, optional
+        If True, the input data is replaced with results. Otherwise, returns
+        a new array.
+        Default is set to True.
 
     Returns
     -------
@@ -257,6 +261,9 @@ def gibbs_removal(vol, slice_axis=2, n_points=3):
 
     """
     nd = vol.ndim
+
+    if not isinstance(inplace, bool):
+        raise TypeError("inplace must be a boolean.")
 
     # check the axis corresponding to different slices
     # 1) This axis cannot be larger than 2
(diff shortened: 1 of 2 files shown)


skoudoro:
Thank you for this @j1c!

After a first quick look, the code looks good, and thanks for adding tests! We will look at the code deeper later in the day and come back to you.

j1c:
Actually, upon second thought, I think its better for `inplace` to default to False. I will let you guys make that decision.

skoudoro:
> Actually, upon second thought, I think its better for inplace to default to False. I will let you guys make that decision.

I do not have any strong opinion about `inplace` to `False`. For backward compatibilities and to not surprise the users, I would keep it to `True`.  

What do you think @arokem, @garyfallidis, others ?

skoudoro:
This PR is ready to be merged, I will wait for another opinion to your question @j1c until the end of the week. If no answer, I will merge it as it is.

arokem:
+1 for inplace=True. You could make pretty good arguments for  either of these  and only one is fully back-compatible

skoudoro:
Ok so here we go! Thanks again @j1c!

PR #2242, opened 2020-09-28
maintenance of bundle_shape_similarity function

BramshQamar:
This PR is for maintenance of bundle_shape_similarity function in segment/bundles.py

It updates line 196 in segment/bundles.py
from: 
```
ba_value = ba_analysis(bundle1_centroids, bundle2_centroids,
                                      threshold)
```
to:
```
ba_value = ba_analysis(recognized_bundle=bundle1_centroids,
                                      expert_bundle=bundle2_centroids,
                                      threshold=threshold)
```


arokem:
Just making sure: doesn't look like this changes the logic of the call. It just makes it explicit which input is which. Is that correct? I think it's alright. The CI failure seems unrelated.

skoudoro:
Azure Pipeline is under maintenance so down. That's why it is failing

arokem:
OK - I think it's OK to merge this.

PR #2247, opened 2020-10-01
BF: In  LiFE, set nan signals to 0.

arokem:
When the mean-removed relative signal is a nan (this can arise when the b0 is 0), the sensible thing to do is to ignore it, which means setting it to 0.


skoudoro (approved):
Thanks @arokem,

Make sense to me so +1.

I will wait for any additional comments  until tomorrow and then merge it

skoudoro:
Thank you @arokem

PR #2250, opened 2020-10-09
[ENH] Add parallelization to gibbs denoising

### LINKED ISSUE #2236: Parallelize gibbs_removal
## Description
`gibbs_removal` in essence does fft on each 2d slice of either a 3d or 4d volume. This particular this line, https://github.com/dipy/dipy/blob/c1c43b4ea71086682a5b408b539274495a9052cb/dipy/denoise/gibbs.py#L290 just loops through each slice sequentially. It should be fairly trivial to parallelize the process via `multiprocessing`.

Of course this can be done at the user level by simply running the function with either 2d or 3d inputs, which is what I am doing, but seems like a good addition to dipy itself.

Thoughts?

j1c:
Added parallelization to gibbs denoising. I used the utility class `MapWrapper` from scipy for parallelization, which really just a wrapper for `Pool` in `multiprocessing` module with some convenient functionality for dealing with number of workers, etc. Let me know if this is acceptable.

Main thing I changed is how the axis of the input data are ordered within the `gibbs_denoise` function since its much easier to work with the array as an iterable if the axis we are iterating is the first one, not the last one. Some lines had to be changed to accommodate for this change in axis ordering.

Closes #2236

### DIFF
--- a/dipy/denoise/gibbs.py
+++ b/dipy/denoise/gibbs.py
@@ -241,9 +241,9 @@ def gibbs_removal(vol, slice_axis=2, n_points=3, inplace=True, workers=1):
         a new array.
         Default is set to True.
     workers : int, optional
-        Computation is subdivided into ``workers`` sections and evaluated in 
-        parallel (uses ``multiprocessing.Pool <multiprocessing>``). Supply 
-        ``-1`` to use all cores available to the Process. 
+        Computation is subdivided into ``workers`` sections and evaluated in
+        parallel (uses ``multiprocessing.Pool <multiprocessing>``). Supply
+        ``-1`` to use all cores available to the Process.
         Default is set to 1.
 
     Returns
@@ -311,7 +311,9 @@ def gibbs_removal(vol, slice_axis=2, n_points=3, inplace=True, workers=1):
         vol[:, :] = _gibbs_removal_2d(vol, n_points=n_points, G0=G0, G1=G1)
     else:
         mapwrapper = MapWrapper(workers)
-        partial_func = partial(_gibbs_removal_2d, n_points=n_points, G0=G0, G1=G1)
+        partial_func = partial(
+            _gibbs_removal_2d, n_points=n_points, G0=G0, G1=G1
+        )
         vol[:, :, :] = list(mapwrapper(partial_func, vol))
 
     # Reshape data to original format
--- a/dipy/denoise/gibbs.py
+++ b/dipy/denoise/gibbs.py
@@ -241,11 +241,9 @@ def gibbs_removal(vol, slice_axis=2, n_points=3, inplace=True, num_threads=1):
         a new array.
         Default is set to True.
     num_threads : int or None, optional
-        Split the calculation to a pool of children processes. This only
-        applies to 3D or 4D `data` arrays. If a positive integer then it
-        defines the size of the multiprocessing pool that will be used. If
-        None, then the size of the pool will equal the number of cores 
-        available.
+        Number of threads. Only applies to 3D or 4D `data` arrays. If None then
+        all available threads will be used. Otherwise, must be a positive
+        integer.
         Default is set to 1.
 
     Returns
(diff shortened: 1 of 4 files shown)


j1c:
Ok I see that `MapWrapper` wont work with the scipy version in travis so I'll work on changing to using `Pool`

skoudoro (commented):
Nice! Thank you again @j1c for doing that!

Do you have a performance comparison before and after this update? just to get an idea.
Overall, it looks good to me, so +1 for merging this. But I just need to try your PR.

dipy/denoise/gibbs.py
```diff
@@ -262,33 +271,40 @@ def gibbs_removal(vol, slice_axis=2, n_points=3, inplace=True):
...
+        raise ValueError("Data is not an image")
+
     if not isinstance(inplace, bool):
         raise TypeError("inplace must be a boolean.")
 
+    if not isinstance(num_processes, int):
+        raise TypeError("num_processes must be an int.")
+    else:
+        if num_processes < 0:
+            raise ValueError("num_processes must be >= 0.")
```
skoudoro, line 287: Can you add new tests for all this block?

skoudoro:
Also, Can you add this option to our command-line interface: https://github.com/dipy/dipy/blob/master/dipy/workflows/denoise.py#L244. Thanks!

dipy/denoise/gibbs.py
```diff
@@ -220,7 +223,7 @@ def _gibbs_removal_2d(image, n_points=3, G0=None, G1=None):
     return imagec
 
 
-def gibbs_removal(vol, slice_axis=2, n_points=3, inplace=True):
+def gibbs_removal(vol, slice_axis=2, n_points=3, inplace=True, num_processes=1):
```
grlee77, line 226: Please use `num_threads` instead for consistency with other multi-threaded dipy functions such as (e.g. `dipy.denoise.nlmeans_3d`, functions in `dipy.align.streamlinear, etc.)

grlee77: The default of 1 here looks good to me as this is compatible with users potentially applying other means of parallelization to call dipy functions (e.g. joblib, etc).  I did see that some functions in dipy do default to `num_threads=None`, though. Also, it looks like other functions use `num_threads=None` rather than `num_threads=0` to indicate the maximum should be used so it is probably best to do the same here.

grlee77:
> Do you have a performance comparison before and after this update? just to get an idea.

I ran it locally on a 10-core Intel Skylake-X system and it seems the acceleration provided is equal to the number of physical cores! With `num_threads=0`, 20 threads get used due to hyperthreading and the time to run the following line in `denoise_gibbs.py` was reduced from 107 seconds to 10.7 seconds.

https://github.com/dipy/dipy/blob/1dc577be4611bacf7b7f6de6f3fbd0703d4e2a8b/doc/examples/denoise_gibbs.py#L171

skoudoro:
That's great news!  Thank you @grlee77 for the benchmark test. So After the update, it should be ready to go.

grlee77:
Can you also update `doc/examples/denoise_gibbs.py` as follows:

The following line needs `inplace=False` or the plot shows no difference (this problem in the demo was introduced in #2239, but we may as well fix it here)
https://github.com/dipy/dipy/blob/ace9c7a349c763d914e3f25de552efad891c84d9/doc/examples/denoise_gibbs.py#L82

Add `num_threads=None` to the 3D call at the following line so the demo completes more quickly:
https://github.com/dipy/dipy/blob/ace9c7a349c763d914e3f25de552efad891c84d9/doc/examples/denoise_gibbs.py#L171

dipy/denoise/gibbs.py
j1c: All tests added.

dipy/denoise/gibbs.py
j1c: Changed to `num_threads` with `None` being use all cores.

skoudoro (approved):
All comments have been addressed.

I will wait until tomorrow for any additional comments otherwise it is ready to go. Thanks @j1c!

skoudoro:
No additional comment so here we go! Thanks @j1c! merging

PR #2252, opened 2020-10-09
BUG: Set tau factor to parameter value in local PCA

### LINKED ISSUE #2251: localpca tau_factor is hard coded to 2.3
## Description
`localpca` will always use 2.3 as `tau_factor` since the line below doesn't reference the input argument. According to the docs, seems like the user should be able to set this.

https://github.com/dipy/dipy/blob/8c4b76d7ece74cb06d24dfaa1ce4f2b4fb274be9/dipy/denoise/localpca.py#L318

jhlegarreta:
Set tau factor to parameter value in local PCA denoising.

Fixes #2251.


skoudoro (approved):
Thanks @jhlegarreta. LGTM. Waiting for the CI's

PR #2253, opened 2020-10-09
Improve FFT efficiency in gibbs_removal

grlee77:
This PR makes two changes that give an additional 35% reduction in runtime for me when applied on top of the existing parallel Gibbs PR (#2250).

1.) The first commit uses uncentered frequency coordinates in `k` so that no calls to `fftshift` are required in the inner loops.
2.) The second swaps out `numpy.fft` for `scipy.fft` (or `scipy.fftpack` on older SciPy). The primary benefit of this is that SciPy's FFTs support single precision which ends up being faster when the input is single precision.

### DIFF
--- a/dipy/denoise/gibbs.py
+++ b/dipy/denoise/gibbs.py
@@ -1,5 +1,15 @@
 
 import numpy as np
+from numpy.lib import NumpyVersion as Version
+import scipy
+
+
+if Version(scipy.__version__) >= Version('1.4.0'):
+    import scipy.fft
+    _fft = scipy.fft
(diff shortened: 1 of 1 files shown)


skoudoro (approved):
Hi @grlee77,

I just checked it and the performance improvement is impressive! 🚀 🚀 🚀 

Thank you for this nice trick. Always good to know. 

It is ready to go to me. As usual, I will wait until tomorrow to see if there are any additional comments before merging

arokem:
Good stuff! +1 and merging!

PR #2257, opened 2020-10-16
[Fix] warning if not the same number of points

### LINKED ISSUE #342: Missing a warning if not the same number of points
In dipy.tracking.distances, in bundles_distances_mdf and possibly other metric too, it would be nice to add a verification of the streamlines; if A and B don't have the same number of points, add a downsample or at least a warning. Thank you!

skoudoro:
This PR fixes #342 (a very old "issue"). This is just adding a warning to the user when `bundles_distances_mdf` or `bundles_distances_mam` do not have 2 tracks with the same number of points.

Also, pep8 and docstring has been updated too.

### DIFF
--- a/dipy/core/interpolation.pyx
+++ b/dipy/core/interpolation.pyx
@@ -64,7 +64,7 @@ def interp_rbf(data, sphere_origin, sphere_target,
     if norm == "angle":
         norm = angle
     elif norm == "euclidean_norm":
-        w_s = "The Euclidian norm used for interpolation is inaccurate "
+        w_s = "The Euclidean norm used for interpolation is inaccurate "
         w_s += "and will be deprecated in future versions. Please consider "
         w_s += "using the 'angle' norm instead"
         warnings.warn(w_s, PendingDeprecationWarning)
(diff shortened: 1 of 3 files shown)


dipy/core/interpolation.pyx
```diff
@@ -64,7 +64,7 @@ def interp_rbf(data, sphere_origin, sphere_target,
     if norm == "angle":
         norm = angle
     elif norm == "euclidean_norm":
-        w_s = "The Eucldian norm used for interpolation is inaccurate "
+        w_s = "The Euclidian norm used for interpolation is inaccurate "
```
jhlegarreta, line 67: The correct spelling is "Euclidean" if I'm not wrong.

jhlegarreta (commented):
Thanks for doing this :+1:.

An in-line comment.

dipy/core/interpolation.pyx
skoudoro: Thanks!

jhlegarreta (approved):

arokem:
CI failures unrelated I believe. Merging!

PR #2262, opened 2020-10-18
[FIX] update tests to respect numpy NEP 34 

### LINKED ISSUE #2132: Generating ndarrays with different shapes generates NumPy warning at testing
## Description
When running the [test_clustering](https://github.com/dipy/dipy/blob/master/dipy/segment/tests/test_clustering.py) suite, a number of warnings are generated due to NumPy testing methods needing to create arrays inferring their data type:
```
(which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes)
is deprecated.
If you meant to do this, you must specify 'dtype=object' when creating the ndarray
```
triggered at
https://github.com/numpy/numpy/blob/master/numpy/testing/_private/utils.py#L703
and explained [here](https://numpy.org/neps/nep-0034-infer-dtype-is-object.html).

## Way to reproduce
Test the `test_clustering` suite. No input data required.

The methods where this warning is triggered are
```
test_cluster_map_getitem()
test_cluster_map_comparison_with_object()
```

See for example
https://travis-ci.org/github/dipy/dipy/jobs/676101427#L5543

Not sure about the specific dependency versions or python versions that trigger the warning. `NEP 34` was filed/created on 2019-10-10. Reproducing it locally is still elusive to me.

Looks like for the `test_cluster_map_getitem()` method the warning is triggered only once. There are multiple calls to `assert_equal` and [`assert_arrays_equal`](https://github.com/dipy/dipy/blob/master/dipy/testing/__init__.py#L41), which in turn calls to `assert_array_equal`. Maybe the warning is triggered for the cluster that is empty (i.e. the first one) (?), since otherwise the arrays that the NumPy testing statement needs to create have the same length.

Could be loosely related to #2130.

Suggestions are welcome.

skoudoro:
This PR fixes #2132.  [NEP 34](https://numpy.org/neps/nep-0034-infer-dtype-is-object.html) disallow inferring `dtype=object` from sequences. So the tests have been updated to follow this rule.

### DIFF
--- a/dipy/segment/tests/test_clustering.py
+++ b/dipy/segment/tests/test_clustering.py
@@ -412,7 +412,8 @@ def test_cluster_map_getitem():
 
     # Test slicing and negative indexing
     assert_equal(cluster_map[-1], clusters[-1])
-    assert_array_equal(cluster_map[::2], clusters[::2])
+    assert_array_equal(np.array(cluster_map[::2], dtype=object),
+                       np.array(clusters[::2], dtype=object))
     assert_arrays_equal(cluster_map[::-1], clusters[::-1])
     assert_arrays_equal(cluster_map[:-1], clusters[:-1])
     assert_arrays_equal(cluster_map[1:], clusters[1:])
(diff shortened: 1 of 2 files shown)


jhlegarreta:
The warning seems not to be present for the referred tests in the travis CI builds :+1:.

Two identical warnings persist in other methods: `tracking/tests/test_streamline.py::test_deform_streamlines` and `tracking/tests/test_streamline.py::test_streamlines_generator`, e.g. https://travis-ci.com/github/dipy/dipy/jobs/401441828#L3471

Shall these be addressed here or in a separate PR?

skoudoro:
> Shall these be addressed here or in a separate PR?

I just addressed them on this PR. Thank you for pointing them out!

jhlegarreta:
The related warnings have disappeared :+1:. Merging.

On an unrelated note, it looks like PR #2257 has introduced some warnings on some tests. Maybe they should be detected?

PR #2264, opened 2020-10-19
BF: Fix `dipy_align_syn` default value assumptions

jhlegarreta:
Fix `dipy_align_syn` default value assumptions: ensure that the `mopt_smooth`,
`mopt_inner_iter` and `mopt_step_type`/`step_type` parameters get appropriate
values for all possible combinations of the optional input arguments.

Fixes:
```
INFO:Starting Diffeomorphic Registration
INFO:Using CC Metric
  File "./bin/dipy_align_syn", line 7, in <module>
    run_flow(SynRegistrationFlow())
  File "/path/to/dipy/dipy/workflows/flow_runner.py", line 89, in run_flow
    return flow.run(**args)
  File "/path/to/dipy/dipy/workflows/align.py", line 869, in run
    mopt_smooth = mopt_smooth or init_param[metric]['mopt_smooth']
KeyError: 'cc'
```

The workflow was failing with the above error when being called with default
parameters, e.g.
```
dipy_align_syn /path/to/data/static.nii.gz /path/to/data/moving.nii.gz
```

This happened because the `cc` metric does not require the same arguments as the
`em` and `ssd` metric, and hence the `init_param` variable contained no
information for the `cc` metric.

The fix ensures that if using the `cc` metric or the `mopt_smooth` and
`mopt_inner_iter` have their default values (which evaluate to `False`), the
sensible default values from `init_param` are used to initialize the `em` and
`ssd` metrics.

Similarly, the `mopt_step_type` now gets a default initialization depending on
the metric type: if using the `cc` metric, its value is set to an empty string
if provided, since `cc` does not use it, and `em` and `ssd`'s `step_type`
parameter is initialized to their corresponding values in `init_param`. If
specifying `mopt_step_type` and the `em` or `ssd` metric, its value is only
applied to the specified metric, the other one taking its default initialization
value from `init_param`.

Rather than initializing the value of the step type to a common or shared
default value, the proposed solution ensures that if at some point either
the `em` or the `ssd` metric support an exclusive step type, their preferred
default is exclusive or if all their step types were exclusive, the workflow
will initialize each to the appropriate value.

Take advantage of the commit to make the `mopt_inner_iter` parameter have a
default integer value.

### DIFF
--- a/dipy/workflows/align.py
+++ b/dipy/workflows/align.py
@@ -733,7 +733,7 @@ class SynRegistrationFlow(Workflow):
     def run(self, static_image_files, moving_image_files, prealign_file='',
             inv_static=False, level_iters=[10, 10, 5], metric="cc",
             mopt_sigma_diff=2.0, mopt_radius=4, mopt_smooth=0.0,
-            mopt_inner_iter=0.0, mopt_q_levels=256, mopt_double_gradient=True,
+            mopt_inner_iter=0, mopt_q_levels=256, mopt_double_gradient=True,
             mopt_step_type='', step_length=0.25,
             ss_sigma_factor=0.2, opt_tol=1e-5, inv_iter=20,
             inv_tol=1e-3, out_dir='', out_warped='warped_moved.nii.gz',
@@ -864,11 +864,20 @@ class SynRegistrationFlow(Workflow):
                              'mopt_step_type': 'gauss_newton'
                              }
                       }
-        mopt_smooth = mopt_smooth or init_param[metric]['mopt_smooth']
-        mopt_inner_iter = mopt_inner_iter or  \
-            init_param[metric]['mopt_inner_iter']
-        mopt_step_type = mopt_step_type or \
-            init_param[metric]['mopt_step_type']
+
+        mopt_smooth = mopt_smooth \
+            if mopt_smooth or metric == 'cc' \
+            else init_param[metric]['mopt_smooth']
+        mopt_inner_iter = mopt_inner_iter \
+            if mopt_inner_iter or metric == 'cc' \
+            else init_param[metric]['mopt_inner_iter']
+
+        # If using the 'cc' metric, force the `mopt_step_type` parameter to an
+        # empty value since the 'cc' metric does not use it; for the rest of
+        # the metrics, the `step_type` parameter will be initialized to their
+        # corresponding default values in `init_param`.
+        if metric == 'cc':
+            mopt_step_type = ''
 
         for (static_file, moving_file, owarped_file, oinv_static_file,
              omap_file) in io_it:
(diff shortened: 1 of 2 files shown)


jhlegarreta:
Maybe testing the `cc` and `em` cases should be [added](https://github.com/dipy/dipy/blob/master/dipy/workflows/tests/test_align.py#L283)?

jhlegarreta:
> Maybe testing the cc and em cases should be added?

Tests added in 3776a6b. The error can be reproduced if that commit is applied without the 91f15a0 commit. The test is passing locally.

jhlegarreta:
The reported CI build failure is due to a timeout; the rest are passing, so this is ready to be merged.

dipy/workflows/align.py
```diff
@@ -886,21 +895,31 @@ def run(self, static_image_files, moving_image_files, prealign_file='',
...
+            # Note that `step_type` is initialized to the default value in
+            # `init_param` for the metric that was not specified as a
+            # parameter or if the `mopt_step_type` is empty.
+            l_metric = {
+                "ssd": SSDMetric(
+                    static_image.ndim, smooth=mopt_smooth,
+                    inner_iter=mopt_inner_iter,
+                    step_type=mopt_step_type
+                    if(mopt_step_type and mopt_step_type.strip())
+                    and metric == 'ssd'
```
grlee77, line 907:
```suggestion
                    if mopt_step_type.strip() and metric == 'ssd'
```

dipy/workflows/align.py
```diff
@@ -886,21 +895,31 @@ def run(self, static_image_files, moving_image_files, prealign_file='',
...
+                ),
+                "cc": CCMetric(
+                    static_image.ndim, sigma_diff=mopt_sigma_diff,
+                    radius=mopt_radius),
+                "em": EMMetric(
+                    static_image.ndim, smooth=mopt_smooth,
+                    inner_iter=mopt_inner_iter,
+                    step_type=mopt_step_type
+                    if (mopt_step_type and mopt_step_type.strip())
+                    and metric == 'em'
```
grlee77, line 918:
```suggestion
                    if mopt_step_type.strip() and metric == 'em'
```

grlee77 (changes requested):
Looks good to me. I just have the following style suggestion:

I think a better option than my suggested simplifications is probably to just not create the `l_metric` dictionary at all. Why not just directly set `current_metric` using logic like:

```Python
metric = metric.lower()
if metric == 'ssd':
    current_metric = SSDMetric(...
elif metric == 'cc':
    current_metric = CCMetric(...
elif metric == 'em':
    current_metric = EMMetric(...
else:
    raise ValueError("unrecognized metric: {}".format(metric))
```
otherwise you create all of the metrics in `l_metric` but then only use one of them.

jhlegarreta:
> otherwise you create all of the metrics in l_metric but then only use one of them.

Thanks for the review @grlee77 :+1:. I agree, but I did not want to change the way things were done at that level. Also, with your proposal, we would not need to generate any `init_param` dictionary either, and we'd rather set the parameters directly to each metric instance.

I do not have a preference, but thought the way things had been done was elegant, even if a little bit more convoluted than a straightforward `if/elif/else` solution.

grlee77 (approved):

grlee77:
@jhlegarreta: I changed my review to "Approve" as the comments above are a matter of style and shouldn't hold up potentially merging this for 1.3.0.

jhlegarreta:
Thanks @grlee77 . Not sure if [the suggested changes in these series of comments](https://github.com/dipy/dipy/pull/2264#discussion_r509538578) are strictly equivalent to what I proposed (I have not tested; e.g. if passing a `None` to the `metric` would make your solution to fail, if passing an empty string would to as well, etc.). If they are, I would be willing to address them.

skoudoro (approved):
Looks good to me. Thanks @jhlegarreta and thanks @grlee77  for the review.

The use of `dict` was done by laziness. Sorry about that. If/else complicate the test case and coverage so the use of `dict` is a workaround. Unfortunately, it fails sometimes...

skoudoro:
The PR has been approved twice so merging. Thanks @jhlegarreta!

PR #2265, opened 2020-10-20
Gibbs denoising: fft only along the axis of interest

grlee77:
Here is one more efficiency improvement I came across for the Gibbs artifact removal last week.

The FFT only needs to be applied along the axis where the subvoxel shifts are being applied. There is no need to FFT along the orthogonal direction as well. Because of an earlier transpose at https://github.com/dipy/dipy/blob/9c5769912e50e6c7639ada4d928fd0add33cc604/dipy/denoise/gibbs.py#L93, the axis to FFT is always axis 1.

This gives an additional ~25% performance improvement on the example for me.

cc @j1c, @skoudoro and @arokem who were involved with other recent Gibbs PRs.

### DIFF
--- a/dipy/denoise/gibbs.py
+++ b/dipy/denoise/gibbs.py
@@ -103,19 +103,19 @@ def _gibbs_removal_1d(x, axis=0, n_points=3):
     sp = np.zeros(xs.shape, dtype=dtype_float)
     sn = np.zeros(xs.shape, dtype=dtype_float)
     N = xs.shape[1]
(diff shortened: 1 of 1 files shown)


skoudoro (approved):
I confirm obtaining the same result as before.

Concerning performance, I got a ~15% performance improvement on the example for me (macOS) so 10% less than @grlee77 but still really valuable.

+1 for merging. I will wait until tomorrow for additional comments.

Thank you @grlee77!

PR #2266, opened 2020-10-20
STYLE: minor refactoring in TissueClassifierHMRF

grlee77:
This PR is mainly a style-based refactor of some things in the tissue segmentation code that I had modified some time ago when working on a GPU-based implementation. It will be best to review the various commits in isolation as they tend to address different aspects.

Overall, I don't consider this PR very high priority as there is no new feature or substantial performance impact. Commit 
4a1678d does result in a net decrease in code to maintain, however, as there were two code paths with nearly identical code in each.

The one behavior change here is that in bca8123 I switched to checking the tolerance at every iteration rather than every 10 iterations, so it is possible the loop will might terminate earlier than before if the tolerance is reached at a number of iterations that s not a multiple of 10. The added time for the tolerance check is negligible, so I don't see a strong reason to only check it every 10th iteration.

### DIFF
--- a/dipy/segment/mrf.pyx
+++ b/dipy/segment/mrf.pyx
@@ -80,9 +80,9 @@ class ConstantObservationModel(object):
 
         Returns
         -------
-        mu, std : ndarrays
+        mu, var : ndarrays
             1 x nclasses dimension
-            Mean and standard deviation for each class
+            Mean and variance for each class
 
         """
         cdef:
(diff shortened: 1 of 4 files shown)


skoudoro:
Thank you for this update @grlee77. After a quick look, the refactoring makes complete sense to me.

However, it seems that behavior change has an impact on the tests. Can you check them?

grlee77:
Thanks @skoudoro, I should have checked that prior to uploading (I had only checked that the demo segmentation still looked good). The issue was just needing to account for the change from std to variance in 47274ea in the test cases.

skoudoro:
ok, thank you for the update @grlee77. Can you add a small note on [API_Change.rst](https://github.com/dipy/dipy/blob/master/doc/api_changes.rst) for this behavior change?

After that,  +1 👍  for merging

PR #2271, opened 2020-10-21
Increase Azure pipeline timeout

skoudoro:
This is strange, the default timeout for Azure pipelines should be 6h. However, it is 1h for us.  It seems that we hit this limit with our windows Matrix. So, this PR increases the limit to 2h.  

Let's see if it behaves correctly...


skoudoro:
All green, perfect! 

The shortest matrix: 32 min
The longuest matrix:  1h10 min

timeout increase: OK, This is ready to go!

PR #2280, opened 2020-11-04
DOC: Improve the denoising workflow documentation

jhlegarreta:
Improve the denoising workflow documentation:
- Move general comments about PCA to the place where it is introduced for the
  first time.
- Improve the explanations about the command arguments.
- Fix grammar.
- Use a given line length consistently.

### DIFF
--- a/doc/interfaces/denoise_flow.rst
+++ b/doc/interfaces/denoise_flow.rst
@@ -14,39 +14,48 @@ You can check how to :ref:`fetch the DIPY data<data_fetch>`.
 Denoising using Local PCA
 -------------------------
 
-PCA-based denoising algorithms are effective denoising methods because they explore 
-the redundancy of the multi-dimensional information of diffusion-weighted datasets.
+Denoising algorithms based on principal components analysis (PCA) are effective
+denoising methods because they explore the redundancy of the multi-dimensional
+information of diffusion-weighted datasets. The basic idea behind the PCA-based
+denoising algorithms is to perform a low-rank approximation by thresholding the
+eigenspectrum of the noisy signal matrix.
 
-The algorithm to perform the PCA-based denoising involves the following steps:
+The algorithm to perform a local PCA-based (LPCA) denoising involves the following
+steps:
 
-* First, we estimate the local noise variance at each voxel.
+* Estimating the local noise variance at each voxel.
+* Applying a PCA in local patches around each voxel over the gradient
+directions.
+* Thresholding the eigenvalues based on the local estimate of the noise
+variance, and then doing a PCA reconstruction.
 
-* Then, we apply PCA in local patches around each voxel over the gradient directions.
+The local PCA algorithm turns out to work well on diffusion MRI owing to the 4D
+structure of DWI acquisitions where the q-space is typically oversampled giving
+highly correlated 3D volumes of the same subject.
 
-* Finally, we threshold the eigenvalues based on the local estimate of sigma and then do a PCA reconstruction
+For illustrative purposes of the local PCA denoising method, we will be using
+the ``stanford_hardi`` dataset, but you can also use your own data.
 
-We are using ``stanford_hardi`` dataset. You can also use your own data.
+The workflow for the LPCA denoising requires the paths to the diffusion input
+file, as well as the b-values and b-vectors files.
 
-The workflow for the LPCA denoising requires the paths to the diffusion input file,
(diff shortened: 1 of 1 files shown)


jhlegarreta:
Making sure this is consistent with the review that @arokem and @ShreyasFadnavis did on the blog post document.

ShreyasFadnavis:
Hi @jhlegarreta ! Looks good -- 1 minor change in general: we should probably stick to `Overcomplete Local PCA` and `Marchenko-Pastur PCA`  (both of which are variants of the local PCA method first (AFAIK) introduced for DWIs here: https://onlinelibrary.wiley.com/doi/full/10.1002/mrm.22748)

ShreyasFadnavis:
> Hi @jhlegarreta ! Looks good -- 1 minor change in general: we should probably stick to `Overcomplete Local PCA` and `Marchenko-Pastur PCA` (both of which are variants of the local PCA method first (AFAIK) introduced for DWIs here: https://onlinelibrary.wiley.com/doi/full/10.1002/mrm.22748)

This can be addressed on a later PR of course!

jhlegarreta:
> Hi @jhlegarreta ! Looks good -- 1 minor change in general: we should probably stick to `Overcomplete Local PCA` and `Marchenko-Pastur PCA` (both of which are variants of the local PCA method first (AFAIK) introduced for DWIs here: https://onlinelibrary.wiley.com/doi/full/10.1002/mrm.22748)

Thanks for the review @ShreyasFadnavis. Done in commint 3407c52.

jhlegarreta:
Build failures are unrelated.

areeshatariq (approved):

doc/interfaces/denoise_flow.rst
```diff
@@ -10,43 +10,53 @@ Multiple denoising methods are available in DIPY.
...
-The algorithm to perform the PCA-based denoising involves the following steps:
+The algorithm to perform an Overcomplete Local PCA-based (LPCA) denoising
+involves the following steps:
 
-* First, we estimate the local noise variance at each voxel.
+* Estimating the local noise variance at each voxel.
+* Applying a PCA in local patches around each voxel over the gradient
+directions.
+* Thresholding the eigenvalues based on the local estimate of the noise
+variance, and then doing a PCA reconstruction.
```
skoudoro, line 30: As you can see below, there is an issue with the rendering. Can you fix it

Capture d’écran 2020-11-10 à 23 24 25

skoudoro (commented):
Hi @jhlegarreta,

Thank you for this. When I generate this tutorial, I have the following error:

> :scale:option is ignored

Could you fix it too?

Thanks

skoudoro:
To generate the doc and see the result:   `make -C doc html-after-examples`

doc/interfaces/denoise_flow.rst
jhlegarreta: Good catch. Sorry. Fixed in ca07e4a.

jhlegarreta:
> Thank you for this. When I generate this tutorial, I have the following error:
> :scale:option is ignored
> Could you fix it too?
> Thanks

I assume that this warning/error was already present before this PR. It looks like Sphinx needs `Pillow` for the `scale` option to work. The [`doc-requirements.txt`](https://github.com/dipy/dipy/blob/master/doc/doc-requirements.txt) does not list it.

> To generate the doc and see the result: make -C doc html-after-examples

Thanks for the pointer :+1:. I believe the corresponding CI build should put in place to free developers from the burden of doing that additional step.

skoudoro:
> I assume that this warning/error was already present before this PR. It looks like Sphinx needs Pillow for the scale option to work. The doc-requirements.txt does not list it.

`Pillow` is installed in my environment, I do not think this is the reason. Could you fix this issue here or you prefer a dedicated PR? 

> Thanks for the pointer 👍. I believe the corresponding CI build should put in place to free developers from the burden of doing that additional step.

Like writing tests, I think it is a good practice to generate the documentation locally to check your change before pushing or creating a PR. Although I agree we should have a dedicated CI's, I think the developer should always check his work.

jhlegarreta:
For the time being, maybe the `scale` issue can be addressed in a separate PR as it is a warning I believe. Thanks for having checked the first version of the generated documentation.

skoudoro:
After some thinking, the issue should be fixed on another PR.  Thank you for this @jhlegarreta. Merging

PR #2284, opened 2020-11-08
[DOC] Fix incorrect default output directory in doctsrings

areeshatariq:
If no output directory is specified, the results are saved to the current directory, not the input data directory.


jhlegarreta (approved):

skoudoro:
Thanks @areeshatariq

PR #2287, opened 2020-11-09
[DOC] Remove default values from doctstrings for reconstruction methods

areeshatariq:
It becomes difficult to maintain default values in docstrings. Default values are mentioned in function parameters.
Similarly, will remove default values from docstrings for the remaining methods


jhlegarreta (approved):
Looks good to me. As the message body says, duplicating values is error-prone and involves heavier maintenance burden.

Making this extensive to the rest of the workflows would be welcome.

skoudoro:
thanks! merging

PR #2296, opened 2020-11-19
[DOC] Fix the 'None' value in num_threads attribute

areeshatariq:
Only positive integers can be passed to the num_threads attribute. ``None`` does not work.

### DIFF
--- a/doc/interfaces/gibbs_unringing_flow.rst
+++ b/doc/interfaces/gibbs_unringing_flow.rst
@@ -17,8 +17,8 @@ Magnetic Resonance (MR) images are reconstructed from the Fourier coefficients
 of acquired k-space images. Since only a finite number of Fourier coefficients
 can be acquired in practice, reconstructed MR images can be corrupted by Gibbs
 artefacts, which is manifested by intensity oscillations adjacent to edges of
-different tissue types [Veraart15]_. Although this artefact affects MR images in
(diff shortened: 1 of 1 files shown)


dipy/workflows/denoise.py
```diff
@@ -257,10 +257,9 @@ def run(self, input_files, slice_axis=2, n_points=3, num_threads=1,
             third axis.
         n_points : int, optional
             Number of neighbour points to access local TV (see note).
-        num_threads : int or None, optional
-            Number of threads. Only applies to 3D or 4D `data` arrays. If None
-            then all available threads will be used. Otherwise, must be a
-            positive integer.
+        num_threads : int, optional
+            Number of threads. Only applies to 3D or 4D `data` arrays. Must be
+            a positive integer.
```
jhlegarreta, line 262: Has this been agreed upon? Not sure this change is desired: this would mean that there would not be a way to use all available threads from the CLIs unless the users know the limits of their hardware. @skoudoro ?

skoudoro: I confirm this change is not desired. Can you put it back as it was @areeshatariq? Thank you

areeshatariq: Addressed in 4ff61ad

jhlegarreta (approved):
Thanks.

PR #2333, opened 2021-03-17
Add __str__ to GradientTable

### LINKED ISSUE #2331: gtab.info does not print anything
https://github.com/dipy/dipy/blob/10621cd0ffaf8c1f16b76931486b669a4da33a81/dipy/core/gradients.py#L215 info from GradientTable doesn't print anything in Spyder and Jupyter. Perhaps because it uses logger which might be disabled by Spyder and Jupyter and works best for CLIs only?

skoudoro:
This is PR is to fix #2331. The work done here is:
- replace `logging` by `print`
- add `__str__` function to `GradientTable` class

This PR should be merged after #2332

### DIFF
--- a/dipy/core/gradients.py
+++ b/dipy/core/gradients.py
@@ -212,13 +212,18 @@ class GradientTable(object):
         return self.gradients / denom
 
     @property
-    def info(self):
-        logger.info('B-values shape (%d,)' % self.bvals.shape)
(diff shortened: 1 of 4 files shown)


skoudoro:
This is ready to go. 

if someone could have a look 😄

arokem:
Yep! Thanks!

PR #2341, opened 2021-03-25
Allow use of all threads in the gibbs ringing workflow

### LINKED ISSUE #2300: NF: Add "None" options in the CLIs
## Description
The ``None`` value does not work in the CLIs as the value of attributes. 
For example: Adding ``num_threads`` attribute with value ``None`` using ``dipy_gibbs_ringing`` command should be appropriately processed to use all the available threads but currently only positive integers can be passed to the ``num_threads`` attribute.

drombas:
Closes #2300 

The dipy_gibbs_ringing workflow raised an error when the user specified `num_threads = None` to use all available cores. This PR fixes that by configuring `0` instead of `None` as the value for using all cores.

## Proposed Changes

  - Modify the behaviour of gibbs_removal to choose all available threads when  the user sets `num_threads` to `0`
  - Update corresponding docstrings, tests, example and documentation

### DIFF
--- a/dipy/denoise/gibbs.py
+++ b/dipy/denoise/gibbs.py
@@ -255,7 +255,7 @@ def gibbs_removal(vol, slice_axis=2, n_points=3, inplace=True, num_threads=1):
         a new array.
         Default is set to True.
     num_threads : int, optional
-        Number of threads. Only applies to 3D or 4D `data` arrays. If 0 then
+        Number of threads. Only applies to 3D or 4D `data` arrays. If <=0 then
         all available threads will be used. Otherwise, must be a positive
         integer.
         Default is set to 1.
@@ -295,9 +295,6 @@ def gibbs_removal(vol, slice_axis=2, n_points=3, inplace=True, num_threads=1):
 
     if not isinstance(num_threads, int):
         raise TypeError("num_threads must be an int.")
-    else:
-        if num_threads < 0:
-            raise ValueError("num_threads must be >= 0.")
 
     # check the axis corresponding to different slices
     # 1) This axis cannot be larger than 2
@@ -327,7 +324,7 @@ def gibbs_removal(vol, slice_axis=2, n_points=3, inplace=True, num_threads=1):
     if nd == 2:
         vol[:, :] = _gibbs_removal_2d(vol, n_points=n_points, G0=G0, G1=G1)
     else:
-        if num_threads == 0:
+        if num_threads <= 0:
             pool = Pool()
         else:
             pool = Pool(num_threads)
--- a/dipy/denoise/gibbs.py
+++ b/dipy/denoise/gibbs.py
@@ -255,10 +255,8 @@ def gibbs_removal(vol, slice_axis=2, n_points=3, inplace=True, num_threads=1):
         a new array.
         Default is set to True.
     num_threads : int, optional
-        Number of threads. Only applies to 3D or 4D `data` arrays. If <=0 then
-        all available threads will be used. Otherwise, must be a positive
-        integer.
-        Default is set to 1.
+        Number of threads. If -1 then all available threads will be used.
+        Default is 1. Only applies to 3D or 4D `data` arrays.
 
     Returns
     -------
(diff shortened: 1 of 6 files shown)


doc/interfaces/gibbs_unringing_flow.rst
```diff
@@ -32,17 +32,15 @@ ringing-free volume::
...
+You can also specify optional arguments such as:
+
+- ``num_threads``: the number of threads to be used in parallel to accelerate processing. Set it to ``0``  to use all available threads.
+- ``out_dir``: directory to save the output.
 
 To run the Gibbs unringing on the data it suffices to execute the
 ``dipy_gibbs_ringing`` command, e.g.::
 
-    dipy_gibbs_ringing data/tissue_data/t1_brain_denoised.nii.gz --num_threads 4 --out_dir "gibbs_ringing_output"
+    dipy_gibbs_ringing data/tissue_data/t1_brain_denoised.nii.gz --num_threads 0 --out_dir "gibbs_ringing_output"
```
skoudoro, line 43: Can you keep this to 4? Thank you

skoudoro (commented):
Thank you for doing this @drombas!

See below a tiny comment. A more complete review will come a bit later.

doc/interfaces/gibbs_unringing_flow.rst
drombas: sure

drombas:
Hi @skoudoro. As in #2347 the failed tests do not seem to be related to this. Could we restart them?

ShreyasFadnavis:
> Hi @skoudoro. As in #2347 the failed tests do not seem to be related to this. Could we restart them?

Restarted!

drombas:
Thanks @ShreyasFadnavis, It worked!

dipy/denoise/gibbs.py
```diff
@@ -293,12 +293,11 @@ def gibbs_removal(vol, slice_axis=2, n_points=3, inplace=True, num_threads=1):
...
-    if (not isinstance(num_threads, int)) and (num_threads is not None):
-        raise TypeError("num_processes must be an int or None.")
+    if not isinstance(num_threads, int):
+        raise TypeError("num_threads must be an int.")
     else:
-        if isinstance(num_threads, int):
-            if num_threads <= 0:
-                raise ValueError("num_processes must be > 0.")
+        if num_threads < 0:
+            raise ValueError("num_threads must be >= 0.")
```
skoudoro, line 300: I think, you do not need this.  `-2`, `-1` or `0` should be the same: All core use. it gives a bit more flexiblity

dipy/denoise/gibbs.py
```diff
@@ -328,7 +327,7 @@ def gibbs_removal(vol, slice_axis=2, n_points=3, inplace=True, num_threads=1):
     if nd == 2:
         vol[:, :] = _gibbs_removal_2d(vol, n_points=n_points, G0=G0, G1=G1)
     else:
-        if num_threads is None:
+        if num_threads == 0:
```
skoudoro, line 330: to follow-up the comment above  `num_thread =< 0:`

skoudoro (commented):
Hi @drombas,

Overall, it looks good to me. we need to apply the same logic in different places.
Can you add a note on the [api_change.rst](https://github.com/dipy/dipy/blob/master/doc/api_changes.rst) files. you will have to create a new section for dipy 1.5.

Thank you!

drombas:
A couple of non-related tests still fail, should we restart them?

ShreyasFadnavis:
> A couple of non-related tests still fail, should we restart them?

Done!

skoudoro:
Hi @drombas,

Since you will have a more generic function in #2352 to manage threads, should we close this PR in favor of #2352. 
You will apply your change here to the other branch.

What do you think?

drombas:
Hi! I'm ok with doing everything on #2352. 

Just to mention, after digging a bit more I noticed that [this](https://github.com/dipy/dipy/blob/e96c440ccfc0df550e9baa5c2f6b3d0e579bfa11/dipy/denoise/gibbs.py#L332) and [reslice.py](https://github.com/dipy/dipy/blob/e96c440ccfc0df550e9baa5c2f6b3d0e579bfa11/dipy/align/reslice.py#L100) rely on `Pool()` instead of OpenMP (as they are in python, I guess) and I am not sure we should/could use the same generic function to handle them too. 

Indeed, that might explain the difference between the default value of `num_threads` (`None` for cython functions using OpenMP and  `1` for those in python using `Pool()`). Anyway, we can discuss it in #2352.

skoudoro:
> Just to mention, after digging a bit more I noticed that this and reslice.py rely on Pool() instead of OpenMP (as they are in python, I guess) and I am not sure we should/could use the same generic function to handle them too.

I agree, we should use the same function to handle them. 

> Anyway, we can discuss it in #2352.

and yes let's discuss it on this PR. So I let you close this one.

PR #2348, opened 2021-04-03
RF: Use new name for this function.

arokem:
This shouid eliminate warnings that now routinely appear

### DIFF
--- a/dipy/direction/closest_peak_direction_getter.pyx
+++ b/dipy/direction/closest_peak_direction_getter.pyx
@@ -190,7 +190,7 @@ cdef class PmfGenDirectionGetter(BasePmfDirectionGetter):
             selecting the tracking direction.
         basis_type : name of basis
             The basis that ``shcoeff`` are associated with.
-            ``dipy.reconst.shm.real_sym_sh_basis`` is used by default.
+            ``dipy.reconst.shm.real_sh_descoteaux`` is used by default.
         relative_peak_threshold : float in [0., 1.]
             Used for extracting initial tracking directions. Passed to
             peak_directions.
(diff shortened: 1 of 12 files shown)


skoudoro (commented):
Hi @arokem,

Thank you for this update!

it seems you are missing 2-3  `real_sph_harm` in `test_shm.py`. Also, you will have to update `test_default_lambda_csdmodel`.

Thanks !

arokem:
Thanks for taking a look! I also took the opportunity to replace `real_sym_sh_basis` with `real_sh_descoteaux`.

skoudoro (commented):
Hi @arokem,

You forgot to address the problem with `test_default_lambda_csdmodel`. You just need to update the number of warnings to be caught. 

After that, it should be ready to be in.

Thanks!

arokem:
OK - I think I got it all now. Thanks for the quick review!

skoudoro (approved):
All Green, CI's Happy, I will merge this if there is no additional comment by the end of the day.

Thank you for this @arokem!

jhlegarreta:
@arokem By changing the method name in here:
https://github.com/dipy/dipy/commit/343f81159f1a648178d1c58d8bf56241e440e350#diff-bcbfa34194cf36d07616911bb135804418eb5903801f5c0f2d65f24a88c6ada1R125

a duplication was introduced with an existing method name:
https://github.com/dipy/dipy/blob/master/dipy/reconst/tests/test_shm.py#L164

Thus, only the second one is being tested:
https://github.com/dipy/dipy/blob/master/dipy/reconst/tests/test_shm.py#L125
https://github.com/dipy/dipy/blob/master/dipy/reconst/tests/test_shm.py#L164

If the first one, which is calling the deprecated `real_sym_sh_mrtrix` method is renamed and kept, the deprecation message should be dealt with appropriately (cross-referencing PR #2478).

PR #2349, opened 2021-04-05
[ENH] Allow for other statistics, like median, in afq_profile

36000:
Median can be a useful way to calculate tract profiles.

### DIFF
--- a/dipy/stats/analysis.py
+++ b/dipy/stats/analysis.py
@@ -218,7 +218,7 @@ def gaussian_weights(bundle, n_points=100, return_mahalnobis=False,
     return w / np.sum(w, 0)
 
 
-def afq_profile(data, bundle, affine, n_points=100,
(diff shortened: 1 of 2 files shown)


skoudoro (commented):
Hi @36000,

Thank you for this update! 

Overall, it looks good to me. Could you just address [the pep8 above](https://github.com/dipy/dipy/pull/2349#issuecomment-813554625) and then it should be ready to be merged. 

@arokem, can you give a look at this extended feature for `afq_profile`?

Thank you both!

ps: Codecov comments are back! 🥳  🎉

arokem:
This looks good to me! I think that the CI failures are unrelated. I will try to restart it.

Looks like codecov might be a bit set too sensitive.

skoudoro:
thank you for this @36000. merging

PR #2350, opened 2021-04-05
[FIX] Use npy_intp variables instead of int and size_t to iterate over numpy arrays

### LINKED ISSUE #423: Use npy_intp variables instead of int and size_t to iterate over numpy arrays
Verify that, in all cython modules, we use npy_intp variables to iterate over numpy arrays instead of int and size_t. According to this thread:
http://mail.scipy.org/pipermail/numpy-discussion/2010-July/051901.html

size_t is even more dangerous. This is the exact comment:

"An initial attempt to implement this feature tried to use size_t. It
quickly turned out that this cannot work: Python uses negative indices
in many places (to indicate counting from the end). Even in places where
size_t would be usable, too many reformulations of code where necessary,
e.g. in loops like:

for(index = length-1; index >= 0; index--)

This loop will never terminate if index is changed from int to size_t."

skoudoro:
This PR fixes #423.

- [x] replace `cimport numpy as np`  by `cimport numpy as cnp` to respect the convention and be explicit
- [x] replace `size_t` by `npy_intp` (for more info, look at #423)
- [x] replace `int` by `npy_intp` when we use it for index

### DIFF
--- a/dipy/align/bundlemin.pyx
+++ b/dipy/align/bundlemin.pyx
@@ -352,10 +352,10 @@ def distance_matrix_mdf(streamlines_a, streamlines_b):
         distance matrix
     """
     cdef:
-        size_t i, j, lentA, lentB
+        cnp.npy_intp i, j, lentA, lentB
     # preprocess tracks
     cdef:
-        size_t longest_track_len = 0, track_len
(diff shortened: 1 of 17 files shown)


Garyfallidis:
Nice work. I forgot about ``size_t``. :+1:

PR #2351, opened 2021-04-06
Updating Vec2vec_rotmat to deal with numerical issues

nasimanousheh:
``vec2vec_rotmat`` function can provide incorrect results when the dot product of the input vectors is slightly a bit beyond one. This is happening when both vectors (u, v) are close to collinear. Or at least this is what I currently observe. I found this problem when I used this function for a FURY application. In both bellow examples, I used ``vec2vec_rotmat`` function to rotate the arrows. In these examples, one arrow that is rotating around the center (0,0,0) is the leader and the other arrow is follower. 

Here in  ``vec2vec_rotmat`` function, using ``cosa = np.dot(u, v)`` . Notice that at some point the follower vector (purple) is flipping which is not correct.
without_clip

Here in  ``vec2vec_rotmat`` function, ``cosa = np.clip(np.dot(u, v), -1, 1)``. The follower vector does not flip anymore.
with_clip

### DIFF
--- a/dipy/core/tests/test_geometry.py
+++ b/dipy/core/tests/test_geometry.py
@@ -219,6 +219,17 @@ def test_vec2vec_rotmat():
         R = vec2vec_rotmat(a, b)
         assert_array_almost_equal(np.dot(R, a), b)
 
+    a = np.array([0.33950729, 0.92041729, 0.1938216])
+    a_norm = np.linalg.norm(a)
+    print(a_norm)
+    b = np.array([0.40604787, 0.97518325, 0.2057731])
+    b_norm = np.linalg.norm(b)
+    print(b_norm)
+    R = vec2vec_rotmat(a, b)
+    res = np.dot(a, b)
+    if (res > 1):
+        print("cosa is greater than 1!!")
+    tmp = 1
 
 def test_compose_transformations():
 
--- a/dipy/core/tests/test_geometry.py
+++ b/dipy/core/tests/test_geometry.py
@@ -221,14 +221,19 @@ def test_vec2vec_rotmat():
 
     a = np.array([0.33950729, 0.92041729, 0.1938216])
     a_norm = np.linalg.norm(a)
+    a2 = a/a_norm
     print(a_norm)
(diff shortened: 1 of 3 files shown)


dipy/core/tests/test_geometry.py
```diff
@@ -219,6 +219,22 @@ def test_vec2vec_rotmat():
         R = vec2vec_rotmat(a, b)
         assert_array_almost_equal(np.dot(R, a), b)
 
+    a = np.array([0.33950729, 0.92041729, 0.1938216])
```
arokem, line 222: Are these vectors where this numerical instability happens? It would be good to build a test around them if so.

arokem:
Thanks for doing this! I think that the clipping makes sense. And adding a test would be great.

dipy/core/tests/test_geometry.py
nasimanousheh: Yes. These two vectors are the examples that their dot product is slightly more than 1 and create instability:
 a = np.array([0.33950729, 0.92041729, 0.1938216])
 b = np.array([0.40604787, 0.97518325, 0.2057731])
Sure. I will build a test for these vectors.

nasimanousheh:
A bit of explanation of the progress here. What is happening is that: 
a) when two vectors (u and v) are numerically slightly larger than unit vectors and  
b) not exactly the same (close to colinear but not exactly colinear), 
using `cosa = np.dot(u, v)` will make the cosine bigger than 1 or smaller than -1. To avoid this issue, we can use `np.clip(np.dot(u, v), -1, 1)` instead of `np.dot(u, v)`.
I added the tests and included the fix . 
In the future you may consider always checking if the vectors are unit norm and otherwise raise a warning however this might be a bit of overkill as you are calling this function very often.

nasimanousheh:
Tests failures seem irrelevant to my changes.

skoudoro (approved):
Does anyone want to have a last look? If not, I will go ahead and merge it tomorrow.

skoudoro:
thank you @nasimanousheh

Garyfallidis:
:+1:

PR #2352, opened 2021-04-06
RF: configure num_threads==-1 as the value to use all cores

drombas:
Related to #2300 and a continuation of #2341.

### Proposed Changes

  - Configure  ``num_threads<=0`` as the option to use all cores across the codebase
  - Modify the name of the argument ``num_proceses --> num_threads`` in **reslice.py** and **test_reslice.py**
  - Delete the ``num_threads`` argument where it is not used

### Important point
Unlike in #2341, most of the functions edited here used all cores by default (using ``None``) so, to keep this behavior, I set the new default to ``0``.  This implies that:
1. The number of used cores by default differs between functions. Does it make sense to use all cores by default in some functions and only ``1`` in others?
2. This PR modifies the default value of ``num_threads`` in several functions, which was one of the main concerns discussed in #2300. What do you think @jhlegarreta,@skoudoro?

### DIFF
--- a/dipy/align/bundlemin.pyx
+++ b/dipy/align/bundlemin.pyx
@@ -96,7 +96,7 @@ def _bundle_minimum_distance_matrix(double [:, ::1] static,
         Distance matrix
     num_threads : int, optional
         Number of threads. If -1 (default) then all available threads will be
-        used.        
+        used.
 
     Returns
     -------
@@ -160,7 +160,7 @@ def _bundle_minimum_distance(double [:, ::1] static,
     num_threads : int, optional
         Number of threads. If -1 (default) then all available threads will be
         used.
-        
+
     Returns
     -------
     cost : double
--- a/dipy/align/bundlemin.pyx
+++ b/dipy/align/bundlemin.pyx
@@ -74,7 +74,7 @@ def _bundle_minimum_distance_matrix(double [:, ::1] static,
                                     cnp.npy_intp moving_size,
                                     cnp.npy_intp rows,
                                     double [:, ::1] D,
-                                    num_threads=None):
+                                    num_threads=0):
     """ MDF-based pairwise distance optimization function
 
     We minimize the distance between moving streamlines of the same number of
@@ -94,8 +94,8 @@ def _bundle_minimum_distance_matrix(double [:, ::1] static,
         Number of points per streamline
     D : 2D array
         Distance matrix
-    num_threads : int
-        Number of threads. If None (default) then all available threads
+    num_threads : int, optional
+        Number of threads. If <=0 (default) then all available threads
         will be used.
 
     Returns
@@ -108,7 +108,7 @@ def _bundle_minimum_distance_matrix(double [:, ::1] static,
         int all_cores = openmp.omp_get_num_procs()
         int threads_to_use = -1
 
-    if num_threads is not None:
+    if num_threads > 0:
         threads_to_use = num_threads
     else:
         threads_to_use = all_cores
@@ -126,7 +126,7 @@ def _bundle_minimum_distance_matrix(double [:, ::1] static,
                                                &moving[j * rows, 0],
                                                rows)
 
-    if have_openmp and num_threads is not None:
+    if have_openmp and num_threads >0:
         openmp.omp_set_num_threads(all_cores)
 
     return np.asarray(D)
@@ -137,7 +137,7 @@ def _bundle_minimum_distance(double [:, ::1] static,
                              cnp.npy_intp static_size,
                              cnp.npy_intp moving_size,
                              cnp.npy_intp rows,
-                             num_threads=None):
+                             num_threads=0):
     """ MDF-based pairwise distance optimization function
 
     We minimize the distance between moving streamlines of the same number of
@@ -155,8 +155,8 @@ def _bundle_minimum_distance(double [:, ::1] static,
         Number of moving streamlines
     rows : int
         Number of points per streamline
-    num_threads : int
-        Number of threads. If None (default) then all available threads
+    num_threads : int, optional
+        Number of threads. If <=0 (default) then all available threads
         will be used.
 
     Returns
@@ -180,7 +180,7 @@ def _bundle_minimum_distance(double [:, ::1] static,
         int all_cores = openmp.omp_get_num_procs()
         int threads_to_use = -1
 
-    if num_threads is not None:
+    if num_threads > 0:
         threads_to_use = num_threads
     else:
         threads_to_use = all_cores
@@ -236,7 +236,7 @@ def _bundle_minimum_distance(double [:, ::1] static,
 
         dist = 0.25 * dist * dist
 
-    if have_openmp and num_threads is not None:
+    if have_openmp and num_threads > 0:
         openmp.omp_set_num_threads(all_cores)
 
     return dist
--- a/dipy/align/bundlemin.pyx
+++ b/dipy/align/bundlemin.pyx
@@ -13,7 +13,9 @@ from safe_openmp cimport have_openmp
 from cython.parallel import prange
 from libc.stdlib cimport malloc, free
 from libc.math cimport sqrt, sin, cos
-from multiprocessing import cpu_count
+
+from dipy.utils.omp import cpu_count, determine_num_threads
+from dipy.utils.omp cimport set_num_threads, restore_default_num_threads
 
 cdef cnp.dtype f64_dt = np.dtype(np.float64)
 
@@ -95,8 +97,11 @@ def _bundle_minimum_distance_matrix(double [:, ::1] static,
     D : 2D array
         Distance matrix
     num_threads : int, optional
-        Number of threads. If -1 (default) then all available threads will be
-        used.
+        Number of threads to be used for OpenMP parallelization. If None
+        (default) the value of OMP_NUM_THREADS environment variable is used
+        if it is set, otherwise all available threads are used. If < 0 the
+        maximal number of threads minus |num_threads + 1| is used (enter -1 to
+        use as many threads as possible). 0 raises an error.
 
     Returns
     -------
@@ -105,19 +110,10 @@ def _bundle_minimum_distance_matrix(double [:, ::1] static,
 
     cdef:
         cnp.npy_intp i=0, j=0, mov_i=0, mov_j=0
-        int all_cores = openmp.omp_get_num_procs()
         int threads_to_use = -1
 
-    if num_threads in (None, -1):
-        threads_to_use = all_cores
-    elif num_threads > 0:
-        threads_to_use = num_threads
-    else:
-        raise ValueError("num_threads must be > 0 or -1 (all cores)")
-
-    if have_openmp:
-        openmp.omp_set_dynamic(0)
-        openmp.omp_set_num_threads(threads_to_use)
+    threads_to_use = determine_num_threads(num_threads)
+    set_num_threads(threads_to_use)
 
     with nogil:
 
@@ -128,8 +124,8 @@ def _bundle_minimum_distance_matrix(double [:, ::1] static,
                                                &moving[j * rows, 0],
                                                rows)
 
-    if have_openmp and num_threads not in (None, -1):
-        openmp.omp_set_num_threads(all_cores)
+    if num_threads is not None:
+        restore_default_num_threads()
 
     return np.asarray(D)
 
@@ -158,8 +154,11 @@ def _bundle_minimum_distance(double [:, ::1] static,
     rows : int
         Number of points per streamline
     num_threads : int, optional
-        Number of threads. If -1 (default) then all available threads will be
-        used.
+        Number of threads to be used for OpenMP parallelization. If None
+        (default) the value of OMP_NUM_THREADS environment variable is used
+        if it is set, otherwise all available threads are used. If < 0 the
+        maximal number of threads minus |num_threads + 1| is used (enter -1 to
+        use as many threads as possible). 0 raises an error.
 
     Returns
     -------
@@ -179,19 +178,10 @@ def _bundle_minimum_distance(double [:, ::1] static,
         double * min_j
         double * min_i
         openmp.omp_lock_t lock
-        int all_cores = openmp.omp_get_num_procs()
         int threads_to_use = -1
 
-    if num_threads in (None, -1):
-        threads_to_use = all_cores
-    elif num_threads > 0:
-        threads_to_use = num_threads
-    else:
-        raise ValueError("num_threads must be > 0 or -1 (all cores)")
-
-    if have_openmp:
-        openmp.omp_set_dynamic(0)
-        openmp.omp_set_num_threads(threads_to_use)
+    threads_to_use = determine_num_threads(num_threads)
+    set_num_threads(threads_to_use)
 
     with nogil:
 
@@ -240,8 +230,8 @@ def _bundle_minimum_distance(double [:, ::1] static,
 
         dist = 0.25 * dist * dist
 
-    if have_openmp and num_threads not in (None, -1):
-        openmp.omp_set_num_threads(all_cores)
+    if num_threads is not None:
+        restore_default_num_threads()
 
     return dist
 
--- a/dipy/align/bundlemin.pyx
+++ b/dipy/align/bundlemin.pyx
@@ -74,7 +74,7 @@ def _bundle_minimum_distance_matrix(double [:, ::1] static,
                                     cnp.npy_intp moving_size,
                                     cnp.npy_intp rows,
                                     double [:, ::1] D,
-                                    num_threads=0):
+                                    num_threads=None):
     """ MDF-based pairwise distance optimization function
 
     We minimize the distance between moving streamlines of the same number of
@@ -95,8 +95,8 @@ def _bundle_minimum_distance_matrix(double [:, ::1] static,
     D : 2D array
         Distance matrix
     num_threads : int, optional
-        Number of threads. If <=0 (default) then all available threads
-        will be used.
+        Number of threads. If -1 (default) then all available threads will be
+        used.        
 
     Returns
     -------
@@ -108,10 +108,12 @@ def _bundle_minimum_distance_matrix(double [:, ::1] static,
         int all_cores = openmp.omp_get_num_procs()
         int threads_to_use = -1
 
-    if num_threads > 0:
+    if num_threads in (None, -1):
+        threads_to_use = all_cores
+    elif num_threads > 0:
         threads_to_use = num_threads
     else:
-        threads_to_use = all_cores
+        raise ValueError("num_threads must be > 0 or -1 (all cores)")
 
     if have_openmp:
         openmp.omp_set_dynamic(0)
@@ -126,7 +128,7 @@ def _bundle_minimum_distance_matrix(double [:, ::1] static,
                                                &moving[j * rows, 0],
                                                rows)
 
-    if have_openmp and num_threads >0:
+    if have_openmp and num_threads not in (None, -1):
         openmp.omp_set_num_threads(all_cores)
 
     return np.asarray(D)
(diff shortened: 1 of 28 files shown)


grlee77:
As an alternative, I would also consider using `num_threads=-1` to mean the maximum number of workers (similar array[-1] giving th last element in an array). That is the approach taken for the `workers` argument in SciPy functions ([example](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.differential_evolution.html)) and for the `n_jobs` argument in [joblib.Parallel](https://joblib.readthedocs.io/en/latest/generated/joblib.Parallel.html#joblib-parallel).

Unfortunately there is not good consensus on either this behavior or the name for the `num_threads`/`workers`/`n_jobs` argument across scientific Python libraries though! (see some discussion in https://github.com/scikit-image/scikit-image/issues/4876)

jhlegarreta:
Thanks for taking care of this @drombas !

> Configure num_threads<=0 as the option to use all cores across the codebase

IMO using two different values to mean the same thing is misleading.

> The number of used cores by default differs between functions. Does it make sense to use all cores by default in some functions and only 1 in others?

At first sight, this looks inconsistent/undesirable to me, but have not investigated at the reasons, if any. Maybe others can elaborate on this.

> This PR modifies the default value of num_threads in several functions, which was one of the main concerns discussed in #2300. What do you think @jhlegarreta,@skoudoro?

Although not desirable, I'd definitely be for it if the new values are the ones that make sense/are sensible defaults.

Interesting what @grlee77 says https://github.com/dipy/dipy/pull/2352#issuecomment-815026058 . That also broadens the scope of our issue. As for the terms used, I wouldn't know to tell which term is more accurate/honors its purpose, but have seen mixed use in the past elsewhere with the terms `process`, `thread`, `job`, `worker`.

dipy/align/reslice.py
```diff
@@ -13,7 +13,7 @@ def _affine_transform(kwargs):
 
 
 def reslice(data, affine, zooms, new_zooms, order=1, mode='constant', cval=0,
-            num_processes=1):
+            num_threads=1):
```
skoudoro, line 16: keep `num_processes` in this function

dipy/align/reslice.py
```diff
@@ -97,7 +96,7 @@ def reslice(data, affine, zooms, new_zooms, order=1, mode='constant', cval=0,
                     _kwargs = {'input': data[..., i]}
                     _kwargs.update(kwargs)
                     params.append(_kwargs)
-                pool = Pool(num_processes)
+                pool = Pool(num_threads)
```
skoudoro, line 99: same comments as above

skoudoro (commented):
Hi @drombas,

Thank you for doing this again!

After all comments and more thinking, see below some suggestions:

> Modify the name of the argument num_proceses --> num_threads in reslice.py and test_reslice.py

you should keep `num_processes` here. This is different than `num_threads`. To simplify it means we spawn processes (run new python interpreter to parallelize). 

 > Delete the num_threads argument where it is not used

you will have to keep it and start a deprecation cycle to warn the users

> The number of used cores by default differs between functions. Does it make sense to use all cores by default in some functions and only 1 in others?

To me it makes sense. It depends a lot on the algorithms. Some of them are greedy and we do not want to kill/freeze the laptop of the user (they will blame the algo which is wrong)

> This PR modifies the default value of num_threads in several functions, which was one of the main concerns discussed in #2300. What do you think @jhlegarreta,@skoudoro?

After @jhlegarreta and @grlee77 comments, I recommend that you keep the default value at `None` or the value indicated, and inside each function just initialize the `num_threads` by doing:
```python
num_threads = num_threads or -1
```
Does it make sense to you @drombas?

drombas:
Thank you all for your comments! TBH I don't have a strong opinion on this kind of conventions.

On your points @skoudoro:

> you should keep `num_processes` here. This is different than `num_threads`. To simplify it means we spawn processes (run new python interpreter to parallelize).

Sorry for that, thought it was just a different naming for the same.

> you will have to keep it and start a deprecation cycle to warn the users
 
Ok, I'll keep those.

> I recommend that you keep the default value at `None` or the value indicated, and inside each function just initialize the `num_threads` by doing:
> 
> ```python
> num_threads = num_threads or -1
> ```

Just to clarify, you mean we keep `None` as default:
```python
def anyFunction(...,num_threads=None):
```
and then inside the function we initialize it to `-1` or `1` (depending on the case):
```python
if num_threads is None:
    num_threads = -1
```
It sounds reasonable as it doesn't change the default but incorporates `-1` as the option for all cores.

As for selecting between `threads`,`jobs`,`workers`,... we could leave it for another issue/discussion ans stick to `num_threads` for the moment.

skoudoro:
> It sounds reasonable as it doesn't change the default but incorporates -1 as the option for all cores.

Exactly and this `num_threads = num_threads or -1` is equivalent to

```python
if num_threads is None:
    num_threads = -1
```

drombas:
Before reviewing the code, please notice that I added some extra tests to account for invalid `num_threads` values. 

Could we restart those failing tests? (All related tests passed locally )

jhlegarreta:
Thanks for the effort @drombas . Restarted the tests.

dipy/align/bundlemin.pyx
```diff
@@ -94,9 +94,9 @@ def _bundle_minimum_distance_matrix(double [:, ::1] static,
         Number of points per streamline
     D : 2D array
         Distance matrix
-    num_threads : int
-        Number of threads. If None (default) then all available threads
-        will be used.
+    num_threads : int, optional
+        Number of threads. If -1 (default) then all available threads will be
+        used.
```
jhlegarreta, line 99: Please remove the trailing whitespaces.

dipy/denoise/denspeed.pyx
```diff
@@ -57,7 +57,7 @@ def nlmeans_3d(arr, mask=None, sigma=None, patch_radius=1,
 
     if sigma.ndim != 3:
         raise ValueError('sigma needs to be a 3D ndarray', sigma.shape)
-
+
```
jhlegarreta, line 60: Remove the unnecessary whitespaces.

dipy/denoise/shift_twist_convolution.pyx
```diff
@@ -89,9 +89,9 @@ def convolve_sf(odfs_sf, kernel, test_mode=False, num_threads=None, normalize=Tr
         The 5D lookup table
     test_mode : boolean
         Reduced convolution in one direction only for testing
-    num_threads : int
-            Number of threads. If None (default) then all available threads
-            will be used.
+    num_threads : int, optional
+        Number of threads. If -1 (default) then all available threads will be
+        used.
```
jhlegarreta, line 94: Remove the unnecessary whitespaces/tabs.

dipy/denoise/shift_twist_convolution.pyx
```diff
@@ -132,10 +132,10 @@ cdef double [:, :, :, ::1] perform_convolution (double [:, :, :, ::1] odfs,
...
     test_mode : boolean
         Reduced convolution in one direction only for testing
-    num_threads : int
-            Number of threads. If None (default) then all available threads
-            will be used.
-        
+    num_threads : int, optional
+        Number of threads. If -1 (default) then all available threads will be
+        used.
+
```
jhlegarreta, line 138: Remove the unnecessary whitespaces/tabs.

jhlegarreta (changes requested):
I had a quick look at the changes: the docstrings and the default values in the method signatures are contradicting, e.g.:
```
def _bundle_minimum_distance_matrix(double [:, ::1] static,
                                    double [:, ::1] moving,
                                    cnp.npy_intp static_size,
                                    cnp.npy_intp moving_size,
                                    cnp.npy_intp rows,
                                    double [:, ::1] D,
                                    num_threads=None):
(...)
    num_threads : int, optional
        Number of threads. If -1 (default) then all available threads will be
        used.
```

Default is `None`. Have not looked at all signatures, but I had a look at a few of them, and all them fall into this contradiction. I'd dare to say that for people that will be looking at the documentation this is quite confusing, and IMHO it is as confusing from a developer's point of view.

drombas:
Thanks @jhlegarreta !

> Default is None. Have not looked at all signatures, but I had a look at a few of them, and all them fall into this contradiction. I'd dare to say that for people that will be looking at the documentation this is quite confusing,

The intention was to keep `None` as the default but also tell the user that `None` in practice behaves as `-1` (all cores). Any suggestion for a less confusing docstring ?

We could also return to the original plan and get rid of `None` by setting `-1` as the default in the method signature.

jhlegarreta:
> The intention was to keep None as the default but also tell the user that None in practice behaves as -1 (all cores). Any suggestion for a less confusing docstring ?

I still believe that using two values with the same meaning is confusing, and this means that I'd be revisiting the use of `None` in the CLI that triggered the issue this PR tries to fix. I think @grlee77 's proposal https://github.com/dipy/dipy/pull/2352#issuecomment-815026058, beyond the terminology, would not have any such contradiction.

Also, it looks like every method involved is forced to do its `if/else` block or implent its own logic to potentially arrive at the same conclusion (provided that the convention is clear); I'd have one single utility function named e.g. `determine_threads(num_threads)`, `compute_threads(num_threads)` or similar, that is defined at a single place, and shared and used across all involved methods. Might not be straightforward, and might need some deprecation cycle if the logic changes for some methods, but from that moment on, we'd only need to have a look at/change things at a single place (less prone to errors, etc).

Sorry, but I feel that I cannot comment further being still not convinced by the solution.

Thanks for the effort @drombas .

skoudoro:
> this means that I'd be revisiting the use of None in the CLI that triggered the issue this PR tries to fix

This PR does not try to fix this anymore @jhlegarreta. it tries to standardize how we set up `num_threads` since there are different rules in each function. Also, it tries to keep the backward compatibility with the `None` so it is normal that 2 values will have the same meaning. 

> I think @grlee77 's proposal #2352 (comment), beyond the terminology, would not have any such contradiction.

that's why, the first step here is `-1` for all cores. Then, a new PR will be needed to deprecate `None` in this specific case. (concerning the other case, we need to manage None at the worflow level). The proposal does not say what is the behavior when you have 0 or None as a parameter. Should we raise an error? or interpret it as all_cores which means 3 values for the same behavior (-1, 0, None). I need to look at scikit-learn code or maybe @grlee77 have an answer.

> also, it looks like every method involved is forced to do its if/else block or implement its own logic to potentially arrive at the same conclusion (provided that the convention is clear); I'd have one single utility function named e.g. determine_threads(num_threads), compute_threads(num_threads) or similar, that is defined at a single place, and shared and used across all involved methods

I agree with this point. could you create a function @drombas in `dipy.utils.omp.pyx`? something like `determine_num_threads()` as @jhlegarreta  propose and then use it everywhere. Would be easier to maintain in the future. Thanks a lot!

skoudoro:
@jhlegarreta @drombas @grlee77 : More details about their rules below or [in this link](https://github.com/scikit-learn/scikit-learn/blob/114616d9f6ce9eba7c1aacd3d4a254f868010e25/sklearn/utils/_openmp_helpers.pyx#L18). I am ok to follow the same rules

- For ``n_threads = None``,
  - if the ``OMP_NUM_THREADS`` environment variable is set, return
        ``openmp.omp_get_max_threads()``
   - otherwise, return the minimum between ``openmp.omp_get_max_threads()``
        and the number of cpus, taking cgroups quotas into account. Cgroups 
        quotas can typically be set by tools such as Docker.
      The result of ``omp_get_max_threads`` can be influenced by environment
      variable ``OMP_NUM_THREADS`` or at runtime by ``omp_set_num_threads``.
- For ``n_threads > 0``, return this as the maximal number of threads for
      parallel OpenMP calls.
    - For ``n_threads < 0``, return the maximal number of threads minus
      ``|n_threads + 1|``. In particular ``n_threads = -1`` will use as many
      threads as there are available cores on the machine.
- Raise a ValueError for ``n_threads = 0``.

jhlegarreta:
We might want to check how the above (https://github.com/dipy/dipy/pull/2352#issuecomment-819819348, https://github.com/dipy/dipy/pull/2352#issuecomment-819827813) would work when calling the CLIs, but looks like a step forward.

drombas:
> I agree with this point. could you create a function @drombas in `dipy.utils.omp.pyx`? something like `determine_num_threads()` as @jhlegarreta propose and then use it everywhere. 

okey, let's try that. I think we already have something similar in [omp.pyx](https://github.com/dipy/dipy/blob/e96c440ccfc0df550e9baa5c2f6b3d0e579bfa11/dipy/utils/omp.pyx#L27) that we can adapt to the scikit-learn logic. I will give it a try during next days.

skoudoro:
Hi @drombas, 

Do you think you will have time to finish this PR before Friday and the new DIPY release? or should I move this PR for the next release cycle in June?

Thank you for your feedback

drombas:
Hi @skoudoro,

By tomorrow I should have finished the changes so, if it is ok we can decide tomorrow.

skoudoro:
sounds like a plan 👍🏾 . no problem

dipy/utils/multiproc.py
```diff
@@ -0,0 +1,37 @@
...
+
+    if num_processes == 0:
+        raise ValueError("num_processes cannot be 0")
+
+    if num_processes is None:
+        return cpu_count()
+
+    if num_processes < 0:
+        return max(1, cpu_count() + num_processes + 1)
+    else:
```
skoudoro, line 36: this `else` is not necessary. Can you remove it?

dipy/utils/omp.pyx
```diff
@@ -42,6 +42,43 @@ def _get_default_threads():
...
+
+    if num_threads == 0:
+        raise ValueError("num_threads cannot be 0")
+
+    if num_threads is None:
+        return default_threads
+
+    if num_threads < 0:
+        return max(1, cpu_count() + num_threads + 1)
+    else:
```
skoudoro, line 78: this `else` is not necessary. Can you remove it?

dipy/workflows/reconst.py
```diff
@@ -456,9 +456,11 @@ def run(self, input_files, bvalues_files, bvectors_files, mask_files,
         parallel : bool, optional
             Whether to use parallelization in peak-finding during the
             calibration procedure.
-        nbr_processes : int, optional
+        nbr_processes: int, optional
```
skoudoro, line 459: Please, keep the space here before the colon

dipy/workflows/reconst.py
```diff
@@ -618,9 +620,11 @@ def run(self, input_files, bvalues_files, bvectors_files, mask_files,
...
```
skoudoro, line 623: same as above, Please, keep the space here before the colon

skoudoro (commented):
This is very nice work @drombas. Thanks a lot for that. 

Overall, it looks good. I still need to look at it more carefully. See below some comments. 

Also, Can you add your docstring as a note on the `[api_change.rst](https://github.com/dipy/dipy/blob/master/doc/api_changes.rst)`. You can create a section for DIPY 1.4.1.

Thanks!

drombas:
Thanks for the comments @skoudoro. 

In summary, the selection of the number of cores is now centralized in two files:
- omp.py: for OpenMP parallelization
- multiproc.py: for multiprocessing parallelization

Finally I split it in two as the logic is slightly different: for OpenMP the environment variable OMP_NUM_THREADS is considered while for multiprocessing it is not. I also thought it could be confusing to use omp.pyx to define the logic of parallelization using multiprocessing package. To help with the review a bit [here](https://docs.google.com/spreadsheets/d/16rMzVirivx5tWD39gAnrdPSQ7MhHjBcvuXpNEIXgmjA/edit?usp=sharing) is an organized list of the main changed files.

skoudoro (approved):
It is good to go! 

I will wait until Friday evening to see if there is any additional comment and then go ahead and merge it.

doc/interfaces/gibbs_unringing_flow.rst
```diff
@@ -32,12 +32,11 @@ ringing-free volume::
 
 To run the Gibbs unringing method, we need to specify the path to the input
 data. This path may contain wildcards to process multiple inputs at once.
-You can also specify the optional arguments. In this case, we will be
-specifying the number of threads (``num_threads``) and output directory
-(``out_dir``). The number of threads allows to exploit the available
-computational power and accelerate the processing. The maximum number of
-threads available depends on the CPU of the computer, so users are expected
-to set an appropriate value based on their platform.
```
jhlegarreta, line 40: I think this was written this way having in mind that part of the workflows audience/users might not be that familiar with what threads are, what they represent or how a value that fits in their case can be determined. It does not aim to be a thorough explanation, but I found it helpful.

I think re-using it below could be nice.

Also, although not written in stone, note that we try to wrap at 79 characters.

jhlegarreta (commented):
Thanks for the hard work and for having persevered @drombas.

Thanks for the spreadsheet. Very helpful. On a related note, automatically knowing which methods are multi-threaded can be helpful to developers and users at some point. But that is a separate endeavor.

Not sure if I followed the removal in here
https://github.com/dipy/dipy/pull/2352/files#diff-5fcc99cde72e6ea9a640d32a91756c178d916d26ae479e0da4c766a61c45dd4cL226

But if it's OK, then dismiss the observation.

I am missing dedicated test methods for the accepted values in the `determine_num_processes` and `determine_num_threads` methods.

A minor in-line comment.

The last two comments can be addressed at a latter time if it is preferred to have this merged as soon as possible.

doc/interfaces/gibbs_unringing_flow.rst
drombas: I found the tutorial a bit lengthy but I see your point about the broader audience, let's keep it.

dipy/utils/tests/test_multiproc.py
```diff
@@ -0,0 +1,37 @@
...
+    # A string should raise an error
+    assert_raises(TypeError, determine_num_processes, "0")
+
+    # 1 should be 1
+    assert_equal(determine_num_processes(1), 1)
+
+    # A positive integer should not change
+    assert_equal(determine_num_processes(4), 4)
+
+    # None and -1 shoul be equal (all cores)
```
jhlegarreta, line 23: shoul: should

dipy/utils/tests/test_multiproc.py
```diff
@@ -0,0 +1,37 @@
+""" Testing multiproc utilities
+"""
+
+from dipy.utils.multiproc import determine_num_processes
+from numpy.testing import assert_equal, assert_raises, run_module_suite
+
+
+def test_determine_num_processs():
```
jhlegarreta, line 8: Can this be done for the Cython code as well?

dipy/utils/tests/test_multiproc.py
drombas: damm

jhlegarreta (approved):
:100: for the hard work @drombas.

drombas:
Thanks for the quick feedback!

> Not sure if I followed the removal in here
> https://github.com/dipy/dipy/pull/2352/files#diff-5fcc99cde72e6ea9a640d32a91756c178d916d26ae479e0da4c766a61c45dd4cL226

After the changes that part is not reached. It is true, though, that there is a catch of NotImplementedError exception that I did not implement (not sure how critical it is as it is the only place I saw that check).

skoudoro:
thank you @drombas! merging

jhlegarreta:
I'd say that the exception added in 1e439ce should get tested. Thanks.

drombas:
That exception is raised when the number of cores is undetermined and TBH I don't know how we could test it.

jhlegarreta:
> That exception is raised when the number of cores is undetermined and TBH I don't know how we could test it.

Can the exception be forced to be raised and the expected message or result be checked?

drombas:
> Can the exception be forced to be raised and the expected message or result be checked?

I imagine it should be possible if we knew exactly how the number of cores is determined.

PR #2363, opened 2021-04-07
[ENH] Adding cython file(*.pyx) in documentation

### LINKED ISSUE #366: Some doc missing

### LINKED ISSUE #1302: [DOC] cython (pyx) files are not parsed

skoudoro:
This PR adds `pyx` files in the doc generation. Currently, many functions are not documented in the API because of that. This fixes the problem. I used this opportunity to clean also the documentation

- supersedes and close #1311
- fixes #1302 
- fixes #366
- relax a bit Codecov

### DIFF
--- a/doc/tools/apigen.py
+++ b/doc/tools/apigen.py
@@ -1,5 +1,4 @@
-"""
-Attempt to generate templates for module reference with Sphinx
+"""Attempt to generate templates for module reference with Sphinx.
 
 To include extension modules, first identify them as valid in the
 ``_uri2path`` method, then handle them in the ``_parse_module_with_import``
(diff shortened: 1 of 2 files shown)


doc/tools/apigen.py
```diff
@@ -108,15 +107,18 @@ def set_package_name(self, package_name):
...
+        """Import namespace package."""
         mod = __import__(name)
         components = name.split('.')
         for comp in components[1:]:
             mod = getattr(mod, comp)
         return mod
 
     def _get_object_name(self, line):
-        """ Get second token in line
+        """.Get second token in line.
```
arokem, line 118:
```suggestion
        """Get second token in line.
```

arokem (commented):
Great! I spotted a typo, but other than that no issues here

skoudoro:
Good catch!

PR #2365, opened 2021-04-08
[DOC]: Change defaults in Patch2Self example

ShreyasFadnavis:
Hi everyone,

I know that the defaults in `Patch2Self` were changed recently by @arokem, but they won't be visible to the users until the next release. I am making this small PR for adding it to the documentation to ensure that people still use the correct options!

@skoudoro : would it be possible to update the documentation with these changes before the next release?

### DIFF
--- a/doc/examples/denoise_patch2self.py
+++ b/doc/examples/denoise_patch2self.py
@@ -68,10 +68,20 @@ data.
 hardi_fname, hardi_bval_fname, hardi_bvec_fname = get_fnames('stanford_hardi')
 data, affine = load_nifti(hardi_fname)
 bvals = np.loadtxt(hardi_bval_fname)
-denoised_arr = patch2self(data, bvals)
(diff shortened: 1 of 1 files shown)


skoudoro (approved):
Thank you @ShreyasFadnavis.

If no other comment by the end of the day, I will go ahead and merge this PR.

arokem:
This additional documentation is valuable! But most users will not see this either, until it makes it into a release. 

Should we cut a quick 1.4.1 to propagate these changes out? @skoudoro : would that be terribly onerous?

skoudoro:
> would that be terribly onerous?

Since we did not do any announcement concerning the release of 1.4.0 because of the workshop, we can do that. It will be easier. 

Let's cut this release 1-2 weeks from now, it will let us time to finalize some small issues

arokem:
Super!

PR #2372, opened 2021-04-20
Expand patch radius if input is int

FelixLiu-SF:
if patch radius is an int, it needs to be expanded to (1,3) array before calling _extract_3d_patches.

### DIFF
--- a/dipy/denoise/patch2self.py
+++ b/dipy/denoise/patch2self.py
@@ -236,7 +236,7 @@ def patch2self(data, bvals, patch_radius=[0, 0, 0], model='ridge',
     """
     if isinstance(patch_radius, int):
         patch_radius = np.ones(3, dtype=int) * patch_radius
-        
+
     patch_radius = np.asarray(patch_radius, dtype=int)
 
     if not data.ndim == 4:
--- a/dipy/denoise/patch2self.py
+++ b/dipy/denoise/patch2self.py
@@ -234,6 +234,9 @@ def patch2self(data, bvals, patch_radius=[0, 0, 0], model='ridge',
                     Denoising Diffusion MRI with Self-supervised Learning,
                     Advances in Neural Information Processing Systems 33 (2020)
     """
+    if isinstance(patch_radius, int):
+        patch_radius = np.ones(3, dtype=int) * patch_radius
+        
     patch_radius = np.asarray(patch_radius, dtype=int)
 
     if not data.ndim == 4:
--- a/dipy/denoise/tests/test_patch2self.py
+++ b/dipy/denoise/tests/test_patch2self.py
@@ -31,8 +31,8 @@ def test_patch2self_random_noise():
     assert_greater(S0den_clip.min(), S0.min())
     assert_equal(np.round(S0den_clip.mean()), 30)
 
-    # both clip and shift = True
-    S0den_clip = p2s.patch2self(S0, bvals, model='ols',
+    # both clip and shift = True, and int patch_radius
+    S0den_clip = p2s.patch2self(S0, bvals, patch_radius=0, model='ols',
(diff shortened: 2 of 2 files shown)


skoudoro (commented):
Hi @FelixLiu-SF, 

Thank you for doing this fix! 

- Can you update update the unit test also? 
- Can you fix [the pep8 above](https://github.com/dipy/dipy/pull/2372#issuecomment-823523794)?

@ShreyasFadnavis, Can you look into this PR? Thank you

dipy/denoise/patch2self.py
```diff
@@ -234,6 +234,9 @@ def patch2self(data, bvals, patch_radius=[0, 0, 0], model='ridge',
                     Denoising Diffusion MRI with Self-supervised Learning,
                     Advances in Neural Information Processing Systems 33 (2020)
     """
+    if isinstance(patch_radius, int):
+        patch_radius = np.ones(3, dtype=int) * patch_radius
+        
     patch_radius = np.asarray(patch_radius, dtype=int)
```
ShreyasFadnavis, line 240: I think this line becomes redundant as the patch_radius would already be a `numpy array`.

ShreyasFadnavis (commented):
Hi @FelixLiu-SF - Good catch! And thank you for this PR 👍🏽 

1 minor comment from me, otherwise I think the PR looks good! 

@skoudoro I don't think there is a unit test for this functionality.

ShreyasFadnavis:
Maybe hard to unit-test this since it's not a function of its own. We can make a function for it if necessary!

skoudoro:
> https://github.com/dipy/dipy/pull/2372#issuecomment-823523794

In your unit tests, you call many times `patch2self`with the default `patch_radius`.  You just need to pick one of them and add `patch_radius=0` instead of `[0, 0, 0]`.  It should be enough, we should get the same result and the int is tested

ShreyasFadnavis:
> > Maybe hard to unit-test this since it's not a function of its own. We can make a function for it if necessary!
> 
> In your unit tests, you call many times `patch2self`with the default `patch_radius`. You just need to pick one of them and add `patch_radius=0` instead of `[0, 0, 0]`. It should be enough, we should get the same result and the int is tested

Yep -- Good idea! Works :)

dipy/denoise/patch2self.py
FelixLiu-SF: I think it still needs to be there in case the input parameter is an array, such as the default input.

ShreyasFadnavis: Yes! You are right -- my bad!

skoudoro (commented):
Hi @FelixLiu-SF,

Thank you for your pep8 fix. Can you update a test and then your PR will be ready to be merged.

You just need to update [this line](https://github.com/dipy/dipy/blob/master/dipy/denoise/tests/test_patch2self.py#L35) and add `patch_radius=0` as a parameter.

Thank you.

FelixLiu-SF:
> Hi @FelixLiu-SF,
> 
> Thank you for your pep8 fix. Can you update a test and then your PR will be ready to be merged.
> 
> You just need to update [this line](https://github.com/dipy/dipy/blob/master/dipy/denoise/tests/test_patch2self.py#L35) and add `patch_radius=0` as a parameter.
> 
> Thank you.

All set!

ShreyasFadnavis:
Thanks for this @FelixLiu-SF ! I think its ready to be merged 👍🏽 

I will wait till tomorrow if anyone else has any other comments.
Thanks for the review @skoudoro!

skoudoro (approved):

ShreyasFadnavis:
Okay! This is good to go! Merging 🚀

Thanks @FelixLiu-SF !

PR #2382, opened 2021-05-01
Updated docs for using P2S optimally

ShreyasFadnavis:
- [x] Added information about parameters of Patch2Self
- [x] Also added information on applying Patch2Self on batches

### DIFF
--- a/doc/examples/denoise_patch2self.py
+++ b/doc/examples/denoise_patch2self.py
@@ -77,8 +77,8 @@ The ordinary least squares regression (model='ols') tends to be a little slower
 depending on the size of the data. In that case, please consider switching to
 ridge regression (model='ridge').
 
-*Please do note that sometimes using ridge regression can hamper the
-performance of Patch2Self. If so, please use model='ols'.*
+Please do note that sometimes using ridge regression can hamper the
+performance of Patch2Self. If so, please use model='ols'.
 
 The array `denoised_arr` contains the denoised output obtained from Patch2Self.
 
--- a/doc/examples/denoise_patch2self.py
+++ b/doc/examples/denoise_patch2self.py
@@ -158,7 +158,7 @@ After doing this, the 2 denoised batches can be merged as follows:
 `denoised_p2s = np.concatenate((denoised_batch1, denoised_batch2), axis=3)`
 
 One can also consider using the above batching approach to denoise each
-gradient direction separately.
+shell separately if working with multi-shell data.
 """
 
 """
--- a/doc/examples/denoise_patch2self.py
+++ b/doc/examples/denoise_patch2self.py
@@ -91,7 +91,7 @@ Please set `shift_intensity=True` and `clip_negative_vals=False` by default to
 avoid negative values in the denoised output.
 
 The `b0_threshold` is used to separate the b0 volumes from the DWI volumes.
-Changing the value of the b0 threshold is needed is the b0 volumes in the
+Changing the value of the b0 threshold is needed if the b0 volumes in the
 `bval` file have a value greater than the default `b0_threshold`.
 
 The default value of `b0_threshold` in DIPY is set to 50. If using data
--- a/doc/examples/denoise_patch2self.py
+++ b/doc/examples/denoise_patch2self.py
@@ -96,7 +96,7 @@ Changing the value of the b0 threshold is needed is the b0 volumes in the
 
 The default value of `b0_threshold` in DIPY is set to 50. If using data
 such as HCP 7T, the b0 volumes tend to have a higher b-value (>=50)
-assoiciated with them in the `bval` file. Please check the b-values for b0s and
+associated with them in the `bval` file. Please check the b-values for b0s and
 adjust the `b0_threshold` accordingly.
 
 Now lets visualize the output and the residuals obtained from the denoising.
--- a/doc/examples/denoise_patch2self.py
+++ b/doc/examples/denoise_patch2self.py
@@ -69,16 +69,16 @@ hardi_fname, hardi_bval_fname, hardi_bvec_fname = get_fnames('stanford_hardi')
 data, affine = load_nifti(hardi_fname)
 bvals = np.loadtxt(hardi_bval_fname)
 denoised_arr = patch2self(data, bvals, model='ols', shift_intensity=True,
-                          clip_negative_vals=False)
+                          clip_negative_vals=False, b0_threshold=100)
 
 """
 The above parameters should give optimal denoising performance for Patch2Self.
 The ordinary least squares regression (model='ols') tends to be a little slower
-given the size of the data. In that case, please consider switching to ridge
-regression (model='ridge').
+depending on the size of the data. In that case, please consider switching to
+ridge regression (model='ridge').
 
-Please do note that sometimes using ridge regression can hamper the performance
-of the Patch2Self. In such a case, please use model='ols'.
+*Please do note that sometimes using ridge regression can hamper the
+performance of Patch2Self. If so, please use model='ols'.*
 
 The array `denoised_arr` contains the denoised output obtained from Patch2Self.
 
@@ -89,6 +89,10 @@ by using the option `b0_denoising=False` within Patch2Self.
 
 Please set `shift_intensity=True` and `clip_negative_vals=False` by default to
 avoid negative values in the denoised output.
+
+The `b0_threshold` is used to separate the b0 volumes from the DWI volumes.
+Changing the value of the b0 threshold is needed is the b0 volumes in the
+`bval` file have a value greater than the used `b0_threshold`.
 
 Now lets visualize the output and the residuals obtained from the denoising.
 """
@@ -146,7 +150,7 @@ get the same denoising performance. One can simply run Patch2Self using:
 `denoised_batch1 = patch2self(data[..., :150], bvals[:150])`
 `denoised_batch2 = patch2self(data[..., 150:], bvals[150:])`
 
-After doing the 2 denoised batches can be merged as follows:
+After doing this, the 2 denoised batches can be merged as follows:
 `denoised_p2s = np.concatenate((denoised_batch1, denoised_batch2), axis=3)`
 
 If using Patch2Self for research purposes, acquisition design tests, etc., one
--- a/doc/examples/denoise_patch2self.py
+++ b/doc/examples/denoise_patch2self.py
@@ -45,9 +45,9 @@ This is done by using the self-supervised loss:
 fed into the regressor :math:`\Phi` built in phase (A). The prediction is a
 denoised version of held-out volume.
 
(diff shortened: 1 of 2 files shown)


ShreyasFadnavis:
Hi! Sorry for the slowness in this small PR. I have added information about how to use Patch2Self optimally. It is ready to be reviewed 👍🏽

doc/examples/denoise_patch2self.py
```diff
@@ -68,10 +68,18 @@
...
+                          clip_negative_vals=False, b0_threshold=100)
 
 """
+The above parameters should give optimal denoising performance for Patch2Self.
+The ordinary least squares regression (model='ols') tends to be a little slower
+depending on the size of the data. In that case, please consider switching to
+ridge regression (model='ridge').
+
+*Please do note that sometimes using ridge regression can hamper the
+performance of Patch2Self. If so, please use model='ols'.*
```
arokem, line 81:
```suggestion
performance of Patch2Self. If so, please use model='ols'.
```

doc/examples/denoise_patch2self.py
```diff
@@ -130,6 +142,22 @@
...
+directions is very high (>=200 volumes). For instance, if the data has 300
+volumes, one can split the data into 2 batches, (150 directions each) and still
+get the same denoising performance. One can simply run Patch2Self using:
+`denoised_batch1 = patch2self(data[..., :150], bvals[:150])`
+`denoised_batch2 = patch2self(data[..., 150:], bvals[150:])`
+
+After doing this, the 2 denoised batches can be merged as follows:
+`denoised_p2s = np.concatenate((denoised_batch1, denoised_batch2), axis=3)`
+
+If using Patch2Self for research purposes, acquisition design tests, etc., one
```
arokem, line 156:
```suggestion
If using Patch2Self for research purposes, acquisition design tests, etc., one
```
I am not sure what the "research purposes" means here. The following "acquisition design, tests, etc." suggests that this is not something that users should do if they want to analyze the data? But I am not sure.

doc/examples/denoise_patch2self.py
```diff
@@ -68,10 +68,18 @@
 hardi_fname, hardi_bval_fname, hardi_bvec_fname = get_fnames('stanford_hardi')
 data, affine = load_nifti(hardi_fname)
 bvals = np.loadtxt(hardi_bval_fname)
-denoised_arr = patch2self(data, bvals, shift_intensity=True,
-                          clip_negative_vals=False)
+denoised_arr = patch2self(data, bvals, model='ols', shift_intensity=True,
+                          clip_negative_vals=False, b0_threshold=100)
```
arokem, line 72: Sorry -- one more: why is `b0_threshold` set to a higher value than the `gradient_table` default?

ShreyasFadnavis: Okay... Set it back to 50, which is the default (even in the codebase). I added a note that people should change it if they are working with data such as HCP 👍🏽

doc/examples/denoise_patch2self.py
ShreyasFadnavis: Yes! Agreed... Taking that part of the line off. I just wanted to encourage people to experiment with the algorithm.

doc/examples/denoise_patch2self.py
```diff
@@ -82,6 +90,14 @@
 Please set `shift_intensity=True` and `clip_negative_vals=False` by default to
 avoid negative values in the denoised output.
 
+The `b0_threshold` is used to separate the b0 volumes from the DWI volumes.
+Changing the value of the b0 threshold is needed is the b0 volumes in the
+`bval` file have a value greater than the default `b0_threshold`.
+
+The default value of `b0_threshold` in DIPY is set to 50. If using data
+such as HCP, the b0 volume tends to have a higher value, please set it to 100
+or more accordingly.
```
arokem, line 99: Hmmm. The b-values that were used in the HCP 3T dataset are:  

```
array([   5.,  990.,  995., 1000., 1005., 1985., 1990., 1995., 2000.,
       2005., 2010., 2980., 2985., 2990., 2995., 3000., 3005., 3010.])
```

So, setting `b0_threshold=50` and setting `b0_threshold=100` would presumably give you exactly the same thing. I am also confused about the phrase `the b0 volume tends to have a higher value` -- it doesn't seem like this parameter would interact with the values of the image, only with the selection of the image volumes. Or am I missing something?

ShreyasFadnavis: Hi @arokem -- That is correct. If the bvals are less than 50 the `b0_threshold` at 50 and 100 would behave exactly the same way. Here are the bvals for HCP 7T for instance:
image

As you can see they are >=50 somewhere between 55 and 65. Although not common, I do want to warn users if they have such datasets. Let me try rephrasing the sentence to make it clearer.

doc/examples/denoise_patch2self.py
```diff
@@ -68,10 +68,18 @@
 hardi_fname, hardi_bval_fname, hardi_bvec_fname = get_fnames('stanford_hardi')
 data, affine = load_nifti(hardi_fname)
 bvals = np.loadtxt(hardi_bval_fname)
-denoised_arr = patch2self(data, bvals, shift_intensity=True,
-                          clip_negative_vals=False)
+denoised_arr = patch2self(data, bvals, model='ols', shift_intensity=True,
+                          clip_negative_vals=False, b0_denoising=50)
```
arokem, line 72:
```suggestion
                          clip_negative_vals=False, b0_threshold=50)
```
More here: I believe that `b0_denoising` is a boolean.

arokem: Or possibly
```suggestion
                          clip_negative_vals=False, b0_denoising=True)
```

But both are the default values, so it's not clear why you are setting either of them.

doc/examples/denoise_patch2self.py
```diff
@@ -68,10 +68,18 @@
...
 
 """
+The above parameters should give optimal denoising performance for Patch2Self.
```
arokem, line 75: Optimal in this dataset, or generally? If generally, why are these not the default values? For example, the warning below about ridge hampering performance is rather ominous. Should we change the default to `'ols'`?

doc/examples/denoise_patch2self.py
```diff
@@ -82,6 +90,15 @@
 Please set `shift_intensity=True` and `clip_negative_vals=False` by default to
 avoid negative values in the denoised output.
 
+The `b0_threshold` is used to separate the b0 volumes from the DWI volumes.
+Changing the value of the b0 threshold is needed is the b0 volumes in the
```
arokem, line 94:
```suggestion
Changing the value of the b0 threshold is needed if the b0 volumes in the
```

doc/examples/denoise_patch2self.py
```diff
@@ -82,6 +90,15 @@
...
+`bval` file have a value greater than the default `b0_threshold`.
+
+The default value of `b0_threshold` in DIPY is set to 50. If using data
+such as HCP 7T, the b0 volumes tend to have a higher b-value (>=50)
+assoiciated with them in the `bval` file. Please check the b-values for b0s and
```
arokem, line 99:
```suggestion
associated with them in the `bval` file. Please check the b-values for b0s and
```

doc/examples/denoise_patch2self.py
ShreyasFadnavis: My bad! Good catch -- fixing this.

doc/examples/denoise_patch2self.py
ShreyasFadnavis: Done!

arokem: Optimal in this dataset, or generally?

ShreyasFadnavis: 'ols' is optimal generally. Same goes for `shift_intensity=True` and `clip_negative_vals=False` 👍🏽 Ridge does seem to be useful to speedup the regression. I was also looking into using the regression module of `sklearn` and found this comment from the discussion very useful: https://github.com/scikit-learn/scikit-learn/issues/13923#issuecomment-497077229. 

Should I use this? I was thinking of adding an option called `ols_fast` instead of just replacing `ridge`  with a very small `alpha`. What do you think?

ShreyasFadnavis:
While we are at it: should I also add `Patch2Self` in every tutorial as a primary step? I remember @arokem suggesting this in a previous PR 👍🏽 P2S is fast... adds ~3-5 mins for Stanford HARDI data per example. Is this okay? Or will it drive up the compilation time of all tutorials drastically?

skoudoro:
70 tutorials x 5min = 350min =~  6h more....  so I recommend to pick up not more than 10 tutorials.

doc/examples/denoise_patch2self.py
ShreyasFadnavis: > 70 tutorials x 5min = 350min =~ 6h more.... so I recommend to pick up not more than 10 tutorials.

Yes! I don't think it's relevant for all tutorials either :)

Here are the ones where I would add it:
1. CSD and MCSD
2. 3D SHORE
3. MAP-MRI
4. DTI
5. DKI
6. FW-DTI
7. WMTI
8. Basic Tracking EuDX
9. Deterministic Tracking
10. Probabilistic Tracking 
11. Particle Filter Tracking

Does this sound okay?

ShreyasFadnavis:
> 70 tutorials x 5min = 350min =~ 6h more.... so I recommend to pick up not more than 10 tutorials.

Yes! I don't think it's relevant for all tutorials either :)

Here are the ones where I would add it:
1. CSD and MCSD
2. 3D SHORE
3. MAP-MRI
4. DTI
5. DKI
6. FW-DTI
7. WMTI
8. Basic Tracking EuDX
9. Deterministic Tracking
10. Probabilistic Tracking 
11. Particle Filter Tracking

Does this sound okay?

skoudoro:
> Does this sound okay?

+1 👍🏾

skoudoro:
@arokem, Do you want to have a last look? What is the final decision concerning this PR?

doc/examples/denoise_patch2self.py
arokem: I think that it's very odd that ridge is faster than OLS, given that they essentially are the same algorithm with a small tweak that shouldn't really affect performance at all. But 🤷‍♂️

arokem:
I would only change those tutorials that already have denoising in them, but I think that we can take that on another PR (or even better, series of PRs). I think that this PR is ready to go.

doc/examples/denoise_patch2self.py
ShreyasFadnavis: Yep... me too!😂

skoudoro:
Thank you @ShreyasFadnavis, thank you for the review @arokem

PR #2389, opened 2021-05-07
[Fix] bundles_distances_mdf asymmetric values

### LINKED ISSUE #2310: `bundles_distances_mdf` asymmetric values
## Description
Using `bundles_distances_mdf` to calculate minimum average direct-flip distances between streamlines on subjects from HCP dataset results in asymmetric values. Expecting distance to be symmetric: `MDF(i,j)=MDF(j,i)`

```
>>> import numpy as np
>>> from dipy.io.streamline import load_tractogram
>>> import dipy.tracking.streamline as dts
>>> tractogram = load_tractogram(tg_fname, 'same')
>>> streamlines = tractogram.streamlines
>>> fgarray = np.array(dts.set_number_of_points(streamlines, 100))
>>> mdf = dts.bundles_distances_mdf(fgarray, fgarray)
>>> np.count_nonzero(mdf-mdf.T)
2930
```

Asymmetries are small, so this may be a result of numerical error:

```
>>> np.sum(abs(mdf-mdf.T))
0.014194488525390625
```
### Code example
```
>>> import numpy as np
>>> import dipy.tracking.streamline as dts
>>> A = np.load('max_asymmetric_mdf.npy')
>>> dts.bundles_distances_mdf(A,A)
array([[ 0.        , 27.01378632],
       [27.01377106,  0.        ]])
```
[max_asymmetric_mdf.npy.zip](https://github.com/dipy/dipy/files/5777396/max_asymmetric_mdf.npy.zip)

### Relevant images
image

### Operating system and version
```
macOS-10.16-x86_64-i386-64bit
```

### Python version
```
Python 3.8.5 (default, Sep  4 2020, 02:22:02) 
[Clang 10.0.0 ]
```
### dipy version
```
dipy 1.2.0
```

### dependency version
```
NumPy 1.17.5
```

skoudoro:
The goal of this PR is to fix #2310.  `bundles_distances_mdf` had asymmetric values due to some rounding effect. 

Now we have:
```python
>>> import numpy as np
>>> import dipy.tracking.streamline as dts
>>> A = np.load('max_asymmetric_mdf.npy')
>>> dts.bundles_distances_mdf(A,A)
array([[ 0.        , 27.01377869],
       [27.01377869,  0.        ]])
```

Can you try and confirm @bloomdt-uw?  Thank you

### DIFF
--- a/dipy/tracking/distances.pyx
+++ b/dipy/tracking/distances.pyx
@@ -1480,8 +1480,10 @@ cdef void track_direct_flip_dist(float *a,float *b,long rows,float *out) nogil:
     dipy.tracking.distances.local_skeleton_clustering
     """
     cdef:
-        long i=0,j=0
-        float sub=0,subf=0,distf=0,dist=0,tmprow=0, tmprowf=0
+        cnp.npy_intp i=0
+        cnp.npy_intp j=0
+        cnp.float32_t sub=0,subf=0, tmprow=0, tmprowf=0
+        double distf=0,dist=0
 
     for i from 0<=i<rows:
         tmprow=0
@@ -1494,8 +1496,8 @@ cdef void track_direct_flip_dist(float *a,float *b,long rows,float *out) nogil:
         dist+=sqrt(tmprow)
         distf+=sqrt(tmprowf)
 
-    out[0]=dist/<float>rows
-    out[1]=distf/<float>rows
+    out[0]=<cnp.float32_t>dist/<cnp.float32_t>rows
+    out[1]=<cnp.float32_t>distf/<cnp.float32_t>rows
 
 
 @cython.cdivision(True)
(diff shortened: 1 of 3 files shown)


skoudoro:
> Looks good.  A data type issue?

yes a datatype issue

> Will we need to switch all our cython code from float to cnp.float32_t Or double to cnp.float64_t ?

No, it is ok to keep it as it is.  I just did it to be more explicit

> And did the problem appear in all operating systems?

yes,

> But what was the problem?

The problem was coming from `dist` and `distf`. They had to be `double` and not `float` because of a rounding effect following the order of your operation. here a basic example with int (equivalent to float)
```python
int(int(1.2 +3.9)+ 3.2) = 8
int(int(1.2 +3.2)+ 3.9) = 7
```

Garyfallidis:
Thank you @skoudoro! And @bloomdt-uw for reporting!

Issue #2276, opened 2020-10-30
Update Align workflow + create transformation tutorial


skoudoro:
This issue is a follow up of a discussion with @grlee77 in #2223. 3 PRs needs to be created 

> 1.) Perhaps provide a new example based on modification of the existing affine_registration_3d.py? One could choose the b0 volume and then introduce a second copy after a small rotation, translation and/or scaling (possibly via scipy.ndimage's rotate/shift/zoom). The proposed transforms should then be able to co-register these. (I suppose this could also be used as an integration test if you wanted to validate the accuracy of the estimated translations/rotations to some precision).

> 2.) consider adding additional methods for these to workflows/align.py (can always be done as a follow-up PR too)

> The test cases proposed here do appear to verify that the classes produce the expected transformation matrix for a given set of parameters, but to increase confidence, it would be nice to demonstrate that they operate as expected when used in an actual registration example.

drombas:
Hi @skoudoro!

May I try the second point? It would involve just adding options for RigidScaling and RigidIsoScaling to the workflow, right?

skoudoro:
Hi @drombas,

yes, you can go ahead! 

Thank you

PR #2390, opened 2021-05-09
NF: extend the align workflow with Rigid+IsoScaling and Rigid+Scaling

drombas:
Part of #2276 

### Proposed changes
- Add RigidIsoScaling and RigidScaling options to the align workflow
- Update corresponding tests

By the way, I've noticed that the registration parameters defined [here](https://github.com/dipy/dipy/pull/2390/files#diff-b6bf9b9a51f8dfc421e3896b6ea103012d81d9eae8b07e252e0f5a66a220f42fL204) are not used as the [setup_random_transform ](https://github.com/dipy/dipy/blob/master/dipy/align/tests/test_parzenhist.py#L259)function calculates them randomly.

### DIFF
--- a/dipy/align/_public.py
+++ b/dipy/align/_public.py
@@ -36,8 +36,6 @@ from dipy.io.image import load_nifti, save_nifti
 
 __all__ = ["syn_registration", "register_dwi_to_template",
            "write_mapping", "read_mapping", "resample",
-           "center_of_mass", "translation", "rigid", "rigid_isoscaling",
-           "rigid_scaling", "affine",
            "affine_registration", "register_series",
            "register_dwi_series", "streamline_registration"]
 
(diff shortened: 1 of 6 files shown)


drombas:
The major changes I made:
- Adapt `affine_registration()` method from `dipy.align._public` to support new registration methods and workflow logic.
- Add `ret_metric` argument where necessary to return also the optimization coefficients (`xopt`) and metric (`fopt`)

skoudoro:
Apologize for getting back late to you @drombas. I will test and look at your PR this week. but for now, I don't see something bugging me.

skoudoro (commented):
Hi @drombas,

Overall, it looks good. I feel there is a lot of duplicated code in `dipy/align/_public.py` where only the transformation change. We might need to simplify this. See below some comments

skoudoro:
Also, Can you rebase this PR? thank you

drombas:
Hi @skoudoro!

As you said there was a lot of repeated code in `align._public.py`, so I moved all single registration methods to `align._public.affine_registration()` function.

dipy/align/_public.py
```diff
@@ -631,13 +448,36 @@ def affine_registration(moving, static,
...
+            elif func == "rigid":
+                transform = RigidTransform3D()
+            elif func == "rigid_isoscaling":
+                transform = RigidIsoScalingTransform3D()
+            elif func == "rigid_scaling":
+                transform = RigidScalingTransform3D()
+            elif func == "affine":
+                transform = AffineTransform3D()
+            else:
+                raise ValueError("Not supported registration method")
```
skoudoro, line 473: can you add a test for that? (`npt.assert_raises(ValueError, ...)`)

skoudoro (commented):
Hi @drombas, 

Thank you for this, this is much cleaner! I like it and this is almost ready to go.

We should keep `center_of_mass, translation, rigid, affine`  function for backward compatibility and because this is also useful also.

To avoid duplication, I recommend you to code something like:
```python
from functools import partial

center_of_mass  = partial(affine_registration, pipeline=['center_of_mass'])
translation.__doc__ = """ blabla... """
translation = partial(affine_registration, pipeline=['translation'])
translation.__doc__ = """ specific text... """
rigid = partial(affine_registration, pipeline=['rigid'])
rigid.__doc__ = """ specific text... """
affine = partial(affine_registration, pipeline=['affine'])
affine.__doc__ = """ specific... """
```

dipy/align/_public.py
drombas: done!

drombas:
Thanks @skoudoro for the feedback! I tried to add what you suggested. 

Just a comment: since now `center_of_mass, translation, rigid, affine` are based on `affine_registration()`, they do not behave exactly as before (i.e. they now return the resampled image too and do not take any `AffineRegistration` instance as input). 
Not sure how this might impact backward compatibility. Anyway it looks like those single transforms are not used anywhere else across de codebase so I guess we are safe.

skoudoro (approved):
@arokem, is it fine if we break the backward compatibilities with `center_of_mass, translation, rigid, affine` on `dipy.align._public` module.  This is quite new and the @drombas version simplifies the general maintenance and the compatibility with our workflows.

I am +1 for this. Please let us know what you think.

arokem:
Yeah. I think it’s fine.

skoudoro (commented):
All good, I am just waiting for the CI's to be green, and I will go ahead and merge this PR

skoudoro:
Thank you @drombas for the good work

PR #2392, opened 2021-05-14
Add decomposition

weningerleon:
Is there a reason why a 3x3 matrix is assumed to be a rotation-only matrix, but not the 4x4 matrix?
Adding polar decomposition also when the matrix is present in 3x3 form.

### DIFF
--- a/dipy/core/tests/test_gradients.py
+++ b/dipy/core/tests/test_gradients.py
@@ -424,8 +424,7 @@ def test_reorient_bvecs():
     # so that the reorient_bvecs function does not throw an error itself
     new_gt = reorient_bvecs(gt, np.array(shear_affines)[:, :3, :3], atol=1)
     bvecs_close_to_1 = abs(vector_norm(new_gt.bvecs[~gt.b0s_mask]) - 1) <= 0.001
-    npt.assert_equal(bvecs_close_to_1, np.ones(bvecs_close_to_1.shape,
-                                               dtype=bool))
+    npt.assert_(np.all(bvecs_close_to_1)) 
 
 def test_nan_bvecs():
     """
--- a/dipy/core/tests/test_gradients.py
+++ b/dipy/core/tests/test_gradients.py
@@ -14,7 +14,7 @@ from dipy.core.gradients import (gradient_table, GradientTable,
                                  unique_bvals_magnitude,
                                  unique_bvals_tolerance, unique_bvals,
                                  params_to_btens, btens_to_params)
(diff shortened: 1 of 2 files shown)


arokem:
Good catch! I think that you are right. Would you mind adding a test case that fails without this fix and passes with this fix?

skoudoro (commented):
Hi @weningerleon,

It looks good to me too. it would be good if you could update the tests. keep us updated if you need some help.

Thanks!

weningerleon:
Hi, 
Sorry for the late answer, I added a test!

skoudoro:
The failing CI's are not related to your change. It looks good to me. 

Can you have a look @arokem? You can go ahead and merge this PR if everything is fine with you

dipy/core/tests/test_gradients.py
```diff
@@ -412,6 +412,20 @@ def test_reorient_bvecs():
...
+        shear_affines.append(np.array([[1, 0, 1, 0],
+                                       [0, 1, 0, 0],
+                                       [0, 0, 1, 0],
+                                       [0, 0, 0, 1]]))
+    # atol is set to 1 here to do the scaling verification here,
+    # so that the reorient_bvecs function does not throw an error itself
+    new_gt = reorient_bvecs(gt, np.array(shear_affines)[:, :3, :3], atol=1)
+    bvecs_close_to_1 = abs(vector_norm(new_gt.bvecs[~gt.b0s_mask]) - 1) <= 0.001
+    npt.assert_equal(bvecs_close_to_1, np.ones(bvecs_close_to_1.shape,
+                                               dtype=bool))
```
arokem, line 428: I think that this is slightly clearer:

```suggestion
    npt.assert_(np.all(bvecs_close_to_1)) 
```

arokem (commented):
Sorry - I missed the notification of the commit with the test. Looks great! I had a small suggestion for a change.

dipy/core/tests/test_gradients.py
arokem: I think that once we have someone take a look and verify that I did not mess this up, I think that we can merge this. Does this proposed change look OK to you, @skoudoro? In case of doubt, we can also go ahead and merge this without my suggestion.

skoudoro: I agree with the suggestion and just commit it.  

When the CI's will be done, I will go ahead and merge it.

Thank you @arokem!

skoudoro:
All green, Thank you @weningerleon for this fix, and thank you @arokem for the review. merging!

PR #2405, opened 2021-06-01
Small correction on KFA 

RafaelNH:
This should be a small correction on KFA definition - it has to be zero if the kurtosis tensor is null.

### DIFF
--- a/dipy/reconst/dki.py
+++ b/dipy/reconst/dki.py
@@ -1160,9 +1160,6 @@ def axial_kurtosis(dki_params, min_kurtosis=-3./7, max_kurtosis=10,
     return AK.reshape(outshape)
 
 
-
-
-
 def _kt_maximum_converge(ang, dt, md, kt):
     """ Helper function that computes the inverse of the directional kurtosis
     of a voxel along a given direction in polar coordinates.
--- a/dipy/reconst/dki.py
+++ b/dipy/reconst/dki.py
@@ -1160,6 +1160,9 @@ def axial_kurtosis(dki_params, min_kurtosis=-3./7, max_kurtosis=10,
     return AK.reshape(outshape)
 
 
+
+
+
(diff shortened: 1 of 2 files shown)


skoudoro (commented):
Hi @RafaelNH,

Thank you for the update. it seems your test is failing on our CI's:

```python 
>       assert_almost_equal(dkiF.kfa, 0)
E       AssertionError: 
E       Arrays are not almost equal to 7 decimals
E       
E       Mismatched elements: 1 / 1 (100%)
E       Max absolute difference: 0.8582412541188063
E       Max relative difference: inf
E        x: array(0.8582412541188063)
E        y: array(0)
```

more information [here](https://dev.azure.com/dipy/dipy/_build/results?buildId=1163&view=logs&j=23a3abda-4c03-56dd-bb0c-e147b7c0d92b&t=96c255f3-a894-5a77-4222-55009645cdc9&l=2991)

Can you have a quick look? Thanks!

skoudoro:
Hi @RafaelNH,

this a just a small ping concerning my question/comment above.

RafaelNH:
Hi @skoudoro! Thanks for pointing that the test was failing! At least, it was ensuring that the added test detected that issue. This test failure was related to KFA estimates for null kurtosis zeros (in which MKT = 0). MKT was supposed to be zero on the added test, but due to numerical precision is was having a small magnitude. It didn't fail on my pc, because my pc computed a negative residual MKT values, while in the CI tests MKT was giving a small positive residual value. To fix the issue, I am now using a small tolerance to detect null kurtosis tensors (i.e. MKT < tol).

arokem:
Looks like the remaining failures are the ones that will be fixed through #2407, so I think this is good to go.

skoudoro:
I agree, good to go. 

Thank you @RafaelNH for this fix! merging

PR #2406, opened 2021-06-11
Manage Approx_polygon_track with repeated points

### LINKED ISSUE #2314: Approx_polygon_track with repeated points gives an error
## Description
Approx_polygon_track under tracking/distances seems to have an issue with repeated points in a track. 
If the track is the numpy vector [[1,0,0],[2,0,0],[3,0,0],[4,0,0]] the output is, correctly, [[1,0,0],[4,0,0]] 
(I am not adding floating points for readability)

If the track though has a repeated point this happens:

>>> track=np.array([[1.,0.,0.],[1.,0.,0.],[3.,0.,0.],[4.,0.,0.]])
>>> approx_polygon_track(track)
  File "dipy/tracking/distances.pyx", line 1123, in dipy.tracking.distances.approx_polygon_track
ZeroDivisionError: float division

It seems that the code takes in account this possibility through a dpy_isnan but it's not enough as there is no try?

## Way to reproduce
- Debian (Linux-4.19.0-12-amd64-x86_64-with-glibc2.10
- Python version 3.8.5
- [ ] dipy version 1.3.0
- [ ] dependency version (numpy 1.19.4, scipy 1.6.0, nibabel 3.2.1, h5py 3.1.0)

skoudoro:
This PR should fix #2314. 

It just avoids the division by zero when 2 consecutive repeated points appeared.

It should solve your problem @dulietina, can you test it? 

Thank you

### DIFF
--- a/dipy/tracking/distances.pyx
+++ b/dipy/tracking/distances.pyx
@@ -1115,7 +1115,7 @@ def approx_polygon_track(xyz,alpha=0.392):
         float *fvec2
         object characteristic_points
         cnp.npy_intp t_len
-        double angle,tmp
+        double angle,tmp, denom
         float vec0[3]
         float vec1[3]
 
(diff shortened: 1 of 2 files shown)


Garyfallidis:
All good on my end. Merging.

Issue #1687, opened 2018-12-11
Generate random affine


skoudoro:
## Description

As suggested on [this comment](https://github.com/nipy/dipy/pull/1684#issuecomment-445891112), it would be good to have a function for generating random affine matrix. 
This is just a reminder.

jhlegarreta:
@skoudoro Is this still relevant/useful? The proposed method looks straightforward to implement. Should we go for it, or else, should we control the linear and affine (dropping the projective part?) cases separately, like in [here](https://stackoverflow.com/questions/16682965/how-to-generaterandomtransform-with-opencv)?

skoudoro:
Yes @jhlegarreta, still relevant and you can go for it. This function will be really useful for our tests like registration module.

sidkapoor97:
@arokem @skoudoro is there any link which helps in testing whether a unit determinant matrix represents affine transformation or not? Also rotation and translation have determinant 1, whereas shear doesn't.

skoudoro:
close based on this comment: https://github.com/dipy/dipy/pull/1709#issuecomment-5732504140

PR #2407, opened 2021-06-13
from random to deterministic test for deform_streamlines

skoudoro:
The CI's are failing regularly due to the `test_deform_streamlines()` function. 

The reason: I suppose the random affine is sometimes irrealistic

So I converted the test to a deterministic one.


arokem:
This goes back to this issue: https://github.com/dipy/dipy/issues/1687, which would provide a more general solution.  I think this is a good stop-gap solution, though.

Would you mind adding a comment saying where these numbers come from? Are they just a randomly-generated case that happens to work here?

skoudoro:
Yes, I confirm, They are just randomly generated. I took the first one working

skoudoro:
Thank you, and I agree that it would be good if we fix #1687

PR #2409, opened 2021-06-22
NF - Sample PMF for an input position and direction

gabknight:
This PR adds a utility function to the `SHCoeffPmfGen` object. 

- It adds the function `get_pmf_val(point, xyz)`, which return the pmf (ODF/FOD) value at the position `point` and orientation `xyz`. 
- The closest vertex of the sphere is used to approximate the value at `xyz`.

In the current implementation, the pmf is computed for all vertices, to return the value for one vertex. This could be further optimized. 

@guaje @skoudoro For the PTT implementation, I think we need to move this function to the parent class `PmfGen`. but it doesn't work with child `class SimplePmfGen(PmfGen)`, since this class doesn't have a sphere object. I think there is a bit of a design flaw, and simply moving/requiring the sphere for `PmfGen` (as for `SHCoeffPmfGen`) might solve this. I can add this commit to clarify what I mean.

### DIFF
--- a/dipy/direction/pmf.pyx
+++ b/dipy/direction/pmf.pyx
@@ -50,7 +50,7 @@ cdef class SimplePmfGen(PmfGen):
             raise ValueError("pmf should not have negative values.")
         if not pmf_array.shape[3] == sphere.vertices.shape[0]:
             raise ValueError("pmf should have the same number of values as the"
-                             + "number of vertices on sphere.")
+                             + " number of vertices of sphere.")
 
     cpdef double[:] get_pmf(self, double[::1] point):
         if trilinear_interpolate4d_c(self.data, &point[0], self.pmf) != 0:
@@ -123,7 +123,7 @@ cdef class BootPmfGen(PmfGen):
         r, theta, phi = shm.cart2sphere(x, y, z)
         b_range = (r.max() - r.min()) / r.min()
         if b_range > tol:
-            raise ValueError("BootPmfGen only supports single shell data")
+            raise ValueError("BootPmfGen only supports single shell data.")
         B, _, _ = shm.real_sh_descoteaux(self.sh_order, theta, phi)
         self.H = shm.hat(B)
         self.R = shm.lcr_matrix(self.H)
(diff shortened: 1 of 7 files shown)


skoudoro:
Your explanation is clear and +1 for moving/requiring the sphere for `PmfGen`. 

I would recommend something like:

```
cdef class PmfGen:

    def __init__(self,
                 double[:, :, :, :] data,
                 object sphere=None):
```

guaje:
Thanks @gabknight!

My only input here is why don't we use the `default_sphere` as a fixed parameter in the `__init__` of the PmfGen class?

dipy/direction/pmf.pyx
```diff
@@ -14,15 +15,21 @@ from dipy.core.interpolation cimport trilinear_interpolate4d_c
 cdef class PmfGen:
 
     def __init__(self,
-                 double[:, :, :, :] data):
-        self.data = np.asarray(data,  dtype=float)
+                 double[:, :, :, :] data,
+                 object sphere=default_sphere):
```
gabknight, line 19: I'm not comfortable with this. I would prefer having no default value for the sphere. The only sphere that makes sense to use here is the precise sphere corresponding to the spherical data point of the input `data`. Any other sphere will results in an erroneous interpretation of the pmfs. 

If having a default sphere value is necessary, I would advocate for using `None`, and throwing warnings in functions requiring the sphere (i.e. `get_pmf_value(.)`).

skoudoro, line 19: After a second thought, I agree with you, let's remove the `default_sphere`. It should be mandatory and not optional.

gabknight: Ok, (sorry I delete the previous comment as I making additional tests.). Ok, `SHCoeffPmfGen` can use any spheres as input, but the user must be careful with `SimplePmfGen` and setting a default sphere might lead to erroneous results hard to debug.

gabknight:
@skoudoro @guaje Thanks for the feedback. I had to make the `sphere` parameter mandatory in all `PmfGen` sub-classes, but the `sphere` parameter is only mandatory for `PmfGenDirectionGetter.from_pmf(.)`, Other constructors have `default_sphere`.

skoudoro (approved):
Thank you @gabknight, it looks good to me. 

I will wait for the CI's to be all green. does anyone want to have another look?

skoudoro:
Thank you @gabknight, merging!

Issue #1805, opened 2019-04-12
`c_compress_streamline` keeps identical points when it shouldn't


jchoude:
## Description
When trying to compress a streamline, if two or more consecutive points are identical, only the first should be kept, and the subsequent points removed. However, currently, `c_compress_streamline` keeps them. 

When computing the offset in `c_dist_to_line`, 2 identical points yield a `nan`. However, currently, there is a check on `nan`, and points are still added.

Does anyone know what other case could yield a `nan` there? Else, we could remove it or change it.

Tagging @MarcCote since he's the last one to change this part of the code.

FA is the ref anat, and an example track is https://www.dropbox.com/s/v2uxq5b0hywhbbh/853.trk?dl=0

[fa.nii.gz](https://github.com/nipy/dipy/files/3073633/fa.nii.gz)

I can fix it, but I wanted to validate the reasoning behind the check.

MarcCote:
If you have the following 2D streamline `[(0,1), (0,20), (0,1)]`, you don't want to remove the second point which is what would happen if we discard on `nan`.

I think the code should be removing two consecutive points that are identical like `[(0,1), (0,1), (0,20)]`, but won't work when you have 2+ identical consecutive points. Do you think that's your case? (I haven't look at your trk yet).

jchoude:
That's indeed my case. The first 11 points are identical.

I see the use for the nan now.

MarcCote:
But there's clearly something we can do in such case like skipping the point if `curr == prev`.

jchoude:
Yes, I was wondering if that would incur too much overhead, but such a simple check is still less time consuming than the norm computations. I'll work on a fix, thanks @MarcCote  :)

MarcCote:
Cool. Tag me, I'll be happy to review it.

skoudoro:
Hi @jchoude,

Do still plan to create a PR concerning this issue/New feature? Can you give us a feedback? 

Thank you!

PR #2411, opened 2021-06-23
[FIX] `c_compress_streamline` discard identical points

skoudoro:
This PR is to fix issue #1805.

The code is removing some points if they are consecutive and identical like [(0,1), (0,1), (0,20)]

However, If you have the following 2D streamline [(0,1), (0,20), (0,1)], we keep it as it is.

You can check the tests for more examples.

It would be good if @jchoude or @MarcCote  could review this PR since it addresses your discussion.

Thank you

### DIFF
--- a/dipy/tracking/streamlinespeed.pyx
+++ b/dipy/tracking/streamlinespeed.pyx
@@ -468,7 +468,7 @@ cdef double c_dist_to_line(Streamline streamline, cnp.npy_intp prev,
         norm2 += dn*dn
     norm2 = sqrt(norm2)
 
-    return norm1 / norm2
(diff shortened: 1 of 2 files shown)


MarcCote (approved):
The fix looks good to me. Any impact on the speed? It should be negligible, I guess.

skoudoro:
>  Any impact on the speed?

I confirm that it is negligible. I don't remember the exact number, so if needed I can re-run the performance test

MarcCote:
That's fine. I'll let you merge it.

Garyfallidis:
Merging. Nonetheless, this function should switch to use the Streamlines class no?

PR #2413, opened 2021-06-25
Azure pipeline: from ubuntu 1604 to 2004

skoudoro:
Ubuntu 1604 is deprecated on Azure Pipeline so this is just a quick update.


arokem:
Do you understand if the CI failures (e.g., https://dev.azure.com/dipy/dipy/_build/results?buildId=1191&view=logs&jobId=bb1c2637-64c6-57bd-9ea6-93823b2df951&j=bb1c2637-64c6-57bd-9ea6-93823b2df951&t=83ffc4db-f31c-557f-c048-6124f14897a8) are related to this change?

skoudoro:
> are related to this change?

I don't think so. After looking in detail the CI, here the 2 errors: 

- There are no wheels for python3.6 and the latest `h5py`. It was existing before. So the compilation is failing as you can [see here](https://dev.azure.com/dipy/dipy/_build/results?buildId=1191&view=logs&j=c64f2b93-e1ef-5004-dedf-97174a8ceb90&t=b7c4acca-b3bc-5bfc-836f-258060e11062&l=320). it seems that Cython needs to be installed before building this wheel. Another option is to ask h5py to build wheels for python3.6 (end of life in 4month).  We could also just remove the Python3.6 matrix due to its end of life. 

What is your advice concerning the above option? 

-  New VTK release !!! 9.0.2. However, it seems it makes Horizon crash with a memory leak. So we need to release FURY and check what is going on with Horizon. We might need to wait for a FURY release.

What do you think? We can wait until FURY release.

PR #2415, opened 2021-06-27
Implementation multi_voxel_fit progress bar

lcscosta:
Hello everyone,

I started using the dipy for analyze my IVIM images, for fit an entire image will take some time. Without a progress bar, I'll be in the dark.I used tqdm for implement a simple progress bar.

have a good day.

### DIFF
--- a/dipy/reconst/multi_voxel.py
+++ b/dipy/reconst/multi_voxel.py
@@ -1,6 +1,7 @@
 """Tools to easily make multi voxel models"""
 import numpy as np
 from numpy.lib.stride_tricks import as_strided
+from tqdm import tqdm
 
 from dipy.core.ndindex import ndindex
 from dipy.reconst.quick_squash import quick_squash as _squash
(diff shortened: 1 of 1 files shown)


dipy/reconst/multi_voxel.py
```diff
@@ -28,9 +29,13 @@ def new_fit(self, data, mask=None):
 
         # Fit data where mask is True
         fit_array = np.empty(data.shape[:-1], dtype=object)
+        MAXCOUNT = np.prod(data.shape[:-1])
+        bar = tqdm(total=MAXCOUNT, position=0)
         for ijk in ndindex(data.shape[:-1]):
             if mask[ijk]:
                 fit_array[ijk] = single_voxel_fit(self, data[ijk])
+            bar.update()
+        bar.close()
```
arokem, line 38: I would use the total number of voxels in the mask as the total, as I think that gives a more realistic view of the remaining compute to be executed. Something like this: 

```suggestion
        bar = tqdm(total=np.sum(mask), position=0)
        for ijk in ndindex(data.shape[:-1]):
            if mask[ijk]:
                fit_array[ijk] = single_voxel_fit(self, data[ijk])
                bar.update()
        bar.close()
```

arokem (commented):
First of all, thanks for the PR. I think this would indeed be helpful, and tqdm is already a dependency, so there's no downside. I do have a suggestion to refactor this (see comment).

dipy/reconst/multi_voxel.py
lcscosta: I'm agree with your suggestion, thank you for that.

skoudoro (approved):
Thank you @lcscosta!

As soon as the CI's come back green, we will merge this PR

PR #2435, opened 2021-08-19
BUG: Backward compat support for pipeline

larsoner:
Prior to #2390 code like this worked:
```
from dipy.align import (affine_registration, center_of_mass,
                        translation, rigid, affine, resample,
                        imwarp, metrics)
moving_zoomed, reg_affine = affine_registration(
    moving_zoomed, static_zoomed, moving_affine, static_affine,
    nbins=32, metric='MI', pipeline=[center_of_mass, translation],
    level_iters=niter[step], sigmas=sigmas, factors=factors)
```
But due to #2390 moving `pipeline` from `list of callable` to `list of str` arguments, this now raises:
```
../../.local/lib/python3.10/site-packages/dipy/align/_public.py:475: in affine_registration
    raise ValueError("Not supported registration method")
E   ValueError: Not supported registration method
```
I saw two ways out:

1. Allow any `callable` and refactor a bunch of code (probably difficult, doubt many people would require this)
2. Check `if step is center_of_mass` etc. and alias to the string equivalent

This PR:

1. Implements option (2) above
2. Adds tests for it
3. Adds `rigid_isoscaling` and `rigid_scaling` callables to the same public namespace as `center_of_mass`, `rigid`, etc. for completeness
4. Fixes some Sphinx/RST errors (single-backticks are for the default Sphinx role, so a couple I changed to double-backticks to be "code mode")

This is just what I inferred to be the right approach here, but if I'm totally on the wrong track feel free to close, or comment and I can revert some of the changes above if they're not wanted!

### DIFF
--- a/dipy/align/_public.py
+++ b/dipy/align/_public.py
@@ -455,12 +455,11 @@ def affine_registration(moving, static,
     # Convert pipeline to sanitized list of str
     pipeline = list(pipeline)
     for fi, func in enumerate(pipeline):
-        if not isinstance(func, str):
-            if callable(func):
-                for key, val in _METHOD_DICT.items():
-                    if func is val[0]:  # if they passed the callable equiv.
-                        pipeline[fi] = func = key
-                        break
+        if callable(func):
+            for key, val in _METHOD_DICT.items():
+                if func is val[0]:  # if they passed the callable equiv.
+                    pipeline[fi] = func = key
+                    break
         if not isinstance(func, str) or func not in _METHOD_DICT:
             raise ValueError(f'pipeline[{fi}] must be one of '
                              f'{list(_METHOD_DICT)}, got {repr(func)}')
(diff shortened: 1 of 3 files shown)


larsoner:
Two failures, one is a SSL / download timeout and the other is reported in #2436

dipy/align/_public.py
```diff
@@ -450,6 +452,19 @@ def affine_registration(moving, static,
                                 sigmas=sigmas,
                                 factors=factors)
 
+    # Convert pipeline to sanitized list of str
+    pipeline = list(pipeline)
+    for fi, func in enumerate(pipeline):
+        if not isinstance(func, str):
+            if callable(func):
```
skoudoro, line 459: I am not sure that the first `if` is necessary. Why you do not keep only `if callable(func):`

dipy/align/_public.py
```diff
@@ -506,11 +509,28 @@ def affine_registration(moving, static,
 rigid.__doc__ = ("Implements a rigid transform. "
                  "Based on `affine_registration()`.")
 
+rigid_isoscaling = partial(affine_registration, pipeline=['rigid'])
```
skoudoro, line 512: `pipeline=['rigid_isoscaling']`  instead ?

dipy/align/_public.py
```diff
@@ -506,11 +509,28 @@ def affine_registration(moving, static,
...
+rigid_isoscaling.__doc__ = ("Implements a rigid isoscaling transform. "
+                            "Based on `affine_registration()`.")
+
+rigid_scaling = partial(affine_registration, pipeline=['rigid'])
```
skoudoro, line 516: `pipeline=['rigid_scaling']`  instead ?

skoudoro (commented):
Hi @larsoner,

Like the previous PR, apologize for getting back to you so late.

Overall, it looks good to me. See below for a first quick review,  I still need to test your code locally.

I am fine with the proposed solution. It would be good if @arokem or someone else gives their opinion also.

larsoner:
Comments addressed, failures are just connection timeouts I think

skoudoro (approved):
I confirm, timeout issue, CI restarted.

I will wait until tomorrow to see if there are additional comments. If not, I will go ahead and merge it.

Thank you for the update!

skoudoro:
Ok, no additional comments. Thank you @larsoner ! Merging

Issue #2446, opened 2021-09-30
Use of bitwise or with non-bool inputs results in ufunc 'bitwise_or' error


DanNBullock:
https://github.com/dipy/dipy/blob/8898fc962d5aaf7f7cdbf82b027054070fcef49d/dipy/tracking/utils.py#L918

Use of bitwise or with non-bool inputs results in 

> ufunc 'bitwise_or' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''

for array types float64 and

> ufunc 'bitwise_or' output (typecode 'i') could not be coerced to provided output parameter (typecode '?') according to the casting rule ''same_kind''

for int32.

[Function documentation does not mandate use of boolean input array](https://github.com/dipy/dipy/blob/8898fc962d5aaf7f7cdbf82b027054070fcef49d/dipy/tracking/utils.py#L889-L892), and implicitly accepts non-bool inputs.

[Function test](https://github.com/dipy/dipy/blob/8898fc962d5aaf7f7cdbf82b027054070fcef49d/dipy/tracking/tests/test_utils.py#L583-L598) currently only tests for bool inputs, likely overlooking this use case.

Recommended fix:  use `>0` to force input array type to bool and maintain consistency with documentation.

e.g:

https://github.com/dipy/dipy/blob/8898fc962d5aaf7f7cdbf82b027054070fcef49d/dipy/tracking/utils.py#L916-L920

becomes

    for i in range(len(rois)):
        if include[i]:
            include_roi |= rois[i]>0
        else:
            exclude_roi |= rois[i]>0

Sidenote:  seems similar to previously noted issue
https://github.com/dipy/dipy/pull/731

PR #2447, opened 2021-09-30
reduce_rois: Force input array type to bool to avoid bitwise or errors

DanNBullock:
Currently input ROIs arrays are implicitly accepted as non bool.  This results in errors with input arrays of type float and int, as documented in this issue.

https://github.com/dipy/dipy/issues/2446

Using the greater than zero operation should force float and int input arrays to type bool without introducing other troubles.

### DIFF
--- a/dipy/tracking/utils.py
+++ b/dipy/tracking/utils.py
@@ -915,9 +915,9 @@ def reduce_rois(rois, include):
 
     for i in range(len(rois)):
         if include[i]:
-            include_roi |= rois[i]
+            include_roi |= rois[i]>0
         else:
-            exclude_roi |= rois[i]
+            exclude_roi |= rois[i]>0
 
     return include_roi, exclude_roi
 


arokem:
Thanks for the PR! Should those `>` be '!=' instead? What do you want to happen with voxels that have negative values in them? Another option is to raise an error for inputs that "don't make sense" (i.e., anything that isn't boolean), making the user decide how they want to convert into boolean values.

Also, once we resolve that question, would you mind adding a test that fails without this fix? Obviously, our current test-suite doesn't capture the situation described in #2446. The test for this  function is [here](https://github.com/dipy/dipy/blob/master/dipy/tracking/tests/test_utils.py#L583)

DanNBullock:
> Should those > be '!=' instead?

I think that would work.  Personally, I'm only familiar with using ROIs that have positive values, but I suspect there could be use cases involving functional-derived data with an applied threshold.  I suppose if that ends up being overly inclusive for some users, they can apply some preprocessing to their ROIs to obtain the desired behavior/outcome?  Furthermore `!=` is more consistent than `>` with the literal reading of the in-function documentation.

https://github.com/dipy/dipy/blob/8898fc962d5aaf7f7cdbf82b027054070fcef49d/dipy/tracking/utils.py#L892 

Insofar as the typing is concerned, I encountered this issue because an ROI creation function I coded up was creating (unintentionally) float-based ROI values 

https://github.com/DanNBullock/wma_pyTools/blob/145363163f10ea0a0d082ac91f61e17501c0ab63/WMA_pyFuncs.py#L282

Which I was basing off of a nltools method for creating a spherical ROI

https://github.com/cosanlab/nltools/blob/c1788854f753d9b30e0a9ce83c640e76d0f410e2/nltools/mask.py#L64

So I think there are indeed good-faith use cases where ROIs of this sort might arise (I think nilearn is mostly operating with bools though).

> Another option is to raise an error for inputs that "don't make sense" (i.e., anything that isn't boolean)

In either of the above cases the resultant array, though only containing ones and zeros (as either floats or ints)--and therefore in principle convertible to bool--would have been a fairly sensible input in my mind, so I'm not sure a full on error is warranted.  That being said, it may also be desirable to rigidly designate the type of this input (though I'm not sure what further benefit this would obtain).  As I mentioned earlier, I think the documentation itself (implicitly) speaks to the permissibility of non-bool, so throwing an error would likely involve a change of the documentation as well. Which of these options (error vs. conversion) would be more in keeping with the design philosophy of dipy?


With respect to the creation of a test for this, I'm not entirely sure how one would go about creating an np assert-based test for this issue?  As near as I can tell, the tests implemented in the dipy test suite seem to be predicated on the assumption that the existing function runs to completion/without erroring _and then_ tests are performed on the output to assess for fidelity with expected values.  In this case though, there's no opportunity to assess the fidelity of the outputs because the current form of the function (in effect) doesn't accept inputs of the pertinent types (at least, not without erroring).  I've looked through some of the other tests under test_utils, but I haven't found anything I think I could model a test off of (I certainly haven't seen any try--catch usages).  Is there a recommended test standard for this sort of case?

arokem:
First, regarding your last question, there  is a way to test that a particular error is raised. See [here](https://docs.pytest.org/en/6.2.x/assert.html#assertions-about-expected-exceptions)

But in this case, we want to make sure that the error is _not raised_ and that everything proceeds as expected. So, a really good way to fix a bug in DIPY is to: 

1. Implement a test that should pass, but doesn't in the current state of affairs
2. Implement a fix that makes the test pass. 

The benefit of this sequence is that it first demonstrates that there is a problem, and then fixes it, making sure that it doesn't creep back into the code-base, by implementing the test that would catch it.

And regarding the design: I am actually inclined to implement something rather conservative here. The reason is that I could see all kinds of errors happening if people were allowed to pass any input here. Should negative numbers designated as inside the ROI, or outside of it? Well, that depends on what the user expects, and different users might reasonably expect different things. So, my suggestion would be to implement better error handling -- provide an informative message -- rather than change the behavior of the function. WDYT?

DanNBullock:
Sounds good, took me a while to find the `npt.assert_raises()` function (example use [here](https://github.com/dipy/dipy/blob/8898fc962d5aaf7f7cdbf82b027054070fcef49d/dipy/tracking/tests/test_utils.py#L227)), which seems that it would be the appropriate function for this task (e.g. raise error if not [desired type]).

That being said I'm still not sure if I have a solidified intuition about the desired acceptable input datatype.  Sure, we could create a test and in-function check which precludes negative numbers, but this would still leave the matter of (positive) floats and ints undecided.  Does "conservative" (as used above) entail strict enforcement of bool-only inputs?  That'd be a fine and sensible solution as far as I'm concerned, but I'm not sure that's what is implied by the above discussion.  Once that question is adjudicated, I think I will have enough information to proceed.

As a side note though, strict enforcement of bools for this function has some (minor) implications for functions that use reduce_rois() as a sub-function or input.  Namely, this is [near_roi()](https://github.com/dipy/dipy/blob/c846bf5a23b7e95343a9cf231df2653473602456/dipy/tracking/utils.py#L723-L791) and [select_by_rois()](https://github.com/dipy/dipy/blob/c846bf5a23b7e95343a9cf231df2653473602456/dipy/tracking/streamline.py#L200-L324).  Should a similar check/enforcement be implemented for these functions as well, given that a user may not be directly interacting with reduce_rois and that those functions currently have a similarly ambiguous specification of acceptable inputs (e.g. [near_roi](https://github.com/dipy/dipy/blob/c846bf5a23b7e95343a9cf231df2653473602456/dipy/tracking/utils.py#L736-L738) and [select_by_rois](https://github.com/dipy/dipy/blob/c846bf5a23b7e95343a9cf231df2653473602456/dipy/tracking/streamline.py#L213-L216))?

skoudoro:
Hi @DanNBullock,

We plan to talk about your PR during the DIPY online meeting tomorrow (October 7th, 2021 | 1pm EST / 7pm CET / 10am PT).

It would be great if you could attend this meeting if you want to clarify some points of this PR.

More information on https://github.com/dipy/dipy/discussions/2398

Thank you!

DanNBullock:
Sure thing, I'll be there.

arokem:
Yep - looks good to me! And the added testing for this behavior is great. Once the CI runs, I think we can go ahead and merge this.

PR #2453, opened 2021-10-15
make it compatible when number of volume is 2

giulia-berto:
bvecs should be transpose also when number of volume is 2 (bvecs.shape[1] =2)


giulia-berto:
sure, here are the two files I am having problems with (I had to change the file extension to .txt to be able to upload them)
[bval.txt](https://github.com/dipy/dipy/files/7384669/bval.txt)
[bvec.txt](https://github.com/dipy/dipy/files/7384670/bvec.txt)

arokem:
https://github.com/giulia-berto/dipy/pull/1

giulia-berto:
I made an extra change, what do you think? I should now delete the original change in dipy/io/gradients.py, right?

soichih:
We had cases where people try to process b0-dwi data with only 1 or 2 volumes.

PR #2468, opened 2021-11-01
[MNT] Use windows-latest on azure pipeline

skoudoro:
We are currently using `vs2017-win2016` Image on our Azure pipeline. However, The windows-2016 environment will be deprecated on November 15, 2021, and removed on March 15, 2022. 

So, we Migrate to windows-latest instead (which uses vs2019).

For more details see https://github.com/actions/virtual-environments/issues/4312


arokem:
Yep! Thanks

PR #2471, opened 2021-11-02
DOC: Fix coding style guideline link

jhlegarreta:
Fix coding style guideline link and make links consistent.


jhlegarreta:
I guess Sphinx should get this correctly.

Makes more sense to me to point to the file rather than to the deployed URL.

arokem:
Thanks!

PR #2472, opened 2021-11-02
ENH: Return unique invalid streamline removal indices

jhlegarreta:
- ENH: Return unique invalid streamline removal indices
- STYLE: Improve the invalid streamline removal test
- STYLE: Prefer returning a list for indices to keep

### DIFF
--- a/dipy/io/stateful_tractogram.py
+++ b/dipy/io/stateful_tractogram.py
@@ -548,6 +548,8 @@ class StatefulTractogram(object):
             indices_to_remove.append(bisect(
                 self._tractogram.streamlines._offsets, i) - 1)
 
+        indices_to_remove = sorted(set(indices_to_remove))
+
         indices_to_keep = np.setdiff1d(np.arange(len(self._tractogram)),
                                        np.array(indices_to_remove)).astype(int)
 
--- a/dipy/io/stateful_tractogram.py
+++ b/dipy/io/stateful_tractogram.py
@@ -550,8 +550,9 @@ class StatefulTractogram(object):
 
         indices_to_remove = sorted(set(indices_to_remove))
 
-        indices_to_keep = np.setdiff1d(np.arange(len(self._tractogram)),
-                                       np.array(indices_to_remove)).astype(int)
+        indices_to_keep = list(
+            np.setdiff1d(np.arange(len(self._tractogram)),
+                         np.array(indices_to_remove)).astype(int))
 
         tmp_streamlines = self.streamlines[indices_to_keep]
         tmp_dpp = self._tractogram.data_per_point[indices_to_keep]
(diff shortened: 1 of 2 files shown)


jhlegarreta:
Unless I'm missing something, another option would be to replace the loop and the set by:
```
offsets = sorted(set(ic_offsets_indices).intersection(self._tractogram.streamlines._offsets))
indices_to_remove = [np.asarray(self._tractogram.streamlines._offsets == v).nonzero()[0].item() for v in offsets]
```

Looks more intuitive than using the bisection then needing to get the set, but might be subjective.

jhlegarreta:
Build failures are unrelated:
```
2021-11-02T22:59:34.2984859Z FAILED segment/tests/test_mrf.py::test_icm_square - AssertionError
```

dipy/io/stateful_tractogram.py
```diff
@@ -548,8 +548,11 @@ def remove_invalid_streamlines(self, epsilon=1e-3):
             indices_to_remove.append(bisect(
                 self._tractogram.streamlines._offsets, i) - 1)
 
-        indices_to_keep = np.setdiff1d(np.arange(len(self._tractogram)),
-                                       np.array(indices_to_remove)).astype(int)
+        indices_to_remove = sorted(set(indices_to_remove))
```
skoudoro, line 551: I do not understand why you need to sort them. Is it just for the test? Can you explain?

I suppose, the use of `set` is due to multiple duplications, right?

dipy/io/stateful_tractogram.py
```diff
@@ -548,8 +548,11 @@ def remove_invalid_streamlines(self, epsilon=1e-3):
...
+
+        indices_to_keep = list(
+            np.setdiff1d(np.arange(len(self._tractogram)),
+                         np.array(indices_to_remove)).astype(int))
```
skoudoro, line 555: What is the advantage to cast it into `list`?  if it is just styling, I would recommend keeping it as a NumPy array.

skoudoro (commented):
Hi @jhlegarreta,

It looks good to me, thanks! I just have a few questions below.

dipy/io/stateful_tractogram.py
skoudoro: ok, I see for the `set` now.

jhlegarreta: Please read the commit message in 07bf4c2. Although sorting is not strictly necessary, on the one hand it makes sense to return sorted values and on the other hand it helps when you need to explore the values. Also, the indices to keep are sorted by construction.

dipy/io/stateful_tractogram.py
jhlegarreta: The existing docstring said that two lists were being returned, which was untrue. One was a list and the other one was a NumPy array. In either case these are 1D structures, and it makes sense/is more consistent to return the same type for both variables. So it makes more sense to return a list rather than changing the docstring. See the commit message in 3be85d0.

skoudoro:
ok, thanks for the explanation. Merging

PR #2487, opened 2021-12-02
ENH: Better error message

larsoner:
Just a minor update to improve the error message.

Also, are the units of `radius` and `sigma_diff` for CCMetric in voxels (I assume it's this), or mm (i.e., the image's `affine` dims will be factored in to compute the number of voxels)?


skoudoro (approved):
Thank you for this update @larsoner.

I confirm units are in voxels.

PR #2497, opened 2022-01-04
Remove python 3.6 from Azure pipelines

skoudoro:
Python3.6 is officially deprecated, so we had to remove it from Azure Pipelines


skoudoro:
It seems we have an issue with macOs. Azure pipeline updated the mac operating system. We are using BigSur.  Since then, our tests are failing. More precisely, look like a Cython issue. We need to find a machine to test that locally.

Issue #2490, opened 2021-12-15
[BUG] MRI-CT alignment failure


alexrockhill:
## Description

It looks like there is something suboptimal/unstable about the way `Dipy` does the image registration relative to `antspyx`. Hopefully it's just a small fix. Here is how to reproduce the problem.

CT file: https://drive.google.com/file/d/1Epaw_kVLElBIbgo184KIsLSH7YuGliLI/view?usp=sharing
MR file: https://drive.google.com/file/d/1deiJMNH00NU0z3Lr6IPUe0HI7EKjyTo-/view?usp=sharing
```python
import numpy as np
from dipy.align import affine_registration, center_of_mass, translation, rigid
import nibabel as nib
import matplotlib.pyplot as plt


def plot_overlay(image, compare):
    """Define a helper function for comparing plots."""
    image = nib.orientations.apply_orientation(
        np.asarray(image.dataobj), nib.orientations.axcodes2ornt(
            nib.orientations.aff2axcodes(image.affine))).astype(np.float32)
    compare = nib.orientations.apply_orientation(
        np.asarray(compare.dataobj), nib.orientations.axcodes2ornt(
            nib.orientations.aff2axcodes(compare.affine))).astype(np.float32)
    fig, axes = plt.subplots(1, 3, figsize=(12, 4))
    for i, ax in enumerate(axes):
        ax.imshow(np.take(image, [image.shape[i] // 2], axis=i).squeeze().T,
                  cmap='gray')
        ax.imshow(np.take(compare, [compare.shape[i] // 2],
                          axis=i).squeeze().T, cmap='gist_heat', alpha=0.5)
        ax.invert_yaxis()
        ax.axis('off')
    fig.tight_layout()


moving = nib.load('sub-12_CT.nii.gz')
static = nib.load('sub-12_T1.mgz')

_, reg_affine = affine_registration(
    np.array(moving.dataobj),
    np.array(static.dataobj),
    moving_affine=moving.affine,
    static_affine=static.affine,
    nbins=32,
    metric='MI',
    pipeline=[center_of_mass, translation])

moved_img, reg_affine = affine_registration(
    np.array(moving.dataobj),
    np.array(static.dataobj),
    moving_affine=moving.affine,
    static_affine=static.affine,
    nbins=32,
    metric='MI',
    pipeline=[rigid],
    starting_affine=reg_affine)


moved = nib.MGHImage(moved_img.astype(np.float32), static.affine)
plot_overlay(static, moved)

[...]

PR #2498, opened 2022-01-07
Expose optimizer method in affine registration

larsoner:
WIP to explore/work on https://github.com/dipy/dipy/issues/2490. Might be useful to expose these publicly anyway...?


larsoner:
@drombas @Garyfallidis let me know if you think this is useful, in which case I'll keep going. Otherwise I'm happy to close as YAGNI

skoudoro:
Thanks @larsoner for the original idea. I just clean it up

PR #2488, opened 2021-12-13
Minimize memory footprint whereever possible, add joblib support for …

dPys:
@arokem
 
 [PR #53](https://github.com/nipreps/eddymotion/issues/53) and [PR #60](https://github.com/nipreps/eddymotion/pull/60)

Substantial speedup and less virtual memory overhead by avoiding/deleting unneeded objects + garbage collection

PR #2510, opened 2022-02-01
Optimize sfm (reboot)

dPys:
This is a reboot of PR #2488, with fresh commit history + rebased with master (i.e. no more missing utils/multiproc.py file), and a fix on the test per @skoudoro 's request. 

@arokem -- I discovered that when fitting the solver in parallel, the `warm_start` feature was causing numerical divergence. Intuitively, this makes sense because fitting solvers in an embarrassingly parallel manner precludes the possibility of sequentially drawing upon the previous fit estimates for initialization. After disabling, this is no longer an issue, and the results are identical with joblib (when specifying any number of n_procs). And frankly, I propose keeping it disabled in this context anyway because as sklearn notes in their docs: "repeatedly calling an sklearn estimator fit when `warm_start=True` can result in a different solution than when calling fit a single time because of the way the data is shuffled".

Not that it was necessary, but I've also added a `seed` argument to the `fit` method of `SparseFascicleModel` for additional control. 

```
 dipy/reconst/tests/test_sfm.py::test_design_matrix ✓                                                    14% █▌        
 dipy/reconst/tests/test_sfm.py::test_sfm ✓                                                              29% ██▉       
 dipy/reconst/tests/test_sfm.py::test_predict ✓                                                          43% ████▍     
 dipy/reconst/tests/test_sfm.py::test_sfm_background ✓                                                   57% █████▊    
 dipy/reconst/tests/test_sfm.py::test_sfm_stick ✓                                                        71% ███████▎  
 dipy/reconst/tests/test_sfm.py::test_sfm_sklearnlinearsolver ✓                                          86% ████████▋ 
 dipy/reconst/tests/test_sfm.py::test_exponential_iso ✓                                                 100% ██████████
================================================== warnings summary ===================================================
-- Docs: https://docs.pytest.org/en/latest/warnings.html

Results (100.74s):
       7 passed
```

Assuming the disabling of warm_start is acceptable and the CI is happy, then this PR should be ready to merge...

@dPys

### DIFF
--- a/dipy/reconst/sfm.py
+++ b/dipy/reconst/sfm.py
@@ -339,7 +339,7 @@ def sfm_design_matrix(gtab, sphere, response, mode='signal'):
 class SparseFascicleModel(ReconstModel, Cache):
     def __init__(self, gtab, sphere=None, response=[0.0015, 0.0005, 0.0005],
                  solver='ElasticNet', l1_ratio=0.5, alpha=0.001,
-                 isotropic=None):
+                 isotropic=None, seed=42):
         """
         Initialize a Sparse Fascicle Model
 
@@ -408,7 +408,8 @@ class SparseFascicleModel(ReconstModel, Cache):
         self.isotropic = isotropic
         if solver == 'ElasticNet':
             self.solver = lm.ElasticNet(l1_ratio=l1_ratio, alpha=alpha,
-                                        positive=True, warm_start=True)
+                                        positive=True, warm_start=False,
+                                        random_state=seed)
         elif solver == 'NNLS' or solver == 'nnls':
             self.solver = opt.NonNegativeLeastSquares()
 
--- a/dipy/reconst/sfm.py
+++ b/dipy/reconst/sfm.py
@@ -18,6 +18,8 @@ This is an implementation of the sparse fascicle model described in
 import warnings
 
 import numpy as np
+import gc
+from collections import OrderedDict
 
 try:
     from numpy import nanmean
@@ -25,6 +27,7 @@ except ImportError:
     from scipy.stats import nanmean
 
 from dipy.utils.optpkg import optional_package
+from dipy.utils.multiproc import determine_num_processes
 import dipy.core.gradients as grad
 import dipy.core.optimize as opt
 import dipy.sims.voxel as sims
@@ -33,6 +36,7 @@ from dipy.reconst.base import ReconstModel, ReconstFit
 from dipy.reconst.cache import Cache
 from dipy.core.onetime import auto_attr
 
+joblib, has_joblib, _ = optional_package('joblib')
 sklearn, has_sklearn, _ = optional_package('sklearn')
 lm, _, _ = optional_package('sklearn.linear_model')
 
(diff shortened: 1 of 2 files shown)


skoudoro:
Thank you for the update @dPys.

Can you close #2488 to avoid any confusion? since you cited it on the description we can easily go back to our previous comments without letting this PR open.
 
I will have a deeper look but overall,  it looks fine.

CI's failures are not related so they can be ignored. Today, I just started the migration to Github Action, so hopefully, I will fix all these errors at the same time.

Thanks and we will get back to you ASAP

dPys:
> Thank you for the update @dPys.
> 
> Can you close #2488 to avoid any confusion? since you cited it on the description we can easily go back to our previous comments without letting this PR open.
> 
> I will have a deeper look but overall, it looks fine.
> 
> CI's failures are not related so they can be ignored. Today, I just started the migration to Github Action, so hopefully, I will fix all these errors at the same time.
> 
> Thanks and we will get back to you ASAP

Cool, done.

dipy/reconst/sfm.py
```diff
@@ -489,22 +580,28 @@ def fit(self, data, mask=None):
...
+
         S0 : ndarray
             The mean non-diffusion-weighted signal.
+
         iso : IsotropicFit class instance
             A representation of the isotropic signal, together with parameters
             of the isotropic signal in each voxel, that is capable of
             deriving/predicting an isotropic signal, based on a gradient-table.
         """
+        super().__init__(self, model)
```
arokem, line 603: I believe that if you add this, you can remove the assignment `self.model` below on L605.

dipy/reconst/tests/test_sfm.py
```diff
@@ -2,6 +2,7 @@
 import numpy as np
 import numpy.testing as npt
 import pytest
+import time
```
arokem, line 5: I don't think this is being used.

arokem (commented):
Looks good to me! I had just a couple of really small and easy-to-address comments.

dipy/reconst/tests/test_sfm.py
dPys: yep, I had included it initially for benchmarking, but will now remove.

dipy/reconst/sfm.py
dPys: If I do that, then the following occurs:
```
    def odf(self, sphere):
        """
        The orientation distribution function of the SFM
    
        Parameters
        ----------
        sphere : Sphere
            The points in which the ODF is evaluated
    
        Returns
        -------
        odf :  ndarray of shape (x, y, z, sphere.vertices.shape[0])
    
        """
>       odf_matrix = self.model.cache_get('odf_matrix', key=sphere)
E       AttributeError: 'SparseFascicleFit' object has no attribute 'cache_get'
```
for the subsequent `odf` method.

dipy/reconst/tests/test_sfm.py
dPys: yep, removed.

dipy/reconst/sfm.py
arokem: OK - I'm probably wrong then!

arokem:
Assuming the CI failures are unrelated, and no other errors appear, I believe this is ready to go!

skoudoro:
Thanks @dPys, merging !

PR #2512, opened 2022-02-08
Masking for affine registration

samcoveney:
Implementation of masking for affine registration. It is possible to use a static_mask and moving_mask, and to use the pipeline interface. A new example is included.

See #2508, #2179, and #1969 for related discussions and requests. This implementation finished and expanded the implementation referenced in #2179.

### DIFF
--- a/dipy/align/_public.py
+++ b/dipy/align/_public.py
@@ -5,6 +5,7 @@ streamlines.
 
 
 """
+from warnings import warn
 import re
 import collections.abc
 from functools import partial
@@ -343,6 +344,8 @@ def affine_registration(moving, static,
                         sigmas=None,
                         factors=None,
                         ret_metric=False,
+                        moving_mask=None,
+                        static_mask=None,
                         **metric_kwargs):
     """
     Find the affine transformation between two 3D images. Alternatively, find
@@ -403,6 +406,14 @@ def affine_registration(moving, static,
     ret_metric : boolean, optional
         Set it to True to return the value of the optimized coefficients and
         the optimization quality metric.
+
+    moving_mask : array, shape (S', R', C') or (R', C'), optional
+        moving image mask that defines which pixels in the moving image
+        are used to calculate the mutual information.
+
+    static_mask : array, shape (S, R, C) or (R, C), optional
+        static image mask that defines which pixels in the static image
+        are used to calculate the mutual information.
 
     nbins : int, optional
         MutualInformationMetric key-word argument: the number of bins to be
@@ -472,16 +483,31 @@ def affine_registration(moving, static,
     # Go through the selected transformation:
     for func in pipeline:
         if func == "center_of_mass":
-            transform = transform_centers_of_mass(static, static_affine,
-                                                  moving, moving_affine)
+
+            if starting_affine is not None:
+                wm = "starting_affine overwritten by centre_of_mass transform"
+                warn(wm, UserWarning)
+
+            # multiply images by masks for transform_centers_of_mass
+            static_masked, moving_masked = static, moving
+            if static_mask is not None:
+                static_masked = static*static_mask
+            if moving_mask is not None:
+                moving_masked = moving*moving_mask
+
+            transform = transform_centers_of_mass(static_masked, static_affine,
+                                                  moving_masked, moving_affine)
             starting_affine = transform.affine
+
         else:
             transform = _METHOD_DICT[func][1]()
             xform, xopt, fopt \
                 = affreg.optimize(static, moving, transform, None,
                                   static_affine, moving_affine,
                                   starting_affine=starting_affine,
-                                  ret_metric=True)
+                                  ret_metric=True,
+                                  static_mask=static_mask,
+                                  moving_mask=moving_mask)
             starting_affine = xform.affine
 
     # After doing all that, resample once at the end:
@@ -532,7 +558,7 @@ _METHOD_DICT = dict(  # mapping from str key -> (callable, class) tuple
 
 
 def register_series(series, ref, pipeline=None, series_affine=None,
-                    ref_affine=None):
+                    ref_affine=None, static_mask=None):
     """Register a series to a reference image.
 
     Parameters
@@ -554,6 +580,10 @@ def register_series(series, ref, pipeline=None, series_affine=None,
     series_affine, ref_affine : 4x4 arrays, optional.
         The affine. If provided, this input will over-ride the affine provided
         together with the nifti img or file.
+
+    static_mask : array, shape (S, R, C) or (R, C), optional
+        static image mask that defines which pixels in the static image
+        are used to calculate the mutual information.
 
     Returns
     -------
@@ -591,14 +621,16 @@ def register_series(series, ref, pipeline=None, series_affine=None,
                 this_moving, ref,
                 moving_affine=series_affine,
                 static_affine=ref_affine,
-                pipeline=pipeline)
+                pipeline=pipeline,
+                static_mask=static_mask)
             xformed[..., ii] = transformed
             affines[..., ii] = reg_affine
 
     return xformed, affines
 
 
-def register_dwi_series(data, gtab, affine=None, b0_ref=0, pipeline=None):
+def register_dwi_series(data, gtab, affine=None, b0_ref=0, pipeline=None,
+                        static_mask=None):
     """
     Register a DWI series to the mean of the B0 images in that series (all
     first registered to the first B0 volume)
@@ -626,6 +658,9 @@ def register_dwi_series(data, gtab, affine=None, b0_ref=0, pipeline=None):
         The transformations to perform in sequence (from left to right):
         Default: ``[center_of_mass, translation, rigid, affine]``
 
+    static_mask : array, shape (S, R, C) or (R, C), optional
+        static image mask that defines which pixels in the static image
+        are used to calculate the mutual information.
 
     Returns
     -------
@@ -644,7 +679,8 @@ def register_dwi_series(data, gtab, affine=None, b0_ref=0, pipeline=None):
         # First, register the b0s into one image and average:
         b0_img = nib.Nifti1Image(data[..., gtab.b0s_mask], affine)
         trans_b0, b0_affines = register_series(b0_img, ref=b0_ref,
-                                               pipeline=pipeline)
+                                               pipeline=pipeline,
+                                               static_mask=static_mask)
         ref_data = np.mean(trans_b0, -1)
     else:
         # There's only one b0 and we register everything to it
@@ -656,7 +692,8 @@ def register_dwi_series(data, gtab, affine=None, b0_ref=0, pipeline=None):
     series_arr = np.concatenate([ref_data, moving_data], -1)
     series = nib.Nifti1Image(series_arr, affine)
 
-    xformed, affines = register_series(series, ref=0, pipeline=pipeline)
+    xformed, affines = register_series(series, ref=0, pipeline=pipeline,
+                                       static_mask=static_mask)
     # Cut out the part pertaining to that first volume:
     affines = affines[..., 1:]
     xformed = xformed[..., 1:]
--- a/dipy/align/imaffine.py
+++ b/dipy/align/imaffine.py
@@ -43,6 +43,8 @@
 
 """
 
+from warnings import warn
+
 import numpy as np
 import numpy.linalg as npl
 import scipy.ndimage as ndimage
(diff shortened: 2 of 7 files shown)


skoudoro:
Hi @samcoveney,

Thank you again for doing this!

We will look deeper into it this week or next week. Can you address all the pep8 issues above during this time?

Thank you

samcoveney:
I probably can do, or I'll fix it in the final instance once it's been
reviewed. I'll try to get around to it beforehand though

dipy/align/imaffine.py
```diff
@@ -566,6 +575,29 @@ def setup(self, transform, static, moving, static_grid2world=None,
         self.affine_map = AffineMap(P, static.shape, static_grid2world,
                                     moving.shape, moving_grid2world)
 
+        # Masks can only be used with dense sampling
+        if (self.sampling_proportion is None or
+                self.sampling_proportion == 1.0):
```
skoudoro, line 580: `if self.sampling_proportion in [1.0, None]` instead. More pythonic 😄

dipy/align/imaffine.py
```diff
@@ -566,6 +575,29 @@ def setup(self, transform, static, moving, static_grid2world=None,
...
+
+            if static_mask is not None:
+                self.static_mask = static_mask.astype(np.int32)
+            else:
+                self.static_mask = None
+
+            if moving_mask is not None:
+                self.moving_mask = moving_mask.astype(np.int32)
+            else:
+                self.moving_mask = None
```
skoudoro, line 590: same comment as above: Not sure, this `else` is necessary

samcoveney:
Sorry about the syntax errors, I must have introduced these when doing some
final pep8 fixes and forgot to reinstall before testing the example code
once again - silly mistake, as I did run my example before push.

The tests already call these functions, they just don't cover these new
arguments. Can you be very clear about exactly what you would like me to do
here? What would make it clear that these tests were passed?

skoudoro:
> they just don't cover these new arguments. Can you be very clear about exactly what you would like me to do here?

it will be good to handle these new arguments in the tests. So, yes, no problem:

- Check that the new warnings appear  as expected with `np.assert_warns` like [here](https://github.com/dipy/dipy/blob/c846bf5a23b7e95343a9cf231df2653473602456/dipy/workflows/tests/test_reconst_dki.py#L101)
- pass a mask with all zeros and see if the code handles it well. Also the opposite, mask of all  1 and check if we get the same result as before.
- Your tutorial shows the normal case so the 2 cases above should be enough for now 

Thanks !

dipy/align/imaffine.py
```diff
@@ -566,6 +575,29 @@ def setup(self, transform, static, moving, static_grid2world=None,
...
                                     moving.shape, moving_grid2world)
 
+        # Masks can only be used with dense sampling
+        if (self.sampling_proportion is None or
+                self.sampling_proportion == 1.0):
+
+            if static_mask is not None:
+                self.static_mask = static_mask.astype(np.int32)
+            else:
+                self.static_mask = None
```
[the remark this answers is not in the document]
samcoveney: I disagree, this would leave self.static_mask undefined, but this value must be defined.

dipy/align/imaffine.py
samcoveney: I disagree, this would leave self.moving_mask undefined, but this value must be defined.

samcoveney:
I am struggling to adapt test_imaffine.py because there seems to be some issues with testing in general. I will require some help to figure out what is going on. Please try to run the tests (for master branch) with:

    coverage run --source=dipy.align.imaffine -m pytest -s --doctest-modules --verbose dipy/align/tests/test_imaffine.py

I am getting some failures anywhere there is something like array.min(), which gives error "E       TypeError: float() argument must be a string or a number, not '_NoValueType'".

This error will appear even if we define a pointless array inside of imaffine.py, for example in the function AffineRegistration._init_optimizer(), if I do

    tmp = np.zeros([10,10])
    print(tmp.min())

I get the same errors during testing. Specifically for tests test_affreg_all_transforms and test_affreg_defaults

Is there some problem with pytest? Googling this is not seeming to reveal anything, so I need to ask if other people are seeing the same issues please. Otherwise, I am not sure that I can adapt these tests for use with masking, as they seem to fail for no good reason for me.

samcoveney:
I've now updated to fix issues in the last commit, and have included modifications to the tests.

skoudoro:
Interesting you issue with `pytest`, I never encountered it. What is your pytest version?

Otherwise, thanks for updating your PR, I will look into it today.

doc/examples/affine_registration_masks.py
```diff
@@ -0,0 +1,363 @@
...
+"""
+
+from os.path import join as pjoin
+import numpy as np
+import matplotlib.pyplot as plt
+from dipy.viz import regtools
+from dipy.data import fetch_stanford_hardi
+from dipy.data.fetcher import fetch_syn_data
+from dipy.io.image import load_nifti
+from dipy.align.imaffine import (transform_centers_of_mass,
```
skoudoro, line 28: pep8: `transform_centers_of_mass` not used

doc/examples/affine_registration_masks.py
```diff
@@ -0,0 +1,363 @@
...
+                                 AffineMap,
+                                 MutualInformationMetric,
+                                 AffineRegistration)
+from dipy.align.transforms import (TranslationTransform3D,
+                                   RigidTransform3D,
+                                   AffineTransform3D)
```
skoudoro, line 34: pep8: `AffineTransform3D` not used

dipy/align/imaffine.py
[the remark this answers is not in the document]
skoudoro: ok, I see, I got confused with `self.static_mask` and `static_mask`

skoudoro (commented):
ok, Overall it looks good to me. 

Can you address my small comments below?

After that change, I will wait until Wednesday to see if there is any additional comment. If not, I will go ahead and merge it this Wednesday.

Thank you @samcoveney

doc/examples/affine_registration_masks.py
samcoveney: I've removed it now

doc/examples/affine_registration_masks.py
samcoveney: I've removed it now, and also some other imports that were not needed

dipy/align/_public.py
```diff
@@ -404,6 +407,14 @@ def affine_registration(moving, static,
         Set it to True to return the value of the optimized coefficients and
         the optimization quality metric.
 
+    moving_mask : array, shape (S', R', C') or (R', C'), optional
```
arokem, line 410: Do you want to enable passing array nifti or str like for the data? This might also allow passing masks that are in a different space than the data, which could be useful (for example, a brain mask from a T1w image)

dipy/align/imaffine.py
```diff
@@ -215,7 +217,7 @@ def __str__(self):
         return str(self.affine)
 
     def __repr__(self):
-        """Relodable representation - also relies on ndarray's implementation."""
+        """Relodable representation - relies on ndarray's implementation."""
```
arokem, line 220:
```suggestion
        """Reloadable representation - relies on ndarray's implementation."""
```

dipy/align/imaffine.py
```diff
@@ -922,25 +970,50 @@ def _init_optimizer(self, static, moving, transform, params0,
...
 
         """
         self.dim = len(static.shape)
         self.transform = transform
         n = transform.get_number_of_parameters()
         self.nparams = n
 
+        # ensure that masks are not all zeros
+        if np.all(static_mask == 0):
+            warn("static_mask is all zeros, setting to None", UserWarning)
```
arokem, line 988: This means that the entire volume will be used, right? I thiink we can say that explicitly 

```suggestion
            warn("static_mask is all zeros, setting to None (which means that the entire volume will be used)", UserWarning)
```

dipy/align/scalespace.py
```diff
@@ -51,7 +51,7 @@ def __init__(self, image, num_levels,
             mask = np.asarray(image > 0, dtype=np.int32)
 
         # Normalize input image to [0,1]
-        img = (image - image.min())/(image.max() - image.min())
+        img = (image - np.min(image))/(np.max(image) - np.min(image))
```
arokem, line 54: I don't mind the changes made in this file, but is there a particular reason to make these changes? I believe the code path is identical with this syntax?

doc/examples/affine_registration_masks.py
```diff
@@ -0,0 +1,361 @@
...
+This example explains how to compute a transformation to register two 3D
+volumes by maximization of their Mutual Information [Mattes03]_. The
+optimization strategy is similar to that implemented in ANTS [Avants11]_.
+
+We will use masks to define which pixels are used in the Mutual Information.
+Masking can also be done for registration of 2D images rather than 3D volumes.
+
+Masking for registration is much more necessary for cardiac MRI than for brain
+MRI, where it is usually used to specify a region of interest on a 2D static
+image, e.g. the left ventricle in a short axis slice. The prioritizes
```
arokem, line 15:
```suggestion
image, e.g. the left ventricle in a short axis slice. This prioritizes
```

doc/examples/affine_registration_masks.py
```diff
@@ -0,0 +1,361 @@
...
+
+This example explains how to compute a transformation to register two 3D
+volumes by maximization of their Mutual Information [Mattes03]_. The
+optimization strategy is similar to that implemented in ANTS [Avants11]_.
+
+We will use masks to define which pixels are used in the Mutual Information.
+Masking can also be done for registration of 2D images rather than 3D volumes.
+
+Masking for registration is much more necessary for cardiac MRI than for brain
+MRI, where it is usually used to specify a region of interest on a 2D static
```
arokem, line 14:
```suggestion
Masking for registration is useful in a variety of circumstances. For example, in cardiac MRI, 
where it is usually used to specify a region of interest on a 2D static
```

doc/examples/affine_registration_masks.py
arokem, line 15:
```suggestion
image, e.g., the left ventricle in a short axis slice. The prioritizes
```

doc/examples/affine_registration_masks.py
```diff
@@ -0,0 +1,361 @@
...
+
+"""
+.. figure:: transformed_squares_mask_3.png
+   :align: center
+
+   Registration result using both a static mask and a moving mask.
+"""
+
+"""
+In most use cases, it is likely that only a static mask is will be required,
```
arokem, line 313:
```suggestion
In most use cases, it is likely that only a static mask will be required,
```

doc/examples/affine_registration_masks.py
```diff
@@ -0,0 +1,361 @@
...
+e.g. to register a series of images to a single static image.
```
arokem, line 314:
```suggestion
e.g., to register a series of images to a single static image.
```

arokem (commented):
Great work overall! I have a few comments/suggestions. 

The most substantial of these, is that I think that it would be possible to help users even more, by accepting the same set of inputs that we allow for the data: str, nifti or array. In particular, I think that if we accept nifti images, we can enable use of masks that are in a different space than the data (e.g., in the space of a paired T1w acquisition), which could be handy. I am reluctant to ask that you do this now, though, because I am under the impression that this is already useful to you, so this might be a feature to add in a future PR (and would not break back-compatibilty, because the current functionality would still work).

skoudoro:
>  by accepting the same set of inputs that we allow for the data: str, nifti or array

Thank you for your feedback/review. I agree, good catch! indeed, we could do that in a new PR, it should be fast.

dipy/align/_public.py
samcoveney: It's an interesting idea, are you suggesting that this would be useful for aligning images to a T1w image that had a mask already defined, and saved somewhere?

In this case, can't the user just load the mask and pass in the array?

In any case, the public API needs updating as it only accepts 3D data (technically 4D i.e. a series of 3D volumes). I will begin making these modifications later and will consider this change at that stage.

dipy/align/scalespace.py
samcoveney: Changes were made because pytest was having issues with array.min() - see the conversation further up in this pull request. I tried several versions of pytest going back to version 5.0.0, all had the same issue.

samcoveney:
I've accepted most of the suggestion to fix typos - unfortunately these have introduced a few minor pep8 problems...

skoudoro:
Thank you for this work @samcoveney. merging!

Issue #2509, opened 2022-01-31
Easy robustness for streamline_near_roi and near_roi for empty streamlines?


DanNBullock:
## Description
It seems that it would be fairly easy to make [streamline_near_roi](https://github.com/dipy/dipy/blob/be956a529465b28085f8fc435a756947ddee1c89/dipy/tracking/utils.py#L674-L724) and [near_roi](https://github.com/dipy/dipy/blob/be956a529465b28085f8fc435a756947ddee1c89/dipy/tracking/utils.py#L727-L796) robust against various forms of empty streamline inputs.  Is there an intentional desire / design decision that precludes this possibility?

An example of how to implement this would be to use np.any, as demonstrated [here](https://github.com/DanNBullock/dipy/blob/1490492e3f9929a8e28bc682358b957fc8dda328/dipy/tracking/utils.py#L706-L707)

## Way to reproduce
I've implemented tests for this solution in [test_utils here](https://github.com/DanNBullock/dipy/blob/1490492e3f9929a8e28bc682358b957fc8dda328/dipy/tracking/tests/test_utils.py#L347-L359),

However, in a likely unrelated (and possibly local) issue, my current local tests are indicating that the tests associated with the end options "both_end" and "either_end" are returning unexpected results for the *already existing* tests.  The tests specific to this modification seem to be fine.

## Additional notes

- There may be a potential weird edge case, if *literally* every coordinate of every streamline node is 0 (due to the behavior of np.any()), but I would expect such a case to be impossible under normal circumstances (I would not expect a singleton node at 0,0,0).

## Justification / Use case

The suggested modification might be useful for users who may wish to subset (via some criterion) the streamline nodes that that are being passed to streamline_near_roi.  Such an option would reduce the footprint of subsequent cdist computations, in some cases substantially.

An example of such an application would be in the[ implementation of a faster near_roi function, as prototyped here](https://github.com/DanNBullock/wma_pyTools/blob/ee0f96b99f548f661633db0f4cf9068acc90c340/wmaPyTools/segmentationTools.py#L267-L409).  In this example, nodes outside of the ROI bounding box are not passed to streamlines_near_roi, reducing the computation's footprint.  However, this sometimes (or even often) results in passing "empty" streamlines (streamlines with no nodes, as they all occur outside the bounding box).

[...]

PR #2515, opened 2022-02-10
Default to False output for null streamlines in streamline_near_roi

DanNBullock:
Per [the discussion on the issue thread](https://github.com/dipy/dipy/issues/2509), I went ahead and cleaned up [the test_utils section where I had created a test for null streamlines](https://github.com/DanNBullock/dipy/blob/01df7ba4042c90f1d7dc3eedac55bb598c21bb31/dipy/tracking/tests/test_utils.py#L357-L371).  It's slightly inelegant, in that I have to append a non-null streamline to the front in order to bypass issues with near_roi and tol/dtc computations, but it works as implemented.

The [streamline_near_roi change itself](https://github.com/DanNBullock/dipy/blob/01df7ba4042c90f1d7dc3eedac55bb598c21bb31/dipy/tracking/utils.py#L706-L707) appears to be effective and well behaved.

Let me know if additional adjustments are suggested/required.

### DIFF
--- a/dipy/tracking/tests/test_utils.py
+++ b/dipy/tracking/tests/test_utils.py
@@ -360,11 +360,11 @@ def test_near_roi():
     streamlines_null = [np.array([[0., 0., 0.9],
                                  [1.9, 0., 0.],
                                  [3, 2., 2.]]),
-                       np.array([[],
+                        np.array([[],
                                  [],
                                  []]).T,
-                       np.array([]),
-                       []]
+                        np.array([]),
+                        []]
     npt.assert_array_equal(near_roi(streamlines_null, np.eye(4), mask, tol=1),
                            np.array([True, False, False, False]))
     npt.assert_array_equal(near_roi(streamlines_null, np.eye(4), mask),
--- a/dipy/tracking/tests/test_utils.py
+++ b/dipy/tracking/tests/test_utils.py
@@ -335,29 +335,44 @@ def test_target_line_based_out_of_bounds():
 
 
 def test_near_roi():
-    streamlines = [np.array([[0., 0., 0.9],
-                             [1.9, 0., 0.],
-                             [3, 2., 2.]]),
-                   np.array([[0.1, 0., 0],
-                             [0, 1., 1.],
-                             [0, 2., 2.]]),
-                   np.array([[2, 2, 2],
-                             [3, 3, 3]])]
+    # streamlines = [np.array([[0., 0., 0.9],
+    #                          [1.9, 0., 0.],
+    #                          [3, 2., 2.]]),
+    #                np.array([[0.1, 0., 0],
+    #                          [0, 1., 1.],
+    #                          [0, 2., 2.]]),
+    #                np.array([[2, 2, 2],
+    #                          [3, 3, 3]])]
+
+    streamlines =   [np.array([[],
+                              [],
+                              []]).T,
+                      np.array([]),
+                               [],
+                    np.array([[0., 0., 0.9],
+                              [1.9, 0., 0.],
+                              [3, 2., 2.]]),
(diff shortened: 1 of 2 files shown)


dipy/tracking/tests/test_utils.py
```diff
@@ -353,11 +354,27 @@ def test_near_roi():
...
+                           np.array([True, False, False, False]))
+    npt.assert_array_equal(near_roi(streamlinesNULL, np.eye(4), mask),
+                           np.array([False, False, False, False]))
+
     # If there is an affine, we need to use it:
     affine = np.eye(4)
     affine[:, 3] = [-1, 100, -20, 1]
     # Transform the streamlines:
-    x_streamlines = [sl + affine[:3, 3] for sl in streamlines]
+    x_streamlines = [sl + affine[:3, 3] for sl in streamlines[-3:]]
```
arokem, line 377: Why is this needed? It doesn't seem to be referring to the new variable that you created. Is that what you meant to do here? I.e., 

```
 x_streamlines = [sl + affine[:3, 3] for sl in streamlinesNull[-3:]]
```

dipy/tracking/tests/test_utils.py
```diff
@@ -353,11 +354,27 @@ def test_near_roi():
     npt.assert_array_equal(near_roi(streamlines, np.eye(4), mask),
                            np.array([False, True, False]))
 
+    # test for handling of various forms of null streamlines
+    # including a streamline from previous test because near_roi / tol
+    # can't handle completely empty streamline collections
+    streamlinesNULL = [np.array([[0., 0., 0.9],
```
arokem, line 360: PEP8:

```suggestion
    streamlines_null = [np.array([[0., 0., 0.9],
```

arokem (commented):
Thanks for adding this! I had a couple of small questions/suggestions

dipy/tracking/tests/test_utils.py
DanNBullock: I actually don't know what this is... this (to my understanding) doesn't really have anything to do with the changes I'd intended.  It might be a typo?  I believe it should be left in the previous state (i.e. no [-3:])

dipy/tracking/tests/test_utils.py
DanNBullock: Fixed [here]( https://github.com/dipy/dipy/pull/2515/commits/65664c88663003f1b6f25b2c5d22c5a83f6cdbd3) I think.

However, [it seems the pep8 styling is off](https://github.com/dipy/dipy/pull/2515#issuecomment-1038463545)?  In my spyder IDE it didn't seem to have a problem with this, so I'm not entirely sure what the specific fix is.

arokem: From [PEP8](https://www.python.org/dev/peps/pep-0008/#naming-conventions): "mixedCase is allowed only in contexts where that's already the prevailing style (e.g. threading.py), to retain backwards compatibility."

arokem: FWIW, I agree with the automated pep8 that those lines are under-indented. One more space :-)

DanNBullock: I could have sworn I fixed some of this in an earlier push.  In any case [I believe it is fixed now](https://github.com/dipy/dipy/pull/2515/commits/5a59df375109406b4131f9c9d1b692a377b53eab).  Apologies for the false starts on this.  I suppose that even after all these years of working with git I'm still a bit unpracticed.

arokem:
LGTM! Any other comments from anyone? I can go ahead and merge this tomorrow if not.

skoudoro (approved):
+1 for merging.

you can go ahead @arokem

Thanks @DanNBullock !

PR #2123, opened 2020-04-14
WIP: Residual Deep NN

finalelement:
This PR is replacing https://github.com/dipy/dipy/pull/2096

This PR is an implementation of my own paper:

Nath, V., Schilling, K. G., Parvathaneni, P., Hansen, C. B., Hainline, A. E., Huo, Y., ... & Stepniewska, I. (2019). Deep learning reveals untapped information for local white-matter fiber reconstruction in diffusion-weighted MRI. Magnetic resonance imaging, 62, 220-227.

This is currently work in progress but I am sharing to get feedback.

PR #2520, opened 2022-02-17
Resdnn inference

frheault:
This PR is replacing #2123 and contains only the inference, no training.

It contains a model for [1] to predict fODF from raw DWI signal (as SH). It is not a local model like CSD or DTI, but the class has a fit function to make it easy to use on whole volumes.

This PR contains a fetcher for the weights, the predict/fit functions in the class, an example and simple tests.

[1] Nath, V., Schilling, K. G., Parvathaneni, P., Hansen, C. B., Hainline, A. E., Huo, Y., ... & Stepniewska, I. (2019). Deep learning reveals untapped information for local white-matter fiber reconstruction in diffusion-weighted MRI. Magnetic resonance imaging, 62, 220-227.

PS: I tried to run pytest, but I had difficulties. How can I efficiently check if the docstring and tests are right and up to the standard?


skoudoro:
Wow, Thank you for doing this!  I started this morning to reshape, rebase, etc #2123.  Should I continue? Otherwise, I can do a PR on your PR for some changes. 

Concerning this PR, you need to put Tensorflow as an optional dependency and add some tests.

skoudoro:
> I tried to run pytest, but I had difficulties. How can I efficiently check if the docstring and tests are right and up to the standard?

Can you tell us more ?

frheault:
@skoudoro I think this PR should be the main one, I removed the training and changed the organization of the files.

As for the test, I can't seem to run pytest on my computer.
I can run `python dipy/nn/test/test_histo_resdnn.py` but `pytest dipy/nn/test/test_histo_resdnn.py` crashes immediately and I get a numpy error about compatibility... (Weird since I can easily run the test code manually)
image

arokem:
@frheault : this is often the sign of a borked numpy installation. What version of numpy are you running there? What happens if you upgrade numpy in that env?

skoudoro:
Also, have you seen my PR in your branch ? https://github.com/frheault/dipy/pull/1 

Just doing some small correction

Otherwise, I agree with @arokem, Can you share your file to see if we reproduce the same issue.

skoudoro:
I do not succeed to generate your issue. When I do `pytest -svv dipy/nn/tests/test_histo_resdnn.py`, all tests succeed, no crash.

Capture d’écran 2022-02-18 à 17 14 07

Also, do not forget to look and merge at some of my corrections in your PR https://github.com/frheault/dipy/pull/1

frheault:
@skoudoro I will merge your PR, but keep the nibabel load/save.
I personally dislike quite a lot the load_nifti and save_nifti functions.
Thank you for all others fixes!

skoudoro:
> I personally dislike quite a lot the load_nifti and save_nifti functions.

Also, the goal is to keep all tutorials consistent, using the same functionality for now.

skoudoro:
> @skoudoro : what version of numpy are you working with?

just created a new env. So `tensorflow==2.8.0`, `numpy==1.22.2`

dipy/nn/histo_resdnn.py
```diff
@@ -0,0 +1,245 @@
...
+        """
+
+        if x_test.shape[-1] != self.sh_size:
+            raise ValueError('Expected input for the provided model weights '
+                             'do not match the declared model ({})'
+                             .format(self.sh_size))
+
+        return self.model.predict(x_test)
+
+    def fit(self, data, gtab, mask=None):
```
skoudoro, line 159: This needs some thinking. I understand what you are trying to do but there is a clear conflict with `fit` function in the model of this module (`dipy.nn.model`). Look [here](https://github.com/dipy/dipy/blob/master/dipy/nn/model.py#L74) 

I know you are in the middle of both worlds (reconstruction vs neural network). However, in our current architecture, the function name  `fit` has two different goals here. 

We need to put some thinking to fix this issue.

dipy/nn/histo_resdnn.py
```diff
@@ -0,0 +1,245 @@
...
+
+    Parameters
+    ----------
+    log_level : str
+        Log level for the HistoResDNN only
+    """
+    logger.setLevel(level=log_level)
+
+
+class HistoResDNN(object):
```
skoudoro, line 40: No need of `(object)`. Can you remove it?

dipy/nn/histo_resdnn.py
```diff
@@ -0,0 +1,245 @@
...
+    """
+    This class is intended for the ResDNN Histology Network model.
+    """
+
+    def __init__(self, sh_order=8, basis_type='tournier07', verbose=False):
+        r"""
+        The model was re-trained for usage with a different basis function
+        ('tournier07') like the proposed model in [1, 2].
+
+        To obtain the pre-trained model, use:
```
skoudoro, line 50: `use::` instead

2 colons to avoid indentation issue in the docstring

dipy/nn/histo_resdnn.py
```diff
@@ -0,0 +1,245 @@
...
+
+            # Removing negative values from the SF
+            sphere = get_sphere('repulsion724')
+            tmp_sf = sh_to_sf(sh=tmp_sh, sphere=sphere,
+                              basis_type=self.basis_type,
+                              sh_order=self.sh_order)
+            tmp_sf[tmp_sf < 0] = 0
+            tmp_sh = sf_to_sh(tmp_sf, sphere, smooth=0.0006,
+                              basis_type=self.basis_type,
+                              sh_order=self.sh_order)
```
skoudoro, line 239: this is strange to me. Why do you need to convert it and then come back?

doc/examples/histology_resdnn.py
```diff
@@ -0,0 +1,158 @@
+# -*- coding: utf-8 -*-
+"""
+.. histology_resdnn:
+
+==========================================
```
skoudoro, line 5: need to have the same size of th title

doc/examples/histology_resdnn.py
```diff
@@ -0,0 +1,158 @@
...
+Local reconstruction using the Histological ResDNN
+==========================================
```
skoudoro, line 7: need to have the same size of the title

doc/examples/histology_resdnn.py
```diff
@@ -0,0 +1,158 @@
...
+
+resdnn_model = HistoResDNN(verbose=True)
+fetch_model_weights_path = get_fnames('histo_resdnn_weights')
+resdnn_model.load_model_weights(fetch_model_weights_path)
+predicted_sh = resdnn_model.fit(data, gtab, mask=mask)
+
+save_nifti('predicted_sh.nii.gz', predicted_sh, affine)
+
+"""
+Preparing the scene using Fury. The ODF slicer expects spherical function
```
skoudoro, line 77: FURY

skoudoro (commented):
Hi @frheault,

Thank for the update. See below for some comments.

FYI, you did not need to remove the doctest/docstring, you just needed to add `+skip not have_tf`.  Look at the example [here](https://github.com/dipy/dipy/blob/master/dipy/viz/projections.py#L58)

dipy/nn/histo_resdnn.py
frheault: I can change predict() to __predict() et rename fit() to predict()?

The current predict function could (should) be considered private anyway, so this could work nicely.

dipy/nn/histo_resdnn.py
frheault: The model does predict negative fodf value (as SF), simply because this was not a constrained included in the original training/model.

But from SH you cannot remove negative value on SF, so you need to evaluate the SH, threshold, then fit new SH again.

Sub-optimal, yes. But necessary (I think)

dipy/nn/histo_resdnn.py
```diff
@@ -0,0 +1,245 @@
...
+            mask = np.sum(data, axis=-1)
+        mask = mask.astype(bool)
+
+        # Extract B0's and obtain a mean B0
+        b0_indices = np.where(gtab.bvals < 1)[0]
+        if not len(b0_indices) > 0:
+            raise ValueError('b0 must be present for DWI normalization.')
+        logger.info('b0 indices found are: {}'.format(b0_indices))
+
+        mean_b0 = data[..., b0_indices]
```
arokem, line 196: This is confusing because it's not mean_b0 yet. Maybe combine with line below as a one-liner? 
```suggestion
        mean_b0 = np.mean(data[..., b0_indices], axis=-1)
```

dipy/nn/histo_resdnn.py
```diff
@@ -0,0 +1,245 @@
...
+        mean_b0 = np.mean(mean_b0, axis=-1)
+
+        # Detect number of b-values and extract a single shell of DW-MRI Data
+        unique_shells = np.sort(np.unique(gtab.bvals))
```
arokem, line 200: Should use https://github.com/dipy/dipy/blob/master/dipy/core/gradients.py#L795 to be tolerant to small differences

dipy/nn/histo_resdnn.py
```diff
@@ -0,0 +1,245 @@
...
+        dw_sh_coef = sf_to_sh(norm_dw_data, h_sphere, smooth=0.0006,
+                              basis_type=self.basis_type,
+                              sh_order=self.sh_order)
+
+        # Flatten and mask the data (N, SH_SIZE) to facilitate chunks
+        ori_shape = dw_sh_coef.shape
+        flat_dw_sh_coef = dw_sh_coef[mask > 0]
+        flat_pred_sh_coef = np.zeros(flat_dw_sh_coef.shape)
+
+        count = len(flat_dw_sh_coef) // 1000
```
arokem, line 225: Do you want to make 1000 something users can control? We have something like that in DTI.

arokem (commented):
Thanks for implementing this. Can't wait to try this out. 

I had a few comments and suggestions

dipy/nn/histo_resdnn.py
skoudoro: sounds like a good plan

dipy/nn/histo_resdnn.py
```diff
@@ -0,0 +1,245 @@
...
+        Parameters
+        ----------
+        data : np.ndarray
+            DWI signal in a 4D array
+        gtab : GradientTable class instance
+            The acquisition scheme matching the data (must contain at least
+            one b0)
+        mask : np.ndarray (optional)
+            Binary mask of the brain to avoid unnecessary computation and
+            unreliable prediction outside the brain.
```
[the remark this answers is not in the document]
frheault: It simply avoids computation for voxel that do not have any SH coefficient.
I will make it clearer in the description.

dipy/nn/histo_resdnn.py
```diff
@@ -0,0 +1,245 @@
...
+            Predicted fODF (as SH). The volume has matching shape to the input
+            data, but with ``(sh_order + 1) * (sh_order + 2) / 2`` as a last
+            dimension.
+
+        """
+        if mask is None:
+            logger.warning('Mask should be provided to accelerate '
+                           'computation, and because predictions are '
+                           'not reliable outside of the brain.')
+            mask = np.sum(data, axis=-1)
```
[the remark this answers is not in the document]
frheault: I changed the description to add
```
Default: Compute prediction only for nonzero voxels (with at least
            one nonzero SH coefficient).
```
If there is noises my line will be the whole volume, if it is already skull-strip it's going to avoid computation where SH are all zeros.

frheault:
@skoudoro I added the `@doctest_skip_parser` at the top of the __init__ function and everything is crashing still because of the `>>>` in the doctest (that needs tensorflow). Did I do it wrong? (last commit added that)

dipy/nn/histo_resdnn.py
```diff
@@ -0,0 +1,260 @@
...
+    @doctest_skip_parser
+    def __init__(self, sh_order=8, basis_type='tournier07', verbose=False):
+        r"""
+        The model was re-trained for usage with a different basis function
+        ('tournier07') like the proposed model in [1, 2].
+
+        To obtain the pre-trained model, use::
+        >>> resdnn_model = HistoResDNN()
+        >>> fetch_model_weights_path = get_fnames('histo_resdnn_weights')
+        >>> resdnn_model.load_model_weights(fetch_model_weights_path)
```
skoudoro, line 56: for each line, the following comment is missing: ` # skip if not have_tf `. 

The decorator `doctest_skip_parser` will catch it and ignore it for the test.

frheault: Thanks, I missed that part!

skoudoro:
Hi @frheault,

Thank you for the update.

Can you add Tensorflow to the coverage matrix of our CI's? You just need to add TensorFlow [here](https://github.com/dipy/dipy/blob/master/.github/workflows/coverage.yml#L16)

skoudoro:
Is there any additional comment for this PR?

doc/examples/histology_resdnn.py
```diff
@@ -0,0 +1,157 @@
...
+from dipy.viz import window, actor
+import numpy as np
+import scipy.ndimage as ndi
+
+
+# Disable oneDNN warning
+os.environ['TF_CPP_MIN_LOG_LEVEL'] = '2'
+
+"""
+This ResDNN model requires single-shell data with b0, the data is fetched
```
Garyfallidis, line 35: with one or more b0s

doc/examples/histology_resdnn.py
```diff
@@ -0,0 +1,157 @@
...
+# Fetch DWI and GTAB
+hardi_fname, hardi_bval_fname, hardi_bvec_fname = get_fnames('stanford_hardi')
+data, affine = load_nifti(hardi_fname)
+data = np.squeeze(data)
+bvals, bvecs = read_bvals_bvecs(hardi_bval_fname, hardi_bvec_fname)
+gtab = gradient_table(bvals, bvecs)
+
+"""
+To accelerate computation a brain mask must be computed. A median-otsu filter
+isolates the main signal and then, using a floodfill, the main 'blob' is
```
Garyfallidis, line 48: Please remove vague description about median_otsu as not necessary. Just say we use `median_otsu` to extract the brain.

doc/examples/histology_resdnn.py
```diff
@@ -0,0 +1,157 @@
...
+picked. The resulting mask is saved for visual inspection.
+"""
+
+b0_indices = np.where(gtab.bvals == 0)[0]
```
Garyfallidis, line 52: ``gtab.b0s_mask`` is the better alternative to doing this.

doc/examples/histology_resdnn.py
```diff
@@ -0,0 +1,157 @@
...
+"""
+Preparing the scene using FURY. The ODF slicer expects spherical function
+and the mean spherical harmonic amplitude is used as background.
+The ODF slicer and the background image are added as actors and a mid-coronal
+slice is selected.
+"""
+
+interactive = False
+sphere = get_sphere('repulsion724')
+predicted_sf = sh_to_sf(sh=predicted_sh, sphere=sphere,
```
Garyfallidis, line 84: There is a better way to do this here. As long as you provide the ``B_matrix`` in ``odf_slicer`` your can load directly the spherical harmonics (the spherical function is then saved as a memory map -- heavily reducing memory usage).

Garyfallidis (changes requested):
Thank you @frheault this is super close to be merged. Please correct the small requested changes. @skoudoro will make a quick PR on top of yours to help with memory reduction for visualization purposes.

skoudoro:
Here is the PR https://github.com/frheault/dipy/pull/2 @frheault, Feel free to merge it.

skoudoro:
Thank you @frheault for this work. All concerns have been addressed so I am going ahead a merge this PR.

PR #2521, opened 2022-02-18
DOC: Use GitHub actions status badge in README

jhlegarreta:
Use GitHub action status badge in README. Remove the legacy CI services
badges.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -5,11 +5,8 @@
 
 |
 
-.. image:: https://travis-ci.com/dipy/dipy.svg?branch=master
-  :target: https://travis-ci.com/dipy/dipy
-
-.. image:: https://dev.azure.com/dipy/dipy/_apis/build/status/dipy.dipy?branchName=master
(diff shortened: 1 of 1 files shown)


skoudoro:
Thank you @jhlegarreta. I just wonder if we should add only the `stable` badge. There are many other badges available the tests are split into 5. 

When we answer that, it is ready to be merged

jhlegarreta:
> Thank you @jhlegarreta. I just wonder if we should add only the stable badge. There are many other badges available the tests are split into 5.

IMO:
- The `stable` is the one that is closest/is the one from which you would cut a release, and the one that would be uploaded to PyPI, and thus the one that users would `pip` install. Thus, I think it is the most meaningful one.
- Having multiple badges for a build is misleading (especially if some builds fail for some reason; see how difficult was to tell whether things were working prior to transitioning to GitHub Actions when half of our builds were failing).
- Too many badges would cluttering the landing page (especially if they refer to essentially the same feature).

PR #2524, opened 2022-02-18
Add concurrency + timeout to Github Actions (GHA)

skoudoro:
This PR adds new attributes to GHA:

- Concurrency:  kill all the remaining jobs on a PR if there is a more recent commit
- Timeout: our test duration is approximately 40-45 min.  I added a hard limit to 90min in case something weird is happening. We should keep the timeout to 6h for the future doc generation matrix.

### DIFF
--- a/.github/workflows/test_template.yml
+++ b/.github/workflows/test_template.yml
@@ -30,6 +30,11 @@ on:
         type: string
         required: false
         default: '["pip", ]'
+      timeout:
+        description: "Maximum duration of jobs before cancellation"
+        type: number
+        required: false
+        default: 90
       depends:
         description: ""
         type: string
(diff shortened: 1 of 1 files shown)


.github/workflows/test_template.yml
```diff
@@ -30,6 +30,11 @@ on:
         type: string
         required: false
         default: '["pip", ]'
+      # timeout:
```
arokem, line 33: Do you mean for this to be commented out?

skoudoro: I wanted to make it a parameter of this template it but does not seem to work. This is a temporary comment until I understand what is going on

skoudoro:
it seems it does not work. I will look deeper a bit later. Closing this PR for now.

arokem:
BTW: Be careful with the tmate trick. Apparently GitHub considers this to be a TOS violation in some cases: https://github.com/mxschmitt/action-tmate/issues/104

skoudoro:
> BTW: Be careful with the tmate trick. Apparently GitHub considers this to be a TOS violation in some cases: https://github.com/mxschmitt/action-tmate/issues/104

Ok, thank you for the info. Sometimes, this is the only way to understand what is going on and debug. 

I will remove this comment from another PR, it will avoid giving ideas to some people.

PR #2528, opened 2022-02-20
TEST: Remove legacy `nose`-related dead testing code

jhlegarreta:
Remove legacy `nose`-related dead testing code: `run_module_suite` is
a `NumPy` testing method to run a test module that uses the `nose`
testing framework under the hood. DIPY switched to `pytest` as its
testing framework in a series of commits merged on Jan 26, 2019, and
with the 51234a4 merge commit.

PR #2526, opened 2022-02-19
TEST: Filter legacy SH bases warnings in tests

jhlegarreta:
Filter legacy SH bases warnings in tests. Warnings are already being
tested, and there is no further need to ensure that these warnings are
raised.

Filtering them avoids having them printed in the standard output, and
thus, unnecessary clutter.

Fixes:
```
workflows/tests/test_tracking.py: 16 warnings
dipy/venv/lib/python3.8/site-packages/dipy/reconst/shm.py:346:
PendingDeprecationWarning:
The legacy descoteaux07 SH basis uses absolute values for negative harmonic degrees.
It is outdated and will be deprecated in a future DIPY release.
Consider using the new descoteaux07 basis by setting the `legacy` parameter to `False`.
    warn(descoteaux07_legacy_msg, category=PendingDeprecationWarning)
```

and
```
direction/tests/test_peaks.py::test_peaks_shm_coeff
dipy/venv/lib/python3.8/site-packages/dipy/reconst/shm.py:302:
PendingDeprecationWarning:
The legacy tournier07 basis is not normalized.
It is outdated and will be deprecated in a future release of DIPY.
Consider using the new tournier07 basis by setting the `legacy` parameter to `False`.
    warn(tournier07_legacy_msg, category=PendingDeprecationWarning)
```

and similar warnings across modules.

Visible, for example, at:
https://github.com/dipy/dipy/runs/5235022769?check_suite_focus=true#step:9:3686

Follow-up of #2482.

PR #2529, opened 2022-02-20
STYLE: Remove unused import statements

jhlegarreta:
Remove unused import statements.

### DIFF
--- a/dipy/reconst/tests/test_eudx_dg.py
+++ b/dipy/reconst/tests/test_eudx_dg.py
@@ -1,10 +1,7 @@
-import warnings
-
 import numpy as np
 import numpy.testing as npt
 
 from dipy.direction.peaks import default_sphere, peaks_from_model
-import dipy.reconst.eudx_direction_getter
 
 
 def test_EuDXDirectionGetter():
--- a/dipy/align/bundlemin.pyx
+++ b/dipy/align/bundlemin.pyx
@@ -5,16 +5,15 @@
 
 import numpy as np
 cimport numpy as cnp
-cimport cython
 
 cimport safe_openmp as openmp
 from safe_openmp cimport have_openmp
 
 from cython.parallel import prange
 from libc.stdlib cimport malloc, free
-from libc.math cimport sqrt, sin, cos
+from libc.math cimport sqrt
 
-from dipy.utils.omp import cpu_count, determine_num_threads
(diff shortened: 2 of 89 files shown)


jhlegarreta:
Will need to be rebased on `master` to fix potential conflicts with #2526 and #2528 once the latter get hopefully merged.

jhlegarreta:
This PEP8 warning https://github.com/dipy/dipy/pull/2529#issuecomment-1046284682 is unrelated to the patch set; many examples do not have the imports at the top of the files. It is raised here because the file was modified here. This has already been controversial in the past (cros-ref #2103), including examples being left out of PEP8 checks, and the current inconsistency across examples. If it needs to be fixed, it should be done for all examples at once in a separate PR.

dipy/reconst/tests/test_eudx_dg.py
```diff
@@ -1,10 +1,7 @@
-import warnings
-
 import numpy as np
 import numpy.testing as npt
 
 from dipy.direction.peaks import default_sphere, peaks_from_model
-import dipy.reconst.eudx_direction_getter
```
arokem, line 7: Weird. The name of this file and the name of the test make no sense.

arokem (commented):
This is great. I think we can merge this, but along the way I noticed something strange (see comment inline). Should we post a separate issue about that?

dipy/reconst/tests/test_eudx_dg.py
jhlegarreta: Thanks for the review @arokem. I do agree that the file name and contents seem not to match closely the `dipy.reconst.eudx_direction_getter.pyx`, which is the one that I guess should be targeted. Not sure about the intent of the current content: maybe the test's contents should be assimilated to the `direction.tests.test_peaks` or maybe was initially targeted to exercise `direction.closest_peak_direction_getter.pyx` (which does not have a dedicated test). To me, the pointed file should be reworked to exercise the EulerDX direction getter. Something similar happens to the `dipy.reconst.test_peak_finding.py` test (should maybe renamed to `test_recspeed.py`). Best would be to address them in a separate PR or to keep track of them through an issue.

skoudoro: Yes, let's do that in a separate PR. I will go ahead and merge this PR. Thank you @jhlegarreta

Issue #1974, opened 2019-09-27
MAP-MRI optimization & validation errors


l-espana:
I haven't found mention of this issue anywhere so apologies if I'm missing something. We're using the basic reconst mapmri example script (with a few changes) on some of our data, but some subjects have the following errors/warnings:

**/usr/local/lib/python3.6/dist-packages/dipy/reconst/mapmri.py:396: RuntimeWarning: invalid value encountered in true_divide
  data_norm = np.asarray(data / data[self.gtab.b0s_mask].mean())
/usr/local/lib/python3.6/dist-packages/dipy/reconst/mapmri.py:412: UserWarning: Optimization did not find a solution
  warn('Optimization did not find a solution')
/usr/local/lib/python3.6/dist-packages/dipy/reconst/mapmri.py:396: RuntimeWarning: divide by zero encountered in true_divide
  data_norm = np.asarray(data / data[self.gtab.b0s_mask].mean())
map.py:133: RuntimeWarning: invalid value encountered in power
  rtopCbRt = rtop**(1/3.0)
map.py:138: RuntimeWarning: invalid value encountered in sqrt
  rtapSqRt = np.sqrt(rtap)**
Preprocessing: 20018
Running model
>> data.shape (256, 256, 30, 96)
>> reconstruction begin time  0.1 min
>> data_small.shape (256, 256, 30, 96)
>> mask_roi.shape (256, 256, 30)
**DENOISE PROCESS
>> No denoising applied
**MAP MODEL
>> MAP model fit, part1 begin time  0.1 min
>> Applying both laplacian regularization and positivity constraint
**ERROR in validate_data: Lower bound at index 1 is greater than upper bound: inf > 1.0000e+20
ERROR in osqp_setup: Data validation returned failure**
>> MAP model fit, part1 completed  3941.4 min
**CALCULATE PROPERTIES
>> Return to Origin Probability
>> Return to Axis Probability
>> Return to Plane Probability
>> Mean Square Displacement

There still appears to be output, but the validation errors seem concerning.

Running Ubuntu 16.04, python 3.6, dipy 0.16.0. I'm not sure what else may be helpful, but any thoughts on this would be much appreciated.

arokem:
[...]

PR #2531, opened 2022-02-21
NF: MAP+ constraints

tomdelahaije:
This request is not ready to be merged, but I could use some feedback. There are definitely places where the code can be improved, and I am not sure if everything is in the right place -- in particular the changes to dipy/data. Tests and documentation are still missing.

This request implements the MAP+ constraints proposed in https://doi.org/10.1016/j.neuroimage.2019.116405. The constraints can be used by setting `positivity_constraint=True` and `pos_radius='infinity'` when fitting the model. The implementation can be tested by following the MAP-MRI example in the documentation (https://dipy.org/documentation/1.4.1./examples_built/reconst_mapmri/#example-reconst-mapmri), and running for example

```
map_model_plus = mapmri.MapmriModel(gtab, radial_order=radial_order,
                                    laplacian_regularization=False,
                                    positivity_constraint=True,
                                    pos_radius='infinity')
```

It is recommended to use the MOSEK solver for speed by setting `cvxpy_solver='MOSEK'`, cf. https://docs.mosek.com/latest/install/installation.html for installation. 

Some points:
- The current way of using the constraints (by setting `pos_radius='infinity'`) is not very intuitive. 
- The maximum value for `radial_order` when using the constraints is currently 10, but this is not checked anywhere.
- The precomputed constraint matrices are stored as CSV files in dipy/data/files, but there is probably a more efficient way of storing/reading them.


skoudoro:
Hi @tomdelahaije,

Thank you for doing this!

we will need some time to read the paper and look at your code before answering your questions / reviewing your code.
So will get back to you ASAP.

@ShreyasFadnavis @RafaelNH @kerkelae @matteofrigo and @rutgerfick @arokem, could you look at this if you find time and give your opinion/feedback? 

Thank you

skoudoro:
Aslo, 

Could you give your opinion on this issue https://github.com/dipy/dipy/issues/1974 @tomdelahaije. 

Thanks!

dipy/data/__init__.py
```diff
@@ -313,3 +313,38 @@ def matlab_life_results():
...
+    References
+    ----------
+    .. [1] Dela Haije et al. "Enforcing necessary non-negativity constraints
+           for common diffusion MRI models using sum of squares programming".
+           NeuroImage 209, 2020, 116405.
+
+    """
+
+    mf = 'hermite_constraint_' + str(radial_order) + '.csv'
+    coo = np.loadtxt(pjoin(DATA_DIR, mf), delimiter=",")
```
skoudoro, line 342: You should check if the file exists in case someone put a wrong radial order. 
Otherwise, you can add a check ` if 0<= radial_order <= 8`

dipy/reconst/mapmri.py
```diff
@@ -136,13 +142,19 @@ def __init__(self,
             optimal weight from the values in the array.
         positivity_constraint : bool,
             Constrain the propagator to be positive.
+        constraint_type : string,
```
skoudoro, line 145: `str, optional` instead

dipy/reconst/mapmri.py
```diff
@@ -136,13 +142,19 @@ def __init__(self,
...
+            If set to 'local', positivity is enforced on a grid determined by
```
skoudoro, line 146: Are there only 2 constraint_types?  if yes, I would recommend using a boolean value instead of a string value. it avoids a lot of issues. your variable can be renamed `use_local_constraint=True`

skoudoro (commented):
Hi @tomdelahaije,

Thank you again for this. Here is a really quick review. Overall, it looks good but I need to play a bit with it locally.

it would be great if someone could look deeper at the math. @arokem? @garyfallidis?

skoudoro:
Also, the tests should be fixed so if you could rebase or merge master, it would be great

dipy/data/__init__.py
tomdelahaije: `radial_order` is checked in reconst/mapmri.py when a MapmriModel object is created, which I believe should be the only time these files are loaded. Should there be an additional check in data/__init__.py? If so, what kind of error should be raised when the file is missing? (None of the other defs in data/__init__.py have such checks.)

skoudoro: Indeed, good point.

However, I think you should add an additional test `if radial_order > 8`  in `data/__init__.py`.

Some people might do `radial_order=10` which will pass the `MapmriModel` test but fail on this function.

you can raise a `ValueError`

skoudoro (commented):
Thank you for the update!

See below my answer to your questions and here the link of the failing mapmri test: https://github.com/dipy/dipy/runs/6163062683?check_suite_focus=true#step:9:3873

doc/examples/reconst_mapmri.py
```diff
@@ -143,136 +146,198 @@
...
+                 cmap=plt.cm.gray)
+
+ax5 = fig.add_subplot(1, 5, 5)
+ax5.set_axis_off()
+divider = make_axes_locatable(ax5)
+cax = divider.append_axes("left", size="10%")
 plt.colorbar(ind, cax=cax)
 
-plt.savefig('MAPMRI_maps_regularization.png')
+plt.savefig('MAPMRI_rtop.png')
```
skoudoro, line 211: Hi @tomdelahaije,

Like we talked about last time, can you move all `Matplotlib` functions in a new module inside `dipy.viz`?

Thank you

doc/examples/reconst_mapmri.py
```diff
@@ -343,33 +414,62 @@
...
+
+ax6 = fig.add_subplot(2, 3, 6, title=r'NG parallel - Global positivity')
+ax6.set_axis_off()
+ng_parallel2 = np.array(mapfit_plus_ng.ng_parallel()[:, 0, :].T, dtype=float)
+ind = ax6.imshow(ng_parallel2, interpolation='nearest', origin='lower',
+                 cmap=plt.cm.gray)
+divider = make_axes_locatable(ax6)
+cax = divider.append_axes("right", size="5%", pad=0.05)
+plt.colorbar(ind, cax=cax)
+
```
skoudoro, line 453: same comment as above

StongeEtienne:
Hi @tomdelahaije and @skoudoro,
I just looked into the PR, and I think some part could be speed-up / improved,
but I might need to better understand some part.

1) If I'm not wrong, those coefficient are in a "sparse matrix" form,
maybe scipy sparse matrix could used instead.
(and for a similar reason, save them with numpy / scipy ".npz",
It would also avoid having  int loaded as float and casted back to int)

2) how complex / long it is to compute those coefficients ?,
because maybe it could be computed on the first run, and afterward saved and re-used.
Similar to the kernel in the Contextual enhancement : https://github.com/dipy/dipy/blob/master/dipy/denoise/enhancement_kernel.pyx

Thank you for your work, 
and I would be happy to discuss and test the code.

tomdelahaije:
Thanks for the feedback @StongeEtienne!

I've addressed point 1, and the code is much cleaner (and a bit faster) because of it :). Regarding point 2, I agree that this would be the preferred way to go, but the computation of the matrices is too involved for me to implement in dipy right now. If people are interested in doing this I'd be happy to help though.

StongeEtienne (commented):
Here is a few comments, 
do not hesitate to ask if something is unclear.

dipy/data/__init__.py
```diff
@@ -313,3 +314,51 @@ def matlab_life_results():
...
+    file += '.npz'
+    path = pjoin(DATA_DIR, file)
+
+    if not exists(path):
+        raise ValueError("Constraints file '" + file + "' not found.")
+
+    try:
+        array = load_npz(path)
+        n, x = array.shape
+        sdp_constraints = [array[i*x:(i+1)*x] for i in range(n//x)]
```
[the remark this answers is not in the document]
tomdelahaije: Thanks for your comments @StongeEtienne!

What do you mean with "0.0 * v"? 

CVXPY does make use of sparsity, and the code runs faster with sparse matrices than with dense matrices. Because `reshape` makes the matrix dense, and because there are no 3D sparse matrix representations, I had to use this list reconstruction approach. Do you have ideas on how to improve this part while keeping the sparse representation?

dipy/reconst/mapmri.py
```diff
@@ -246,33 +262,47 @@ def __init__(self,
...
+            if global_constraints:
+                if radial_order > 10:
+                    self.sdp_constraints = load_sdp_constraints('hermite', 10)
+                    msg = 'Global constraints are currently supported for'
+                    msg += ' radial_order <= 10.'
+                    warn(msg)
+                else:
+                    self.sdp_constraints = load_sdp_constraints('hermite',
+                                                                radial_order)
+            elif not global_constraints:
```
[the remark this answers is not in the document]
tomdelahaije: The `else` with the ValueError should remain I think, in case the input is something other than a string or float/int.

dipy/data/__init__.py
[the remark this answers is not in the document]
StongeEtienne: Sum optimizer, avoid variables multiplied by 0,
I compared with and without dense matrix and the procession (optimisation) time was about the same for order 6.
I will think of something, but the matrix computation is not the "bottleneck" so I dont really mind.

dipy/reconst/mapmri.py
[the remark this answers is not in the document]
StongeEtienne: global_constraints is not a bool True or False ?,
and even if its a string 
if  my_condition:
elif not my_condition:  
is equivalent to an "if else" in python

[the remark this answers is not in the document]
StongeEtienne: it's look like you already corrected this one !
the other ones where fine  (since they are check on string or type)

StongeEtienne:
@skoudoro it's good for me

skoudoro (commented):
I will go ahead and merge this PR if there are no additional comments by this Friday (May 27th).

The tutorial shows multiple warnings that a (beginner) user could be afraid of. They all come from the solver. Is there a way to hide those warnings @tomdelahaije? or maybe to warn the user in the tutorial that they are expected due to some background voxel...

See below the warning in my case. Thank you for your feedback.

```python
/Users/koudoro/miniconda3/envs/dipy-env-38/lib/python3.8/site-packages/cvxpy/reductions/solvers/conic_solvers/scs_conif.py:333: UserWarning: 
    CVXPY has just called the numerical solver SCS (version 2.1.4),
    which could not accurately solve the problem with the provided solver
    options. No value was specified for the SCS option called
    "acceleration_lookback". That option often has a major impact on
    whether this version of SCS converges to an accurate solution.

    We will try to solve the problem again by setting acceleration_lookback = 0.
    To avoid this error in the future we recommend installing SCS version 3.0
    or higher.

    More information on SCS options can be found at the following URL:
    https://www.cvxgrp.org/scs/api/settings.html
    
  warnings.warn(SCS.ACCELERATION_RETRY_MESSAGE % str(scs_version))
/Users/koudoro/miniconda3/envs/dipy-env-38/lib/python3.8/site-packages/cvxpy/problems/problem.py:1333: UserWarning: Solution may be inaccurate. Try another solver, adjusting the solver settings, or solve with verbose=True for more information.
  warnings.warn(
/Users/koudoro/Software/dipy/dipy/core/optimize.py:520: UserWarning: Solver failed to produce an optimum: optimal_inaccurate.
  warnings.warn(msg)
```

tomdelahaije:
> I will go ahead and merge this PR if there are no additional comments by this Friday (May 27th).
> 
> The tutorial shows multiple warnings that a (beginner) user could be afraid of. They all come from the solver. Is there a way to hide those warnings @tomdelahaije? or maybe to warn the user in the tutorial that they are expected due to some background voxel...
> 
> See below the warning in my case. Thank you for your feedback.
> 
> ```python
> /Users/koudoro/miniconda3/envs/dipy-env-38/lib/python3.8/site-packages/cvxpy/reductions/solvers/conic_solvers/scs_conif.py:333: UserWarning: 
>     CVXPY has just called the numerical solver SCS (version 2.1.4),
>     which could not accurately solve the problem with the provided solver
>     options. No value was specified for the SCS option called
>     "acceleration_lookback". That option often has a major impact on
>     whether this version of SCS converges to an accurate solution.
> 
>     We will try to solve the problem again by setting acceleration_lookback = 0.
>     To avoid this error in the future we recommend installing SCS version 3.0
>     or higher.
> 
>     More information on SCS options can be found at the following URL:
>     https://www.cvxgrp.org/scs/api/settings.html
>     
>   warnings.warn(SCS.ACCELERATION_RETRY_MESSAGE % str(scs_version))
> /Users/koudoro/miniconda3/envs/dipy-env-38/lib/python3.8/site-packages/cvxpy/problems/problem.py:1333: UserWarning: Solution may be inaccurate. Try another solver, adjusting the solver settings, or solve with verbose=True for more information.
>   warnings.warn(
> /Users/koudoro/Software/dipy/dipy/core/optimize.py:520: UserWarning: Solver failed to produce an optimum: optimal_inaccurate.
>   warnings.warn(msg)
> ```

I can't reproduce these warnings on my system, and I don't know how to prevent them from showing. I've added a generic remark about potential warnings in the tutorial.

dipy/core/optimize.py
```diff
@@ -342,3 +347,181 @@ def fit(self, X, y):
...
+            Arguments passed to the CVXPY solve method.
+
+        Returns
+        -------
+        h : array (m)
+             Estimated optimum for problem variables $h$.
+        """
+
+        # Compute and set reduced problem parameters
+        X = np.linalg.cholesky(np.dot(design_matrix.T, design_matrix)).T
```
DenebBoito, line 482: Hello, 

Had a look at the code and seems good to me :) 

I just have a suggestion for this line: doing the cholesky decomposition requires the matrix to be strictly positive definite. While this shouldn't be a problem for MAP, other methods (especially those requiring tensor encoded data) might get a rank deficient matrix which is not directly decomposable with cholesky. If the plan is to reuse this code for several other methods, maybe using the more general way of computing the matrix square root would be better. Or in general, a warning should be issued to inform the user that the system is underdetermined and the solution is not unique.

tomdelahaije: Nice catch. I've added a warning in the case of a rank deficient design matrix. I'm not sure if there is a single satisfactory way to handle rank deficiency for other models, so for now I think just raising a warning to the user is the way to go.

skoudoro:
Thank you for this work @tomdelahaije. your first contribution in !

Thanks to @StongeEtienne and @DenebBoito for helping to review.

Merging

PR #2422, opened 2021-07-16
WIP-Adding math in SLR tutorial 

nasimanousheh:
I'm providing more information on Streamline-based Linear Registration.

Issue #2523, opened 2022-02-18
Doc generation failed


skoudoro:
## Description

It is currently impossible to generate the documentation (`make -C . html-after-examples`).  It seems that some docstrings are incorrect. we need to:

- Update all the dependency packages needed for the documentation
- improve citation management
- Fix quite a lot of docstring in different functions: 
- To avoid this problem in a future, Add a GitHub action to generate the documentation without the example
- Add a GitHub action to generate the documentation of the modified/added tutorial.

All of this can be split into multiple PR.

common docstring error:
```
Parameters 
------------ in 
Save `dix` to `fname` as pickle.
... in the docstring of save_pickle in /Users/koudoro/Software/dipy/dipy/io/pickles.py.
  warn(msg)
/Users/koudoro/miniconda3/envs/dipy-env-38/lib/python3.8/site-packages/numpydoc/docscrape.py:434: UserWarning: potentially wrong underline length... 
Examples 
---------- 
/Users/koudoro/miniconda3/envs/dipy-env-38/lib/python3.8/site-packages/numpydoc/docscrape.py:434: UserWarning: potentially wrong underline length... 
See also 
---------- in 
Save `dix` to `fname` as pickle.
... in the docstring of save_pickle in /Users/koudoro/Software/dipy/dipy/io/pickles.py.
  warn(msg)
/Users/koudoro/miniconda3/envs/dipy-env-38/lib/python3.8/site-packages/numpydoc/docscrape.py:434: 
```

```diff
- WARNING: Literal block expected; none found.
- Bullet list ends without a blank line; unexpected unindent.
- ERROR: [numpydoc] While processing docstring for 'dipy.reconst.dki.DiffusionKurtosisFit.kfa'
```

PR #2533, opened 2022-02-21
[FIX] Docstring cleaning: wrong underline length...

skoudoro:
Related to https://github.com/dipy/dipy/issues/2523.

The goal of this PR is to fix the following warning during doc generation:

```diff
- UserWarning: potentially wrong underline length...
```

it also closes #2422


skoudoro:
> FAILED segment/tests/test_mrf.py::test_icm_square - AssertionError

Failing tests are not related. We need to relax/check `test_icm_square`which seems to fail from time to time.

PR #2534, opened 2022-02-22
TEST: Remove unnecessary `main` method definition in tests

jhlegarreta:
Remove unnecessary `main` method definition in tests: `pytest` collects
any method having the `test` prefix, so for all practical purposes the
`main` method was being disregarded.

### DIFF
--- a/dipy/align/tests/test_crosscorr.py
+++ b/dipy/align/tests/test_crosscorr.py
@@ -212,10 +212,3 @@ def test_compute_cc_steps_3d():
         expected[:, :, -radius::, ...] = 0
         actual, energy = cc.compute_cc_backward_step_3d(gradG, factors, radius)
         assert_array_almost_equal(actual, expected)
-
-
-if __name__ == '__main__':
-    test_cc_factors_2d()
-    test_cc_factors_3d()
-    test_compute_cc_steps_2d()
-    test_compute_cc_steps_3d()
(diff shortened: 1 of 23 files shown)


skoudoro:
Not sure about this one. It is quite useful for everyday debugging and does not disturb `pytest`. This is a bit the goal of `if __name__==_main__`. I understand the code cleaning part here, however, to me, it hurts more than it helps.

jhlegarreta:
> Not sure about this one. It is quite useful for everyday debugging and does not disturb pytest. This is a bit the goal of if __name__==_main__. I understand the code cleaning part here, however, to me, it hurts more than it helps.

I do agree that they might be useful to only execute some method while debugging, but think if a few maintainers would change those lines while debugging a given feature and committed the change. Even if this is not a frequent case, it adds an unnecessary change, some other maintainer might request to revert that change, would trigger another round of checks, delays, etc. So additional noise and maintenance burden.

It looks impractical for all purposes and views. I find it misleading to have them, especially for newcomers, and inconsistent, as we find them only for some tests, and calling only some tests (note that having them all is not the solution). In this case, "less is more": less code to maintain, so ease of maintenance.

Removing this also means sticking to standard `pytest` practice. Finally, I'd dare to say that none of the Python scientific computing packages defines `main` methods in their tests (unless I am not aware of some `pytest` paradigm).

skoudoro:
Fair enough, you get a point. you got me with this sentence:

>  Even if this is not a frequent case, it adds an unnecessary change, some other maintainer might request to revert that change, would trigger another round of checks, delays, etc. So additional noise and maintenance burden.

merging. thank you @jhlegarreta

PR #2374, opened 2021-04-22
ROIs visualizer

guaje:
This PR adds a feature to visualize **binary** ROI images as contours in horizon. The new parameters needed to use this feature are:

- `--roi_images` (boolean, optional): Enables the representation of binary images as contours.
- `--roi_colors` (rgb color, optional): Color applied to the contours. The default color is red (1, 0, 0).

Here are some examples using this new parameters:

| dipy_horizon --roi_images non-binary.nii.gz binary1.nii.gz binary2.nii.gz | dipy_horizon --roi_colors 0 0 1 --roi_images non-binary.nii.gz binary1.nii.gz binary2.nii.gz |
|---|---|
|Issue1|Issue2|

Additionally, the parameter `--random_colors` has been modified to provide random colors to the ROIs without altering its original functionality. The new values supported by this param are `tracts` and `rois` for the tractograms and the ROIs respectively.

Here are some examples showing the change:

| dipy_horizon --cluster --random_colors --roi_images non-binary.nii.gz binary1.nii.gz binary2.nii.gz *.trk | dipy_horizon --cluster --random_colors tracts --roi_images non-binary.nii.gz binary1.nii.gz binary2.nii.gz *.trk | dipy_horizon --cluster --random_colors rois --roi_images non-binary.nii.gz binary1.nii.gz binary2.nii.gz *.trk |
|---|---|---|
|Issue1|Issue2|Issue2|

PR #2536, opened 2022-02-24
Random colors fix in horizon

### LINKED ISSUE #2429: random_colors flag in dipy_horizon does not work as before
## Description
When trying to use  `--random_colors` flag in `dipy_horizon` to visualize two bundles I get the following error:

image

I saw that the usage of this argument was extended in #2374. In fact, if I use `--random_colors tracts` it works well so not sure if this is a bug or something expected.

guaje:
This PR fixes #2429 and restores the `--random_colors` functionality introduced in #2374.

Additionally, a new test is included for cases when optional strings have different configurations.


guaje:
@drombas I think this PR might be of your interest and it would be nice if you could test it and review it.

drombas:
Hi @guaje, I tested it with several bundles/masks and worked perfectly. Thanks!

skoudoro (approved):
LGTM ! Thanks @guaje for adding all these tests. it is really convenient. Merging!

PR #2538, opened 2022-02-25
[WIP] Registration public API for multi-slice 2D data

samcoveney:
The functions in dipy/align/_public, which conveniently allow to register entire dwi series, are only designed to work with 3D volumes, i.e. 4D data, in which the acquisitions are in the last dimension.

In cardiac MRI, data is often stored in a nifti file as if it were volumetric, but the third dimension corresponds to slice positions, i.e., [Nx, Ny, num_slices,  num_acq]. These data for different slices were collected in separate acquisitions, so we cannot treat data[:, :, :, 1] as a volume. The registration needs to be performed on each slice separately, and this is a 2D registration problem.

I have adapted the code in dipy/align/_public to work in such a scenario by giving an optional argument "slice_index", which when provided causes the code to perform 2D registration on the given slice.

One tricky catch with 2D registration, is that the images need to be in the same plane. The 3D affine that can specify static_grid2world and moving_grid2world could be checked to ensure this. But then the resulting data would (in general) have 3D coordinates even if the data was effectively 2 dimensional. For this reason, it's easier just to work in pixel coordinates, and enforce that *_grid2world to np.eye(3) (note that this corresponds to 2D data, which is an additional reason for this change).

This pull request is **work in progress**, but it is pretty much complete. I would like some feedback please. There is an example provided.

### DIFF
--- a/dipy/align/_public.py
+++ b/dipy/align/_public.py
@@ -25,7 +25,12 @@ from dipy.align.transforms import (TranslationTransform3D,
                                    RigidTransform3D,
                                    RigidScalingTransform3D,
                                    RigidIsoScalingTransform3D,
-                                   AffineTransform3D)
+                                   AffineTransform3D,
+                                   TranslationTransform2D,
+                                   RigidScalingTransform2D,
+                                   RigidIsoScalingTransform2D,
+                                   RigidTransform2D,
+                                   AffineTransform2D)
 
 
 import dipy.core.gradients as dpg
@@ -74,8 +79,9 @@ def _handle_pipeline_inputs(moving, static, static_affine=None,
                                                  affine=static_affine)
     moving, moving_affine = read_img_arr_or_path(moving,
                                                  affine=moving_affine)
+
     if starting_affine is None:
-        starting_affine = np.eye(4)
+        starting_affine = np.eye(static.ndim + 1)
 
     return static, static_affine, moving, moving_affine, starting_affine
 
@@ -347,6 +353,7 @@ def affine_registration(moving, static,
                         ret_metric=False,
                         moving_mask=None,
                         static_mask=None,
+                        verbosity=1,
                         **metric_kwargs):
     """
     Find the affine transformation between two 3D images. Alternatively, find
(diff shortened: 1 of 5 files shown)


skoudoro:
Hi @samcoveney

>  But I would like some feedback please. To use, simply provide "slice_index" argument and the return values should be registered 2D data.

We are currently in the release process, doing multiple checks, finalizing some important points, etc... I just want to let you know that we will start to look in detail at your PR next week, after the release.

samcoveney:
> Hi @samcoveney
> 
> > But I would like some feedback please. To use, simply provide "slice_index" argument and the return values should be registered 2D data.
> 
> We are currently in the release process, doing multiple checks, finalizing some important points, etc... I just want to let you know that we will start to look in detail at your PR next week, after the release.

Any chance that this could get looked at soon? Let me know if there's anything I can do to help here.

skoudoro:
Hi @samcoveney ,

Apologize for not giving news. The core maintainers (including me) are really focused on this workshop that we are organizing: https://dipy.org/workshops/dipy-workshop-2022 

That's why we were not so active these past 2 weeks. The workshop finishes this Friday, so everything will get back to normal, next week (review, update, etc...).

samcoveney:
Tests are passing on my system, are they failing here for reasons not related to my PR?

skoudoro:
> Tests are passing on my system, are they failing here for reasons not related to my PR?

yes, fixed in #2566

skoudoro:
Hi @samcoveney,

Can you share some data?

I am sure I understand what you are trying to do. I got a general idea, but something is disturbing me. the `slice_index` looks problematic to me. But I need to go deeper into your PR.

Please, let me know where can I find cardiac MRI data. Thanks

samcoveney:
Hi @skoudoro thanks for taking a look.

Open cardiac MRI datasets (for DWI) are hard to come by, which is why I showed the idea using slices of brain data. Can you take a look at the example file that I wrote?

What is the problem with `slice_index` do you think? It is used to select a slice of 2D data e.g. `data[:, :, slice_index, :]` in the case that the array does not really represent volumes.

This may not be the best implementation of course, it was just meant to allow a simple extension of the API. It could be better to tell the user to simply provide an array of shape `[num_x, num_y, 1, num_acq]` instead of requiring the `slice_index` at all?

I appreciate your thoughts on this.

samcoveney:
I've not heard back on this, but I think I will modify the code to take out the `slice_index` argument - if the user provides data with a spatial dimension of size 1, then do 2D registration, otherwise to 3D registration. This seems much tidier. I will finish that off after the Easter weekend.

samcoveney:
I forgot I put this in... We may as well close it.

skoudoro:
I kept it open, because I believe we need to continue this work at some point. 

but no worries, I tagged it to keep it in mind

PR #2539, opened 2022-02-26
Adds utilities for embarassingly parallel loops.

arokem:
As discussed during the sprint on Thursday, I am adding this utility function that provides a way to do a parallelized map operation (we might want to consider the name "parfor" to reflect this more accurate description). Allows choice of engine and backend. 

Still todo: 

- [x] Implement optional imports. 
- [x] Add and test loky backend


arokem:
I believe this is open for review/discussion

dipy/utils/parallel.py
```diff
@@ -0,0 +1,93 @@
...
+import multiprocessing
+from tqdm.auto import tqdm
+from dipy.utils.optpkg import optional_package
+
+joblib, has_joblib, _ = optional_package('joblib')
+dask, has_dask, _ = optional_package('dask')
+
+
+def parfor(func, in_list, out_shape=None, n_jobs=-1, engine="joblib",
+           backend="threading", func_args=[], func_kwargs={},
```
skoudoro, line 11: `locky` backend should be the default. The other backends do not manage well the threading and we have to manage numpy, openmp, blas, etc global variable

skoudoro (commented):
Thank you @arokem for doing this! Nice!  See below for some comments.

dipy/utils/parallel.py
```diff
@@ -0,0 +1,93 @@
...
+    Examples
+    --------
+    >>> def power_it(num, n=2):
+    ...     return num ** n
+    >>> arr = np.arange(100).reshape(10, 10)
+    >>> out = parfor(power_it, arr, n_jobs=2)
+    >>> out[0, 0] == power_it(arr[0, 0]) # doctest: +SKIP
+    """
+    if engine == "joblib":
+        p = joblib.Parallel(
```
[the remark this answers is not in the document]
arokem: Done in da57a1dda

dipy/utils/parallel.py
```diff
@@ -0,0 +1,93 @@
...
+            n_jobs=n_jobs, backend=backend,
+            **kwargs)
+        d = joblib.delayed(func)
+        d_l = []
+        for in_element in in_list:
+            d_l.append(d(in_element, *func_args, **func_kwargs))
+        results = p(tqdm(d_l))
+
+    elif engine == "dask":
+        if n_jobs == -1:
```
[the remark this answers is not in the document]
arokem: Done in da57a1dda

.github/workflows/test_optional.yml
```diff
@@ -17,4 +17,4 @@ jobs:
     with:
       runs-on: '["macos-latest", "windows-latest"]'
       install-type: '["conda"]'
-      extra-depends: scikit-learn pandas statsmodels pytables scipy
\ No newline at end of file
+      extra-depends: scikit-learn pandas statsmodels pytables scipy dask joblib
```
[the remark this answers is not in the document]
arokem: Done in da57a1dda

dipy/utils/parallel.py
arokem: Done in b76ef7e2a

arokem:
Thanks for taking a look @skoudoro! I just added one more engine (https://github.com/ray-project/ray), because why not? :-) 

What do you think about naming? I think that `parfor` is not so accurate. What do you think about `parmap` or `paramap`?

arokem:
It would be fun to do a little benchmarking study comparing all these different engines/backends to each other with different infra settings (e.g., cloud, big workstation, laptop) and different problem settings (e.g., SFM vs. DTI vs. MSMT).

arokem:
FYI @36000 : I am working to upstream this code to DIPY.

36000:
Looks good, let me know if you need my help with anything.

ShreyasFadnavis:
> It would be fun to do a little benchmarking study comparing all these different engines/backends to each other with different infra settings (e.g., cloud, big workstation, laptop) and different problem settings (e.g., SFM vs. DTI vs. MSMT).

Even Patch2Self -- Could be very beneficial 👍🏽 Will try it out shortly 💯

dipy/utils/tests/test_parallel.py
```diff
@@ -0,0 +1,37 @@
...
+        engines.append("joblib")
+    if para.has_ray:
+        engines.append("ray")
+
+    my_array = np.arange(100).reshape(10, 10)
+    i, j = np.random.randint(0, 9, 2)
+    my_list = list(my_array.ravel())
+    for engine in engines:
+        for backend in ["threading", "multiprocessing"]:
+            npt.assert_equal(para.parfor(power_it,
```
skoudoro, line 25: Why do you test only 1 pixel and not the whole array?

skoudoro (commented):
To me, this PR is ready to go. Thank you @arokem for this.

This is a good first version. In a future PR, we need to:
- Update this function name (`parfor` to ?)
- Update the tests. We are currently not testing the performance improvement. The CI machine has 2 cores, so we should be able to compare 1core vs 2core.
- Pay attention to backend management. With `joblib`, we should avoid threading and multiprocessing and use only locky. I do not know for Dask backend

skoudoro:
It is also interesting to see that `test_omp.py::test_default_threads` is failing in this PR. This needs to be fixed. 

I suppose, installing one of the parallel packages (dask, joblib, ray) has an impact on `OMP_NUM_THREADS` environment variable. It would be good to know which one has updated this variable.

arokem:
Yep! Good point. I just pushed a commit that splits the parallelization test into three separate workflows to test that.

dipy/utils/tests/test_parallel.py
arokem: I am not sure 😄 

Changed in 807ac1f1d to test the whole array.

skoudoro:
We got our answer! Joblib and Ray seems to update `OMP_NUM_THREADS` environment variable. 

Good to know.... To make easy, I propose to just skip the test `test_omp.py::test_default_threads` if `has_joblib or has_ray` are present.  We can't put a number since it will depend on each machine.

ShreyasFadnavis:
Hi @arokem,

Thanks a lot for implementing the parallel system within DIPY! I am quite excited about this 🙂 
Patch2Self is quite parallelizable, especially here: https://github.com/dipy/dipy/blob/master/dipy/denoise/patch2self.py#L313

How do you suggest I can do this? This line essentially calls patch2self on each volume sequentially, but does not need to be sequential!

```
parfor(_vol_denoise, train_dwi,  model, data_dwi.shape, alpha=alpha, # args from P2S
            engine='loky', backend='multiprocessing') # args from parfor
```

Does that look correct? Sorry if I misunderstood something! Also for this task, what engine do you recommend for a standard i7 Windows machine?

skoudoro (approved):
Great! Thank you for the update @arokem!

This first version looks great. I will go ahead and merge it as soon as the CI's come green.

If anyone has a comment or objection, please, let us know.

dipy/utils/tests/test_omp.py
```diff
@@ -27,6 +30,7 @@ def test_set_omp_threads():
         assert_equal(cpu_count(), 1)
 
 
+@pytest.mark.skipif(has_joblib or has_ray)
```
skoudoro, line 33: Arf, I reviewed too fast.  There is a issue here and CI's are failing because of that. it is missing a reason:

```diff
- Error evaluating 'skipif': you need to specify reason=STRING when using booleans as conditions.
```

arokem: Fixed in e0986ecd3

arokem:
@ShreyasFadnavis : I think that the call should look something like: 

```
paramap(_vol_denoise, 
                train_dwi.reshape((-1, train_dwi.shape[-1]),  # Input reshaped so that the function iterates over the first dimension
                out_shape=train_dwi.shape,  # But we can make sure to reshape the results on the way out
                engine="joblib", backend='loky', # Engine is one of: "joblib", "dask", "ray". "loky" is a backend (the default backend) for "joblib" 
                func_args=[model, data_dwi.shape], # These will be passed as arguments to the function on each call
                func_kwargs={alpha=alpha}, # These will be passed as kwargs to the function on each call
                )
```

The last `**kwargs` in the function signature is reserved for passing keyword arguments to `joblib.Parallel` or `dask.compute` (and maybe also to one of the ray functions? I am not sure which it should be).

arokem:
@ShreyasFadnavis : regarding this: 

> Also for this task, what engine do you recommend for a standard i7 Windows machine?

I don't know! But it would be a nice experiment to do (I think we could probably write a computer science paper about that experiment).

Garyfallidis:
Okay I am merging this. Thanks @arokem. Next phase will require quite a bit more testing to see what works best and for which modules/algorithms. In our last test (a couple of days ago) we still see quite a bit of over-subscription with threads and multiprocessing.

Issue #2543, opened 2022-03-02
StatefulTractogram.are_compatible compare data_per_point keys as list instead of set


NotaCS:
## Description
Using `StatefulTractogram.are_compatible(sft1, sft2)` to check if two sft are compatible returns `False` if the keys in data_per_point are the same but not in the same order.

## Easy fix
Line 160 of dipy/io/stateful_tractogram.py
Change
`if sft_1.get_data_per_point_keys() != sft_2.get_data_per_point_keys():`
to
`if set(sft_1.get_data_per_point_keys()) != set(sft_2.get_data_per_point_keys()):`

(and it's probably the same with lines 164-165 `sft_1.get_data_per_streamline_keys() != sft_2.get_data_per_streamline_keys()`)

NotaCS:
The fix proposed by @frheault works well in my case (don't forget to modify the docstrings too).

However, I'm afraid that it might break some codes if people were indexing the results from get_data_per_point_keys and get_data_per_streamline_keys. I'm not sure why someone would want to do that, but who knows...

I do think it makes more sense that they return a set (as is proposed in the fix), though. It is a choice between retro-compatibility (maybe) and proper type for a given object. Your call.

skoudoro:
Closed by https://github.com/dipy/dipy/pull/2545

PR #2545, opened 2022-03-03
Stateful Tractogram DPS and DPP keys ordering

frheault:
Fix for #2543

### DIFF
--- a/dipy/io/stateful_tractogram.py
+++ b/dipy/io/stateful_tractogram.py
@@ -393,11 +393,11 @@ class StatefulTractogram(object):
 
     def get_data_per_point_keys(self):
         """ Return a list of the data_per_point attribute names """
-        return set(self.data_per_point.keys())
+        return list(set(self.data_per_point.keys()))
 
     def get_data_per_streamline_keys(self):
         """ Return a list of the data_per_streamline attribute names """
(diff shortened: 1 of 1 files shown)


dipy/io/stateful_tractogram.py
```diff
@@ -393,11 +393,11 @@ def data_per_streamline(self, data):
 
     def get_data_per_point_keys(self):
         """ Return a list of the data_per_point attribute names """
-        return list(self.data_per_point.keys())
+        return set(self.data_per_point.keys())
```
skoudoro, line 396: can you use ` list(set(...))` to keep the same behavior as before

dipy/io/stateful_tractogram.py
```diff
@@ -393,11 +393,11 @@ def data_per_streamline(self, data):
...
 
     def get_data_per_streamline_keys(self):
         """ Return a list of the data_per_streamline attribute names """
-        return list(self.data_per_streamline.keys())
+        return set(self.data_per_streamline.keys())
```
skoudoro, line 400: same comment as above: can you use ` list(set(...))` to keep the same behavior as before

skoudoro (approved):
Thank you for the update. LGTM, I will merge it as soon as the CI's are happy

PR #2547, opened 2022-03-04
[FIX] option to disable multithreading while multiprocessing

### LINKED ISSUE #2519: Provide an option to disable multithreading while multiprocessing
When we are using multiprocessing for example in ``peaks_from_model`` we can setup the number of processes (for example at 4) but because some of the code will use multi-threading through numpy (Openblas, MKL) we end up using more cores than necessary (for example 12 instead of 4). This can end up delaying execution or just use too many resources freezing the computer. 

A solution for this is to disable multithreading when multiprocessing. This can be done using the following snippet:

```
import os
os.environ['OPENBLAS_NUM_THREADS'] = '1'
os.environ['MKL_NUM_THREADS'] = '1'
```

I suggest that we set the values above before multiprocessing starts and set them back to default after the work is done. In that way we will not affect execution to functions that do use multithreading.

skoudoro:
This PR fixes #2519. It permits to avoid thread oversubscription. In the close future, we need to improve the design and insert it in the parallel module. Small Notes: the `loky` backend from `joblib` solve this issue. This fix might be useful for the other `joblib` backends

### DIFF
--- a/dipy/utils/multiproc.py
+++ b/dipy/utils/multiproc.py
@@ -1,5 +1,6 @@
 """Function for determining the effective number of processes to be used."""
 
+import os
 from multiprocessing import cpu_count
 from warnings import warn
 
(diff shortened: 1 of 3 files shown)


dipy/utils/multiproc.py
```diff
@@ -40,3 +41,40 @@ def determine_num_processes(num_processes):
...
+    many threads when you use multiprocessing module.
+
+    See Also
+    --------
+    ``enable_np_threads``
+    """
+    current_openblas = os.environ.get('OPENBLAS_NUM_THREADS', '')
+    current_mkl = os.environ.get('MKL_NUM_THREADS', '')
+
+    # import ipdb; ipdb.set_trace()
```
arokem, line 61: Remove debugging statement

dipy/utils/multiproc.py
```diff
@@ -40,3 +41,40 @@ def determine_num_processes(num_processes):
         return 1
 
     return num_processes
+
+
+def disable_np_threads():
+    """Reduce OPENBLAS and MKL thread numbers.
```
arokem, line 47: Do you want to make the number of threads an argument to this function? Right now, this always reduces to 1, but maybe sometimes we want to reduce down to number of cpus?

arokem (commented):
I wonder whether it wouldn't be more elegant to implement this as an object that stores `current_openblas` and `current_mkl` as state, rather than storing those in environment variables.

skoudoro:
> I wonder whether it wouldn't be more elegant to implement this as an object that stores current_openblas and current_mkl as state, rather than storing those in environment variables.

Not a bad idea. I will look into it. Also, I will turn this into a work in progress. It needs testing and exploration.  Thank you for the feedback!

PR #2558, opened 2022-03-08
BUG: Fix errant warning about starting_affine

larsoner:
Over in MNE-Python we are getting warnigs (that our test suite [turns into errors](https://github.com/mne-tools/mne-python/runs/5470293960?check_suite_focus=true)) of the form:
```
mne/transforms.py:1724: in _compute_volume_registration
    moved_zoomed, reg_affine = affine_registration(
/opt/hostedtoolcache/Python/3.10.2/x64/lib/python3.10/site-packages/dipy/align/_public.py:491: in affine_registration
    warn(wm, UserWarning)
E   UserWarning: starting_affine overwritten by centre_of_mass transform
```
This is true even though our call passes `starting_affine=None`.

I think this is due to #2512. Not 100% sure this is the right fix, but it seems reasonable. `_handle_pipeline_inputs` returns `np.eye(4)` for `starting_affine`, and it seems more invasive to make that return `None` (which would also avoid the warning at least)...

### DIFF
--- a/dipy/align/_public.py
+++ b/dipy/align/_public.py
@@ -489,7 +489,7 @@ def affine_registration(moving, static,
 
             if starting_affine is not None and starting_was_supplied:
                 wm = "starting_affine overwritten by center_of_mass transform"
-                warn(wm, UserWarning)
+                warn(wm, UserWarning, stacklevel=2)
 
             # multiply images by masks for transform_centers_of_mass
             static_masked, moving_masked = static, moving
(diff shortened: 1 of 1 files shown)


skoudoro (approved):
Ok, make sense and I confirm this is due to https://github.com/dipy/dipy/pull/2512. 

It seems reasonable to me, so, I will go ahead and merge it.

Thanks @larsoner

PR #2561, opened 2022-03-10
[FIX] Motion correction tutorial

skoudoro:
this PR fixes the last failing tutorials.

### DIFF
--- a/dipy/workflows/tests/test_reconst_csa_csd.py
+++ b/dipy/workflows/tests/test_reconst_csa_csd.py
@@ -134,9 +134,4 @@ def reconst_flow_core(flow):
             reconst_flow._force_overwrite = True
             reconst_flow.run(data_path, bval_path, bvec_path, mask_path,
                              out_dir=out_dir,
-                             parallel=True, num_processes=None)
-            reconst_flow = flow()
-            reconst_flow._force_overwrite = True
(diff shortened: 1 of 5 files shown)


dipy/workflows/tests/test_reconst_csa_csd.py
```diff
@@ -132,11 +132,6 @@ def reconst_flow_core(flow):
             # test parallel implementation
             reconst_flow = flow()
             reconst_flow._force_overwrite = True
-            reconst_flow.run(data_path, bval_path, bvec_path, mask_path,
-                             out_dir=out_dir,
-                             parallel=True, num_processes=None)
-            reconst_flow = flow()
-            reconst_flow._force_overwrite = True
```
arokem, line 139: What's the issue with this test?

skoudoro: 1. Not really a test.
2. avoid using all processors to freeze the machine. since there is already one with `num_processes=2`, no need of this one

skoudoro:
Thank you for the review @arokem. Feel free to merge it. I would love to restart doc generation tonight with those fixes.

dipy/workflows/tests/test_reconst_csa_csd.py
arokem: Makes sense! Off we go!

PR #2568, opened 2022-03-29
DOC: Fix the docstring of `write_mapping`

### LINKED ISSUE #2567: This should be a `1`
Looking through this code, I noticed that the docstring is wrong. Posting this issue so that I don't forget to come back and fix it.

https://github.com/dipy/dipy/blob/3687ce9662280869d27480f7752ad663fefc6b28/dipy/align/_public.py#L238

arokem:
Closes #2567.

### DIFF
--- a/dipy/align/_public.py
+++ b/dipy/align/_public.py
@@ -233,9 +233,9 @@ def write_mapping(mapping, fname):
 
     Notes
     -----
-    The data in the file is organized with shape (X, Y, Z, 2, 3, 3), such
-    that the forward mapping in each voxel is in `data[i, j, k, 0, :, :]` and
-    the backward mapping in each voxel is in `data[i, j, k, 0, :, :]`.
+    The data in the file is organized with shape (X, Y, Z, 3, 2), such
+    that the forward mapping in each voxel is in `data[i, j, k, :, 0]` and
+    the backward mapping in each voxel is in `data[i, j, k, :, 1]`.
 
     """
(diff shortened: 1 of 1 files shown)


dipy/align/_public.py
```diff
@@ -233,9 +233,9 @@ def write_mapping(mapping, fname):
 
     Notes
     -----
-    The data in the file is organized with shape (X, Y, Z, 2, 3, 3), such
-    that the forward mapping in each voxel is in `data[i, j, k, 0, :, :]` and
-    the backward mapping in each voxel is in `data[i, j, k, 0, :, :]`.
+    The data in the file is organized with shape (X, Y, Z, 3, 2), such
+    that the forward mapping in each voxel is in `data[i, j, k, :, 0]` and
```
skoudoro, line 237: should not be `data[i, j, k, 3, 0]` instead since you define 3 above?

skoudoro: I mean `3` instead of `:`

arokem: I might be misunderstanding your question, but there are 3 components to the mapping (dx, dy, dz) so the full forward mapping is data[:, :, :, :, 0] and has dimensions `(X,Y,Z,3)`

skoudoro: yes, got it, I should go to sleep... :-)

skoudoro:
Failing CI's not related, merging

PR #2575, opened 2022-04-11
[WIP]  Define curvature and stepsize as default parameter instead of max_angle for tractography

skoudoro:
This PR starts the refactoring/ Redesign of our tracking API.

### DIFF
--- a/dipy/direction/closest_peak_direction_getter.pyx
+++ b/dipy/direction/closest_peak_direction_getter.pyx
@@ -62,8 +62,11 @@ cdef int closest_peak(cnp.ndarray[cnp.float_t, ndim=2] peak_dirs,
 cdef class BasePmfDirectionGetter(DirectionGetter):
     """A base class for dynamic direction getters"""
 
-    @deprecated_params('max_angle', since='1.14', until='1.15',
-                       alternative='use min_curvature instead.')
+    @deprecated_params('max_angle', since='1.5', until='1.99',
+                       alternative='use min_curvature instead. To convert your '
+                                   'max_angle to min_curvature, use '
+                                   'max_angle_from_curvature function in '
+                                   '``dipy.tracking.utils``')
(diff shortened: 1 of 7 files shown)


dipy/direction/closest_peak_direction_getter.pyx
```diff
@@ -60,14 +63,22 @@ cdef int closest_peak(cnp.ndarray[cnp.float_t, ndim=2] peak_dirs,
 cdef class BasePmfDirectionGetter(DirectionGetter):
     """A base class for dynamic direction getters"""
 
-    def __init__(self, pmf_gen, max_angle, sphere, pmf_threshold=.1, **kwargs):
+    @deprecated_params('max_angle', since='1.14', until='1.15',
+                       alternative='use min_curvature instead.')
+    def __init__(self, pmf_gen, max_angle, sphere, min_curvature=None,
+                 step_size=None, pmf_threshold=.1, **kwargs):
```
arokem, line 69: Do we need to handle the case where both max_angle and min_curvature are provided? At least warn the user that `min_curvature` takes precedence?

arokem (commented):
This looks like it's heading in the right direction. A couple of thoughts:  Do we want to give users the option to choose if they'd rather define this in terms of `max_angle` or `min_curvature`? This would mean that we would not deprecate the `max_angle` parameter, but we might throw an error if both are provided (?). An advantage of this approach is that it would be fully backward compatible. A disadvantage is that it makes the API more complicated and potentially raises some user questions. Either way, we might want to consider warning users about the consequences of passing both of these together (see inline comment).

dipy/direction/closest_peak_direction_getter.pyx
skoudoro: yes, we should warn the user. I agree

skoudoro:
For now, closing this PR. a new approach will be proposed later this month.

PR #2579, opened 2022-04-14
Generalized PCA to less than 3 spatial dims

samcoveney:
Generalizes PCA to work when some spatial dimensions have size 1 (i.e. the data array is 4D, but the data is really 3D -- two spatial dimensions * number of acquisitions). 

I have compared the results from this code with the results obtained by padding 3D data with repeated copies of the dimension of size 1, in this case the third dimension:

```
# pad the array with copies of dimension 3
tmp_1 = np.expand_dims(data, 2)[:, :, [0, 0, 0, 0, 0, 0, 0], :]
sigma_1 = pca_noise_estimate(tmp_1, gtab, patch_radius=3, correct_bias=True, smooth=3)

# dimension 3 is size 1
tmp_2 = np.expand_dims(data, 2)
sigma_2 = pca_noise_estimate(tmp_2, gtab, patch_radius=3, correct_bias=True, smooth=3)

# denoise the padded array
denoised_arr_1 = localpca(tmp_1, sigma_1, tau_factor=2.3, patch_radius=3)
denoised_arr_1 = denoised_arr_1[:, :, 3, :]

# denoise the dim3 = 1 array
denoised_arr_2 = localpca(tmp_2, sigma_2, tau_factor=2.3, patch_radius=[2, 2])
denoised_arr_2 = np.squeeze(denoised_arr_2)
```

The results are identical. This would suggest that there is no need to change this code, as 2D data can just be supplemented by padding dimensions. However, this would be misleading, as it would artificially increase the number of samples (even though the new voxels are just copies of existing ones). See related discussion #2578. Therefore, it is preferable to not use these functions with 2D data by simply padding the array along the third dimension, since the user should be warned about the size of `patch_radius` in relation to the number of acquisitions. I have also changed the code to only raise a warning if there are fewer samples (patch_size) than dimensions (number of acquisitions) and adjusted the tests accordingly (see  #2578)

### DIFF
--- a/dipy/denoise/localpca.py
+++ b/dipy/denoise/localpca.py
@@ -1,3 +1,4 @@
+from warnings import warn
 import numpy as np
 try:
     from scipy.linalg.lapack import dgesvd as svd
@@ -51,14 +52,16 @@ def _pca_classifier(L, nvoxels):
 
 
 def genpca(arr, sigma=None, mask=None, patch_radius=2, pca_method='eig',
-           tau_factor=None, return_sigma=False, out_dtype=None):
+           tau_factor=None, return_sigma=False, out_dtype=None,
+           suppress_warning=False):
     r"""General function to perform PCA-based denoising of diffusion datasets.
 
     Parameters
     ----------
     arr : 4D array
         Array of data to be denoised. The dimensions are (X, Y, Z, N), where N
-        are the diffusion gradient directions.
+        are the diffusion gradient directions. The first 3 dimension must have
+        size >= 2 * patch_radius + 1 or size = 1.
     sigma : float or 3D array (optional)
         Standard deviation of the noise estimated from the data. If no sigma
         is given, this will be estimated based on random matrix theory
@@ -93,12 +96,15 @@ def genpca(arr, sigma=None, mask=None, patch_radius=2, pca_method='eig',
     out_dtype : str or dtype (optional)
         The dtype for the output array. Default: output has the same dtype as
         the input.
+    suppress_warning : bool (optional)
+        If true, suppress warning caused by patch_size < arr.shape[-1].
+        Default: False.
 
     Returns
     -------
     denoised_arr : 4D array
         This is the denoised array of the same size as that of the input data,
-        clipped to non-negative values
+        clipped to non-negative values.
 
     References
     ----------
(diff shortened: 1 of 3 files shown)


samcoveney:
Still seeing a lot of failures on tests, not related to my PR at all

dipy/denoise/localpca.py
```diff
@@ -145,16 +150,26 @@ def genpca(arr, sigma=None, mask=None, patch_radius=2, pca_method='eig',
...
+        raise ValueError("Cannot have only 1 sample, please increase patch_radius.")
+    if num_samples < arr.shape[-1]:
+        tmp = np.sum(patch_size == 1)  # count spatial dimensions with size 1
+        if tmp == 0: root = np.ceil(arr.shape[-1] ** (1./3))  # 3D
+        if tmp == 1: root = np.ceil(arr.shape[-1] ** (1./2))  # 2D
+        if tmp == 2: root = arr.shape[-1]  # 1D
+        root = root + 1 if (root % 2) == 0 else root  # make odd
+        spr = int((root - 1) / 2)  # suggested patch_radius
+        e_s = "Number of samples {1} < Dimensionality {0}. ".format(arr.shape[-1], num_samples)
+        e_s += "Consider increasing patch_radius to {0}.".format(spr)
```
skoudoro, line 171: Nice!

dipy/denoise/localpca.py
```diff
@@ -145,16 +150,26 @@ def genpca(arr, sigma=None, mask=None, patch_radius=2, pca_method='eig',
...
+        warn(e_s, UserWarning)
```
skoudoro, line 172: I still wonder whether we should raise an error. Beginner and intermediate users will experiment with patch radius without understanding the consequence. 

We prefer to stop the process, propose a solution, and let the user think.. If we allow the process to run a very long time, we might lose the user or the user come back to us with: "it does not work".

skoudoro (commented):
Hi @samcoveney,

Thank you for this enhancement!  This generalization is very appreciated!

Overall, it looks good. I just have a few comments below. 

Thanks in advance for addressing them. 

ps: CI's have been fixed so can you merge master or rebase this PR. Thank you

skoudoro:
Also, there is some [pep8 issue to address](https://github.com/dipy/dipy/pull/2579#issuecomment-1099306032)
if you could look at them

dipy/denoise/localpca.py
samcoveney: I have included an argument "suppress_warning" to genpca(), which I have not used yet. It was place-holding while we decided what to do concerning this error or warning. We could change it to 'suppress_error' for example, so that only a warning is supplied if this is the case? That way, a user would be forced to consciously deactivate the error. However...

On the topic of whether this should be an error, I am not completely convinced that an error should be raised rather than a warning. From **Diffusion MRI noise mapping using random matrix theory: Diffusion MRI Noise Mapping**, with M number of voxels and N number of data components:

> However, one rather chooses M > N to avoid the situation when M becomes the dominating factor in the deviation from the MP law, which is of order O(1 / min(M, N))

This is the only mention in the paper about having more voxels than components. With P representing the number of components that can represent the bulk of the signal, the key thing is to ensure is that

> the measurements are highly redundant, P << N. First, small P can be achieved by limiting the diffusion-weighting. Indeed, the complexity of the underlying, generally non-Gaussian diffusion process, and as such P, tends to increase with the b-value. Second, P increases with the size M of the sliding window due to the potential heterogeneity in noise level within the set of voxels.

In the case that more voxels are required than signal directions, we may be introducing complexity due to including more variation of signals, also with different noise levels. I think this is likely to degrade performance much more than using M < N. It is often the case that many repeated acquisitions are performed, so many signals in N have been obtained with the same bvec and bval. This is unlike the case in the paper, which does not contain any repeated acquisitions. Repeated acquisitions will assist in denoising, but in theory they aren't adding much complexity to the data, so we would not expect P to grow much even if N were to grow massively (e.g. 10 repeated acquisitions per bvec and bval).

For these reasons, I believe that it is better to raise a warning, possibly allowing the user to suppress this warning if they find it irritating.

samcoveney: Thanks @skoudoro I have done everything requested, but I have avoided using the term `performance degradation` in the warning - I am not sure that we can accurately make this claim. In my first comment in this pull request, I showed how M > N can be achieved simply by supplying copies of existing data, and yet this is not really adding new data. I can change this if it is a sticking point, but I think we would need to be cautious about our wording

skoudoro: It would be great to add a word concerning performance. So you can use `it might have performance impact` or something similar.

skoudoro (commented):
Thank you for the update @samcoveney. 

Can look at the [failing test](https://github.com/dipy/dipy/runs/6226377845?check_suite_focus=true#step:9:3987) of `pca_denoise_estimate`?

dipy/denoise/localpca.py
samcoveney: That's good wording, I will use that

skoudoro (approved):
Thank you for this @samcoveney.

To me, this PR is ready to go!

I will wait for any additional comments until Thursday before merging this PR

skoudoro:
Thank you for this work @samcoveney, merging!  time for me to come back on registration PR

PR #2580, opened 2022-04-15
DOC: Fix typos and grammar

jhlegarreta:
- DOC: Fix typos and grammar in docstrings
- TEST: Fix typos in test method names
- STYLE: Fix typos in local variable naming

### DIFF
--- a/dipy/direction/tests/test_bootstrap_direction_getter.py
+++ b/dipy/direction/tests/test_bootstrap_direction_getter.py
@@ -21,7 +21,7 @@ response = (np.array([1.5e3, 0.3e3, 0.3e3]), 1)
 
 
 def test_bdg_initial_direction():
-    """This test the number of inital direction."
+    """This tests the number of initial directions."
     """
 
     hsph_updated = HemiSphere.from_sphere(unit_icosahedron).subdivide(2)
--- a/dipy/align/_public.py
+++ b/dipy/align/_public.py
@@ -313,7 +313,7 @@ def resample(moving, static, moving_affine=None, static_affine=None,
         will over-ride the affine that is in the nifti.
 
     between_affine: 4x4 array, optional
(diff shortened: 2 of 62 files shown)


jhlegarreta:
Test errors are unrelated (related to the `dipy.io.bvectxt` being deprecated).

dipy/direction/tests/test_bootstrap_direction_getter.py
```diff
@@ -21,7 +21,7 @@
 
 
 def test_bdg_initial_direction():
-    """This test the number of inital direction."
+    """This tests the number of initial directions."
```
skoudoro, line 24: not sure about this one

skoudoro (commented):
thank you for this @jhlegarreta.

That's great! See below, I just have 1 doubt, otherwise this is good to go

dipy/direction/tests/test_bootstrap_direction_getter.py
skoudoro: meaning `test` singular

arokem: I can confirm that the proposed change is 👍

jhlegarreta: :+1: Strictly speaking, I'd say that test methods do not require docstrings. Additionally, if I'm not mistaken PEP8 discourages starting docstrings with wording like "This method/class" etc. At the same time, if the naming or the test contents are not clear enough (which are indicative of bad naming wrt the acronyms/context of the package, or too many things done in a test), I'd opt for a simple comment like `# Test the number of initial directions`. But that is beyond the scope of this PR.

Thanks both for the reviews.

skoudoro:
Failing CI's are not related, so I am going ahead and merging this.

Thanks @jhlegarreta

PR #2581, opened 2022-04-21
Fix logger in SFT

frheault:
I don't know what happens, I think I accidentally swap lines.

As of right now, this forces the logging to be the default when importing dipy.io.stateful_tractogram (rather than the SFT logger specifically)

And users cannot reset it using the classic `logging.basicConfig(level=logging.INFO)`, but they could if using `logging.getLogger().setLevel(logging.INFO)`.

Anyway, this is fixed now.


skoudoro:
Looks good and the failing CI's are not related, so merging. thanks @frheault

PR #2584, opened 2022-04-27
transform_streamlines changes dtype to float64

frheault:
Keep the same dtype when using transform_streamlines (with array_sequence)

This problem was identified when saving VTK files, it was always double precision.


arokem:
Looks great! Any chance you could add a test that would fail when this change is not in place?

skoudoro (approved):
LGTM, can we go ahead and merge this PR @arokem?

arokem:
Yep. This is GTG!

PR #2589, opened 2022-05-06
Switch from using nibabel InTemporaryDirectory to standard library tmpfile module

### LINKED ISSUE #2587: Thread safety concerns with reliance on `nibabel.tmpdirs.InTemporaryDirectory`
## Description

We hit an issue specifically in the [`read_bvals_bvecs`](https://github.com/dipy/dipy/blob/master/dipy/io/gradients.py#L9) implementation where dipy farms out to nibabel's [`InTemporaryDirectory`](https://github.com/nipy/nibabel/blob/a8382384ca24564719ca9bc03488226733c3f2ec/nibabel/tmpdirs.py)

Reading through the nibabel source code its doing some quite sketchy things like:

```python
self._pwd = os.getcwd()
os.chdir(self.name)
```

which seem quite unfriendly toward multithreaded applications.

dipy isn't doing anything fancy at all -- it just needs a tempfile -- so I think switching to the [standard library way](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryFile) would be a safe change to make

## Way to reproduce

- Python version: 3.7.12
- dipy version:  this impacts from at least 0.16.0 to 1.5.0

I can work on a reproduction but logically since Python threads share a single process and `os.chdir` operates at the process level, conceptually I think its obvious nibabel is making some bad assumptions

jacobdr:
Closes #2587

### DIFF
--- a/dipy/io/gradients.py
+++ b/dipy/io/gradients.py
@@ -1,9 +1,8 @@
-
 from os.path import splitext
 import re
+import tempfile
 import warnings
 import numpy as np
-from nibabel.tmpdirs import InTemporaryDirectory
 
 
 def read_bvals_bvecs(fbvals, fbvecs):
@@ -46,11 +45,10 @@ def read_bvals_bvecs(fbvals, fbvecs):
             with open(this_fname, 'r') as f:
                 content = f.read()
             # We replace coma and tab delimiter by space
-            with InTemporaryDirectory():
-                tmp_fname = "tmp_bvals_bvecs.txt"
-                with open(tmp_fname, 'w') as f:
-                    f.write(re.sub(r'(\t|,)', ' ', content))
-                vals.append(np.squeeze(np.loadtxt(tmp_fname)))
+            with tempfile.NamedTemporaryFile(suffix=".txt", mode="w") as f:
+                f.write(re.sub(r'(\t|,)', ' ', content))
+                f.flush()
+                vals.append(np.squeeze(np.loadtxt(f.name)))
         elif ext == '.npy':
             vals.append(np.squeeze(np.load(this_fname)))
         else:
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,4 @@ dipy/.idea/
 .vscode
 .pytest_cache/*
 tmp/
+venv
--- a/dipy/direction/peaks.py
+++ b/dipy/direction/peaks.py
@@ -1,9 +1,9 @@
+import tempfile
 
 from multiprocessing import Pool
 from itertools import repeat
 from os import path
 
-from nibabel.tmpdirs import InTemporaryDirectory
 
 import numpy as np
 import scipy.optimize as opt
@@ -234,8 +234,7 @@ def _peaks_from_model_parallel(model, data, sphere, relative_peak_threshold,
     indices = list(zip(np.arange(0, n, chunk_size),
                        np.arange(0, n, chunk_size) + chunk_size))
 
-    with InTemporaryDirectory() as tmpdir:
-
+    with tempfile.TemporaryDirectory() as tmpdir:
         data_file_name = path.join(tmpdir, 'data.npy')
         np.save(data_file_name, data)
         if mask is not None:


arokem:
Thanks for doing this. And thanks for pinging the nibabel GitHub about this as well. I am curious to see if there's some reason I don't know that they implemented it this way.

matthew-brett:
> Thanks for doing this. And thanks for pinging the nibabel GitHub about this as well. I am curious to see if there's some reason I don't know that they implemented it this way.

Well - as @effigies says in https://github.com/nipy/nibabel/issues/1102#issuecomment-1120122175 - once you've accepted the premise in the class name - that the code context involves changing to temporary directory - I don't think there's any other way to implement it.  But maybe there is - happy to hear if so.

Or you might ask - why write tests that involve changing the directory?  I guess there are various reasons - just to save some code for prefixing the path to lots of temporary files, and testing a process that creates something in the working directory, were two things I thought of.

dipy/io/gradients.py
```diff
@@ -46,11 +45,10 @@ def read_bvals_bvecs(fbvals, fbvecs):
             with open(this_fname, 'r') as f:
                 content = f.read()
             # We replace coma and tab delimiter by space
-            with InTemporaryDirectory():
-                tmp_fname = "tmp_bvals_bvecs.txt"
-                with open(tmp_fname, 'w') as f:
-                    f.write(re.sub(r'(\t|,)', ' ', content))
-                vals.append(np.squeeze(np.loadtxt(tmp_fname)))
+            with tempfile.NamedTemporaryFile(suffix=".txt", mode="w") as f:
```
jacobdr, line 48: Classic windows nit: https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile

> Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows NT or later).

I'll swap out this implementation

dipy/io/gradients.py
```diff
@@ -1,9 +1,9 @@
-
 from os.path import splitext
 import re
+import tempfile
+from typing import final
```
skoudoro, line 4: Can you remove this import? it seems to not be used. Thanks!

jacobdr: yeesh sorry good cash that was some IDE BS

dipy/core/tests/test_gradients.py
```diff
@@ -426,7 +502,8 @@ def test_reorient_bvecs():
     # atol is set to 1 here to do the scaling verification here,
     # so that the reorient_bvecs function does not throw an error itself
     new_gt = reorient_bvecs(gt, np.array(shear_affines)[:, :3, :3], atol=1)
-    bvecs_close_to_1 = abs(vector_norm(new_gt.bvecs[~gt.b0s_mask]) - 1) <= 0.001
+    abs_normalized_vector = abs(vector_norm(new_gt.bvecs[~gt.b0s_mask]) - 1)
```
jacobdr, line 505: This was a pure style change to get the file to pass pep8

jacobdr:
@skoudoro sorry to take so long to get back to this. Just expanded the test coverage in my most recent commits and we'll see how the tests fare now. I was trying to find a way to not have to write back out to disk for performance reasons but I think that is outside the scope of this ticket... so I kept it pretty low-risk and faithful to the OG implementation

dipy/core/tests/test_gradients.py
```diff
@@ -27,6 +30,79 @@ def test_unique_bvals_deprecated():
...
+    case_4_bval, case_4_bvec = read_bvals_bvecs(
+        fbvals=str(case4_bval_numpy_filename),
+        fbvecs=str(case4_bvec_numpy_filename)
+    )
+    assert case_4_bval.shape == expected_bvals_shape
+    assert case_4_bvec.shape == expected_bvec_shape
+    np.testing.assert_allclose(case_4_bval, case_1_bval)
+    np.testing.assert_allclose(case_4_bvec, case_1_bvec)
+
+    # TODO: Expand this test, not sure what would trigger this
```
jacobdr, line 64: I was trying to write a test for this section -- https://github.com/dipy/dipy/pull/2589/files#diff-6e52b8b6894f91b9bc345dba41c51a89ee4bdd2f60d24b77cab81a1fce42fcefR77 -- but I couldn't figure out an easy way to trigger it. Probably because I suck at numpy..... Open to suggestions

jacobdr:
@matthew-brett @skoudoro can I request a re-review given the changes I made to expand the test coverage please? (also need an approval on the pending GH actions workflow for CI)

dipy/core/tests/test_gradients.py
```diff
@@ -27,6 +30,79 @@ def test_unique_bvals_deprecated():
                       np.array([0, 800, 1400, 1401, 1405]))
 
 
+def test_read_bvals_bvecs(tmp_path: pathlib.Path):
```
skoudoro, line 33: this test needs to be moved to `dipy.io.tests.test_gradients.py`.    See [here](https://github.com/dipy/dipy/blob/master/dipy/io/tests/test_io_gradients.py#L13)

You are currently in the `core` module instead of the `io` module.

I know, it can be confusing, but you are currently duplicating some tests. Sorry for that.

dipy/core/tests/test_gradients.py
skoudoro: I did not look at the details, so sorry if I am wrong.

It seems you are doing the same as [here](https://github.com/dipy/dipy/blob/master/dipy/io/tests/test_io_gradients.py#L128).

skoudoro (commented):
Hi @jacobdr,

Your tests are in the wrong file. See below for additional details.
Thank you in advance for the update.

ps: failing CIs are not related to your PR now, it should be fixed in #2595

skoudoro:
Hi @jacobdr,

This PR is almost done, it would be great if you could rebase this PR and address the last comment.
Please, keep us updated. Thank you!

dipy/core/tests/test_gradients.py
jacobdr: Yeah I just removed all my changes to this file as they are duplicated

dipy/core/tests/test_gradients.py
jacobdr: I double checked and I think every case I wrote pretty much already existed so I removed this file and replaced it with the contents from master

jacobdr:
@skoudoro recent commits:

* merged master
* reverted all changes to test_gradients in deference to `dipy/io/tests/test_io_gradients.py` as you astutely pointed out I had re-invented the wheel. the lack of change in coverage should have tipped me off.

dipy/io/gradients.py
```diff
@@ -45,12 +44,21 @@ def read_bvals_bvecs(fbvals, fbvecs):
...
+            try:
+                # We replace coma and tab delimiter by space
+                replaced_content = re.sub(r'(\t|,)', ' ', content)
+                f.write(replaced_content)
+                f.flush()
+                f.seek(0)
+                numpy_data = np.loadtxt(f)
+                vals.append(np.squeeze(numpy_data))
+            finally:
+                f.close()
```
effigies, line 61: Sorry for hopping in late, but if you can `np.loadtxt()` on a filehandle, why not just use `StringIO`?

```suggestion
            munged_content = io.StringIO(re.sub(r'(\t|,)', ' ', content))
            vals.append(np.squeeze(np.loadtxt(munged_content)))
```

skoudoro:
Thanks @jacobdr for doing this! 

Thanks @matthew-brett @effigies @arokem for the review. 

merging

PR #2590, opened 2022-05-06
Fast Streamline Search algorithm implementation

StongeEtienne:
Implementation of the Fast Streamline Search algorithm.
This code include a "Radius search" adapted for streamlines using scipy cKDTree.
Including an example, and basic testing.


skoudoro:
Thank you for adding this nice algorithm @StongeEtienne! we will look into it come back to you ASAP

dipy/segment/search.py
```diff
@@ -0,0 +1,292 @@
...
+        within the given radius. See [StOnge2021]_ for further details.
+
+        Parameters
+        ----------
+        streamlines : Streamlines
+            Streamlines to generate the tree structure.
+        radius : float
+            Search radius (with MDF / average L2 distance)
+            must be smaller than max_radius when FFS was initialized.
+        use_negative : bool
```
skoudoro, line 106: bool, optional

dipy/segment/search.py
```diff
@@ -0,0 +1,292 @@
...
+
+        # Adapting radius for L1 query: sqrt(3) = 1.73205080756887729..
+        # Rounded up for float32 precision to avoid error / false negative
+        l1_sum_dist = 1.73205081 * radius * self.nb_mpts
+
+        # Search for all similar streamlines
+        list_id = []
+        list_id_ref = []
+        list_dist = []
+        for i in range(len(u_bin)):
```
skoudoro, line 148: enumerate instead

dipy/segment/search.py
```diff
@@ -0,0 +1,292 @@
...
+
+            if self.bidirectional:
+                flipped = ids_ref >= self.ref_nb_slines
+                ids_ref[flipped] -= self.ref_nb_slines
+                if use_negative:
+                    dist[flipped] *= -1.0
+
+            return coo_matrix((dist, (ids_in, ids_ref)),
+                              shape=(q_nb_slines, self.ref_nb_slines))
+        else:
```
skoudoro, line 191: not needed this `else`

dipy/segment/search.py
```diff
@@ -0,0 +1,292 @@
...
+
+
+def mean_l2_func(a, b):
+    """Streamlines distance function: Average L2 (MDF without flip)"""
+    return np.mean(np.sqrt(np.sum((a - b) ** 2, axis=-1)), axis=-1)
+
+
+def mean_l1_func(a, b):
+    """Streamlines distance function: Average L1"""
+    return np.mean(np.sum(np.abs(a - b), axis=-1), axis=-1)
```
skoudoro, line 292: we might want to move these distances functions somewhere else..... I need to look where...

doc/examples/fast_streamline_search.py
```diff
@@ -0,0 +1,202 @@
...
+
+ref_in_reach = np.zeros([len(model_af_l)], dtype=bool)
+ref_in_reach[ids_ref] = True
+
+scene = window.Scene()
+scene.SetBackground(1, 1, 1)
+if np.any(ref_in_reach):
+    scene.add(actor.line(model_af_l[ref_in_reach], colors=(0, 1, 0)))
+if np.any(~ref_in_reach):
+    scene.add(actor.line(model_af_l[~ref_in_reach], colors=(1, 0, 0)))
```
skoudoro, line 176: this might be too advanced for a tutorial. Many users do not understand `any` and even less `~`.  Too many negation, can you find an alternative?

skoudoro (commented):
Hi @StongeEtienne,

Here is a first simple review. I need to test and play with your code a bit. I know that you had more distance functions that are not shown here. is it on purpose? 

I will come back with a more detailed review. Thanks in advance for the first cleaning

dipy/segment/search.py
```diff
@@ -0,0 +1,292 @@
...
+
+        Generate the Binned K-D Tree structure with reference streamlines,
+        using streamlines barycenter and mean-points.
+        See [StOnge2021]_ for further details.
+
+        Parameters
+        ----------
+        ref_streamlines : Streamlines
+            Streamlines (ref) to generate the tree structure.
+        max_radius : float
```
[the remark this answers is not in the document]
StongeEtienne: This is the maximum radius (distance) that will be employed for the search.

doc/examples/fast_streamline_search.py
StongeEtienne: its simply to separate in 2 groups, however it needed a check to avoid empty array.
I changed it to only change the color, it should be simplier now.

StongeEtienne:
@skoudoro Yes, i simplified this first implementation to only use "MDF",
I plan to improve this later on.

I realize that one scipy parameter for the ckdtree (return_sorted=False) is only implemented from scipy 1.6,
should we update the requirements or should we do a try / catch ?

skoudoro:
> I realize that one scipy parameter for the ckdtree (return_sorted=False) is only implemented from scipy 1.6,
should we update the requirements or should we do a try / catch ?

Try catch since our minimum requirement is `scipy 1.1`

dipy/segment/search.py
StongeEtienne: moved into   dipy.segment.metric

skoudoro:
@gabknight and @arokem, can you look at/review this PR? thank you

doc/examples/fast_streamline_search.py
```diff
@@ -0,0 +1,203 @@
+"""
```
gabknight, line 1: Please also add it to `doc/example/valid_examples.txt`

doc/examples/fast_streamline_search.py
```diff
@@ -0,0 +1,203 @@
...
+
+from dipy.data import get_target_tractogram_hcp, get_two_hcp842_bundles
+from dipy.io.streamline import load_trk
+from dipy.segment.fss import FastStreamlineSearch, nearest_from_matrix_row
+from dipy.viz import actor, window
+
+"""
+Download and read data for this tutorial
+"""
+
```
gabknight, line 22: Please add a call to the fetches before in case the data was not downloaded by the user
```
from dipy.data import fetch_bundle_atlas_hcp842, fetch_target_tractogram_hcp                         

fetch_bundle_atlas_hcp842()
fetch_target_tractogram_hcp()
```

doc/examples/fast_streamline_search.py
```diff
@@ -0,0 +1,203 @@
...
+recognized_af_l = streamlines[ids_s]
+
+"""
+let's visualize streamlines similar to the Arcuate Fasciculus Left bundle
+"""
+
+scene = window.Scene()
+scene.SetBackground(1, 1, 1)
+scene.add(actor.line(model_af_l, colors=(0, 1, 0)))
```
gabknight, line 118: I suggest removing the model from this figure

gabknight (changes requested):
Thanks @StongeEtienne. Looks good and seems fast. I added a few comments and suggestions.

dipy/segment/fss.py
```diff
@@ -0,0 +1,291 @@
...
+        """
+        if max_radius <= 0.0:
+            raise ValueError("max_radius needs to be a positive value")
+
+        if resampling < 20:
+            warnings.warn("For accurate results, resampling should be"
+                          " at least >= 10 and preferably >= 20")
+
+        if resampling % nb_mpts != 0:
+            raise ValueError("nb_mpts needs to be a factor of resampling")
```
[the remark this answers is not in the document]
StongeEtienne: It's not really a speedup factor, since it could increase the computation time.
However, it could be a "simplification_factor", but refered as mean-points in the paper.

It also needs to be an exact divisor ex:  resampling = 3 can only have 1 or 3 mean-points
so it wont avoid the check by using a int division "//"

dipy/segment/fss.py
```diff
@@ -0,0 +1,291 @@
...
+
+        self.nb_mpts = nb_mpts
+        self.bin_size = bin_size
+        self.bidirectional = bidirectional
+        self.resampling = resampling
+        self.bin_overlap = max_radius
```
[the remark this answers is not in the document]
StongeEtienne: I can change it, it's just that "bin_overlap" for the user is not really descriptive,
and a good "bin_overlap" from my test, is the "radius" (when it's not to small).

dipy/segment/fss.py
```diff
@@ -0,0 +1,291 @@
...
+                              shape=(q_nb_slines, self.ref_nb_slines))
+
+        # No results, return an empty sparse matrix
+        return coo_matrix((q_nb_slines, self.ref_nb_slines))
+
+    def _resample(self, streamlines):
+        """Resample streamlines"""
+        s = np.zeros([len(streamlines), self.resampling, 3], dtype=np.float32)
+        for i, sline in enumerate(streamlines):
+            if len(sline) < 2:
```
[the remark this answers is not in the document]
StongeEtienne: I tested it, because it fixes the error when havin a signle points (in the example it happens),
it does not link to "zero", it duplicates the points (numpy broadcast)
so it becomes a list of the same points, and my algorithms compares it like the rest (average euclidean distance)

dipy/segment/metric.py
```diff
@@ -39,3 +40,49 @@ def mdf(s1, s2):
...
+
+    Returns
+    -------
+    1D array
+        Distance between each S streamlines
+    """
+    return np.mean(np.sum(np.abs(a - b), axis=-1), axis=-1)
+
+
+def mean_l2_func(a, b):
```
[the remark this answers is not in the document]
StongeEtienne: I could rename it to mean_euclidean_dist()

scipy.spatial.distance.euclidean was made for for comparing 2 vectors (not 2 array of vector)
I could use "linalg norm 2" however we would have to reshape the list of streamlines 
in a list of points and reshape it back for the average.

Overall this is a little bit faster (no check, not a general exponent) otherwise it would look like this
```
diff = a - b
dists = np.linalg.norm(diff.reshape((-1, 3)), axis=-1))
return np.mean(dists.reshape((len(diff), -1)), axis=-1))
```

dipy/segment/metric.py
```diff
@@ -39,3 +40,49 @@ def mdf(s1, s2):
                         vol 6, no 175, 2012.
     """
     return dist(MinimumAverageDirectFlipMetric(), s1, s2)
+
+
+def mean_l1_func(a, b):
```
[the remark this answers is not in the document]
StongeEtienne: I could rename it

doc/examples/fast_streamline_search.py
StongeEtienne: done

doc/examples/fast_streamline_search.py
StongeEtienne: done

doc/examples/fast_streamline_search.py
StongeEtienne: done

dipy/segment/metric.py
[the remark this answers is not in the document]
StongeEtienne: it was renamed

dipy/segment/fss.py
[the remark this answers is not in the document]
StongeEtienne: it was renamed

dipy/segment/metric.py
```diff
@@ -39,3 +40,49 @@ def mdf(s1, s2):
...
+    b : 2D or 3D array
+        A streamline or concatenated streamlines
+        (array of S streamlines by P points in N dimension).
+
+    Returns
+    -------
+    1D array
+        Distance between each S streamlines
+    """
+    return np.mean(np.sum(np.abs(a - b), axis=-1), axis=-1)
```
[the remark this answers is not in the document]
StongeEtienne: done for both distance functions, with "np.linalg.norm"

gabknight (approved):
LGTM. Thanks @StongeEtienne

Garyfallidis:
:+1:  Here too!

skoudoro (commented):
Hi @StongeEtienne,

it looks good to me also.

However, I still wonder why we need a class here. everything seems straight forward and some utilities functions can be easily use outside of `FastStreamlineSeach` like `_slines_barycenters` or  `_slines_mean_points`.

Looking forward for your feedback

StongeEtienne:
Thanks all,
@skoudoro 
It could be done as functions, instead of a class.
However there are 2 reasons describing why I decided to use a class:

1.  Most utility function requires a specific formatting (slines_arr), so I don't want people to use them directly thinking they are generic functions for tractograms.
2.  (the main reason) you can more easily re-use the initialization (tree construction) and run multiple time "radius_search()" with different input streamlines. You can imagine initializing "FastStreamlineSeach" with the CST atlas template and afterward apply the "radius_search()" for each subject tractogram. Thus avoiding recomputing the tree.

This is an opinion, if you have another solution (especially for the second reason), I would be happy to discuss it !

skoudoro:
ok, I see

1. For this point (my main reason)  I need to look closer what is this specific formatting. 
2. if it is just for initializing, a closure will have better performance and it is more appropriate than a class. 

Can you remind me why you need this specific formatting? Apologize for this question.  I will look into it, and if it is mandatory, we can go ahead and merge. Otherwise, if we can standardize this point, I would prefer multiple function + closure than this class.

skoudoro (approved):
Thank you @StongeEtienne  for all the updates. 

This is good to go, I will go ahead and merge this PR

PR #2593, opened 2022-05-08
Embed parallelization into the multi_voxel_fit decorator.

arokem:
I've started playing around with the idea that the multi_voxel_fit decorator could use `paramap` instead of iterating over voxels. If we can make this work generally that would be pretty cool. So far, I've only tested this with the fwdti model, and in that case, the change to the additional changes to the code are rather minimal, which gives me hope that we might be able to use this wherever we use this decorator, so in csd, dsi, forecast, fwdti, gqi, ivim, mapmri, mcsd, qtdmri, and shore (!).

### DIFF
--- a/.github/workflows/test_template.yml
+++ b/.github/workflows/test_template.yml
@@ -97,7 +97,7 @@ jobs:
         channels: defaults, conda-forge
         use-only-tar-bz2: true
     - name: Install HDF5 and pytables on macOS
-      if: runner.os == 'macOS'
+      if: ${{ (runner.os == 'macOS')  &&  (env.EXTRA_DEPENDS != '')  &&  (env.INSTALL_TYPE == 'pip') }}
       run: |
         pip install cython
         brew install hdf5
--- a/.github/workflows/test_template.yml
+++ b/.github/workflows/test_template.yml
@@ -96,6 +96,16 @@ jobs:
         python-version: ${{ matrix.python-version }}
         channels: defaults, conda-forge
         use-only-tar-bz2: true
+    - name: Install HDF5 and pytables on macOS
+      if: runner.os == 'macOS'
+      run: |
+        pip install cython
+        brew install hdf5
+        brew install c-blosc
+        export HDF5_DIR=/opt/homebrew/opt/hdf5
+        export BLOSC_DIR=/opt/homebrew/opt/c-blosc
+        pip install tables
+
     - name: Install Dependencies
       run: |
         if [ "${{ inputs.use-pre }}" == "true" ]; then
--- a/dipy/reconst/multi_voxel.py
+++ b/dipy/reconst/multi_voxel.py
@@ -53,14 +53,13 @@ def multi_voxel_fit(single_voxel_fit):
         vox_per_chunk = np.max([data_to_fit.shape[0] // n_jobs, 1])
         chunks = [data_to_fit[ii:ii + vox_per_chunk]
                   for ii in range(0, data_to_fit.shape[0], vox_per_chunk)]
-        fit_array[mask] = np.squeeze(
+        fit_array[mask] = np.concatenate((
             paramap(
                 parallel_fit_worker,
                 chunks,
                 func_args=[single_voxel_with_self],
-                out_shape=fit_array[mask].shape,
                 **kwargs)
-                )
+                ))
         return MultiVoxelFit(self, fit_array, mask)
 
     return new_fit
--- a/dipy/reconst/multi_voxel.py
+++ b/dipy/reconst/multi_voxel.py
@@ -1,4 +1,5 @@
 """Tools to easily make multi voxel models"""
+
 from functools import partial
 import multiprocessing
 
@@ -31,6 +32,7 @@ def multi_voxel_fit(single_voxel_fit):
     """Method decorator to turn a single voxel model fit
     definition into a multi voxel model fit definition
     """
+
     def new_fit(self, data, mask=None, **kwargs):
         """Fit method for every voxel in data"""
         # If only one voxel just return a normal fit
@@ -49,10 +51,12 @@ def multi_voxel_fit(single_voxel_fit):
         # Default to serial execution:
         engine = kwargs.get("engine", "serial")
         if engine == "serial":
-            bar = tqdm(total=np.sum(mask), position=0,
-                       disable=kwargs.get("verbose", True))
-            bar.set_description("Fitting reconstruction model",
-                                "using serial execution")
+            bar = tqdm(
+                total=np.sum(mask), position=0, disable=kwargs.get("verbose", True)
+            )
+            bar.set_description(
+                "Fitting reconstruction model", "using serial execution"
+            )
             for ijk in ndindex(data.shape[:-1]):
                 if mask[ijk]:
                     fit_array[ijk] = single_voxel_fit(self, data[ijk])
@@ -63,17 +67,22 @@ def multi_voxel_fit(single_voxel_fit):
             single_voxel_with_self = partial(single_voxel_fit, self)
             n_jobs = kwargs.get("n_jobs", multiprocessing.cpu_count() - 1)
             vox_per_chunk = kwargs.get(
-                "vox_per_chunk",
-                np.max([data_to_fit.shape[0] // n_jobs, 1]))
-            chunks = [data_to_fit[ii:ii + vox_per_chunk]
-                      for ii in range(0, data_to_fit.shape[0], vox_per_chunk)]
-            fit_array[np.where(mask)] = np.concatenate((
-                paramap(
-                    _parallel_fit_worker,
-                    chunks,
-                    func_args=[single_voxel_with_self],
-                    **kwargs)
-                    ))
+                "vox_per_chunk", np.max([data_to_fit.shape[0] // n_jobs, 1])
+            )
+            chunks = [
+                data_to_fit[ii : ii + vox_per_chunk]
+                for ii in range(0, data_to_fit.shape[0], vox_per_chunk)
+            ]
+            fit_array[np.where(mask)] = np.concatenate(
+                (
+                    paramap(
+                        _parallel_fit_worker,
+                        chunks,
+                        func_args=[single_voxel_with_self],
+                        **kwargs,
+                    )
+                )
+            )
         return MultiVoxelFit(self, fit_array, mask)
 
     return new_fit
--- a/dipy/reconst/multi_voxel.py
+++ b/dipy/reconst/multi_voxel.py
@@ -11,6 +11,22 @@ from dipy.utils.parallel import paramap
 from functools import partial
 import multiprocessing
 
+def parallel_fit_worker(vox_data, single_voxel_fit, **kwargs):
+    """
+    Each pool process calls this worker.
+    Fit model on chunks
+    Parameters
+    -----------
+    arguments: tuple
+        tuple should contains a model and
+        list of indexes
+    Returns
+    --------
+    result: list of tuple
+        return a list of tuple(voxel index, model fitted instance)
+    """
+    return [single_voxel_fit(data) for data in vox_data]
+
 
 def multi_voxel_fit(single_voxel_fit):
     """Method decorator to turn a single voxel model fit
@@ -24,9 +40,7 @@ def multi_voxel_fit(single_voxel_fit):
 
         # Make a mask if mask is None
         if mask is None:
-            shape = data.shape[:-1]
-            strides = (0,) * len(shape)
-            mask = as_strided(np.array(True), shape=shape, strides=strides)
+            mask = np.ones(data.shape[:-1], bool)
         # Check the shape of the mask if mask is not None
         elif mask.shape != data.shape[:-1]:
             raise ValueError("mask and data shape do not match")
@@ -39,7 +53,13 @@ def multi_voxel_fit(single_voxel_fit):
         vox_per_chunk = np.max([data_to_fit.shape[0] // n_jobs, 1])
         chunks = [data_to_fit[ii:ii + vox_per_chunk]
                   for ii in range(0, data_to_fit.shape[0], vox_per_chunk)]
-        fit_array[mask] = paramap(single_voxel_with_self, chunks, **kwargs)
+        fit_array[mask] = np.squeeze(
+            paramap(
+                parallel_fit_worker,
+                chunks,
+                func_args=[single_voxel_with_self],
+                **kwargs)
+                )
         return MultiVoxelFit(self, fit_array, mask)
 
     return new_fit
--- a/dipy/reconst/multi_voxel.py
+++ b/dipy/reconst/multi_voxel.py
@@ -35,7 +35,7 @@ def multi_voxel_fit(single_voxel_fit):
         fit_array = np.empty(data.shape[:-1], dtype=object)
         data_to_fit = data[mask]
         single_voxel_with_self = partial(single_voxel_fit, self)
-        n_jobs = kwargs.get("n_jobs", multiprocessing.cpu_count() -1)
+        n_jobs = kwargs.get("n_jobs", multiprocessing.cpu_count() - 1)
         vox_per_chunk = np.max([data_to_fit.shape[0] // n_jobs, 1])
         chunks = [data_to_fit[ii:ii + vox_per_chunk]
                   for ii in range(0, data_to_fit.shape[0], vox_per_chunk)]
--- a/dipy/reconst/multi_voxel.py
+++ b/dipy/reconst/multi_voxel.py
@@ -3,7 +3,6 @@ from functools import partial
 import multiprocessing
 
 import numpy as np
-from numpy.lib.stride_tricks import as_strided
 from tqdm import tqdm
 
 from dipy.core.ndindex import ndindex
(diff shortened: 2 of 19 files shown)


skoudoro:
Thank you for starting this @arokem? 

Have you looked at https://github.com/dipy/dipy/pull/1418?  I think some ideas can be reused here.

arokem:
Oh yeah - that’s a great pointer! I’ll try to incorporate the ideas you
implemented there into this PR.

arokem:
I ran a benchmark on a beefy 24-cpu compute server with the recent commit.I get a roughly _13x_ speedup for fitting the fwdti model with `engine="joblib"` relative to the default serial mode. I should maybe mention that the server is also doing a bunch of other work, so it's not the cleanest benchmark, but still quite promising.

arokem:
Does anyone understand why half the CI actions are still pending? They have been pending since Friday!

skoudoro:
No, but I will restart them first

arokem:
Does anyone understand these CI failures? I don't _think_ they are related to the content of the PR, but I might be missing something.

skoudoro:
> Does anyone understand these CI failures? I don't think they are related to the content of the PR, but I might be missing something.

Both failures are on the parallelization CI's with a memory leaks issue. This might be due to some of the parallel packages that might change some environment variable flags. These flags could have an impact on this parallelized function. 

All supposition, this is what comes first to my mind.

skoudoro:
the failing function are using `openmp`

arokem:
Hey @skoudoro, I noticed that you did not pin the ray version in #2600, instead pinning only protobuf, but I am seeing this again on the CI: https://github.com/dipy/dipy/runs/6634820045?check_suite_focus=true#step:9:119, which suggests to me that I should pin ray to 0.11 for now. Does that make sense to you? I'll give it a try here.

arokem:
Or possibly 1.11.1

arokem:
We're back to this failure: https://github.com/dipy/dipy/runs/6645881563?check_suite_focus=true#step:9:3751

Interestingly, I can't get this to fail locally on my machine (in an env with dask, ray and joblib installed). I also don't exactly understand how this is related to openmp. Does `set_number_of_points` use openmp?

dipy/reconst/multi_voxel.py
```diff
@@ -5,36 +5,69 @@
...
+                    fit_array[ijk] = single_voxel_fit(self, data[ijk])
                 bar.update()
-        bar.close()
+            bar.close()
+        else:
+            data_to_fit = data[mask]
+            single_voxel_with_self = partial(single_voxel_fit, self)
+            n_jobs = kwargs.get("n_jobs", multiprocessing.cpu_count() - 1)
+            vox_per_chunk = np.max([data_to_fit.shape[0] // n_jobs, 1])
+            chunks = [data_to_fit[ii:ii + vox_per_chunk]
```
arokem, line 62: This might duplicate memory. Need to benchmark.

arokem: We can use this: https://pypi.org/project/memory-profiler/

arokem:
Plan to make progress here: 

- [ ] Set up experimental datasets: All of the models except for DSI can use multi-shell data. Only CSD (I think) can run on single-shell data. For multi-shell datasets we can use HBN and HCP. For DSI, I guess we can use the dsi dataset we have in our data fetchers. We'll need to set up fetchers for HBN data (see #2695) and for HCP (see #2696). 

- [ ] Set up experimental scripts (separate repo, probably): these should run every one of the models that are decorated in this PR with: 
        1. Serial mode.
        2. Parallelized by voxel with dask, ray, joblib.
        3. Parallelized by chunk with dask, ray, joblib.
        4. Parallelized with different backends if possible.
        5. For ray/dask, parallelize on a big distributed AWS cluster.

- [ ] Run the experiments. We'll need to have some uniform hardware settings. We'll want to run this on different OS (Windows, Linux, Mac OS) and maybe on different kinds of computational architectures (e.g., distributed cluster vs. one big machine). 
- [ ] Separately benchmark timing (this is straightforward) and memory (using https://github.com/pythonprofilers/memory_profiler). 
- [ ] Compare and contrast 😄

skoudoro (commented):
For now, it works with `engin in ["serial", "dask"]

- my laptop crash with `ray`
- See below for joblib issue. 

I will share the timing when those 2 are fixed.

Thanks @arokem

arokem:
The errors on the CI are really puzzling and hard to reproduce locally and we worry that it might be some funky interaction with joblib, so I am removing the multi_voxel decorator for that model and we can reinstate it on a separate PR later on, if we fix everything else here.

arokem:
OK - having removed the decorator from the SFM model, it seems that the only remaining error on the CI is completely unrelated. Or am I missing something?

arokem:
At any rate, @asagilmore has now completed an extensive set of experiments with this PR and we are glad to say that [Ray](https://www.ray.io/) in particular provides a substantial speedup for reconstruction models (on the order of 10x) across two different models, and across a pretty wide range of hardware setups and chunking schemes. We're writing up a report about this [here](https://nrdg.github.io/2024-dipy-parallelization/) and we'd be happy to have input on the results and ideas that we are developing there (the repo for that report is here: https://github.com/nrdg/2024-dipy-parallelization). 

With that said, I think that this code and the results we report are ready for a review, and are hopefully close to a shape where they can be merged for an upcoming release.

arokem:
Looks like installing pytables on mac is (newly?) broken: https://github.com/dipy/dipy/actions/runs/8847440312/job/24295269319?pr=2593#step:6:125

skoudoro:
Yes, I saw that with pytable.

Not sure what we can do apart from reporting.

The last release was 6month ago so not sure what changed last week.

Maybe a release of h5py...

arokem:
There was one on April 10th: https://pypi.org/project/h5py/3.11.0/

I'm trying to pin to 3.10 in https://github.com/dipy/dipy/pull/2593/commits/cd0e653bdb56e6bb6accdce637483bc1d1837f5b. Let's see what we learn.

arokem:
Following #3202, what's the right way to catch a warning thrown when importing one of the optional dependencies? I am getting: 

```
reconst/tests/test_multi_voxel.py::test_multi_voxel_fit
  /home/runner/work/dipy/dipy/venv/lib/python3.10/site-packages/ray/_private/pydantic_compat.py:2: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
    from pkg_resources import packaging

reconst/tests/test_multi_voxel.py::test_multi_voxel_fit
  /home/runner/work/dipy/dipy/venv/lib/python3.10/site-packages/pkg_resources/__init__.py:2832: DeprecationWarning: Deprecated call to `pkg_resources.declare_namespace('google')`.
  Implementing implicit namespace packages (as specified in PEP 420) is preferred to `pkg_resources.declare_namespace`. See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages
    declare_namespace(pkg)
```

I believe this is emitted when importing ray. Should I explicitly ignore it in a context manager around that import?

skoudoro:
The "right way" would be to add the warning message with `ignore` status in the `pyproject.toml`.

see here: https://github.com/dipy/dipy/blob/60669b888e7bc322a8b973f3a0b8121e4fd7f7d1/pyproject.toml#L184-L186

skoudoro:
if it is to much for the `pyproject.toml`, we could add/create a specific python file for that.

skoudoro:
For example, `mne-python` do it directly in the `conftest.py` instead of the `pyproject.toml`.

see here: https://github.com/mne-tools/mne-python/blob/main/mne/conftest.py#L131-L178


So, this is something to decide together, what is your opinion @arokem and @jhlegarreta

arokem:
I like it better in the conftest

jhlegarreta:
Hats off for this work, Ariel.

> So, this is something to decide together, what is your opinion @arokem and @jhlegarreta ?

Having filtering rules both in the `pyproject.toml` and the `conftest.py` file would make things easier to follow, as we would need to look at two files instead of one. Also, not sure if `pytest` would end up by overriding the rules in one file by the ones that it reads in the last place.

So probably as Serge says, if the list of rules becomes too lengthy, better to keep them all in the `conftest.py` file.

arokem:
Thanks! I will wait for your PR to see how this is done. 

In the meanwhile, it doesn't look like pinning h5py helped with the pytables installation.

oesteban:
@skoudoro @arokem @jhlegarreta -- please let me know if I can be of help, if you need review or anything

arokem:
Looks like there are some conflicts with master, so let me rebase first (I'll do that in a couple of hours when I have a little bit more quiet...). I think this is ready to be merged, in principle, but the CI keeps acting up.

arokem:
Looks like the remaining failure is unrelated. Other than that, I think that this is ready for a review and, if everything looks good, to be merged.

dipy/reconst/multi_voxel.py
```diff
@@ -1,42 +1,88 @@
...
-            if mask[ijk]:
-                fit_array[ijk] = single_voxel_fit(self, data[ijk])
+        # Default to serial execution:
+        engine = kwargs.get("engine", "serial")
+        if engine == "serial":
+            bar = tqdm(
+                total=np.sum(mask), position=0, disable=kwargs.get("verbose", True)
+            )
+            bar.set_description(
+                "Fitting reconstruction model", "using serial execution"
```
jhlegarreta, line 58: Can this be put into a single string?

dipy/utils/parallel.py
```diff
@@ -42,17 +46,21 @@ def paramap(
     engine : str
```
jhlegarreta, line 46: str, optional

dipy/utils/parallel.py
```diff
@@ -42,17 +46,21 @@ def paramap(
...
         {"dask", "joblib", "ray", "serial"}
         The last one is useful for debugging -- runs the code without any
-        parallelization. Default: "joblib"
+        parallelization. Default: "ray"
```
jhlegarreta, line 49: I'd remove altogether the `Default: ...` part: it is already documented in the method signature, and reduces maintenance burden, no risk to make them go out of sync if the signature changes, etc.

dipy/utils/parallel.py
```diff
@@ -42,17 +46,21 @@ def paramap(
...
     backend : str, optional
         What joblib or dask backend to use. For joblib, the default is "loky".
         For dask the default is "threading".
+    clean_spill : bool, optional
+        If True, clean up the spill directory after the computation is done.
+        Only applies to "ray" engine. Otherwise has no effect.
+        Default: True.
```
jhlegarreta, line 56: Same as above concerning the default value.

jhlegarreta (commented):
Some minor inline comments. Have not tested.

About pinning the dependencies, can they be made more flexible (`>=` or `>`)?

Also, DIPY's compatibility to NumPy 2.0.0 has been requested in an issue, so not sure if this can be made work with NumPy 2.0.0 once PR #3266 (if it we make it successful) gets merged.

pyproject.toml
```diff
@@ -96,11 +96,11 @@ dev = [
     # Also update [build-system] -> requires
     'meson-python>=0.13',
     'wheel',
-    'setuptools>=67',
+    'setuptools==69.5.1',
```
oesteban, line 99: Perhaps more flexible, but still ensuring it's locked in to the 69 major.
```suggestion
    'setuptools~=69.5',
```

pyproject.toml
```diff
@@ -96,11 +96,11 @@ dev = [
...
     'packaging>=21',
     'ninja',
     'Cython>=0.29.35',
-    'numpy>=1.22',
+    'numpy>=1.22, <2.0',
```
oesteban, line 103: Equivalent:
```suggestion
    'numpy~=1.22',
```

tools/ci/install_dependencies.sh
```diff
@@ -7,7 +7,7 @@ set -ex
 
 echo "Display Python version"
 python -c "import sys; print(sys.version)"
-python -m pip install -U pip setuptools>=30.3.0 wheel
+python -m pip install -U pip setuptools==69.5.1 wheel
```
oesteban, line 10:
```suggestion
python -m pip install -U pip 'setuptools~=69.5' wheel
```

oesteban (commented):
Added some suggestions to make version pinnings more flexible but still specific, following suggestion of @jhlegarreta

arokem:
Thanks for the reviews! For now I've accepted Oscar's suggestions for more flexible versions and will address Jon's suggestions later when I'm not on my phone 😀

dipy/reconst/multi_voxel.py
arokem: Done in faf69f4a7

dipy/utils/parallel.py
arokem: Done in 444503737

dipy/utils/parallel.py
arokem: Done in 0140c8cf0

arokem:
Looks like there's a quite a bit more work to get to numpy 2.0 compatibility. I wonder whether we shouldn't cut a release that pins numpy < 2.0 for now, so that users aren't running into issues.

jhlegarreta:
Re https://github.com/dipy/dipy/pull/2593#issuecomment-2180609120 PR #3267 will bring us closer.

pyproject.toml
```diff
@@ -96,11 +96,11 @@ dev = [
     # Also update [build-system] -> requires
     'meson-python>=0.13',
     'wheel',
-    'setuptools>=67',
+    'setuptools~=69.5',
     'packaging>=21',
     'ninja',
     'Cython>=0.29.35',
-    'numpy>=1.22',
+    'numpy~=1.22',
```
skoudoro, line 103: the update of pyproject.toml is not needed after the merge of #3274.

it should works with numpy 2.0.0 and below.

.github/workflows/test_template.yml
```diff
@@ -96,6 +96,16 @@ jobs:
         python-version: ${{ matrix.python-version }}
         channels: defaults, conda-forge
         use-only-tar-bz2: true
+    - name: Install HDF5 and pytables on macOS
+      if: runner.os == 'macOS'
```
skoudoro, line 100: we should test if we are in optional dependency also. We do not want to install pytables all the times

skoudoro (commented):
Hi @arokem,

can you rebase this PR and remove your change in `pyproject.toml`.

the update of pyproject.toml is not needed after the merge of #3274.

it should works with numpy 2.0.0 and below now, it has been fixed in https://github.com/dipy/dipy/pull/3274.

See my other comment below.

After that, this PR should be good to go

NOTE: also, I see that you do a lot of stuff for pytable and h5py.  I do not know if we should keep it since pytables will release a new version with macOS arm64 wheels. see https://github.com/PyTables/PyTables/issues/1165.  
Second thought, we should keep it until they release, then, easy to remove.

skoudoro:
> Sounds reasonable. 1.9.0 was thought to be compatible, though:
https://github.com/dipy/dipy/blame/d5e34fbdca1a30829060827f944b5aeb9f145e75/doc/index.rst#L95

it was for a long moment with the numpy2dev, however, we did not update it with `numpy2rc1` and up

.github/workflows/test_template.yml
arokem: Could you please use the "add suggestion" button to make a suggested change here? I don't know how to test for this.

arokem:
Rebased. Could you help me with the CI? I am not sure how to check that we're in the "optional" case.

.github/workflows/test_template.yml
skoudoro:
```suggestion
      if: ${{ (runner.os == 'macOS')  &&  (env.EXTRA_DEPENDS != '')  &&  (env.INSTALL_TYPE == 'pip') }}
```

skoudoro (commented):
> Rebased. Could you help me with the CI? I am not sure how to check that we're in the "optional" case.

Thanks! It should be something like the suggestion below. I hope there is no typo in this syntax

arokem:
Thanks!

pyproject.toml
```diff
@@ -37,7 +37,7 @@ classifiers = [
 ]
 dynamic = ['version']
 dependencies = [
-    "numpy>=1.22.4",
+    "numpy>=1.22.4, <2.0",
```
skoudoro, line 40:
```suggestion
    "numpy>=1.22.4",
```

pyproject.toml
```diff
@@ -96,11 +96,11 @@ dev = [
     # Also update [build-system] -> requires
     'meson-python>=0.13',
     'wheel',
-    'setuptools>=67',
+    'setuptools~=69.5',
     'packaging>=21',
     'ninja',
     'Cython>=0.29.35',
-    'numpy>=1.22',
+    'numpy<=2.0.0',
```
skoudoro, line 103:
```suggestion
    'numpy>=1.22.4',
```

skoudoro (commented):
Hi @arokem,

I was going to merge but I saw that a previous comment has been remove but not addressed. see below the suggestion and then I can go ahead.

Thank you

arokem:
Sorry about that and thanks for spotting this.

skoudoro (approved):
okay, all good, I will go ahead and merge this PR. 

Thank you for this amazing work @arokem and @asagilmore!

**NOTE**: we need to document it in details somewhere. It will take time.  So, for now, I recommend a follow-up PR to add a small recipe in dipy documentation (see https://docs.dipy.org/stable/recipes).  Just add a small section: `How to accelerate the fitting in all DIPY reconstruction models?` or something similar.

> We're writing up a report about this [here](https://nrdg.github.io/2024-dipy-parallelization/) and we'd be happy to have input on the results and ideas that we are developing there (the repo for that report is here: https://github.com/nrdg/2024-dipy-parallelization).

Where can we find the comparison with joblib, dask and ray? I see a lot of work with Ray and this is clearly the recommended backend. However, it would be great to see the advantage against the others backend. Thank you for your feedback

skoudoro:
Following up with my questions above @arokem and @asagilmore

Issue #2594, opened 2022-05-17
DIPY compilation fails with the last release of cython (0.29.29)


skoudoro:
> Avoid acquiring the GIL at the end of nogil functions. This change was backported in order to avoid generating wrong C code that would trigger C compiler warnings with tracing support enabled

This new cython feature might have an impact on our current code. We need to introspect.  see below:

cython 0.29.29 release notes: https://cython.readthedocs.io/en/latest/src/changes.html#id27

the current error:
```python
Installing collected packages: dipy
  Attempting uninstall: dipy
    Found existing installation: dipy 1.6.0.dev0
    Uninstalling dipy-1.6.0.dev0:
      Successfully uninstalled dipy-1.6.0.dev0
  Running setup.py develop for dipy
    ERROR: Command errored out with exit status 1:
     command: /Users/koudoro/miniconda3/envs/dipy-env-38/bin/python -c 'import io, os, sys, setuptools, tokenize; sys.argv[0] = '"'"'/Users/koudoro/Software/dipy/setup.py'"'"'; __file__='"'"'/Users/koudoro/Software/dipy/setup.py'"'"';f = getattr(tokenize, '"'"'open'"'"', open)(__file__) if os.path.exists(__file__) else io.StringIO('"'"'from setuptools import setup; setup()'"'"');code = f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' develop --no-deps
         cwd: /Users/koudoro/Software/dipy/
    Complete output (112 lines):
    running develop
    running egg_info
    writing dipy.egg-info/PKG-INFO
    writing dependency_links to dipy.egg-info/dependency_links.txt
    writing requirements to dipy.egg-info/requires.txt
    writing top-level names to dipy.egg-info/top_level.txt
    reading manifest file 'dipy.egg-info/SOURCES.txt'
    reading manifest template 'MANIFEST.in'
    /Users/koudoro/miniconda3/envs/dipy-env-38/lib/python3.8/site-packages/setuptools/dist.py:502: UserWarning: Normalizing '1.6.0dev0' to '1.6.0.dev0'
      warnings.warn(tmpl.format(**locals()))
    /Users/koudoro/miniconda3/envs/dipy-env-38/lib/python3.8/site-packages/setuptools/command/easy_install.py:158: EasyInstallDeprecationWarning: easy_install command is deprecated. Use build and pip and other standards-based tools.
[...]

PR #2595, opened 2022-05-17
Temporary skip Cython 0.29.29 

skoudoro:
We currently have an issue with cython 0.29.29 described in the issue https://github.com/dipy/dipy/issues/2594.

To avoid blocking some PRs, this PR skips this Cython version (0.29.29) until we solve our issue.

### DIFF
--- a/.github/workflows/test_compat.yml
+++ b/.github/workflows/test_compat.yml
@@ -12,13 +12,13 @@ jobs:
     with:
       runs-on: '["ubuntu-latest", ]'
       python-version: '["3.7", ]'
-      depends: cython==0.29 numpy==1.15.0 scipy==1.1 nibabel==3.0.0 h5py==2.8.0 tqdm
+      depends: cython==0.29.24 numpy==1.15.0 scipy==1.1 nibabel==3.0.0 h5py==2.8.0 tqdm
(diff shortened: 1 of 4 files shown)


arokem:
Seems fine as a stop-gap. Should we also report our findings to the Cython team?

arokem:
Looks like there's a new Cython release out that might fix our issue: https://cython.readthedocs.io/en/latest/src/changes.html#id27. WDYT?

skoudoro:
> Looks like there's a new Cython release out that might fix our issue:

Not sure I understand what you mean. This is exactly this new release that creates the issue and makes our build fail.

arokem:
Sorry -- to be explicit: I think that 0.29.29 causes issues. Does the even newer 0.29.30 resolve them?

arokem:
I swear I am not making this up:
Screen Shot 2022-05-17 at 10 03 46 AM

arokem:
Maybe your browser is caching a previous state of this webpage?

skoudoro:
Thanks!!! I could not understand, refresh all my tabs and indeed, I can see 0.29.30 !!!

I suppose the release will come soon and fix the issue. Thanks and sorry for that.

PR #2450, opened 2021-10-13
[NF] New tracking Algorithm:  Parallel Transport Tractography (PTT)

baranaydogan:
The goal of this PR is to implements a state-of-the-art tractography algorithm, parallel transport tractography (PTT). PTT is capable of reconstructing geometrically smooth and topographically organized fiber bundles.

More information on this page: https://dmritrekker.github.io/tutorials/tutorial_2.html

**paper:** [aydogan2021](https://ieeexplore.ieee.org/abstract/document/9239977/)  and [aydogan2019](https://www.researchgate.net/publication/336847169_A_novel_fiber-tracking_algorithm_using_parallel_transport_frames)

TODO:
-------
- [ ] Clean a bit the code
- [ ] Add more tests
- [ ] Update DIPY Tracking API to increase the flexibility.
- [ ] add tutorials
- [ ] finalize comparison with the original implementation

PR #2596, opened 2022-05-20
NF - Parallel Transport Tractography (PTT)

gabknight:
The goal of this PR is to implements a state-of-the-art tractography algorithm, parallel transport tractography (PTT). PTT is capable of reconstructing geometrically smooth and topographically organized fiber bundles.

More information on this page: https://dmritrekker.github.io/tutorials/tutorial_2.html

paper: [aydogan2021](https://ieeexplore.ieee.org/abstract/document/9239977/) and [aydogan2019](https://www.researchgate.net/publication/336847169_A_novel_fiber-tracking_algorithm_using_parallel_transport_frames)

Opened following the development in PR #2450

### DIFF
--- a/dipy/direction/ptt_direction_getter.pyx
+++ b/dipy/direction/ptt_direction_getter.pyx
@@ -50,6 +50,8 @@ cdef class PTTDirectionGetter(ProbabilisticDirectionGetter):
     cdef double       probe_step_size
     cdef double[9]    propagator
     cdef double       step_size
+    cdef int          rejection_sampling_max_try
+    cdef int          rejection_sampling_nbr_sample
 
 
     def __init__(self, pmf_gen, max_angle, sphere, pmf_threshold=None,
@@ -110,7 +112,10 @@ cdef class PTTDirectionGetter(ProbabilisticDirectionGetter):
                                             * self.probe_count)
         self.angular_separation = 2.0 * M_PI / float(self.probe_count)
         self.data_support_exponent = data_support_exponent
+
         self.k_small = 0.0001
+        self.rejection_sampling_max_try = 100
+        self.rejection_sampling_nbr_sample = 10 # Adaptively set in Trekker.
 
         ProbabilisticDirectionGetter.__init__(self, pmf_gen, max_angle, sphere,
                                        pmf_threshold, **kwargs)
@@ -333,16 +338,15 @@ cdef class PTTDirectionGetter(ProbabilisticDirectionGetter):
             Returns 0 if the initialization was successful, or
             1 otherwise.
         """
+        cdef double data_support = 0
+        cdef double max_posterior = 0
+        cdef int tries
 
         self.position[0] = seed_point[0]
         self.position[1] = seed_point[1]
         self.position[2] = seed_point[2]
 
-        cdef double data_support = 0
-        cdef double max_posterior = 0
-
-        cdef int tries
-        for tries in range(1000):
+        for tries in range(self.rejection_sampling_nbr_sample):
             self.initialize_candidate(seed_direction)
             data_support = self.calculate_data_support()
             if data_support > max_posterior:
@@ -353,7 +357,7 @@ cdef class PTTDirectionGetter(ProbabilisticDirectionGetter):
 
         # Initialization is successful if a suitable candidate can be sampled
         # within the trial limit
-        for tries in range(1000):
+        for tries in range(self.rejection_sampling_max_try):
             self.initialize_candidate(seed_direction)
             if (random() * max_posterior <= self.calculate_data_support()):
                 self.last_val = self.last_val_cand
@@ -377,7 +381,7 @@ cdef class PTTDirectionGetter(ProbabilisticDirectionGetter):
         cdef double max_posterior = 0
         cdef double data_support = 0
         cdef double[3] tangent
-        cdef int tries = 0
+        cdef int tries
 
         self.prepare_propagator(self.step_size)
 
@@ -401,8 +405,7 @@ cdef class PTTDirectionGetter(ProbabilisticDirectionGetter):
         self.frame[0][1] = tangent[1]
         self.frame[0][2] = tangent[2]
 
-        # This is adaptively set in Trekker.
-        for tries in range(20):
+        for tries in range(self.rejection_sampling_nbr_sample):
             self.k1, self.k2 = random_point_within_circle(self.max_curvature)
             data_support = self.calculate_data_support()
             if data_support > max_posterior:
@@ -413,7 +416,7 @@ cdef class PTTDirectionGetter(ProbabilisticDirectionGetter):
 
         # Propagation is successful if a suitable candidate can be sampled
         # within the trial limit
-        for tries in range(1000):
+        for tries in range(self.rejection_sampling_max_try):
             self.k1, self.k2 = random_point_within_circle(self.max_curvature)
             if random() * max_posterior <= self.calculate_data_support():
                 self.last_val = self.last_val_cand
--- a/dipy/direction/ptt_direction_getter.pyx
+++ b/dipy/direction/ptt_direction_getter.pyx
@@ -5,12 +5,14 @@
 """
 Implementation of parallel transport tractography (PTT).
 """
-
-import numpy as np
 cimport numpy as cnp
+from libc.math cimport M_PI, pow, sin, cos
 
-from dipy.direction.probabilistic_direction_getter cimport ProbabilisticDirectionGetter
-from dipy.utils.fast_numpy cimport (copy_point, random, norm, normalize, cross)
+from dipy.direction.probabilistic_direction_getter cimport \
+        ProbabilisticDirectionGetter
+from dipy.utils.fast_numpy cimport (copy_point, cross, norm, normalize, random,
+                                    random_vector, random_perpendicular_vector,
+                                    random_point_within_circle)
 from dipy.tracking.stopping_criterion cimport (StreamlineStatus,
                                                StoppingCriterion,
                                                TRACKPOINT,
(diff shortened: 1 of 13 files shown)


gabknight:
@skoudoro @baranaydogan The PR is ready for review :). The performance is suboptimal, especially regarding the rejection sampling algorithm. It is approximatively 100x slower than probabilistic tracking with the default parameters. We should target improving that in a future PR.

My tests show better connectivity results on a basic setup on disco phantom when using ptt (r=0.88) compared to prob (r=0.84) or det (r=0.84).

gabknight:
`PTTDirectionGetter` can be used in `ParticleFilteringTracking`, but the results would be ideoverride `PTTDG.get_direction(point, direction)` to return a PTT-sampled propagation direction.ntical to using `ProbabilisticDirectionGetter` as input, i.e. it doesn't use the DirectionGetter.generate_streamline(.) to propagate the streamlines. We would need to override `PTTDG.get_direction(point, direction)` to return a PTT-sampled propagation direction.

dipy/direction/ptt_direction_getter.pyx
```diff
@@ -0,0 +1,472 @@
...
+
+         """
+        if probe_length <= 0:
+            raise ValueError("probe_length must be greater than 0.")
+        if probe_radius < 0:
+            raise ValueError("probe_radius must be greater or equal to 0.")
+        if probe_quality < 2:
+            raise ValueError("probe_quality must be greater than 2.")
+        if probe_count < 1:
+            raise ValueError("probe_count must be greater than 1.")
```
skoudoro, line 101: `greater or equal` since default value is 1.

dipy/direction/ptt_direction_getter.pyx
```diff
@@ -0,0 +1,472 @@
...
+            probe (minimum=2).
+        probe_count : integer
+            The number of probes. This parameter sets the number of parallel
+            lines used to model the cylinder (minimum=1).
+        data_support_exponent : double
+            Data support to the power dataSupportExponent is used for
+            rejection sampling.
+
+         """
+        if probe_length <= 0:
```
skoudoro, line 94: can it be 0 really ?

dipy/direction/ptt_direction_getter.pyx
```diff
@@ -0,0 +1,472 @@
...
+
+        self.position[0] = seed_point[0]
+        self.position[1] = seed_point[1]
+        self.position[2] = seed_point[2]
+
+        cdef double data_support = 0
+        cdef double max_posterior = 0
+
+        cdef int tries
+        for tries in range(1000):
```
skoudoro, line 345: 1000 could be a variable if we want to reduces tries number. 

Also, the loop is never break. Does it means that it will do the 1000 tries even if max_posterior is good ?

dipy/direction/ptt_direction_getter.pyx
```diff
@@ -0,0 +1,472 @@
...
+        self.frame[0][0] = tangent[0]
+        self.frame[0][1] = tangent[1]
+        self.frame[0][2] = tangent[2]
+
+        # This is adaptively set in Trekker.
+        for tries in range(20):
+            self.k1, self.k2 = random_point_within_circle(self.max_curvature)
+            data_support = self.calculate_data_support()
+            if data_support > max_posterior:
+                max_posterior = data_support
```
skoudoro, line 409: same questions as above

skoudoro (commented):
Overall it looks good , thanks @gabknight!  

For sure, we will have to optimise this version but this is a good first version! thanks !

dipy/direction/ptt_direction_getter.pyx
gabknight: Thanks. I added the variable. It should have been 20, as in the `propagate(.)` function. The loop never breaks because it is estimating the maximum likelihood for the rejection sampling algorithm.

dipy/direction/ptt_direction_getter.pyx
gabknight: I added the variable and reduced the value to 10, which almost 2X the speed of PTT. In my tests, the results are not different. We need optimizing this in a future PR.

dipy/direction/ptt_direction_getter.pyx
gabknight: `< 1` is ok here, as if `True` it triggers the `ValueError`. Should I write the test as `if not probe_count >= 1` ?

dipy/direction/ptt_direction_getter.pyx
gabknight: It cannot be, and will trigger the `ValueError`. Similarly, maybe it would be clearer to have `if not probe_lenght > 0`. What do you think?

skoudoro: yes, it would be clearer

dipy/direction/ptt_direction_getter.pyx
skoudoro: no, sorry for the confusion, it is good

gabknight: thanks, fixed.

skoudoro (approved):
All good from my side. A good first version.

Of course, it needs to be optimized. but we can do it in a new PR. Thanks @gabknight for this work !

@baranaydogan Can you test it from your side ???   

Any additional comments ?  I will go ahead and merge it tomorrow morning if no additional comment

dipy/direction/ptt_direction_getter.pyx
```diff
@@ -0,0 +1,472 @@
...
+    cdef double       probe_radius
+    cdef double       probe_step_size
+    cdef double[9]    propagator
+    cdef double       step_size
+
+
+    def __init__(self, pmf_gen, max_angle, sphere, pmf_threshold=None,
+                 double probe_length=0.5, double probe_radius=0,
+                 int probe_quality=3, int probe_count=1,
+                 double data_support_exponent=1, **kwargs):
```
baranaydogan, line 60: Because the default values are slightly different in Trekker, the results will not be the same, which is of course OK. But this might be good to mention in the documentation later on. (For example, probe_length's default value 0.25 x voxelSize in Trekker.)

dipy/direction/ptt_direction_getter.pyx
baranaydogan: This loop estimates the max_posterior. The estimation is done by picking the maximum data_support among 1000 random trials. So all 1000 trials are used that is why the loop should not break. Here, 1000 is quite an arbitrary choice. In my opinion, this should be sufficient but we have not thoroughly investigated this. 20 on the other hand sounds very small in my opinion, it is quite likely that the max_posterior will be underestimated in that case. Maybe, for now we can stick to a high value, like 1000, to be on the safe side.

baranaydogan: 20 might be sufficient during propagation though, because the tangent is fixed. But at the initialization, the tangent is also most of the times not fixed. So if it is the case that there is an initial direction, then 20 can be used instead of 1000.

dipy/direction/ptt_direction_getter.pyx
baranaydogan: If speed is a big concern 10 will work too, although in that case max_posterior will be underestimated quite often.

doc/examples/tracking_ptt.py
```diff
@@ -0,0 +1,101 @@
...
+Start the local tractography using PTT direction getter.
+"""
+
+sphere = get_sphere(name='repulsion724')
+fod = csd_fit.odf(sphere)
+pmf = fod.clip(min=0)
+ptt_dg = PTTDirectionGetter.from_pmf(pmf, max_angle=15, probe_length=0.5,
+                                     sphere=sphere)
+
+# Parallele Transport Tractography
```
baranaydogan, line 69: Parallel

dipy/direction/ptt_direction_getter.pyx
skoudoro: Can you address this @gabknight ? thank you

doc/examples/tracking_ptt.py
skoudoro: good catch

baranaydogan:
This looks great! I went through the code, all seems fine to me. Next, I will run some tests.

dipy/direction/ptt_direction_getter.pyx
gabknight: Thanks for the clarification @baranaydogan. The rejection sampling is suboptimally implemented now and can be improved. I set the `tries` to 20 (now 10) into the initialization loop, as in the dipy implementation the initial tangent is always set to one of the peaks at the seed location. This is also the case for probabilistic DG. This is something to generalize in a future the tracking module overhaul.

dipy/direction/ptt_direction_getter.pyx
gabknight: ok. Currently, tries for suitable propagation direction is set to 10. The max posterior estimation is set to 100 samples.

dipy/direction/ptt_direction_getter.pyx
gabknight: Ok thanks. I clarified this in the doc.

gabknight:
I noticed the tracking was performed in voxel unit, while the input parameter were given in mm. I made the appropriate changes to track in mm. @skoudoro can you have a look at commit 555064f.

dipy/direction/ptt_direction_getter.pyx
baranaydogan: Things differ in that case of course. This is good know. Thank you for the information.

Garyfallidis:
Great start @gabknight ! The tutorial needs more explanation. And it seem there are multiple trivial tricks to apply to make the PTT implementation faster. But for now this is great! Let's all start using it and start providing feedback for improvements.

PR #2599, opened 2022-05-27
Support empty ArraySequence in transform_streamlines

frheault:
This emerged from testing conversion of empty tck/trk to vtk and its was crashing

### DIFF
--- a/.github/workflows/test_template.yml
+++ b/.github/workflows/test_template.yml
@@ -75,10 +75,10 @@ jobs:
       TEST_WITH_XVFB: ${{ inputs.enable-viz-tests }}
       PRE_WHEELS: "https://pypi.anaconda.org/scipy-wheels-nightly/simple"
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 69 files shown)


arokem:
Thanks for doing this! As always, a test that fails without this functionality would be a good addition to this PR.

skoudoro:
2 CI matrices are failing because of `ray`. Are you aware of a change @arokem?  it failed by just importing the module....

arokem:
Yeah. We need to pin the ray version for now, as in: https://github.com/dipy/dipy/pull/2593/commits/7628a780b5c077ae1a933a7cad45abce192d8ca7

skoudoro:
> Yeah. We need to pin the ray version for now, as in: https://github.com/dipy/dipy/commit/7628a780b5c077ae1a933a7cad45abce192d8ca7

ok, good to know. Thanks!

skoudoro (commented):
Hi @frheault,

it looks good to me. We might need to create a patch release of FURY before merging this PR since it fails [here](https://github.com/dipy/dipy/runs/6630377017?check_suite_focus=true#step:9:4087).

I also restarted some CI's matrix to make sure that everything is ok

PR #2600, opened 2022-05-27
Pin Ray

skoudoro:
Quick PR to pin ray to version 1.11 in order to avoid CI's failure (Version 1.12 fails to be imported for now).


skoudoro:
Still failing with 1.11, do you know at which version should I pin @arokem?

arokem:
I was working off of https://github.com/ray-project/ray/issues/24169, and pinning to 1.11, but it looks like the issue you are seeing here is different from the one I was trying to address (see https://github.com/dipy/dipy/runs/6428492396?check_suite_focus=true#step:9:119). Looks like you ran into something more like this one: https://github.com/ray-project/ray/pull/25211

arokem:
Thanks!

PR #2618, opened 2022-07-06
Use np.linalg.multi_dot instead of multiple np.dot routines

maldil:
The Numpy [API documentation](https://numpy.org/doc/stable/reference/generated/numpy.linalg.multi_dot.html) recommends using `np.linalg.multi_dot` instead of multiple `np.dot` calls since it is more succinct and effective. This PR is to make the required changes.


skoudoro (commented):
Hi @maldil,

Thank you for this!

our Numpy minimum version is 1.14.5 so do you know in which version `multi_dot` has been introduced?

Also, can you address the [pep8 above](https://github.com/dipy/dipy/pull/2618#issuecomment-1175844821)

arokem:
Looks like it's been around for a while: https://github.com/numpy/numpy/blame/da6297b9f799301a109f478f9056dc9f5b7c6d27/numpy/linalg/linalg.py#L2618. 

The optional `out` argument was added in 1.19, so we might want to avoid using that for now.

skoudoro (commented):
Great! Looks good to me, we will merge it as soon as the CI's come back green.

Thanks again @maldil!

PR #2626, opened 2022-07-22
Adding Synb0

pjsjongsung:
I've added files required by Synb0. The format was edited so that it would match how resdnn was implemented.

Side note: Should we change Synb0 and resdnn to be directly imported from dipy.nn? (Right now it's dipy.nn.synb0.Synb0 and likewise on resdnn)


pjsjongsung:
Will change according to the style and ping.

pjsjongsung:
Passed PEP8!

dipy/nn/synb0.py
```diff
@@ -0,0 +1,417 @@
...
+                self.activation = tf.keras.layers.LeakyReLU(0.01)
+
+            def call(self, input):
+                x = self.conv3d(input)
+                x = self.instnorm(x)
+                x = self.activation(x)
+
+                return x
+
+    def fetch_default_weights(self, idx):
```
frheault, line 217: Could this be idx=0?
What are there 5 choices? Is it ok to always pick the first one?

dipy/nn/synb0.py
```diff
@@ -0,0 +1,417 @@
...
+
+    def predict(self, b0, T1, batch_size=None, average=True):
+        r"""
+        Wrapper function to faciliate prediction of larger dataset.
+        The function will scale the data to meet the required shape of image.
+        Note that the b0 and T1 image should have the same shape
+
+        Parameters
+        ----------
+        b0 : np.ndarray (batch, 77, 91, 77) or (77, 91, 77)
```
frheault, line 300: This shape is a whole image? Does it mean it won't work with the vast majority of images out there?

What is the solution when someone has an image like (100,90,100?) Does it have to be cropped (and loose tissue) or resample to an arbitrary voxel size?

Or is the expected preprocessing to rigidly register the volume to a template with these dimensions?

dipy/nn/synb0.py
```diff
@@ -0,0 +1,417 @@
...
+        -------
+        np.ndarray
+            unnormalized image from range min_img to max_img
+        """
+        return (image+1)/2*(max_img-min_img) + min_img
+
+    def predict(self, b0, T1, batch_size=None, average=True):
+        r"""
+        Wrapper function to faciliate prediction of larger dataset.
+        The function will scale the data to meet the required shape of image.
```
frheault, line 295: Do you mean pad instead of scale? I see at the start of the code that it will crash if not in the right shape. So to me this statement isn't true?

dipy/nn/synb0.py
```diff
@@ -0,0 +1,417 @@
...
+        Note that the b0 and T1 image should have the same shape
+
+        Parameters
+        ----------
+        b0 : np.ndarray (batch, 77, 91, 77) or (77, 91, 77)
+            For a single image, input should be a 3D array. If multiple images,
+            there should also be a batch dimension.
+
+        T1 : np.ndarray (batch, 77, 91, 77) or (77, 91, 77)
```
frheault, line 304: It should be explained somewhere if the data has to be skull-strip or not.

dipy/nn/synb0.py
```diff
@@ -0,0 +1,417 @@
...
+        """
+        if all([b0.shape[1:] != (77, 91, 77), b0.shape != (77, 91, 77)]) or \
+                b0.shape != T1.shape:
+            raise ValueError('Expected shape (batch, 77, 91, 77) or \
+                             (77, 91, 77) for both inputs')
+
+        shape = b0.shape
+        dim = len(shape)
+
+        if dim == 3:
```
frheault, line 330: Some documentation about this pre-processing would be useful.

dipy/nn/synb0.py
```diff
@@ -0,0 +1,417 @@
...
+                        temp = self.__unnormalize(prediction[j], p99[j], 0)
+                        prediction[j] = temp
+
+                    prediction = prediction[:, 2:-1, 2:-1, 3:-2, 0]
+                    prediction = np.moveaxis(prediction, 1, -1)
+
+                mean_pred[..., i] = prediction
+            prediction = np.mean(mean_pred, axis=-1)
+        else:
+            if dim == 3:
```
frheault, line 384: Would it be simpler at the beginning to verify if 3D and to simply add an extra dimension to fake a batch size of 1?
This would simplify the whole thing from the start, no?

frheault (commented):
Hello @pjsjongsung, are the tests passing on your personal computer? And since the last commit either did not run or crashed I don't know if its working <normally>.

I was also wondering if it was possible to make an example such as:
https://github.com/dipy/dipy/blob/master/doc/examples/histology_resdnn.py

A bit more detailed than tests with screenshots showing input/output/difference would be nice (if possible).
It would also be really nice if the example was made from <typical data> and prepared the data so the shape matches instead of simply downloading the test data that is already magically the exact expected shape.
(Showcasing resampling, cropping or registration so if someone wants to use on their data they don't get an immediate raise ValueError)

dipy/nn/synb0.py
pjsjongsung: The synb0 pipeline averages the 5 outputs of the 5 trained model weights. While the user can use this function to try just one of the models, it is more intended as an internal class function for the predict function. I can add more description about the function if that helps.

dipy/nn/synb0.py
pjsjongsung: I am not sure how the shape is created. This model is intended to be a part of the pipeline, hence assumes it gets the intended input shape.

dipy/nn/synb0.py
pjsjongsung: You are correct. I'll fix the description.

dipy/nn/synb0.py
pjsjongsung: This model is not intended to be used as a stand-alone model at the moment, so should follow whatever the previous steps of the pipeline gives. However, I do believe the input needs to be skull stripped.

dipy/nn/synb0.py
pjsjongsung: Good idea. Will add it on the next commit.

dipy/nn/synb0.py
pjsjongsung: It could. I'll look into it and see if that's possible.

pjsjongsung:
@frheault Thanks for the comments. I can add some descriptions and try to simplify the batch dimension part, but a lot of the problems come from the fact that this is not yet a stand-alone model. As mentioned in the meeting a few weeks ago, next steps will be about how to implement rest of the pipeline in DIPY.

pjsjongsung:
Fixed issues mentioned by @frheault . I ran pytest on my windows 10 python 3.8 with no errors. I also verified that the model is returning the correct result.

dipy/nn/synb0.py
Garyfallidis: Moving forward we have to think about these constraints.

dipy/nn/synb0.py
```diff
@@ -0,0 +1,410 @@
...
+            If False, it uses the loaded weights for prediction.
+            Default is True.
+        Returns
+        -------
+        pred_output : np.ndarray (...) or (batch, ...)
+            Reconstructed b-inf image(s)
+
+        """
+        # Check if shape is as intended
+        if all([b0.shape[1:] != (77, 91, 77), b0.shape != (77, 91, 77)]) or \
```
Garyfallidis, line 335: This fixed numbers are problematic.

Garyfallidis (changes requested):
We should make the parameters variables and set default values but not fix numbers in the class.

Garyfallidis:
@pjsjongsung ping!

pjsjongsung:
@Garyfallidis I think we can change after we have the preprocessing steps? As of now if we fit an image by resizing or padding to match the shape required by the model, the output will likely be very unexpected.

pjsjongsung:
@skoudoro It seems the error is coming from trying to load Synb0 without Tensorflow. Is there a way to skip the error or file itself if the user does not have TF2.x? I'm confused because it shouldn't be different from the historesdnn model already inside the nn module, and pytest passed on my local pc.

dipy/nn/synb0.py
```diff
@@ -0,0 +1,410 @@
...
+
+        log_level = 'INFO' if verbose else 'CRITICAL'
+        set_logger_level(log_level)
+
+        # Synb0 network load
+
+        self.model = self.UNet3D()
+        self.model.build(input_shape=(None, 80, 80, 96, 2))
+    
+    class UNet3D(Model):
```
skoudoro, line 91: Hi @pjsjongsung,

When TensorFlow is not present,  `Model` is not imported, so this generates an error because it does not know what it is.....

Actually, it is not a test which is failing but when pytest is parsing all the files to find tests

skoudoro: BTW, in resdnn, they never inherit from a model, that is the difference.

pjsjongsung: Is this line something I can add to above the class definition to solve this?
`@doctest_skip_parser`

pjsjongsung: I see. I'll make changes so that it doesn't do that then. Thank you!

skoudoro:
Hi @pjsjongsung,

thank you for the update. Can you also fix [the pep8 issue](https://github.com/dipy/dipy/pull/2626#issuecomment-1192947641)?

pjsjongsung:
@skoudoro The above pep8 issue should have been resolved in the previous commits. I don't think I introduced any other pep8 issues.

skoudoro:
> @skoudoro The above pep8 issue should have been resolved in the previous commits. I don't think I introduced any other pep8 issues.

I have checked on my laptop, and on github and I confirm that they are not addressed. I also looked at your previous commit, and you addressed only a part of them.

Let me know if you have issues to address the last pep8. sometimes, it is due to the editor.

pjsjongsung:
> 

Apologies for the delay @skoudoro. All warnings from the files/lines I added should have been removed now.

pjsjongsung:
@skoudoro @Garyfallidis Should be done!

pjsjongsung:
The result can change based on available memory. I might as well remove the test then, if you think that's ok. @skoudoro @Garyfallidis . The files still detect corruption when downloading the weights, so shouldn't be a problem.

pjsjongsung:
I removed the tests. Pytest worked without errors in my local machine, so it should be good to go.

Garyfallidis:
Everything usually works on your local machine... but when you try on other machines .... thunderstorms!

Garyfallidis:
There is a similar issue reported here https://github.com/tensorflow/tensorflow/issues/19200 For some people the solution was to use TF 2.3.0. This seems resolving incompatibilities between using the CPU or using the GPU. The DIPY CIs only use CPUs for now.

Garyfallidis:
Please bring back the tests and try to run them with TF 2.3.0 if you haven't already. If the problem persists then we will need to ask the TF team for a fix on their end. Keep it up!

pjsjongsung:
I think reverting to TF 2.3 might cause problems with other models. My machine gave an insufficient CPU memory warning, but still managed to get the error within 2 decimals. I'm hoping this would work.

skoudoro:
No additional comments. Thank you @pjsjongsung, merging

PR #2634, opened 2022-08-25
Fix small difference between pdfs dense 2d/3d

rtaiello:
in` _compute_pdfs_3d` it's used `total_sum` to normalize the joint, whereas in` _compute_pdfs_2d` it was used `valid_points`.


skoudoro:
Thank you for doing this @rtaiello!

Can you update the tests or add a new one?

rtaiello:
@skoudoro I think that was a small typo, which can of test do I have to address?
Thanks in advance!

skoudoro:
yes, it is a small typo. However,  it is strange that we did not catch this small typo. it means that our unit tests do not check this point. 

Can you add/complete this test in the file https://github.com/dipy/dipy/blob/master/dipy/align/tests/test_parzenhist.py#L116

PR #2668, opened 2022-10-29
ENH: Make the DTI fit CLI metric saving message accurate

jhlegarreta:
Make the DTI fit CLI metric saving message accurate: print the actual output filenames of the metrics that are saved.

Previously, only the dirname of the eigen values was being printed (if it existed), printing the default saving directory name otherwise. However, if the metrics' output filenames were specified using the `--out_*` output optional arguments, the above message would result inaccurate/misleading.


jhlegarreta:
`codecov/patch` failure is unrelated.

skoudoro (approved):
make sense. LGTM, thanks @jhlegarreta

PR #2669, opened 2022-10-29
DOC: Remove inaccurate `patch2self` docstring default values

jhlegarreta:
Remove inaccurate `patch2self` docstring default values: rely on the default values in the method prototype for documentation.


skoudoro:
Thanks !  merging

PR #2670, opened 2022-10-29
ENH: Allow non-default parameters to `Patch2Self` CLI

jhlegarreta:
Allow specifying non-default parameters to `Patch2Self` CLI.

### DIFF
--- a/dipy/workflows/denoise.py
+++ b/dipy/workflows/denoise.py
@@ -21,8 +21,10 @@ class Patch2SelfFlow(Workflow):
     def get_short_name(cls):
         return 'patch2self'
 
-    def run(self, input_files, bval_files, model='ols', verbose=False,
-            out_dir='', out_denoised='dwi_patch2self.nii.gz'):
+    def run(self, input_files, bval_files, model='ols',
+            b0_threshold=50, alpha=1.0, verbose=False, b0_denoising=True,
+            clip_negative_vals=False, shift_intensity=True, out_dir='',
+            out_denoised='dwi_patch2self.nii.gz'):
         """Workflow for Patch2Self denoising method.
 
         It applies patch2self denoising on each file found by 'globing'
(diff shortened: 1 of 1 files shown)


jhlegarreta:
@ShreyasFadnavis @skoudoro See if this makes sense: I ignore whether these were left out purposely when adding the workflow; I assume it is safe to add them as they are/no special care needs to be taken depending on other parts of the workflow/parameter values. No new tests were added. The patch radius is not added as I think lists are not correctly dealt by the workflows yet.

ShreyasFadnavis:
Hi @jhlegarreta ! Thanks a lot for this. I think the PR is good to go 😄 @skoudoro -- can you take a quick look and see if anything else is needed? If not, I can merge this on Monday morning.

skoudoro (commented):
Can you update the tests @jhlegarreta? 

After that, it should be good to go.

ShreyasFadnavis:
> Can you update the tests @jhlegarreta? 
> 
> 
> 
> After that, it should be good to go.

Not sure how we would test this apart from what we already have, since it's just exposing the options available...

skoudoro:
@ShreyasFadnavis, after this change, CI's are failing, not sure why. 

Maybe one of the parameter (like `out_dtype=None`) can not be `None` because it is not handle by your algorithms. 

But I will let @jhlegarreta to look deeper at the default parameters and see why tests are failing

jhlegarreta:
> Maybe one of the parameter (like out_dtype=None) can not be None because it is not handle by your algorithms.

The issue is not the `None` value in itself: other workflows do contain some `None` values by default for other parameters, e.g.
`SlrWithQbxFlow`, `SlrWithQbxFlow`, `ReconstDtiFlow`, `ReconstCSAFlow`, `ReconstIvimFlow`, or `MedianOtsuFlow`, and are tested in `DummyWorkflowOptionalStr` (if I read this correctly).

> But I will let @jhlegarreta to look deeper at the default parameters and see why tests are failing

The real issue lies in the assumption that all `out_` parameters correspond to filenames (except `out_dir` which is parsed to a dirname separately). Finding a fix requires some discussion and time: issue #2671.

In the meantime, I've removed the `out_dtype` parameter from the workflow.

skoudoro (approved):
LGTM, I will let @ShreyasFadnavis to have a last look and merge this PR. 

Thanks @jhlegarreta

skoudoro:
And thanks for the issue too!

ShreyasFadnavis:
Thanks @jhlegarreta ! Also thanks @skoudoro :) 

LGTM :) I will wait till tomorrow evening to merge this, in case anyone else has any other comments...

ShreyasFadnavis:
Thanks @jhlegarreta @skoudoro and @araikes -- Merging!

PR #2688, opened 2022-12-07
DOC: Document Dirac delta generation method missing member

jhlegarreta:
Document Dirac delta generation method missing member.

### DIFF
--- a/dipy/reconst/shm.py
+++ b/dipy/reconst/shm.py
@@ -146,6 +146,10 @@ def gen_dirac(m, n, theta, phi, legacy=True):
         The polar (colatitudinal) coordinate.
     phi : float [0, 2*pi]
         The azimuthal (longitudinal) coordinate.
+    legacy: bool, optional
+        If true, uses DIPY's legacy descoteaux07 implementation (where |m|
+        is used for m < 0). Else, implements the basis as defined in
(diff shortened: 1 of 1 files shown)


dipy/reconst/shm.py
```diff
@@ -146,6 +146,10 @@ def gen_dirac(m, n, theta, phi, legacy=True):
         The polar (colatitudinal) coordinate.
     phi : float [0, 2*pi]
         The azimuthal (longitudinal) coordinate.
+    legacy: bool, optional
+        If true, uses DIPY's legacy descoteaux07 implementation (where |m|
+        is used for m < 0). Else, implements the basis as defined in
+        Descoteaux et al. 2007 (without the absolute value).
```
arokem, line 152:
```suggestion
        Descoteaux et al. 2007 (without the absolute value). Default: True.
```

jhlegarreta: I would prefer not to add this, as it adds a maintenance burden and increases the risk of inconsistencies in case the default value changes in the method signature, but the value is not changed in the documentation. Do you agree @arokem ?

arokem: That's fine. I thought this was required by the numpy docstring standard, but I see that it is merely an option: https://numpydoc.readthedocs.io/en/latest/format.html#parameters

PR #2690, opened 2022-12-07
STYLE: Remove unnecessary b-val print in CSD reconstruction flow

jhlegarreta:
Remove unnecessary b-value threshold/min value print in CSD reconstruction flow.


arokem:
Yep! LGTM.

PR #2694, opened 2022-12-11
NF: Add RUMBA-SD reconstruction workflow

jhlegarreta:
Add RUMBA-SD reconstruction workflow.

Add the accompanying test.

### DIFF
--- a/dipy/workflows/reconst.py
+++ b/dipy/workflows/reconst.py
@@ -8,7 +8,7 @@ from warnings import warn
 import nibabel as nib
 
 from dipy.core.gradients import gradient_table
-from dipy.data import default_sphere
+from dipy.data import default_sphere, get_sphere
 from dipy.io.gradients import read_bvals_bvecs
 from dipy.io.peaks import save_peaks, peaks_to_niftis
 from dipy.io.image import load_nifti, save_nifti, load_nifti_data
@@ -24,6 +24,7 @@ from dipy.reconst.shm import CsaOdfModel
 from dipy.workflows.workflow import Workflow
 from dipy.reconst.dki import DiffusionKurtosisModel, split_dki_param
 from dipy.reconst.ivim import IvimModel
+from dipy.reconst.rumba import RumbaSDModel
 
 from dipy.reconst import mapmri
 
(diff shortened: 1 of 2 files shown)


jhlegarreta:
A few notes:
- The workflow does not contain all possible workflow paths to compute the RUMBA-SD fODFs, e.g. the global fit is not currently possible.
- The FRF computation is done using the single-shell, single-tissue model, and the recursive response or the multi-shell, multi-tissue workflows are not implemented.
- The test is limited to a single set of parameters, so there is no guarantee that the workflow will not fail if using another set of parameters.

However, I think that it is better to have at least a working version, and to add features as required/time permits/fix issues as they are discovered without attempting to have a complete model from the beginning.

jhlegarreta:
`segment/tests/test_mrf.py::test_icm_square` is failing on and off lately. It would need some attention:
https://github.com/dipy/dipy/actions/runs/3670429923/jobs/6205950456#step:9:4473

jhlegarreta:
The test enters the list of slowest tests: it adds 2m30s:
https://github.com/dipy/dipy/actions/runs/3670429918/jobs/6206230576#step:9:3734

So maybe we should pick a different dataset and/or a slice as in the example. If anybody has a suggestion of a ROI that would serve better the purpose, I'm ready to use it.

dipy/workflows/reconst.py
```diff
@@ -1032,3 +1033,193 @@ def get_fitted_ivim(self, data, mask, bval, bvec, b0_threshold=50):
...
+        sphere : Sphere, optional
+            Sphere on which to construct fODF. If None, uses `repulsion724`.
+        verbose : bool, optional
+            If true, logs updates on estimated signal-to-noise ratio after each
+            iteration. This only takes effect in a global fit (`voxelwise` is
+            set to `False`).
+        relative_peak_threshold : float
+            Only return peaks greater than ``relative_peak_threshold * m`` where m
+            is the largest peak.
+        min_separation_angle : float
```
skoudoro, line 1135: `, optional`

dipy/workflows/reconst.py
```diff
@@ -1032,3 +1033,193 @@ def get_fitted_ivim(self, data, mask, bval, bvec, b0_threshold=50):
...
+        b0_threshold : float, optional
+            Threshold used to find b0 volumes.
+        bvecs_tol : float, optional
+            Bvecs should be unit vectors.
+        roi_center : variable int, optional
+            Center of ROI in data. If center is None, it is assumed that it is
+            the center of the volume with shape `data.shape[:3]`.
+        roi_radii : int or array-like, optional
+            radii of cuboid ROI in voxels.
+        fa_thr : float
```
skoudoro, line 1081: `, optional`

dipy/workflows/reconst.py
```diff
@@ -1032,3 +1033,193 @@ def get_fitted_ivim(self, data, mask, bval, bvec, b0_threshold=50):
...
+            integer.
+        voxelwise : bool, optional
+            If true, performs a voxelwise fit. If false, performs a global fit
+            on the entire brain at once. The global fit requires a 4D brain
+            volume in `fit`.
+        use_tv : bool, optional
+            If true, applies total variation regularization. This only takes
+            effect in a global fit (`voxelwise` is set to `False`). TV can only
+            be applied to 4D brain volumes with no singleton dimensions.
+        sphere : Sphere, optional
```
skoudoro, line 1126: not sure what `Sphere`means for the user.  sphere_name so `str` ?

dipy/workflows/reconst.py
```diff
@@ -1032,3 +1033,193 @@ def get_fitted_ivim(self, data, mask, bval, bvec, b0_threshold=50):
...
+
+            if b0_threshold < bvals.min():
+                warn("b0_threshold (value: {0}) is too low, increase your "
+                     "b0_threshold. It should be higher than the first b0 value "
+                     "({1}).".format(b0_threshold, bvals.min()))
+
+            gtab = gradient_table(
+                bvals, bvecs, b0_threshold=b0_threshold, atol=bvecs_tol)
+
+            sphere = sphere if sphere else default_sphere
```
skoudoro, line 1186: Still not sure what you expect as a sphere for the user

dipy/workflows/reconst.py
```diff
@@ -1032,3 +1033,193 @@ def get_fitted_ivim(self, data, mask, bval, bvec, b0_threshold=50):
...
+                gtab, wm_response=wm_response[0], gm_response=gm_response,
+                csf_response=csf_response, n_iter=n_iter,
+                recon_type=recon_type, n_coils=n_coils, R=R,
+                voxelwise=voxelwise, use_tv=use_tv, sphere=sphere,
+                verbose=verbose)
+
+            rumba_peaks = peaks_from_model(
+                model=rumba, data=data, sphere=sphere,
+                relative_peak_threshold=relative_peak_threshold,
+                min_separation_angle=min_separation_angle, mask=None,
```
skoudoro, line 1204: `mask=None` ?  you might want to load `maskfile` and pass it here

dipy/workflows/reconst.py
```diff
@@ -1032,3 +1033,193 @@ def get_fitted_ivim(self, data, mask, bval, bvec, b0_threshold=50):
...
+                     "({1}).".format(b0_threshold, bvals.min()))
+
+            gtab = gradient_table(
+                bvals, bvecs, b0_threshold=b0_threshold, atol=bvecs_tol)
+
+            sphere = sphere if sphere else default_sphere
+
+            # Compute the FRF
+            wm_response, _ = auto_response_ssst(
+                gtab, data, roi_center=roi_center, roi_radii=roi_radii,
```
skoudoro, line 1190: here, I suppose it will fail if `roi_center` is an `int` like you propose in the docstring

skoudoro (commented):
Hi @jhlegarreta,

Thank you for this! I need to run it locally but here is a first review.

dipy/workflows/reconst.py
```diff
@@ -1032,3 +1033,193 @@ def get_fitted_ivim(self, data, mask, bval, bvec, b0_threshold=50):
...
+            If true, applies total variation regularization. This only takes
+            effect in a global fit (`voxelwise` is set to `False`). TV can only
+            be applied to 4D brain volumes with no singleton dimensions.
+        sphere : Sphere, optional
+            Sphere on which to construct fODF. If None, uses `repulsion724`.
+        verbose : bool, optional
+            If true, logs updates on estimated signal-to-noise ratio after each
+            iteration. This only takes effect in a global fit (`voxelwise` is
+            set to `False`).
+        relative_peak_threshold : float
```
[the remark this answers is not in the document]
jhlegarreta: Done in f4e10d3.

dipy/workflows/reconst.py
jhlegarreta: Done in f4e10d3.

dipy/workflows/reconst.py
jhlegarreta: Done in f4e10d3.

dipy/workflows/reconst.py
jhlegarreta: You're right, my mistake, I changed it to a `sphere_name`, `str` and defaulted to `repulsion724` in f4e10d3.

dipy/workflows/reconst.py
jhlegarreta: Addressed in f4e10d3.

dipy/workflows/reconst.py
jhlegarreta: It's a `variable int`; whatever we mean by that: I borrowed the wording from the CSD workflow.

skoudoro: I agree that `variable int` is obscure.  At some point, we need to update the workflow framework and use the `typing` module.  `variable int` = `List[int]` which make more sense.

In general, we should start to introduce some typing in DIPY.

jhlegarreta: I agree with this. And to define as many docstrings as possible in a single place and to borrow them from there.

dipy/workflows/reconst.py
```diff
@@ -1032,3 +1033,198 @@ def get_fitted_ivim(self, data, mask, bval, bvec, b0_threshold=50):
...
+            many cores as possible). 0 raises an error.
+        gm_response : float, optional
+            Mean diffusivity for GM compartment. If `None`, then grey
+            matter volume fraction is not computed.
+        csf_response : float, optional
+            Mean diffusivity for CSF compartment. If `None`, then CSF
+            volume fraction is not computed.
+        n_iter : int, optional
+            Number of iterations for fODF estimation. Must be a positive int.
+        recon_type : {'smf', 'sos'}, optional
```
skoudoro, line 1109: `str,` instead of `{'smf', 'sos'}`. you can add this in the docstring

jhlegarreta: Done in 8603b93.

skoudoro:
Thank you @jhlegarreta

dipy/workflows/reconst.py
jhlegarreta: Resolving. These issues are more appropriate for a dedicated thread/issue/PR.

PR #2697, opened 2022-12-13
NF: Adds a fetcher for HCP data.

### LINKED ISSUE #2696: Port HCP fetcher from pyAFQ into here
Use this code: https://github.com/yeatmanlab/pyAFQ/blob/master/AFQ/data/fetch.py#L1367-L1483

The fetcher requires that users set up their own access to HCP data on AWS, the documentation includes an explanation and links to that.

arokem:
Closes #2696.

### DIFF
--- a/dipy/data/fetcher.py
+++ b/dipy/data/fetcher.py
@@ -1761,9 +1761,13 @@ def fetch_hcp(subjects,
                            f'sub-{subject}_aparc+aseg_seg.nii.gz')] =\
             f'{study}/{subject}/T1w/aparc+aseg.nii.gz'
 
-    for k in data_files.keys():
-        if not op.exists(k):
-            bucket.download_file(data_files[k], k)
+    with tqdm(total=len(data_files.keys())) as pbar:
+        for k in data_files.keys():
+            pbar.set_description_str(f"Downloading {k}")
+            if not op.exists(k):
+                bucket.download_file(data_files[k], k)
+            pbar.update()
+
     # Create the BIDS dataset description file text
     hcp_acknowledgements = """Data were provided by the Human Connectome Project, WU-Minn Consortium (Principal Investigators: David Van Essen and Kamil Ugurbil; 1U54MH091657) funded by the 16 NIH Institutes and Centers that support the NIH Blueprint for Neuroscience Research; and by the McDonnell Center for Systems Neuroscience at Washington University.""",  # noqa
     to_bids_description(pjoin(my_path, study),
--- a/dipy/data/fetcher.py
+++ b/dipy/data/fetcher.py
@@ -1671,7 +1671,7 @@ def fetch_hcp(subjects,
     profile_name : string, optional
         The name of the AWS profile used for access. Default: "hcp"
     path : string, optional
-        Path to save files into. Default: '~/AFQ_data'
+        Path to save files into. Default: '~/.dipy'
     study : string, optional
         Which HCP study to grab. Default: 'HCP_1200'
     aws_access_key_id : string, optional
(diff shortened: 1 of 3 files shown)


dipy/data/fetcher.py
```diff
@@ -1628,8 +1635,141 @@ def read_five_af_bundles():
...
+
+    Returns
+    -------
+    dict with remote and local names of these files,
+    path to BIDS derivative dataset
+
+    Notes
+    -----
+    To use this function with its default setting, you need to have a
+    file '~/.aws/credentials', that includes a section:
```
skoudoro, line 1692: What will be the path on Windows?

dipy/data/fetcher.py
```diff
@@ -1628,8 +1635,141 @@ def read_five_af_bundles():
...
+    Parameters
+    ----------
+    subjects : list
+        Each item is an integer, identifying one of the HCP subjects
+    hcp_bucket : string, optional
+        The name of the HCP S3 bucket. Default: "hcp-openaccess"
+    profile_name : string, optional
+        The name of the AWS profile used for access. Default: "hcp"
+    path : string, optional
+        Path to save files into. Default: '~/AFQ_data'
```
skoudoro, line 1674: default path to update

skoudoro (commented):
Hi @arokem,

Thank you for this, it works like a charm. We might need to create a tutorial later in case this [link](https://wiki.humanconnectome.org/display/PublicData/How+To+Connect+to+Connectome+Data+via+AWS) stop to work.

is it possible to add `tqdm` during the download? it takes some times and we have currently no information if it is downloading or freezing or whatever....

Same when the data already exist in our disk, it would great to add a message like we did with the other fetcher. 

For Windows, where should we put the credentials ?

skoudoro:
Also, CI's are currently failing because you need to add the fetcher in our workflow list. see here: https://github.com/dipy/dipy/blob/master/dipy/workflows/tests/test_io.py#L55

dipy/data/fetcher.py
```diff
@@ -1628,8 +1635,141 @@ def read_five_af_bundles():
...
+              aws_access_key_id=None,
+              aws_secret_access_key=None):
+    """
+    Fetch HCP diffusion data and arrange it in a manner that resembles the
+    BIDS [1]_ specification.
+
+    Parameters
+    ----------
+    subjects : list
+        Each item is an integer, identifying one of the HCP subjects
```
skoudoro, line 1668: can you add a example here, I made the stupid mistake to just add an int instead of a list[int]

dipy/data/fetcher.py
arokem: This is what AWS has to say about that: https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html#cli-configure-files-where, but I am not sure that I understand it. What's the deal with `%UserProfile%`? Maybe I should just include a link to this webpage?

dipy/data/fetcher.py
arokem: Whoops. Updated in d6662003

dipy/data/fetcher.py
arokem: Instead of adding the example, I added a bit of code to recover from the case where users provide just one subject (either as integer or as string), because I think that's a typical failure mode.

arokem:
I added a progress bar in [c0bdd5e](https://github.com/dipy/dipy/pull/2697/commits/c0bdd5eb039e0847c8ac21ed46ae6b6b7a11462e). It's a bit coarse, documenting the relative number of files downloaded, instead of relative number of bytes.

skoudoro:
> I added a progress bar in [c0bdd5e](https://github.com/dipy/dipy/pull/2697/commits/c0bdd5eb039e0847c8ac21ed46ae6b6b7a11462e). It's a bit coarse, documenting the relative number of files downloaded, instead of relative number of bytes.

much better than nothing 👍🏾

dipy/data/fetcher.py
skoudoro: yes, can you add this link ?

skoudoro (approved):
Thank you for this @arokem. I am going ahead and merge it

PR #2698, opened 2022-12-14
Replace CENIR with HBN

arokem:
For now, I have only addressed the first part of this, which is to create a fetcher for preprocessed HBN data. 

This is implemented on top of #2696, because it uses a function introduced there.

### DIFF
--- a/dipy/data/fetcher.py
+++ b/dipy/data/fetcher.py
@@ -1793,7 +1793,8 @@ def fetch_hbn(subjects, path=None):
     ----------
     subjects : list
         Identifiers of the subjects to download.
-        For example: ["NDARAA112DMH", "NDARAA117NEJ"].
+        For example: ["NDARAA948VFH", "NDAREK918EC2"].
+
     path : string, optional
         Path to save files into. Default: '~/.dipy'
 
@@ -1819,8 +1820,8 @@ def fetch_hbn(subjects, path=None):
         from botocore import UNSIGNED
         from botocore.client import Config
     else:
-        TripWire("The `fetch_hbn` function requires the boto3",
-                 "library, but that is not installed.")
+        TripWire("The `fetch_hbn` function requires the boto3" +
+                 " library, but that is not installed.")
 
     # Anonymous access:
     client = boto3.client('s3', config=Config(signature_version=UNSIGNED))
--- a/dipy/data/fetcher.py
+++ b/dipy/data/fetcher.py
@@ -20,7 +20,7 @@ from dipy.core.gradients import (gradient_table,
 from dipy.io.gradients import read_bvals_bvecs
 from dipy.io.image import load_nifti, load_nifti_data, save_nifti
 from dipy.io.streamline import load_trk
-from dipy.utils.optpkg import optional_package
+from dipy.utils.optpkg import optional_package, TripWire
 
 
 from urllib.request import urlopen
@@ -1814,6 +1814,14 @@ def fetch_hbn(subjects, path=None):
         Scientific Data. 2022;9(1):1-27.
 
     """
+
+    if has_boto3:
+        from botocore import UNSIGNED
+        from botocore.client import Config
+    else:
+        TripWire("The `fetch_hbn` function requires the boto3",
+                 "library, but that is not installed.")
+
     # Anonymous access:
     client = boto3.client('s3', config=Config(signature_version=UNSIGNED))
 
@@ -1863,7 +1871,7 @@ def fetch_hbn(subjects, path=None):
         with tqdm(total=len(download_files.keys())) as pbar:
             for k in download_files.keys():
                 pbar.set_description_str(f"Downloading {k}")
-                bucket.download_file(data_files[k], k)
+                client.download_file("fcp-indi", download_files[k], k)
                 pbar.update()
 
     # Create the BIDS dataset description file text
--- a/dipy/data/fetcher.py
+++ b/dipy/data/fetcher.py
@@ -1785,9 +1785,9 @@ def fetch_hcp(subjects,
     return data_files, pjoin(my_path, study)
 
 
-def fetch_hbn_preproc(subjects, path=None):
+def fetch_hbn(subjects, path=None):
     """
-    Fetches data from the Healthy Brain Network POD2 study [1, 2]_.
+    Fetches preprocessed data from the Healthy Brain Network POD2 study [1, 2]_.
 
     Parameters
     ----------
--- a/dipy/data/fetcher.py
+++ b/dipy/data/fetcher.py
@@ -1787,7 +1787,7 @@ def fetch_hcp(subjects,
 
 def fetch_hbn(subjects, path=None):
     """
-    Fetches preprocessed data from the Healthy Brain Network POD2 study [1, 2]_.
+    Fetch preprocessed data from the Healthy Brain Network POD2 study [1, 2]_.
 
     Parameters
     ----------
@@ -1795,7 +1795,7 @@ def fetch_hbn(subjects, path=None):
         Identifiers of the subjects to download.
         For example: ["NDARAA112DMH", "NDARAA117NEJ"].
     path : string, optional
-        Path to save files into. Default: '~/AFQ_data'
+        Path to save files into. Default: '~/.dipy'
 
     Returns
     -------
--- a/dipy/data/fetcher.py
+++ b/dipy/data/fetcher.py
@@ -1855,12 +1855,16 @@ def fetch_hbn(subjects, path=None):
             local = op.join(dipy_home, full)
             data_files[local] = remote
 
-    with tqdm(total=len(data_files.keys())) as pbar:
-        for k in data_files.keys():
-            pbar.set_description_str(f"Downloading {k}")
-            if not op.exists(k):
-                client.download_file("fcp-indi", data_files[k], k)
-            pbar.update()
+    download_files = {}
+    for k in data_files.keys():
+        if not op.exists(k):
+            download_files[k] = data_files[k]
+    if len(download_files.keys()):
+        with tqdm(total=len(download_files.keys())) as pbar:
+            for k in download_files.keys():
+                pbar.set_description_str(f"Downloading {k}")
+                bucket.download_file(data_files[k], k)
+                pbar.update()
 
     # Create the BIDS dataset description file text
     to_bids_description(op.join(my_path, "HBN"),
--- a/dipy/data/fetcher.py
+++ b/dipy/data/fetcher.py
@@ -1814,7 +1814,6 @@ def fetch_hbn_preproc(subjects, path=None):
         Scientific Data. 2022;9(1):1-27.
 
     """
-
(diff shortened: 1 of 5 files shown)


skoudoro (commented):
LGTM, thanks @arokem.

I think #2697 should be merged before this PR.

Concerning the failing CI's, you need to add the fetcher in our workflow list. see here: https://github.com/dipy/dipy/blob/master/dipy/workflows/tests/test_io.py#L55

Also, same comment as #2697, it would be great to have `tqdm`during the download and if the file is already downloaded, it would be great to have a message, like the current behavior of our fetcher. 

Thanks !

skoudoro:
#2697 has been merged. You can rebase this PR, add tqdm and add the fetcher in our workflow list. Thanks!

arokem:
Yep. Just pushed that.

dipy/data/fetcher.py
```diff
@@ -1782,3 +1779,88 @@ def fetch_hcp(subjects,
                            "PipelineDescription": {'Name': 'hcp_pipeline'}})
 
     return data_files, pjoin(my_path, study)
+
+
+def fetch_hbn_preproc(subjects, path=None):
```
skoudoro, line 1784: Can you rename it `fetch_hbn`?   You can add the `preproc` information in the docstring. 

Also, workflow is failing because `hbn` != `hbn_preproc`

dipy/data/fetcher.py
```diff
@@ -1742,22 +1742,19 @@ def fetch_hcp(subjects,
...
     for subject in subjects:
-        # We make a single session folder per subject for this case, because
-        # AFQ api expects session structure:
         sub_dir = pjoin(base_dir, f'sub-{subject}')
-        sess_dir = pjoin(sub_dir, "ses-01")
         if not op.exists(sub_dir):
-            os.makedirs(pjoin(sess_dir, 'dwi'), exist_ok=True)
-            os.makedirs(pjoin(sess_dir, 'anat'), exist_ok=True)
-        data_files[pjoin(sess_dir, 'dwi', f'sub-{subject}_dwi.bval')] =\
+            os.makedirs(pjoin(sub_dir, 'dwi'), exist_ok=True)
```
skoudoro, line 1747: good catch

dipy/data/fetcher.py
```diff
@@ -1782,3 +1779,88 @@ def fetch_hcp(subjects,
...
+        my_path = path
+
+    base_dir = op.join(my_path, "HBN", 'derivatives', 'qsiprep')
+
+    if not os.path.exists(base_dir):
+        os.makedirs(base_dir, exist_ok=True)
+
+    data_files = {}
+
+    for subject in subjects:
```
skoudoro, line 1830: Like hcp, I recommend to check subjects if it is just a `str` to avoid mistake

dipy/workflows/tests/test_io.py
```diff
@@ -63,7 +63,8 @@ def test_io_fetch_fetcher_datanames():
                      'stanford_labels', 'stanford_pve_maps', 'stanford_t1',
                      'syn_data', 'taiwan_ntu_dsi', 'target_tractogram_hcp',
                      'tissue_data', 'qte_lte_pte', 'resdnn_weights',
-                     'DiB_217_lte_pte_ste', 'DiB_70_lte_pte_ste', "hcp"]
+                     'DiB_217_lte_pte_ste', 'DiB_70_lte_pte_ste', "hcp",
+                     "hbn"]
```
skoudoro, line 67: `hbn` or `hbn_preproc`. it should be the same name as the function

skoudoro (commented):
Hi @arokem,

This PR is almost ready to be merged. Can you address the small comments below?

arokem:
Hey Serge - unfortunately, this will require a bit more work, because I also need to figure out how to do anonymous downloads from S3 using boto3. I am working on it...

arokem:
Could the test errors now appearing here be related to the recent release of a new version of numpy yesterday?

skoudoro:
> Could the test errors now appearing here be related to the recent release of a new version of numpy yesterday?

it seems so

arokem:
This now includes replacing CENIR in some of the documentation examples. I believe it's ready for review/merge.

arokem:
Note that CENIR is still used in some places. Specifically, in the denoising example (I believe that the HBN data is denoised during processing).

dipy/data/fetcher.py
```diff
@@ -1786,3 +1783,105 @@ def fetch_hcp(subjects,
...
+    """
+    Fetches preprocessed data from the Healthy Brain Network POD2 study [1, 2]_.
+
+    Parameters
+    ----------
+    subjects : list
+        Identifiers of the subjects to download.
+        For example: ["NDARAA112DMH", "NDARAA117NEJ"].
+    path : string, optional
+        Path to save files into. Default: '~/AFQ_data'
```
skoudoro, line 1798: "DIPY_HOME" instead of "AFQ_data"

dipy/data/fetcher.py
```diff
@@ -1786,3 +1783,105 @@ def fetch_hcp(subjects,
                            "PipelineDescription": {'Name': 'hcp_pipeline'}})
 
     return data_files, pjoin(my_path, study)
+
+
+def fetch_hbn(subjects, path=None):
+    """
+    Fetches preprocessed data from the Healthy Brain Network POD2 study [1, 2]_.
```
skoudoro, line 1790: `Fetch` imperative mood in docstring

dipy/data/fetcher.py
```diff
@@ -1786,3 +1783,105 @@ def fetch_hcp(subjects,
...
+
+    for subject in subjects:
+        initial_query = client.list_objects(
+            Bucket="fcp-indi",
+            Prefix=f"data/Projects/HBN/BIDS_curated/sub-{subject}/")
+        ses = initial_query['Contents'][0]["Key"].split('/')[5]
+        query = client.list_objects(
+            Bucket="fcp-indi",
+            Prefix=f"data/Projects/HBN/BIDS_curated/derivatives/qsiprep/sub-{subject}/{ses}/")  # noqa
+        file_list = [kk["Key"] for kk in query["Contents"]]
```
skoudoro, line 1855: Not sure what you are trying to do here.

In my case, it fails:

```python
In [3]: te = fetch_hbn(["NDARAA112DMH", "NDARAA117NEJ"])
KeyError                                  Traceback (most recent call last)
Cell In[3], line 1
----> 1 te = fetch_hbn(["NDARAA112DMH", "NDARAA117NEJ"])

File ~/devel/dipy/dipy/data/fetcher.py:1855, in fetch_hbn(subjects, path)
   1851 ses = initial_query['Contents'][0]["Key"].split('/')[5]
   1852 query = client.list_objects(
   1853     Bucket="fcp-indi",
   1854     Prefix=f"data/Projects/HBN/BIDS_curated/derivatives/qsiprep/sub-{subject}/{ses}/")  # noqa
-> 1855 file_list = [kk["Key"] for kk in query["Contents"]]
   1856 sub_dir = op.join(base_dir, f'sub-{subject}')
   1857 ses_dir = op.join(sub_dir, ses)

KeyError: 'Contents'
```

skoudoro: Additional information:

here what contains `query`:

```
{'ResponseMetadata': {'RequestId': 'N3CXBA1XSTRTR6HE', 'HostId': 'ju6bxhE4xxTXDU0OO7BfObTIefXF97G58A3EstSWmY8Q49OLB8BIxwiNMGWrlrZtAESHlhfxQc0=', 'HTTPStatusCode': 200, 'HTTPHeaders': {'x-amz-id-2': 'ju6bxhE4xxTXDU0OO7BfObTIefXF97G58A3EstSWmY8Q49OLB8BIxwiNMGWrlrZtAESHlhfxQc0=', 'x-amz-request-id': 'N3CXBA1XSTRTR6HE', 'date': 'Thu, 12 Jan 2023 17:13:44 GMT', 'x-amz-bucket-region': 'us-east-1', 'content-type': 'application/xml', 'transfer-encoding': 'chunked', 'server': 'AmazonS3'}, 'RetryAttempts': 0}, 'IsTruncated': False, 'Marker': '', 'Name': 'fcp-indi', 'Prefix': 'data/Projects/HBN/BIDS_curated/derivatives/qsiprep/sub-NDARAA112DMH/ses-HBNsiteRU/', 'MaxKeys': 1000, 'EncodingType': 'url'}
```

Which is strange because `initial_query` work as expected. So either:
- there is an issue in the path 
- there is a limit, cannot have 2 requests in a row so fast.
- or something else.

dipy/data/fetcher.py
```diff
@@ -1786,3 +1783,105 @@ def fetch_hcp(subjects,
...
+    # If user provided incorrect input, these are typical failures that
+    # are easy to recover from:
+    if isinstance(subjects, int) or isinstance(subjects, str):
+        subjects = [subjects]
+
+    for subject in subjects:
+        initial_query = client.list_objects(
+            Bucket="fcp-indi",
+            Prefix=f"data/Projects/HBN/BIDS_curated/sub-{subject}/")
+        ses = initial_query['Contents'][0]["Key"].split('/')[5]
```
skoudoro, line 1851: To handle potential issue, 
I recommend using  
` ses = initial_query.get('Contents', [])` and check if everything is ok.

skoudoro (changes requested):
Hi @arokem,

Thank you for the update. Can you rebase? Also, I had some issue. See below, I can not fetch the data and tests the tutorial.

It would great if you could have a look today that I can release tomorrow with this fetcher.

Thanks!

dipy/data/fetcher.py
arokem: I think this is because the subject ID doesn't exist in the qsiprep derivatives. I believe this is a subject that doesn't even have raw dwi data! https://fcp-indi.s3.amazonaws.com/index.html#data/Projects/HBN/BIDS_curated/sub-NDARAA112DMH/ses-HBNsiteRU/. 

I'll fix the documentation example to point to a subject that has processed data. We might want to think about how we provide information about the full list of subjects that users can point to (it's a long list!). We also have QC data for all these subjects, which users might want to refer to, so we might also want to add a fetcher to that. It's in a csv file in the same s3 bucket.

dipy/workflows/tests/test_io.py
arokem: I think this is correct now.

dipy/data/fetcher.py
arokem: Addressed in 7e6153ac5

dipy/data/fetcher.py
arokem: Addressed in f974774f4

dipy/data/fetcher.py
arokem: Addressed in 85da8476d, I believe

arokem:
I think that I addressed all the previous comments, but let me know if I missed anything.

skoudoro (approved):
Thank you for the update @arokem.

Merging

PR #2701, opened 2022-12-19
FIX: Don't print a progbar for downloading unless you need to.

arokem:


skoudoro (approved):
Thanks @arokem, merging

Issue #2655, opened 2022-10-06
dipy pca_noise_estimate is wrong?


samcoveney:
## Description
Manjon 2013 paper "Diffusion Weighted Image Denoising Using Overcomplete Local PCA" provides a method to estimate noise in a diffusion weighted image series using PCA. I believe that the DIPY interpretation of this work is incorrect. Apologies in advance if I am wrong.

The function `pca_noise_estimate` uses a data matrix with rows = voxels and columns = acquisitions, which is the formulation used to apply PCA to denoising data. However, for _estimation of noise level_ in the images, the formulation provided by Manjon reads as follows:

> Given a set of n b0 images (n. = 2), we first compute global PCA decomposition of image series (i.e. we perform a PCA decomposition of a matrix formed by as many columns as image voxels and as many rows as images, see Fig. 2) - Manjon 2013

In other words, the voxels are supposed to be in the columns of X. This is why in Manjon 2013, Fig 2 plots the eigenvectors as images, because each eigenvector is an image. And it is why it is suggested to use B0 images if possible, since these images ought to have very few principle components corresponding to features in the data (if there was no noise and no other imaging artefacts etc, then these images would all be the same).

In `pca_noise_estimate`, the PCA is done on X with rows = voxels and columns = acquisitions. The data X is then projected onto these new axes using X.dot(W), and the variance of the coordinates in the new space is evaluated and used to calculate the noise.

I believe this function is therefore wrong, and should be rewritten. I am happy to do this, if we agree that a correction is needed.

skoudoro:
@ShreyasFadnavis, Can you look at this if you have time to answer to @samcoveney? 

Thank you

samcoveney:
I should also point out that if all pixels are used (as they are here),
with the current implementation being that pixels are samples, then N > P
(number of features = images), so the amount of variance in the last
component (i.e. the smallest eigenvalue) will be basically zero (even if
N=P this will be true, because the mean is subtracted from the columns).

This would not be the case for the original idea in Manjon paper, because
the number of images is the number of samples and this is less than the
number of features (pixels).

PR #2707, opened 2023-01-04
Fixdenoise

samcoveney:
Following on from #2678 I have attempted to do several things:
1) fix `genpca` to work when the number of samples (voxels) is less than the number of features (images) by modifying `_pca_classifier`
2) fix `pca_noise_estimate` to use the canonical algorithm (which was that specified in the function docs), although I have left the original DiPy method available
3) modified `localpca` to use the `pca_noise_estimate` (which would make this function canonical) and to raise a warning if an estimate of sigma is provided to this function

Comments and thoughts welcome. _Especially_ where I have made some modifications of estimated values that attempt to take account of the centering of the data matrix used for PCA (there are some NOTEs left in the code highlighting areas that may need looking at, I can always change things back).

### DIFF
--- a/dipy/denoise/localpca.py
+++ b/dipy/denoise/localpca.py
@@ -1,5 +1,9 @@
+from packaging.version import Version
 from warnings import warn
+
 import numpy as np
+import scipy
+
 try:
     from scipy.linalg.lapack import dgesvd as svd
     svd_args = [1, 0]
@@ -202,6 +206,8 @@ def genpca(arr, sigma=None, mask=None, patch_radius=2, pca_method='eig',
         var = np.zeros(arr.shape[:-1], dtype=calc_dtype)
         thetavar = np.zeros(arr.shape[:-1], dtype=calc_dtype)
 
+    SCIPY_LESS_1_5_0 = Version(scipy.__version__) < Version('1.5.0')
+    kw_eigh = {'turbo': True} if SCIPY_LESS_1_5_0 else {}  # {'driver': 'gvd'}
     # loop around and find the 3D patch for each direction at each pixel
     for k in range(patch_radius[2], arr.shape[2] - patch_radius[2]):
         for j in range(patch_radius[1], arr.shape[1] - patch_radius[1]):
@@ -238,7 +244,7 @@ def genpca(arr, sigma=None, mask=None, patch_radius=2, pca_method='eig',
                     # PCA using an Eigenvalue decomposition
                     C = np.transpose(X).dot(X)
                     C = C / X.shape[0]
-                    [d, W] = eigh(C, turbo=True)
+                    [d, W] = eigh(C, **kw_eigh)
 
                 if sigma is None:
                     # Random matrix theory
--- a/dipy/denoise/localpca.py
+++ b/dipy/denoise/localpca.py
@@ -36,7 +36,13 @@ def _pca_classifier(L, nvoxels):
 
     Notes
     -----
-    This is based on the algorithm described in [1]_.
+    This is based on the algorithm described in [1]_. Note that the
+    condition expressed in the while-loop is expressed in terms of the
+    variance of equation (12), not equation (11) as in [1]_. Also,
+    this code implements ascending eigenvalues, unlike [1]_.
+    Also unlike [1]_, the data matrix X is assumed to be centered,
+    therefore nvoxels has been replaced with nvoxels - 1 to account
+    for the lost degree of freedom.
 
     References
     ----------
@@ -49,11 +55,10 @@ def _pca_classifier(L, nvoxels):
     # if num_samples - 1 (to correct for mean subtraction) is less than number
     # of features, discard the zero eigenvalues
     if L.size > nvoxels - 1:
-        L = L[-(nvoxels - 1):] 
+        L = L[-(nvoxels - 1):]
 
     var = np.mean(L)
     c = L.size - 1
-    # NOTE: data is centered, at most (nvoxels - 1) non-zero eigenvalues
     r = L[c] - L[0] - 4 * np.sqrt((c + 1.0) / (nvoxels - 1)) * var
     while r > 0:
         var = np.mean(L[:c])
@@ -85,7 +90,7 @@ def genpca(arr, sigma=None, mask=None, patch_radius=2, pca_method='eig',
         outside of those voxels.
     patch_radius : int or 1D array (optional)
         The radius of the local patch to be taken around each voxel (in
-        voxels). Default: 2 (denoise in blocks of 5x5x5 voxels).
+        voxels). E.g. patch_radius=2 gives 5x5x5 patches.
     pca_method : 'eig' or 'svd' (optional)
         Use either eigenvalue decomposition (eig) or singular value
         decomposition (svd) for principal component analysis. The default
@@ -102,16 +107,13 @@ def genpca(arr, sigma=None, mask=None, patch_radius=2, pca_method='eig',
         \tau_{factor} can be set to a predefined values (e.g. \tau_{factor} =
         2.3 [3]_), or automatically calculated using random matrix theory
         (in case that \tau_{factor} is set to None).
-        Default: None.
     return_sigma : bool (optional)
         If true, the Standard deviation of the noise will be returned.
-        Default: False.
     out_dtype : str or dtype (optional)
         The dtype for the output array. Default: output has the same dtype as
         the input.
     suppress_warning : bool (optional)
         If true, suppress warning caused by patch_size < arr.shape[-1].
-        Default: False.
 
     Returns
     -------
@@ -175,8 +177,8 @@ def genpca(arr, sigma=None, mask=None, patch_radius=2, pca_method='eig',
     if num_samples == 1:
         raise ValueError("Cannot have only 1 sample,\
                           please increase patch_radius.")
-    # NOTE: account for mean substraction by testing #samples - 1
-    if num_samples - 1 < arr.shape[-1] and not suppress_warning:
+    # account for mean substraction by testing #samples - 1
+    if (num_samples - 1) < arr.shape[-1] and not suppress_warning:
         tmp = np.sum(patch_size == 1)  # count spatial dimensions with size 1
         if tmp == 0:
             root = np.ceil(arr.shape[-1] ** (1./3))  # 3D
@@ -207,7 +209,7 @@ def genpca(arr, sigma=None, mask=None, patch_radius=2, pca_method='eig',
 
     dim = arr.shape[-1]
     if tau_factor is None:
-        tau_factor = 1 + np.sqrt(dim / (num_samples - 1)) # NOTE: account for mean subtraction
+        tau_factor = 1 + np.sqrt(dim / num_samples)
 
     theta = np.zeros(arr.shape, dtype=calc_dtype)
     thetax = np.zeros(arr.shape, dtype=calc_dtype)
@@ -258,7 +260,7 @@ def genpca(arr, sigma=None, mask=None, patch_radius=2, pca_method='eig',
 
                 if sigma is None:
                     # Random matrix theory
-                    this_var, ncomps = _pca_classifier(d, num_samples)  # NOTE: ncomps not usable as index if num_samples < num_features, but is recalculated below anyway 
+                    this_var, _ = _pca_classifier(d, num_samples)
                 else:
                     # Predefined variance
                     this_var = var[i, j, k]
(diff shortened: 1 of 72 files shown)


dipy/denoise/localpca.py
```diff
@@ -40,14 +41,22 @@ def _pca_classifier(L, nvoxels):
...
     var = np.mean(L)
     c = L.size - 1
-    r = L[c] - L[0] - 4 * np.sqrt((c + 1.0) / nvoxels) * var
+    # NOTE: data is centered, at most (nvoxels - 1) non-zero eigenvalues
+    r = L[c] - L[0] - 4 * np.sqrt((c + 1.0) / (nvoxels - 1)) * var
     while r > 0:
         var = np.mean(L[:c])
         c = c - 1
-        r = L[c] - L[0] - 4 * np.sqrt((c + 1.0) / nvoxels) * var
+        r = L[c] - L[0] - 4 * np.sqrt((c + 1.0) / (nvoxels - 1)) * var
```
jhlegarreta, line 57: Following https://github.com/dipy/dipy/discussions/2678#discussioncomment-4336266, https://github.com/dipy/dipy/discussions/2678#discussioncomment-4342309 on the number of directions vs. voxels, and Rafael's comment
>  You can easily rewrite the equation in Veraart et al., 2016 to match the code in function _pca_classifier.

maybe such re-arrangements could be explicitly put allowing to identify the equations. Otherwise, without writing those oneself, looks difficult to review. It also requires to go through the issue comments, which is unlikely to happen if one simply opens the DIPY documentation.

dipy/denoise/localpca.py
```diff
@@ -242,7 +252,7 @@ def genpca(arr, sigma=None, mask=None, patch_radius=2, pca_method='eig',
 
                 if sigma is None:
                     # Random matrix theory
-                    this_var, ncomps = _pca_classifier(d, num_samples)
+                    this_var, ncomps = _pca_classifier(d, num_samples)  # NOTE: ncomps not usable as index if num_samples < num_features, but is recalculated below anyway
```
jhlegarreta, line 255: Generally, if the value  `ncomps` is not used here, it could be substituted by `_`.

Then, and having a look at a few lines below, the returned `ncomps` is never used, which is questionable. It is only used in the test. Also, following the comment, in the test there is a single voxel, so if that means `num_samples=1`, then I am not sure the value can be tested; alternatively, if
> the PCA is done on X with rows = voxels and columns = acquisitions.

as said in https://github.com/dipy/dipy/issues/2655#issue-1399383100, then again not sure if the wording is accurate (e.g. should B0 be used as suggested in the same issue ?).

A style comment: I would avoid adding in-line comments unless they are short, otherwise the large horizontal space hinders readability. Same goes to all in-line notes above and below.

dipy/denoise/pca_noise_estimate.pyx
```diff
@@ -47,6 +48,9 @@ def pca_noise_estimate(data, gtab, patch_radius=1, correct_bias=True,
     smooth : int
         Radius of a Gaussian smoothing filter to apply to the noise estimate
         before returning. Default: 2.
+    image_as_samples : bool
+        Whether to use images as rows (samples) for PCA (algorithm in [1]_) or
+        to use images as columns (features). Default: True (images as samples).
```
jhlegarreta, line 53: Following

> I don't believe that transposing matrix X is a good idea, since this then turns the voxels into features, and then the algorithms really would be quite different? While this would remove the zero-eigenvalues (for cases where they would appear before the transpose), I don't think this problem is equivalent.

in https://github.com/dipy/dipy/discussions/2678#discussioncomment-4342309, then what would be a use case for the case where the variable value is `false`? (despite https://github.com/dipy/dipy/pull/2707/files#diff-33a3d513671c2a32557d87cc9b4c1ef1c7fb068ed4a4110abb2a222cde658669R103 and having added a test)

dipy/denoise/pca_noise_estimate.pyx
```diff
@@ -92,18 +100,32 @@ def pca_noise_estimate(data, gtab, patch_radius=1, correct_bias=True,
...
-    V = X.dot(W)
 
-    # Grab the column corresponding to the smallest eigen-vector/-value:
-    I = V[:, -1].reshape(n0, n1, n2)
+    if images_as_samples:
+        # use second-to-last eigenvector, because of X is centered
+        W = W.astype('double')
+        V = W[:, n3-2].reshape(n0, n1, n2)
+
+        # eigenvalue = variance gives scale (ref [1]_ method is ambigious)
```
jhlegarreta, line 120: Not sure if I understand the comment.

jhlegarreta (commented):
Thanks for this work. Denoising is not my field of study, so sorry if my review is inaccurate/misleading or only focusing on non-essential things, but a few general comments:
- I have not taken the time to re-write and re-arrange the equations across the referenced papers.
- Can the equations be put in LaTeX form in the method documentation, and be pointed to the exact equivalents across the referenced papers? I find that besides being helpful to users, it would also help reviewing this.
- Error plotting the results prior to this fix and after may be helpful.

dipy/denoise/localpca.py
samcoveney: We cannot do PCA with num_samples=1 and I believe this is prevented elsewhere in the code. The code tests for how many components to keep based on the variance, rather than `ncomps`, I believe that `ncomps` was likely left in the code for testing reasons, and so you are probably right and we should remove it here, to make it clear that it is not used

Point taken on the comment, this was mainly to highly things for reviewers

dipy/denoise/localpca.py
```diff
@@ -281,24 +291,33 @@ def genpca(arr, sigma=None, mask=None, patch_radius=2, pca_method='eig',
...
     patch_radius : int or 1D array (optional)
         The radius of the local patch to be taken around each voxel (in
         voxels). Default: 2 (denoise in blocks of 5x5x5 voxels).
+    gtab: gradient table object (optional if sigma is provided)
+        gradient information for the data gives us the bvals and bvecs of
+        diffusion data, which is needed to calculate noise level if sigma is
+        not provided.
+    patch_radius_sigma : int (optional)
+        The radius of the local patch to be taken around each voxel (in
+        voxels) for estimating sigma. Default: 1 (estimate in blocks of 3x3x3 voxels).
```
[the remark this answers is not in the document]
samcoveney: good point thanks, I will remove from the docs

dipy/denoise/localpca.py
jhlegarreta: Either in the original form or the re-arranged form, but documenting either form (method docstring and before each relevant block) would help users and reviewers. I find it helpful writing those using LaTeX syntax.

dipy/denoise/pca_noise_estimate.pyx
samcoveney: Allowing for `False` allows the code to work as previously (before my changes). In truth, the two approaches are sort of doing something quite similar, and it is not obvious what the best approach really is. But it is clear that `True` is the canonical algorithm.

This is for estimating the noise level, rather than for doing the denoising using that estimate noise level. I don't believe that transposing X is a good idea for the denoising step (and this certainly isn't what the literature prescribes)

(I will remove the default argument from the documentation)

dipy/denoise/pca_noise_estimate.pyx
samcoveney: Basically, the original paper is extremely unclear on a certain point. I emailed them, to no avail. I will try to find a better worded comment, but the current code (i.e. with my changes for estimating the noise level) does seem to estimate noise properly

dipy/denoise/pca_noise_estimate.pyx
jhlegarreta: Then either the application case `True` vs. `False` should be documented (if relevant) or a warning should be raised if `False`. Also, if the default has changed, this should also go through some cycle of warning users that the default behavior has changed (regardless of the previous comment).

dipy/denoise/pca_noise_estimate.pyx
jhlegarreta: > I emailed them, to no avail

Thanks for the effort Sam.

samcoveney:
@RafaelNH and @ShreyasFadnavis any thoughts on this please, in addition to @jhlegarreta ?

dipy/denoise/pca_noise_estimate.pyx
samcoveney: I have decided to leave the non-canonical algorithm, which was the original algorithm in the code, as the default. I have documented this thoroughly now.

dipy/denoise/pca_noise_estimate.pyx
```diff
@@ -92,18 +100,32 @@ def pca_noise_estimate(data, gtab, patch_radius=1, correct_bias=True,
...
     W = Vt.T
-    # Project into the data space
-    V = X.dot(W)
 
-    # Grab the column corresponding to the smallest eigen-vector/-value:
-    I = V[:, -1].reshape(n0, n1, n2)
+    if images_as_samples:
+        # use second-to-last eigenvector, because of X is centered
+        W = W.astype('double')
+        V = W[:, n3-2].reshape(n0, n1, n2)
```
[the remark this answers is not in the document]
samcoveney: good idea, done

dipy/denoise/pca_noise_estimate.pyx
```diff
@@ -92,18 +100,32 @@ def pca_noise_estimate(data, gtab, patch_radius=1, correct_bias=True,
     if (dsm != 1) and (dsm < 2 * patch_radius + 1):
         raise ValueError("Array 'data' is incorrect shape")
 
-    X = data0.reshape(nsamples, n3)
+    if images_as_samples:
```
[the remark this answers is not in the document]
samcoveney: I have documented, hopefully this will be sufficient.

dipy/denoise/pca_noise_estimate.pyx
samcoveney: Although I have tested both the canonical (newly added) and non-canonical (originally in DiPy) code and found they perform extremely similarly, I have decided to leave the original algorithm as the default, and have improved these comments, and the documentation, to explain.

dipy/denoise/localpca.py
```diff
@@ -317,6 +336,13 @@ def localpca(arr, sigma, mask=None, patch_radius=2, pca_method='eig',
         set to None, it will be automatically calculated using the
         Marcenko-Pastur distribution [2]_.
         Default: 2.3 (according to [1]_)
+    return_sigma : bool (optional)
+        If true, a noise standard deviation estimate based on the
+        Marcenko-Pastur distribution is returned [2]_.
+        Default: False.
+    correct_bias : bool
```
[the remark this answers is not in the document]
samcoveney: agreed

dipy/denoise/localpca.py
samcoveney: I have added documentation to explain that this while loop, and the condition therein, is expressed differently to the paper, and I have identified the corresponding equation numbers so people can look themselves.

samcoveney:
Really sorry guys but the branch has become completely corrupted and I cannot figure out why, or how to fix it. I am closing this pull request and will make a new one.

dipy/denoise/localpca.py
jhlegarreta: OK. I'd have preferred to write the equations explicitly in the documentation using LaTeX, but OK. Resolving.

dipy/denoise/localpca.py
jhlegarreta: Done.

dipy/denoise/localpca.py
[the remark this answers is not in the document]
jhlegarreta: Done.

dipy/denoise/localpca.py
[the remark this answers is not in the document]
jhlegarreta: Done.

dipy/denoise/pca_noise_estimate.pyx
jhlegarreta: :+1: for me, but probably other reviewers or people more knowledgeable should weigh in.

dipy/denoise/pca_noise_estimate.pyx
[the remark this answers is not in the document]
jhlegarreta: :+1: looks OK to me.

dipy/denoise/pca_noise_estimate.pyx
jhlegarreta: OK. As said in the related above comment, other reviewers might want to weigh in.

PR #2711, opened 2023-01-10
[FIX] use tempfile module instead of nibabel for TemporaryDirectory

skoudoro:
`TemporaryDirectory` has been deprecated in nibabel. We should use the standard module `tempfile`. This PR fix this issue. Also, some import reshaping


arokem:
LGTM! Please go ahead and merge when the CI comes back green.

skoudoro:
Ok, CI's are green, Thank you @arokem, I will go ahead and merge this PR.

PR #2713, opened 2023-01-16
Empty vtk support

frheault:
This emerged from testing conversion of empty tck/trk to vtk and its was crashing

I forgot the first PR existed and when I updated the code I messed up my command. I did not have the patience to clean out my branch.


skoudoro:
Thanks @frheault. we will look into it. As always, if you could anticipate by adding a test that fails without this functionality would be a good addition to this PR.

ref :  #2599

skoudoro (approved):
Thanks for the update @frheault!

Looks good, I will just wait for the CI's to finish

PR #2716, opened 2023-01-20
NF - Added cython utility functions 

gabknight:
Utility functions for PTT tracking PR #2596.

Is there a good way to add unit tests on these?

### DIFF
--- a/dipy/utils/fast_numpy.pxd
+++ b/dipy/utils/fast_numpy.pxd
@@ -4,6 +4,13 @@
 
 cimport numpy as np
 
+from libc.stdlib cimport rand
+from libc.math cimport sqrt, fabs, M_PI, pow, sin, cos
+
+
+cdef extern from "limits.h":
+    int INT_MAX
+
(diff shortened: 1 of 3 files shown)


skoudoro:
Thanks @gabknight,

For the test, you can create a `test_fast_numpy.py` in `utils/tests` folder.

Inside the tests, check if np.dot is slower than your function and obtain the same result

gabknight:
ok, thanks. I needed to change the function declaration from cdef to cpdef for the tests. Cross is 25x faster, dot is 1.5x on the length-3 vectors. There will more gain when doing the tracking in c.

skoudoro (approved):
Just waiting for the CI to finish and this PR is ready to be merged.

Thank you @gabknight

skoudoro:
# SMALL NOTE

We need a system to allow pytest to run test on `cdef` files. In this file, `cpdef` are used just for the tests. This is not necessary. I suppose we need test as `pyx` files. Something to explore

PR #2717, opened 2023-01-25
fixed bug for non-linear fitting with masks

samcoveney:
When using NLLS or RESTORE, there was a bug with using a mask if wanting to return S0. This wasn't caught because the tests did not cover non-linear fitting when using both masks and S0.


arokem:
LGTM! If one more person could take a look over, we can merge this in quickly.

skoudoro (approved):
Looks good to me also, thanks @samcoveney! merging

PR #2727, opened 2023-02-18
Fix EXTRAS_REQUIRE

### LINKED ISSUE #2726: Setup broken on Python 3.10.9 setuptools 67.2.0
## Description
Setup is broken. Reason is that dipy/info.py is missing some commas between entries of `EXTRAS_REQUIRE`, causing strings to be implicitly concatenated and resulting in invalid dependency specifications.

## Way to reproduce
* Arch Linux
* Python 3.10.9
* Setuptools 1:67.2.0
```
$ python setup.py build
/usr/lib/python3.10/site-packages/setuptools/config/setupcfg.py:520: SetuptoolsDeprecationWarning: The license_file parameter is deprecated, use license_files instead.
  warnings.warn(msg, warning_class)
error in dipy setup command: 'extras_require' must be a dictionary whose values are strings or lists of strings containing valid project/version requirement specifiers.
```

m-pilia:
Resolves #2726

### DIFF
--- a/dipy/info.py
+++ b/dipy/info.py
@@ -125,7 +125,7 @@ EXTRAS_REQUIRE = {
         "pandas",
         "tables",
         "matplotlib",
-        "fury>=0.8.0"
+        "fury>=0.8.0",
         "scikit-learn",
         "scikit-image",
         "statsmodels",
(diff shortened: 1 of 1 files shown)


skoudoro (approved):
Good catch, Thanks @m-pilia!

PR #2609, opened 2022-05-31
NF: DKI+ constraints

tomdelahaije:
This request implements the DKI+ constraints proposed in https://doi.org/10.1016/j.neuroimage.2019.116405. The constraints can be used by setting `fit_method='CLS'` (constrained ordinary least squares) or `fit_method='CWLS'` (constrained weighted least squares). The implementation can be tested by following the DKI example in the documentation (https://dipy.org/documentation/1.4.1./examples_built/reconst_dki), and running for example

```
dkimodel_plus = dki.DiffusionKurtosisModel(gtab, fit_method='CWLS')
dkifit_plus = dkimodel_plus.fit(data_smooth, mask=mask)
```

It is recommended to use the MOSEK solver for speed by setting `cvxpy_solver='MOSEK'`, cf. https://docs.mosek.com/latest/install/installation.html for installation.

The implementation of these constraints is essentially identical to the implementation provided by #2531 for MAP-MRI.

One point: the SCS solver that CVXPY uses by default is more stable and gives more accurate results when the b-values are given in natural units [ms / micron^2], essentially dividing the b-values by 1000 when they are given in standard units of [s / mm^2]. Is there a recommended way of using these natural units when setting up a gradient table, or should I just try to rescale the gradients in the fitting function?

PR #2730, opened 2023-02-24
Fix weighted Jacobians, return extra fit data, add adjacency function

samcoveney:
Added the ability to return and store extra information from the fitting functions, in this case applied to retrieving a mask of non-outlier voxels from the RESTORE algorithm, but which could be used to return other data from different functions (for example, the leverages for linear fitting algorithms). This data is stored in TensorFit.model.extra to avoid changing the interface.

I have also added a function to calculate adjacency of voxels, taking account of arbitrary masks. This currently has no use, but it works correctly for a new fitting function I am working on, and will make it possible to include algorithms that utilize multiple voxels for fitting.

The code should work exactly as before as far as the user is concerned. I can add some additional tests, but am currently a little unsure about what they would be, so comments and suggestions are very welcome.

EDIT: following on from #2746, added fixes to get the Jacobian correct for weighted non-linear least squares. This also required modifications to RESTORE to (1) calculate the Jacobian correctly; (2) bring it more in line with the canonical algorithm.

### DIFF
--- a/dipy/reconst/dki_micro.py
+++ b/dipy/reconst/dki_micro.py
@@ -393,9 +393,14 @@ class KurtosisMicrostructureModel(DiffusionKurtosisModel):
         if mask is None:
             out_shape = data.shape[:-1] + (-1,)
             params = params_all_mask.reshape(out_shape)
+            if extra is not None:
+                self.extra = extra.reshape(data.shape)
         else:
             params = np.zeros(data.shape[:-1] + (params_all_mask.shape[-1],))
             params[mask, :] = params_all_mask
+            if extra is not None:
+                self.extra = np.zeros(data.shape)
+                self.extra[mask, :] = extra
 
         return KurtosisMicrostructuralFit(self, params)
 
--- a/dipy/reconst/dki_micro.py
+++ b/dipy/reconst/dki_micro.py
@@ -393,14 +393,14 @@ class KurtosisMicrostructureModel(DiffusionKurtosisModel):
         if mask is None:
             out_shape = data.shape[:-1] + (-1,)
             params = params_all_mask.reshape(out_shape)
-            if extra is not None:
-                self.extra = extra.reshape(data.shape)
+            #if extra is not None:
+            #    self.extra = extra.reshape(data.shape)
         else:
             params = np.zeros(data.shape[:-1] + (params_all_mask.shape[-1],))
             params[mask, :] = params_all_mask
-            if extra is not None:
-                self.extra = np.zeros(data.shape)
-                self.extra[mask, :] = extra
+            #if extra is not None:
+            #    self.extra = np.zeros(data.shape)
+            #    self.extra[mask, :] = extra
 
         return KurtosisMicrostructuralFit(self, params)
 
--- a/dipy/reconst/dti.py
+++ b/dipy/reconst/dti.py
@@ -1371,7 +1371,11 @@ def iter_fit_tensor(step=1e4):
                                   return_S0_hat=return_S0_hat,
                                   *args, **kwargs)
             data = data.reshape(-1, data.shape[-1])
-            dtiparams = np.empty((size, 12), dtype=np.float64)
+            if 'return_lower_triangular' in kwargs:
+                sz = 7 if kwargs['return_lower_triangular'] else 12
+            else:
+                sz = 12
+            dtiparams = np.empty((size, sz), dtype=np.float64)
             if return_S0_hat:
                 S0params = np.empty(size, dtype=np.float64)
             extra = np.empty(data.shape)
@@ -1389,10 +1393,10 @@ def iter_fit_tensor(step=1e4):
                                      data[i:i + step],
                                      *args, **kwargs)
             if return_S0_hat:
-                return (dtiparams.reshape(shape + (12, )),
+                return (dtiparams.reshape(shape + (sz, )),
                         S0params.reshape(shape + (1, ))), extra.reshape(shape + (-1,))
             else:
-                return dtiparams.reshape(shape + (12, )), extra.reshape(shape + (-1,))
+                return dtiparams.reshape(shape + (sz, )), extra.reshape(shape + (-1,))
 
         return wrapped_fit_tensor
 
@@ -1468,7 +1472,7 @@ def wls_fit_tensor(design_matrix, data, return_S0_hat=False, adjacency=None):
     log_s = np.log(data)
 
     # calculate weights
-    fit_result = ols_fit_tensor(design_matrix, data,
+    fit_result, _ = ols_fit_tensor(design_matrix, data,
                                 return_lower_triangular=True)
     w = np.exp(fit_result @ design_matrix.T)
 
@@ -1548,7 +1552,7 @@ def ols_fit_tensor(design_matrix, data, return_S0_hat=False,
                            np.log(data))
 
     if return_lower_triangular:
-        return fit_result
+        return fit_result, None
 
     if return_S0_hat:
         return (eig_from_lo_tri(fit_result,
@@ -1789,7 +1793,7 @@ def nlls_fit_tensor(design_matrix, data, weighting=None,
     flat_data = data.reshape((-1, data.shape[-1]))
 
     # Use the OLS method parameters as the starting point for the optimization:
-    D = ols_fit_tensor(design_matrix, flat_data, return_lower_triangular=True)
+    D, _ = ols_fit_tensor(design_matrix, flat_data, return_lower_triangular=True)
 
     # Flatten for the iteration over voxels:
     ols_params = np.reshape(D, (-1, D.shape[-1]))
@@ -1928,7 +1932,7 @@ def restore_fit_tensor(design_matrix, data, sigma=None, jac=True,
             sigma = np.reshape(sigma, (-1, 1))
 
     # calculate OLS solution
-    D = ols_fit_tensor(design_matrix, flat_data, return_lower_triangular=True)
+    D, _ = ols_fit_tensor(design_matrix, flat_data, return_lower_triangular=True)
 
     # Flatten for the iteration over voxels:
     ols_params = np.reshape(D, (-1, D.shape[-1]))
@@ -2007,7 +2011,7 @@ def restore_fit_tensor(design_matrix, data, sigma=None, jac=True,
                     robust[vox] = (cond == False)
 
                     # recalculate OLS solution with clean data
-                    new_start = ols_fit_tensor(clean_design, clean_data,
+                    new_start, _ = ols_fit_tensor(clean_design, clean_data,
                                                return_lower_triangular=True)
 
                     if np.iterable(sigma_vox):
--- a/dipy/reconst/dti.py
+++ b/dipy/reconst/dti.py
@@ -708,14 +708,14 @@ def adjacency_calc(img_shape, mask):
             if flat_mask[idx]:
                 cond = dists[idx, :]
                 cond = cond * flat_mask
-                cond = cond[flat_mask]  # so indices will match masked array 
+                cond = cond[flat_mask]  # so indices will match masked array
                 adj.append(np.argwhere(cond).flatten().tolist())
     else:
         for idx in range(dists.shape[0]):
             cond = dists[idx, :]
             adj.append(np.argwhere(cond).flatten().tolist())
 
-    return adj   
+    return adj
 
 
 class TensorModel(ReconstModel):
@@ -814,7 +814,7 @@ class TensorModel(ReconstModel):
             A boolean array used to mark the coordinates in the data that
             should be analyzed that has the shape data.shape[:-1]
 
-        adjacency : bool
+        adjacency : bool, optional
             Boolean to calculate voxel adjacency accounting for mask.
         """
 
@@ -828,7 +828,8 @@ class TensorModel(ReconstModel):
             mask = np.array(mask, dtype=bool, copy=False)
         data_in_mask = np.reshape(data[mask], (-1, data.shape[-1]))
 
-        # NOTE: build adjacency list, if required - could make a list somewhere that notes whether this is required for a particular algorithm
+        # build adjacency list, if required
+        # NOTE: could make a list somewhere that notes whether this is required for a particular algorithm
         if adjacency:
             self.kwargs["adjacency"] = adjacency_calc(img_shape, mask)
 
@@ -1382,7 +1383,7 @@ def iter_fit_tensor(step=1e4):
             for i in range(0, size, step):
                 if return_S0_hat:
                     (dtiparams[i:i + step], S0params[i:i + step]),\
-                    extra[i:i + step]\
+                     extra[i:i + step]\
                         = fit_tensor(design_matrix,
                                      data[i:i + step],
                                      return_S0_hat=return_S0_hat,
@@ -1419,7 +1420,7 @@ def wls_fit_tensor(design_matrix, data, return_S0_hat=False, adjacency=None):
         dimension should contain the data. It makes no copies of data.
     return_S0_hat : bool
         Boolean to return (True) or not (False) the S0 values for the fit.
-    adjacency : list
+    adjacency : list, optional
         Adjacency list matching the flattened and masked array data array.
 
     Returns
@@ -1473,7 +1474,7 @@ def wls_fit_tensor(design_matrix, data, return_S0_hat=False, adjacency=None):
 
     # calculate weights
     fit_result, _ = ols_fit_tensor(design_matrix, data,
-                                return_lower_triangular=True)
+                                   return_lower_triangular=True)
     w = np.exp(fit_result @ design_matrix.T)
 
     # the weighted problem design_matrix * w is much larger (differs per voxel)
@@ -1512,7 +1513,7 @@ def ols_fit_tensor(design_matrix, data, return_S0_hat=False,
         dimension should contain the data. It makes no copies of data.
     return_S0_hat : bool
         Boolean to return (True) or not (False) the S0 values for the fit.
-    adjacency : list
+    adjacency : list, optional
         Adjacency list matching the flattened and masked array data array.
     return_lower_triangular : bool
         Boolean to return (True) or not (False) the coefficients of the fit.
@@ -1771,7 +1772,7 @@ def nlls_fit_tensor(design_matrix, data, weighting=None,
     return_S0_hat : bool
         Boolean to return (True) or not (False) the S0 values for the fit.
 
-    adjacency : list
+    adjacency : list, optional
         Adjacency list matching the flattened and masked array data array.
 
     fail_is_nan : bool
@@ -1902,7 +1903,7 @@ def restore_fit_tensor(design_matrix, data, sigma=None, jac=True,
     return_S0_hat : bool
         Boolean to return (True) or not (False) the S0 values for the fit.
 
-    adjacency : list
+    adjacency : list, optional
         Adjacency list matching the flattened and masked array data array.
 
     fail_is_nan : bool
--- a/dipy/reconst/dti.py
+++ b/dipy/reconst/dti.py
@@ -8,6 +8,7 @@ import functools
 import numpy as np
 
 import scipy.optimize as opt
+from scipy.spatial.distance import pdist, squareform
 
 from dipy.utils.arrfuncs import pinv
 from dipy.data import get_sphere
@@ -673,6 +674,50 @@ def tensor_prediction(dti_params, gtab, S0):
     return np.exp(np.dot(lower_tri, D.T))
 
 
+def adjacency_calc(img_shape, mask):
+    r"""Create adjacency list for voxels, accounting for the mask.
+
+    Parameters
+    ----------
+
+    img_shape : list
+        Spatial shape of the image data.
+
+    mask : array
+        A boolean array used to mark the coordinates in the data that
+        should be analyzed that should have the same shape as the images.
+
+    Returns
+    -------
+
+    adj : list
+        List, one entry per voxel, giving the indices of adjacent voxels.
+        The list will correspond to the data array once masked and flattened.
+
+    """
+
+    # Image-space coordinates of voxels, flattened
+    XYZ = np.meshgrid(*[range(ds) for ds in img_shape], indexing='ij')
+    XYZ = np.column_stack([xyz.ravel() for xyz in XYZ])
+    dists = squareform(pdist(XYZ))
+    dists = (dists < 2)  # NOTE: adjaceny list will contain the voxel itself
+    adj = []
+    if mask is not None:
+        flat_mask = mask.reshape(-1)
+        for idx in range(dists.shape[0]):
+            if flat_mask[idx]:
+                cond = dists[idx, :]
+                cond = cond * flat_mask
+                cond = cond[flat_mask]  # so indices will match masked array 
+                adj.append(np.argwhere(cond).flatten().tolist())
+    else:
+        for idx in range(dists.shape[0]):
+            cond = dists[idx, :]
+            adj.append(np.argwhere(cond).flatten().tolist())
+
+    return adj   
+
+
 class TensorModel(ReconstModel):
     """ Diffusion Tensor
     """
@@ -757,7 +802,7 @@ class TensorModel(ReconstModel):
             e_s += " positive."
             raise ValueError(e_s)
 
-    def fit(self, data, mask=None):
+    def fit(self, data, mask=None, adjacency=False):
         """ Fit method of the DTI model class
 
         Parameters
@@ -769,15 +814,23 @@ class TensorModel(ReconstModel):
             A boolean array used to mark the coordinates in the data that
             should be analyzed that has the shape data.shape[:-1]
 
+        adjacency : bool
+            Boolean to calculate voxel adjacency accounting for mask.
         """
+
         S0_params = None
 
+        img_shape = data.shape[:-1]
         if mask is not None:
             # Check for valid shape of the mask
-            if mask.shape != data.shape[:-1]:
+            if mask.shape != img_shape:
                 raise ValueError("Mask is not the same shape as data.")
             mask = np.array(mask, dtype=bool, copy=False)
         data_in_mask = np.reshape(data[mask], (-1, data.shape[-1]))
+
+        # NOTE: build adjacency list, if required - could make a list somewhere that notes whether this is required for a particular algorithm
+        if adjacency:
+            self.kwargs["adjacency"] = adjacency_calc(img_shape, mask)
 
         if "sigma" in self.kwargs:
             sigma = self.kwargs["sigma"]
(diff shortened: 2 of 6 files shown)


skoudoro:
I like the idea to add extra information. I wonder if it should be generalized to all model.

I will have a closer look at this PR this afternoon. Thank you for this @samcoveney

skoudoro:
Hi @samcoveney,

Can you check why it is failing before starting a review.  I see the following error:

```python
        if mask is None:
            out_shape = data.shape[:-1] + (-1, )
>           dki_params = params_in_mask.reshape(out_shape)
E           AttributeError: 'tuple' object has no attribute 'reshape'

```

Thanks for looking into it

dipy/reconst/dti.py
```diff
@@ -769,16 +814,24 @@ def fit(self, data, mask=None):
             A boolean array used to mark the coordinates in the data that
             should be analyzed that has the shape data.shape[:-1]
 
+        adjacency : bool
```
skoudoro, line 817: `bool, optional` instead

dipy/reconst/dti.py
```diff
@@ -1354,6 +1419,8 @@ def wls_fit_tensor(design_matrix, data, return_S0_hat=False):
         dimension should contain the data. It makes no copies of data.
     return_S0_hat : bool
         Boolean to return (True) or not (False) the S0 values for the fit.
+    adjacency : list
```
skoudoro, line 1422: `list, optional` instead

dipy/reconst/dti.py
```diff
@@ -1702,6 +1771,9 @@ def nlls_fit_tensor(design_matrix, data, weighting=None,
     return_S0_hat : bool
         Boolean to return (True) or not (False) the S0 values for the fit.
 
+    adjacency : list
```
skoudoro, line 1774: same as above

dipy/reconst/dti.py
```diff
@@ -1829,6 +1902,9 @@ def restore_fit_tensor(design_matrix, data, sigma=None, jac=True,
...
```
skoudoro, line 1905: same as above

skoudoro (commented):
some quick additional comments.

dipy/reconst/dti.py
samcoveney: done!

dipy/reconst/dti.py
samcoveney: done

dipy/reconst/dti.py
samcoveney: done

dipy/reconst/dti.py
samcoveney: done

samcoveney:
Following from #2746 I may need to make some fixes to the non-linear fitting functions. It would be ideal, if there are no problems with the changes here, to build these fixes on top of this pull request, rather than the other way around (mainly for my convenience, sorry - let me know if this is a bad idea).

samcoveney:
Also included fixes for #2746 here. In order to have certain results needed for the Jacobian cached, rather than recalculated, I have made use of a new class. This seemed like the best approach, since there is no implementation in Scipy / Numpy to use the Levenberg–Marquardt algorithm using a function that returns both the function and its derivative.

samcoveney:
Any progress on this? I understand that the conference is coming up, so no worries if it's on the back-burner for now. Thanks

skoudoro:
@RafaelNH @arokem @Garyfallidis, did you have time to look at this to answer to @samcoveney?

arokem:
Not before next week, no.

samcoveney:
While rebasing this on master, I noticed changes in dki.py and dki_micro.py, and while I think that I have correctly rebased on master, I have a couple of concerns about some inconsistency in the use of masks in the DKI fitting functions. However, those things were not introduced by me, so it might be something to look into later.

samcoveney:
Hi all, this has been sitting around for a really long time, any chance of action please?

skoudoro:
Hi @samcoveney,

Sorry for this delay, I will look into it today.

However, I will need someone else to look at your modification in RESTORE. I can not move forward without this. @arokem @RafaelNH @Garyfallidis ? maybe @gabknight ?

> I have a couple of concerns about some inconsistency in the use of masks in the DKI fitting functions

to be faster, can you point some line number? I suppose this happened when we merged #2609

thank you again

arokem:
I can take a look but it will not be before the last week of August.

dipy/reconst/dti.py
```diff
@@ -769,16 +814,25 @@ def fit(self, data, mask=None):
...
         if mask is not None:
             # Check for valid shape of the mask
-            if mask.shape != data.shape[:-1]:
+            if mask.shape != img_shape:
                 raise ValueError("Mask is not the same shape as data.")
             mask = np.array(mask, dtype=bool, copy=False)
         data_in_mask = np.reshape(data[mask], (-1, data.shape[-1]))
 
+        # build adjacency list, if required
+        # NOTE: could make a list somewhere that notes whether this is required for a particular algorithm
```
skoudoro, line 832: to remove

dipy/reconst/dti.py
```diff
@@ -1313,25 +1372,32 @@ def wrapped_fit_tensor(design_matrix, data, return_S0_hat=False,
...
-                                                       *args, **kwargs)
+                    dtiparams[i:i + step], extra[i:i + step]\
+                        = fit_tensor(design_matrix,
+                                     data[i:i + step],
+                                     *args, **kwargs)
             if return_S0_hat:
-                return (dtiparams.reshape(shape + (12, )),
-                        S0params.reshape(shape + (1, )))
+                return (dtiparams.reshape(shape + (sz, )),
+                        S0params.reshape(shape + (1, ))), extra.reshape(shape + (-1,))
```
skoudoro, line 1398: pep8 too long

dipy/reconst/dti.py
```diff
@@ -1313,25 +1372,32 @@ def wrapped_fit_tensor(design_matrix, data, return_S0_hat=False,
...
             else:
-                return dtiparams.reshape(shape + (12, ))
+                return dtiparams.reshape(shape + (sz, )), extra.reshape(shape + (-1,))
```
skoudoro, line 1400: pep8 too long

dipy/reconst/dti.py
```diff
@@ -1313,25 +1372,32 @@ def wrapped_fit_tensor(design_matrix, data, return_S0_hat=False,
...
+                sz = 12
+            dtiparams = np.empty((size, sz), dtype=np.float64)
             if return_S0_hat:
                 S0params = np.empty(size, dtype=np.float64)
+            extra = np.empty(data.shape)
             for i in range(0, size, step):
                 if return_S0_hat:
-                    dtiparams[i:i + step], S0params[i:i + step] \
+                    (dtiparams[i:i + step], S0params[i:i + step]),\
+                     extra[i:i + step]\
```
skoudoro, line 1386: pep8 alignment issue

dipy/reconst/dti.py
```diff
@@ -1514,111 +1580,129 @@ def _ols_fit_matrix(design_matrix):
     return np.dot(U, U.T)
 
 
-def _nlls_err_func(tensor, design_matrix, data, weighting=None,
-                   sigma=None):
-    r"""
-    Error function for the non-linear least-squares fit of the tensor.
+class _nlls_class():
```
skoudoro, line 1583: Not sure to see the benefit of a class here.  is there a specific state that you want to keep? maybe a closure is enough. Can you explain ? thank you

dipy/reconst/dti.py
```diff
@@ -1514,111 +1580,129 @@ def _ols_fit_matrix(design_matrix):
...
+        .. [1] Koay, CG, Chang, L-C, Carew, JD, Pierpaoli, C, Basser PJ (2006).
+            A unifying theoretical and algorithmic framework for least squares
+            methods of estimation in diffusion tensor imaging. MRM 182, 115-25.
 
-    Notes
-    -----
-    This is an implementation of equation 14 in [1]_.
+        """
+        # minus sign, because derivative of residuals = data - y
+        # sqrt(w) because w coresponds to the squared residuals
```
skoudoro, line 1696: typo: coresponds ==> corresponds

dipy/reconst/tests/test_dti.py
```diff
@@ -511,37 +511,56 @@ def test_mask():
...
-    # Test Jacobian at zero
-    D = np.zeros_like(D)
-    args = [X, Y]
-    analytical = dti._nlls_jacobian_func(D, *args)
-    for i in range(len(X)):
-        args = [X[i], Y[i]]
-        approx = opt.approx_fprime(D, dti._nlls_err_func, 1e-8, *args)
-        assert np.allclose(approx, analytical[i])
+    Y = np.exp(np.dot(X, D_orig))
+    # NOTE: added error here - maybe remove later
```
skoudoro, line 528: comment to address and/or remove

dipy/reconst/tests/test_dti.py
```diff
@@ -511,37 +511,56 @@ def test_mask():
...
+    scale = 10
+    error = np.random.normal(scale=scale, size=Y.shape)
+    Y = Y + error
+
+    # NOTE: although sigma and gmm gradients seem correct from inspection, they are not accurate enough to pass the tests
```
skoudoro, line 533: comment to address and/or remove

dipy/reconst/dti.py
```diff
@@ -824,12 +878,17 @@ def fit(self, data, mask=None):
             dti_params = params_in_mask.reshape(out_shape)
             if self.return_S0_hat:
                 S0_params = model_S0.reshape(out_shape[:-1])
+            if extra is not None:
+                self.extra = extra.reshape(data.shape)
```
skoudoro, line 882: I would recommend to initialize `self.extra = {}` in the init function

skoudoro (changes requested):
Hi @samcoveney, 

The goal of this PR is providing the ability to return and store extra information from the fitting functions. 

However, you return a numpy array.  I would expect something like a dict to be your storage/container object. Then, on this storage, you can have a key like 'adjacency' or your numpy array as a value. it could be a dynamic dataclass also. This should allow other user to provide and combine other extra information if needed. Currently, I feel like it is not possible or I missed something.

> This data is stored in TensorFit.model.extra to avoid changing the interface

Currently, sometimes it exist, sometimes not. I would recommend to make it more consistent.

Can you add a test for `adjacency_calc` function? 

Thanks !

skoudoro:
Concerning the mask issue, I see what you mean, we need to update this. I think it would be better in a new PR

dipy/reconst/dki_micro.py
```diff
@@ -393,9 +393,14 @@ def fit(self, data, mask=None, sphere='repulsion100', gtol=1e-2,
...
             out_shape = data.shape[:-1] + (-1,)
             params = params_all_mask.reshape(out_shape)
+            #if extra is not None:
+            #    self.extra = extra.reshape(data.shape)
         else:
             params = np.zeros(data.shape[:-1] + (params_all_mask.shape[-1],))
             params[mask, :] = params_all_mask
+            #if extra is not None:
+            #    self.extra = np.zeros(data.shape)
+            #    self.extra[mask, :] = extra
```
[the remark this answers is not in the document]
samcoveney: as we discuss below, there seems to be an issue with masking and new dki.py and dki_micro.py, so I left this in as a placeholder/reminder about the sort of modifications that will be needed down the line one the masking is sorted

dipy/reconst/dti.py
```diff
@@ -673,6 +674,50 @@ def tensor_prediction(dti_params, gtab, S0):
...
+
+    """
+
+    # Image-space coordinates of voxels, flattened
+    XYZ = np.meshgrid(*[range(ds) for ds in img_shape], indexing='ij')
+    XYZ = np.column_stack([xyz.ravel() for xyz in XYZ])
+    dists = squareform(pdist(XYZ))
+    dists = (dists < 2)  # NOTE: adjaceny list will contain the voxel itself
+    adj = []
+    if mask is not None:
```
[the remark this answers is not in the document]
samcoveney: thanks, I will make this optional

dipy/reconst/dti.py
samcoveney: done

dipy/reconst/dti.py
```diff
@@ -673,6 +674,50 @@ def tensor_prediction(dti_params, gtab, S0):
     return np.exp(np.dot(lower_tri, D.T))
 
 
+def adjacency_calc(img_shape, mask):
+    r"""Create adjacency list for voxels, accounting for the mask.
```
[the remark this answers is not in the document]
samcoveney: done

dipy/reconst/dti.py
samcoveney: good idea.

The reason that the 'extra' field was a numpy array, is that at various times arrays become flattened due to masks and due to wrappers for iteration over voxels, which makes it quite tricky to generalize these extra returns to something other than a numpy array with the same shape as the data.

Thinking through possible use cases, it is hard to see where an extra return wouldn't be the same shape as the data array, so for now I will try to generalize to a dictionary of items that are the same shape as the data array. This allows to return whether a data-point is robust, to return the leverages of linear problems, the weights used in various fitting functions, the fitting residuals (although I see no point in returning these).

dipy/reconst/dti.py
samcoveney: done

dipy/reconst/tests/test_dti.py
samcoveney: done, will leave in the addition of error in the test

dipy/reconst/tests/test_dti.py
```diff
@@ -585,7 +604,9 @@ def test_nlls_fit_tensor():
     npt.assert_almost_equal(tensor_est.md[0], md)
 
     # Using the gmm weighting scheme:
-    tensor_model = dti.TensorModel(gtab, fit_method='NLLS', weighting='gmm')
+    # NOTE: Not using RESTORE here, using NLLS, in which case the provided
```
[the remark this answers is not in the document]
samcoveney: done

samcoveney:
I think I have made the requested changes. I understand that we may need to wait for RESTORE to be checked over, but I also hope it's clear that the previous implementation must have been wrong and that this PR at least sets the ground for a correct and canonical implementation

skoudoro:
Thank you for the update, I will look into it later today.

can you quickly address the first failing CI: https://github.com/dipy/dipy/actions/runs/5821525834/job/15784136517?pr=2730#step:4:17

>  I understand that we may need to wait for RESTORE to be checked over, but I also hope it's clear that the previous implementation must have been wrong and that this PR at least sets the ground for a correct and canonical implementation

Yes ! and since the mistake have been done one, let's make sure that we do not repeat. That's why I requested a deeper look.
Thank you again!

samcoveney:
I completely forgot to add the test for calc_adjacency() , done that now!

dipy/reconst/dti.py
```diff
@@ -769,16 +825,25 @@ def fit(self, data, mask=None):
...
+        img_shape = data.shape[:-1]
         if mask is not None:
             # Check for valid shape of the mask
-            if mask.shape != data.shape[:-1]:
+            if mask.shape != img_shape:
                 raise ValueError("Mask is not the same shape as data.")
             mask = np.array(mask, dtype=bool, copy=False)
         data_in_mask = np.reshape(data[mask], (-1, data.shape[-1]))
 
+        if adjacency > 0:
```
skoudoro, line 843: strange that this work. I did not know that you could do ` False > 0` or `True >0`. Interesting.  I need to get some info for that.

For now, I would recommend to do `if adjacency and adjacency > 0`

dipy/reconst/dti.py
```diff
@@ -1514,111 +1604,129 @@ def _ols_fit_matrix(design_matrix):
...
+        .. [1] Chang, L-C, Jones, DK and Pierpaoli, C (2005). RESTORE: robust
+        estimation of tensors by outlier rejection. MRM, 53: 1088-95.
+        """
+        # This is the predicted signal given the params:
+        y = np.exp(np.dot(design_matrix, tensor))
+        self.y = y  # cache the results
+
+        # Compute the residuals
+        residuals = data - y
+        self.residuals = residuals  # cache the results
```
skoudoro, line 1667: why residuals in the cache ? the cached value seems to not be used

skoudoro (changes requested):
Thank you for the update @samcoveney , See below my last comments.

dipy/reconst/dti.py
```diff
@@ -756,8 +811,9 @@ def __init__(self, gtab, fit_method="WLS", return_S0_hat=False, *args,
             e_s = "The `min_signal` key-word argument needs to be strictly"
             e_s += " positive."
             raise ValueError(e_s)
+        self.extra = {}
 
-    def fit(self, data, mask=None):
+    def fit(self, data, mask=None, adjacency=False):
```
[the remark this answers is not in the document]
samcoveney: How about initializing with zero then, therefore we only calculate adjacency if a value above zero is given?

dipy/reconst/dti.py
samcoveney: How about initializing with zero then, therefore we only calculate adjacency if a value above zero is given?

dipy/reconst/dti.py
samcoveney: sorry about that, left over from a previous setup where we were updating the GMM weights on each call of the error function (rather than updating them after each optimization, which is the correct thing to do)

samcoveney:
Have made these modifications, and an extra one that makes it easier to understand how the GMM weights are passed from `restore_fit_tensor()` to `err_func()`

dipy/reconst/dti.py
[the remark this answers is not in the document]
skoudoro: Yes, zero is also good for me

dipy/reconst/dti.py
```diff
@@ -1313,25 +1378,43 @@ def wrapped_fit_tensor(design_matrix, data, return_S0_hat=False,
                                   return_S0_hat=return_S0_hat,
                                   *args, **kwargs)
             data = data.reshape(-1, data.shape[-1])
-            dtiparams = np.empty((size, 12), dtype=np.float64)
+            if 'return_lower_triangular' in kwargs:
+                sz = 7 if kwargs['return_lower_triangular'] else 12
+            else:
+                sz = 12
```
skoudoro, line 1384: this is a bit repetitive. `sz = 7 if kwargs.get('return_lower_triangular') else 12` should be enough

skoudoro (approved):
Thank you @samcoveney, 

Overall, it looks good. I added a small comment but it is not mandatory.
Now, waiting for @arokem to check in details RESTORE. I suppose he will do it next week. 
Thanks for your patience.

dipy/reconst/dti.py
samcoveney: What if `return_lower_triangular` is in the dictionary but its value is `False` ?

dipy/reconst/dti.py
arokem: I think a test should be added of the adjacency calculation in the context of the model.

dipy/reconst/dti.py
arokem, line 1384:
```suggestion
sz = 7 if kwargs.get('return_lower_triangular', False) else 12
```

dipy/reconst/dki_micro.py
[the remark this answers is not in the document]
arokem: Can you point to where these issues were discussed? I can't find this discussion now.

dipy/reconst/dti.py
```diff
@@ -673,6 +674,53 @@ def tensor_prediction(dti_params, gtab, S0):
...
+
+    img_shape : list
+        Spatial shape of the image data.
+
+    mask : array, optional
+        A boolean array used to mark the coordinates in the data that
+        should be analyzed that should have the same shape as the images.
+
+    distance : float, optional
+        Cutoff distance for finding adjacent voxels.
```
arokem, line 691:
```suggestion
        Cutoff distance for finding adjacent voxels. Default 1.99 
```

Where does the number 1.99 come from?

dipy/reconst/dti.py
```diff
@@ -756,8 +804,9 @@ def __init__(self, gtab, fit_method="WLS", return_S0_hat=False, *args,
             e_s = "The `min_signal` key-word argument needs to be strictly"
             e_s += " positive."
             raise ValueError(e_s)
+        self.extra = {}
 
-    def fit(self, data, mask=None):
+    def fit(self, data, mask=None, adjacency=0):
```
arokem, line 809: I don't like the setting of adjacency to 0 here. I think it should be set to `False` per default and serve as a threshold/cutoff otherwise.

dipy/reconst/dti.py
```diff
@@ -824,12 +882,19 @@ def fit(self, data, mask=None):
...
         else:
             dti_params = np.zeros(data.shape[:-1] + (12,))
             dti_params[mask, :] = params_in_mask
             if self.return_S0_hat:
                 S0_params = np.zeros(data.shape[:-1])
                 S0_params[mask] = model_S0.squeeze()
+            if extra is not None:
+                for key in extra:
+                    self.extra[key] = np.zeros(data.shape)
+                    self.extra[key][mask, :] = extra[key]
```
arokem, line 897: Just to clarify: the assumption is that `extra` always returns as an array the shape of the data?

arokem: I wonder if `extra` should be named something more specific.

dipy/reconst/dti.py
```diff
@@ -1856,17 +1953,23 @@ def restore_fit_tensor(design_matrix, data, sigma=None, jac=True,
...
     params = np.empty((flat_data.shape[0], npa))
 
+    # For storing whether image is used in final fit for each voxel
+    robust = np.ones(flat_data.shape, dtype=int)
+
     # For warnings
     resort_to_OLS = False
 
+    # Instance of _nlls_class, need for nlls error func and jacobian
+    nlls = _nlls_class()
```
arokem, line 1971: Calling this object `nlls` here could be slightly confusing, because there is also a scipy optimization function of that name.

dipy/reconst/dti.py
arokem: I'd name it so that it's clear that it's a class in subsequent usage (when it's initialized further down). How about: `_NllsHelper`?

arokem (changes requested):
Sorry that it's taken so long to get to this. I think that the fixes to RESTORE look correct, and the additional implementation details (e.g., the _nlls_class) are clever (but see some small suggestions there as well). I apparently missed the discussion of some issues with dki / dki_micro, so I'd appreciate a pointer to this discussion.

dipy/reconst/dti.py
samcoveney: It is there so that the default produces a 3x3 patch (I suppose < 2 would be fine, but I just used 1.99 to be 'safe')

dipy/reconst/dti.py
samcoveney: Thanks, I can't commit from here due to the indent, but this makes sense (and I didn't know we could do this, it's very useful!)

dipy/reconst/dti.py
arokem: If this is voxels, I wonder if you just want to fix it to 1.99 or 2, so that adjacency means one and only one thing.

dipy/reconst/dti.py
```diff
@@ -673,6 +674,53 @@ def tensor_prediction(dti_params, gtab, S0):
     return np.exp(np.dot(lower_tri, D.T))
 
 
+def adjacency_calc(img_shape, mask=None, distance=1.99):
```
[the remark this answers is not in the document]
samcoveney: Probably a good idea, in both cases... I'll take a look into finding a good home for it

dipy/reconst/dti.py
samcoveney: Yeah that would be more intuitive for the user if it was `False`, thanks

dipy/reconst/dti.py
samcoveney: Thanks, @skoudoro had similar thoughts, and originally `extra` was only a single array, but I have now generalized to a dictionary whose entries must be the same shape as the data. The reason for this is that reshaping / flattening / masking that goes on.

I did consider to generalize further, but having made a fair bit of use of this extra functionality myself, I can't really see any use-cases where the extra information we want is not the same shape as the data. At least for now, this seems to be a good compromise that lays a template for future additions.

IMO, I think `extra` probably shouldn't be more specific, the entries could be quite different depending on the fitting function

dipy/reconst/dti.py
samcoveney: that's true, but within the function scope, surely it's okay to use this name?

samcoveney:
Thanks for taking a look @arokem and @skoudoro I will go through and make suggested changes soon.

re: the dki masking - somewhere in this thread I mentioned that, since I made this pull request originally, somebody changed some implementations of dki modules which do not appear to correctly make use of masks. I needed to rebase this PR, and have had to leave some suggested code changes (commented out) in the dki code. I think these modifications should come later, everything seems to work at the moment

dipy/reconst/dti.py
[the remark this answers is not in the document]
samcoveney: okay, all done, has a new home in utils/volume.py and the test moved to utils/tests/test_volume.py

dipy/reconst/dti.py
samcoveney: Voxels, not mm, and I believe the user should be able to select the number of adjacent voxels in a neighbourhood of specified size. This is certainly what I need to do at the moment

samcoveney: Additionally, I was told previously to leave default values out of the documentation, since it inevitably gets missed later and leads to docs errors?

dipy/reconst/tests/test_dti.py
```diff
@@ -623,6 +643,17 @@ def test_nlls_fit_tensor():
     tmf = npt.assert_warns(UserWarning, tensor_model.fit, Y_less)
     npt.assert_equal(tmf[0].S0_hat, np.nan)
 
+    # Test sigma with an array
+    sigma = np.ones(Y_less.shape[-1])
+    tensor_model = dti.TensorModel(gtab_less, fit_method='NLLS',
+                                   weighting='sigma', sigma=sigma)
```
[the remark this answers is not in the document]
samcoveney: good spot, done this

dipy/reconst/dti.py
samcoveney: I don't think there is a scipy opt functions with this name... there is one called `nnls` which is different. Leaving this for now

samcoveney:
I believe I have incorporate (most of) these suggestions, thanks for the help. If things look okay, please let me know so I can squash everything together. Do you need a rebase?

skoudoro:
I think a rebase is not necessary. 

@arokem, Can you have a last look before merging?

Also, I am going to have a last look this afternoon.

arokem:
LGTM

skoudoro (approved):
Thank you for this work and sorry for the delays @samcoveney.

Thank you for the review @arokem 

Looks good to me also, merging

close #2820 also

samcoveney:
Thanks @skoudoro and @arokem for your help and input on this

Issue #2648, opened 2022-09-22
Issues with dipy_align_syn


pjsjongsung:
## Description
I have been following the tutorial for dipy_align_syn, and it seems that the --out_inv_static option in dipy_align_syn does not work and the inverse transformed static image is not saved.

Also, I spent a lot of time figuring out that without the --force option, if there is already a file with the output image's name, the command terminates without any information. I think it would be nice if it at least printed a warning/error that one should use the --force option if they want to overwrite.

I think these are just minor issues and can be resolved with some small fixes, but is important since it is related to how convenient the method is to use.

## Way to reproduce
- [ ] Code example
`dipy_align_syn moving.nii.gz static.nii.gz --force --metric "cc" --out_dir "syn_reg_output" --out_warped "warped.nii.gz" --out_inv_static "inv.nii.gz" --out_field "field.nii.gz"`
- [ ] Relevant images (if any)
- [ ] Operating system and version (run `python -c "import platform; print(platform.platform())"`)
Windows-10-10.0.19043-SP0
- [ ] Python version (run `python -c "import sys; print("Python", sys.version)"`)
Python 3.8.12
- [ ] dipy version (run `python -c "import dipy; print(dipy.__version__)"`)
1.5.0
- [ ] dependency version (numpy, scipy, nibabel, h5py, cvxpy, fury)
    * import numpy; print("NumPy", numpy.__version__)
    1.22.4
    * import scipy; print("SciPy", scipy.__version__)
    1.8.0
    * import nibabel; print("Nibabel", nibabel.__version__)
    3.2.2
    * import h5py; print("H5py", h5py.__version__)
    3.6.0
    * import cvxpy; print("Cvxpy", cvxpy.__version__)
    1.2.1
    * import fury; print("fury", fury.__version__)
    0.8.0

skoudoro:
Thank you for the feedback @pjsjongsung, 

we will look into it

lb-97:
@skoudoro, I had made a commit for this issue and raised a pull request. Kindly review and let me know what you think.

skoudoro:
Can I close this issue @pjsjongsung?  it seems it has been fixed by https://github.com/dipy/dipy/pull/2733.

Please, try and keep us updated. Thanks !

pjsjongsung:
I think we can. Thank you @lb-97 @skoudoro !

PR #2733, opened 2023-02-26
Update SynRegistrationFlow for #2648

lb-97:
Updated the class to save inverse transformed static image


skoudoro:
Thank you @lb-97, merging

PR #2749, opened 2023-03-19
adding a new getitem method

shilpiprd:
I figured that get_item method is most closely related to gradients.py file which is inside dipy/core/gradients.py folder because i saw the bvech bvals method. 
# Example usage
#since we can call a getItem method by simply using square brackets, 
gtab = MyGradientTable(gradients)
gtab_selected = gtab[2:10]  # Select b-values from 2 to 10. 

Please let me know if I need to make any changes to this method.

### DIFF
--- a/dipy/core/gradients.py
+++ b/dipy/core/gradients.py
@@ -211,7 +211,7 @@ class GradientTable(object):
         denom = denom.reshape((-1, 1))
         return self.gradients / denom
     
-    def __getitem__(self,bmin=0, bmax=np.inf):
+    def __getitem__(self, idx, threshold = 200):
         # Create a mask for selecting b-values greater than threshold
         if isinstance(idx, int):
             idx = [idx]
@@ -224,10 +224,13 @@ class GradientTable(object):
         bvecs_selected = self.bvecs[:, idx][:, mask]
         
         # Create a new MyGradientTable object with the selected b-values and b-vectors
-        return gradient_table_from_bvals_bvecs(bvals_selected, bvecs_selected, big_delta=self.big_delta,
-                                           small_delta=self.small_delta,
-                                           b0_threshold=self.b0_threshold,
-                                           atol=self.atol, btens=self.btens)
+        return gradient_table_from_bvals_bvecs(
+            bvals_selected, 
+            bvecs_selected, 
+            big_delta=self.big_delta,
+            small_delta=self.small_delta,
+            b0_threshold=self.b0_threshold,
+            atol=self.atol, btens=self.btens)
     
     @property
     def info(self, use_logging=False):
--- a/dipy/core/gradients.py
+++ b/dipy/core/gradients.py
@@ -210,7 +210,26 @@ class GradientTable(object):
         denom = self.bvals + (self.bvals == 0)
         denom = denom.reshape((-1, 1))
         return self.gradients / denom
-
+    
+    def __getitem__(self, idx):
+        # Create a mask for selecting b-values greater than threshold
+        if isinstance(idx, int):
+            idx = [idx]
+        elif isinstance(idx, slice):
+            idx = range(*idx.indices(len(self.bvals)))
+        mask = self.bvals[idx] > 200
+        
+        # Apply the mask to select the desired b-values and b-vectors
+        bvals_selected = self.bvals[idx][mask]
+        bvecs_selected = self.bvecs[:, idx][:, mask]
+        
+        # Create a new MyGradientTable object with the selected b-values and b-vectors
+        return gradient_table_from_bvals_bvecs(bvals_selected, bvecs_selected, big_delta=0.03,
+                                           small_delta=0.01,
+                                           b0_threshold=50,
+                                           atol=1e-2, btens=None)
+    #gave default value for small_delta, big_delta and b0_threshold
+    
     @property
     def info(self, use_logging=False):
         show = logging.info if use_logging else print
--- a/dipy/core/gradients.py
+++ b/dipy/core/gradients.py
@@ -212,15 +212,24 @@ class GradientTable(object):
         denom = denom.reshape((-1, 1))
         return self.gradients / denom
     
-    def getitem(self, bmin=0, bmax=np.inf):
-        if bmin != self.b0_threshold:
-            self.b0_threshold = bmin
-            warn("Updating b0_threshold to {} for slicing.".format(bmin), UserWarning, stacklevel=2)
-        # priority to bmin in all other cases
-        mask = (self.bvals >= bmin) & (self.bvals <= bmax)
+    def __get_item__(self, idx):
+        if isinstance(idx, int):
+            idx = [idx]#convert in a list if integer.
+            
+        elif isinstance(idx, slice):
+        # Get the lower bound of the slice
+            slice_start = idx.start if idx.start is not None else 0
+        # Check if it is different from b0_threshold
+            if slice_start != self.b0_threshold:
+                # Update b0_threshold and warn the user
+                self.b0_threshold = slice_start
+                warn("Updating b0_threshold to {} for slicing.".format(slice_start), UserWarning, stacklevel=2)
+                idx = range(*idx.indices(len(self.bvals)))
+
+        mask = self.bvals[idx] > self.b0_threshold
         # Apply the mask to select the desired b-values and b-vectors
-        bvals_selected = self.bvals[mask]
-        bvecs_selected = self.bvecs[mask, :]#since bvecs has shape:(N,3)
+        bvals_selected = self.bvals[idx][mask]
+        bvecs_selected = self.bvecs[idx, :][mask, :]
     
         # Create a new MyGradientTable object with the selected b-values and b-vectors
         return gradient_table_from_bvals_bvecs(bvals_selected, 
--- a/dipy/core/gradients.py
+++ b/dipy/core/gradients.py
@@ -220,7 +220,7 @@ class GradientTable(object):
         mask = (self.bvals >= bmin) & (self.bvals <= bmax)
         # Apply the mask to select the desired b-values and b-vectors
         bvals_selected = self.bvals[mask]
-        bvecs_selected = self.bvecs[:, mask]
+        bvecs_selected = self.bvecs[mask, :]#since bvecs has shape:(N,3)
     
         # Create a new MyGradientTable object with the selected b-values and b-vectors
         return gradient_table_from_bvals_bvecs(bvals_selected, 
@@ -228,8 +228,8 @@ class GradientTable(object):
                                                big_delta=self.big_delta,
                                                small_delta=self.small_delta,
                                                b0_threshold=self.b0_threshold,
-                                               atol=self.atol, 
                                                btens=self.btens)
+        #removing atol parameter as it's not in the constructor of GradientTable.
     
     @property
     def info(self, use_logging=False):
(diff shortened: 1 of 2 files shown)


arokem:
Yeah - I like it! Could you please add a couple of tests of this new functionality?

shilpiprd:
> Yeah - I like it! Could you please add a couple of tests of this new functionality?

I've added the tests, please review them.

shilpiprd:
Sure

dipy/core/gradients.py
```diff
@@ -210,7 +210,26 @@ def bvecs(self):
...
+        # Apply the mask to select the desired b-values and b-vectors
+        bvals_selected = self.bvals[idx][mask]
+        bvecs_selected = self.bvecs[:, idx][:, mask]
+        
+        # Create a new MyGradientTable object with the selected b-values and b-vectors
+        return gradient_table_from_bvals_bvecs(bvals_selected, bvecs_selected, big_delta=0.03,
+                                           small_delta=0.01,
+                                           b0_threshold=50,
+                                           atol=1e-2, btens=None)
+    #gave default value for small_delta, big_delta and b0_threshold
```
skoudoro, line 231: Why default value?  I recommend to get the value already initialized like  `self.big_delta , self.small_delta , etc`

shilpiprd: You rea right. I'll change this as well.

shilpiprd:
I updated the getitem method, and realised that the test file needed to be changed as well.I changed the test file in this branch. Should I close the #2750 now? 
Please review and let me know if any other changes needs to be made.

arokem:
Yeah. I'll close the other one.

dipy/core/gradients.py
```diff
@@ -210,7 +210,25 @@ def bvecs(self):
         denom = self.bvals + (self.bvals == 0)
         denom = denom.reshape((-1, 1))
         return self.gradients / denom
-
+    
+    def __getitem__(self, idx, threshold = 200):
```
arokem, line 214: I would set the threshold to be 0 per default. 

One more thought is to implement a lower and upper bound, so that if you index: 

```
gtab[:, 100:1000]
```

you would get a new gradient table with all the b-values between 100 and 1,000. I can see some use-cases for this kind of selection.

arokem (commented):
I've thought about this a bit more. What do you think about indexing with a lower and upper bound on b-values?

dipy/core/gradients.py
shilpiprd: I was thinking that we can probably remove the threshold argument and associated code, as it is no longer needed with the new indexing scheme.
Since now we could have: def __getitem__(self, bmin=0, bmax=np.inf):
And to create mask we can simply use: mask = (self.bvals >= bmin) & (self.bvals <= bmax). Using this to create mask we can remove a lot of unnecessary code as well. 
and # Select b-values between 100 and 1000, 
gtab = gtab[bmin=100, bmax=1000]
What do you think of this?

skoudoro: I like the idea! 

however, a deprecation cycle will be needed. a lot people are using this default behavior so we will have to keep both behavior for 1 year (2-3 releases) at least.

shilpiprd: So are you sugggesting something like this? 
def __getitem__(self, bmin=0, bmax=np.inf, threshold=0):
    # Use the new behavior if bmin and bmax are provided
    if bmin != 0 or bmax != np.inf:
        mask = (self.bvals >= bmin) & (self.bvals <= bmax)
    else:
        # Use the old behavior if bmin and bmax are not provided
        mask = self.bvals > threshold

skoudoro: > But I don't quite understand the part where the interp parameter is being deprecated from version 1.19 onwards. Can you explain more on that.

typo, replaced by `threshold`

shilpiprd: Oh right, so should i create a commit for this? I mean the method looks correct.

skoudoro: yes, update the code and test it to make sure it works as expected.

shilpiprd: I modified the gradients file and tests_gradients.py as well. Please review it.

shilpiprd: Could somebody please manually trigger the checks on my last commit?

shilpiprd:
The test file is passing all checks. It is failing one check but I can't seem to figure out where the error is because it says PASSED everywhere. 
Is there a need to look into this ? or is it mergable?

dipy/core/gradients.py
```diff
@@ -210,7 +211,26 @@ def bvecs(self):
         denom = self.bvals + (self.bvals == 0)
         denom = denom.reshape((-1, 1))
         return self.gradients / denom
-
+    
+    @deprecated_params('threshold', None, since='1.17', until='1.19')
+    def getitem(self, bmin=0, bmax=np.inf, threshold=0):
```
skoudoro, line 216: I think, we got lost here.  

The goal was to have the slicing capabilities. However, we loose it from your initial implementation.  
we can not do anymore gtab[100:200].

we should get back to that. we just need a warning if the slicing is really different from the `b0_treshold`.

Sorry for that. Can you revert your code, use `__get_item__`, and in case of a slice, you check if the lower bound is similar to `b0_thresold`. If yes, nothing to do, if no, just update self.b0_thresold to this value.

Does it make sense?   we do not need(bmin, bmax, thresold)

skoudoro (commented):
Hi @shilpiprd,

See below my comment and sorry for the late review.

dipy/core/gradients.py
shilpiprd: Sure, you're right. We did get deviated from the original method. I changed it, Did you mean something like this? If you'd agree with this, i'll change the test cases as well.

dipy/core/gradients.py
```diff
@@ -210,7 +211,26 @@ def bvecs(self):
         denom = self.bvals + (self.bvals == 0)
         denom = denom.reshape((-1, 1))
         return self.gradients / denom
-
+    
+    def getitem(self, bmin=0, bmax=np.inf):
+        if bmin != self.b0_threshold:
+            self.b0_threshold = bmin
+            warn("Updating b0_threshold to {} for slicing.".format(bmin), UserWarning, stacklevel=2)
```
skoudoro, line 218: yes, I mean something like this

dipy/core/gradients.py
```diff
@@ -210,7 +211,26 @@ def bvecs(self):
         denom = self.bvals + (self.bvals == 0)
         denom = denom.reshape((-1, 1))
         return self.gradients / denom
-
+    
+    def getitem(self, bmin=0, bmax=np.inf):
```
skoudoro, line 215: here it should be __get_item__(self, idx):

dipy/core/gradients.py
```diff
@@ -210,7 +211,26 @@ def bvecs(self):
...
+        if bmin != self.b0_threshold:
+            self.b0_threshold = bmin
+            warn("Updating b0_threshold to {} for slicing.".format(bmin), UserWarning, stacklevel=2)
+        # priority to bmin in all other cases
+        mask = (self.bvals >= bmin) & (self.bvals <= bmax)
```
skoudoro, line 220: like you did before, check if `slice` or just `int`

dipy/core/gradients.py
shilpiprd: Sure, I'll change this

dipy/core/gradients.py
shilpiprd: I'm sorry I didn't understand. Can u elaborate on what needs to be checked?

shilpiprd: Were you referring to bmin value?

shilpiprd: Oh, sorry. I understood what you meant.

shilpiprd:
I've added the idx parameter and modified the test file as well. Can someone please review it and allow for workflow checks to run?

shilpiprd:
I noticed that the failing checks are not from the method that I"ve implemented.

dipy/core/tests/test_gradients.py
```diff
@@ -461,6 +460,26 @@ def test_generate_bvecs():
...
+    # Create a GradientTable object with some test b-values and b-vectors
+    bvals = np.array([0, 100, 200, 300, 400])
+    #value should be in increasing order as b-value affects the diffusion weighting of the image, and the amount of diffusion weighting increases with increasing b-value.
+    bvecs = np.array([[0, 0, 1], [1, 0, 0], [0, 1, 0], [0, 0, 1], [1, 0, 0]])
+    #the b-vectors should be unit-length vectors
+    gradients = bvals[:, None] * bvecs
+    gtab = GradientTable(gradients) 
+
+    # Test with a single index
+    gtab_slice1 = gtab.__get_item__(1)
```
skoudoro, line 473: the goal was to do `gtab[1]` for example

dipy/core/tests/test_gradients.py
```diff
@@ -461,6 +460,26 @@ def test_generate_bvecs():
...
+    assert np.array_equal(gtab_slice1.bvals, np.array([100.]))
+    assert np.array_equal(gtab_slice1.bvecs, np.array([[1., 0., 0.]]))
+
+    # Test with a range of indices
+    gtab_slice2 = gtab.__get_item__(slice(2, 5))
```
skoudoro, line 478: same, the goal is to do `gtab[2:5]` and not `gtab.__get_item__`.. Please can you update

skoudoro (commented):
Hi @shilpiprd,

Can you address my comments below? Thank you

dipy/core/tests/test_gradients.py
shilpiprd: Sure

shilpiprd: It seems there was a typo in the test file. I changed it, it should pass all the tests now.

shilpiprd:
I don't think it's my method that's failing. It says 1 check failing, but inside it says PASSED for all.

skoudoro:
Thank you @shilpiprd, great job for your first contribution! it works like a charm. merging.

jhlegarreta:
I was debugging some code and had a look at this line:
https://github.com/dipy/dipy/commit/3fae50e79d6830627633ba84839499134ad63c6b#diff-bbd3ad0756880932ac2142d609f873674431fa6ad7a187ae379bba84ab2ec06fR225

This certainly looks wrong (or weird to say the least): `self.b0_threshold` is a b-value, and `slice_start` is an index, so it makes no sense to assign an index value to `self.b0_threshold`.

The comparison of the `slice_start` variable (an index) to the b0 threshold does not look right as well.

I have gone through this:
https://github.com/dipy/dipy/pull/2749#discussion_r1153704938

And it is unclear to me why the b0 threshold value should be updated if we are just doing a slicing operation (at least in the most general case of the slicing operation).

Should I be wrong and the implementation be OK, @shilpiprd @skoudoro can this be explained in a comment in the code, please?

skoudoro:
> This certainly looks wrong (or weird to say the least): self.b0_threshold is a b-value, and slice_start is an index, so it makes no sense to assign an index value to self.b0_threshold.

yep, looks wrong to me too after a quick look. I suppose the idea was 
`self.self.b0_threshold = self.b0_threshold[slice_start]`. but I do not remember the context.  I am not even sure if doing that make sense.

I will go through the discussion later tonight to refresh my memory

shilpiprd:
@jhlegarreta , Actually, I think you might be right. I'm not sure of the context since I didn't write that code. But at the first glance, it does seem like index assignment is wrong.

PR #2751, opened 2023-03-21
[ENH] Asymmetric peak_directions

CHrlS98:
Hello,

Add a `is_symmetric` flag to `dipy.direction.peaks.peak_directions` to enable peak extraction on asymmetric ODFs. `is_symmetric` is `True` by default, so that the default behaviour does not change. When set to `False`, the sphere directions `v` and `-v` are considered distinct, such that it is possible to find a maxima along `v` but not along `-v`. 

I simply removed the absolute value in `remove_similar_vertices` so that opposite sphere directions are not considered close to each other.

Have a good one!

### DIFF
--- a/dipy/direction/peaks.py
+++ b/dipy/direction/peaks.py
@@ -112,7 +112,7 @@ def peak_directions(odf, sphere, relative_peak_threshold=.5,
     min_separation_angle : float in [0, 90]
         The minimum distance between directions. If two peaks are too close
         only the larger of the two is returned.
-    is_symmetric : bool
+    is_symmetric : bool, optional
         If True, v is considered equal to -v.
 
     Returns
--- a/dipy/direction/peaks.py
+++ b/dipy/direction/peaks.py
@@ -113,7 +113,7 @@ def peak_directions(odf, sphere, relative_peak_threshold=.5,
         The minimum distance between directions. If two peaks are too close
         only the larger of the two is returned.
     is_symmetric : bool
-        When True, vertices v and -v are considered the same
+        If True, v is considered equal to -v.
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


dipy/reconst/recspeed.pyx
```diff
@@ -57,6 +58,8 @@ def remove_similar_vertices(
     return_indices : {False, True}, optional
         If True, return `indices` as well as `vertices` and maybe `mapping`
         (see below).
+    is_symmetric : {False, True}, optional
```
gabknight, line 61: Since this is in independent function, I suggest renaming the parameter `is_symmetric` to `remove_antipodale`

dipy/reconst/recspeed.pyx
```diff
@@ -57,6 +58,8 @@ def remove_similar_vertices(
...
+        If False, v is considered different from -v.
```
gabknight, line 63: I suggest changing to `If True, v is considered equal to -v.`

gabknight: Could you also edit lines 45-46 and remove the sentence "Vertex v and -v are considered the same so if v and -v are both in  `vertices` only one is kept."

dipy/direction/peaks.py
```diff
@@ -112,6 +112,8 @@ def peak_directions(odf, sphere, relative_peak_threshold=.5,
     min_separation_angle : float in [0, 90]
         The minimum distance between directions. If two peaks are too close
         only the larger of the two is returned.
+    is_symmetric : bool
+        When True, vertices v and -v are considered the same
```
gabknight, line 116: I suggest changing to : `If True, v is considered equal to -v.`

gabknight (commented):
Thanks @CHrlS98. Looks good to me. Just a few minor suggestions.

dipy/direction/peaks.py
```diff
@@ -112,6 +112,8 @@ def peak_directions(odf, sphere, relative_peak_threshold=.5,
     min_separation_angle : float in [0, 90]
         The minimum distance between directions. If two peaks are too close
         only the larger of the two is returned.
+    is_symmetric : bool
```
skoudoro, line 115: `bool, optional` instead

skoudoro:
Thanks @CHrlS98, merging!

Issue #2754, opened 2023-03-22
[Question] dipy/denoise/gibbs.py


tashrifbillah:
https://github.com/dipy/dipy/blob/77afd6098c70939dde1647e71fa1d6c7be1ef48e/dipy/denoise/gibbs.py#L322-L324

The copied volume is overwritten below anyway:

https://github.com/dipy/dipy/blob/77afd6098c70939dde1647e71fa1d6c7be1ef48e/dipy/denoise/gibbs.py#L327-L335

So what is the point of using `inplace` parameter?

skoudoro:
Good catch, I think this parameter is useless. Either we need to remove the parameter or make ` inplace=False` work.  Low priority. 

Feel free to fix this issue @tashrifbillah !

skoudoro:
After playing with the code, I think I made a mistake. The `inplace` argument works as expected and avoid you to modify the input volume. 

So closing the issue and the PR

PR #2756, opened 2023-03-23
Remove unusued inplace param from gibbs_removal()

tashrifbillah:
Closes issue #2754


skoudoro (commented):
Hi @tashrifbillah,

Thank you cleaning up!

You might need to update test files and tutorials.
Also, we need to give time to user to remove this parameter from their script. 

So, you need to keep the variable and the code, and add it as a deprecated parameter
something like:
```python
@deprecated_params(['num_threads', 'inplace'], ['num_processes', None], since='1.4', until='1.8')
```

skoudoro:
closing in favor of my last comment [here](https://github.com/dipy/dipy/issues/2754#issuecomment-1516642508)

PR #2759, opened 2023-03-24
TRX integration

frheault:
Integration of TRX into the StatefulTractogram API to allow for easy loading and saving in Dipy.

This does not really use the memmap from the core repo, since the first thing it does is put it into memory and use the SFT class. Still, this will facilitate the use of TRX.

The next goal is to integrate: https://github.com/tee-ar-ex/trx-python/blob/master/scripts/tff_convert_tractogram.py
as a workflow in Dipy.

This PR is simply to add the dependency (and make tests work) and simply support TRX in the SFT (`load`/`save`) and in utils function (`get_reference_info`)


dipy/io/streamline.py
```diff
@@ -73,6 +74,10 @@ def save_tractogram(sft, filename, bbox_valid_check=True):
         dpy_obj = Dpy(filename, mode='w')
         dpy_obj.write_tracks(sft.streamlines)
         dpy_obj.close()
+    elif extension in ['.trx']:
+        trx = tmm.TrxFile.from_sft(sft)
+        print(trx)
```
skoudoro, line 79: print to remove

requirements.txt
```diff
@@ -6,3 +6,4 @@ nibabel>=3.0.0
 h5py>=2.8.0
 packaging>=19.0
 tqdm>=4.30.0
+trx-python==0.1
```
skoudoro, line 9: why not >=0.1 ?

skoudoro (commented):
Thank you for this work @frheault, this looks good to me, I will test it locally , and give you feedback

requirements.txt
frheault: I think that at least while we 'move' things it's safer to fix it until it's done.

Especially since we will be cleaning the 'core' TRX repo as we move things into Dipy.

(The goal is not to break anything, but this is only a precaution)

frheault:
It seems like there is a few problems. I see error on windows due to the tmp folder being read-only and a problem because trx imports from numpy.typing (I think that's 3.9-3.10). So I will have to look into that (how to remove it from trx).

I am not sure how @skoudoro or @arokem since this is a problem right at the start. @arokem maybe if you want to remove all need for numpy.typing and do a PR in trx?

Otherwise, there is not much to do in relation to this PR.

arokem:
Are you sure that typing is not supported in 3.8? The discussion in the numpy PR that introducted np.typing seems to suggest that it might be: https://github.com/numpy/numpy/pull/16515. 

And it would be great to be able to include typing going forward.

arokem:
Maybe we could drop 3.7 support? I think that would be fine for most users to that?

arokem:
The trx-python library is being tested on 3.7 - 3.10 (https://github.com/tee-ar-ex/trx-python/actions/runs/4469447927/jobs/7851571051), so I don't think that python version is an issue.

arokem:
It is however new to numpy: version 1.20, I believe so we would have to up the numpy minimal version by quite a bit: https://github.com/dipy/dipy/blob/master/dipy/info.py#L78

skoudoro:
Not sure what you are trying to do by removing `src`folder.  It is needed by all our cython packages.

Also, not sure why you removed `packaging` and replaced it by  pkg_resources.

Can you remove some of your last commit, it will help for debugging your issue.

Thanks!

frheault:
I will remove a few of the last commit, the fact that dipy cannot be installed without a release (with pip and a URL)  is making my testing really hard.
There is no way for pip (when using an URL) to know it has to install cython and packaging first and everything crash is very annoying for this parallel integration.

skoudoro:
> the fact that dipy cannot be installed without a release (with pip and a URL) is making my testing really hard.

What do you mean by that?  if you have a compiler and do `pip install .` on the source directory. it just works. You need to provide more details.

Not sure I understand your complain. it is like many scientific package (numpy, scipy, scikit-image, etc....). you need a compiler if you want to install it without a wheel and it just works.

Garyfallidis:
@frheault what you say does not make sense. Of course it works. For ages we have this option.

arokem:
Thanks Eleftherios. Maybe you can tell us why this is not working? https://github.com/tee-ar-ex/trx-python/pull/57

skoudoro:
Maybe related to the last issue pointed on TRX @arokem?  https://github.com/tee-ar-ex/trx-python/issues/56

arokem:
Yeah. Good call. I'll give that a try

frheault:
I mean using something like pip install `dipy@git+https://git@github.com/dipy/dipy@ba3ce0c59#egg=dipy`

From an URL, of course it works locally when cloned first with `pip install -e .`

arokem:
I'm trying to debug this on https://github.com/tee-ar-ex/trx-python/pull/57 but no luck so far.

skoudoro:
> I'm trying to debug this on https://github.com/tee-ar-ex/trx-python/pull/57 but no luck so far.

Not sure why `Cython` is not found since you just install it  anything else.  I will try to look deeper when I have time.

arokem:
Yeah. It's really quite confounding.

skoudoro:
Hi @frheault,

I confirm that `pip install dipy@git+https://git@github.com/dipy/dipy@ba3ce0c59#egg=dipy`  is working. I use it in fury and never had any issues.

After looking with Ariel, it seems that the issue is coming from `pyproject.toml`. We do not have this file yet, this is a work in progress. More information in the discussion on this PR https://github.com/tee-ar-ex/trx-python/pull/57

skoudoro:
Hi @frheault,

what is the status of this PR? how can we help you? 

We missed the last release, but I would this to be in before the next one.

please, keep us updated. Thanks !

dipy/info.py
```diff
@@ -81,6 +81,7 @@
 H5PY_MIN_VERSION = '3.7.0'
 PACKAGING_MIN_VERSION = '19.0'
 TQDM_MIN_VERSION = '4.30.0'
+TRX_MIN_VERSION = '0.0'
```
arokem, line 84: Should we bump this to the already-released 0.1.2?

arokem (commented):
Of potential interest here is SPEC0, which specifies minimally supported versions of the various components of the scientific python ecosystem: https://scientific-python.org/specs/spec-0000/. This suggests we can bump up the numpy dependency to 1.22. I'll try implementing that as a change and pushing it into this branch (heads up @frheault that you might need to pull down your branch after I do that).

.github/workflows/test_compat.yml
```diff
@@ -16,16 +16,16 @@ jobs:
     with:
       runs-on: '["ubuntu-latest", ]'
       python-version: '["3.8", ]'
-      depends: cython==0.29.24 numpy==1.17.5 scipy==1.3.2 nibabel==3.0.0 h5py==3.0.0 tqdm
+      depends: cython==0.29.24 numpy==1.20.3 scipy==1.3.2 nibabel==3.0.0 h5py==3.0.0 tqdm
```
arokem, line 19:
```suggestion
      depends: cython==0.29.24 numpy==1.22.0 scipy==1.3.2 nibabel==3.0.0 h5py==3.0.0 tqdm
```

.github/workflows/test_compat.yml
```diff
@@ -16,16 +16,16 @@ jobs:
...
       python-version: '["3.8", ]'
-      depends: cython==0.29.24 numpy==1.17.5 scipy==1.3.2 nibabel==3.0.0 h5py==3.0.0 tqdm
+      depends: cython==0.29.24 numpy==1.22.0 scipy==1.3.2 nibabel==3.0.0 h5py==3.0.0 tqdm
   minimal-py39:
     uses: ./.github/workflows/test_template.yml
     with:
       runs-on: '["ubuntu-latest", ]'
       python-version: '["3.9", ]'
-      depends: cython==0.29.24 numpy==1.19.4 scipy==1.5.4 nibabel==3.0.0 h5py==3.1.0 tqdm
+      depends: cython==0.29.24 numpy==1.20.3 scipy==1.5.4 nibabel==3.0.0 h5py==3.1.0 tqdm
```
arokem, line 25:
```suggestion
      depends: cython==0.29.24 numpy==1.22.0 scipy==1.8 nibabel==4.0.0 h5py==3.1.0 tqdm
```

.github/workflows/test_compat.yml
```diff
@@ -16,16 +16,16 @@ jobs:
     with:
       runs-on: '["ubuntu-latest", ]'
       python-version: '["3.8", ]'
-      depends: cython==0.29.24 numpy==1.17.5 scipy==1.3.2 nibabel==3.0.0 h5py==3.0.0 tqdm
+      depends: cython==0.29.24 numpy==1.22.0 scipy==1.3.2 nibabel==3.0.0 h5py==3.0.0 tqdm
```
arokem, line 19:
```suggestion
      depends: cython==0.29.24 numpy==1.22.0 scipy==1.8 nibabel==4.0.0 h5py==3.0.0 tqdm
```

arokem:
Also updated nibabel and scipy dependencies. Nibabel keeps track of this via a script: https://github.com/nipy/nibabel/blob/master/tools/update_requirements.py

frheault:
I don't understand the error
ImportError: numpy.core.multiarray failed to import (auto-generated because you didn't call 'numpy.import_array()' after cimporting numpy; use '<void>numpy._import_array' to disable if you are certain you don't need it).

Its not happening in my code and I don't think I changed anything close to it and it's installing the same numpy version as any other PR

.github/workflows/test_compat.yml
```diff
@@ -16,16 +16,16 @@ jobs:
     with:
       runs-on: '["ubuntu-latest", ]'
       python-version: '["3.8", ]'
-      depends: cython==0.29.24 numpy==1.17.5 scipy==1.3.2 nibabel==3.0.0 h5py==3.0.0 tqdm
+      depends: cython==0.29.24 numpy==1.22.0 scipy==1.8 nibabel==4.0.0 h5py==3.0.0 tqdm
```
skoudoro, line 19: I suppose it is due to this change @frheault. 

you can look at my PR #2715 to update those values.

Also, my PR #2715 for integrating pyproject.toml is almost done. So I suppose, it will help TRX

frheault: Should I wait or copy the values in all yml files under workflow?

arokem:
Looks like the remaining failures are somehow related to the Windows filesystem. See comments by user named Eryk Sun at the bottom of this SO post: https://stackoverflow.com/questions/40130958/python-2-error-32-the-process-cannot-access-the-file-because-it-is-being-used?rq=3

frheault:
@arokem Assuming this is <fine> for now, could you try loading and saving a few of your TRX using the SFT from Dipy and see if you get anything unexpected?

skoudoro:
> Assuming this is "fine" for now, 

this is hard to have this assumption when all Windows CI's are failing 😅. I will connect later to a windows machine to see what is going on.

I saw that TRX repo has no test on windows or macOS. Maybe there is a deeper issue to look at.

arokem:
I wonder if the fact that Windows tests are failing here has to do with the concurrency in the testing: 

https://github.com/dipy/dipy/blob/master/.github/workflows/test.yml#L15 

Could you try removing that to see if it makes a difference? It will be a bit slower, but will help us narrow this down a bit.

skoudoro:
> I wonder if the fact that Windows tests are failing here has to do with the concurrency in the testing:https://github.com/dipy/dipy/blob/master/.github/workflows/test.yml#L15

No for sure. I do not think they share the same environment (that's why each one has a checkout).
This line of concurrency is to avoid to build when you have 2 successive push commits.  So it kills the build of the oldest commit and build the new one.

I will be surprised if it is related

skoudoro:
What I do not understand: 

when you call `trx.close` during `load_tractogram`, does it really close?  After the close, you are still using sft which contains a memory map from the file that you just close...

Not sure I am clear, but this part of TRX confuse me. I feel like you can not close if you do not finish with memorrymap operation. Otherwise, some explicit copy need to be done. 

Can you clarify ?

frheault:
Ahh, this is a great question. For now this PR simply allow to load the TRX and put it in memory into the form of the SFT. It is not really a support of the memmap (as soon as it is loaded it is all in memory).

To really use the memmap capacities I would have to use tmm.load() only and not transfer it into the SFT class.

I think that within the next few months (future PR) I will try to add some scripts to do some basic operation memory free (convert, compress, concat, etc.). But for now this PR is basically not using memmap, straight to SFT in memory.

skoudoro:
> So to answer your question: Since the SFT is all in memory, closing the TRX is fine.

ok, thanks for the clarification!  So now, I am lost with the issue you encounter... 😅

frheault:
I tried it on my personal Windows partition and everything is being deleted correctly and our tests are now much more robust thanks to your comments!

Do you think that @arokem suggestions to put the tmp directories in the .dipy folder could be an acceptable workaround?

It seems like it is working as intended, but since we (at least I) dont have a way of inspecting the folders of the runners to see what is happening (is data accumulating when the scope is exited) I dont know if it is fine or catastrophic.

frheault:
I tried it on my personal Windows partition and everything is being deleted correctly and our tests are now much more robust thanks to your comments!

Do you think that @arokem suggestions to put the tmp directories in the .dipy folder could be an acceptable workaround?

It seems like it is working as intended, but since we (at least I) dont have a way of inspecting the folders of the runners to see what is happening (is data accumulating when the scope is exited) I dont know if it is fine or catastrophic.

arokem:
At least it's a different error now! 🙃

frheault:
I discovered that my code was running in Pycharm and in Conda differently and depending if I used Conda create vs pip install in one or the other I was getting a ton of previously unseen errors!

I did not keep track of exactly which errors were from where/when/why, but overall it seems to be due to an overwriting problem. Writing over an opened file in Linux and MacOS seems to be alright, but in Windows sometimes the IDE/environments make it either impossible (PermissionError) or let you do it, but then it is impossible to close the initial files. 

Not sure exactly what is happening and why it was passing on trx-python GitHub action and not Dipy, but anyway... I am tired of this for now!

@arokem @skoudoro So now all tests are passing in: Github trx-python (all 3 OS) and on my personal Linux (Ubuntu 22.04, Python 3.8 - 3.11) and on my personal Windows 11 - Python3.10. After discovering that behavior from Conda, Pycharm was different I ran all tests in a few virtualenv created from Pycharm, Conda, manually, inside a Powershell terminal, Spyder, Pycharm Terminal and AnacondaPrompt and they all are passing now.

I was getting weird/different errors on Pycharm/Conda and I think it is due to the IDE/environnement trying to make the execution of code safer (artificially or something)

arokem:
So, is there a discrepancy because on DIPY we use conda for installation
and on TRX we don't?

skoudoro:
> So, is there a discrepancy because on DIPY we use conda for installation and on TRX we don't?

To be correct, only 1 CI matrix in windows is using conda, all the others CI matrix are using, pip + virtualenv.

frheault:
> > So, is there a discrepancy because on DIPY we use conda for installation and on TRX we don't?
> 
> To be correct, only 1 CI matrix in windows is using conda, all the others CI matrix are using, pip + virtualenv.

That's true, then I have no idea why our trx-python test were working, but all Dipy Windows-Latest were failing.

But now it's working, and now we have 15 new tests on our sides and a few new ones in Dipy thanks to this craziness!

arokem:
> But now it's working, and now we have 15 new tests on our sides and a few new ones in Dipy thanks to this craziness!

Wait, what?! I now also see this on the tests for https://github.com/dipy/dipy/pull/2759/commits/9550451a7e6a87e3d56e91a9042522c49a08329b. 

That's amazing!! 🎉 🥳

frheault:
Yeah they all passed, but one Coverage check failed, so I tried a new test to put it back up a bit.

But I'm done I think 😅

skoudoro:
Amazing ! Good job @frheault !  What a intensive and tricky problem ! So finally, what was the issue?  

Conclusion is: test in DIPY seems to be useful lol 😅 

I recommend you to yank some TRX version in Pypi, it will avoid some user to fall in this trap, in case some of them pinned a wrong TRX version.

skoudoro:
> I would be curious to wait after https://github.com/dipy/dipy/pull/2715 get merged to see if everything works well with the new building with meson!

I will rebase #2715 after we merge this PR and we will see but I believe it should be fine.

arokem:
> I recommend you to yank some TRX version in Pypi, it will avoid some user to fall in this trap, in case some of them pinned a wrong TRX version.

I doubt anyone is depending on any of those versions (or on trx at all). It's all so new... 

I believe this is ready to merge! @skoudoro (or others) : any objections? Feel free to go ahead if not.

arokem:
I should add: I see that the coverage test is failing, but I suggest that we swallow the decrease of 0.02% coverage, given how much effort has already gone into testing this...

requirements.txt
```diff
@@ -5,4 +5,5 @@ scipy>=1.8.1
 nibabel>=4.0.0
 h5py>=3.7.0
 packaging>=19.0
-tqdm>=4.30.0
\ No newline at end of file
+tqdm>=4.30.0
+trx-python>=0.2.3
```
skoudoro, line 9: Can you update the version here ? thanks

dipy/io/utils.py
```diff
@@ -223,15 +225,42 @@ def is_reference_info_valid(affine, dimensions, voxel_sizes, voxel_order):
...
+    filename: str
+        The filename to clean
+
+    Returns
+    -------
+        base, ext : tuple(str, str)
+        Clean basename and the full extension
+    """
+    base, ext = os.path.splitext(filename)
+
```
skoudoro, line 244: Can you add here `ext = ext.lower()`.   

Some software do weird stuff (capitalize) and do `myfile.NII.GZ`. your code will failed here because of capitalization.

skoudoro (commented):
Hi @frheault,

Here some additional quick comments. After that, we can go ahead and merge this PR.

Thanks !

skoudoro:
> I should add: I see that the coverage test is failing, but I suggest that we swallow the decrease of 0.02% coverage, given how much effort has already gone into testing this...

Otherwise, I agree with this.

dipy/io/utils.py
frheault: I'm going to fix it in trx-python library too, tomorrow and I'll free!

frheault:
CodeCov (+0.05%)!

🎉

skoudoro:
Yeah !!! 🚀  Thank you for this work @frheault and @arokem!   

I am going ahead and merge this PR. We can now start the phase 2 of #2760.

.github/workflows/test.yml
```diff
@@ -21,4 +21,4 @@ jobs:
     uses: ./.github/workflows/test_template.yml
     with:
       runs-on: '["ubuntu-latest", "macos-latest", "windows-latest"]'
-      python-version: '["3.8", "3.9", "3.10", 3.11]'
+      python-version: '["3.9", "3.10", 3.11]'
```
jhlegarreta, line 24: @skoudoro So we've lost 3.8 here. Re: https://github.com/dipy/dipy/pull/2715#discussion_r1348580722. Will it be brought back?

skoudoro: > Will it be brought back?

No, I do not think so.

I might add it in here, https://github.com/dipy/dipy/blob/268b7182c273808e9d6eadb8e90c591a42ef6e02/.github/workflows/test_compat.yml#L14. but just until december.

jhlegarreta: > No, I do not think so.

Is there a particular reason? Is `trx` not compatible with 3.8 or is it that it was not tested? According to https://github.com/tee-ar-ex/trx-python/blob/master/.github/workflows/python-package.yml#L15, it is. Were DIPY tests failing for Python 3.8? Was it hard to keep it?

> . but just until december.

As mentioned in https://github.com/dipy/dipy/pull/2715#discussion_r1348580722,
- dropping a Python version should have been announced in the previous release.
- I am fine if from now one we will stick to the Scientific Python specs, but that should be made clear in the docs, and for now, keep 3.8 until the next release or EOL.

jhlegarreta: > Python3.8 end of life is in less than 1y. It is not a good idea to stop it at the same time as python EOL. if we do it, many user will still use a version without security patch.

Still unclear; the Python folks will patch versions as they maintain them. It is up to the users to use the latest patched version; that is not DIPY's task.

>  Better to stop it before and let them the time to migrate before the EOL.

But users have to be warned beforehand.

skoudoro: I recommend you to look at 

- https://numpy.org/neps/nep-0029-deprecation_policy.html
- https://github.com/nipy/nibabel/issues/803

this is what we approximately follow since a long time. Dropping python 3.8 in December does not sound so terrible. Numpy plan to do it in April, other package has already done it. Not sure what is the concern?  Yes documentation need to be updated concerning this topic and ok for announcing it.
Like I said above, we will keep it for this release and announce that we plan to drop it in the december release. 

@arokem, any opinion concerning this point ?

arokem: My opinion is that dropping 3.8 now is not a big deal, but that our documentation should also state explicitly that we are following https://scientific-python.org/specs/spec-0000/, so that users are not caught unawares. In fact, according to that, we should also be dropping 3.9 support pretty soon. So, dropping 3.8 now seems really OK.

jhlegarreta: Relevant to the discussion: https://endoflife.date/ and an example of a Python version dropping notice https://endoflife.date/numpy#python-support.

Issue #2736, opened 2023-02-27
dipy_horizon needs --force option if there is tmp.png


pjsjongsung:
## Description
dipy_horizon fails to open the file if there is an output file unless you use the --force option. (e.g. tmp.png) However, this is also the case even if you have no intention of creating an output file and just want to visualize something. I am not sure why it needs to be cautious about overwriting tmp.png, but judging by the name of the file, this does not seem necessary.

Below is the warning message:
INFO:The following output files already exist, the workflow will not continue processing any further. Add the --force flag to allow output files overwrite.
INFO:tmp.png

## Way to reproduce
[If reporting a bug, please include the following important information:]
- [ ] Code example : dipy_horizon file_path (while you have a file named tmp.png in the current directory)
- [ ] Operating system and version: Windows-10-10.0.19045-SP0
- [ ] Python version: 3.8.11
- [ ] dipy version 1.6.0
- [ ] dependency version (numpy, scipy, nibabel, h5py, cvxpy, fury)
- [ ] numpy : 1.22.4
- [ ] scipy: 1.7.1
- [ ] nibabel: 3.2.1
- [ ] h5py 3.4.0
- [ ] no cvxpy installed
- [ ] fury 0.8.0

pjsjongsung:
On a side note, we have cvxpy as dependency in the issue template but cvxpy does not seem to be a requirement when installing dipy. Is this intentional?

lb-97:
@pjsjongsung, there's a 'overwrite' checker in the HorizonFlow workflow that throws this warning. I added a helper function to avoid this warning in interactive mode.

maharshi-gor:
I have opened the PR https://github.com/dipy/dipy/pull/2969
which allow user to work without the --force option

maharshi-gor:
The PR is merged and the behavior is fixed as closing this issue.

PR #2770, opened 2023-03-31
BF - update viz.py

lb-97:
Issue #2736
Updated viz.py to stump warnings of overwriting 'tmp.png' file in interactive mode

### DIFF
--- a/dipy/workflows/viz.py
+++ b/dipy/workflows/viz.py
@@ -24,7 +24,15 @@ class HorizonFlow(Workflow):
     @classmethod
     def get_short_name(cls):
         return 'horizon'
-
+    
+    def manage_output_overwrite_helper(self):
+        '''
+        Overriding parent function to stump warnings of over writing existing
+        files as interactive mode of HorizonFlow doesn't attempt to write any 
+        output files
+        '''
+        return True
+    
     def run(self, input_files, cluster=False, cluster_thr=15.,
             random_colors=None, length_gt=0, length_lt=1000,
             clusters_gt=0, clusters_lt=10**8, native_coords=False,
@@ -140,6 +148,9 @@ class HorizonFlow(Workflow):
         emergency_ref = create_nifti_header(hdr['affine'], hdr['dims'],
                                             hdr['vox_size'])
 
+        if interactive:
+            self.manage_output_overwrite = self.manage_output_overwrite_helper
+            
         io_it = self.get_io_iterator()
 
         for input_output in io_it:


dipy/workflows/viz.py
```diff
@@ -24,7 +24,15 @@ class HorizonFlow(Workflow):
...
         return 'horizon'
-
+    
+    def manage_output_overwrite_helper(self):
+        '''
+        Overriding parent function to stump warnings of over writing existing
+        files as interactive mode of HorizonFlow doesn't attempt to write any 
+        output files
+        '''
+        return True
```
skoudoro, line 34: Hi @lb-97,

This is not a solution to overwrite this overwrite the helper. 

Please, look deeper to check why it is needed

lb-97: Hi @skoudoro,
Here's a summary of my understanding: _manage_output_overwrite_, is a native method of class Workflow and HorizonFlow is inherited from this class. The method above is responsible for generating an iterator of tuples of input, output files(to be generated). The output file names are identified by the arguments 'out_*', the default value of these variables i.e., 'tmp.png' is considered if nothing is explicitly passed. Now, this method generators an empty iterator if 'tmp.png' already exists by throwing a WARN message. 

In the interactive mode of HorizonFlow, the default 'out_*' argument is picked up by the parent method, although not desired and returns an empty iterator with a WARN message. When the iterator is empty, no input files can be visualized. 

The parent _manage_output_overwrite_ method is used at many other places in the code, so one way to generate an iterator but also override WARN message, which is how interactive mode should work, is to make change in the inherited class. That's exactly what my PR does. 

I believe I made this decision after careful consideration and looking deeper for hours before making a PR.

So, if there's a better way to do it I'm all ears for it. Looking forward.

dipy/workflows/viz.py
```diff
@@ -140,6 +148,9 @@ def run(self, input_files, cluster=False, cluster_thr=15.,
         emergency_ref = create_nifti_header(hdr['affine'], hdr['dims'],
                                             hdr['vox_size'])
 
+        if interactive:
+            self.manage_output_overwrite = self.manage_output_overwrite_helper
+
```
lb-97, line 153: @skoudoro, this part of the commit makes sure the change is effective only in interactive mode.

dipy/workflows/viz.py
skoudoro: - why using `tmp.png` and not `None`?
- why not just initialize `HorizonFlow` with `self._force_overwrite`?
- With your solution, you hide any overwrite which seems to not be desired behavior ?
- Are you sure that "HorizonFlow doesn't attempt to write any output files in interactive mode" ? 

Also, please:
- add tests to reproduce the issue and show the correction
- follow the library coding style (double quotes instead of single quotes for example)

all this note should help you to go deeper into it and follow the library standard.

lb-97: Hi @skoudoro ,
1. `tmp.png` is the default filename used in stealth mode. It's the expected name in other examples such as `save_tractogram()`.  Adding the below change does work but it's more of a hack than understanding root cause. My PR addresses the root cause of this warning, as well as override it.
`if interactive:
    out_stealth_png = ''
`
2. Initializing with `self._force_overwrite`, will output the message:
`The following output files are about to be overwritten.`
Although it's actually not overriding anything
3. The expected behavior of interactive mode of HorizonFlow is to open an interactive window but not save/override any file. I verified this by implementing it locally.
4. Yes.

Sure, I'd be happy to add any test cases. Since this is not a New Feature type issue, I am not sure what kind of new test cases I can introduce. 
Yes, I will change it to double quotes, my bad. I apologize.

Looking forward to hearing from you.

PR #2780, opened 2023-04-17
Added option to set Legacy=False in PmfGenDirectionGetter.from_shcoeff

### LINKED ISSUE #2775: NF - Add option to set Legacy=False in PmfGenDirectionGetter.from_shcoeff(.)
There is current the option to select the basis, but it then uses `Legacy=True`. A parameter should be added to `PmfGenDirectionGetter.from_shcoeff(.)` with the value passed to `dipy.direction.pmf.SHCoeffPmfGen`, line 76, when generating the `B` matrix.

MickyG03:
Fixed #2775

I have added the legacy parameter to all the connecting functions with default value set to True.

### DIFF
--- a/dipy/direction/pmf.pyx
+++ b/dipy/direction/pmf.pyx
@@ -74,7 +74,7 @@ cdef class SHCoeffPmfGen(PmfGen):
             basis = shm.sph_harm_lookup[basis_type]
         except KeyError:
             raise ValueError("%s is not a known basis type." % basis_type)
-        self.B, _, _ = basis(sh_order, sphere.theta, sphere.phi,legacy=legacy)
+        self.B, _, _ = basis(sh_order, sphere.theta, sphere.phi, legacy=legacy)
         self.coeff = np.empty(shcoeff_array.shape[3])
         self.pmf = np.empty(self.B.shape[0])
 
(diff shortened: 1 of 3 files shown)


skoudoro:
Thank you for this @MickyG03. 

@gabknight, can you have a quick look?  thank you

dipy/direction/pmf.pyx
```diff
@@ -62,7 +62,8 @@ cdef class SHCoeffPmfGen(PmfGen):
     def __init__(self,
                  double[:, :, :, :] shcoeff_array,
                  object sphere,
-                 object basis_type):
+                 object basis_type
```
skoudoro, line 65: missing coma here

MickyG03:
I have corrected that, can you please re-run the tests?

gabknight:
Thanks @MickyG03 !

Just one thing. Why the change is `dipy.reconst.csdeconv` file? I think the `dipy.reconst.csdconv.basis` function should never be called by `SHCoeffPmfGen`.  The `basis` function in `dipy/direction/pmf.pyx` line 77 is from `shm.sph_harm_lookup[basis_type]`, which return two basis functions of the `dipy.reconst.shm` module, both having the legacy option.

MickyG03:
I didn't realize that, when I was tracking the use of legacy parameter from function to function, I thought that the parameter ended up into the function of dipy.reconst.csdconv.basis. I'll change that back. I am not sure why installing dependencies (the coverage test ) is failing.

skoudoro:
the coverage test failing is not related, I will fix it in a new PR.

gabknight:
LGTM. Thanks @MickyG03

skoudoro:
Thank you @MickyG03, and thank you for the review @gabknight ! merging

Issue #2777, opened 2023-04-07
Error using dipy_motion_correct


araikes:
Hi @skoudoro ,

I'm attempting to use `dipy_motion_correct` on some data. As a test case, I have a file with 8 b=0 images at the front of the series and then 64 directions. When I run either `dipy_motion_correct` or follow the tutorial (https://dipy.org/documentation/1.6.0./examples_built/motion_correction/#example-motion-correction) using my data, I get the following error:

```
Traceback (most recent call last):
  File "/home/u26/adamraikes/.conda/envs/dipy/bin/dipy_correct_motion", line 7, in <module>
    run_flow(MotionCorrectionFlow())
  File "/home/u26/adamraikes/.conda/envs/dipy/lib/python3.10/site-packages/dipy/workflows/flow_runner.py", line 89, in run_flow
    return flow.run(**args)
  File "/home/u26/adamraikes/.conda/envs/dipy/lib/python3.10/site-packages/dipy/workflows/align.py", line 764, in run
    reg_img, reg_affines = motion_correction(data=data, gtab=gtab,
  File "/home/u26/adamraikes/.conda/envs/dipy/lib/python3.10/site-packages/dipy/align/_public.py", line 696, in register_dwi_series
    series_arr = np.concatenate([ref_data, moving_data], -1)
  File "<__array_function__ internals>", line 200, in concatenate
ValueError: all the input arrays must have same number of dimensions, but the array at index 0 has 3 dimension(s) and the array at index 1 has 4 dimension(s)
```

Functionally, I can see that this is a mismatch between the dimensions of the mean b0 image shape (64 96 64) and the moving data shape (64 96 64 64). What I can't see is how to fix it.

This is a clean conda install from today with Python 3.10.0 and only DIPY 1.6.0 (and it's requirements, particularly numpy `1.21.6`) installed.

pjsjongsung:
@araikes @skoudoro It seems to be a bug in the register_dwi_series function. Think line 648 of dipy.align._public should have keepdims=True. Some parts of the tutorial might need fixing too.

For the meanwhile, a workaround might be running something like this below

```
import nibabel as nib
from dipy.align._public import register_series
import numpy as np

pipeline = ["center_of_mass", "translation", "rigid", "affine"]
b0_img = nib.Nifti1Image(data[..., gtab.b0s_mask], affine)
trans_b0, b0_affines = register_series(b0_img, ref=0, pipeline=pipeline)
[...]

Issue #2563, opened 2022-03-10
Some tutorials take way too much time


skoudoro:
- In Green, tutorial < 5 min, they are fine
- in black, tutorial between 5-15min,  warning zone, we need to review some of them
- in red, tutorial more than 15min, danger zone, we need to do something

```diff
+ quick_start.py  That took 6.85 seconds to run
+ brain_extraction_dwi.py That took 32.29 seconds to run
+ reconst_csa_parallel.py  That took 141.54 seconds to run
+ reconst_csa.py That took 67.55 seconds to run
+ reconst_csd_parallel.py That took 33.74 seconds to run
+ reconst_csd.py That took 151.74 seconds to run
+ reconst_forecast.py That took 79.97 seconds to run
+ reconst_dki.py  That took 174.17 seconds to run
+ reconst_dki_micro.py That took 168.40 seconds to run
+ reconst_msdki.py That took 220.21 seconds to run
+ reconst_dsi_metrics.py That took 30.10 seconds to run
+ reconst_dsi.py That took 47.04 seconds to run
+ reconst_dti.py took 59.90 seconds to run
+ reconst_fwdti.py That took 72.81 seconds to run
+ reconst_gqi.py That took 5.33 seconds to run
+ reconst_dsid.py That took 0.56 seconds to run
- reconst_ivim.py That took 1250.27 seconds to run
- reconst_mapmri.py That took 1286.67 seconds to run
reconst_mcsd.py That took 583.18 seconds to run
reconst_qtdmri.py 545.49 seconds to run
reconst_rumba.py That took 898.77 seconds to run
+ reconst_sfm.py That took 178.08 seconds to run
+ reconst_sh.py That took 10.23 seconds to run
+ reconst_shore.py That took 3.41 seconds to run
+ reconst_shore_metrics.py That took 6.01 seconds to run
+ reconst_qti.py That took 87.10 seconds to run
+ kfold_xval.py That took 3.38 seconds to run
+ reslice_datasets.py That took 1.21 seconds to run
+ segment_quickbundles.py That took 1.87 seconds to run
+ segment_extending_clustering_framework.py That took 1.26 seconds to run
+ segment_clustering_features.py That took 2.35 seconds to run
+ segment_clustering_metrics.py That took 0.63 seconds to run
+ snr_in_cc.py That took 41.26 seconds to run
+ streamline_formats.py That took 11.30 seconds to run
+ gradients_spheres.py That took 7.50 seconds to run
[...]

PR #2781, opened 2023-04-17
Small fixes in functions

pjsjongsung:
I have added the keep_dims option to fix the error on register_dwi_series in #2777 

I have also included np.moveaxis to make the affine output correspond to other functions by having a shape of (n, 4, 4) instead of (4, 4, n). Also mentioned in the issue.

Another small fix is adding an interpretable error to dipy.align.reslice.reslice if the input has a dimension other than 3 or 4.


skoudoro (commented):
Thank you for this @pjsjongsung.

As usual, unitests are more than welcomed :-)!

pjsjongsung:
This should be good to go. I also changed a few lines in the tutorial which was skipping some lines hence the long tutorial running time mentioned in #2563

pjsjongsung:
Just a reminder to myself that I should revert the change on the affine dimension and make the change other way around according to the aforementioned issue.

Garyfallidis:
PR needs a rebase @pjsjongsung .

pjsjongsung:
Working on the rebase and some additional changes. I think I messed up bad here with the pull and everything. Trying to fix it right now.

pjsjongsung:
@skoudoro While working on it, I realized that bvecs are (N, 3) instead of (3, N). So wouldn't it make more sense for reorient_bvecs function to also have (N, 4, 4) as affine input rather than (4, 4, N)? For reference, it is currently (N, 4, 4) but was mentioned in #2777 maybe it should be changed to (4, 4, N).

skoudoro:
@arokem, Can we have your opinion on that? 

it might be an issue in the whole codebase. We do not have the same rules which can be confusing.

Thank you in advance for your feedback

skoudoro:
> @skoudoro While working on it, I realized that bvecs are (N, 3) instead of (3, N). So wouldn't it make more sense for reorient_bvecs function to also have (N, 4, 4) as affine input rather than (4, 4, N)? For reference, it is currently (N, 4, 4) but was mentioned in https://github.com/dipy/dipy/issues/2777 maybe it should be changed to (4, 4, N).

> @arokem, Can we have your opinion on that?

it might be an issue in the whole codebase. We do not have the same rules which can be confusing.

Thank you in advance for your feedback

@arokem, do you have an opinion on this ? I think we should agree on a standard.

pjsjongsung:
@arokem Thank you for the reply! The problem was that some function use affine as (4, 4, N) and some as (N, 4, 4). As you said we don't have to match the dimension of N for bvecs and affines. I can change the code so that every function returns and expects (4, 4, N) as the dimension of affine.

arokem:
That would be great. Thanks!

skoudoro:
Hi @pjsjongsung,

will you find time to finalize this PR before the end of this month ?

pjsjongsung:
Oh I didn't realize it was getting errors. My bad. I will get it done soon, at least by this month.

pjsjongsung:
I think the code is in working state now! (Hopefully) @skoudoro

dipy/core/gradients.py
```diff
@@ -667,11 +667,14 @@ def reorient_bvecs(gtab, affines, atol=1e-2):
     """
     new_bvecs = gtab.bvecs[~gtab.b0s_mask]
```
skoudoro, line 668: we might need to add a warning if we detect that the shape is `(n, 4, 4) or (n, 3, 3)`. For a long time, it was like that, this change can break some code. So better to warn them

skoudoro:
Not sure why the CI's are failing here, the code looks good. Can you rebase this PR, address my comment above and it should be ready to go !

dipy/core/gradients.py
pjsjongsung: Should I correct the dimension if the shape is wrong and throw a warning that it will be not compatible in future releases?

skoudoro: Yes !

pjsjongsung:
Should be good to go @skoudoro !

skoudoro (commented):
Thank you for the update @pjsjongsung.

Can you update the message 
```
r
eorient_bvecs has changed to require affines as (4, 4, n) or (3, 3, n). Shape of (n, 4, 4) or (n, 3, 3) will be deprecated in the future.
```
by 
```
reorient_bvecs has changed to require affines as (4, 4, n) or (3, 3, n). Shape of (n, 4, 4) or (n, 3, 3) is deprecated and will fail in the future.
```

Can you also add this in [api_changes](https://github.com/dipy/dipy/blob/master/doc/api_changes.rst) ? 

Thank you

skoudoro:
Also, can you add a small test in `test_gradients` using `npt.assert_warns(...)`

pjsjongsung:
Done! @skoudoro

skoudoro (approved):
Thank you for this update @pjsjongsung, looks good!

merging

PR #2795, opened 2023-05-01
bump dependencies mininal version

skoudoro:
Small PR to bump our minimal version. 

numpy 1.14.5 -> 1.22.4
scipy 1.1 -> 1.8.1
nibabel 3.0.0 -> 4.0.0
h5py 2.8.0 -> 3.7.0
fury 0.8 -> 0.9.0

python_requires 3.6 -> 3.8

Note: This will give us access to typing module which we should start to do.


skoudoro:
Good catch, thank you! I just updated that

PR #2800, opened 2023-05-03
Remove the Deprecating nisext 

### LINKED ISSUE #2706: FYI: Deprecating nisext in nibabel
The nisext tools are obsolete and unmaintained, so we are planning on deprecating them. AFAIK the only place they are still used is the dipy Makefile:

https://github.com/dipy/dipy/blob/089a02ad0520895bd30370591f965c685293e480/Makefile#L80-L97

Do you all still use these, or are they cruft? We can probably help with alternative implementations using modern build tools, if needed.

This is not urgent; we will be raising a `DeprecationWarning` in the next release, but the tools will remain for another year at least (until 6.0).

skoudoro:
Nibabel has deprecated `nisext`. We were almost not using those tools. After checking, only `check-files` was used `sometimes`. Also, those tools depends on `distutils` that we plan to remove completely. 

This is why this PR remove completely `nisext` since this is a package that we will not use in a close future. it also allow `Nibabel` to move forward. 

Thanks for all the good services `nisext`!

closes #2706

### DIFF
--- a/version_helpers.py
+++ b/version_helpers.py
@@ -1,7 +1,5 @@
 """ Distutils / setuptools helpers for versioning
 
-This code started life in the nibabel package as nisexts/sexts.py
-
 Code transferred by Matthew Brett, who holds copyright.
 
 This version under the standard dipy BSD license.
--- a/Makefile
+++ b/Makefile
@@ -77,25 +77,6 @@ distclean: clean
 %.html : %.pyx
 	cython -a $<
 
-# Check for files not installed
-check-files:
-	$(PYTHON) -c 'from nisext.testers import check_files; check_files("dipy")'
-
-# Print out info for possible install methods
-check-version-info:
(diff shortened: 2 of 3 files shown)


version_helpers.py
```diff
@@ -1,7 +1,5 @@
 """ Distutils / setuptools helpers for versioning
 
-This code started life in the nibabel package as nisexts/sexts.py
```
arokem, line 3: I'd suggest leaving this comment in, maybe adding a comment that it has been deprecated upstream in nibabel.

arokem (commented):
LGTM. One small and not crucial comment, but otherwise, I think you can merge this when it comes back green from the CI.

version_helpers.py
skoudoro: This file will be removed before the end of the summer (september).  This is a way to start the process 😄

Issue #2805, opened 2023-05-05
Incorrect initial direction for the backward segment of local tracking


gabknight:
The initial direction set for the backward segment in `localtracking` in wrongly updated after 045a1ee.

Removing
```
 for i in range(3):
        dir[i] = first_step[i]
        vs[i] = voxel_size[i]
        seed[i] = seed_pos[i]
```
and passing directly vectors allowed for direction getters to modify those variables. `voxel_size` and `seed_pos` are not modify, but `first_step` is currently overwritten in the cython code of `dipy.tracking.direction_getter.generate_streamline` line 120, which in turn changes the `first_step` input to reconstruct the backward segment of the streamlines from the seed position. This modifies the set of initial propagation orientations and allows the tracking to go above the angular threshold at the seed location. After the initial backward step, the tracking proceed normally. This affects all local tracking algorithms since version 1.5. PFT is unaffected.

gabknight:
fixed with PR #2806

PR #2806, opened 2023-05-05
BF - initial backward orientation of local tracking

gabknight:
This address issue #2805.

- Reverts changes in `dipy/tracking/localtrack.pyx` from https://github.com/dipy/dipy/commit/045a1ee5485af33675726675a010c3a6fda5d9b3
- fix a minor infrequent bug in probabilistic tracking where the forward and backward direction could be sample in such a way it is bigger that `max_angle`. This was fixed by changing the initial direction to the true first direction selected in the forward segment, instead of the tentative initial direction obtained from the peaks.

### DIFF
--- a/dipy/direction/probabilistic_direction_getter.pyx
+++ b/dipy/direction/probabilistic_direction_getter.pyx
@@ -178,7 +178,6 @@ cdef class DeterministicMaximumDirectionGetter(ProbabilisticDirectionGetter):
                 (direction[0], direction[1], direction[2])]
         except KeyError:
             # if direction is not a sphere vertex, we find the closest vertex
-            print("oh no!")
             for i in range(3):
                 vertex[i] = direction[i]
             idx = self.sphere.find_closest(vertex)
--- a/dipy/direction/probabilistic_direction_getter.pyx
+++ b/dipy/direction/probabilistic_direction_getter.pyx
@@ -93,14 +93,23 @@ cdef class ProbabilisticDirectionGetter(PmfGenDirectionGetter):
         cdef:
             cnp.npy_intp i, idx, _len
             double[:] newdir, pmf
+            double vertex[3]
             double last_cdf, random_sample
             cnp.uint8_t[:] bool_array
 
         pmf = self._get_pmf(point)
         _len = pmf.shape[0]
-
-        bool_array = self._adj_matrix[
-            (direction[0], direction[1], direction[2])]
+        try:
+            bool_array = self._adj_matrix[
+                (direction[0], direction[1], direction[2])]
+        except KeyError:
+            # if direction is not a sphere vertex, we find the closest vertex
+            for i in range(3):
+                vertex[i] = direction[i]
+            idx = self.sphere.find_closest(vertex)
+            bool_array = self._adj_matrix[(self.vertices[idx, :][0],
+                                           self.vertices[idx, :][1],
+                                           self.vertices[idx, :][2])]
 
         for i in range(_len):
             if bool_array[i] == 0:
@@ -157,14 +166,25 @@ cdef class DeterministicMaximumDirectionGetter(ProbabilisticDirectionGetter):
         cdef:
             cnp.npy_intp _len, max_idx
             double[:] newdir, pmf
+            double vertex[3]
             double max_value
             cnp.uint8_t[:] bool_array
 
         pmf = self._get_pmf(point)
         _len = pmf.shape[0]
 
-        bool_array = self._adj_matrix[
-            (direction[0], direction[1], direction[2])]
+        try:
+            bool_array = self._adj_matrix[
+                (direction[0], direction[1], direction[2])]
+        except KeyError:
+            # if direction is not a sphere vertex, we find the closest vertex
+            print("oh no!")
+            for i in range(3):
+                vertex[i] = direction[i]
+            idx = self.sphere.find_closest(vertex)
+            bool_array = self._adj_matrix[(self.vertices[idx, :][0],
+                                           self.vertices[idx, :][1],
+                                           self.vertices[idx, :][2])]
 
         max_idx = 0
         max_value = 0.0
--- a/dipy/direction/probabilistic_direction_getter.pyx
+++ b/dipy/direction/probabilistic_direction_getter.pyx
@@ -6,12 +6,14 @@
 Implementation of a probabilistic direction getter based on sampling from
 discrete distribution (pmf) at each step of the tracking.
 """
+from random import random
+
 import numpy as np
 cimport numpy as cnp
 
 from dipy.direction.closest_peak_direction_getter cimport PmfGenDirectionGetter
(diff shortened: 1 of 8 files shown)


skoudoro:
Can you create a test to make sure we detect it next time ?

gabknight:
I added the tests, which identify a minor bug, again at the junction of the forward and backward streamline segments for probabilistic tracking. For the backward segment, I changed the tentative initial direction (peaks) to the true initial orientation (vector connecting the first and second points of the forward segment.). Since orientations are always sphere vertices for `PmfGenDirectionGetter`, I needed calling `sphere.find_closest()` with the update initial direction, when querying the dictionary of neighboring orientations. I did it through a try/catch to remove the check at each iteration, what do you think?

dipy/direction/probabilistic_direction_getter.pyx
```diff
@@ -93,14 +93,23 @@ cdef class ProbabilisticDirectionGetter(PmfGenDirectionGetter):
...
+            bool_array = self._adj_matrix[
+                (direction[0], direction[1], direction[2])]
+        except KeyError:
+            # if direction is not a sphere vertex, we find the closest vertex
+            for i in range(3):
+                vertex[i] = direction[i]
+            idx = self.sphere.find_closest(vertex)
+            bool_array = self._adj_matrix[(self.vertices[idx, :][0],
+                                           self.vertices[idx, :][1],
+                                           self.vertices[idx, :][2])]
```
skoudoro, line 112: Not sure about this.  `try except` python clause inside a `cdef` functions. 

You might loose all the cedf benefit.  I need to look deeper at this.

dipy/direction/probabilistic_direction_getter.pyx
```diff
@@ -157,14 +166,24 @@ cdef class DeterministicMaximumDirectionGetter(ProbabilisticDirectionGetter):
...
```
skoudoro, line 186: same as above. you might need to find an alternative.

dipy/tracking/local_tracking.py
```diff
@@ -143,7 +143,13 @@ def _generate_tractogram(self):
                         stream_status == StreamlineStatus.ENDPOINT or
                         stream_status == StreamlineStatus.OUTSIDEIMAGE):
                     continue
-                first_step = -first_step
+                if stepsF > 1:
+                    # Use the opposite of the first selected orientation for
+                    # the backward tracking segment
+                    first_step = F[0] - F[1]
+                    first_step = first_step / np.linalg.norm(first_step)
```
skoudoro, line 150: let's try to avoid a zerodivision error via a small check

dipy/tracking/localtrack.pyx
```diff
@@ -56,14 +56,21 @@ def local_tracker(
...
+        double seed[3]
 
     if (seed_pos.shape[0] != 3 or first_step.shape[0] != 3 or
             voxel_size.shape[0] != 3 or streamline.shape[1] != 3):
         raise ValueError('Invalid input parameter dimensions.')
 
+    for i in range(3):
+        dir[i] = first_step[i]
+        vs[i] = voxel_size[i]
+        seed[i] = seed_pos[i]
```
skoudoro, line 70: Why this copy is needed? I do not get it .

We might need to rename `dir` since this is a python keyword

skoudoro (commented):
Hi @gabknight,

Thank you for fixing this bug. See below my comments.  I will look for an alternative to `try/except` clause

dipy/tracking/localtrack.pyx
skoudoro: if copy is needed, I will use some like:

```python
# Memory copy 
memcpy(dir, &first_step[0], sizeof(double) * 3)
```

you also have the `copy_point` function in `dipy.utils.fast_numpy`

`copy_point(&first_step[0], dir)`

dipy/direction/probabilistic_direction_getter.pyx
skoudoro: I just realized that `self._adj_matrix` is a dict, so a python object.

This needs to change. otherwise the `cdef` is quite useless.

dipy/tracking/localtrack.pyx
gabknight: Yes, thanks, I will update the code. 

We need copies because `generate_streamline(.)` can modify them and the input values are further needed in `_generate_tractogram(.)`.  The bug here was that `dir` was modified with the current tracking orientation in `generate_streamline(.)`, and later used as input for the computation of the backward streamline segment in `generate_tractogram(.)`. It was supposed to be the initial orientation, but instead was the last tracking orientation. Making copies solves the issue.

dipy/tracking/local_tracking.py
gabknight: thanks, fixed.

dipy/direction/probabilistic_direction_getter.pyx
gabknight: Replacing the `dict` with a cythonized search for neighbors vertices turn out as fast for probabilistic tracking and faster for deterministic maximum tracking. The code is shorter and more clear this way. Further optimization will be done in another PR.

skoudoro: That's great! Thank you @gabknight. waiting for the CI's to finish and I will go ahead to merge this

dipy/direction/probabilistic_direction_getter.pyx
```diff
@@ -93,40 +82,49 @@ cdef class ProbabilisticDirectionGetter(PmfGenDirectionGetter):
...
-
-            direction[0] = newdir[0]
-            direction[1] = newdir[1]
-            direction[2] = newdir[2]
-        else:
-            direction[0] = -newdir[0]
-            direction[1] = -newdir[1]
-            direction[2] = -newdir[2]
+        with nogil:
+            # TO DO call fast numpy fcts.
```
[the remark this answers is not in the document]
skoudoro: I believe a cast should do the job:   `my_func(<double[:3]>direction)`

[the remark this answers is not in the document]
gabknight: yes, thanks!

skoudoro:
Thank you for the update @gabknight! 

it seems there is still some issues with Windows as you can see below
```
FAILED tracking/tests/test_tracking.py::test_tracking_max_angle - AssertionError
FAILED tracking/tests/test_tracking.py::test_particle_filtering_tractography - AssertionError: 
Arrays are not almost equal to 7 decimals
 ACTUAL: 0.0
 DESIRED: 0.2
```

I suppose `srand` have different result between windows and linux. you might need to find an alternative for that.

gabknight:
I created issue #2867 to address the random fct call problem in another PR.

skoudoro (approved):
Thank you @gabknight, 

Looks good! Thanks for this work. merging

PR #2808, opened 2023-05-08
[DOC] Fix cross referencing

skoudoro:
This PR updates all cross referencing which were broken due to sphinx-gallery. 

Before, you could reference an example by doing `:ref:example_{filename}`

Now, you need to reference an example by doing`:ref:sphx_glr_examples_built_{section_name}_{filename}.py`

Also, I force the inclusion of `link_name.inc` in all tutorials. some link were not working due to this.


skoudoro:
All green and ready for review or to be merged !

PR #2810, opened 2023-05-11
[DOC] introducing sphinxcontrib-Bibtex to improve reference management

skoudoro:
> build succeeded, 1871 warnings.

A lot of those warning are due to references. This PR add `sphinxcontrib-bibtex` to manage references, avoid duplication, avoid formatting error, etc. 

This PR will take a bit of time to be finalize so I wonder if I should split it in multiple PR or not.  It should make review process easier. Looking forward for your feedback.

### DIFF
--- a/dipy/reconst/dki.py
+++ b/dipy/reconst/dki.py
@@ -423,8 +423,10 @@ def _F2m(a, b, c):
 
 
 def directional_diffusion(dt, V, min_diffusivity=0):
-    r"""Calculate the apparent diffusion coefficient (ADC) in each direction
-    of a sphere for a single voxel [1]_
+    r"""Compute apparent diffusion coefficient (adc).
+
+    Calculate the apparent diffusion coefficient (adc) in each direction of
+    a sphere for a single voxel :footcite:t:`NetoHenriques2015`.
 
     Parameters
     ----------
@@ -446,18 +448,8 @@ def directional_diffusion(dt, V, min_diffusivity=0):
 
     References
     ----------
-    .. [1] Jensen JH, Helpern JA, Ramani A, Lu H, Kaczynski K, (2005).
-           Diffusional kurtosis imaging: The quantification of non-gaussian
-           water diffusion by means of magnetic resonance imaging. Magnetic
-           Resonance in Medicine 53(6): 1432-1440
-    .. [2] Henriques RN, Correia MM, Marrale M, Huber E, Kruper J, Koudoro S,
-           Yeatman JD, Garyfallidis E, Rokem A (2021). Diffusional Kurtosis
-           Imaging in the Diffusion Imaging in Python Project. Frontiers in
-           Human Neuroscience 15: 675433.
-    .. [3] Neto Henriques R, Correia MM, Nunes RG, Ferreira HA (2015).
-           Exploring the 3D geometry of the diffusion kurtosis tensor -
-           Impact on the development of robust tractography procedures and
-           novel biomarkers, NeuroImage 111: 85-99
+    .. footbibliography::
+
     """
     adc = (
         V[:, 0] * V[:, 0] * dt[0]
--- a/doc/references.bib
+++ b/doc/references.bib
@@ -114,7 +114,7 @@
   number    = {3},
   pages     = {209-219},
   year      = {1996},
-  url       = {10.1006/jmrb.1996.0086},
+  doi       = {10.1006/jmrb.1996.0086},
   url       = {https://doi.org/10.1006/jmrb.1996.0086}
 }
 
@@ -346,7 +346,7 @@
   number    = {5},
   pages     = {1051-1058},
   year      = {2007},
-  url       = {10.1002/jmri.20905},
+  doi       = {10.1002/jmri.20905},
   url       = {https://doi.org/10.1002/jmri.20905}
 }
 
@@ -706,7 +706,7 @@
 }
 
 @article{Hansen2016b,
-  author    = {Brian Hansen and Noam Shemesh and Sune N\o{}rh\o{}j Jespersen}
+  author    = {Brian Hansen and Noam Shemesh and Sune N\o{}rh\o{}j Jespersen},
   title     = {{Fast imaging of mean, axial and radial diffusion kurtosis}},
   journal   = {NeuroImage},
   volume    = {142},
@@ -776,7 +776,7 @@
 }
 
 @article{Hui2008,
-  author    = {Edward S. Hui and Matthew M. Cheung and Liqun Qi and Ed X. Wu}
+  author    = {Edward S. Hui and Matthew M. Cheung and Liqun Qi and Ed X. Wu},
   title     = {{Towards better MR characterization of neural tissues using directional diffusion kurtosis analysis}},
   journal   = {NeuroImage},
   volume    = {42},
@@ -796,7 +796,7 @@
   pages     = {1432-1440},
   year      = {2005},
   doi       = {https://doi.org/10.1002/mrm.20508},
-  doi       = {https://onlinelibrary.wiley.com/doi/abs/10.1002/mrm.20508}
+  url       = {https://onlinelibrary.wiley.com/doi/abs/10.1002/mrm.20508}
 }
 
 @article{Jeurissen2014,
@@ -821,6 +821,18 @@
   year      = {2011},
   doi       = {10.1002/hbm.21032},
   url       = {https://doi.org/10.1002/hbm.21032}
+}
+
+@article{Jones1999,
+  author    = {Derek K. Jones and Mark A. Horsfield and Andrew Simmons},
+  title     = {{Optimal strategies for measuring diffusion in anisotropic systems by magnetic resonance imaging}},
+  journal   = {Magnetic Resonance in Medicine},
+  volume    = {42},
+  number    = {3},
+  pages     = {515--25},
+  year      = {1999},
+  doi       = {10.1002/(SICI)1522-2594(199909)42:3<515::AID-MRM14>3.0.CO;2-Q},
+  url       = {http://www.ncbi.nlm.nih.gov/pubmed/10467296}
 }
 
 @article{Jones2013,
@@ -910,7 +922,7 @@
 }
 
 @article{Koay2009a,
-  author    = {Cheng Guan Koay and Evren {\"O}zarslan and Peter J. Basser}
+  author    = {Cheng Guan Koay and Evren {\"O}zarslan and Peter J. Basser},
   title     = {{A signal transformational framework for breaking the noise floor and its applications in MRI}},
   journal   = {Journal of Magnetic Resonance},
   volume    = {197},
@@ -1002,7 +1014,7 @@
   journal   = {IEEE Transactions on Medical Imaging},
   volume    = {22},
   number    = {1},
-  year      = {120-128},
+  pages     = {120-128},
   year      = {2003},
   doi       = {10.1109/TMI.2003.809072}
 }
@@ -1059,7 +1071,7 @@
   journal   = {Magnetic Resonance in Medicine},
   volume    = {81},
   number    = {5},
-  year      = {2019}
+  year      = {2019},
   pages     = {3245-3261},
   doi       = {10.1002/mrm.27606},
   url       = {https://doi.org/10.1002/mrm.27606}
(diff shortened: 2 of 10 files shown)


arokem:
Splitting it up makes sense to me. There's no reason to wait until it's all done to merge these additions, right?

skoudoro:
> There's no reason to wait until it's all done to merge these additions, right?

right, so you can merge, and I will do that step by step starting with the examples and then the docstring in the codebase

dipy/reconst/dki.py
```diff
@@ -380,8 +380,10 @@ def _F2m(a, b, c):
 
 
 def directional_diffusion(dt, V, min_diffusivity=0):
-    r""" Calculate the apparent diffusion coefficient (adc) in each direction of
-    a sphere for a single voxel [1]_
+    r"""Compute apparent diffusion coefficient (adc).
+
+    Calculate the apparent diffusion coefficient (adc) in each direction of
+    a sphere for a single voxel :footcite:t:`NETOHENRIQUES201585`.
```
arokem, line 386: This is a bit of an eyesore. Maybe we could we use a uniform format for bibtex keys? For example, "NetoHenriques2015NeuroImage"

arokem (commented):
Looks great! What do you think about normalizing the bibtex keys?

dipy/reconst/dki.py
skoudoro: > This is a bit of an eyesore

ahahah, you are right

> Maybe we could we use a uniform format for bibtex keys?

I agree, even if the bibtex file is big. I like the beginning of your example but I do not think we should finish with the journal. so just {firstname}{name}{year}{letter | optional if multiple paper} is that ok ?

arokem: Yeah. That would work well.

doc/diffusion.bib
```diff
@@ -319,14 +319,14 @@ @Article{Carlsson2009
 }
 
 @Article{Avram2008NMRBiomed,
-  Author         = {Avram, Liat and \�{O} zarslan, Evren and Assaf, Yaniv
+  Author         = {Avram, Liat and \�{O} zarslan, Evren and Assaf, Yaniv
```
jhlegarreta, line 322: This symbol and other similar ones introduced are not plain, BIbTeX symbols; e.g. the correct BibTeX spelling for Evren's surname would be `\"{O}zarslan`. I can have a go to the file to fix them. I assume the changes were done inadvertently introduced when running some tool.

doc/diffusion.bib
```diff
@@ -319,14 +319,14 @@ @Article{Carlsson2009
...
                    and Bar-Shir, Amnon and Cohen, Yoram and Basser, Peter
                    J.},
   Title          = {Three-dimensional water diffusion in impermeable
                    cylindrical tubes: theory versus experiments},
   Journal        = {NMR IN BIOMEDICINE},
   Volume         = {21},
-  Pages          = {888�898},
+  Pages          = {888�898},
```
jhlegarreta, line 329: Similar to the above, but in this case the hyphen does exist in BibTeX; usually BibTeX is smart enough to change it to the en dash for page ranges.

jhlegarreta (changes requested):
Thanks for doing this @skoudoro.

Not sure if `sphinxcontrib-bibtex` is actively maintained, but I do agree that this or other similar mechanism is required.

Some comments in-line.

jhlegarreta:
I agree to the conversation in https://github.com/dipy/dipy/pull/2810#discussion_r1191500875. I am sure there is also a way to generate the keys automatically using a self-defined pattern, but we may leave that as a separate task/PR.

Also, making all entries consistent (fields, names, etc.) would ease reading the file (e.g. when maintainers need to go through it). Something that might be partially automated? Probably best as a separate PR.

doc/diffusion.bib
jhlegarreta: I can have a look at this and amend the commit.

jhlegarreta:
Also, If BibTeX references will be/are split across modules (e.g. segment, tracking, docs, etc.), it may be harder to maintain, as the same reference may fall into different categories, duplicated entries may exist (unless the automated tool prevents them, but then it may not be evident where a reference should belong to).

jhlegarreta:
I had forgotten that we had started this. Thanks for reviving it :+1:. I had done a partial local test during the last few days, and the tool seems to be working as expected. 

Re: https://github.com/dipy/dipy/pull/2810#pullrequestreview-1432731863
> Not sure if sphinxcontrib-bibtex is actively maintained, but I do agree that this or other similar mechanism is required.

There has been a release recently, so it looks like it is.

Other thoughts:
- The current bibtex is long and hard to read (layout, contains the abstracts, and lots of other additional fields, etc.). It was generated by some bibliography tool, and unless we keep the underlying file/collection (whatever that is) before the exportation in a shared repository, it will be hard to keep up with that. And even in that case, whenever a ref needs to be changed or added, we would depend on a separate repository. So manually adding refs does not look that hard (the time-consuming part is the current transition).
- It is also hard to locate references in the current file as there does not seem to be any sorting; having a type (article, inproceedings) and alphabetical (key -author surname, year) sorting would help a lot IMO.
- So I would clearly remove the current references.
- I have gathered a significant amount of references (close to 150) as I've been going through the doc warnings and best would be to add them as an additional commit to the bibtex file. Having them would save time and get us closer to fixing the doc issues. In the same commit, I can then remove the existing refs in the bibtex file (many of the refs do not appear at all across the code base, some are incomplete, and as noted, they have other problems).

Let me know what you think about adding the refs I've gathered @skoudoro.

Re https://github.com/dipy/dipy/pull/2810#issuecomment-1543985963

> I will do that step by step starting with the examples and then the docstring in the codebase

- The quickest ones to fix (as there are very few files and refs) would be the ones in the `theory` doc module, and maybe at that point the PR can be already merged, to then address the examples, as you say, then other modules as time permits. I can event push that as I tested on these locally.

A few other things to note/test:
  - Use of `.. bibliography::` vs `.. footbibliography::` (and `:cite:t:`, `:cite:p:` vs the `footcite` counterparts).
  - I've read that there needs to be a single directive across the entire code base (not sure/not tested).
  - We also have footnotes (different from notes) that `sphinxcontrib-bibtex` can manage.
  - Cross-refs in figure captions may need to be tested.


I have another batch of doc warning fixes that I will push in the coming days so that we can further decrease the number and be closer to having the citation-related ones only (but I also see in the API that blocks that do not contain warnings have at times formatting problems, especially when math blocks are involved, but that will be for another round of PRs).

skoudoro:
> I had forgotten that we had started this. Thanks for reviving it 👍. I had done a partial local test during the last few days, and the tool seems to be working as expected.

thanks 👍 

> There has been a release recently, so it looks like it is.

yep, it is and it seems it does not need a lot of update

> The current bibtex is long and hard to read.......It is also hard to locate references in the current file.......So I would clearly remove the current references.

I agree with this statement. For now, do you want me to just rename it and then, when we have the new file, we can remove it?

> I have gathered a significant amount of references (close to 150) 

Please, create a PR, after merging your new references, I can rebase this one and start the proper way.

> Use of .. bibliography::

Use of `.. bibliography::` needs to be use only one time. That why we should use `.. footbibliography::` in priority. I need to read again the documentation, it was a long time again. the `mne-python` project use it the same way if you have time to have a look

> I've read that there needs to be a single directive across the entire code base (not sure/not tested).

yes, and that is the `.. bibliography::`.   see my comment just above.

> We also have footnotes (different from notes) that sphinxcontrib-bibtex can manage.

another steps after fixing bibliography

> I have another batch of doc warning fixes that I will push in the coming days so that we can further decrease the number and be closer to having the citation-related ones only

ok 👍 



Thank you for looking into this @jhlegarreta

jhlegarreta:
@skoudoro OK, thanks for the feedback. See PR #3308 for the bibliography file:
- There may be a few references missing, but tried to go over all files to gather all references.
- @arokem @skoudoro Simplified the key convention to `lastNameYear(Letter)`. IMO it should be enough.

jhlegarreta:
Did not dare to touch the `doc/diffusion.bib` Serge, but I guess we would be good rebasing on `master`, and fixing the conflicts by dropping the changes in here.

I will slowly work on the transition of the references in the core modules in a separate PR.

skoudoro:
Hi @jhlegarreta,

I did the rebase. However, `sphinxcontrib-bibtex` is failing due to many unknown entries.  (like `@electronics`, `@dataset`)

Indeed, they do not exist when I look at this page: https://www.bibtex.com/e/entry-types/.

Can you tell me where they come from and which tool is recognizing these types of entries? Do you want to create a PR to simplify that or should I go ahead here?

Thank you for your feedback

jhlegarreta:
Sorry for the headaches this might have caused @skoudoro: PR #3312.

doc/references.bib
```diff
@@ -823,6 +823,18 @@ @article{Jeurissen2011
   url       = {https://doi.org/10.1002/hbm.21032}
 }
 
+@article{Jones1999,
+  author    = {Jones, D K and Horsfield, M a and Simmons, a},
+  title     = {{Optimal strategies for measuring diffusion in anisotropic systems by magnetic resonance imaging.}},
```
jhlegarreta, line 828: {{Optimal strategies for measuring diffusion in anisotropic systems by magnetic resonance imaging}},
```suggestion
  title     = {{Optimal strategies for measuring diffusion in anisotropic systems by magnetic resonance imaging}},
```

doc/references.bib
```diff
@@ -823,6 +823,18 @@ @article{Jeurissen2011
...
+  journal   = {Magnetic resonance in medicine : official journal of the Society of Magnetic Resonance in Medicine / Society of Magnetic Resonance in Medicine},
```
jhlegarreta, line 829: Magnetic Resonance in Medicine
```suggestion
  journal   = {Magnetic Resonance in Medicine},
```

doc/references.bib
```diff
@@ -823,6 +823,18 @@ @article{Jeurissen2011
   url       = {https://doi.org/10.1002/hbm.21032}
 }
 
+@article{Jones1999,
+  author    = {Jones, D K and Horsfield, M a and Simmons, a},
```
jhlegarreta, line 827:
```suggestion
  author    = {Derek K. Jones and Mark A. Horsfield and Andrew Simmons},
```

jhlegarreta (commented):
Sorry, I missed this one.

skoudoro:
> Sorry for the headaches this might have caused @skoudoro: PR https://github.com/dipy/dipy/pull/3312.

I have to admit, not easy to find the issue 😅 

all back to normal, I think this PR is ready to go

jhlegarreta (commented):
:+1: thanks for doing this: the number of warnings has decreased in the documentation CI builds.

I would have preferred to see this
https://github.com/dipy/dipy/pull/2810/files#diff-c1e15eead4a02baaf8049c7bc83badef5892399b556e74a6e35a84c5926b0a33L343-L344

as a separate commit. If you are not willing to change that, and if your local tests have shown that we are safe to make the transition across the code base, let me know and I will merge as is.

Edit: thanks for having fixed a number of oversights of mine in `references.bib`.

skoudoro:
> I would have preferred to see this

Sorry for that... if you have a quick instruction line to change that, I can do it, otherwise, I am not willing to spend time on this. Sorry. 

>  if your local tests have shown that we are safe to make the transition across the code base, let me know

Yes, it is safe. you can go ahead ! see some screenshot below

image
image
image

Now, we need to update all our codebase (docstring, rst and examples). Let's start with docstring and/or to avoid conflicting with #3306.

jhlegarreta:
> Sorry for that... if you have a quick instruction line to change that, I can do it, otherwise, I am not willing to spend time on this. Sorry.

:+1: Sounds good.

> Yes, it is safe. you can go ahead ! see some screenshot below

Looks :100:. Merging.

jhlegarreta (approved):

doc/diffusion.bib
jhlegarreta: Removed existing entries were removed in PR #3308. Resolving this.

doc/diffusion.bib
jhlegarreta: Removed existing entries were removed in PR #3308. Resolving this.

PR #2816, opened 2023-05-23
Correlation Tensor Imaging

shilpiprd:
CTI  is an advanced diffusion MRI technique that focuses on capturing the correlation between diffusion signals in different directions.
CTI provides improved accuracy in characterizing tissue microstructure, enhanced detection of crossing fibers, and better mapping of complex white matter connections.
CTI specifically addresses the challenges associated with crossing fibers, which can confound interpretation and analysis. 
CTI provides more in-depth information about the microscopic anisotropy of cells, whether they are aligned or crossing.


skoudoro:
Thanks! @lb-97 and @RafaelNH, Can you look at this?

Also, I would recommend to make sure that the tests are running locally before pushing and creating a pull request. if you have any questions concerning this point, do not hesitate

shilpiprd:
Actually the cti.py file isn't complete at all. So I can't really create tests for any of the methods yet because there's a little uncertainity in each of the method's code.

RafaelNH:
@skoudoro - Indeed! I asked Shilpi to start this PR, so that everyone can see the code progress of her GSoC project in "real time". The idea is to include tests at the same time as the code is being implemented. Keep tuned!

dipy/reconst/tests/test_cti.py
```diff
@@ -0,0 +1,30 @@
...
+bvecs1 = np.append(bvecs1 , [[0, 0, 0]], axis=0)
+bvals1 = np.array([2] * 20 + [1] * 20 + [1] * 20 + [1] * 20 + [0])
+#deifining DDE protocol for CTI
+btens = np.array(['LTE' for i in range(20)] + ['PTE' for i in range(20)] + ['STE' for i in range(20)] + ['CTE' for i in range(21)])
+
+#on providing btens, (bvals1,bvecs1) is ignored.
+gtab = gradient_table(bvals1, bvecs1, btens = btens) #gradients (81,3) will be defined, earlier on, not sure
+#we've isotropic and anisotropic diffusion tensor distribution (DTD)
+ 
+angles = [(0, 0), (90, 0)] #both tesnsors are perpendicular to each other: to simulate the presence of crossing fibers
```
RafaelNH, line 27: To make things easier, for the CTI tests, I will suggest to generate the signals exactly from the QTI model, rather than for the multi_voxel simulations. To see how this is done give a look to the lines 291 and 297 in test_qti.py. Important aspects that you need to understand: 1) what is the variable DTD; 2) what qti.dtd_covariance(DTD) does; 3) what qti.from_3x3_to_6x1 does; 3) what qti.from_6x&_to_21x1 does; and 4) what qti.qti_signal does. Any questions please let me know.

RafaelNH (commented):
Hi @shilpiprd! Here are my first comments to the CTI tests. Let me know if something is not clear.

dipy/reconst/tests/test_cti.py
shilpiprd: You mean multi_tensor simulation right ? "Rather than for the multi_tensor simulations. "

dipy/reconst/utils.py
```diff
@@ -49,6 +49,74 @@ def dki_design_matrix(gtab):
 
     return B
 
+def cti_design_matrix(gtab1, gtab2): 
+	r"""Construct B design matrix for CTI.
+	
+	Parameters
+	----------
+	gtab1: dipy.core.gradients.GradientTable
+		A GradientTable class instance
```
RafaelNH, line 58: Specific that this is "A GradientTable class instance for first DDE diffusion epoch"

RafaelNH (changes requested):
Here are some more comments for the CTI_design_matrix. Please don't forget to address also my previous comments.

RafaelNH (commented):
Just leaving some comments to the latest advances of CTI implementation.

RafaelNH (commented):
Here are some more comments.

dipy/reconst/utils.py
```diff
@@ -50,6 +50,102 @@ def dki_design_matrix(gtab):
...
+    b1 = gtab1.bvals
+    b2 = gtab2.bvals
+    n1 = gtab1.bvecs
+    n2 = gtab2.bvecs
+    B = np.zeros((len(b1), 43))
+
+    B[:, 0] = -b1 * (n1[:, 0]**2) - b2 * (n2[:, 0]**2)
+    B[:, 1] = -2 * b1 * n1[:, 0] * n1[:, 1] - 2 * b2 * n2[:, 0] * n2[:, 1]
+    B[:, 2] = -b1 * n1[:, 1]**2 - b2 * n2[:, 1]**2
+    B[:, 3] = -2 * b1 * n1[:, 0] * n1[:, 2] + 2 * b2 * n2[:, 0] * n2[:, 2]
```
RafaelNH, line 78: You have here a wrong sign. This code line should be:
B[:, 3] = -2 * b1 * n1[:, 0] * n1[:, 2] - 2 * b2 * n2[:, 0] * n2[:, 2]

dipy/reconst/utils.py
```diff
@@ -50,6 +50,102 @@ def dki_design_matrix(gtab):
...
+    B[:, 4] = - 2 * b1 * n1[:, 1] * n1[:, 2] + 2 * b2 * n2[:, 1] * n2[:, 2]
```
RafaelNH, line 79: You have here a wrong sign. This code line should be:
B[:, 4] = -2 * b1 * n1[:, 1] * n1[:, 2] - 2 * b2 * n2[:, 1] * n2[:, 2]

dipy/reconst/tests/test_cti.py
```diff
@@ -0,0 +1,195 @@
...
+    DTD_labels = ["anisotropic_DTD", "isotropic_DTD", "combined_DTD"]
+    for i, DTD in enumerate(DTDs):
+        D = np.mean(DTD, axis=0)
+        D_flat = np.squeeze(from_3x3_to_6x1(D)) #has shape:(6, )    #pretty useless, not needed
+        evals, evecs = decompose_tensor(D)  #evals:shape: (3, ) & evecs.shape: (3, 3)
+        C = qti.dtd_covariance(DTD)
+        C = qti.from_6x6_to_21x1(C)
+
+        #changing up cti : 
+        const = np.sqrt(2)
```
RafaelNH, line 140: After looking to the maths with @kerkelae. The correct conversion between the CTI's and QTI's covariance tensors are: 

const = np.sqrt(2)
Ccti = np.zeros((21,1))
Ccti[0] = C[0]
Ccti[1] = C[1] 
Ccti[2] = C[2]
Ccti[3] = C[3] / const 
Ccti[4] = C[4] / const 
Ccti[5] = C[5] / const 
Ccti[6] = C[6] / 2
Ccti[7] = C[7] / 2
Ccti[8] = C[8] / 2
Ccti[9] = C[9] / 2
Ccti[10] = C[10] / 2
Ccti[11] = C[11] / 2
Ccti[12] = C[12] / 2
Ccti[13] = C[13] / 2
Ccti[14] = C[14] / 2
Ccti[15] = C[15] / 2 
Ccti[16] = C[16] / 2 
Ccti[17] = C[17] / 2 
Ccti[18] = C[18] / (2 * const)
Ccti[19] = C[19] / (2 * const)
Ccti[20] = C[20] / (2 * const)

Perhaps it cool be useful to have this conversion implemented in a function.

dipy/reconst/tests/test_cti.py
```diff
@@ -0,0 +1,195 @@
...
+        K[7] = 3 * ccti[13] / (MD ** 2)
+        K[8] = 3 * ccti[12] / (MD ** 2)
+        K[9] = (ccti[5] + 2 * ccti[17]) / (MD**2)
+        K[10] = (ccti[4] + 2 * ccti[16]) / (MD**2)
+        K[11] = (ccti[3] + 2 * ccti[15]) / (MD**2)
+        K[12] = (ccti[6] + 2 * ccti[19]) / (MD**2)
+        K[13] = (ccti[10] + 2 * ccti[20]) / (MD**2)
+        K[14] = (ccti[14] + 2 * ccti[18]) / (MD**2)
+
+        cti_params = construct_cti_params(evals, evecs, K, C)
```
RafaelNH, line 183: Here you have to input the CTI's C elements, i.e. 
cti_params = construct_cti_params(evals, evecs, K, ccti)

RafaelNH (changes requested):
Hi @shilpiprd! Figured out some of the typos of the code. Please give a look. After correcting these typos. I think the next step is to test the cti fitting. I.e use the pred_signals, gtab1, and gtab2, and test if given cti_params_fitted are the same as cti_params.

dipy/reconst/tests/test_cti.py
```diff
@@ -0,0 +1,195 @@
...
+            qti.from_3x3_to_6x1(D),
+            qti.from_6x6_to_21x1(C))).T
+    data = qti.qti_signal(gtab, D, C, S0=S0)[np.newaxis, :]
+
+    # use labels that describe the signal
+    plt.plot(data.ravel(), label=DTD_labels[i])
+
+plt.legend()
+plt.show()
+
```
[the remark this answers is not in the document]
shilpiprd: Yes, actually I think from line 82 till 108 can be removed, since we have this whole thing later inside the function test_cti_prediction anyway

dipy/reconst/tests/test_cti.py
```diff
@@ -0,0 +1,421 @@
+import matplotlib.pyplot as plt
```
RafaelNH, line 1: No need to import matplotlib in tests - remove this line of code.

dipy/reconst/tests/test_cti.py
```diff
@@ -0,0 +1,421 @@
...
+        #        min_diffusivity=0):
+        inverse_design_matrix = np.linalg.pinv(design_matrix(gtab1, gtab2))
+        cti_return = ls_fit_cti(design_matrix(gtab1, gtab2), cti_pred_signals, inverse_design_matrix )
+        # OLS fitting
+        ctiM = cti.CorrelationTensorModel(gtab1, gtab2, fit_method="OLS")
+        # ctiF = ctiM.fit(cti_pred_signals)
+        # ctiF = ctiM.fit(cti_params) #this is turning out to be NoneType                     #error 
+
+        # assert_array_almost_equal(ctiF.model_params, cti_params)
+        assert_array_almost_equal(cti_return, cti_params)
```
RafaelNH, line 335: This is failing but it is not an issue of the 'ls_fit_cti' function. 

'cti_return' does not match 'cti_params' because of the eigenvalue decomposition. 

Since the total diffusion tensors of your tests are isotropic, when the tensor are decomposed inside function 'ls_fit_cti' it creates a different eigenvector due to float precision (note that away set of eigenvectors will create the same isotropic diffusion tensor if the tree eigenvalues are equal which seem to be the case). 

If you assert the evals, kurtosis tensor elements and covariance tensor elements, you will see that everything correctly matches. So to fix this issue just replace this line of code by:

    evals, evecs, kt, cvt = split_cti_params(cti_params)
    evals_return, evecs_return, kt_return, cvt_return = split_cti_params(cti_return)
    assert np.allclose(evals, evals_return), "evals do not match!"
    assert np.allclose(kt, kt_return), "K do not match!"
    assert np.allclose(cvt, cvt_return), "C do not match!"

RafaelNH (changes requested):
Hi @shilpiprd! I think I find the issue about the output of cti fit function not matching the gt. See my comments.

dipy/reconst/cti.py
```diff
@@ -0,0 +1,634 @@
...
+        #             msg = "Maximum convexity_level supported is 4."
+        #             warnings.warn(msg)
+        #             self.convexity_level = 4
+        #         self.sdp_constraints = load_sdp_constraints(
+        #             'dki', self.convexity_level)
+        #     self.sdp = PositiveDefiniteLeastSquares(22, A=self.sdp_constraints)
+
+        self.weights = fit_method in {'WLS', 'WLLS', 'UWLLS', 'CWLS'}
+        self.is_multi_method = fit_method in ['WLS', 'OLS', 'UWLLS', 'ULLS',
+                                              'WLLS', 'OLLS', 'CLS', 'CWLS']
```
RafaelNH, line 204: Same here - remove 'CLS' and 'CWLS'.

dipy/reconst/cti.py
```diff
@@ -0,0 +1,634 @@
...
+
+    def fit(self, data, mask=None): #here data is cti_params of shape : (n, 48)
```
RafaelNH, line 206: We want to fit multi voxels so add the line @multi_voxel_fit in the top of this function.

dipy/reconst/cti.py
```diff
@@ -0,0 +1,634 @@
...
+        data : array
+            The measured signal from one voxel.
+
+        mask : array
+            A boolean array used to mark the coordinates in the data that
+            should be analyzed that has the shape data.shape[-1]
+
+        """
+        data_thres = np.maximum(data, self.min_signal)
+        params = self.fit_method(self.design_matrix, data_thres,
```
RafaelNH, line 220: here the input self.inverse_design_matrix is missing.

dipy/reconst/cti.py
```diff
@@ -0,0 +1,634 @@
...
+                                 *self.args, **self.kwargs)
+        #we need to somehow obtian cti_params from data (it's actually cti_pred_signals)
+        print('this is data.shape: ',data.shape)
+        return CorrelationTensorFit(self, params) #ig there's a need to somehow define cti_params here. 
+    
+    @multi_voxel_fit
+    def multi_fit(self, data_thres, mask=None):
```
RafaelNH, line 227: Unlike the case of the DKI, we don't need to have a separate multi_fit function since we should not aim to implement the NLS fit in this first PR. Let's only focus on the OLS and WLS fits.

RafaelNH: Alternatively, you can leave the multi_fit function here, but make sure that it is being called by the fit function (defined line 206 in analogous that it is done in dki.py)

RafaelNH (changes requested):
Hi @shilpiprd ! I added some extra comments related to the reason why the fitting is not working when being called from the 'fit' function of the CorrelationTensorModel class object.

dipy/reconst/cti.py
```diff
@@ -0,0 +1,658 @@
...
+
+    @property
+    def kt(self):  # self.model_params.shape = (48, )
+        """
+        Return the 15 independent elements of the kurtosis tensor as an array
+        """
+        return self.model_params[12:27, ...]  
+
+    @property
+    def dft(self):  # created
```
RafaelNH, line 310: I think you don't need to define this property because you will be importing an analogous property from the DiffusionKurtosisFit / TensorFit objects. See property lower_triangular of the TensorFit object in dti.py. 

Also, the code of the property that you are defining here only works for single voxels.

dipy/reconst/cti.py
```diff
@@ -0,0 +1,658 @@
...
+                                diffusion_tensor[1, 1],
+                                diffusion_tensor[2, 2],
+                                diffusion_tensor[0, 1],
+                                diffusion_tensor[0, 2],
+                                diffusion_tensor[1, 2]])
+
+        return dt_elements
+    
+    @property
+    def cvt(self):  # created
```
RafaelNH, line 332: Perhaps it is better to call this property just ct.

dipy/reconst/cti.py
```diff
@@ -0,0 +1,658 @@
...
+                2) Three lines of the eigenvector matrix each containing the
+                   first, second and third coordinates of the eigenvector
+                3) Fifteen elements of the kurtosis tensor
+                4) Twenty One elements of the covariance tensor
+
+        """
+        DiffusionKurtosisFit.__init__(self, model, model_params)
+
+    @property
+    def kt(self):  # self.model_params.shape = (48, )
```
RafaelNH, line 303: I think you don't need to redefine this property here since it is already being imported from the DiffusionKurtosisFit class object. 

Also you have a typo here - the different model parameters should correspond to the last dimension of model_params and not the first. This is being correctly defined in he DiffusionKurtosisFit class object. 

return self.model_params[12:27, ...]  ==> return self.model_params[..., 12:27]

dipy/reconst/cti.py
```diff
@@ -0,0 +1,658 @@
...
+                                diffusion_tensor[1, 2]])
+
+        return dt_elements
+    
+    @property
+    def cvt(self):  # created
+        """
+        Returns the 21 independent elements of the covariance tensor as an array
+        """
+        return self.model_params[27:48, ...]
```
RafaelNH, line 336: model parameters should correspond to the last dimension of model_params and not the first. 

return self.model_params[27:48, ...] ==> return self.model_params[..., 27:48]

dipy/reconst/cti.py
```diff
@@ -0,0 +1,658 @@
...
+        # we separate the kurtosis in 3 parts: isotropic+anisotropoic+microscopic. Do we need methods for this? : REFER VIDEO ON THIS
+
+    def predict(self, gtab1, gtab2, S0=100):  # created
+        """Given a CTI model fit, predict the signal on the vertices of a gradient table
+
+        Parameters:
+        -----------
+        params: numpy.ndarray (...,43)
+                All parameters estimated from the correlation tensor model.
+                Paramters are ordered as follows::
```
RafaelNH, line 351: Paramters ==> Parameters

dipy/reconst/cti.py
```diff
@@ -0,0 +1,658 @@
...
+        """
+        return self.model_params[27:48, ...]
+
+   # def cvt(__):  # calculates the mean of all covariance parameters. required ? Formula ?
+
+        # There are 4 (radial,mean,axial,fractional) do we have something similar for covariance in qti ?
+
+        # we separate the kurtosis in 3 parts: isotropic+anisotropoic+microscopic. Do we need methods for this? : REFER VIDEO ON THIS
+
+    def predict(self, gtab1, gtab2, S0=100):  # created
```
RafaelNH, line 344: S0 default is 1

dipy/reconst/cti.py
```diff
@@ -0,0 +1,658 @@
...
+            :math:: 
+            
+            \[K_{aniso} = \frac{6}{5} \cdot \frac{\langle V_{\lambda}(D_c) \rangle}{\overline{D}^2}\]
+            
+        where: \(K_{aniso}\) is the anisotropic kurtosis, 
+            \(\langle V_{\lambda}(D_c) \rangle\) represents the mean of the variance of eigenvalues of the diffusion tensor,
+            \(\overline{D}\) is the mean of the diffusion tensor.
+        """ 
+
+        D = self.dft
```
RafaelNH, line 392: For D use the information provided in self.lower_triangular

dipy/reconst/cti.py
```diff
@@ -0,0 +1,658 @@
...
+        C = self.cvt
+        matrix = np.array([[D[0], D[3], D[4]],[D[3], D[1], D[5]],[D[4], D[5], D[2]]])
```
RafaelNH, line 394: Avoid this matrix conversion, since you can access the individual diffusion tensor elements directly from self.lower_triangular

dipy/reconst/cti.py
```diff
@@ -0,0 +1,658 @@
...
+        D  = matrix 
+        
+        Variance = 2/9 * (C[0] + D[0, 0] ** 2 + C[1] + D[1, 1] ** 2 + C[2] + D[2, 2]**2 - C[5]
```
RafaelNH, line 397: this will only work for single voxel data. For multi voxel data you have to call the tensor elements in the following way:

C[..., 0]
D[..., 0] (assuming that tensor elements are in the last dimension according to self.lower_triangular).

...
(and so on)

dipy/reconst/cti.py
```diff
@@ -0,0 +1,658 @@
...
+                 - D[0, 0] * D[1, 1] - C[4] - D[0, 0] * D[2, 2]
+                 - C[3] - D[1, 1] * D[2, 2]
+                 + 3 * (C[17] + D[0, 1] ** 2 + C[16] + D[0, 2] ** 2 + C[15] + D[1, 2] ** 2))
+
+        mean_D =  np.trace(D) / 3#trace(D) / 3
+        K_aniso = (6/5) * (Variance / (mean_D **2))
+        return K_aniso 
+    
+    @auto_attr
+    def K_iso(self):
```
RafaelNH, line 407: Adapt this function according to the comments made for K_aniso.

RafaelNH (changes requested):
Hi @shilpiprd! I did a more detailed code revision, now including some comments about documentation formatting since we are reaching the end of the GSoC. 

Apart from my comments about documentation formatting, I noticed that you haven't added tests for multi voxel simulations. Indeed, the code of the new kurtosis sources at the moment will not work for multi voxel scenarios as is generally the case of real diffusion MRI data. Please add that tests and fix the kurtosis source attributes to be general to multi voxel cases. See my in line comments for more details.

dipy/reconst/cti.py
```diff
@@ -0,0 +1,658 @@
...
+    return evals, evecs, kt, cvt
+
+
+def cti_prediction(cti_params, gtab1, gtab2, S0=100):
+    """Predict a signal given correlation tensor imaging parameters
+
+        Parameters
+        ----------
+        cti_params: numpy.ndarray (...,48)
+                All parameters estimated from the correlation tensor model.
```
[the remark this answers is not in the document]
shilpiprd: are you sure that's ok because in DKI, we've: 
dki_params : ndarray (x, y, z, 27) or (n, 27)
        All parameters estimated from the diffusion kurtosis model.
        Parameters are ordered as follows:
            1) Three diffusion tensor's eigenvalues
            2) Three lines of the eigenvector matrix each containing the first,
               second and third coordinates of the eigenvectors respectively
            3) Fifteen elements of the kurtosis tensor

dipy/reconst/cti.py
```diff
@@ -0,0 +1,658 @@
...
+    evals = cti_params[..., :3]
+    evecs = cti_params[..., 3:12].reshape(cti_params.shape[:-1] + (3, 3))
+    # kt = cti_params[12:27, ...] #original
+    kt = cti_params[..., 12:27]
+    # cvt = cti_params[27:48, ...]
+    cvt = cti_params[..., 27:48]
+    return evals, evecs, kt, cvt
+
+
+def cti_prediction(cti_params, gtab1, gtab2, S0=100):
```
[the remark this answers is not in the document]
shilpiprd: are you sure ? but didn't we decide to pass 100 as default to make cti signals match with the qti ones ?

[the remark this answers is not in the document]
RafaelNH: Yes. It is better to set the default to 1. We can manually set S0 to 100 in the tests.

doc/examples/reconst_cti.py
```diff
@@ -0,0 +1,152 @@
+"""
+========================================================================
+Reconstruction of the diffusion signal with the correlation tensor model
+========================================================================
+
+The Correlation Tensor Model (CTI) is an advanced extension of the Diffusion Kurtosis Model (see :ref:sphx_glr_examples_built_reconstruction_reconst_dki.py). [ not sure]
```
RafaelNH, line 6: Write instead: The Correlation Tensor MRI (CTI) is a method that uses double diffusion encoding data to resolve sources of kurtosis. Compared to the  Q-space Trajectory Imaging (see :ref:sphx_glr_examples_built_reconstruction_reconst_qti.py) it does not assume that tissues can be represented by a sum of multiple Gaussian compartments.

RafaelNH: Here we also have to add the reference to the paper (Henriques et al. Neuroimage 2020). See how references are being done in other DIPY tutorials.

doc/examples/reconst_cti.py
```diff
@@ -0,0 +1,152 @@
...
+In addition to the DKI, CTI introduces a more complete representation of the underlying microstructural complexity. Specifically, the CTI differentiates between different sources of kurtosis by distinguishing between isotropic, anisotropic, and microscopic contributions. This refined characterization provides richer insights into tissue microstructure and can reveal features that may be obscured or conflated in simpler models.
```
RafaelNH, line 7: Write instead: Therefore, in addition to the kurtosis sources related to diffusion variance across compartments (K_aniso and K_iso related to the microscopic anisotropy and the variance of compartments mean diffusivities respectively), CTI also measures K_micro which quantifies non-Gaussian diffusion effects deviating for the multiple Gaussian component tissue representation (effects like restricted diffusion, exchange, compartments structural disorder like cross-sectional variance).

RafaelNH: Here cite the work by Lisa Novello and Rita Alves:

Novello, L., Henriques, R.N., Ianus, A., Feiweier, T., Shemesh, N., Jovicich, J. 2022. 
In vivo Correlation Tensor MRI reveals microscopic kurtosis in the human brain on 
a clinical 3T scanner. Neuroimage 254: 119137 doi: 10.1016/j.neuroimage.2022.119137 

 Alves, R., Henriques, R.N., Kerkelä, L., Chavarrías C., Jespersen, S.N., Shemesh, N. 
2022. Correlation Tensor MRI deciphers underlying kurtosis sources in stroke, 
Neuroimage 247: 118833. doi: 10.1016/j.neuroimage.2021.118833

doc/examples/reconst_cti.py
```diff
@@ -0,0 +1,152 @@
...
+Also, similar to DTI, assumes that the diffusion properties within each Gaussian compartment can be described by a diffusion tensor. However, unlike DTI, CTI allows for the presence of multiple diffusion tensors within a voxel, representing different Gaussian compartments.
```
RafaelNH, line 8: Remove this sentence which is not correct.

doc/examples/reconst_cti.py
```diff
@@ -0,0 +1,152 @@
...
+
+In DTI, we’ve diffusion tensor which measure the average diffusion properties within each voxel. in case of DKI we’ve diffusion tensor as well as kurtosis tensor. kurtosis tensor measures the dialation of diffusion properties from the gaussaian behaviour. But in case of CTI we’ve diffusion tensor, kurtosis tensor as well as covariance tensor. The covariance tensor describes the correlation between diffusion measurements at different b-values, offering insights into the relationship between diffusion processes at different scales.
```
RafaelNH, line 10: Remove this sentence.

doc/examples/reconst_cti.py
```diff
@@ -0,0 +1,152 @@
...
+
+CTI is also very suitable for situations which involve crossing fibers. As unlike DKI, it provides additional information in those cases by providing information about microscopic anisotropy.
```
RafaelNH, line 12: Remove this sentence, this is not the main advantage of CTI. Also, at the moment, we did not implement code to extract the directions from crossing fibers from CTI reconstructed data.

doc/examples/reconst_cti.py
```diff
@@ -0,0 +1,152 @@
...
+
+[ give maths for formulas ] 
+The CorrelationTensorModel expresses the diffusion-weighted signal in the same way as the DiffusionKurtosisModel.
```
RafaelNH, line 15: This is not correct. The CorrelationTensorModel does not express the diffusion-weighted signal in the same way as the DiffusionKurtosisModel. We started the CorrelationTensorFit by class instance of the DiffusionKurtosisTensorFit just to capture the common attributes, but it has a different expression reason why we had to change de "design matrix" for CTI.

RafaelNH: The different equations for DKI and CTI can be found in the original CTI paper (Henriques et al., 2020, equation 2 for DKI and equation 16 for CTI)

doc/examples/reconst_cti.py
```diff
@@ -0,0 +1,152 @@
...
+
+In DTI, we’ve diffusion tensor which measure the average diffusion properties within each voxel. in case of DKI we’ve diffusion tensor as well as kurtosis tensor. kurtosis tensor measures the dialation of diffusion properties from the gaussaian behaviour. But in case of CTI we’ve diffusion tensor, kurtosis tensor as well as covariance tensor. The covariance tensor describes the correlation between diffusion measurements at different b-values, offering insights into the relationship between diffusion processes at different scales.
+
+CTI is also very suitable for situations which involve crossing fibers. As unlike DKI, it provides additional information in those cases by providing information about microscopic anisotropy.
+
+[ give maths for formulas ] 
+The CorrelationTensorModel expresses the diffusion-weighted signal in the same way as the DiffusionKurtosisModel. 
+
+However it differentiates from the DiffusionKurtosis Model by calculating the different sources of kurtosis. The isotropic source is calculated as:
```
RafaelNH, line 17: I think for now we can ignore the maths in this tutorial, the information above is enough for the users. At the moment, if they want to know more they can go to the original papers. Let's keep it simple for this PR and for your GSoC project.

doc/examples/reconst_cti.py
```diff
@@ -0,0 +1,152 @@
...
+            
+and 
+
+:math:: 
+	K_{\text{intra}} = K_{\text{T}} - K_{\text{aniso}} - K_{\text{iso}}
+	
+	where K_{\text{intra}} refers to intra-compartmental kurtosis, 
+	K_{\text{T}} refers to total kurtosis, 
+	K_{\text{aniso}} refers to anisotropic source of kurtosis, 
+	K_{\text{iso}} refers to isotropic source of kurtosis
```
RafaelNH, line 56: So you can remove the lines from 8 to here

doc/examples/reconst_cti.py
```diff
@@ -0,0 +1,152 @@
...
+import matplotlib.pyplot as plt
+import numpy as np
+import math
+
+from dipy.reconst.tests.test_qti import _anisotropic_DTD, _isotropic_DTD
+from dipy.core.sphere import Sphere, HemiSphere
+from dipy.reconst.tests.test_qti import _anisotropic_DTD, _isotropic_DTD
+import dipy.reconst.qti as qti
+from dipy.reconst.qti import (from_3x3_to_6x1)
+from dipy.reconst.dti import (decompose_tensor, mean_diffusivity)
```
RafaelNH, line 71: Remove the above 6 lines of code

doc/examples/reconst_cti.py
```diff
@@ -0,0 +1,152 @@
...
+import dipy.reconst.cti as cti
+from dipy.io import read_bvals_bvecs
+from dipy.core.gradients import gradient_table
+from dipy.io.image import load_nifti
+"""
+CTI analysis requires specific gradient information from the data, often acquired from different gradient tables. In this example, we manually load b-values and b-vectors from two separate sets of files, 'bvals1.bval' and 'bvec1.bvec', and 'temp_data/bvals2.bval' and 'temp_data/bvec2.bvec'.
```
RafaelNH, line 77: Write instead. CTI analysis requires data acquired with double diffusion encoding data acquired with different pairs of b-values and gradient directions between the two diffusion epochs according to the considerations by Henriques et al., Magn Reson Med 2021

RafaelNH: Add a reference to Henriques et al., Magn Reson Med 2021.

RafaelNH: Add also the following sentence. To run CTI we need to have separate b-value and b-vector files for each DDE diffusion epoch. In this tutorial, a sample DDE dataset and respective b-value and b-vector files for the two epochs are fetched. If you want to process your own DDE data compatible with CTI, you will have to change the above lines of code to add the paths of your data.

doc/examples/reconst_cti.py
```diff
@@ -0,0 +1,152 @@
...
+The users should ensure that the data is formatted correctly for the CTI analysis they are performing.
+""" 
+data, affine = load_nifti('temp_data/RB_invivo_cti_data_f3.nii')
+bvals1, bvecs1 = read_bvals_bvecs('temp_data/bvals1.bval',
+                                  'temp_data/bvec1.bvec')
+bvals2, bvecs2 = read_bvals_bvecs('temp_data/bvals2.bval',
+                                  'temp_data/bvec2.bvec')
+                                  """ 
+In this example, the function `load_nifti` is used to load the CTI data from the file 'RB_invivo_cti_data_f3.nii' and returns the data as a nibabel Nifti1Image object along with the affine transformation. The b-values and b-vectors for two different gradient tables are loaded from 'bvals1.bval' and 'bvec1.bvec', and 'bvals2.bval' and 'bvec2.bvec' respectively using the `read_bvals_bvecs` function. These are then  further converted into gradient tables `gtab1` and `gtab2` which are essential for the CTI analysis being conducted.
```
RafaelNH, line 86: Replace
 "These are then  further converted into gradient tables `gtab1` and `gtab2` which are essential for the CTI analysis being conducted." 

by

For CTI reconstruction in DIPY, we need to define the b-values and b-vectors for each diffusion epoch on separate gradient table, as done in the above line of code.

doc/examples/reconst_cti.py
```diff
@@ -0,0 +1,152 @@
...
+"""
+gtab1 = gradient_table(bvals1, bvecs1)
+gtab2 = gradient_table(bvals2, bvecs2)
+"""
+Before fitting the data, we perform some data pre-processing. We first compute a brain mask to avoid unnecessary calculations on the background of the image.
```
RafaelNH, line 91: Here say that the data of these example is already pre-processed according to the description of the 2021 paper. Use the data and mask provided in https://zenodo.org/record/8276773

doc/examples/reconst_cti.py
```diff
@@ -0,0 +1,152 @@
...
+"""
+maskdata, mask = median_otsu(data, vol_idx=[0, 1], median_radius=4, numpass=2,
+                             autocrop=False, dilate=1)
+"""
+Now that we have loaded and created a mask for the data we can go forward with CTI fitting. For this, the CTI model is first defined for the data’s GradientTable object by instantiating the CorrelationTensorModel object in the following way:
```
RafaelNH, line 96: Now that we loaded the data and generated the two gradient tables we can go forward with CTI fitting. For this, the CTI model is first defined for GradientTable objects gtab1 and gtab2 by instantiating the CorrelationTensorModel object in the following way:

doc/examples/reconst_cti.py
```diff
@@ -0,0 +1,152 @@
...
+"""
+ctimodel = cti.CorrelationTensorModel(gtab1, gtab2)
+"""
+To fit the data using the defined model object, we call the fit function of this object.
+"""
+ctifit = ctimodel.fit(data, mask = mask)
```
RafaelNH, line 102: mask =  mask ==> mask=mask

doc/examples/reconst_cti.py
```diff
@@ -0,0 +1,152 @@
...
+"""
+The fit method for the CTI model produces a CorrelationTensorFit object, which contains the attributes of both the DKI and DTI models. Given that CTI is a layered model built upon the DKI, which itself extends the DTI model, the CorrelationTensorFit instance captures a comprehensive set of parameters and attributes from these underlying models.
+
+For instance, since the DKI model inherently estimates the diffusion tensor parameters, all standard DTI statistics, such as fractional anisotropy (FA), mean diffusivity (MD), axial diffusivity (AD), and radial diffusivity (RD), can be directly extracted from the CorrelationTensorFit instance. Additionally, the CTI offers further insights by evaluating correlation tensors, offering a richer perspective on the tissue's microstructural environment.
```
RafaelNH, line 106: Write instead. For instance, since the CTI model inherently estimates all DTI and DKI statistics, such as the mean, axial and radial diffusivities (MD, AD, RD) and the mean, axial and radial kurtosis (MK, RK, AK).

Add some lines of code retrieving the main DTI/DKI parameters to illustrate what you are saying.

doc/examples/reconst_cti.py
```diff
@@ -0,0 +1,152 @@
...
+axarr[1].imshow(kaniso_map[:, :, slice_idx], cmap='gray', origin='lower', vmin = 0, vmax = 1)
+axarr[1].set_title('Kaniso Map')
+
+axarr[2].imshow(kmicro_map[:, :, slice_idx], cmap='gray', origin='lower', vmin = 0, vmax = 1)
+axarr[2].set_title('Kmicro Map')
+
+plt.tight_layout()
+plt.show()
+
+
```
RafaelNH, line 128: Remove the extra blank spaces below

RafaelNH (changes requested):
I am now submitting some comments about the cti tests and the tutorial that you uploaded yerterday. I think I found why some test are not working and detected a mistake on the implementation of K_total, see my comments.

dipy/reconst/tests/test_cti.py
```diff
@@ -0,0 +1,381 @@
...
+        variance_of_mean_diffusivities = np.var(mean_diffusivities)
+        mean_D = np.mean(mean_diffusivities) 
+        ground_truth_K_iso = 3 * variance_of_mean_diffusivities / (mean_D ** 2)
+        
+        error_msg = (
+            f"Calculated K_iso {K_iso} for anisotropicDTD does not match the "
+            f"ground truth {ground_truth_K_iso}"
+        )
+        assert np.isclose(K_iso, ground_truth_K_iso), error_msg
+
```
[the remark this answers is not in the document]
shilpiprd: do you mean that I should just put ground_truth = 0 for all DTDs and then compare ? 
Shouldn't we actually calculate ground truth the way with other sources ?

doc/examples/reconst_cti.py
shilpiprd: I think I'll only add eq16.

dipy/reconst/tests/test_cti.py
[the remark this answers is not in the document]
RafaelNH: Yes, for all DTDs ground_truth_K_micro = 0

dipy/reconst/cti.py
```diff
@@ -0,0 +1,467 @@
...
+            \(D_{ij}\) are elements of the diffusion tensor,
+            \(\overline{D}\) is the mean of the diffusion tensor.
+            \{\overline{W}} is the mean kurtosis,
+            \(W_{ijkl}\) are elements of the kurtosis tensor.
+        """
+
+        mean_K = self.mkt()
+        D = self.quadratic_form
+        mean_D = np.trace(D) / 3
+        psi = 2 / 5 * ((np.sqrt(D[..., 0, 0]) + np.sqrt(D[..., 1, 1])
```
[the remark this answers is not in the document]
RafaelNH: This was marked as resolved but is still not right. Right psi calculation is 

psi = 2 / 5  * (D[..., 0, 0] ** 2 + D[..., 1, 1] ** 2 + D[..., 2, 2] ** 2 + 2 * D[..., 0, 1] ** 2 + 2*D[..., 1, 2] ** 2 + 2*D[..., 1, 2] ** 2) / (mean_D ** 2) - 6/5

dipy/reconst/cti.py
```diff
@@ -0,0 +1,467 @@
...
+                        + np.sqrt(D[..., 2, 2])
+                        + 2 * np.sqrt(D[..., 0, 1]) + 2 * np.sqrt(D[..., 0, 2])
+                        + np.sqrt(D[..., 1, 2])) / (mean_D ** 2)) - (6/5)
+
+        excess_K = 1/5 * mean_K + psi
```
RafaelNH, line 385: In addition to the mistake in the calculation in psi. Why are you multiplying mean_K by 1/5? See equation 9 of the CTI neuroimage 2020 paper. According to that Ktotal = mean_K + psi. Avoid also using the variable name excess_K since it may be misleading. I would just combine the mean_K and psi directly in the function return, i.e.:

return mean_K + psi

doc/examples/reconst_cti.py
shilpiprd: Should I put the link in the documentation as well ? So that people can download the data themselves ?

RafaelNH: Leave it like this for now, since the tutorial have to be updated after PR #2879 is merged anyway.

arokem (approved):
Looks great! I didn't have time to run the example on my end, but thought I would provide all this feedback anyway. I believe this can be merged very soon.

shilpiprd:
> Looks great! I didn't have time to run the example on my end, but thought I would provide all this feedback anyway. I believe this can be merged very soon.

Thanks Arokem for your feedback, I'll be sure to look throught the whole thing and submit a commit soon.

dipy/reconst/cti.py
```diff
@@ -0,0 +1,465 @@
...
+            A GradientTable class instance for second DDE diffusion epoch
+
+        S0 : float or ndarray (optional)
+            The non diffusion-weighted signal in every voxel, or across all
+            voxels. Default: 1
+
+        Returns
+        -------
+        S : ndarray
+            Simulated signal based on the CTI model:
```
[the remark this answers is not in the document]
shilpiprd: Technically, there is a signal which is expressed by the CorrelationTensor Model, however it's quite long. Also, I don't think it's necessary to mention it there as it has been mentioned in the tutorial. So I think I'll just remove the colon.

dipy/reconst/cti.py
```diff
@@ -0,0 +1,465 @@
...
+
+        where: \(K_{aniso}\) is the anisotropic kurtosis,
+            \(\langle V_{\lambda}(D_c) \rangle\) represents the mean of the
+            variance of eigenvalues of the diffusion tensor,
+            \(\overline{D}\) is the mean of the diffusion tensor.
+        """
+        C = self.ct
+        D = self.quadratic_form
+
+        Variance = 2/9 * (C[..., 0] + D[..., 0, 0] ** 2 + C[..., 1]
```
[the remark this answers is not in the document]
shilpiprd: It is actually,  This is the equation of K_aniso, 
Screenshot from 2023-09-21 19-10-39
And this is the numerator: 
Screenshot from 2023-09-21 19-10-52
I've taken these from the original CTI paper.

dipy/reconst/cti.py
```diff
@@ -0,0 +1,465 @@
...
+
+        where: \(K_{iso}\) is the isotropic kurtosis,
+            \(V({\overline{D}^c})\) represents the variance of the diffusion
+            tensor raised to the power c,
+            \(\overline{D}\) is the mean of the diffusion tensor.
+
+        """
+        C = self.ct
+        mean_D = self.md
+        Variance = 1/9 * (C[..., 0] + C[..., 1] + C[..., 2] + 2 * C[..., 5]
```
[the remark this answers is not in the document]
shilpiprd: So, K_iso is: 
Screenshot from 2023-09-21 19-13-02
And it's numerator is: 
Screenshot from 2023-09-21 19-13-29

dipy/reconst/cti.py
```diff
@@ -0,0 +1,465 @@
...
+
+        K_total = self.K_total
+        K_aniso = self.K_aniso
+        K_iso = self.K_iso
+        micro_K = K_total - K_aniso - K_iso
+        return micro_K
+
+
+def params_to_cti_params(result, min_diffusivity=0):
+    # Extracting the diffusion tensor parameters from solution
```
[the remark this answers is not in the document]
shilpiprd: Actually even in DKI module, it's written like this (Line no 2175 of dki.py file). So, I'm not sure if a docstring is required here.

dipy/reconst/tests/test_cti.py
```diff
@@ -0,0 +1,458 @@
...
+    ccti[15] = C[15] / 2
+    ccti[16] = C[16] / 2
+    ccti[17] = C[17] / 2
+    ccti[18] = C[18] / (2 * const)
+    ccti[19] = C[19] / (2 * const)
+    ccti[20] = C[20] / (2 * const)
+    return ccti
+
+
+def multi_gaussian_k_from_c(ccti, MD):
```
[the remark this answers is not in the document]
shilpiprd: Yes, I think that can be done. @RafaelNH what do you think ?

doc/examples/reconst_cti.py
```diff
@@ -0,0 +1,171 @@
+"""
+========================================================================
+Reconstruction of the diffusion signal with the correlation tensor model
+========================================================================
+
+The Correlation Tensor MRI (CTI) is a method that uses double diffusion
+encoding data to resolve sources of kurtosis. Compared to the Q-space
+Trajectory Imaging
+(see :ref:`sphx_glr_examples_built_reconstruction_reconst_qti.py`) [NetoHe2020]_.
```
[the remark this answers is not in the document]
shilpiprd: Yes, you're right.

dipy/reconst/utils.py
```diff
@@ -50,6 +50,106 @@ def dki_design_matrix(gtab):
     return B
 
 
+def cti_design_matrix(gtab1, gtab2):
```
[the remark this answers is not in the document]
shilpiprd: I'm not entirely sure how this could be done. Rafael do you have some suggestions?

arokem:
OK - looks like most of my comments are addressed with two outstanding: 

- [ ] https://github.com/dipy/dipy/pull/2816#discussion_r1327719389
- [ ] https://github.com/dipy/dipy/pull/2816#discussion_r1327720568

I also just ran the example, and it works great on my laptop.

dipy/reconst/tests/test_cti.py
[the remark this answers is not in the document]
RafaelNH: Yes, we can added it as part of the main CTI code in cti.py. Actually, I intend to use that function on future developments of CTI (in a future PR), where I intent to include main DKI quantities decoupled from microscopic kurtosis effects, noting that the kurtosis tensor produced from CTI's covariance tensor are decoupled from microscopic kurtosis effects.

dipy/reconst/utils.py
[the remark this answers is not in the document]
RafaelNH: A good test for this function is just checking if we get the same design matrix in the following two conditions:

A1 = cti_design_matrix(gtab1, gtab2)

and 

A2 = cti_design_matrix(gtab2, gtab1)

A1 has to be the same as A2 due to the symmetry of the design matrix, the order that you apply two different gradient directions should not matter. Actually this test was used in my local computer in a early version of the code to detect some wrong signs in the design matrix, so including the test in test_cti.py will be useful to make sure that the design matrix does not break in possible future code updates.

RafaelNH:
> OK - looks like most of my comments are addressed with two outstanding:
> 
> * [ ]  [Correlation Tensor Imaging #2816 (comment)](https://github.com/dipy/dipy/pull/2816#discussion_r1327719389)
> * [ ]  [Correlation Tensor Imaging #2816 (comment)](https://github.com/dipy/dipy/pull/2816#discussion_r1327720568)
> 
> I also just ran the example, and it works great on my laptop.

@shilpiprd - Regarding the missing comments, I added some suggestions on how to address them. See my comment response above.

RafaelNH (approved):

doc/examples/reconst_cti.py
```diff
@@ -0,0 +1,172 @@
...
+The users should also ensure that the data is formatted correctly for the CTI
+analysis they are performing.
+"""
+fdata, fbvals1, fbvecs1, fbvals2, fbvecs2, fmask = get_fnames('cti_rat1')
+data, affine = load_nifti(fdata)
+bvals1, bvecs1 = read_bvals_bvecs(fbvals1, fbvecs1)
+bvals2, bvecs2 = read_bvals_bvecs(fbvals2, fbvecs2)
+"""
+In this example, the function ``load_nifti`` is used to load the CTI data from
+the file ``RB_invivo_cti_data_f3.nii`` and returns the data as a nibabel
```
RafaelNH, line 80: Replace "In this example, the function ``load_nifti`` is used to load the CTI data from
the file ``RB_invivo_cti_data_f3.nii`` and returns the data as a nibabel..."
by 

"In this example, the function ``load_nifti`` is used to load the CTI data saved in file``fdata`` and returns the data as a nibabel...

skoudoro (commented):
Hi @shilpiprd,

Overall, it looks good! it seems there are some indentation issue concerning your function docstring. Can you fix that ?

Also, By running your example, I encounter this warnings:
```python
/Users/skoudoro/devel/dipy/dipy/reconst/cti.py:347: RuntimeWarning: divide by zero encountered in scalar divide
  K_iso = 3 * (Variance / (mean_D ** 2))
/Users/skoudoro/devel/dipy/dipy/reconst/cti.py:323: RuntimeWarning: divide by zero encountered in scalar divide
  K_aniso = (6/5) * (Variance / (mean_D ** 2))
/Users/skoudoro/devel/dipy/dipy/reconst/cti.py:379: RuntimeWarning: invalid value encountered in scalar divide
  psi = 2 / 5 * ((D[..., 0, 0]**2 + D[..., 1, 1]**2
```

it would be great to fix those warning by making sure that you avoid to divide by 0. 
Otherwise, see my small comment below.

Thanks for the future fixes !

dipy/reconst/cti.py
```diff
@@ -0,0 +1,470 @@
...
+        Parameters are ordered as follows:
+
+            1. Three diffusion tensor's eigenvalues
+            2. Three lines of the eigenvector matrix each containing the
+            first, second and third coordinates of the eigenvector
+            3. Fifteen elements of the kurtosis tensor
+            4. Twenty-One elements of the covariance tensor
+        S0 : float or ndarray (optional)
+            The non diffusion-weighted signal in every voxel, or across all
+            voxels. Default: 1
```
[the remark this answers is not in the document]
shilpiprd: You're right, S0 shouldn't be there in the docstring.

shilpiprd:
I think I've made all the suggested changes, but please let me know if anything else needs to be added .

skoudoro (approved):
Ok, I think this PR is ready to be merged.

@RafaelNH, Should we wait for the review of @kerkelae ? or we can go ahead ?

arokem:
I think that with 3 approving reviews, we should go ahead and merge. If @kerkelae wants to take a look later, we can address comments that may arise through a separate PR.

skoudoro:
ok, great job @shilpiprd !  Thank you for mentoring @RafaelNH  and thank you for the review @arokem!

merging

Issue #2790, opened 2023-04-26
DTI fitting function with NLLS method raises an error.


GalKepler:
## Description
Changes introduced to the DTI fitting function for NLLS method raise an error.

I've narrowed the issue to a single line of code that was changed sometime in the last few months:
In [here](https://github.com/dipy/dipy/blob/master/dipy/reconst/dti.py#L1724), the OLS parameters were previously estimated as:

```python
inv_design = np.linalg.pinv(design_matrix)
log_s = np.log(flat_data)
D = np.dot(inv_design, log_s.T).T
```
This behavior changed into:
````python
D = ols_fit_tensor(design_matrix, flat_data, return_lower_triangular=True)
````

Which, for some reason, raises the following error:
```python
ValueError: could not broadcast input array from shape (10000,7) into shape (10000,12)
```
Changing this line of code to what it previously was fixed this issue (while passing the same inputs for DTI reconstruction), but I'm not sure why was this change introduced in the first place so I'm not sure which is the correct way to go.

## Way to reproduce
- [X] Code example
```python
from dipy.workflows.reconst import ReconstDtiFlow
a = {"input_files" : path_to_dwi_nifti,
"bvalues_files": path_to_dwi_bval,
"bvectors_files": path_to_dwi_bvec,
"mask_files": path_to_dwi_mask,
"out_dir": path_to_output_directory,
"fit_method": "NLLS"}

runner = ReconstDtiFlow()
runner.run(**a)
```

- [X] Operating system and version (run `python -c "import platform; print(platform.platform())"`)
Linux-5.10.0-21-amd64-x86_64-with-glibc2.31
- [X] Python version (run `python -c "import sys; print("Python", sys.version)"`)
3.10.2
- [X] dipy version (run `python -c "import dipy; print(dipy.__version__)"`)
1.7.0

skoudoro:
Hi @GalKepler,

Thank you for your feedback.

@samcoveney, can you explain your recent change to @GalKepler. He had some issue. 

Thank you

samcoveney:
sorry @GalKepler and @skoudoro  I completely missed this conversation.

I haven't seen this error myself, and I use the NLLS fitting all the time.

I am having some problems reproducing this at the moment, since I keep seeing the error

```
ModuleNotFoundError: No module named 'dipy.direction.ptt_direction_getter'
```

when trying to run this workflow, or when trying to run some of the tests. But the NLLS fitting routines are working for me.

[...]

PR #2820, opened 2023-06-05
fixed bug for nlls fitting

samcoveney:
Fixed bug in NLLS, see  #2790 for details.


samcoveney:
_Alternatively_ we could just finally deal with #2730 which already contains this fix plus others

skoudoro:
implemented and closed by #2730

PR #2821, opened 2023-06-05
NF -  added pft min wm parameter

gabknight:
This is the part 1 of 2 of refactoring PR #1627, adding a new parameter to PFT.

The new optional parameter (default value is 0), `min_wm_pve_before_stopping`, improves PFT starting in voxel with non-zero GM partial volume estimate (PVE). The 'ENDPOINT' streamline status is ignored until the tracking reaches a position with `WM PVE > min_wm_pve_before_stopping`. This is useful when seeding from the GM-WM interface or in subcortical GM, where tractography may stop prematurely in the GM. This parameter forces the tracking to reach the deep WM before stopping, or to reach voxels with WM PVE of 0.

### DIFF
--- a/dipy/tracking/local_tracking.py
+++ b/dipy/tracking/local_tracking.py
@@ -234,7 +234,7 @@ class ParticleFilteringTracking(LocalTracking):
             random.seed).
         save_seeds : bool
             If True, return seeds alongside streamlines
-        min_wm_pve_before_stopping : int
+        min_wm_pve_before_stopping : int, optional
             Minimum white matter pve (1 - stopping_criterion.include_map -
             stopping_criterion.exclude_map) to reach before allowing the
             tractography to stop.
@@ -265,7 +265,7 @@ class ParticleFilteringTracking(LocalTracking):
         if particle_count <= 0:
             raise ValueError("The particle count must be greater than 0.")
 
-        if min_wm_pve_before_stopping < 0 or min_wm_pve_before_stopping > 1:
+        if 0 < min_wm_pve_before_stopping < 1:
             raise ValueError("The min_wm_pve_before_stopping value must be "
                              "between 0 and 1.")
 
--- a/dipy/tracking/local_tracking.py
+++ b/dipy/tracking/local_tracking.py
@@ -265,7 +265,7 @@ class ParticleFilteringTracking(LocalTracking):
         if particle_count <= 0:
             raise ValueError("The particle count must be greater than 0.")
 
-        if not 0 < min_wm_pve_before_stopping < 1:
+        if not 0 <= min_wm_pve_before_stopping <= 1:
             raise ValueError("The min_wm_pve_before_stopping value must be "
                              "between 0 and 1.")
 
--- a/dipy/tracking/local_tracking.py
+++ b/dipy/tracking/local_tracking.py
@@ -182,7 +182,8 @@ class ParticleFilteringTracking(LocalTracking):
                  step_size, max_cross=None, maxlen=500,
                  pft_back_tracking_dist=2, pft_front_tracking_dist=1,
                  pft_max_trial=20, particle_count=15, return_all=True,
-                 random_seed=None, save_seeds=False):
+                 random_seed=None, save_seeds=False,
+                 min_wm_pve_before_stopping=0):
         r"""A streamline generator using the particle filtering tractography
         method [1]_.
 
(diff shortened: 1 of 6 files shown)


dipy/tracking/local_tracking.py
```diff
@@ -221,6 +222,10 @@ def __init__(self, direction_getter, stopping_criterion, seeds, affine,
             random.seed).
         save_seeds : bool
             If True, return seeds alongside streamlines
+        min_wm_pve_before_stopping : int
```
skoudoro, line 225: `int, optional` instead

dipy/tracking/local_tracking.py
```diff
@@ -248,8 +253,12 @@ def __init__(self, direction_getter, stopping_criterion, seeds, affine,
         if particle_count <= 0:
             raise ValueError("The particle count must be greater than 0.")
 
-        self.directions = np.empty((maxlen + 1, 3), dtype=float)
+        if min_wm_pve_before_stopping < 0 or min_wm_pve_before_stopping > 1:
```
skoudoro, line 256: `if not 0 < min_wm_pve_before_stopping < 1` instead

dipy/tracking/localtrack.pyx
```diff
@@ -54,7 +54,7 @@ def local_tracker(
         Ending state of the streamlines as determined by the StoppingCriterion.
     """
     cdef:
-        cnp.npy_intp i
+        int i
```
skoudoro, line 57: why this change? 

it is recommend to use npy_intp variables to iterate over numpy arrays instead of int and size_t.

> "An initial attempt to implement this feature tried to use size_t. It
quickly turned out that this cannot work: Python uses negative indices
in many places (to indicate counting from the end). Even in places where
size_t would be usable, too many reformulations of code where necessary,
e.g. in loops like: `for(index = length-1; index >= 0; index--)` This loop will never terminate if index is changed from int to size_t."

dipy/tracking/localtrack.pyx
```diff
@@ -144,7 +148,7 @@ def pft_tracker(
 
     """
     cdef:
-        cnp.npy_intp i
+        int i
```
skoudoro, line 151: same comment, see above

> why this change?  it is recommend to use npy_intp variables to iterate over numpy arrays instead of int and size_t.

dipy/tracking/localtrack.pyx
```diff
@@ -188,10 +192,12 @@ cdef _pft_tracker(DirectionGetter dg,
                   cnp.float_t[:, :, :, :] particle_dirs,
                   cnp.float_t[:] particle_weights,
                   cnp.int_t[:, :] particle_steps,
-                  cnp.int_t[:, :] particle_stream_statuses):
+                  cnp.int_t[:, :] particle_stream_statuses,
+                  double min_wm_pve_before_stopping):
     cdef:
-        int i, pft_trial, pft_streamline_i, back_steps, front_steps
+        int i, j, pft_trial, back_steps, front_steps
```
skoudoro, line 198: same comment, see above

> why this change?  it is recommend to use npy_intp variables to iterate over numpy arrays instead of int and size_t.

skoudoro (changes requested):
Overall, Looks good to me. 

it would be great to update one of the tutorial to explain this parameter. It does not seems obvious to me when I read the code, but the PR description is super clear. 

I do not know if you prefer to do that here or on a new PR. Otherwise, see below few comments

skoudoro:
Hi @gabknight,

Can you address the conflict issue by a rebase and address my comment above. Thanks !

dipy/tracking/local_tracking.py
gabknight: fixed. thx.

dipy/tracking/localtrack.pyx
gabknight: ok. I revert it.

dipy/tracking/local_tracking.py
gabknight: done.

gabknight:
Hi @skoudoro,

Thank you. I addressed your comments and added a few lines in the PFT example.

skoudoro:
All looks good. 

Before merging, I still need to check what is going on with `test_horizon_events` (https://github.com/dipy/dipy/actions/runs/5927644456/job/16189637170?pr=2821#step:9:3119).

The CI failed with this PR, and I do not know if it is related or not.

skoudoro:
Ok, I spent some times, and I can not reproduce it. Also, it does not seems related so I am going ahead and merge this PR. 

Thank you @gabknight!

PR #2860, opened 2023-08-11
Updates HCP fetcher dataset_description to be compatible with current BIDS

arokem:
Small update to reflect changes to the BIDS spec since this was first written.


skoudoro:
thank you for the update @arokem, merging

PR #2864, opened 2023-08-16
Change internal space/origin when using sft.to_x() with an empty sft.

EmmaRenauld:
When using an empty SFT, internal space (sft._space) is never modified. Changing this ensures that we can compare the compatibility of many SFT, even when some are empty.

Same with origin.

### DIFF
--- a/dipy/io/tests/test_stateful_tractogram.py
+++ b/dipy/io/tests/test_stateful_tractogram.py
@@ -12,6 +12,8 @@ import pytest
 from dipy.data import fetch_gold_standard_io
 from dipy.io.stateful_tractogram import Origin, Space, StatefulTractogram
 from dipy.io.streamline import load_tractogram, save_tractogram
+from dipy.io.utils import is_header_compatible
+
 
 from dipy.utils.optpkg import optional_package
 fury, have_fury, setup_module = optional_package('fury')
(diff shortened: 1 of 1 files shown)


arokem:
This makes sense to me! Any chance to get a test of this functionality?

frheault:
Good catch ! This was indeed an oversight !

EmmaRenauld:
What type of test would you like? Here is an example of previous problematic usage:

```
sft1 = load_tractogram(some_path)
sft2 = sft1.from_sft([], sft)

print(sft1.space) --> RASMM
print(sft2.space) --> RASMM

sft1.to_vox()
sft2.to_vox()

print(sft1.space) --> VOX
print(sft2.space) --> RASMM

print(StatefulTractogram.are_compatible(sft1, sft2))  --> FALSE
```

I could add these lines somewhere, they should now be compatible. Where should I add it?


Can you help me understand why the Github tests were not successful?

skoudoro:
> Can you help me understand why the Github tests were not successful?

You can ignore, this is not due to your PR. it seems there is an incompatibility between `cvxpy` and `scipy` on conda. We might need to pin specific version to avoid this issue

skoudoro (approved):
Thank you for the update @EmmaRenauld.

Looks good to me. I will wait until Monday to merge it in case there are any additional comments.

PR #2871, opened 2023-08-23
Corrected paths to static files

lb-97:


skoudoro (approved):
Thanks, seems ok. merging

PR #2875, opened 2023-08-24
Increased PTT performances

gabknight:
-  cythonized `find_closest(.)` from `dipy.core.shpere` to `dipy.direction.pmf`

-  optimized `pmf.get_pmf_value(.)`: only interpolated the direction of interest instead of the whole pmf.

-  improved cython code in `dipy.direction.ptt_direction_getter.pyx`

On my tests on the disco dataset, the performances improved by 5X.


gabknight:
@skoudoro @Garyfallidis  I had to modify the `cpdef` to `cdef` in the `fast_numpy` functions to change the inputs from `double[:]` to `double*` . As such, I also had to remove the tests. Do we move forward like this? Another option would be to add `_c` to the `cdef` function names, and create `cpdef` functions with the same names that calls their respective `_c` functions. This option would be fine, but solely used for the testing. What do you think?

skoudoro:
I am ok, with that only for fast_numpy module. 

In parallel, we need to figure out a way to add unit test for cdef functions

skoudoro:
Can you comment then instead or deleting them for now ?

gabknight:
This is ready for review. Further improvements will come in another PR.

skoudoro (approved):
Thank you @gabknight,

I am going ahead and merge this PR. Looks good to me.

PR #2879, opened 2023-08-26
Add fetcher for a sample CTI dataset

RafaelNH:
Hi all! I've added a fetcher for a sample CTI dataset that can be used for the implementation in PR #2816.

To fetch and load the data, you can use the following lines of code:
```
from dipy.data import get_fnames
from dipy.io import read_bvals_bvecs
from dipy.io.image import load_nifti

fdata, fbvals1, fbvecs1, fbvals2, fbvecs2, fmask = get_fnames('cti_rat1')

data, affine = load_nifti(fdata)
bvals1, bvecs1 = read_bvals_bvecs(fbvals1, fbvecs1)
bvals2, bvecs2 = read_bvals_bvecs(fbvals2, fbvecs2)
mask, affine = load_nifti(fmask)
```

Full details of the dataset can be found in the manuscript: Henriques et al., (2021). Evidence for microscopic kurtosis in neural tissue revealed by correlation tensor MRI. Magn Reson Med 86 (6): 3111-3130.

Fetched dataset correspond to the data for Rat 1 of that manuscript.

Please check if the fetcher is working for you and let me know if you have any comments.

PR #2881, opened 2023-08-28
Adding fetcher in the test_file for #2879

shilpiprd:
Hey Serge ! I felt I could do this so I added what was needed in the tests, #2879


skoudoro:
Hi @shilpiprd,

Thank you for this but both changes need to be done on the same PR. you could restart this PR on the top of rafael branch or you need to ask rafael to update this. Otherwise, it looks good.

RafaelNH:
Indeed, it is better to have this directly in PR #2879. Since it is only a line of code, I will commit it myself in that PR. Thanks anyway @shilpiprd to give a look to it - I really appreciated it.

PR #2885, opened 2023-08-31
[ENH] Add minimum length to streamline generator

36000:
This can be useful if you are going to use the streamlines for bundle recognition, and you don't want to have to deal with small streamlines.

### DIFF
--- a/dipy/tracking/local_tracking.py
+++ b/dipy/tracking/local_tracking.py
@@ -186,7 +186,7 @@ class LocalTracking(object):
 class ParticleFilteringTracking(LocalTracking):
 
     def __init__(self, direction_getter, stopping_criterion, seeds, affine,
-                 step_size, max_cross=None, maxlen=500,
+                 step_size, max_cross=None, maxlen=500, minlen=0,
                  pft_back_tracking_dist=2, pft_front_tracking_dist=1,
                  pft_max_trial=20, particle_count=15, return_all=True,
                  random_seed=None, save_seeds=False,
@@ -218,6 +218,9 @@ class ParticleFilteringTracking(LocalTracking):
         maxlen : int
             Maximum number of steps to track from seed. Used to prevent
             infinite loops.
+        minlen : int
+            Minimum number of steps to track from seed. Can be useful
+            for filtering out useless streamlines.
         pft_back_tracking_dist : float
             Distance in mm to back track before starting the particle filtering
             tractography. The total particle filtering tractography distance is
@@ -296,6 +299,7 @@ class ParticleFilteringTracking(LocalTracking):
                                                         step_size,
                                                         max_cross,
                                                         maxlen,
+                                                        minlen,
                                                         True,
                                                         return_all,
                                                         random_seed,
--- a/dipy/tracking/local_tracking.py
+++ b/dipy/tracking/local_tracking.py
@@ -86,9 +86,8 @@ class LocalTracking(object):
             raise ValueError("step_size must be greater than 0.")
         if maxlen < 1:
             raise ValueError("maxlen must be greater than 0.")
-        if minlen >= maxlen:
-            maxlen = minlen + 2
-            warn("maxlen updated to be greater than minlen")
+        if minlen + 1 >= maxlen:
+            raise ValueError("maxlen must be greater than minlen + 1")
         if not isinstance(seeds, Iterable):
             raise ValueError("seeds should be (N,3) array.")
 
--- a/dipy/tracking/local_tracking.py
+++ b/dipy/tracking/local_tracking.py
@@ -86,8 +86,8 @@ class LocalTracking(object):
             raise ValueError("step_size must be greater than 0.")
         if maxlen < 1:
             raise ValueError("maxlen must be greater than 0.")
-        if minlen + 1 >= maxlen:
-            raise ValueError("maxlen must be greater than minlen + 1")
+        if minlen > maxlen:
+            raise ValueError("maxlen must be greater than or equal to minlen")
         if not isinstance(seeds, Iterable):
             raise ValueError("seeds should be (N,3) array.")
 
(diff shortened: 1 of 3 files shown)


skoudoro:
Hi @36000, 

Thank you for this. Yes, I think it will be useful.

It will allow us to avoid this extra step that I am personally doing:

```python
from dipy.tracking.streamlinespeed import length

streamlines_generator = LocalTracking(csa_peaks, stopping_criterion, seeds,
                                          affine=resliced_affine, step_size=.5)
# remove short streamlines (< 30mm)
streamlines_generator = (s for s in streamlines_generator if length(s) > 30.0)
streamlines = Streamlines(streamlines_generator)
``

**Comment by skoudoro** [comment]
You will have to add and update some tests.  Can you anticipate this?  thank you

File: dipy/tracking/local_tracking.py [hunk 1]
```diff
@@ -31,8 +31,9 @@ def _get_voxel_size(affine):
         return np.sqrt(dotlin.diagonal())
 
     def __init__(self, direction_getter, stopping_criterion, seeds, affine,
-                 step_size, max_cross=None, maxlen=500, fixedstep=True,
-                 return_all=True, random_seed=None, save_seeds=False):
+                 step_size, max_cross=None, maxlen=500, minlen=0,
```
skoudoro, line 34: As you can see in this link: https://github.com/dipy/dipy/blob/eede25d73dcff7fff758d64dbdb829533de6a05a/dipy/tracking/local_tracking.py#L172

`minlen` was 1.  I would recommend to keep this value to avoid breaking previous code

dipy/tracking/local_tracking.py
```diff
@@ -59,6 +60,9 @@ def __init__(self, direction_getter, stopping_criterion, seeds, affine,
         maxlen : int
             Maximum number of steps to track from seed. Used to prevent
             infinite loops.
+        minlen : int
```
skoudoro, line 63: `int, optional` instead

dipy/tracking/local_tracking.py
```diff
@@ -169,17 +174,19 @@ def _generate_tractogram(self):
 
                 # move to the next streamline if only the seed position
                 # and not return all
-                if len(streamline) > 1 or self.return_all:
-                    if self.save_seeds:
-                        yield streamline, s
-                    else:
-                        yield streamline
+                len_sl = len(streamline)
+                if len_sl >= self.min_length and len_sl <= self.max_length \
```
skoudoro, line 178: same here, As you can see in this link: https://github.com/dipy/dipy/blob/eede25d73dcff7fff758d64dbdb829533de6a05a/dipy/tracking/local_tracking.py#L172

it was `> minlen`  with minlen default to 1.  I would recommend to keep this condition and default value to avoid breaking previous code

dipy/tracking/local_tracking.py
```diff
@@ -211,6 +218,9 @@ def __init__(self, direction_getter, stopping_criterion, seeds, affine,
         maxlen : int
             Maximum number of steps to track from seed. Used to prevent
             infinite loops.
+        minlen : int
```
skoudoro, line 221: same : `int, optional`

dipy/tracking/local_tracking.py
```diff
@@ -169,17 +174,19 @@ def _generate_tractogram(self):
...
+                            yield streamline, s
+                        else:
+                            yield streamline
 
 
 class ParticleFilteringTracking(LocalTracking):
 
     def __init__(self, direction_getter, stopping_criterion, seeds, affine,
-                 step_size, max_cross=None, maxlen=500,
+                 step_size, max_cross=None, maxlen=500, minlen=0,
```
skoudoro, line 189: same as above

skoudoro (changes requested):
Hi @36000,

Thank you for adding a test. See below my comments

dipy/tracking/local_tracking.py
36000: Good point, actually I am going to set it to 2 because I am now using `>=` instead of `>`.

36000:
I think this is good to go, thoughts?

dipy/tracking/local_tracking.py
```diff
@@ -91,6 +95,7 @@ def __init__(self, direction_getter, stopping_criterion, seeds, affine,
         self.fixed_stepsize = fixedstep
         self.max_cross = max_cross
         self.max_length = maxlen
+        self.min_length = minlen
```
skoudoro, line 98: just to avoid weird setup for someone not paying attention.
On the top of the `__init__py`, can you add a check, something like (pseudo code to validate):
```python
if minlen > maxlen:
    maxlen = minlen + 2
   warn("max length updated due to strange value")
```

skoudoro (commented):
Looks good to me. Thanks @36000.

Just added a small request below then I think it is ready to go. 

it would be great if @gabknight or @arokem could have a look also. Thanks

dipy/tracking/local_tracking.py
```diff
@@ -81,6 +86,9 @@ def __init__(self, direction_getter, stopping_criterion, seeds, affine,
             raise ValueError("step_size must be greater than 0.")
         if maxlen < 1:
             raise ValueError("maxlen must be greater than 0.")
+        if minlen >= maxlen:
```
arokem, line 89: I would consider being a bit more strict here and raising an error in this case. If the user is asking for minlen that is larger than maxlen they must be confused about something, but I don't think that we can assume that what they want is `maxlen = minlen + 2`

arokem (approved):
Overall, looks good to me. 

I wonder whether we want to be a bit more strict in input validation, though.

skoudoro:
All Looks good, but not sure why everything is failing. 

I believe it is not related to your PR. I suppose, a dependency update just break everything. I will look into it a bit later today then I will go ahead and merge your PR if I confirm my guess

dipy/tracking/local_tracking.py
skoudoro: I am ok with this plan 👍, let raise an error

skoudoro (approved):
Thanks @36000 for this,

merging

PR #2889, opened 2023-09-14
Apply pyupgrade suggestions

DimitriPapadopoulos:
* Inheritance from "object" is implicit in Python 3.
  See also Pylint's [useless-object-inheritance / R0205](https://pylint.readthedocs.io/en/latest/user_guide/messages/refactor/useless-object-inheritance.html).
* IOError is kept for compatibility with previous versions; starting from Python 3.3, it is an alias of OSError.
  https://docs.python.org/3/library/exceptions.html#IOError
* The default encoding for Python source code is UTF-8.
  https://docs.python.org/3/tutorial/interpreter.html#source-code-encoding


skoudoro (approved):
Looks good to me. Thank you @DimitriPapadopoulos.

I will go ahead and merge this as soon as the CI's come back green

skoudoro:
ok, all good, merging.

PR #2890, opened 2023-09-14
Attempt to fix error in conda jobs

DimitriPapadopoulos:
AttributeError: type object 'spmatrix' has no attribute '__div__'. Did you mean: '__dir__'?

https://github.com/OpenBB-finance/OpenBBTerminal/issues/5208

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 # Check against .travis.yml file and dipy/info.py
 cython>=0.29.24, !=0.29.29
 numpy>=1.22.4
-scipy>=1.8.1
+scipy=1.10.1
 nibabel>=4.0.0
 h5py>=3.7.0
 packaging>=19.0


requirements.txt
```diff
@@ -1,8 +1,8 @@
 # Check against .travis.yml file and dipy/info.py
 cython>=0.29.24, !=0.29.29
 numpy>=1.22.4
-scipy>=1.8.1
+scipy=1.10.1
```
skoudoro, line 4: Hi @DimitriPapadopoulos,

We do not want to pin `scipy` in `requirement.txt`. It should be flexible. 

However, it is ok to pin scipy or numpy in our github workflows. [look at this file](https://github.com/dipy/dipy/blob/master/.github/workflows/test_optional.yml).

I suppose the issue is coming from cvxpy version on conda.  cvxpy version on Pypi is 1.3.2 and accept new version of numpy/scipy.`cvxpy` version on our conda CI is 1.2.2 for whatever reason.  So either, we need to find a way to update this version OR pin `numpy/scipy` OR make sure to use conda-forge which have the right version (1.3.2)

skoudoro (changes requested):
Hi @DimitriPapadopoulos,

Thank you for this attempt. See below my comment.

requirements.txt
DimitriPapadopoulos: This was just a test, not a real solution. Unfortunately, it doesn't work, either because pinning down scipy here doesn't work, or because the error is different (unrelated to the version of scipy in cvxpy).

skoudoro: > Unfortunately, it doesn't work, either because pinning down scipy here doesn't work

yes, as you can see [here](https://github.com/dipy/dipy/actions/runs/6188638350/job/16801067780?pr=2890#step:6:154), scipy version is 1.11.1.  You need to pin it on the github worflows file (see my comment above).

You can also see [here](Unfortunately, it doesn't work, either because pinning down scipy here doesn't work) that we do not use the `requirements.txt` for this CI matrix.

PR #2891, opened 2023-09-14
Apply pyupgrade suggestions

DimitriPapadopoulos:
* In Python 3, string literals are Unicode by default.
* Use `yield from` instead of a loop with  `yield`.
* Delete empty 1st line.
* In `subprocess.check_output` arguments., replace `universal_newlines` with `text`.

~~I tried to replace `stdout=subprocess.PIPE`/`stderr=subprocess.PIPE` with `capture_output=True`, but it doesn't work. Yet `capture_output` is documented to be available in Python >= 3.7:
https://docs.python.org/3/library/subprocess.html#subprocess.run~~


arokem (approved):
LGTM!

PR #2897, opened 2023-09-15
BF: Bug when downloading hbn data.

arokem:
This is for the use-case where the user wants to define the output path. This wasn't working, because it always downloaded into the ~/.dipy path until this change.


skoudoro (approved):
Good catch, thanks ! merging

Issue #1853, opened 2019-06-12
Load seeds from trk (followup on #1792)


arokem:
In #1792, we discussed adding I/O options for the seeds (see https://github.com/nipy/dipy/pull/1792#issuecomment-477295813). 

To complete the picture, we need to implement something like: 

`save_tractogram(fname, streamlines, affine, seeds=seeds)

`streamlines, header, seeds = load_tractogram(filename, seeds=True)`

Here: https://github.com/nipy/dipy/blob/master/dipy/io/streamline.py

AntoineTheb:
On it !

skoudoro:
Hi @AntoineTheb,

Any update concerning this new feature? it would be good to have it before the future [Workshop in March](https://workshop.dipy.org).

Please, let us know! I tag this issue for the next release 1.2.

AntoineTheb:
Sure ! I must admit I had totally forgotten about this, but I'll get back on this in the coming days and this should be ready soon.

skoudoro:
ok, great! Thank you! looking forward to it!

AntoineTheb:
Sorry for the delay, things got in the way. Quick question @skoudoro @arokem: With the StatefulTractogram now implemented, should the seeds also change space and shift when `sft.to_space()` and `stf.to_center` is called ?

AntoineTheb:
Or, another thought, should I make other functions to save and load seeds to/from tractograms so as not to add extra parameters to `save/load_tractogram` ?

skoudoro:
> Or, another thought, should I make other functions to save and load seeds to/from tractograms so as not to add extra parameters to save/load_tractogram ?

Personally, I prefer this last option @AntoineTheb

skoudoro:
Hi @AntoineTheb,

Can you give us an update concerning this issue? Thank you

PR #2898, opened 2023-09-19
Remove zip operation in transform_tracking_output

CHrlS98:
This PR simply removes the zip operation in `transform_tracking_output` when `save_seeds` is enabled. Else, it is not possible to perform lazy saving of (streamline, seed) tuples using `nibabel LazyTractogram`.

However if anyone knows why it is there and/or how to lazily save streamlines and seeds using the actual implementation, this fix might not be required.

@gabknight @frheault

### DIFF
--- a/dipy/tracking/utils.py
+++ b/dipy/tracking/utils.py
@@ -853,7 +853,7 @@ def transform_tracking_output(tracking_output, affine, save_seeds=False):
 
     Parameters
     ----------
-    streamlines : Streamlines generator
+    tracking_output : Streamlines generator
         Either streamlines (list, ArraySequence) or a tuple with streamlines
         and seeds together
     affine : array (4, 4)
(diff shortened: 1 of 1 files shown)


skoudoro:
I am tagging also @AntoineTheb, #1792 #1853 just to keep track.

Maybe you can make it optional @CHrlS98 instead of completely remove it.  I do not have strong opinion about that just in case it is used and for backward compatibility

AntoineTheb:
From what I see, the zip removed at line 883 consumes the generator and forces streamlines and seeds to be loaded in memory. This prevents lazy tractograms down the line from actually being lazy. I don't think this breaks anything @skoudoro and can be removed without consequences.

arnaudbore (approved):

gabknight:
lgtm. thx @CHrlS98

skoudoro (approved):
ok, Thanks @CHrlS98, merging

PR #2902, opened 2023-09-22
Short import for horizon

maharshi-gor:
- This PR introduces the smaller import for horizon.

- Old import: `from dipy.viz.horizon.app import horizon`
- New import: `from dipy.viz import horizon`.

##### Note: The old import is still working. It should not break any existing imports.

A test is also included to test that both the import have same signature.

### DIFF
--- a/dipy/viz/__init__.py
+++ b/dipy/viz/__init__.py
@@ -1,6 +1,8 @@
 # Init file for visualization package
 import warnings
 
+from dipy.viz.horizon.app import horizon
+
 from dipy.utils.optpkg import optional_package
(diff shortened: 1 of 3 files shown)


skoudoro (commented):
Thank you for the API simplification. 

Can you update this line also : https://github.com/dipy/dipy/blob/f30d96d4d1a3b98d7aedf9531a439c6e0427f5f1/dipy/workflows/viz.py#L5.

Thank you!

skoudoro (approved):
Thanks @maharshi-gor! merging

PR #2904, opened 2023-09-24
Apply refurb suggestions

DimitriPapadopoulos:
Apply some [refurb](https://github.com/dosisod/refurb) suggestions.

I have omitted those that don't seem to improve readability, as well as suggestions that seem too intrusive and would be better left for later.


skoudoro (approved):
LGTM ! They all make sense. I am going ahead and merge this PR. Thanks @DimitriPapadopoulos 
It is very appreciated !  this cleaning / update is needed

DimitriPapadopoulos:
Note that `len(x) == 0` → `not x` is valid for Python arrays, but not for NumPy arrays, that's why _refurb_ applies this change to Python native objects only.

PR #2906, opened 2023-09-25
[FIX] Pin scipy for the conda failing CI's

skoudoro:
All in the Title. 

Our Conda CI's are failing due to `cvxpy` version  we have to pin scipy. 

Supersed #2890

### DIFF
--- a/.github/workflows/test_optional.yml
+++ b/.github/workflows/test_optional.yml
@@ -21,4 +21,5 @@ jobs:
     with:
       runs-on: '["macos-latest", "windows-latest"]'
       install-type: '["conda"]'
-      extra-depends: scikit-learn pandas statsmodels pytables scipy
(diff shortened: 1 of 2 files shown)


skoudoro:
ok, this is strange, all the dependencies are the same as https://github.com/dipy/dipy/actions/runs/5747080622/job/15577531014#step:6:59  which should be all greens but one CI's matrix is still failing.

still investigating... and if I do not find the reason, I will handle this error:
```python
FAILED reconst/tests/test_dti.py::test_restore - numpy.linalg.LinAlgError: Eigenvalues did not converge
```

skoudoro:
Yeah! All green, this is ready to go @arokem. Feel free to check and merge it when you can. Thanks

arokem:
Nice work!

Issue #2900, opened 2023-09-20
format → f-strings?


DimitriPapadopoulos:
## Description

Among other _pyugrade_ suggestions, changing `.format()` to [f-strings](https://docs.python.org/3/tutorial/inputoutput.html#tut-f-strings). Would you agree with a PR to apply such a change?

arokem:
I'd be 👍 for that. f-strings have been in Python since 3.6, so that shouldn't be a problem.

skoudoro:
+1 👍 , you can go ahead @DimitriPapadopoulos

PR #2910, opened 2023-09-26
STYLE: f-strings

DimitriPapadopoulos:
Change to f-strings where possible.

Batch **1**

Addresses #2900.

### DIFF
--- a/dipy/reconst/mapmri.py
+++ b/dipy/reconst/mapmri.py
@@ -237,22 +237,21 @@ class MapmriModel(ReconstModel, Cache):
         """
 
         if np.sum(gtab.b0s_mask) == 0:
-            msg = "gtab does not have any b0s, check in the gradient_table "
-            msg += "if b0_threshold needs to be increased."
-            raise ValueError(msg)
+            raise ValueError("gtab does not have any b0s, check in the "
+                             "gradient_table if b0_threshold needs to be "
+                             "increased.")
         self.gtab = gtab
 
         if radial_order < 0 or radial_order % 2:
-            msg = "radial_order must be a positive, even number."
-            raise ValueError(msg)
+            raise ValueError("radial_order must be a positive, even number.")
         self.radial_order = radial_order
 
         self.bval_threshold = bval_threshold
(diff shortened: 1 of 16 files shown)


skoudoro:
Hi @DimitriPapadopoulos,

Do you an update concerning this PR? Thanks!

DimitriPapadopoulos:
Also note that I have modified only a subset of the files. It will take some time and possibly multiple PRs to convert all files.

skoudoro:
> Also note that I have modified only a subset of the files. It will take some time and possibly multiple PRs to convert all files.

make sense and I am fine with splitting in multiple PR.

dipy/reconst/mapmri.py
```diff
@@ -267,21 +266,19 @@ def __init__(self,
                 raise ValueError('CVXPY package needed to enforce constraints.')
             if cvxpy_solver is not None:
                 if cvxpy_solver not in cvxpy.installed_solvers():
-                    msg = "Input `cvxpy_solver` was set to %s." % cvxpy_solver
-                    msg += " One of %s" % ', '.join(cvxpy.installed_solvers())
-                    msg += " was expected."
-                    raise ValueError(msg)
+                    installed_solvers = ', '.join(cvxpy.installed_solvers())
+                    raise ValueError(f"Input `cvxpy_solver` was set to"
```
skoudoro, line 270: I do not think the f-string is necessary at this specific line. but I agree it is necessary below. 

There are others cases like that in your change, but that fine, I will go ahead and merge it. 

Thanks !

skoudoro (approved):
LGTM, Thanks @DimitriPapadopoulos 

merging this first batch.

dipy/reconst/mapmri.py
DimitriPapadopoulos: You're right, it's not really necessary. [Horizontally align quotes in multi-line f-string breaks](https://github.com/psf/black/issues/2915) is a good summary of how I ended up adding `f` to lines that do not really need it, without being overly happy with it.

skoudoro: I see, thanks for the pointer!

PR #2912, opened 2023-09-26
MAINT: the symbol for second is s, not sec.

DimitriPapadopoulos:
https://www.bipm.org/en/si-base-units/second


skoudoro (approved):
I am ok with this change. Thank you @DimitriPapadopoulos.

What do you think @arokem ?

arokem:
+1

PR #2914, opened 2023-09-26
GitHub Actions

DimitriPapadopoulos:


.github/workflows/test_template.yml
```diff
@@ -107,7 +107,7 @@ jobs:
           tools/ci/run_tests.sh
         fi
     - name: Upload coverage to Codecov
-      uses: codecov/codecov-action@v3
+      uses: codecov/codecov-action@v4
```
skoudoro, line 110: Can you put it back to v3.   it seems v4 is still in a beta version following this link https://github.com/codecov/codecov-action

skoudoro (commented):
Thanks @DimitriPapadopoulos. 

See below my small comment to fix and then, it is ready to go.

.github/workflows/test_template.yml
DimitriPapadopoulos: Ah indeed. I think I had enabled `@v4`of this action in another project and then it was abruptly removed because of a major bug...

DimitriPapadopoulos: I had to force-pushed to keep the two commits. Hope it's OK as the changes are minimal.

skoudoro: No problem 👍 , thanks for the update

skoudoro (approved):
All good, thank you @DimitriPapadopoulos, merging

PR #2921, opened 2023-09-29
[FIX] tiny fix to HBN fetcher to also grab T1 for each subject

36000:
@arokem


36000:
@arokem i think this is ready to merge?

arokem:
I think that you are right. Merging.

PR #2926, opened 2023-10-05
DOC: Add MSMT auto response method `tol` parameter missing docstring

jhlegarreta:
Add MSMT auto response method `tol` parameter missing docstring.


dipy/reconst/mcsd.py
```diff
@@ -739,6 +739,8 @@ def auto_response_msmt(gtab, data, tol=20, roi_center=None, roi_radii=10,
     gtab : GradientTable
     data : ndarray
         diffusion data
+    tol : int
```
skoudoro, line 742: `int, optional` instead

You can also update the other parameters of this function and add `, optional` then, we can merge.

Thanks!

jhlegarreta: Done in 4d97d58.

skoudoro (approved):
LGTM, merging

skoudoro:
Thanks @jhlegarreta

Issue #2928, opened 2023-10-06
`test_icm_square` failing on and off


jhlegarreta:
## Description

`test_icm_square` is failing on and off, e.g.
https://github.com/dipy/dipy/actions/runs/6398397281/job/17368457748#step:9:3961

Looks like sometimes the difference is 0. So there is either a bug or some randomness in the process that need to be fixes.

FWIW, two consecutive local runs show that the result of `np.abs(np.sum(difference_map))` is not the same across runs.

## Way to reproduce

`$ pytest dipy/segment/tests/test_mrf.py`

DIPY `master`.
Probably affects to all OS, and not only the Python 3.11 version pointed by the build above. Regardless of the dependencies versions probably.

skoudoro:
Thank you for this report.

I agree, this is hard to reproduce locally but I am pretty sure this is due to a cast

jhlegarreta:
Might be the case that it is due to a cast, but results should be consistent across runs as well, isn't it?

skoudoro:
Do you have time to take the lead on that @pjsjongsung?  I am ok with the proposed solution.

pjsjongsung:
It shouldn't take too long. I can do that

jhlegarreta:
I'm fine with the result being 0 for a given random value, but this value should be fixed across runs,  the very same way we get the exact same result when fixing a random seed, if I am not missing something.

Changing the check of the test will not resolve the underlying issue (provided that the above assumption is correcta). Fixing it and adding the new additional check may be more appropriate IMO.

pjsjongsung:
It seems we can set a seed, which will be same across platform, but not across numpy versions.

However, it might still be stable enough, and all we would have to do is change the seed if it does not work on newer numpy version.

Alternatively, we could give a few more chances for test by setting a for loop with say 10 iters, and randomizing the number every loop. If the test does not take a lot of time, it won't be a bottleneck but still make the test much more stable.

pjsjongsung:
The ultimate problem is that there is a noticeable chance of the result being 0 than expected.

I am not entirely sure what the test is supposed to check, but if such comparison between two beta values are necessary, and want to completely remove any possibility of having a random assertion failure, we could also load the file with some random values that we know will pass the check.

I did a similar thing with the DL models using figshare.

jhlegarreta:
[...]

PR #2929, opened 2023-10-09
Stabilizing some test functions with set random seeds

pjsjongsung:
This pull request was created according to #2928 
I have added a decorator to fix numpy's random seed.
To use the decorator, I have moved the global variable in `dipy/segment/tests/test_mrf.py` to a function that creates it.

Are there any more changes I should make/revert? Also, do we want to add this decorator to every test function that has a numpy random function?

Thanks!

### DIFF
--- a/dipy/segment/tests/test_mrf.py
+++ b/dipy/segment/tests/test_mrf.py
@@ -5,6 +5,7 @@ from dipy.sims.voxel import add_noise
 from dipy.segment.mrf import (ConstantObservationModel,
                               IteratedConditionalModes)
 from dipy.segment.tissue import (TissueClassifierHMRF)
+from dipy.testing.decorators import set_random_seed
 
 
 # Load a coronal slice from a T1-weighted MRI
@@ -28,32 +29,42 @@ square[42:213, 42:213, :] = 1
 square[71:185, 71:185, :] = 2
 square[99:157, 99:157, :] = 3
 
-square_gauss = np.zeros((256, 256, 3)) + 0.001
-square_gauss = add_noise(square_gauss, 10000, 1, noise_type='gaussian')
-square_gauss[42:213, 42:213, :] = 1
-noise_1 = np.random.normal(1.001, 0.0001,
-                           size=square_gauss[42:213, 42:213, :].shape)
-square_gauss[42:213, 42:213, :] = square_gauss[42:213, 42:213, :] + noise_1
-square_gauss[71:185, 71:185, :] = 2
-noise_2 = np.random.normal(2.001, 0.0001,
-                           size=square_gauss[71:185, 71:185, :].shape)
-square_gauss[71:185, 71:185, :] = square_gauss[71:185, 71:185, :] + noise_2
-square_gauss[99:157, 99:157, :] = 3
-noise_3 = np.random.normal(3.001, 0.0001,
-                           size=square_gauss[99:157, 99:157, :].shape)
-square_gauss[99:157, 99:157, :] = square_gauss[99:157, 99:157, :] + noise_3
 
-square_1 = np.zeros((256, 256, 3)) + 0.001
-square_1 = add_noise(square_1, 10000, 1, noise_type='gaussian')
-temp_1 = np.random.randint(1, 21, size=(171, 171, 3))
-temp_1 = np.where(temp_1 < 20, 1, 3)
-square_1[42:213, 42:213, :] = temp_1
-temp_2 = np.random.randint(1, 21, size=(114, 114, 3))
-temp_2 = np.where(temp_2 < 19, 2, 1)
-square_1[71:185, 71:185, :] = temp_2
-temp_3 = np.random.randint(1, 21, size=(58, 58, 3))
-temp_3 = np.where(temp_3 < 20, 3, 1)
-square_1[99:157, 99:157, :] = temp_3
+@set_random_seed
+def create_square_uniform():
+    square_1 = np.zeros((256, 256, 3)) + 0.001
+    square_1 = add_noise(square_1, 10000, 1, noise_type='gaussian')
+    temp_1 = np.random.randint(1, 21, size=(171, 171, 3))
+    temp_1 = np.where(temp_1 < 20, 1, 3)
+    square_1[42:213, 42:213, :] = temp_1
+    temp_2 = np.random.randint(1, 21, size=(114, 114, 3))
+    temp_2 = np.where(temp_2 < 19, 2, 1)
+    square_1[71:185, 71:185, :] = temp_2
+    temp_3 = np.random.randint(1, 21, size=(58, 58, 3))
+    temp_3 = np.where(temp_3 < 20, 3, 1)
+    square_1[99:157, 99:157, :] = temp_3
+
+    return square_1
+
+
+@set_random_seed
+def create_square_gauss():
+    square_gauss = np.zeros((256, 256, 3)) + 0.001
+    square_gauss = add_noise(square_gauss, 10000, 1, noise_type='gaussian')
+    square_gauss[42:213, 42:213, :] = 1
+    noise_1 = np.random.normal(1.001, 0.0001,
+                            size=square_gauss[42:213, 42:213, :].shape)
+    square_gauss[42:213, 42:213, :] = square_gauss[42:213, 42:213, :] + noise_1
+    square_gauss[71:185, 71:185, :] = 2
+    noise_2 = np.random.normal(2.001, 0.0001,
+                            size=square_gauss[71:185, 71:185, :].shape)
+    square_gauss[71:185, 71:185, :] = square_gauss[71:185, 71:185, :] + noise_2
+    square_gauss[99:157, 99:157, :] = 3
+    noise_3 = np.random.normal(3.001, 0.0001,
+                            size=square_gauss[99:157, 99:157, :].shape)
+    square_gauss[99:157, 99:157, :] = square_gauss[99:157, 99:157, :] + noise_3
+
+    return square_gauss
 
 
 def test_grayscale_image():
@@ -231,6 +242,7 @@ def test_square_iter():
     icm = IteratedConditionalModes()
 
     initial_segmentation = square
+    square_gauss = create_square_gauss()
 
     mu, sigmasq = com.seg_stats(square_gauss, initial_segmentation,
                                 nclasses)
@@ -323,6 +335,7 @@ def test_icm_square():
     icm = IteratedConditionalModes()
 
     initial_segmentation = square
+    square_1 = create_square_uniform()
 
     mu, sigma = com.seg_stats(square_1, initial_segmentation,
                               nclasses)
--- a/dipy/testing/decorators.py
+++ b/dipy/testing/decorators.py
@@ -7,7 +7,7 @@ Decorators for dipy tests
 import re
 import os
 import platform
-
+import numpy as np
 
 SKIP_RE = re.compile("(\s*>>>.*?)(\s*)#\s*skip\s+if\s+(.*)$")
 
@@ -73,3 +73,18 @@ def xvfb_it(my_test):
     # Plant it back in and return the new function:
     test_with_xvfb.__name__ = fname
     return test_with_xvfb if not is_windows else my_test
+
+
+def set_random_seed(func):
+    """Decorator to set equal seeds.
+
+    This will help us avoid random errors
+    that occur with low probabilitiy during testing.
+
+    """
+    def test_with_seed(*args, **kwargs):
+        np.random.seed(0)
+        output = func(*args, **kwargs)
+        return output
+    
+    return test_with_seed


skoudoro:
Hi @pjsjongsung,

Thank you for taking care of this! I like the decorator approach. For sure there are other functions like the ones in tracking module but for now, it is ok. let's just focus on ICM.

dipy/testing/decorators.py
```diff
@@ -73,3 +73,18 @@ def test_with_xvfb(*args, **kwargs):
...
+
+def set_random_seed(func):
+    """Decorator to set equal seeds.
+
+    This will help us avoid random errors
+    that occur with low probabilitiy during testing.
+
+    """
+    def test_with_seed(*args, **kwargs):
+        np.random.seed(0)
```
skoudoro, line 86: instead of zero, I would recommend to put a number., whatever it is.... something like this year (`np.random.seed(20223)`)

dipy/testing/decorators.py
```diff
@@ -73,3 +73,18 @@ def test_with_xvfb(*args, **kwargs):
     # Plant it back in and return the new function:
     test_with_xvfb.__name__ = fname
     return test_with_xvfb if not is_windows else my_test
+
+
+def set_random_seed(func):
+    """Decorator to set equal seeds.
+
+    This will help us avoid random errors
+    that occur with low probabilitiy during testing.
```
jhlegarreta, line 82: The primary purpose of this is not to prevent errors, but to make results reproducible.

dipy/testing/decorators.py
```diff
@@ -73,3 +73,18 @@ def test_with_xvfb(*args, **kwargs):
     # Plant it back in and return the new function:
     test_with_xvfb.__name__ = fname
     return test_with_xvfb if not is_windows else my_test
+
+
+def set_random_seed(func):
+    """Decorator to set equal seeds.
```
jhlegarreta, line 79: Decorator to use a fixed value for the random generator seed.

dipy/testing/decorators.py
```diff
@@ -73,3 +73,18 @@ def test_with_xvfb(*args, **kwargs):
...
+
+    This will help us avoid random errors
+    that occur with low probabilitiy during testing.
+
+    """
+    def test_with_seed(*args, **kwargs):
+        np.random.seed(2023)
```
jhlegarreta, line 86: This has been deprecated in favor of:
```
rng = np.random.default_rng(<seed_value>)
rng.random(<value_count>)
```

so the decorator should be changed. Not sure how hard it is to make the `rng` variable be exposed to the testing method.

https://numpy.org/doc/stable/reference/random/generator.html
https://numpy.org/neps/nep-0019-rng-policy.html

dipy/testing/decorators.py
```diff
@@ -73,3 +73,18 @@ def test_with_xvfb(*args, **kwargs):
     # Plant it back in and return the new function:
     test_with_xvfb.__name__ = fname
     return test_with_xvfb if not is_windows else my_test
+
+
+def set_random_seed(func):
```
jhlegarreta, line 78: I would rename this to something like `fix_random_number_generator`, as the seed is fixed, and cannot be fixed from the outside. The name should then be made extensive to the inner method.

dipy/segment/tests/test_mrf.py
```diff
@@ -28,32 +29,42 @@
...
-square_1[42:213, 42:213, :] = temp_1
-temp_2 = np.random.randint(1, 21, size=(114, 114, 3))
-temp_2 = np.where(temp_2 < 19, 2, 1)
-square_1[71:185, 71:185, :] = temp_2
-temp_3 = np.random.randint(1, 21, size=(58, 58, 3))
-temp_3 = np.where(temp_3 < 20, 3, 1)
-square_1[99:157, 99:157, :] = temp_3
+
+@set_random_seed
+def create_square_uniform():
```
jhlegarreta, line 34: Why is this change part of the commit?

jhlegarreta (changes requested):
Thanks for giving this a try @pjsjongsung. A few comments inline.

dipy/segment/tests/test_mrf.py
pjsjongsung: Because I had to add a decorator when creating the global variables. Adding the decorator to the test functions will be too late. Before, it was just defining them globally, which I do not think I can add a decorator to it.

jhlegarreta: OK, then please explain in the commit message. Thanks Jong Sung.

dipy/testing/decorators.py
pjsjongsung: Yeah I understand it is deprecated, but as you said I am not sure how to use the decorator with the rng variable. Alternatively, we could just change each random function individually.

jhlegarreta: The risk of using a deprecated value is the removal in a future `NumPy` version, so better IMO it is better to stick to the current convention now.

skoudoro: I did not know that `np.random.seed` has been deprecated. Thank you for the info @jhlegarreta .

We need to update that in the whole code based. I prefer the new standard.

pjsjongsung:
I have changed the icm tests to the random generator. If this looks ok, we could change all random functions accordingly?

dipy/segment/tests/test_mrf.py
```diff
@@ -28,32 +29,47 @@
...
-temp_1 = np.where(temp_1 < 20, 1, 3)
-square_1[42:213, 42:213, :] = temp_1
-temp_2 = np.random.randint(1, 21, size=(114, 114, 3))
-temp_2 = np.where(temp_2 < 19, 2, 1)
-square_1[71:185, 71:185, :] = temp_2
-temp_3 = np.random.randint(1, 21, size=(58, 58, 3))
-temp_3 = np.where(temp_3 < 20, 3, 1)
-square_1[99:157, 99:157, :] = temp_3
+# Setting seed value
+seed_v = 1234
```
jhlegarreta, line 33: Is this needed?

pjsjongsung: Meant to remove that. Thank you!

pjsjongsung:
It seems to be disrupting assertions of various tests using np.random.seed. I will try changing all of them to generators, and if that doesn't work, we should not change this until we figure out what the reason for the assertion errors are.

dipy/testing/decorators.py
```diff
@@ -73,3 +73,18 @@ def test_with_xvfb(*args, **kwargs):
...
     return test_with_xvfb if not is_windows else my_test
+
+
+def fix_random_number_generator(func):
+    """Decorator to use a fixed value for the random generator seed.
+
+    This will make the tests that use random functions reproducible.
+
+    """
+    def wrapper_with_fixed_seed(*args, **kwargs):
```
jhlegarreta, line 84: I'd change this name so that it more closely follows the callable one.

jhlegarreta (commented):
Thanks for the effort: I would split this into two/three commits so that it becomes crystal clear the purpose, benefit, effects of each commit (otherwise, it becomes hard to find and explain things when the changes are revisited in the future):
- The first one would transition to the current NumPy recommended way to use the random generator across the whole code base.
- The decorator is added and used across all tests (could be done together with the first commit)
- A last commit would try to address the ICM test failure/instability.

pjsjongsung:
I am working on the new first commit, which is transitioning everything to a numpy recommended way of random generator. However, there seems to be tests that were only working on particular seeds, and is now disrupted by the different random values. (Even with the same seed, np.random.seed and random generator have different results). I will keep looking at what the reasons might be, but it seems it is not an easy solution.

For now, I will create a separate PR for test stability of ICM.

jhlegarreta:
Re https://github.com/dipy/dipy/pull/2929#issuecomment-1768504561

> (Even with the same seed, np.random.seed and random generator have different results)

Maybe this is expected/documented somewhere or can be asked to the NumPy folks.

Thanks for effort @pjsjongsung.

pjsjongsung:
> Re [#2929 (comment)](https://github.com/dipy/dipy/pull/2929#issuecomment-1768504561)
> 
> > (Even with the same seed, np.random.seed and random generator have different results)
> 
> Maybe this is expected/documented somewhere or can be asked to the NumPy folks.
It seems they have changed Bit Generator that is used to create random values. So until they change that again, (which I believe  is far away since they changed it only once ever) the seed values should be stable enough.

pjsjongsung:
It seems all the tests related to the files I changed are passing.

jhlegarreta (changes requested):
> It seems they have changed Bit Generator that is used to create random values. So until they change that again, (which I believe is far away since they changed it only once ever) the seed values should be stable enough.

OK. Thanks for the investigation Jong Sung.

> It seems all the tests related to the files I changed are passing.

:ok_hand: Just curious: then, why were you reporting that some tests were failing in https://github.com/dipy/dipy/pull/2929#issuecomment-1753653282? Maybe because the seed was set to a different value wrt the previous one?

> fix_random_number_generator

I now realize that it is not the best name (my bad, as I proposed it): can it be changed to `set_random_number_generator`, the private method just be `_set_random_number_generator`, and its descendant be `_set_random_number_generator_wrapper`?

Please, use a single commit: `TEST: Changing numpy random usage in tests to random generator` with the meaningful commit body message you wrote, the rest of the commit are intermediate work and are not informative when navigating the git history. You will need to squash all commits, and push force.

pjsjongsung:
> 👌 Just curious: then, why were you reporting that some tests were failing in [#2929 (comment)](https://github.com/dipy/dipy/pull/2929#issuecomment-1753653282)? Maybe because the seed was set to a different value wrt the previous one?

So `numpy.random.seed` seems to use a different random number generator than `numpy.random.Generator`. Thus even with the same seed, completely different values were being created. So there needed to be some searching on new seed values that worked.
 
> > fix_random_number_generator
> 
> I now realize that it is not the best name (my bad, as I proposed it): can it be changed to `set_random_number_generator`, the private method just be `_set_random_number_generator`, and its descendant be `_set_random_number_generator_wrapper`?
> 

> Please, use a single commit: `TEST: Changing numpy random usage in tests to random generator` with the meaningful commit body message you wrote, the rest of the commit are intermediate work and are not informative when navigating the git history. You will need to squash all commits, and push force.

Will do.

jhlegarreta:
> So numpy.random.seed seems to use a different random number generator than numpy.random.Generator. Thus even with the same seed, completely different values were being created. So there needed to be some searching on new seed values that worked.

In a quick inspection I saw the same seeds being used/no baselines updated. So if you had to look for new seeds, among the infinite values, how did you find the right values in a finite amount of time :thinking:?

pjsjongsung:
> > So numpy.random.seed seems to use a different random number generator than numpy.random.Generator. Thus even with the same seed, completely different values were being created. So there needed to be some searching on new seed values that worked.
> 
> In a quick inspection I saw the same seeds being used/no baselines updated. So if you had to look for new seeds, among the infinite values, how did you find the right values in a finite amount of time 🤔?

Most of the tests were stable enough to work without change. Others, I had to randomly search a few times, but it worked out. I just ran a for loop with different seed values until it worked :)

pjsjongsung:
Commit squashed and there seems to be no errors from this PR!

jhlegarreta:
:ok_hand: Beautiful @pjsjongsung. Some failing tests seem to be related to the use of random generators, although they were not modified in this patch set:
https://github.com/dipy/dipy/actions/runs/6594238121/job/17917878390?pr=2929#step:9:4397

Do you have an intuition about these?

pjsjongsung:
@jhlegarreta It must be connected to the change in using seeds. The change in seeds are global, so even if the function was not specifically using the seed within the function, it would have been benefiting from a fixed number. (Is my guess)

I have changed the tests that failed and while doing that have changed some other functions that were using some kind of numpy random function but was not using a fixed seed generator.

However, once this PR is merged, I will create another PR to change every numpy random functions used in tests (even if it was not using a seed before) to use some kind of seed.

pjsjongsung:
I think the PR is good to go? The mlp test function is failing for some reason, which does not even use any type of random function. My guess is the change in seeds changed something in the order of the training data being loaded. (IMO it is strange to have a test function on the training accuracy of a DL model after 1 epoch)

jhlegarreta:
> I think the PR is good to go? The mlp test function is failing for some reason, which does not even use any type of random function. My guess is the change in seeds changed something in the order of the training data being loaded.

Have seen the MLP failure; did not spot where the randomness is involved, other than the possible weights, probably internally/out of the sight of the end user. If the seeds were/are being set somehow globally, this may affect the test at issue.

>  (IMO it is strange to have a test function on the training accuracy of a DL model after 1 epoch)

Agreed.

Thoughts:
- Change the test to test only that the model can be instantiated (in a separate commit and/or PR).
- Relax the accuracy tolerance.
- Training an MLP that is not large with simple data should not be that long, so increasing the number of epochs may be reasonable to find a better accuracy.

pjsjongsung:
I have increased the number of epochs to 5 as @jhlegarreta suggested, so it should be more stable than before.

jhlegarreta:
9afe970 Commit message subject prefix, please.

pjsjongsung:
Sorry, my bad. Still getting used to it. Corrected

jhlegarreta (approved):
Build failure is unrelated.

Thanks for this @pjsjongsung.

skoudoro (approved):
Thank you for this huge work @pjsjongsung and Thank you for reviewing @jhlegarreta .

Looks good to me too. merging

PR #2932, opened 2023-10-12
OPT - Optimized pmfgen

gabknight:
This PR cythonize pmfgen.

- Optimized pmfgen object. On my tests, this 2X Det, Prob and PTT DG tracking.
- Added a new optional parameter `sh_to_pmf=False` to `PmfGenDirectionGetter.from_shcoeff(.)`. If `True`, it speeds up tracking when using sh coefficients, but uses more memory. I updated the examples to set `sh_to_pmf=True`


gabknight:
The parameter `sh_to_pmf` of `PmfGenDirectionGetter.from_shcoeff()` is `False` by default to not change behaviors. If `True` it speeds up det/prob tracking by 2X and PTT by 10X. I updated the examples to set `sh_to_pmf=True`.

It should be good to go if tests succeed.

skoudoro (approved):
Thanks you for the speedup @gabknight!

I do not have specific comment since I followed this work closely. It looks good. 

is there any comments on this PR? If not, I will go ahead and merge it on Thursday 26 afternoon ET timezone.

skoudoro:
Thank you @gabknight for this work. merging

PR #2933, opened 2023-10-13
fixed bug with fit extra returns

samcoveney:
Fixed a small bug in the iteration wrapper for fitting, introduced after #2730 allowing fitting functions to return more than just the parameters.

### DIFF
--- a/dipy/reconst/dti.py
+++ b/dipy/reconst/dti.py
@@ -1352,7 +1352,7 @@ def iter_fit_tensor(step=1e4):
                 if extra_i is not None:
                     for key in extra_i:
                         if i == 0: extra[key] = np.empty(data.shape)
-                        extra[key][i:i + step] = extra_i
+                        extra[key][i:i + step] = extra_i[key]
                 else:
                     if i == 0: extra = None
 


jhlegarreta:
Thanks for doing this @samcoveney. A test should be added so that this does not happen again.

samcoveney:
The problem is that most of the tests for DiPy are so small that they don't trigger the use of the iterative decorators for the fitting routines... so potentially would need a bigger fix.

jhlegarreta:
> The problem is that most of the tests for DiPy are so small that they don't trigger the use of the iterative decorators for the fitting routines... so potentially would need a bigger fix.

Did not have a look at the test module that is related to this, but nothing prevents us from writing other tests as needed, isn't it?

samcoveney:
You are right @jhlegarreta we can write additional checks if needed on a case by case basis. At the moment, the only fitting routine making use of extra returns is the RESTORE algorithm, which does not use the iterative wrapper. This is partly why this was missed in testing.

I have a branch I am working on that has other fitting functions returning extra useful information, which is why I discovered the mistake. And until I make the pull request adding this extra functionality, there isn't really an ideal way to write these tests at the moment. I'll put these tests in that pull request.

jhlegarreta:
Sounds good. Thanks for the effort @samcoveney.

samcoveney:
Thanks @jhlegarreta for your feedback, and thanks @skoudoro

PR #2938, opened 2023-10-18
TEST: Adding random generator with seed to icm tests

pjsjongsung:
This commit adds seeds to test_square_iter and test_icm_square to stabilize the tests. square_gauss and square_1, which were originally global variables are now loaded through a function which has a numpy random generator inside with fixed seed. The variables were wrapped in a function because 1. there is no need for the variables to be global 2. this complies with a future PR that will add decorators on tests that require seed values.

Note that this is a separate PR created from a discussion in #2929 for immediate increase in stability of the tests while a PR to change all random functions in DIPY to the recent numpy convention is underway.

### DIFF
--- a/dipy/segment/tests/test_mrf.py
+++ b/dipy/segment/tests/test_mrf.py
@@ -28,32 +28,42 @@ square[42:213, 42:213, :] = 1
 square[71:185, 71:185, :] = 2
 square[99:157, 99:157, :] = 3
 
-square_gauss = np.zeros((256, 256, 3)) + 0.001
-square_gauss = add_noise(square_gauss, 10000, 1, noise_type='gaussian')
-square_gauss[42:213, 42:213, :] = 1
-noise_1 = np.random.normal(1.001, 0.0001,
-                           size=square_gauss[42:213, 42:213, :].shape)
-square_gauss[42:213, 42:213, :] = square_gauss[42:213, 42:213, :] + noise_1
-square_gauss[71:185, 71:185, :] = 2
-noise_2 = np.random.normal(2.001, 0.0001,
-                           size=square_gauss[71:185, 71:185, :].shape)
-square_gauss[71:185, 71:185, :] = square_gauss[71:185, 71:185, :] + noise_2
-square_gauss[99:157, 99:157, :] = 3
-noise_3 = np.random.normal(3.001, 0.0001,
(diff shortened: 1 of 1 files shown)


skoudoro (commented):
Looks good to me. Thanks @pjsjongsung 

> 1. there is no need for the variables to be global 

I agree with this statement, I do not know why they were global. Actually, I should check the whole test codebase to reduce this.

Waiting for the CI's  to finish and if it is ok, I will go ahead. 

> Note that this is a separate PR created from a discussion in https://github.com/dipy/dipy/pull/2929 for immediate increase in stability of the tests while a PR to change all random functions in DIPY to the recent numpy convention is underway.

Instead of a big PR, maybe you should do it step by step, small PRs by small PRs. You could create an issue to identify all np.random that needs to be replace in multiple PR. 
But if you feel you do it all in once, please, feel free to go ahead.

pjsjongsung:
Thank you @skoudoro 

I think I will push one PR because even though the changes are in multiple files, it is just a few lines per file. I feel like it will create too much small PRs.

jhlegarreta (approved):
Thanks for doing this @pjsjongsung.

> Instead of a big PR, maybe you should do it step by step, small PRs by small PRs. You could create an issue to identify all np.random that needs to be replace in multiple PR.
> But if you feel you do it all in once, please, feel free to go ahead.

I don't think splitting this across multiple PRs is the best approach.

All methods that need to be changed can be identified and changed easily (automatically, or almost) I believe (despite involving a given amount of work), so best is to do all at once.

> I think I will push one PR because even though the changes are in multiple files, it is just a few lines per file. I feel like it will create too much small PRs.

:+1:.

pjsjongsung:
Sorry I missed that. I will make sure to do that in the later commits.

jhlegarreta:
> @pjsjongsung one thing I missed: please use the appropriate commit message subject codes.

@pjsjongsung not only the PR title, but the commit itself: you'll need to:
```
$ git commit --amend
(change the commit message subject)
$ git push -f origin icm_test_stable
```

pjsjongsung:
Oh I was afraid it might disrupt the commit. Done! Thanks

skoudoro:
All green, great !  merging, thanks. @pjsjongsung and @jhlegarreta !

PR #2939, opened 2023-10-18
DOC: Update jhlegarreta affiliation in developers

jhlegarreta:
Update jhlegarreta affiliation in developers.


skoudoro:
Thanks !

PR #2942, opened 2023-10-18
OPT: Delegate to NumPy creating a random matrix

jhlegarreta:
Delegate to NumPy creating a random matrix, instead of creating each row independently using a list comprehension.

Additionally, fixes:
```
Local variable 'si' value is not used
```


arokem (approved):
LGTM. Thanks for doing this!

skoudoro (approved):
LGTM, waiting for. CI's before merging

skoudoro:
Thanks @jhlegarreta !

PR #2943, opened 2023-10-18
BF: Fix bundlewarp shape analysis profile values for all `False` mask

jhlegarreta:
Fix bundlewarp shape analysis profile values for the case where all mask values are `False`: set the value to `np.nan` instead of relying on NumPy to compute the mean and the std values on an empty array.

Fixes:
```
align/tests/test_streamwarp.py::test_bundle_shape_profile
  /home/runner/work/dipy/dipy/venv/lib/python3.11/site-packages/numpy/core/fromnumeric.py:3504:
RuntimeWarning: Mean of empty slice.
    return _methods._mean(a, axis=axis, dtype=dtype,

align/tests/test_streamwarp.py::test_bundle_shape_profile
  /home/runner/work/dipy/dipy/venv/lib/python3.11/site-packages/numpy/core/_methods.py:129:
RuntimeWarning: invalid value encountered in scalar divide
    ret = ret.dtype.type(ret / rcount)

align/tests/test_streamwarp.py::test_bundle_shape_profile
  /home/runner/work/dipy/dipy/venv/lib/python3.11/site-packages/numpy/core/_methods.py:206:
RuntimeWarning: Degrees of freedom <= 0 for slice
    ret = _var(a, axis=axis, dtype=dtype, out=out, ddof=ddof,

align/tests/test_streamwarp.py::test_bundle_shape_profile
  /home/runner/work/dipy/dipy/venv/lib/python3.11/site-packages/numpy/core/_methods.py:163:
RuntimeWarning: invalid value encountered in divide
    arrmean = um.true_divide(arrmean, div, out=arrmean,

align/tests/test_streamwarp.py::test_bundle_shape_profile
  /home/runner/work/dipy/dipy/venv/lib/python3.11/site-packages/numpy/core/_methods.py:198:
RuntimeWarning: invalid value encountered in scalar divide
    ret = ret.dtype.type(ret / rcount)
```

Raised for example at:
https://github.com/dipy/dipy/actions/runs/6565543562/job/17834291163?pr=2938#step:9:4645

### DIFF
--- a/dipy/align/streamwarp.py
+++ b/dipy/align/streamwarp.py
@@ -238,6 +238,15 @@ def bundlewarp_shape_analysis(moving_aligned, deformed_bundle, no_disks=10,
     Bundle shape difference analysis using magnitude from BundleWarp
     displacements and BUAN.
 
+    Depending on the number of points of a streamline, and the number of
+    segments requested, multiple points may be considered for the computation
+    of a given segment; a segment may contain information from a single point;
+    or some segments may not contain information from any points. In the latter
+    case, the segment will contain an ``np.nan`` value. The point-to-segment
+    mapping is defined by the :func:`assignment_map`: for each segment index,
+    the point information of the matching index positions, as returned by
+    :func:`assignment_map`, are considered for the computation.
+
     Parameters
     ----------
     moving_aligned : Streamlines
@@ -279,8 +288,13 @@ def bundlewarp_shape_analysis(moving_aligned, deformed_bundle, no_disks=10,
 
     for i in range(n):
 
-        shape_profile[i] = np.mean(offsets[indx == i])
-        stdv[i] = np.std(offsets[indx == i])
+        mask = indx == i
+        if sum(mask):
+            shape_profile[i] = np.mean(offsets[mask])
+            stdv[i] = np.std(offsets[mask])
+        else:
+            shape_profile[i] = np.nan
+            stdv[i] = np.nan
 
     if plotting:
         bundle_shape_profile(x, shape_profile, stdv)


dipy/align/streamwarp.py
```diff
@@ -278,8 +278,13 @@ def bundlewarp_shape_analysis(moving_aligned, deformed_bundle, no_disks=10,
 
     for i in range(n):
 
-        shape_profile[i] = np.mean(offsets[indx == i])
-        stdv[i] = np.std(offsets[indx == i])
+        mask = indx == i
+        if sum(mask):
```
jhlegarreta, line 282: I am not sure if all values being `False` is something that can be expected. It looks weird to me.

dipy/align/streamwarp.py
```diff
@@ -278,8 +278,13 @@ def bundlewarp_shape_analysis(moving_aligned, deformed_bundle, no_disks=10,
...
+            shape_profile[i] = np.mean(offsets[mask])
+            stdv[i] = np.std(offsets[mask])
+        else:
+            shape_profile[i] = np.nan  # or pass and leave it to zero ?
```
jhlegarreta, line 286: If the case makes sense, should these be left to their original zero values, and use a simple `pass` instead?

jhlegarreta:
The warnings are gone:
https://github.com/dipy/dipy/actions/runs/6567803665/job/17841150778?pr=2943#step:9:4655

dipy/align/streamwarp.py
skoudoro: @BramshQamar, can you answer @jhlegarreta  concerning this point.

dipy/align/streamwarp.py
skoudoro: @BramshQamar, can you answer @jhlegarreta  concerning this point.

dipy/align/streamwarp.py
BramshQamar: Hi @jhlegarreta, good catch!

You are right, and it shouldn't return all `False`. I see there is a logical problem in the test function. I set the number of points to 20 in each streamlines https://github.com/dipy/dipy/blob/d2a05444f318cbe1397d492fc83bbd3b898da979/dipy/align/tests/test_streamwarp.py#L62  
and then in the second test case, I am setting n=100, which means 100 segments to be created along the length of the bundle. Since streamlines only have 20 points in them, some segments will be empty (i.e. actual data points are less than the segments).
https://github.com/dipy/dipy/blob/d2a05444f318cbe1397d492fc83bbd3b898da979/dipy/align/tests/test_streamwarp.py#L77

We could make changes in the `def bundlewarp_shape_analysis` function to handle this case and can have another test for it.

I suppose keeping the mask condition should be fine even if the number of points in a streamline is greater or equal to segment/disk numbers.

dipy/align/streamwarp.py
BramshQamar: @jhlegarreta @skoudoro 
We should keep this line because 0 values imply minimum to no shape differences in the two bundles' segments. This would be the case when looking at shape differences between two identical bundles.

dipy/align/streamwarp.py
jhlegarreta: > We could make changes in the def bundlewarp_shape_analysis function to handle this case and can have another test for it.
> I suppose keeping the mask condition should be fine even if the number of points in a streamline is greater or equal to segment/disk numbers.

I am not sure if I see this: if streamlines have N vertices, but you request an analysis over 100, then streamlines should be resampled within this function, regardless of N. Not sure if following this rationale has undesired side-effects.

dipy/align/streamwarp.py
jhlegarreta: The expected values of the analysis should be explained in the docstring: what a 0 means, what an `np.nan` means, and what any value in between means.

Kept `np.nan`  and removed the temporary comment in 48a6110.

dipy/align/streamwarp.py
BramshQamar: @jhlegarreta We ideally would like to keep the original number of points in a streamline. And we allow multiple points from the same streamline to belong to one segment. We don't expect one point per segment from each streamline. We also allow for a streamline not to have any points belonging to some segments (streamlines of different lengths are not sampled to have the same N number of points). Since this is a test function, I set them to have 20, and passing 100 segment numbers caused an issue here.

jhlegarreta: So then your proposal is to set them to 20 in the test? How do you propose to handle the situation of passing 100: should the method raise an exception or do you have another proposal?

BramshQamar:
Hi @jhlegarreta 
Sorry, I was on break for some time. There are two ways:

Easy solution: change `n` to `20` segments, so we don't get errors. Or we check in the `test_bundle_shape_profile` function; if `shape_profile` has more than `k ` NaN values, we raise an error or warning to let the user know `n `is not an optimal number of segments for their data where k can be defined as `k = n*0.2`. If 20% of the segments are empty, a warning/error is raised.

Let me know if you have any other suggestions to solve this.

Thank you,
Bramsh

jhlegarreta:
- Changing it to `20` does not solve the issue: we want to test all possible cases, because users will certainly use such cases.
- Raising a warning/error in the caller function (the test) is not useful: we want the method to be self-contained, and to manage all possible scenarios if applicable.

Anyways, I think I have finally figured out that all this stems from the lack of proper documentation (https://github.com/dipy/dipy/pull/2943#discussion_r1375276688), so I have added some documentation that hopefully makes it clear for the users what an `np.nan` means in the result and how the point-to-segments binning is done.

jhlegarreta:
`Optional Deps / pip / build (3.10, macos-latest, pip)` failure is unrelated

skoudoro:
thank you for adding the documentation. it looks good to me. I will go ahead and merge this PR. Thanks @jhlegarreta and thanks @BramshQamar for the review.

PR #2945, opened 2023-10-20
DOC: Fix miscellaneous docstrings

jhlegarreta:
- DOC: Fix parameter name in `msdki` function
- DOC: Fix return parameter names in miscellaneous `msdki` functions


skoudoro (approved):
All good, thanks @jhlegarreta, merging

PR #2947, opened 2023-10-20
RF - BootDirectionGetter

gabknight:
This PR prepares `PmfGen` cython optimization.

- Move `bootPmfGen` functionalities from `dipy.direction.pmf` to `dipy.direction.bootstrap_direction_getter.pyx`
- Change `BootDirectionGetter` parent class from `PmfGenDirectionGetter` to `DirectionGetter`
- Remove `bootPmfGen`  class from `dipy.direction.pmf`
- Update tests

There are no noticeable performance impacts.


skoudoro (commented):
Great! thank you for refactoring this!  I like the simplification and the gain of speed. 

it looks good to me for now. Can you check that we do not break any examples ?

Thank you

gabknight:
It did break the example! The data array was missing a `dtype` check/transformation to float. I also added an optional `b_tol` parameter for more control.

skoudoro:
Thank you for this work @gabknight , merging

Issue #2803, opened 2023-05-04
Type annotations integration


skoudoro:
## Description

it would be great if we could start to integrate type annotations in the whole DIPY codebase. 
Benefits are multiples (Improve code readibility so easier to maintain/debug, review and reduce cognitive load) and this summer can be an opportunity for that.

it would be great to know if someone is against this task.

Thank you all for the future feedback

Modules to to do
-----------------

- [ ] align
- [ ] boots
- [ ] core
- [ ] data
- [ ] denoise
- [ ] direction
- [ ] io
- [ ] nn
- [ ] reconst
- [ ] segment
- [ ] sims
- [ ] stats
- [ ] testing
- [ ] tests
- [ ] tracking
- [ ] utils
- [ ] viz
- [ ] workflows

skoudoro:
Great @WassCodeur !

You should start with `io` modules of the project. 

You start by annotating the basic types, like `str`, `bool`. 

Also, I suppose there is library that can do the work for you, you will just need to check if it make sense. Let me know if you have any questions !

WassCodeur:
sure

skoudoro:
After the last DIPY Open Meeting, we decided to wait and see the evolution of type annotation adoption.

So closing this issue for now. 

Sorry @WassCodeur, I will have to close your PR also. Thanks for your nice work.

PR #2975, opened 2023-11-16
connectivity_matrix code speed up

DelinteNicolas:
Following discussion #2972 , here is a pull request with the new code. The main objective was to speed up the code when 'inclusive' was set to True.

For a tractogram of 500 000 streamlines, the code now runs in less than 20s with all parameters set to True, compared to more than 2 minutes previously. The code is also shorter and subjectively more readable.

Main differences:
- 'inclusive' is now faster, with a x3 speed up for a tractogram of 2000 streamlines and >x6 speed up for a tractogram of 500 000 streamlines. Speed when 'inclusive' is set to False is unchanged.
- 'symmetric' now returns only the top half of the matrix when set to False. The results thus differ from the previous implementation when set to False. To be honest, 'symmetric'=False does not make sense to me since streamlines are direction agnostic. Feel free to change this behavior.
- 'return_mapping' is identical
- 'mapping_as_streamlines' was slightly changed. The results are identical but the function now returns a 'mapping_as_streamlines' map even when 'return_mapping' is set to False. Feel free to also change this behavior. It would make sense to me to change the parameter name to 'return_mapping_as_streamlines' to make the behavior more explicit. 

The results were only tested with a few examples, more testing should be performed to ensure reproducibility between both implementations.

### DIFF
--- a/dipy/tracking/utils.py
+++ b/dipy/tracking/utils.py
@@ -107,11 +107,12 @@ def density_map(streamlines, affine, vol_dims):
     return counts
 
 
-def connectivity_matrix(streamlines, affine, label_volume,
-                        inclusive: bool = True, symmetric: bool = True,
-                        return_mapping: bool = False,
-                        mapping_as_streamlines: bool = False):
-    """Count the streamlines that start and end at each label pair.
+def connectivity_matrix_mod2(streamlines, affine, label_volume,
+                             inclusive: bool = False, symmetric: bool = True,
+                             return_mapping: bool = False,
+                             mapping_as_streamlines: bool = False):
+    '''
+    Count the streamlines that start and end at each label pair.
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


skoudoro:
Thank you for this @DelinteNicolas!  Not sure yet why the CI's are failing. it seems that your news implementation does not provide identical results as the previous implementation since tests are failing. 

I will look why a bit later.

Thanks !

DelinteNicolas:
Hello, I have modified the pull request to simplify the changes you would want to make. The implementation now has the same behavior as the previous implementation for the 'symmetric' and 'mapping_as_streamlines' parameters.

Main changes:
- changed matrix type to int64 to fit with previous implementation
- removed np.floor and astype(int) which should be handled by _to_voxel_coordinates function
- changed behavior of mapping_as_streamlines to work similarly the previous implementation
- inclusive default set to False
- symmetry matrix has been changed to fit with previous behavior. It no longer returns the top half of the symmetric matrix
- mapping incorrectly returned streamlines endpoints when mapping_as_streamlines as True. This has been fixed

The tests should now be correct but **/!\ results still differ when inclusive is set to True and symmetric to False**

Hope this helps

skoudoro (approved):
LGTM. Thank you @DelinteNicolas. 

I will wait until Tuesday if there is another reviewer. If not, I will go ahead and merge it.

arokem:
This looks fine to me. I just noticed the type annotations implemented here. I think that these annotations are pretty harmless, but just wanted to make sure the this is not in conflict with our decision on #2803?

skoudoro:
> I think that these annotations are pretty harmless, but just wanted to make sure the this is not in conflict with our decision on https://github.com/dipy/dipy/issues/2803?

I agree, they are harmless, but they are conflicting with our decision. I will remove them now and merge this PR.

Thank you for the feedback @arokem and thank you for this work @DelinteNicolas

PR #2976, opened 2023-11-20
[MTN] Update index url for PRE-Wheels dependencies

skoudoro:
This is a small PR to use the new index URL for the pre-wheel. 

We want to anticipate Numpy 2.0


skoudoro:
The pre wheels does not fail with Numpy 2.0.dev, it is a great news for our codebase. Just more warnings appear for now. 

Now, I will update with python 3.12 and if all ok, I will go ahead and merge it.

skoudoro:
ok, it does not work with python 3.12 yet. I will investigate later. removing my last commit for now. 

Also, it seems that numpy 2.0.dev is installed, but a dependency removed this installation and force numpy 1.26.2.  That's why it was working something to investigate also.

skoudoro:
For now, I merge this PR. I will tackle the numpy issue on another one.

PR #2978, opened 2023-11-28
DOC: Fix issue template [ci skip]

larsoner:
The interior quotes either need to be escaped or changed to single quotes, the latter seemed simpler


larsoner:
... otherwise you get:
```
$ python -c "import sys; print("Python", sys.version)"
  File "<string>", line 1, in <module>
NameError: name 'Python' is not defined
```

Issue #2979, opened 2023-11-28
Prerelease wheels not NumPy 2.0.0.dev compatible


larsoner:
## Description

Dipy prerelease wheels at https://pypi.anaconda.org/scipy-wheels-nightly/simple are not compatible with the NumPy 2.0 wheels at https://pypi.anaconda.org/scientific-python-nightly-wheels/simple, so downstream libs like MNE-Python have to choose which prerelease dependency to test against (or split into multiple runs).

## Way to reproduce
[If reporting a bug, please include the following important information:]
- [x] Code example

```
$ pip install --upgrade --pre --extra-index-url "https://pypi.anaconda.org/scientific-python-nightly-wheels/simple" "numpy>=2.0.0.dev0" --extra-index-url "https://pypi.anaconda.org/scipy-wheels-nightly/simple" dipy
...
Successfully installed dipy-1.8.0.dev0 numpy-2.0.0.dev0
$ python -c "import dipy.align"
```

<details>
<summary>Traceback</summary>

```
$ python -c "import dipy.align"

A module that was compiled using NumPy 1.x cannot be run in
NumPy 2.0.0.dev0 as it may crash. To support both 1.x and 2.x
versions of NumPy, modules must be compiled against NumPy 2.0.

If you are a user of the module, the easiest solution will be to
either downgrade NumPy or update the failing module (if available).

NOTE: When testing against pre-release versions of NumPy 2.0
or building nightly wheels for it, it is necessary to ensure
the NumPy pre-release is used at build time.
The main way to ensure this is using no build isolation
and installing dependencies manually with NumPy.
For cibuildwheel for example, this may be achieved by using
the flag to pip:
    CIBW_BUILD_FRONTEND: pip; args: --no-build-isolation
installing NumPy with:
    pip install --pre --extra-index-url https://pypi.anaconda.org/scientific-python-nightly-wheels/simple
in the `CIBW_BEFORE_BUILD` step.  Please compare with the
solutions e.g. in astropy or matplotlib for how to make this
conditional for nightly wheel builds using expressions.
If you do not worry about using pre-releases of all
dependencies, you can also use `--pre --extra-index-url` in the
build frontend (instead of build isolation).
This will become unnecessary as soon as NumPy 2.0 is released.

If your dependencies have the issue, check whether they
have nightly wheels build against NumPy 2.0.

pybind11 note: You may see this message if using pybind11,
this is not problematic at pre-release time
it indicates the need for a new pybind11 release.

[...]

PR #2981, opened 2023-11-28
[MTN] replace the deprecated sctypes

skoudoro:
`np.sctypes` is removed from Numpy 2.0 (https://numpy.org/devdocs/numpy_2_0_migration_guide.html#main-namespace)

This is something that we used only one time. So this PR provide an alternative to fix the future issue. 

follow up of https://github.com/dipy/dipy/issues/2979

PR #2980, opened 2023-11-28
[FIX] int_t to npy_intp

skoudoro:
In Numpy 2.0, `int_t` type is not available anymore and it is not recommended to use it. 

In our case, it can be safely replaced by `npy_intp`.

This fix DIPY compilation with Numpy 2.0. 

 A follow up PR should come today for the tests

PS: #2979

PR #2982, opened 2023-11-29
[FIX] Force the use of pre-wheels

skoudoro:
Currently, our `PRE_WHEELS` are not using Numpy 2.0 because a dependency uninstall it. 

So, this update should force the use of `PRE_WHEELS`.

ref: #2979

### DIFF
--- a/.github/workflows/test_pre.yml
+++ b/.github/workflows/test_pre.yml
@@ -11,7 +11,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  PRE-py311:
+  PRE-py312:
     uses: ./.github/workflows/test_template.yml
     with:
       runs-on: '["ubuntu-latest", ]'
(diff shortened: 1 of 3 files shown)


skoudoro:
it seems to be ok now. there are still a lot of rooms for improvement.

#2980 and #2981 needs to be merged, then I will rebase this PR to make sure everything is ok.

skoudoro:
All good. Now numpy 2.0 is used and DIPY build is OK.

`PRE` tests are failing because of `cvxpy`. [see here](https://github.com/dipy/dipy/actions/runs/8442506642/job/23124062840?pr=2982#step:10:129)

```
RuntimeError: module compiled against ABI version 0x1000009 but this version of numpy is 0x2000000
```

I created an issue to follow up with that (https://github.com/cvxpy/cvxpy/issues/2389) but this PR can be merged. The initial issue is fixed

PR #2987, opened 2023-12-04
STYLE: Make `cvxpy`-dependent test checking consistent in `test_mcsd`

jhlegarreta:
Make `cvxpy`-dependent test checking consistent in `test_mcsd`: use the `needs_cvxpy` decorator consistently in `test_mcsd`.

### DIFF
--- a/dipy/reconst/tests/test_mcsd.py
+++ b/dipy/reconst/tests/test_mcsd.py
@@ -4,7 +4,6 @@ from dipy.reconst.mcsd import (mask_for_response_msmt,
                                response_from_mask_msmt,
                                auto_response_msmt)
 from dipy.reconst.mcsd import MultiShellDeconvModel, multi_shell_fiber_response
-from dipy.reconst import mcsd
 import numpy as np
 import numpy.testing as npt
 import pytest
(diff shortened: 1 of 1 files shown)


dipy/reconst/tests/test_mcsd.py
```diff
@@ -56,7 +55,7 @@ def _expand(m, iso, coeff):
     return params
 
 
-@pytest.mark.skipif(not mcsd.have_cvxpy, reason="Requires CVXPY")
+@needs_cvxpy
```
skoudoro, line 58: It seems you also need to update the decorator line 19 by adding the reason (`needs_cvxpy = pytest.mark.skipif(not have_cvxpy, reason="Requires CVXPY")`

see the error here: https://github.com/dipy/dipy/actions/runs/7094023985/job/19308524800?pr=2987#step:9:5832

skoudoro (commented):
Hi @jhlegarreta,

It seems you also need to update the decorator line 19 by adding the reason (`needs_cvxpy = pytest.mark.skipif(not have_cvxpy, reason="Requires CVXPY")`

see the error here: https://github.com/dipy/dipy/actions/runs/7094023985/job/19308524800?pr=2987#step:9:5832

Thanks for the future update

dipy/reconst/tests/test_mcsd.py
jhlegarreta: Done in 7ef0df1.

skoudoro:
Thanks @jhlegarreta !

PR #2989, opened 2023-12-04
BF: Mask `1` values in leveraged, residual matrix computation

jhlegarreta:
Mask strict `1` values in leveraged, residual matrix computation: when the diagonal of `H` turns out to have `1`s, `1 - H.diagonal()` gives `0`s, and then a divide by 0 operation is encountered and makes NumPy raise a warning.

Fixes:
```
direction/tests/test_bootstrap_direction_getter.py::test_boot_pmf
  python3.10/site-packages/dipy/reconst/shm.py:1023: RuntimeWarning:
   divide by zero encountered in divide
    R = (np.eye(len(H)) - H) / leverages
```

and
```
direction/tests/test_bootstrap_direction_getter.py::test_boot_pmf
  python3.10/site-packages/dipy/reconst/shm.py:1024: RuntimeWarning:
   invalid value encountered in subtract
    return R - R.mean(0)
```

raised for example at:
https://github.com/dipy/dipy/actions/runs/7073407086/job/19253289839#step:9:4400

Masking `<=1` diagonal values was introduced in commit 6a01bf6.

### DIFF
--- a/dipy/reconst/shm.py
+++ b/dipy/reconst/shm.py
@@ -1018,7 +1018,7 @@ def lcr_matrix(H):
     if H.ndim != 2 or H.shape[0] != H.shape[1]:
         raise ValueError('H should be a square matrix')
 
-    leverages = np.sqrt(1 - H.diagonal(), where=H.diagonal() <= 1)
+    leverages = np.sqrt(1 - H.diagonal(), where=H.diagonal() < 1)
     leverages = leverages[:, None]
     R = (np.eye(len(H)) - H) / leverages
     return R - R.mean(0)


jhlegarreta:
The issue is raised because the diagonal of `H` turns out to be all `1`s, `1 - H.diagonal()` gives `0`s, and then a divide by 0 happens. Not sure what a diagonal of all 1's means in the `H` matrix in [shm::hat](https://github.com/dipy/dipy/blob/master/dipy/reconst/shm.py#L1008).

Instead of changing the mask, maybe this is indicative of something deeper that needs to be addressed.

skoudoro:
Hi @jhlegarreta,

Why do you keep this PR as a draft?

jhlegarreta:
> Why do you keep this PR as a draft?

Because I am not sure the solution proposed is the right one: https://github.com/dipy/dipy/pull/2989#issuecomment-1839749477

Was expecting/waiting for maintainers to comment.

skoudoro:
> Because I am not sure the solution proposed is the right one: https://github.com/dipy/dipy/pull/2989#issuecomment-1839749477 Was expecting/waiting for maintainers to comment.

Ok, I see

> Not sure what a diagonal of 1's/all 1s (which is the case happening in the test) means in the H matrix in [shm::hat](https://github.com/dipy/dipy/blob/master/dipy/reconst/shm.py#L1008).

I am not sure too.  @arokem or @Garyfallidis might know better for this case. We will wait for their feedback.

Thank you for this @jhlegarreta

Garyfallidis:
Hi @jhlegarreta this looks good but are you sure that R of 0s is okay? Can you please add extra tests? This function was made long time ago. We need to be extra careful. Thank you for the great work.

skoudoro (approved):
After a deeper look, I believe that `R=[0,0,0,0]` for the matrix `np.eye(4)` is ok and seems to make sense. Before, we were having a `nan` matrix. 

This PR is good to go for me.

PR #2992, opened 2023-12-05
RGB support for images

maharshi-gor:
### Changes
This PR enables the user to visualize the images with rgb channel support.

### How to use
User needs to send a `--rgb` flag in the command while invoking the Horizon
```
dipy_horizon <image_path> --rgb 
```
### Limitations
1. This only supports rgb data and not rgba data.
2. User can only use either rgb supported data or volume data image along side normal nifti images.
3. The colormap intensities and volumes slider will be hidden in the case of rgb images.

### What to expect
- If a rgb channel data passed along side with without volume nifti image without `--rgb` flag. The data will be shown in as volume data.
- If a volume data is passed with `--rgb` flag. This will result in two scenarios:
        **1. if size of the 4th dimension is not 3**
            In this scenario, there will be a warning for incorrectly showing the data as rgb and the data will be processed as 
volume data
        **2. if size of the 4th dimension is equal to 3**
             In this scenario, the data will be considered rgb and shown as the rgb image.

### Screenshots of the workings
Screenshot from 2023-12-05 14-40-34
Screenshot from 2023-12-05 14-40-26
Screenshot from 2023-12-05 14-36-55


dipy/viz/horizon/app.py
```diff
@@ -455,18 +456,20 @@ def build_show(self, scene):
...
                             self.show_m.iren, scene, data, affine=affine,
                             world_coords=self.world_coords,
-                            percentiles=[0, 100])
+                            percentiles=[0, 100], rgb=self.rgb)
                         self.__tabs.append(SlicesTab(
                             slices_viz, slice_id=img_count + 1,
                             force_render=self._show_force_render))
                         img_count += 1
                 else:
+                    # print(self.rgb)
```
skoudoro, line 469: print to remove

dipy/viz/horizon/app.py
```diff
@@ -39,7 +39,7 @@
 class Horizon:
 
     def __init__(self, tractograms=None, images=None, pams=None, cluster=False,
-                 cluster_thr=15.0, random_colors=None, length_gt=0,
+                 rgb=False, cluster_thr=15.0, random_colors=None, length_gt=0,
```
skoudoro, line 42: need to add to the docstring

dipy/workflows/viz.py
```diff
@@ -168,6 +170,9 @@ def run(self, input_files, cluster=False, cluster_thr=15.,
             if ends('.nii.gz') or ends('.nii'):
 
                 data, affine = load_nifti(fname)
+                # data = np.stack(np.array([data, data, data]), axis=-1)
+                # data[data >= 0.2] = 0.9
+                # print(data)
```
skoudoro, line 175: to remove

dipy/workflows/viz.py
```diff
@@ -42,6 +42,8 @@ def run(self, input_files, cluster=False, cluster_thr=15.,
         input_files : variable string
         cluster : bool, optional
             Enable QuickBundlesX clustering.
+        rgb : bool, optional
+            Enable the color image.
```
skoudoro, line 46: `Enable the color image (rgb only, alpha channel will be ignored)` instead

skoudoro (commented):
Hi @maharshi-gor,

Here is the initial review. I need to test it locally. Thanks!

dipy/viz/horizon/app.py
pjsjongsung, line 1: There should be some checks if someone passed rgb parameters with something other than 3 channels. You could throw a warning and ignore the rgb parameter.

maharshi-gor: This functionality is already available.

pjsjongsung (approved):
Looks good!

skoudoro (commented):
Hi @maharshi-gor,

using the following command: `dipy_horizon --rgb /Users/skoudoro/Downloads/RGB\ Tensor.nii.gz`

I got the following message:
```python
If the volumes are not shown you might want to remove --rgb option to process the volume data properly.
Original shape: (71, 88, 62, 3)
Resized to RAS shape: (71, 88, 62, 3)
/Users/skoudoro/devel/dipy/dipy/viz/horizon/tab/base.py:90: UserWarning: Images are of different dimensions, synchronization of slices will not work
  warnings.warn(msg)
```

the warning does not make sense, I have only 1 image.
the rgb message does not make sense, I use it. 

Can you fix this? or there are some specific reason for this message? 
thank you for your feedback.

skoudoro:
could you also rebase this PR again? Sorry for that.

skoudoro (approved):
all good, merging

PR #2995, opened 2023-12-09
[MTN] remove custom module `_importlib`

skoudoro:
This PR just remove our custom `_importlib` module.  It was necessary for python2 which we do not support anymore


skoudoro:
I am going ahead and merge this PR which is just cleaning some old module /functions

PR #2999, opened 2023-12-10
TEST: Set explicitly `CLARABEL` as the CVXPY solver

jhlegarreta:
Set explicitly `CLARABEL` as the CVXPY solver instead of relying on CVXPY choosing the solver: CVXPY resorts to `ECOS` by default, which will be deprecated starting CVXPY 1.5.0.

Fixes:
```
reconst/tests/test_dki.py::test_dki_fits
  python3.10/site-packages/cvxpy/reductions/solvers/solving_chain.py:336: FutureWarning:
      Your problem is being solved with the ECOS solver by default. Starting in
      CVXPY 1.5.0, Clarabel will be used as the default solver instead. To continue
      using ECOS, specify the ECOS solver explicitly using the ``solver=cp.ECOS``
      argument to the ``problem.solve`` method.

    warnings.warn(ECOS_DEPRECATION_MSG, FutureWarning)
```

raised, for example, at:
https://github.com/dipy/dipy/actions/runs/7146907000/job/19465502675#step:9:4433

### DIFF
--- a/dipy/reconst/tests/test_dki.py
+++ b/dipy/reconst/tests/test_dki.py
@@ -2,6 +2,7 @@
 
 import numpy as np
 import random
+
 import dipy.reconst.dki as dki
 import dipy.reconst.dti as dti
 from numpy.testing import (assert_array_almost_equal, assert_array_equal,
@@ -23,7 +24,7 @@ from dipy.core.geometry import sphere2cart
 from dipy.utils.optpkg import optional_package
 from dipy.utils.tripwire import TripWireError
 
-_, have_cvxpy, _ = optional_package("cvxpy", min_version="1.4.1")
+cvxpy, have_cvxpy, _ = optional_package("cvxpy", min_version="1.4.1")
 
 fimg, fbvals, fbvecs = get_fnames('small_64D')
 bvals, bvecs = read_bvals_bvecs(fbvals, fbvecs)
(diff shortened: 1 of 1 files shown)


dipy/reconst/tests/test_dki.py
```diff
@@ -4,6 +4,7 @@
 import random
 import dipy.reconst.dki as dki
 import dipy.reconst.dti as dti
+from cvxpy import settings as s
```
skoudoro, line 7: Thank you for this! We try to avoid 1 letter variable. The minimum is 3 letters when possible. 

I do not think you need `settings`. you can just do `import cvxpy` and then `cvxpy.ECOS` or `cvxpy.CLARABEL`.

Also, the default one was ECOS, why do you change it to Clarabel?  I recommend to keep ecos for now, so `cvpy_solver=cvxpy.ECOS` since the warning just tell us to be explicit. We might change the default solver when we know more about the difference between both

skoudoro (changes requested):
Hi @jhlegarreta,

Thank you for this. see below some comment

dipy/reconst/tests/test_dki.py
jhlegarreta: Changed the import in 8f5988c. Using `cvxpy.ECOS` (or previously `s.CLARABEL`) results in the test failing: please check it locally. I assume that what the warning _only_ means is that ECOS is the current default solver, but it will be changed: despite ECOS being the default solver, if I'm not mistaken, CVXPY _may_ internally pick another one that best suits the problem. I ignore whether in its verbose mode it would tell us the solver that it has been using.

jhlegarreta:
The warning is gone.

dipy/reconst/tests/test_dki.py
jhlegarreta: This fails now: https://github.com/dipy/dipy/actions/runs/7158982614/job/19491729377#step:9:4498
```
FAILED reconst/tests/test_dki.py::test_dki_fits - AttributeError: module 'cvxpy' has no attribute 'CLARABEL'
```

However, it looks like we should be able to call it:
https://github.com/cvxpy/cvxpy/blob/f8e3b2c29da53712cad6b89314588f10c3a96afa/cvxpy/__init__.py#L66

Any intuition @skoudoro ?

skoudoro: it fails with anaconda, so we need to check cvxpy version on Anaconda and when this change has been introduced, I will look into it

skoudoro: We might need to update our minimal version for cvxpy due to this following information: https://www.cvxpy.org/updates/index.html#ecos-deprecation

skoudoro: I plan to do 2 PRs to improve the minimum version management. After that, you will need to rebase and it should be ok.

The strategy is to skip the tests if we do not respect the minimum version

jhlegarreta: a970b79 rebased on master.

skoudoro (approved):
Ok, The conda issue has been fixed. The warning disappeared also. So this is ready to go. 

Thanks @jhlegarreta for this work. Merging

dipy/reconst/tests/test_dki.py
jhlegarreta: https://github.com/dipy/dipy/pull/2999#pullrequestreview-1778264214 Resolving.

PR #3001, opened 2023-12-10
RF: Create PCA denoising utils methods

jhlegarreta:
Create PCA denoising utils methods: refactor code blocks into their own methods to improve encapsulation.

Add the corresponding tests.

Add a method to report about the dimensionality problem message.

The above allows for the dimensionality problem message to be checked and filtered in tests.

Fixes:
```
denoise/tests/test_lpca.py::test_mppca_returned_sigma
 denoise/localpca.py:194: UserWarning:
 Number of samples 27 - 1 < Dimensionality 104.
 This might have a performance impact.
 Increase patch_radius to 2 to avoid this warning,
 or supply suppress_warning=True to your function call.
    warn(e_s, UserWarning)
```

reported, for example, at:
https://github.com/dipy/dipy/actions/runs/7146907000/job/19465502675#step:9:4410

### DIFF
--- a/dipy/denoise/localpca.py
+++ b/dipy/denoise/localpca.py
@@ -1,3 +1,5 @@
+import copy
+
 from packaging.version import Version
 from warnings import warn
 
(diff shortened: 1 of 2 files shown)


dipy/denoise/localpca.py
```diff
@@ -66,6 +94,99 @@ def _pca_classifier(L, nvoxels):
...
+    Returns
+    -------
+    int
+        Number of samples.
+    """
+
+    return np.prod(patch_size)
+
+
+def compute_suggested_patch_radius(arr, patch_size):
```
jhlegarreta, line 162: This could have taken `arr.shape[-1]` as the first argument, but I thought it was less clear from the point of view of the caller.

skoudoro (approved):
I am ok with this refactoring.

Thanks @jhlegarreta, merging

PR #3008, opened 2023-12-12
NF: add SH basis conversion between dipy and mrtrix3

### LINKED ISSUE #2993: Add conversion utility between DIPY and MRtrix3 spherical harmonic basis
## Description

For interoperability between DIPY and MRtrix3, it would be useful to connect the two spherical harmonic bases. The basis used by mrtrix can be found [here](https://mrtrix.readthedocs.io/en/latest/concepts/spherical_harmonics.html). It [exists in dipy](https://github.com/dipy/dipy/blob/618b39525ad5276f774b9920da983920d59cf340/dipy/reconst/shm.py#L300-L305) as `dipy.reconst.shm.real_sh_tournier` with `legacy=False`. However, dipy is locked into the legacy `descoteaux07` basis.

The enhancement suggested [here](https://github.com/dipy/dipy/discussions/2959#discussioncomment-7480673) is:

- a conversion function in `reconst/shm.py`
- potentially also a workflow to provide a cli

[A lot more detail can be found here.](https://github.com/dipy/dipy/discussions/2959#discussioncomment-7481675)

Please assign me and add appropriate labels.

ebrahimebrahim:
Close #2993

This adds a function `convert_sh_descoteaux_tournier` to convert SH representations between DIPY and MRtrix3.

(There was also a plan to add something to add an associated workflow to `dipy.workflows.io`, but I don't feel I understand the workflow concept well enough to add that. Not doing it now.)


ebrahimebrahim:
> Hello @ebrahimebrahim, Thank you for submitting the Pull Request !
> 
> * In the file [`dipy/reconst/shm.py`](https://github.com/dipy/dipy/blob/f6cc0273ee9d1ac98558b65a2dff8062c587ef30/dipy/reconst/shm.py):
> 
> > [Line 1615:81](https://github.com/dipy/dipy/blob/f6cc0273ee9d1ac98558b65a2dff8062c587ef30/dipy/reconst/shm.py#L1615): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (94 > 80 characters)
> > [Line 1616:81](https://github.com/dipy/dipy/blob/f6cc0273ee9d1ac98558b65a2dff8062c587ef30/dipy/reconst/shm.py#L1616): [E501](https://duckduckgo.com/?q=pep8%20E501) line too long (96 > 80 characters)
> 
> Do see the [DIPY coding Style guideline](https://github.com/dipy/dipy/blob/master/doc/devel/coding_style_guideline.rst)

(these are URLs so it will have to violate pep8)

skoudoro:
Hi @ebrahimebrahim,

Thank you doing this!

concerning the violating pep8, just add `# noqa: E501` at the end of the violating line.

concerning the workflow, you need
- to add a class that will call your function. [look at this tutorial](https://docs.dipy.org/1.8.0/examples_built/workflows/workflow_creation.html#sphx-glr-examples-built-workflows-workflow-creation-py)
- define the command line name in `pyproject.toml`. look [here](https://github.com/dipy/dipy/blob/master/pyproject.toml#L51-L83)
- add your workflow name and cli name in `dipy.workflows.cli`.  look [here](https://github.com/dipy/dipy/blob/master/dipy/workflows/cli.py#L8-L43)

ebrahimebrahim:
I tried out the CLI workflow on an example case. It seems to be all set. Thanks for explaining workflows!

dipy/workflows/io.py
```diff
@@ -257,3 +258,39 @@ def run(self, input_files, vol_idx=0, out_dir='',
             save_nifti(osplit, split_vol, affine, image.header)
 
             logging.info('Split volume saved as {0}'.format(osplit))
+
+
+class ConvertDipyMrtrixFlow(Workflow):
```
jhlegarreta, line 263: Please use a more meaningful name for the workflow: what does it convert ? e.g.  `ConvertDipyMrtrixSHFlow`

dipy/workflows/cli.py
```diff
@@ -35,6 +35,7 @@
     "dipy_median_otsu": ("dipy.workflows.segment", "MedianOtsuFlow"),
     "dipy_recobundles": ("dipy.workflows.segment", "RecoBundlesFlow"),
     "dipy_reslice": ("dipy.workflows.align", "ResliceFlow"),
+    "dipy_shm_convert_mrtrix": ("dipy.workflows.io", "ConvertDipyMrtrixFlow"),
```
jhlegarreta, line 38: `dipy_shm_convert_mrtrix` : `dipy_sh_convert_mrtrix`

Despite the name of the module being `shm`, `shm` is less identifiable than `sh` to name spherical harmonics, so at least for the workflow, `sh` means more easily recognizable.

jhlegarreta:
Thanks for doing this @ebrahimebrahim.

dipy/workflows/io.py
skoudoro: I agree, maybe just `ConvertSHFlow`

dipy/workflows/cli.py
skoudoro: or `dipy_convert_sh` you might need to add a parameter as a string to select the output basis type

ebrahimebrahim: Output basis type does not need to be selected because the same conversion function works both ways. It equals its own inverse. I can explain that in the docstring.

ebrahimebrahim:
Hmm the code coverage thing is not happy with me because of "patch" percent coverage. Maybe it does not count mock objects?

dipy/workflows/tests/test_io.py
```diff
@@ -91,3 +92,28 @@ def test_split_flow():
...
+        "dipy.workflows.io.save_nifti"
+    ), patch.object(
+        ConvertSHFlow, "get_io_iterator", return_value=iter([(0, 0)])
+    ):
+
+        workflow = ConvertSHFlow()
+        workflow.run("fake_input_path", "fake_output_dir")
+
+        # just test that the convert function gets called
+        mock_convert.assert_called()
```
skoudoro, line 119: Can you look how we tests our workflows and follow the standard way ? (meaning avoid `mock`)

Thank you

skoudoro (changes requested):
Hi @ebrahimebrahim,

No worries for codecov. you are ok following this: https://app.codecov.io/gh/dipy/dipy/pull/3008/commits

see below my comment.

dipy/workflows/tests/test_io.py
ebrahimebrahim: The advantage of mocking is that unit tests become more independent. Then when someone breaks something fewer tests fail and the ones that do fail make it more clear what actually broke. 

If I follow the other examples then I have to use some of the example data from `get_fnames` right? I don't think there are there test FOD images saved as nifti, so I am not sure how to follow those examples without at least mocking `load_nifti`. Any advice?

skoudoro: > The advantage of mocking is that unit tests become more independent. Then when someone breaks something fewer tests fail and the ones that do fail make it more clear what actually broke.

I understand the advantage. However, to keep consistency and facilitate the maintenance, we try to keep the same standard. When we decided to switch to mock, we will put an effort to change all worflows using mock. Feel free to create an issue to discuss to this point.

> If I follow the other examples then I have to use some of the example data from get_fnames right? I don't think there are there test FOD images saved as nifti, so I am not sure how to follow those examples without at least mocking load_nifti. Any advice?

I will follow your `test_convert_sh_descoteaux_tournier`.   Meaning:
- Do it all in a temporary directory (`with TemporaryDirectory() as tmpdir:`)
- use the same input array (`sh_coeff`), and save it as a nifti, with identity matrix as an affine
- run the workflows
- load the output nifti
- compare the output nifiti with expected_sh_coeff

Does it make sense ?

dipy/workflows/tests/test_io.py
```diff
@@ -91,3 +94,33 @@ def test_split_flow():
         split_data, split_affine = load_nifti(split_path)
         npt.assert_equal(split_data.shape, volume[..., 0].shape)
         npt.assert_array_almost_equal(split_affine, affine)
+
+
+def test_convert_sh_flow():
+    with TemporaryDirectory() as out_dir:
+        out_dir = Path(out_dir)
+        filepath_in = out_dir/'sh_coeff_img.nii'
```
skoudoro, line 102: you should use `os.path.join(out_dir,'sh_coeff_img.nii')` 

I suppose, Windows CI are failing because of `/`

dipy/workflows/tests/test_io.py
```diff
@@ -91,3 +94,33 @@ def test_split_flow():
...
+        filepath_out = out_dir/'sh_coeff_img_converted.nii'
```
skoudoro, line 103: same comment as above

skoudoro (commented):
Thank you for the update @ebrahimebrahim.

See below for my comment. 

is there any additional comment concerning this PR? @arokem?

dipy/workflows/tests/test_io.py
ebrahimebrahim: ah thanks for catching this. I didn't know pathlib doesn't take the os differences into account

skoudoro: I think there is a way to do it correctly with pathlib also but I need to look. maybe later today

ebrahimebrahim:
Any idea why the windows tests are failing after using `os.path.join`?

```txt
FAILED workflows/tests/test_io.py::test_convert_sh_flow - NotADirectoryError: [WinError 267] The directory name is invalid: 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmp4f8uynzy\\sh_coeff_img_converted.nii'
```

skoudoro:
> Any idea why the windows tests are failing after using os.path.join?

Interesting.... no idea, I need to connect to Windows machine and see what's going on. Unfortunately, I can do it only next week...

ebrahimebrahim:
> > Any idea why the windows tests are failing after using os.path.join?
> 
> Interesting.... no idea, I need to connect to Windows machine and see what's going on. Unfortunately, I can do it only next week...

Thanks, and no rush at all! I appreciate the testing effort as I don't have easy access to a windows workstation for a while.

There may be something I've misunderstood about workflows and `out_dir`. From the error message it looks like the path `C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\tmp4f8uynzy\\sh_coeff_img_converted.nii` was expected to be a _directory_  for some reason.

dipy/workflows/tests/test_io.py
```diff
@@ -91,3 +93,33 @@ def test_split_flow():
         split_data, split_affine = load_nifti(split_path)
         npt.assert_equal(split_data.shape, volume[..., 0].shape)
         npt.assert_array_almost_equal(split_affine, affine)
+
+
+def test_convert_sh_flow():
+    with TemporaryDirectory() as out_dir:
+        filepath_in = os.path.join(out_dir, 'sh_coeff_img.nii')
+        filename_out = 'sh_coeff_img_converted.nii'
```
skoudoro, line 101: I just tested on Windows and I got the same error. 

Not sure what's going on but I had to add the `gz`extension (meaning `sh_coeff_img_converted.nii.gz`) for the test to pass. 

it seems the cleanup between `*.nii` and `*.nii.gz` differ. I need to find out why, but for now, this is your quick fix.

Maybe @arokem has an idea.

skoudoro (changes requested):
Hi @ebrahimebrahim,

Happy New Year !  sorry for the late feedback see below my comment. this should fix your issue for now.

skoudoro:
Also, Can you update this line in the documentation: https://github.com/dipy/dipy/blob/master/doc/recipes.rst?plain=1#L35-L36

Thank you

dipy/workflows/tests/test_io.py
ebrahimebrahim: Happy new year and thank you for debugging on this!

ebrahimebrahim:
This should be ready for another look.

If I understand the output correctly from the two failing checks, I think in both cases it's not something related to this PR.

dipy/reconst/shm.py
```diff
@@ -1575,3 +1575,52 @@ def convert_sh_to_legacy(sh_coeffs, sh_basis, full_basis=False):
...
 
     return out_sh_coeffs
+
+
+def convert_sh_descoteaux_tournier(sh_coeffs):
+    """Convert SH coefficients between the legacy ``descoteaux07`` SH basis and
+    the non-legacy ``tournier07`` SH basis. Because this conversion is equal to
+    its own inverse, it can be used to convert in either direction:
+    legacy-descoteaux to non-legacy-tournier or non-legacy-tournier to
+    legacy-descoteaux.
```
skoudoro, line 1585: Can you update the docstring to respect our standard. see [here](https://github.com/dipy/dipy/blob/master/dipy/direction/peaks.py#L94-L131) for an example

skoudoro (approved):
Hi @ebrahimebrahim,

Thank you again for this.  it looks good to me.

I will wait for until friday to see if we have a feedback from @jhlegarreta or @arokem or someone else. If not, I will go ahead and merge it.

Thank you for this work.

ps: during this time, if you could fix the small docstring issue, it would be great. Thanks!

dipy/reconst/shm.py
```diff
@@ -1575,3 +1575,52 @@ def convert_sh_to_legacy(sh_coeffs, sh_basis, full_basis=False):
...
+    .. [descoteaux07] Descoteaux, M., Angelino, E., Fitzgibbons, S. and
+           Deriche, R. Regularized, Fast, and Robust Analytical Q-ball Imaging.
+           Magn. Reson. Med. 2007;58:497-510.
+    .. [tournier19] Tournier J-D, Smith R, Raffelt D, Tabbara R, Dhollander T,
+           Pietsch M, et al. MRtrix3: A fast, flexible and open software
+           framework for medical image processing and visualisation.
+           NeuroImage. 2019 Nov 15;202:116-137.
+    .. [mrtrixbasis] https://mrtrix.readthedocs.io/en/latest/concepts/spherical_harmonics.html
+    .. [mrtrixdipybases] https://github.com/dipy/dipy/discussions/2959#discussioncomment-7481675
+    """  # noqa: E501
```
jhlegarreta, line 1617: Adding a blank line after the docstring and immediately before the first line of code gives some more breath to the code, improves readability.

jhlegarreta (commented):
I have a style comment. Not a show-stopper, but would be nice to address as Serge has also requested a change.

Thanks for doing this @ebrahimebrahim.

dipy/reconst/shm.py
ebrahimebrahim: Can you be more specific about what needs to be changed? It seems to follow the example already.

jhlegarreta (commented):
IMO, all commits in this PR should be squashed and all meaningful changes be explained in that single commit as this is a new feature: e.g. without commits 035858f and 58d6cba we would be introducing commits that do not work/do not pass tests in the history.

Would reduce the number of commits in the history, and thus make it less complex and save time when reading through it.

dipy/reconst/shm.py
skoudoro: Sorry, the first summary should be one-line summary. 

So, it would be great to shorten this sentence `Convert SH coefficients between the legacy ``descoteaux07`` SH basis and the non-legacy ``tournier07`` SH basis.`

and move the sentence starting with `because this conversion ....` to the extended summary (so in a new line, separated by an empty line.)

we follow Numpy  docstring standard: https://numpydoc.readthedocs.io/en/latest/format.html

skoudoro:
Concerning the squash, I agree with @jhlegarreta, but it is a bonus, I think it is fine if you can not do it.

ebrahimebrahim:
Comments addressed and commits squashed

skoudoro:
Great, thanks! waiting for the CI's, then we can merge

jhlegarreta (commented):
I am not going to oppose to merging this, but the commit message is not faithful to the change or NF: it has simply used the previous commit message subjects. Thus, e.g. saying that `.gz` was added to the `.nii` extensions, that long line or URL PEP8 is bypassed, that the function naming is improved or that `os.path` was used instead of `pathlib` is now meaningless. This is now a single commit, and readers are interested in the final result or added feature; the intermediate steps or process does not add meaningful information to the commit history; it is only relevant to the GitHub conversation.

So a self-contained commit message could be along the following lines:
```
NF: Add DIPY to MRtrix3 SH basis conversion capabilities

Add legacy descoteaux07 (DIPY) and tournier07 (MRtrix3) SH basis conversion capabilities.

Add the corresponding tests.

Add and test the corresponding workflow.

Document the feature in the DIPY how-to recipes file.
```

Hope this makes sense.

ebrahimebrahim:
> So a self-contained commit message could be along the following lines:

Thanks, I agree with your reasoning and like your rephrasing. I copied your message :)

jhlegarreta (approved):
Thanks for doing this and for addressing the comments @ebrahimebrahim :100:.

jhlegarreta:
macOS VIZ failure is unrelated. Merging.

PR #3012, opened 2023-12-14
DOC: Document observance for Scientific Python min supported versions

jhlegarreta:
Document observance for Scientific Python minimum supported versions recommendation.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -73,6 +73,14 @@ or using `conda`::
 For detailed installation instructions, including instructions for installing
 from source, please read our `installation documentation <https://dipy.org/documentation/latest/installation/>`_.
 
+Python versions and dependencies
+--------------------------------
+
+DIPY follows the `Scientific Python`_ `SPEC 0 — Minimum Supported Versions`_
+recommendation as closely as possible, including the supported Python and
+dependencies versions.
+
+Further information can be found in :ref:`toolchain-roadmap`.
 
 License
 =======
@@ -92,3 +100,7 @@ Reference
     S. Van Der Walt, M. Descoteaux, I. Nimmo-Smith and DIPY contributors,
     "DIPY, a library for the analysis of diffusion MRI data",
     Frontiers in Neuroinformatics, vol. 8, p. 8, Frontiers, 2014.
+
+
+.. _`Scientific Python`: https://scientific-python.org/
+.. _`SPEC 0 — Minimum Supported Versions`: https://scientific-python.org/specs/spec-0000/
--- a/doc/links_names.inc
+++ b/doc/links_names.inc
@@ -81,6 +81,10 @@
 .. _cython: https://cython.org
 .. _travis-ci: https://travis-ci.com/
 
+.. Scientific Python practices
+.. _`Scientific Python`: https://scientific-python.org/
+.. _`SPEC 0 — Minimum Supported Versions`: https://scientific-python.org/specs/spec-0000/
+
 .. Other python projects
 .. _numpy: https://numpy.org
 .. _scipy: https://www.scipy.org
--- a/doc/user_guide/dependencies.rst
+++ b/doc/user_guide/dependencies.rst
@@ -1,8 +1,17 @@
 .. _dependencies:
 
-============
+================================
+Python versions and dependencies
+================================
+
+DIPY follows the `Scientific Python`_ `SPEC 0 — Minimum Supported Versions`_
+recommendation as closely as possible, including the supported Python and
+dependencies versions.
+
+Further information can be found in :ref:`toolchain-roadmap`.
+
 Dependencies
-============
+------------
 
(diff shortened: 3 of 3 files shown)


skoudoro (commented):
Hi @jhlegarreta,

Thank you for this. We talk about it in https://github.com/dipy/dipy/blob/master/doc/devel/toolchain.rst,  So maybe you can also add a reference to this internal document.

skoudoro:
What do you think ?

jhlegarreta:
Wasn't aware of that bit. Thanks for doing this. A few comments:
- To me, this information does not only apply to developers, but to users as well (i.e. those trying to use it through a `pip install`), so maybe the location of the file should be rethought. Also, I've seen the term "toolchain" be used in other context, like what you would use to compile a C program, so it is misleading to me to use it to include dependencies.
- Have not gone through the entire document, but for example, I am not sure the table in https://github.com/dipy/dipy/blame/master/doc/devel/toolchain.rst#L50 is clear: what does that line mean: that Python 3.8+, including Python 3.8, is required starting mid-2023? When was 3.7 effectively dropped? We were about to drop 3.8 CI a few weeks ago, so I am not sure about all these timelines. Also, how they overlap with the Scientific Python ecosystem timelines? If they only reproduce them, I would try to redirect to these tables to avoid the risk of going out of sync.

Once the above details get sorted out, I could rebase this PR to cross-reference this document.

skoudoro:
> To me, this information does not only apply to developers, but to users as well (i.e. those trying to use it through a pip install), so maybe the location of the file should be rethought. 

I see what you mean. With your current PR, I think it can stay where it is. They seems to be complementary. 

> Also, I've seen the term "toolchain" be used in other context, like what you would use to compile a C program, so it is misleading to me to use it to include dependencies.

What is your naming proposition? Currently nothing else come to my mind. I will look at this renaming for next release.

> Have not gone through the entire document, but for example, I am not sure the table in https://github.com/dipy/dipy/blame/master/doc/devel/toolchain.rst#L50 is clear: what does that line mean: that Python 3.8+, including Python 3.8, is required starting mid-2023? 

starting mid-2023, DIPY support python 3.8+

> When was 3.7 effectively dropped?

mid-2023

> We were about to drop 3.8 CI a few weeks ago, so I am not sure about all these timelines. Also, how they overlap with the Scientific Python ecosystem timelines? If they only reproduce them, I would try to redirect to these tables to avoid the risk of going out of sync.

the timeline is currently based on the previous releases (existing releases). I have checked all the previous release compatibility. From 2024, we will try to follow scientific-python recommendation. They are recommendation, so we will do our best to follow them but we allow ourselves to make some modification and not be strict on them. for each releases, we would like to support 4 python versions.

jhlegarreta:
> What is your naming proposition?

"dipy_ecosystem"? Have not thought enough, but "toolchain" is misleading I'd say.

> starting mid-2023, DIPY support python 3.8+

Python 3.8 has been supported since a long time ago, and hence my question.

> mid-2023

:+1:

skoudoro:
> Python 3.8 has been supported since a long time ago, and hence my question.

Yes, and previous line implied it by saying 
mid-2020          Py3.6+
 2022              Py3.7+

so the line means: starting mid-2023, DIPY support python 3.8+ or do not support python 3.7 or less

the table show the python minimal supported version

But I see your confusion, Maybe it is better to put in the table when we dropped a python version

jhlegarreta:
Re https://github.com/dipy/dipy/pull/3012#issuecomment-1856114949 I see: maybe it was my viewpoint the biased one. Maybe it is enough if it is clarified when referring to the paragraph; if this is the usual way this is put, maybe it is not necessary. Thanks.

jhlegarreta:
@skoudoro 20649ad cross-references the document you contributed in commit 3616798.

jhlegarreta:
The coverage build failure is unrelated to this patch set.

README.rst
```diff
@@ -73,6 +73,14 @@ or using `conda`::
...
 from source, please read our `installation documentation <https://dipy.org/documentation/latest/installation/>`_.
 
+Python versions and dependencies
+--------------------------------
+
+DIPY follows the `Scientific Python`_ `SPEC 0 — Minimum Supported Versions`_
+recommendation as closely as possible, including the supported Python and
+dependencies versions.
+
+Further information can be found in :ref:`toolchain-roadmap`.
```
skoudoro, line 83: I am not sure this reference will work (render correctly) on DIPY Github. 

I will go ahead and merge this PR but can you please check and do a follow up PR if it does not render correctly on https://github.com/dipy/dipy

skoudoro: thank you!

skoudoro (approved):

skoudoro:
LGTM, merging!

README.rst
jhlegarreta: @skoudoro :+1: PR #3019.

PR #3013, opened 2023-12-14
Added Fibonacci spiral and test for it

Atharva-Shah-2298:
## Fibonacci Sphere Method for Sphere Vertices

The `fibonacci_sphere` method is designed for efficiently generating points on the surface of a sphere, such as in the iterations of disperse charges. The time complexity is reduced because of the analytical solution as opposed to an optimization problem solution in disperse charges

### Method Signature:

```python
def fibonacci_sphere(n_points: int, randomize: bool) -> numpy.ndarray of shape (n_points,3)
```

### Example Usage

```python
from dipy.core.sphere import fibonacci_sphere,Sphere
points = fibonacci_sphere(n_points=1000, randomize=True)
sph = Sphere(xyz = points)
```

### DIFF
--- a/dipy/core/sphere.py
+++ b/dipy/core/sphere.py
@@ -496,6 +496,43 @@ def disperse_charges(hemi, iters, const=.2):
     return HemiSphere(xyz=charges), potential
 
 
+def fibonacci_sphere(n_points, randomize=True):
+    """
+    Generate points on the surface of a sphere using Fibonacci Spiral.
+
+    Parameters
+    ----------
+    n_points : int
+        The number of points to generate on the sphere surface.
+    randomize : bool, optional
+        If True, randomize the starting point on the sphere. Default is True.
+
+    Returns
+    -------
+    points : ndarray
+        An array of 3D points representing coordinates on the sphere surface.
+
+    """
+    random_shift = 0
+    if randomize:
+        random_generator = np.random.default_rng()
+        random_shift = random_generator.integers(0, n_points)
+
+    indices = np.arange(n_points)
+    offset = 2.0 / n_points
+    increment = np.pi * (3. - np.sqrt(5.))
+
+    y = ((indices * offset) - 1) + (offset / 2)
+    r = np.sqrt(1 - y**2)
+    phi = ((indices + random_shift) % n_points) * increment
+
+    x = np.cos(phi) * r
+    z = np.sin(phi) * r
+
+    points = np.column_stack((x, y, z))
+    return points
+
+
 def _equality_constraints(vects):
     """Spherical equality constraint. Returns 0 if vects lies on the unit
     sphere. Note that a flattened array is returned because `scipy.optimize`
--- a/dipy/core/sphere.py
+++ b/dipy/core/sphere.py
@@ -513,6 +513,11 @@ def fibonacci_sphere(n_points, randomize=True):
         An array of 3D points representing coordinates on the sphere surface.
 
     """
+    if not isinstance(n_points, int) or n_points <= 4:
+        raise ValueError(
+            "Number of points must be a positive integer greater than 4."
+        )
+
     random_shift = 0
     if randomize:
         random_generator = np.random.default_rng()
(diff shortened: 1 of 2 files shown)


dipy/core/sphere.py
```diff
@@ -496,6 +497,44 @@ def disperse_charges(hemi, iters, const=.2):
...
+    """
+    Generate points on the surface of a sphere using Fibonacci Spiral Sampling.
+
+    Parameters
+    ----------
+
+    - n_points (int): The number of points to generate on the sphere surface.
+      Default is 724.
+    - randomize (bool): If True, randomize the starting point on the sphere.
+      Default is True.
```
skoudoro, line 510: Can you remove the default value. 

Also, Can you look at the other function to respect our docstring standard?  currently your docstring does not follow it

dipy/core/sphere.py
```diff
@@ -496,6 +497,44 @@ def disperse_charges(hemi, iters, const=.2):
...
+
+    Returns
+    -------
+
+    array-like (N,3): An array of 3D points representing coordinates
+    on the sphere surface.
```
skoudoro, line 516: same comment as above

dipy/core/sphere.py
```diff
@@ -496,6 +497,44 @@ def disperse_charges(hemi, iters, const=.2):
...
+    offset = 2.0 / n_points
+    increment = np.pi * (3. - np.sqrt(5.))
+
+    for i in range(n_points):
+        y = ((i * offset) - 1) + (offset / 2)
+        r = np.sqrt(1 - pow(y, 2))
+        phi = ((i + random_shift) % n_points) * increment
+        x = np.cos(phi) * r
+        z = np.sin(phi) * r
+        points.append([x, y, z])
```
skoudoro, line 533: this can be done with numpy array and it will be way faster !  I do not see the need of a loop here. so this fucntion can be done without loop. 

Please, can you update ? thanks

skoudoro (commented):
Hi @Atharva-Shah-2298,

Thank you for this!  See below an initial review of your PR. Thank you for your future update

skoudoro:
Also, Can you squash your pep8 commit ? thank you

skoudoro (changes requested):
Also, Can you squash your pep8 commit ? thank you

dipy/core/sphere.py
Atharva-Shah-2298: Now the array is pre populated and updates each point. Thanks for the suggestion

dipy/core/sphere.py
```diff
@@ -496,6 +496,43 @@ def disperse_charges(hemi, iters, const=.2):
...
+        An array of 3D points representing coordinates on the sphere surface.
+
+    """
+    random_shift = 0
+    if randomize:
+        random_generator = np.random.default_rng()
+        random_shift = random_generator.integers(0, n_points)
+
+    indices = np.arange(n_points)
+    offset = 2.0 / n_points
```
skoudoro, line 522: is there a minimal number of points  needed? Can you add a small tests (`if`) to check that and avoid invalid value like a division by zero.

```python
if nb_points < 0:
   raise ValueError("my message")
```

skoudoro (commented):
Hi @Atharva-Shah-2298,

Overall, looks good and much better! thank you for the update. 

See below my last comment but it seems ready to be merged.

skoudoro (approved):
Looks good to me. here the results:


image

with the following code:
```python
In [1]: from dipy.core.sphere import fibonacci_sphere,Sphere

In [2]: points = fibonacci_sphere(n_points=1000, randomize=True)

In [3]: points = fibonacci_sphere(n_points=323, randomize=True)

In [4]: sph = Sphere(xyz = points)

In [5]: from dipy.viz import window, actor

In [6]: scene = window.Scene()
   ...: scene.SetBackground(1, 1, 1)

In [7]: scene.add(actor.point(sph.vertices, window.colors.red, point_radius=0.05))

In [8]: window.show(scene)
```

I will go ahead and merge this PR.

Thank you for this work @Atharva-Shah-2298

Issue #2647, opened 2022-09-18
Allow to run DIPY unittest offline


skoudoro:
When I am offline, I can not run the DIPY unitest because some test needs to fetch data online. 

We should avoid that for unit testing and allow it only for tutorial.

kunalKumar-13:
This is smaller than it looks, so here is the measurement in case it helps someone pick it up.

**18 of 178 `test_*.py` files reach the network.** I classified every `get_fnames` name by whether its branch calls a `fetch_*` (20 names are packaged under `dipy/data/files/`, 37 download), then scanned the test tree for those names plus direct `fetch_*` / `read_stanford_*` / `read_ivim` calls:

| file | needs |
|---|---|
| `dipy/align/tests/test_api.py` | `sherbrooke_3shell`, `syn_data`, `read_stanford_hardi` |
| `dipy/data/tests/test_fetcher.py` | `fetch_data` |
| `dipy/direction/tests/test_ptt_direction_getter.py` | `ptt_minimal_dataset` |
| `dipy/io/tests/test_stateful_surface.py` | `gold_standard_io`, `real_data_io` |
| `dipy/io/tests/test_stateful_tractogram.py` | `gold_standard_io` |
| `dipy/io/tests/test_streamline.py` | `gold_standard_io` |
| `dipy/io/tests/test_surface.py` | `real_data_io` |
| `dipy/io/tests/test_utils.py` | `gold_standard_io` |
| `dipy/nn/tests/test_deepn4.py` | `fetch_default_weights`, `deepn4_test_data` |
| `dipy/nn/tests/test_evac.py` | `evac_test_data` |
| `dipy/nn/tests/test_histo_resdnn.py` | `fetch_default_weights`, `stanford_hardi` |
| `dipy/nn/tests/test_synb0.py` | `fetch_default_weights`, `synb0_test_data` |
| `dipy/nn/tests/test_synthseg.py` | `synthseg_test_data` |
| `dipy/reconst/tests/test_csdeconv.py` | `read_stanford_hardi` |
| `dipy/reconst/tests/test_ivim.py` | `read_ivim` |
| `dipy/utils/tests/test_tractogram.py` | `gold_standard_io` |
| `dipy/workflows/tests/test_io.py` | `gold_standard_io` |
| `dipy/workflows/tests/test_nn.py` | `evac_test_data`, `stanford_t1` |

It clusters more than the count suggests:

[...]

PR #3025, opened 2024-01-02
[TEST] Manage http errors for stateful tractograms

skoudoro:
This PR provides an alternative when there is http issue with `fetch_gold_standard_io`. 

Instead of failing, we just skip the tests and provide a reason. It also one step further concerning offline testing (see https://github.com/dipy/dipy/issues/2647)

It also cleanup some global variable.


skoudoro:
Anyone to look at this ? 

I will wait until Wednesday and then go ahead and merge if no objection.

skoudoro:
ok, I am going ahead and merge this.

Issue #2760, opened 2023-03-24
TRX integration 


arokem:
We have started integrating the TRX file format into DIPY. The goal is to remove entirely the dependence of TRX on DIPY, while retaining much of the useful functionality in the many scripts and functions that are currently in the trx-python library directly into DIPY.

Work on this has started in #2759, which adds the trx-python library as a dependence of DIPY. This is a very light-weight dependency, and should help us move forward with the rest of the work described below.

To follow up, we would like to: 

- [ ] Port a few of the scripts from the trx-python library. Particularly: https://github.com/tee-ar-ex/trx-python/blob/master/scripts/tff_concatenate_tractograms.py, https://github.com/tee-ar-ex/trx-python/blob/master/scripts/tff_convert_tractogram.py, and https://github.com/tee-ar-ex/trx-python/blob/master/scripts/tff_validate_trx.py (which could be renamed as "validate_tractogram"). @skoudoro is in charge of this
- [ ] Port other library code in trx that depends directly on DIPY. @arokem is in charge of this.
- [ ] Clean up things in the trx-python library. @arokem and @frheault are in charge of this. 

But before we can move on with these, we should sort out remaining issues on #2759.

skoudoro:
Hi @arokem and @frheault,

What is the status of this? Do you have an ETA ?

skoudoro:
From the first item, All are on the last release apart from [tff_validate_trx.py](https://github.com/tee-ar-ex/trx-python/blob/master/scripts/tff_validate_trx.py). it will be in the release with the name `dipy_filter_tractogram`

samcoveney:
I have helped someone update their DiPy, and now we see the error `from trx import trx_file_memmap`. I recall that, after updating a while back, I had the same problem. In the end I just made a new conda environment and installed everything from scratch, because I couldn't fine a solution to the error. Any idea?

skoudoro:
> ImportError: cannot import name 'split_name_with_gz' from 'trx.utils'

We encounter this problem one time during the workshop.

it seems there is a incompatibility between a package named `trx` (not related to medical imaging) and our dependency named `trx-python`.

both use `import trx`, so when they are both installed, it picked the first package and not `trx-python`.

This is something to report in `trx-python` package. In our case, we just remove the other package and it works.

Let us know if it was the same issue @samcoveney

samcoveney:
[...]

PR #3027, opened 2024-01-04
[NF] Add Concatenate tracks workflows

skoudoro:
This PR is a follow up of #2760.  

It migrates the script `concatenate_tractogram` from TRX-Python library to DIPY.

More PRs should come to continues the migration.


skoudoro:
@frheault can you have a look ?

frheault:
That's great/useful to transfer! I haven't tested yet due to the fact I can't build Dipy right now.

Am I missing instructions since the new 3.12 update? I tried the good old `pip install -e .` and did not work so I tried:

```
1. pip install -r requirements/build.txt
2. pip install -e .
3. meson setup --reconfigure builddir/
4. meson -C compile builddir/
5. pip install -e .
```

And when I try to launch the script I get `FileNotFoundError: [Errno 2] No such file or directory: '/tmp/pip-build-env-ogw3qo_g/normal/local/bin/ninja'`. Am I missing a package? a `sudo apt install`? Or is there instructions I did not see?

When I got that figure out I will be able to run tests, try the script, etc. Also I think the script should be name **dipy_concatenate_tractograms**

skoudoro:
```
pip install -r requirements/build.txt
pip install --no-build-isolation -e .
```

should be enough to install DIPY.

Concerning the name, I might rename it just `dipy_concatenate` and allow to concatenate tractograms or volumes

frheault:
Thanks, I will try that!

About a script for both, it would be confusing to see parameters that applies to only tractograms or only volumes or some for both, no?

Like the datatype, the verification of header (can be turned off for tractography, but shape has to match for volume, or data_per_*, etc.) that would make the help twice as long and more confusing

dipy/workflows/io.py
```diff
@@ -257,3 +261,81 @@ def run(self, input_files, vol_idx=0, out_dir='',
...
+                has_group = True
+            trx_list.append(tractogram_obj)
+
+        trx = concatenate_tractogram(
+            trx_list, delete_dpv=delete_dpv, delete_dps=delete_dps,
+            delete_groups=delete_groups or not has_group,
+            check_space_attributes=check_space_attributes,
+            preallocation=preallocation)
+
+        valid_extensions = ['trk', 'trx', "tck,", "fib", "dpy", "vtk"]
```
frheault, line 335: There is a comma in between the quote for tck

skoudoro:
> About a script for both, it would be confusing to see parameters that applies to only tractograms or only volumes or some for both, no?

you are right. I was thinking in subcommand, like "dipy_concat. tractogram args" a bit like git which should avoid input parameter conflict/problem/confusion

But let's keep it simple for now.  I will update that. Thanks for the review

skoudoro:
Hi @frheault,

Did you succeed to install DIPY?  Did you have time to try this PR? 

Thank you in advance for you feedback

frheault:
Ah there is a missing comment, I forgot to click somewhere. 
Sorry about that.
I installed it correctly and tested the script. I tried 5 TRX/TRK/TCK -> TRX/TRK/TCK (all combinaison) 

The TRK had data_per_point and it was conserved correctly, also the header were alright and the position was matching the reference anatomy.

So it is working well!

skoudoro:
Thank you for the feedback!

I will wait until Friday in case someone else want to try it and then, I will go ahead and merge it

skoudoro:
working on `validate_tractogram` and `convert_tractogram`, I start to wonder if `concatenate_tractogram` should go in `dipy.tracking.utils` instead of `dipy.io.streamline`.

In general, it is not clear to me where all operations on Tractogram should go..... maybe a new module named `dipy.utils.streamline`

any opinion @frheault and @arokem?

frheault:
Recently in Scilpy we added like 5 modules and changed the name and position of everything because we were tired of these challenges, so I believe it is a great question.

We created a tractogram module for operation at the tractogram level as opposed to the streamlines level. In this situation, I do not like utils (because everything is ending in there). So without a tractogram module, I would say streamline (but it is not great either)

skoudoro:
Thank you for your review and feedback @frheault.

Issue #2789, opened 2023-04-25
Horizon image's dtype validation


guaje:
## Description

When using horizon 

```python
from dipy.viz.app import horizon
```

In a Python script, the data types are not validated compared to when used via the respective workflow. This makes sense since the tool was originally intended to be used from the command line, but now we are promoting its usage through Python code, and we should ensure we can account for such cases. Here is the traceback I managed to get:

```bash
Original shape: (5, 5, 5)
Segmentation fault (core dumped)
```

It seems the `actor.slicer()` only supports `float64`, `float32`, `int32` but not `int64`.

## Way to reproduce

- Code example:
```python
import numpy as np
from dipy.viz.horizon.app import horizon

horizon(images=[((10 * np.random.rand(5, 5, 5)).astype(np.int64), np.eye(4))])
```

- Operating system and version:
`Linux-5.15.108-1-MANJARO-x86_64-with-glibc2.35`
But it was initially reported originally in MX MacOS.

- Python version:
`Python 3.8.10`
But it was found in other versions too.

- dipy version: `1.7.0`

- dependency version (numpy, scipy, nibabel, h5py, cvxpy, fury)
    * import numpy; print("NumPy", numpy.__version__): `NumPy 1.24.3`
    * import scipy; print("SciPy", scipy.__version__): `SciPy 1.10.1`
    * import nibabel; print("Nibabel", nibabel.__version__): `Nibabel 5.1.0`
    * import h5py; print("H5py", h5py.__version__): `H5py 3.8.0`
    * import cvxpy; print("Cvxpy", cvxpy.__version__): `Cvxpy 1.3.1`
    * import fury; print("fury", fury.__version__): `fury 0.9.0`

skoudoro:
Can you look at this @maharshi-gor?

skoudoro:
fixed by https://github.com/dipy/dipy/pull/3031, closing

PR #3031, opened 2024-01-06
Bugfix: Horizon image's dtype validation

maharshi-gor:
This PR resolves the issue https://github.com/dipy/dipy/issues/2789 about checking dtype of the passed image arrays.

#### Current behavior
- When horizon is invoked from either we do not check the type of the data passed in the images. It is implicitly assumed to be an appropriate numerical type supported by horizon.
- This causes issues when users are using the horizon from the python code as the type is not verified.

#### Proposed behavior
All the images will be checked for dtype and the following cases can be the result.
1. Data is of type int32 or float32 or float64, data will be added to the valid images.
2. Data is of a different integer subtype than int32, data will be cast to int32 and then will be added to the valid images. There will be a warning provided.
3. Data is of a different floating subtype than float32 or float64, data will be cast to float64 and hen will be added to the valid images. There will be a warning provided.
4. Data is of non-numeric type, a warning will be provided and data will not be added to the valid images.

### DIFF
--- a/dipy/viz/horizon/util.py
+++ b/dipy/viz/horizon/util.py
@@ -1,9 +1,20 @@
+import warnings
+import numpy as np
+
+
 def check_img_shapes(images):
-    """
-    Checks if the images have same shapes
+    """Check if the images have same shapes.
+
+    Parameters
+    ----------
+    images : list
+
+    Returns
+    -------
+    boolean
+        True, if shapes are equal.
     """
 
-    # No need to check if there is less than 2 images
     if len(images) < 2:
         return True
     base_shape = images[0][0].shape[:3]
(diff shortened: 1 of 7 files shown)


dipy/viz/tests/test_apps.py
```diff
@@ -133,6 +134,29 @@ def test_horizon(rng):
             world_coords=True, interactive=False)
 
 
+def check_for_warnings(warn_printed, w_msg):
+    selected_w = [w for w in warn_printed if issubclass(w.category,
+                                                        UserWarning)]
+    assert len(selected_w) >= 1
+    msg = [str(m.message) for m in selected_w]
+    npt.assert_equal(w_msg in msg, True)
```
skoudoro, line 142: this function is duplicated. 

Maybe you should moved this function in `dipy.testing.__init__.py`

dipy/viz/tests/test_util.py
```diff
@@ -0,0 +1,78 @@
...
+    ]
+    npt.assert_equal(check_img_shapes(images), False)
+
+
+def check_for_warnings(warn_printed, w_msg):
+    selected_w = [w for w in warn_printed if issubclass(w.category,
+                                                        UserWarning)]
+    assert len(selected_w) >= 1
+    msg = [str(m.message) for m in selected_w]
+    npt.assert_equal(w_msg in msg, True)
```
skoudoro, line 34: duplicated, to move somewhere else

dipy/viz/horizon/util.py
```diff
@@ -12,3 +23,42 @@ def check_img_shapes(images):
         if base_shape != data.shape[:3]:
             return False
     return True
+
+
+def check_img_dtype(images):
+    """Check for supported dtype. If not supported numerical type, fallback to
```
skoudoro, line 29: after the first `.` you should return to a new line. see [here](https://github.com/dipy/dipy/blob/master/dipy/direction/peaks.py#L94-L131) for an example

skoudoro (commented):
Hi @maharshi-gor,

Looks good to me. See below a very small comment but this PR is ready to be merged. 

Thanks!

PR #3042, opened 2024-01-15
pyproject.toml: no cython at run time.

emollier:
As first identified in [Debian bug #1060438], the pyproject.toml declares a run time dependency on cython for dipy, but running the program and suite after cythonizing the code does not require cython anymore.  In such configuration, the cython requirement for the build system should be sufficient.

[Debian bug #1060438]: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1060438

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -37,7 +37,6 @@ classifiers = [
 ]
 dynamic = ['version']
 dependencies = [
-    "cython>=0.29.24, !=0.29.29",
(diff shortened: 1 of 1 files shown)


skoudoro:
Hi @emollier,

Make sense, thank you for catching that. Waiting for the CI's then we will go ahead and merge it.

skoudoro:
all green, merging

PR #3043, opened 2024-01-15
d/d/t/test_data.py: endian independent dtype.

emollier:
test_sphere_dtypes produces consistently little endian float64 numpy objects.  This is usually not an issue, but big endian systems, the little endian value is compared against a big endian float64, causing the test to fail.  This fixes/worksaround the test failure on big endian CPUs.

### DIFF
--- a/dipy/data/tests/test_data.py
+++ b/dipy/data/tests/test_data.py
@@ -5,4 +5,4 @@ import numpy as np
 def test_sphere_dtypes():
     for sphere_name, sphere_path in SPHERE_FILES.items():
         sphere_data = np.load(sphere_path)
-        npt.assert_equal(sphere_data['vertices'].dtype, np.float64)
+        npt.assert_equal(sphere_data['vertices'].dtype, np.dtype('<f8'))


skoudoro:
Thank you for doing this.

We might need to set up a Github Action with big endian systems to avoid those recurring issue (something using QEMU I suppose, I need to look into it - if you have an example, feel free to share! )

emollier:
I don't know how to bind the below on Github Actions, but maybe this will give you ideas.

To test on big endian, my go to platform is to use qemu-system-s390x coupled to binfmt kernel settings, provided by Debian packages qemu-system-misc and qemu-user-static.  Once installed, I can spawn a Debian s390x system root tree with e.g.:
```
# debootstrap --arch s390x sid /srv/chroot/sid-390x http://ftp.fr.debian.org/debian
```
Once done, I can chroot straight into the tree and, thanks to binfmt, qemu-system-s390x will "interpret" the s390x executables:
```
# uname -m
x86_64
# chroot /srv/chroot/sid-390x
# uname -m
s390x
```
Crude yet effective.  Main drawback is, that emulation for s390x is slow: running the test suite took some eight hours to run on my machine, to compare to one quarter of hour on the native cpu.

I'm using actually a configuration a bit more elaborate, but you get the idea.

skoudoro:
all green, merging. I will create an issue to remember the s390x github action (to do later during the year)

PR #3045, opened 2024-01-17
[DOC] fix `multi_shell_fiber_response` docstring array dims [ci skip]

ebrahimebrahim:
Fix array dimensions in the `multi_shell_fiber_response` docstring

### DIFF
--- a/dipy/reconst/mcsd.py
+++ b/dipy/reconst/mcsd.py
@@ -439,11 +439,12 @@ def multi_shell_fiber_response(sh_order, bvals, wm_rf, gm_rf, csf_rf,
     bvals : ndarray
         Array containing the b-values. Must be unique b-values, like outputted
         by `dipy.core.gradients.unique_bvals_tolerance`.
-    wm_rf : (4, len(bvals)) ndarray
-        Response function of the WM tissue, for each bvals.
-    gm_rf : (4, len(bvals)) ndarray
+    wm_rf : (N-1, 4) ndarray
+        Response function of the WM tissue, for each bvals,
+        where N is the number of unique b-values including the b0.
+    gm_rf : (N-1, 4) ndarray
         Response function of the GM tissue, for each bvals.
-    csf_rf : (4, len(bvals)) ndarray
+    csf_rf : (N-1, 4) ndarray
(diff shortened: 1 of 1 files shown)


dipy/reconst/mcsd.py
```diff
@@ -439,11 +439,11 @@ def multi_shell_fiber_response(sh_order, bvals, wm_rf, gm_rf, csf_rf,
     bvals : ndarray
         Array containing the b-values. Must be unique b-values, like outputted
         by `dipy.core.gradients.unique_bvals_tolerance`.
-    wm_rf : (4, len(bvals)) ndarray
+    wm_rf : (len(bvals)-1, 4) ndarray
```
skoudoro, line 442: not sure it is correct `-1`.  maybe `(len(nonzero(bvals)), 4)`. 

I have a doubt here

skoudoro (commented):
Hi @ebrahimebrahim, 

Thank you for this update. See below my comment

dipy/reconst/mcsd.py
ebrahimebrahim: agreed that `(len(nonzero(bvals)), 4)` better communicates the idea.

It seems that `bvals` is expected to have exactly one zero in it, because [it's expected to have been generated by `unique_bvals_tolerance`](https://github.com/dipy/dipy/blob/9153a852a511b07ffc141becfbd9a96ca42e9a90/dipy/reconst/mcsd.py#L441). 

it's allowed for it to not have that zero, but [with mild complaining](https://github.com/dipy/dipy/blob/9153a852a511b07ffc141becfbd9a96ca42e9a90/dipy/reconst/mcsd.py#L520)

what about writing `(N-1, 4) where N is the number of unique b-values including the b0 `
since the docstring later makes this definition with the notation `N`

skoudoro: > what about writing (N-1, 4) where N is the number of unique b-values including the b0 
since the docstring later makes this definition with the notation N

I am ok with that 👍

ebrahimebrahim: How's this?

skoudoro:
Thanks @ebrahimebrahim , merging

PR #3055, opened 2024-02-01
[RF] replace Bunch by Enum

skoudoro:
This is a small PR to replace our custom class `Bunch` by the standard object `Enum`.  

`Bunch` was created because `Enum` was not available with python2. It is not anymore the case.


skoudoro:
Finally, the solution with Enum is not better. You need to implement comparison which is a bit overkill for what we need.

keeping as it is and closing

PR #3056, opened 2024-02-01
[WIP][RF] Use lazy loading

skoudoro:
Proof of concept for adopting [Scientific Python SPEC 1](https://discuss.scientific-python.org/t/spec-1-lazy-loading-for-submodules/25).

> This has several advantages:
>   - It exposes a nested namespace that behaves as a flat namespace. This avoids carefully having to import exactly the right combination of submodules, and allows interactive exploration of the namespace in an interactive terminal.
>   - It avoids having to optimize for import cost. Currently, developers often move imports inside of functions to avoid slowing down importing their module. Lazy importing, when implemented through out a library, makes all imports cheap.
>   - It provides direct access to submodules, avoiding local namespace conflicts. Instead of doing `import dipy.reconst as reconst` to avoid clobbering a local `reconst`, one can now import each library and access its members directly: `import dipy; dipy.reconst.`

# Todo

- [x] Add CI run that uses `EAGER_IMPORT=true` env var to smoke out circular import issues
- [ ] Convert all `__init__.py`

Feedbacks are welcomed

When this PR will be done, a follow-up PR will be created to simplify all tutorials import

### DIFF
--- a/.github/workflows/test_template.yml
+++ b/.github/workflows/test_template.yml
@@ -66,6 +66,7 @@ jobs:
       PYTHON_VERSION: ${{ matrix.python-version }}
       VENV_ARGS: "--python=python"
       USE_PRE: ${{ inputs.use-pre }}
+      EAGER_IMPORT: true
       COVERAGE: ${{ inputs.coverage }}
       CODECOV_TOKEN: ${{ secrets.codecov-token }}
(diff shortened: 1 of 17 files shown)


skoudoro:
superseded by #3288. 

Closing

PR #3059, opened 2024-02-05
Generalize special casing while loading bvecs, to include the case of transposed 2,3 vectors

johnccshen:


skoudoro:
Thank you for this @johnccshen!

Can you explain more your specific case and can you add unit tests for this case. You can look at some unitest examples : https://github.com/dipy/dipy/blob/master/dipy/core/tests/test_gradients.py

Thanks !

johnccshen:
Hi @skoudoro ,

When there are only 2 bvecs in the array, the array is unexpectedly transposed.
That says we have a bvec array with shape (2,3). The logic here ([Link](https://github.com/dipy/dipy/pull/3059/files#diff-bbd3ad0756880932ac2142d609f873674431fa6ad7a187ae379bba84ab2ec06fR707))

if (bvecs.shape[1] > bvecs.shape[0]) and bvecs.shape[0] > 1 -> converts the the bvec array to (3,2)

Though I know that in practical, it's rare to see only 2 vecs in the bvec array.

skoudoro:
Hi @arokem,

Can you give your opinion concerning this PR? Not sure we want to address all small and specific cases which increase code maintenance and complexity.

Hi @johnccshen,

What is the status of the unit test? will you find time to finalize this? Thank you for your feedback.

pjsjongsung:
I don't think there should be an exception like that, but it seems to be important to figure out why that if statement is there in the first place.

johnccshen:
HI all, 

Sorry for the late response, I was on Chinese Lunar New Year and just back online.

I accidentally sent a DTI image with data shape (256, 256, 256, 2) into the fwdti model. 
The popup error was irrelevant, then I realized that the shape of gtab.bvec was unexpectedly converted to (3,2).

Looking at the unit test I added might give you more sense of what I am trying to describe.

pjsjongsung:
This makes sense if the reason for the if statement was to catch transposed bvecs, but I think we still need to make sure that the if statement was indeed there for that particular reason.

arokem:
I don't think this complicates the code or increases its complexity. I'm +1 for this change.

arokem:
BTW: I changed the title of PR, because the word "skip" confused me.

johnccshen:
> This makes sense if the reason for the if statement was to catch transposed bvecs, but I think we still need to make sure that the if statement was indeed there for that particular reason.

Hi I believe the change I made is the same as what it is just being more generalized. 

@arokem , Thanks for updating the title. It's a better understanding now.

However, I found that I broke the unit test. 
I had a hard time setting up the environment in my local machine. Do we have any documentation for developers to reference?

johnccshen:
@skoudoro Thank you! I rebased to the master.

skoudoro (approved):
Thank you for the rebase @johnccshen, 

merging

PR #3063, opened 2024-02-09
[RF] remove cpdef in PmfGen

skoudoro:
This PR is a small refactoring to remove the `cpdef` in  pmf.pyx.   

`cpdef = def + cdef` , so it does not make sense to have cpdef + cdef.     it is either `def +cdef ` or `cpdef`

I also use the opportunity to replace some pointer by memoryview. When using memoryview, we should make sure to use aligned memory by using [::1] for speed reason.

@gabknight, can you have a look and try?


gabknight:
LGTM. This makes the code much nicer, thanks. Quick testing show no performance changes.

skoudoro:
Thank you for the feedback! So I am going ahead and merge it

PR #3070, opened 2024-02-20
[DOC] Fix installation link in README  [ci skip]

pjsjongsung:
This PR is to simply change the broken installation link in the README file.


skoudoro:
Good catch! Thanks @pjsjongsung!

PR #3073, opened 2024-02-22
[NF] Add DSI workflow

skoudoro:
This PR just add DSI Workflow. 

You can test it by run this cli on `taiwan_ntu_dsi` DIPY Dataset


skoudoro:
Not sure why so many tests are failing. 

Failing tests are not related to this PR.

Its seems related to `setup_module`and `teardown_module`.

I also see that there is a new release of Pytest (from last week). I will try to reproduce that locally.

PR #3083, opened 2024-02-26
Np.unique check removed.

maharshi-gor:
This PR introduces the change to the `np.unique` check whether an image is binary or not.

#### Issue
- This creates a lot of overhead on the system as the images can be in the size of gigabytes.

#### Solution
- A function is introduced to check an image is binary or not.
- This function samples 100 points (default, can be changed via parameter) and verifies if the image provided is a binary image.

### DIFF
--- a/dipy/viz/horizon/app.py
+++ b/dipy/viz/horizon/app.py
@@ -11,7 +11,8 @@ from dipy.viz.gmem import GlobalHorizon
 from dipy.viz.horizon.tab import (ClustersTab, PeaksTab, ROIsTab, SlicesTab,
                                   TabManager, build_label)
 from dipy.viz.horizon.visualizer import ClustersVisualizer, SlicesVisualizer
-from dipy.viz.horizon.util import check_img_dtype, check_img_shapes, is_binary_image
+from dipy.viz.horizon.util import (check_img_dtype, check_img_shapes,
+                                   is_binary_image)
 
 fury, has_fury, setup_module = optional_package('fury', min_version="0.9.0")
 
--- a/dipy/viz/horizon/app.py
+++ b/dipy/viz/horizon/app.py
@@ -11,7 +11,7 @@ from dipy.viz.gmem import GlobalHorizon
 from dipy.viz.horizon.tab import (ClustersTab, PeaksTab, ROIsTab, SlicesTab,
                                   TabManager, build_label)
 from dipy.viz.horizon.visualizer import ClustersVisualizer, SlicesVisualizer
-from dipy.viz.horizon.util import check_img_dtype, check_img_shapes
+from dipy.viz.horizon.util import check_img_dtype, check_img_shapes, is_binary_image
 
 fury, has_fury, setup_module = optional_package('fury', min_version="0.9.0")
 
@@ -451,8 +451,7 @@ class Horizon:
             for img in self.images:
                 data, affine = img
                 self.vox2ras = affine
-                img_val_range = np.unique(data).shape[0]
-                if img_val_range == 2:
+                if is_binary_image(data):
                     if self.__roi_images:
                         if 'rois' in self.random_colors:
                             roi_color = next(self.color_gen)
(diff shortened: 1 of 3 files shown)


dipy/viz/horizon/util.py
```diff
@@ -63,3 +63,32 @@ def check_img_dtype(images):
...
+        indices.append(
+            np.random.choice(
+                np.arange(0, dim - 1),
+                replace=dim <= unique_points,
+                size=(unique_points)
+                )
+            )
+
+    indices = np.asarray(indices)
+    return np.unique(data[*indices]).shape[0] <= 2
```
skoudoro, line 94: Not sure to understand the goal of this PR since you are still using `np.unique` here. I also do not understand the initial request.

pjsjongsung:
I think one of the concern was that np.unique might be too heavy on a big data. I think we should check whether if it is binary **if the roi argument** is True, not the other way around. Especially since we are repeating the same code right now (code for roi but not binary = code for not binary)

dipy/viz/horizon/util.py
Garyfallidis: He is simply checking if an image is binary. Now only running unique for 100 values. Not the entire image.

Garyfallidis:
It is a bit more complicated. The issue is combining data and binary masks and roi masks.

skoudoro:
Ok, I see. indeed, the logic remain a bit strange and I think it needs update.

dipy/viz/horizon/util.py
skoudoro: why not `np.histogram`?

dipy/viz/horizon/util.py
```diff
@@ -63,3 +63,32 @@ def check_img_dtype(images):
...
+    Returns
+    -------
+    boolean
+        Whether the image is binary or not
+    """
+    indices = []
+
+    for dim in data.shape:
+        indices.append(
+            np.random.choice(
```
skoudoro, line 86: Numpy suggest the migration from `numpy.random` to `numpy.random.Generator`. All the codebase have been migrated. Can you do it here also? 

Thank you

dipy/viz/horizon/util.py
skoudoro: ok, for now

dipy/viz/tests/test_util.py
```diff
@@ -69,3 +70,16 @@ def test_check_img_dtype():
         check_img_dtype(images)
         check_for_warnings(l_warns, 'skipping image 1, passed image is not in '
                            + 'numerical format')
+
+
+def test_is_binary_image():
```
pjsjongsung, line 75: Serge's comment on random also applies here. Looking at some test files (e.g. dipy/core/tests/test_gradients.py) should tell you what you need to do

dipy/viz/horizon/app.py
```diff
@@ -451,8 +452,7 @@ def build_show(self, scene):
             for img in self.images:
                 data, affine = img
                 self.vox2ras = affine
-                img_val_range = np.unique(data).shape[0]
-                if img_val_range == 2:
+                if is_binary_image(data):
```
pjsjongsung, line 455: > I think one of the concern was that np.unique might be too heavy on a big data. I think we should check whether if it is binary **if the roi argument** is True, not the other way around. Especially since we are repeating the same code right now (code for roi but not binary = code for not binary)

I think this logic still stands here regardless of why we need the binary image.

maharshi-gor:
Thank you so much @pjsjongsung @Garyfallidis @skoudoro  for your valuable input.

I have resolved all the mentioned comments

dipy/viz/horizon/app.py
maharshi-gor: @pjsjongsung If you can create an issue for me that would be amazing! so we can keep track of it.

dipy/viz/horizon/util.py
maharshi-gor: @skoudoro np.histogram will try to scan the entire data which is almost equal to checking np.unique, here we are simply generating 100 indices for the data and testing them.

Issue #2970, opened 2023-11-08
spherical harmonic degree/order terminology swapped 


ebrahimebrahim:
For functions related to spherical harmonics, I noticed that the terminology "degree" and "order" is often used backwards in docstrings and variable names.

- in [this function](https://github.com/dipy/dipy/blob/b5efe4f8cd2fd4d88c0a5c08936f1b4ab314f8f0/dipy/reconst/shm.py#L578) for example `m` should be order and `n` should be degree
- in many places the maximum `n` (sometimes called $\ell$) value is a function argument called `sh_order`. it should be called `sh_degree` (or better `sh_degree_max`)

---

For a spherical harmonic $Y^m_n$ (sometimes called $Y^m_\ell$) , the $n$ ($\ell$)  is called _degree_ and the $m$  is called order.
- see terminology for [spherical harmonics](https://en.wikipedia.org/wiki/Spherical_harmonics)
- also see terminology for the [associated legendre polynomials](https://en.wikipedia.org/wiki/Associated_Legendre_polynomials#Definition_for_non-negative_integer_parameters_%E2%84%93_and_m) that show up up in the definition of spherical harmonics
- in [mrtrix docs](https://mrtrix.readthedocs.io/en/latest/concepts/spherical_harmonics.html#what-are-spherical-harmonics) we do sometimes see "order" used for $n$ (or $\ell$) but in that context the term _phase_ is then used for $m$.

itellaetxe:
Hi,

I am Iñigo Tellaetxe, newcomer, and I would like to solve this issue.
I read the resources you included, and it does not seem to be very difficult to tackle. Can I get it assigned?

Thanks,

Iñigo

skoudoro:
Yes, you can go ahead @itellaetxe,

thanks

itellaetxe:
Thanks!
I have been reviewing the issue in more detail, and I have a couple of suggestions. I'd like to know your opinion before proceeding:

- Since `m` (order) and `n` (degree) are 1) visually similar symbols 2) they are subsequent in the alphabet, I would stop using `n` for the degree, and I would stay with `l` for the degree in all cases.
- _order_ and _degree_ are not very long words, so instead of using symbols like `m` and `l`, we could simply write the words themselves, for most (conceptual) clarity. It could result in slightly longer lines of code, though.

What are your thoughts on this? @ebrahimebrahim

ebrahimebrahim:
- I agree that `m,l` is better than `m,n`. Good idea!
[...]

PR #3086, opened 2024-02-28
Fix spherical harm terminology swap

itellaetxe:
## Summary
For functions related to spherical harmonics, the terminology "degree" and "order" is often used backwards in docstrings and variable names. This PR ensures consistency when referring to "degree" and "order" in the code.
Closes Issue #2970 

## Main changes
- Standardized the symbols for the spherical harmonic concepts of order and degree with m_value and l_value, respectively. In case these are arrays, we use m_values and l_values to be more informative of the type of the variable.
- Changed sh_order to sh_degree_max as @ebrahimebrahim suggested in the issue.
- Updated the examples according to the new input argument names (sh_degree_max).


itellaetxe:
My linter was not showing me the E501 problem. Fixing it. Sorry

skoudoro:
Hi @itellaetxe,

We fixed the issue in master,  Can you rebase this branch or merge master to this branch?

Thanks

CHrlS98:
Hi @skoudoro @itellaetxe @ebrahimebrahim,
The terminology swap was introduced in a previous PR of mine (#2206). We also briefly discussed the swap here https://github.com/dipy/dipy/pull/2206#pullrequestreview-484290448. However I now realize that what we called the "literature" at the time is actually limited to the diffusion MR literature and from a mathematician's/physicist's point-of-view it might be wrong. Still, both bases we support in DIPY (`descoteaux07` and `tournier07`) use the term _order_ to describe $\ell$ in the cited papers.

If we do proceed with this change we might want to add a note somewhere in the doc clearly explaining the issue. It looks like the dMRI community have settled on the terminology **order** and all the papers I came across using CSD reconstruction report the maximum SH **order** when they talk about $\ell_{max}$ (although it is true that the maximum SH order and degrees will be equal). 

This is very confusing and unclear to me why there is a swap between the dMRI literature and wikipedia. I'll also tag @karanphil here who might have something to add.

karanphil:
> Hi @skoudoro @itellaetxe @ebrahimebrahim, The terminology swap was introduced in a previous PR of mine (#2206). We also briefly discussed the swap here https://github.com/dipy/dipy/pull/2206#pullrequestreview-484290448. However I now realize that what we called the "literature" at the time is actually limited to the diffusion MR literature and from a mathematician's/physicist's point-of-view it might be wrong. Still, both bases we support in DIPY (`descoteaux07` and `tournier07`) use the term _order_ to describe ℓ in the cited papers.
> 
> If we do proceed with this change we might want to add a note somewhere in the doc clearly explaining the issue. It looks like the dMRI community have settled on the terminology **order** and all the papers I came across using CSD reconstruction report the maximum SH **order** when they talk about ℓmax (although it is true that the maximum SH order and degrees will be equal).
> 
> This is very confusing and unclear to me why there is a swap between the dMRI literature and wikipedia. I'll also tag @karanphil here who might have something to add.

It is indeed confusing. I remember being confused at first, coming from a physics background. However, when you read the literature on SH bases and CSD in dMRI (like tournier04, tournier07, descoteaux07 or descoteaux09), they all refer to `l` (or `n` in the case of Tournier/MRtrix) as the order of the SH function.

While Wikipedia follows the inverse terminology, other references such as Brilliant (https://brilliant.org/wiki/spherical-harmonics/) follow the current way of Dipy and the dMRI literature.

I think we need to be very careful if changing this.

skoudoro:
Thank you for your feedback @CHrlS98. Indeed, Looking back https://github.com/dipy/dipy/pull/2206 reminds me that I have already this question as a comment.

@Garyfallidis and @arokem, Can you also give your opinion here? it is important. thank you! 

maybe @RafaelNH could provide also a feedback

ebrahimebrahim:
> It looks like the dMRI community have settled on the terminology **order** and all the papers I came across using CSD reconstruction report the maximum SH **order** when they talk about $\ell_\text{max}$ (although it is true that the maximum SH order and degrees will be equal).

Hmm this is more confusing that I thought when I opened #2970

Sorry if I opened a can of worms that should have remained closed :grimacing: 

(In mrtrix "order" is used for $\ell$ but then it's not "degree" but "_phase_" that is used for $m$. So while using "order" for $\ell$ could match some dmri community conventions, I am questioning whether _both_ $\ell$=order _and_ $m$=degree is ever used together. IDK anymore...)

itellaetxe:
> We fixed the issue in master, Can you rebase this branch or merge master to this branch?

Hi @skoudoro, due to the ongoing discussion, I think I will leave it like this until we arrive to a conclusion. I will adapt the code to the final decision.

arokem:
Hi! I think that we should use the terminology that is used in the dMRI papers. I would suggest putting in a Note that explains that this is a choice that we made, so readers of wikipedia and scipy should be careful when relying on these as resources. Does that makes sense?

ebrahimebrahim:
Do dMRI papers swap degree and order or do they just use order for $\ell$ and then something different (like phase) for m?

CHrlS98:
In Maxime Descoteaux (@mdesco) PhD thesis, there is a chapter on spherical harmonics (section 5.2) where it is said that $\ell$ is the order and $m$, the degree (https://theses.hal.science/tel-00457458):
image
I am however having a hard time finding a peer-reviewed journal article referring to $m$ as the degree.

In Tournier *et al*, 2004 (https://doi.org/10.1016/j.neuroimage.2004.07.037), they refer to $\ell$ as the order, but $m$ is indeed called a phase factor:
> Each spherical harmonic can be denoted using two numbers: its harmonic order n (n ≥ 0) and phase factor m (−n ≤ m ≤ n).

Actually, even in the DIPY paper from 2014 (https://doi.org/10.3389/fninf.2014.00008), $m$ is called the phase:
> The SH of order l and phase m, Yml(θ, ϕ), arises from the angular solution to Laplace's equation in spherical coordinates and they form an orthonormal basis for complex functions defined on the unit sphere. 

So, while there appears to be a consensus on $\ell$, I'm not really sure what we should call $m$ anymore...

arokem:
@desco also used the term "phase factor" in [his seminal 2007 paper]:(https://onlinelibrary.wiley.com/doi/pdf/10.1002/mrm.21277)

Screenshot 2024-03-04 at 8 30 25 PM

I think part of the confusion stems from the fact that the words "order" and "degree"  are used interchangeably to refer to the highest exponent in a polynomial function (see e.g., [here](https://math.stackexchange.com/questions/3323294/degree-and-order-of-a-polynomial)). So, a nomenclature that uses "degree" and "phase factor" is potentially much less confusing. In fact, one could argue that if you ever use "order" that should refer to the variable referred to as $\ell$ in this writing of the equation. 

In other words, potentially with a lot of explanation, I might suggest to go with `order` and `phase_factor` (I think that just `phase` may be confusing, because it's an integer and phases are usually thought of as continuous things.

skoudoro:
> In other words, potentially with a lot of explanation, I might suggest to go with order and phase_factor (I think that just phase may be confusing, because it's an integer and phases are usually thought of as continuous things.

+1 with this conclusion. Thanks all for the debate / clarification, I think we have a consensus.

@itellaetxe, can you go ahead with  `order` (for $\ell$) and `phase_factor / phase` for $m$  in the code. Concerning the documentation (rst file), you can refer to this PR for the discussion and add some additional note.

I plan to release DIPY tomorrow or in 2 days so it will be great if you could update this PR on ASAP. Thank you in advance!

ps: it is ok if you can not make it.

skoudoro:
PS: do not forget to fix the conflict and pep8 issues above. Thanks again !

itellaetxe:
Thanks to everyone that took part in the discussion. Shortly:

- I followed the indications given by @arokem and @skoudoro, regarding the new namings.
- `sh_degree_max` is now `sh_order_max`.
- Included the mentioned note in the theory rst file explaining our convention choice.
- Fixed a couple of bad namings I overlooked the first time I opened the PR.
- Passed tests locally.
- Rebased my branch onto master.


I just saw that I made a typo with `order` in a couple of locations. Fixing and pushing now.

doc/theory/sh_basis.rst
```diff
@@ -161,6 +160,15 @@ are also taken into account when reconstructing a spherical function. These
...
 
+NOTE: 
+The definition of spherical harmonics that DIPY utilizes does not match the one
+in Wikipedia and scipy. Instead, DIPY follows the dMRI literature conventions,
+like in ``descoteaux07`` and ``tournier07``.
+The code in DIPY also follows the following convention:
+Let the SH be noted as $Y_{l}^m$. Then, $l$ is referred to as either order or
+l_value(s), and $m$ is referred to as either phase factor or m_value(s).
+This decisions were made as a result of the PR in https://github.com/dipy/dipy/pull/3086
+
```
skoudoro, line 171: Perfect !

karanphil:
I am questioning the change from `sh_order` to `sh_order_max`. This would require at least deprecated versions for all functions with `sh_order=` as optional argument. For instance, `reconst.shm.sf_to_sh` will break now for any users using `sh_order=`.

I understand `sh_order_max` is more accurate, but is it really a necessary change?

skoudoro:
> I understand sh_order_max is more accurate, but is it really a necessary change?

Let's try to be the most accurate as possible but you got a very good point @karanphil 

@itellaetxe, everytime you change from `sh_order` to `sh_order_max` in a function argument, you need to add the decorator below:

```python
@deprecated_params('sh_order', 'sh_order_max', since='1.9', until='2.0')
```

see an example: https://github.com/dipy/dipy/blob/379062e3704088bcc998e3736a52212ca41627f5/dipy/align/imaffine.py#L249

doc/theory/sh_basis.rst
```diff
@@ -161,6 +160,15 @@ are also taken into account when reconstructing a spherical function. These
...
 symmetric signals.
 
+NOTE: 
+The definition of spherical harmonics that DIPY utilizes does not match the one
+in Wikipedia and scipy. Instead, DIPY follows the dMRI literature conventions,
+like in ``descoteaux07`` and ``tournier07``.
+The code in DIPY also follows the following convention:
+Let the SH be noted as $Y_{l}^m$. Then, $l$ is referred to as either order or
+l_value(s), and $m$ is referred to as either phase factor or m_value(s).
+This decisions were made as a result of the PR in https://github.com/dipy/dipy/pull/3086
```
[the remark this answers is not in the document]
karanphil: Since this PR is cited there, I would also update the section "Main changes" (at the beginning of the PR) to represent the actual final changes. It would be less confusing if someone looks at it in the future.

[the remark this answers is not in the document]
itellaetxe: Thank you both for the suggestions :)

- @karanphil I just updated the beginning of the PR. It should reflect the current changesnow.
- @CHrlS98 thanks for the grammar check. I will leave it as _"These decisions..."_. Pushing it in the next commit.

itellaetxe:
> @itellaetxe, everytime you change from `sh_order` to `sh_order_max` in a function argument, you need to add the decorator below:

It should be good to go now!

dipy/data/__init__.py
```diff
@@ -261,7 +261,7 @@ def mrtrix_spherical_functions():
     -------
     func_coef : array (2, 3, 4, 45)
         Functions represented by the coefficients associated with the
-        mrtrix spherical harmonic basis of order 8.
+        mrtrix spherical harmonic basis of order (l) 8.
```
karanphil, line 264: Should it be "basis of maximal order"? Following the change from `sh_order` to `sh_order_max`.

dipy/reconst/csdeconv.py
```diff
@@ -150,15 +150,15 @@ def __init__(self, S0, dwi_response, bvalue=None):
         self.S0 = S0
         self.dwi_response = dwi_response
         self.bvalue = bvalue
-        self.m = np.zeros(len(dwi_response))
-        self.sh_order = 2 * (len(dwi_response) - 1)
-        self.n = np.arange(0, self.sh_order + 1, 2)
+        self.m_value = np.zeros(len(dwi_response))
+        self.sh_order_max = 2 * (len(dwi_response) - 1)
+        self.l_value = np.arange(0, self.sh_order_max + 1, 2)
```
karanphil, line 155: `self.l_value` and `self.m_value` are arrays here, shouldn't they have `s` at the end?

dipy/reconst/forecast.py
```diff
@@ -79,8 +80,8 @@ def __init__(self,
         ----------
         gtab : GradientTable,
             gradient directions and bvalues container class.
-        sh_order : unsigned int,
-            an even integer that represent the SH order of the basis (max 12)
+        sh_order_max : unsigned int,
+            an even integer that represent the SH order (l) of the basis (max 12)
```
karanphil, line 84: maximal SH order

dipy/reconst/mcsd.py
```diff
@@ -83,69 +86,70 @@ def __init__(self, response, sh_order, shells, S0=None):
...
     """Used to inflate the response for the `multiplier_matrix` in the
     `MultiShellDeconvModel`.
     Parameters
     ----------
     response : MultiShellResponse object
     gtab : GradientTable
-    n : int ``>= 0``
-        The degree of the harmonic.
+    sh_order_max : int ``>= 0``
+        The order (l) of the harmonic.
```
karanphil, line 109: maximal order

dipy/reconst/shm.py
```diff
@@ -162,21 +163,22 @@ def gen_dirac(m, n, theta, phi, legacy=True):
...
     This may take scalar or array arguments. The inputs will be broadcast
     against each other.
 
     Parameters
     ----------
-    m : int ``|m| <= n``
-        The degree of the harmonic.
-    n : int ``>= 0``
-        The order of the harmonic.
+    m_value : int ``|m| <= l``
```
karanphil, line 178: I believe that most of the `m_value` and `l_value` in `reconst.shm` are arrays. The old docstring might be erroneous. It might be "array of int" instead of "int".

dipy/reconst/shm.py
```diff
@@ -409,33 +411,33 @@ def real_sh_tournier(sh_order, theta, phi,
            framework for medical image processing and visualisation.
            NeuroImage. 2019 Nov 15;202:116-137.
     """
-    m, n = sph_harm_ind_list(sh_order, full_basis)
+    m_value, l_value = sph_harm_ind_list(sh_order_max, full_basis)
```
karanphil, line 414: These are arrays I believe.

dipy/reconst/shm.py
```diff
@@ -747,26 +755,28 @@ def sampling_matrix(self, sphere):
...
         """Creates a model that can be used to fit or sample diffusion data
 
         Parameters
         ----------
         gtab : GradientTable
             Diffusion gradients used to acquire data
-        sh_order : even int >= 0
-            the spherical harmonic order of the model
+        sh_order_max : even int >= 0
+            the spherical harmonic order (l) of the model
```
karanphil, line 779: maximal

karanphil (commented):
I noted a couple of places where `l/m_value` should be `l/m_values`, and where the docstring is incomplete. Other than that, it looks good to me, great job!

dipy/reconst/csdeconv.py
```diff
@@ -463,17 +466,17 @@ def estimate_response(gtab, evals, S0):
     return single_tensor(gtab, S0, evals, evecs, snr=None)
 
 
-def forward_sdt_deconv_mat(ratio, n, r2_term=False):
+def forward_sdt_deconv_mat(ratio, l_values, r2_term=False):
```
skoudoro, line 469: here, it is missing `@deprecated_params('n', 'l_values', since='1.9', until='2.0')

dipy/reconst/shm.py
```diff
@@ -56,18 +57,18 @@ def bandit(f):
     return bandit
 
 
-def forward_sdeconv_mat(r_rh, n):
+def forward_sdeconv_mat(r_rh, l_values):
```
skoudoro, line 60: it is missing `@deprecated_params('n', 'l_values', since='1.9', until='2.0')`

dipy/reconst/shm.py
```diff
@@ -201,42 +203,42 @@ def spherical_harmonics(m, n, theta, phi, use_scipy=True):
...
+    val = val * np.exp(1j * m_values * theta)
     return val
 
 
 @deprecate_with_version('dipy.reconst.shm.real_sph_harm is deprecated, '
                         'Please use '
                         'dipy.reconst.shm.real_sh_descoteaux_from_index '
                         'instead', since='1.3', until='2.0')
-def real_sph_harm(m, n, theta, phi):
+def real_sph_harm(m_values, l_values, theta, phi):
```
skoudoro, line 224: same here

dipy/reconst/shm.py
```diff
@@ -245,33 +247,33 @@ def real_sph_harm(m, n, theta, phi):
...
     See Also
     --------
     scipy.special.sph_harm
     """
-    return real_sh_descoteaux_from_index(m, n, theta, phi, legacy=True)
+    return real_sh_descoteaux_from_index(m_values, l_values, theta, phi, legacy=True)
 
 
-def real_sh_tournier_from_index(m, n, theta, phi, legacy=True):
+def real_sh_tournier_from_index(m_values, l_values, theta, phi, legacy=True):
```
skoudoro, line 259: same here, see above

itellaetxe:
@skoudoro, I completely overlooked the deprecator decorator in the `m` and `n` cases. I will fix it. Sorry!

dipy/reconst/shm.py
```diff
@@ -297,35 +299,35 @@ def real_sh_tournier_from_index(m, n, theta, phi, legacy=True):
...
-        real_sh *= np.where(m == 0, 1., np.sqrt(2))
+        real_sh *= np.where(m_values == 0, 1., np.sqrt(2))
     else:
         warn(tournier07_legacy_msg, category=PendingDeprecationWarning)
 
     return real_sh
 
 
-def real_sh_descoteaux_from_index(m, n, theta, phi, legacy=True):
+def real_sh_descoteaux_from_index(m_values, l_values, theta, phi, legacy=True):
```
skoudoro, line 314: same here, see above

skoudoro (commented):
Thank you for this great work @itellaetxe. 

This PR is almost ready to go. See my additional comments.

is there any additional comments @CHrlS98 ? @karanphil ?  @arokem ? @ebrahimebrahim ?

skoudoro:
If not, I might edit this PR, let the CI's run and merge it today to be part of the release.

skoudoro:
Ok, all seems ok, merging!

Thank you @itellaetxe for this work

Thank you @arokem @CHrlS98 @karanphil @ebrahimebrahim  for the review/debate

itellaetxe:
Thank you @skoudoro for the help with those last decorators!
Also thank you for the patience and the good advice to everyone that participated in the discussion. I will try to do it better next time :)

Issue #1404, opened 2018-01-20
RF - DirectionGetter.get_direction function return


gabknight:
The function `get_direction(.)` of the `DirectionGetter` class currently returns False if it succeeds in updating the direction. `get_direction(.)` returns 0 if there is no error and 1 otherwise. Since we don't have various error code here, I suggest returning 1 if it succeeds in updating the direction, 0 if it fails.

Another option would be to rename the function to something like `update_direction(.)`, so it is a bit more intuitive as it uses the 0/non-0 success/error-code as returning value.

Briefly discussed in #1384. 

What do you think?

skoudoro:
closing due to new architecture.  Feel free to reopen it if you plan to work on this.

PR #3089, opened 2024-02-29
WIP - Parallel Tracking Framework

gabknight:
**WIP**

Re-implementation of dipy's tracking algorithms using cython `prange` function over all seeds. This allows for efficient and transparent use of multi-threading for tracking.

### DIFF
--- a/dipy/tracking/fast_tracking.pyx
+++ b/dipy/tracking/fast_tracking.pyx
@@ -27,13 +27,13 @@ from libc.time cimport time, time_t
 
 
 #need cpdef to access cdef functions?
-cpdef list generate_tractogram(double[:,::1] seed_positons,
+cpdef list generate_tractogram(double[:,::1] seed_positions,
                                double[:,::1] seed_directions,
                                StoppingCriterion sc,
                                ProbabilisticTrackingParameters params,
                                PmfGen pmf_gen):
     cdef:
-        cnp.npy_intp _len=seed_positons.shape[0]
+        cnp.npy_intp _len=seed_positions.shape[0]
         double[:,:,:] streamlines_arr
         double[:] status_arr
         cnp.npy_intp i, j
@@ -44,7 +44,7 @@ cpdef list generate_tractogram(double[:,::1] seed_positons,
     status_arr =  np.array(np.zeros((_len)), order='C')
     print("1")
     t1 = time(NULL)
-    generate_tractogram_c(seed_positons, seed_directions, _len, sc, params,
+    generate_tractogram_c(seed_positions, seed_directions, _len, sc, params,
                           pmf_gen, &probabilistic_tracker, streamlines_arr, status_arr)
     print("2")
     t2 = time(NULL)
@@ -64,7 +64,7 @@ cpdef list generate_tractogram(double[:,::1] seed_positons,
     return streamlines
 
 
-cdef int generate_tractogram_c(double[:,::1] seed_positons,
+cdef int generate_tractogram_c(double[:,::1] seed_positions,
                                double[:,::1] seed_directions,
                                int nbr_seeds,
                                StoppingCriterion sc,
@@ -76,7 +76,6 @@ cdef int generate_tractogram_c(double[:,::1] seed_positons,
     cdef:
         cnp.npy_intp i, j, k
 
-
     for i in prange(nbr_seeds, nogil=True):
     #for i in range(nbr_seeds):
         stream = <double*> malloc((params.max_len * 3 * 2 + 1) * sizeof(double))
@@ -86,7 +85,7 @@ cdef int generate_tractogram_c(double[:,::1] seed_positons,
         for j in range(params.max_len * 3 * 2 + 1):
             stream[j] = 0
 
-        status[i] = generate_local_streamline(&seed_positons[i][0],
+        status[i] = generate_local_streamline(&seed_positions[i][0],
                                               &seed_directions[i][0],
                                               stream,
                                               tracker,
@@ -130,7 +129,7 @@ cdef int generate_local_streamline(double* seed,
     # forward tracking
     stream_status_forward = TRACKPOINT
     for i in range(1, params.max_len):
-        if probabilistic_tracker(&point[0], &voxdir[0], params, pmf_gen):  # , pmf_gen_func):
+        if tracker(&point[0], &voxdir[0], params, pmf_gen):  # , pmf_gen_func):
             break
         # update position
         for j in range(3):
@@ -150,7 +149,7 @@ cdef int generate_local_streamline(double* seed,
         voxdir[j] = voxdir[j] * -1
     stream_status_backward = TRACKPOINT
     for i in range(1, params.max_len):
-        if probabilistic_tracker(&point[0], &voxdir[0], params, pmf_gen):
+        if tracker(&point[0], &voxdir[0], params, pmf_gen):
             break
         # update position
         for j in range(3):
@@ -316,10 +315,10 @@ cdef int deterministic_maximum_tracker(double* point,
     return 1
 
 #get_direction_c of the DG
-cdef int paralle_transport_tracker(double* point,
-                                   double* direction,
-                                   ParalleTransportTrackingParameters params,
-                                   PmfGen pmf_gen):
+cdef int parallel_transport_tracker(double* point,
+                                    double* direction,
+                                    ParallelTransportTrackingParameters params,
+                                    PmfGen pmf_gen):
     # update point and dir with new position and direction
 
     # return 1 if the propagation failed.
--- a/dipy/tracking/fast_tracking.pyx
+++ b/dipy/tracking/fast_tracking.pyx
@@ -1,0 +1,20 @@
+cimport numpy as cnp
+
+from dipy.tracking.stopping_criterion cimport StoppingCriterion
+from dipy.direction.pmf cimport PmfGen
+
+cpdef generate_tractogram(cnp.ndarray[cnp.float_t, ndim=2] seed_positons,
+                          cnp.ndarray[cnp.float_t, ndim=2] seed_directions,
+                          StoppingCriterion sc,
+                          PmfGen pmf_gen):
+    cdef:
+        cnp.npy_intp _len=seed_positons.shape[0]
+
+    return generate_tractogram_c(<double[:_len,:3]> seed_positons,
+                                 <double[:_len,:3]> seed_directions,
+                                 sc, pmf_gen)
+
+cdef double[:, :] generate_tractogram_c(double[:, :] seed_positons,
+                                        double[:, :] seed_directions,
+                                        StoppingCriterion sc,
+                                        PmfGen pmf_gen)
--- a/dipy/tracking/fast_tracking.pyx
+++ b/dipy/tracking/fast_tracking.pyx
@@ -1,7 +1,9 @@
 # cython: boundscheck=False
 # cython: initializedcheck=False
 # cython: wraparound=False
+# cython: Nonecheck=False
 
+from cython.parallel import prange
 import numpy as np
 cimport numpy as cnp
 
@@ -24,7 +26,7 @@ from libc.stdlib cimport malloc, free
 cpdef list generate_tractogram(double[:,::1] seed_positons,
                                double[:,::1] seed_directions,
                                StoppingCriterion sc,
-                               TrackingParameters params,
+                               ProbabilisticTrackingParameters params,
                                PmfGen pmf_gen):
     cdef:
         cnp.npy_intp _len=seed_positons.shape[0]
@@ -57,7 +59,7 @@ cdef int generate_tractogram_c(double[:,::1] seed_positons,
                                double[:,::1] seed_directions,
                                int nbr_seeds,
                                StoppingCriterion sc,
-                               TrackingParameters params,
+                               ProbabilisticTrackingParameters params,
                                PmfGen pmf_gen,
                                func_ptr tracker,
                                double[:,:,:] streamlines,
@@ -66,8 +68,8 @@ cdef int generate_tractogram_c(double[:,::1] seed_positons,
         cnp.npy_intp i, j, k
 
 
-    # <<cython.parallel.prange>>
-    for i in range(nbr_seeds):
+    for i in prange(nbr_seeds, nogil=True):
+    # for i in range(nbr_seeds):
         stream = <double*> malloc((params.max_len * 3 * 2 + 1) * sizeof(double))
 
         # initialize to 0. It will be replaced when better handling various
@@ -103,13 +105,14 @@ cdef int generate_local_streamline(double* seed,
                                    double* stream,
                                    func_ptr tracker,
                                    StoppingCriterion sc,
-                                   TrackingParameters params,
-                                   PmfGen pmf_gen):
+                                   ProbabilisticTrackingParameters params,
+                                   PmfGen pmf_gen) noexcept nogil:
     cdef:
         cnp.npy_intp i, j
         double point[3]
         double voxdir[3]
         StreamlineStatus stream_status_forward, stream_status_backward
+
     # set the initial position
     copy_point(seed, point)
     copy_point(direction, voxdir)
@@ -118,7 +121,7 @@ cdef int generate_local_streamline(double* seed,
     # forward tracking
     stream_status_forward = TRACKPOINT
     for i in range(1, params.max_len):
-        if tracker(&point[0], &voxdir[0], params, pmf_gen):  # , pmf_gen_func):
+        if probabilistic_tracker(&point[0], &voxdir[0], params, pmf_gen):  # , pmf_gen_func):
             break
         # update position
         for j in range(3):
@@ -139,7 +142,7 @@ cdef int generate_local_streamline(double* seed,
         voxdir[j] = voxdir[j] * -1
     stream_status_backward = TRACKPOINT
     for i in range(1, params.max_len):
-        if tracker(&point[0], &voxdir[0], params, pmf_gen):
+        if probabilistic_tracker(&point[0], &voxdir[0], params, pmf_gen):
             break
         # update position
         for j in range(3):
@@ -160,7 +163,7 @@ cdef int generate_local_streamline(double* seed,
 cdef double* get_pmf(double* point,
                      PmfGen pmf_gen,
                      double pmf_threshold,
-                     int pmf_len):
+                     int pmf_len) noexcept nogil:
     cdef:
         cnp.npy_intp i
         double* pmf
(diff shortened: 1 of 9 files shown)


skoudoro:
Great! Thank you for this. 

It is nice to show that it exists. there still some work like:

- [ ] docstring
- [ ] unittest
- [ ] adding more algorithms
- [ ] simplify tracking parameter
- [ ] deprecated the current interface
- [ ] adding examples
- [ ] adding benchmarks

but it should not take too long.  It will not be in the next release, but it will be in the release in June/July

skoudoro:
We also need to look at #1404 in this PR

dipy/tracking/stopping_criterion.pyx
```diff
@@ -277,19 +279,22 @@ cdef class CmcStoppingCriterion(AnatomicalStoppingCriterion):
...
+        # rng = np.random.default_rng()
+
         # test if the tracking continues
         if include_result + exclude_result <= 0:
             return TRACKPOINT
         num = max(0, (1 - include_result - exclude_result))
         den = num + include_result + exclude_result
         p = (num / den) ** self.correction_factor
-        if rng.random() < p:
+        # if rng.random() < p:
```
jhlegarreta, line 290: Remove this rather than leaving as a comment?

dipy/tracking/stopping_criterion.pyx
```diff
@@ -277,19 +279,22 @@ cdef class CmcStoppingCriterion(AnatomicalStoppingCriterion):
...
+        if random() < p:
             return TRACKPOINT
 
         # test if the tracking stopped in the include tissue map
         p = (include_result / (include_result + exclude_result))
-        if rng.random() < p:
+        # if rng.random() < p:
```
jhlegarreta, line 296: Same here?

dipy/tracking/fast_tracking.pyx
```diff
@@ -0,0 +1,235 @@
...
+
+        # copy the streamlines points from the buffer to a 1d vector of the streamline length
+        lengths[i] = stream_idx[1] - stream_idx[0]
+        if lengths[i] > 1:
+            streamlines[i] = <double*> malloc(lengths[i] * 3 * sizeof(double))
+            memcpy(&streamlines[i][0], &stream[stream_idx[0] * 3], lengths[i] * 3 * sizeof(double))
+        free(stream)
+        free(stream_idx)
+
+    return 0
```
jhlegarreta, line 123: Looks like this is the only possible return value. Maybe define a constant called e.g. ``TRACKING_SUCCESS` and assign it a 0 value and use it?

dipy/tracking/tracker.py
```diff
@@ -0,0 +1,93 @@
...
+
+def deterministic_directions():
+    pass
+
+
+def eudx_directions():
+    pass
+
+
+def build_stopping_criterion(mask=None, check_point=None, include=None,
```
jhlegarreta, line 16: Documentation is missing. Same applies to all methods in his file.

dipy/tracking/tracker.py
```diff
@@ -0,0 +1,93 @@
...
+                             exclude=None):
+    sc_element = {
+        "mask": mask,
+        "wm_map": None,
+        "gm_map": None,
+        "csf_map": None,
+        "checkpoint": check_point,
+        "include": include,
+        "exclude": exclude,
```
jhlegarreta, line 25: I know we usually do no use enums, but I would define these into an enum or an SC class; otherwise, they are not exposed/documented in the API.

dipy/tracking/fast_tracking.pyx
```diff
@@ -0,0 +1,235 @@
...
+cdef extern from "stdlib.h" nogil:
+    void *memset(void *ptr, int value, size_t num)
+
+def generate_tractogram(double[:,::1] seed_positions,
+                        double[:,::1] seed_directions,
+                        StoppingCriterion sc,
+                        TrackerParameters params,
+                        PmfGen pmf_gen,
+                        int nbr_threads=0,
+                        float buffer_frac=1.0):
```
jhlegarreta, line 44: Do we document Cython functions? If yes, these should be probably documented.

dipy/tracking/tracker.py
```diff
@@ -0,0 +1,93 @@
...
+                                 tractogram_func=generate_tractogram)
+probabilistic_tracking_python = partial(local_tracking,
+                                        tractogram_func=generate_tractogram_py)
+
+deterministic_tracking = partial(
+    local_tracking, direction_getter="deterministic")
+eudx_tracking = partial(local_tracking, direction_getter="eudx")
+fact_tracking = partial(local_tracking, direction_getter="fact")
+ptt_tracking = partial(local_tracking, direction_getter="ptt")
+pft_tracking = partial(local_tracking, direction_getter="pft")
```
jhlegarreta, line 87: Should the algorithm names be exposed in the API as well?

dipy/tracking/tracker_deterministic.pyx
```diff
@@ -0,0 +1,70 @@
...
+    cdef:
+        cnp.npy_intp i, max_idx
+        double max_value=0
+        double* newdir
+        double* pmf
+        double cos_sim
+        cnp.npy_intp len_pmf=pmf_gen.pmf.shape[0]
+
+    if norm(direction) == 0:
+        return 1
```
jhlegarreta, line 34: Similar to the comment about the `return 0` statements, should these be declared as a constant and be used across the code?

dipy/tracking/tracker_parameters.pyx
```diff
@@ -0,0 +1,125 @@
...
+                                   max_angle=max_angle,
+                                   probe_length=probe_length,
+                                   probe_radius=probe_radius,
+                                   probe_quality=probe_quality,
+                                   probe_count=probe_count,
+                                   data_support_exponent=data_support_exponent,
+                                   random_seed=random_seed)
+        params.set_tracker_c(parallel_transport_tracker)
+        return params
+    #elif algo_name == 'eudx':
```
jhlegarreta, line 54: Remove for now?

dipy/tracking/tracker_parameters.pyx
```diff
@@ -0,0 +1,125 @@
...
+        self.sh = None
+        self.ptt = None
+
+        if pmf_threshold is not None:
+            self.sh = ShTrackerParameters(pmf_threshold)
+
+        if probe_length is not None and probe_radius is not None and probe_quality is not None and probe_count is not None and data_support_exponent is not None:
+            self.ptt = ParallelTransportTrackerParameters(probe_length, probe_radius, probe_quality, probe_count, data_support_exponent)
+
+    # def set_tracker(self, tracker):
```
jhlegarreta, line 94: Remove?

dipy/tracking/tracker_ptt.pyx
```diff
@@ -0,0 +1,386 @@
...
+    stream_data : double*
+        Streamline data persitant across tracking steps.
+    pmf_gen : PmfGen
+        Orientation data.
+
+    Returns
+    -------
+    status : int
+        Returns 0 if the propagation was successful, or
+        1 otherwise.
```
jhlegarreta, line 60: Update these if the aforementioned constants are defined.

dipy/tracking/tracker_ptt.pyx
```diff
@@ -0,0 +1,386 @@
...
+        seed_point : double[3]
+            Initial point
+        seed_direction : double[3]
+            Initial direction
+
+        Returns
+        -------
+        status : int
+            Returns 0 if the initialization was successful, or
+            1 otherwise.
```
jhlegarreta, line 149: Same as above.

dipy/tracking/tracker_ptt.pyx
```diff
@@ -0,0 +1,386 @@
...
+            if (random() * max_posterior <= calculate_data_support(params, stream_data, pmf_gen)):
+                stream_data[22] = stream_data[23] # last_val = last_val_cand
+                return 0
+        return 1
+
+cdef void initialize_candidate(TrackerParameters params,
+                               double* stream_data,
+                               PmfGen pmf_gen,
+                               double* init_dir) noexcept nogil:
+    """"Initialize the parallel transport frame.
```
jhlegarreta, line 182: One double tick in excess.

dipy/tracking/tracker_ptt.pyx
```diff
@@ -0,0 +1,386 @@
...
+    plus 2 scalars, i.e. k1 and k2).
+
+    A point and parallel transport frame parametrizes a curve that is named
+    the "probe". Using probe parameters (probe_length, probe_radius,
+    probe_quality, probe_count), a short fiber bundle segment is modelled.
+
+    Parameters
+    ----------
+    params : TrackerParameters
+        PTT tracking parameters.
```
jhlegarreta, line 195: Define PTT somewhere?

dipy/tracking/tracker_ptt.pyx
```diff
@@ -0,0 +1,386 @@
...
+                               double* init_dir) noexcept nogil:
+    """"Initialize the parallel transport frame.
+
+    After initial position is set, a parallel transport frame is set using
+    the initial direction (a walking frame, i.e., 3 orthonormal vectors,
+    plus 2 scalars, i.e. k1 and k2).
+
+    A point and parallel transport frame parametrizes a curve that is named
+    the "probe". Using probe parameters (probe_length, probe_radius,
+    probe_quality, probe_count), a short fiber bundle segment is modelled.
```
jhlegarreta, line 190: Are these parameters defined, documented or exposed somewhere?

dipy/tracking/tracker_ptt.pyx
```diff
@@ -0,0 +1,386 @@
...
+                                   PmfGen pmf_gen) noexcept nogil:
+    """Calculates data support for the candidate probe.
+        Parameters
+    ----------
+    params : TrackerParameters
+        PTT tracking parameters.
+    stream_data : double*
+        Streamline data persitant across tracking steps.
+    pmf_gen : PmfGen
+        Orientation data.
```
jhlegarreta, line 308: Return value missing.

jhlegarreta (commented):
Will try the example locally as time permits.

Some in-line comments, mostly about documentation.

As for the style formatting, maybe in your last push you had not rebased on `master`? The below recipe should work locally and give the same result as the `code-format`build:
https://github.com/dipy/dipy/blob/master/doc/devel/coding_style_guideline.rst#coding-style
https://github.com/dipy/dipy/actions/runs/11281141299/job/31375708287?pr=3089

jhlegarreta:
Two more comments:
- Would be a good idea to document the new code (especially the Cython part) a little bit, at least the main blocks. Otherwise, it is hard to follow :book:.
- I have executed the example, and it runs OK, and the outputs look OK :100:. If the progress bars indicate the time to download the dataset, then it is notably long, even more so compared to the time spent on tracking. I know there is an issue to try to specify which parts to download. As a temporary solution, maybe a "disco_small" dataset could be created. If the progress bars include the tracking process, then maybe the events they belong to should be indicated, as otherwise the user does not know what the progress bars mean.

I think for a first round of comments the above is enough.

dipy/tracking/stopping_criterion.pyx
gabknight: thanks, removed.

dipy/tracking/stopping_criterion.pyx
gabknight: yes

dipy/tracking/tracker_ptt.pyx
gabknight: fixed.

dipy/tracking/tracker_ptt.pyx
gabknight: yes, fixed. thanks

dipy/tracking/fast_tracking.pyx
gabknight: I change the return from int to void.

dipy/tracking/tracker_deterministic.pyx
gabknight: We have many cython functions with 0/1+ status returns, e.g. `dipy.core.interpolation.trilinear_interpolate4d_c`.

I agree constant could clarify the code. @skoudoro what do you suggest?

dipy/tracking/tracker_ptt.pyx
jhlegarreta: My bad, sorry. There is not return section then. I think there is multiple functions that do not have a return value, so no return section should exist for those.

dipy/tracking/tracker.py
gabknight: solve, added a enum `TrackerStatus`

thanks

skoudoro:
Hi @gabknight,

It does not compile on Windows due to the following error:

```
  dipy/tracking/fast_tracking.cp311-win_amd64.pyd.p/fast_tracking.c(1236): fatal error C1083: Cannot open include file: 'sys/time.h': No such file or directory
```

In all operating systems, we have this warning which is problematic:
```
 [82/150] Generating 'dipy\\tracking\\fast_tracking.cp312-win_amd64.pyd.p\\fast_tracking.c'
      performance hint: C:\Users\Eleftherios\Devel\dipy\dipy\tracking\fast_tracking.pyx:164:19: Exception check after calling 'seed' will always require the GIL to be acquired.

      Possible solutions:

          1. Declare 'seed' as 'noexcept' if you control the definition and you're sure you don't want the function to raise exceptions.

          2. Use an 'int' return type on 'seed' to allow an error code to be returned.
```

Let me know when you start to look at this

dipy/tracking/fast_tracking.pyx
```diff
@@ -0,0 +1,256 @@
...
+                                               FAIL)
+
+from nibabel.streamlines import ArraySequence as Streamlines
+
+from libc.stdlib cimport malloc, free
+from libc.string cimport memcpy
+from libc.math cimport floor, ceil
+from libc.stdio cimport printf
+
+from posix.time cimport clock_gettime, timespec, CLOCK_REALTIME
```
skoudoro, line 38: `posix.time` : Unix-specific header file and is not available in the standard library of Windows, which is what Visual Studio targets.

That's why it fails on windows.

dipy/tracking/fast_tracking.pyx
```diff
@@ -0,0 +1,256 @@
...
+        cnp.npy_intp _len = seed_positions.shape[0]
+        cnp.npy_intp _plen = int(ceil(_len * buffer_frac))
+        cnp.npy_intp i, seed_start, seed_end
+        double** streamlines_arr
+        int* length_arr
+        StreamlineStatus* status_arr
+
+    # srand/rand don't play well with multi-threading
+    if params.random_seed > 0 and not nbr_threads == 1:
+        raise ValueError("random_seed > 0 do not work with nbr_threads != 1.")
```
skoudoro, line 67: Does this still make sense. Especially when you look later in the code this (line 157): 

```
    if params.random_seed > 0:
        s_random_seed = int(
            (seed[0] * 2 + seed[1] * 3 + seed[2] * 5) * params.random_seed
            )
    else:
        clock_gettime(CLOCK_REALTIME, &ts)
        s_random_seed = int(ts.tv_sec + (ts.tv_nsec / 1000000000.))
    fast_numpy.seed(s_random_seed)
```

gabknight: I think we can get rid of this initialization. If the `.random_seed <= 0`, we should do nothing in the `else`. 

The objective of this code was to allow for the generation of multiple probabilistic streamlines from the same initial coordinate. Otherwise, the code always generates the same streamline for a given initial coordinate. Because the code doesn't support parallel tracking for fix random seeds, there is no need for a time-based random seed initialization.

I will propose a fix for this.

dipy/tracking/tracker.py
```diff
@@ -159,6 +175,10 @@ def probabilistic_tracking(
         will all produce the same streamline trajectory for a given seed coordinate
         (will forces nbr_threads=1). A value of 0 may produces various streamline
         tracjectories for a given seed coordinate.
+    seed_buffer_fraction: float, optional
+        Fraction of the seed buffer to use. A value of 1.0 will use the entire seed
+        buffer. A value of 0.5 will use half of the seed buffer then the other half.
+        a way to reduce memory usage.
```
jhlegarreta, line 181: Capitalize A: A way to reduce memory usage. There is multiple appearances.

Garyfallidis:
@gabknight 
I was checking this PR for any issues and I was ready to merge but then noticed something unexpected. The result of PTT is  worse than PT and DT. See below.

Running fast Deterministic Tractography...
DiSCo ground-truth correlation (fast deterministic tractography):  0.9050590977389915
Running fast Probabilistic Tractography...
DiSCo ground-truth correlation (fast probabilistic tractography):  0.9146863983957279
Running fast Parallel Transport Tractography...
DiSCo ground-truth correlation (fast PTT tractography):  0.5394093789432843

gabknight:
@Garyfallidis good catch! There was change in the default parameters. My last commit addresses this.

Running fast Deterministic Tractography...
DiSCo ground-truth correlation (fast deterministic tractography):  0.9090724543541376
Running fast Probabilistic Tractography...
DiSCo ground-truth correlation (fast probabilistic tractography):  0.9246372195897025
Running fast Parallel Transport Tractography...
DiSCo ground-truth correlation (fast PTT tractography):  0.8924315441639352

Garyfallidis:
Can we make PTT surpass 0.92? Any particular parameter choices?

gabknight:
@skoudoro if you have time, I found 2 issues I will need your help to solve

- `doc/examples/tracking_fast_tracking.py`, changing line 82 to `sphere = default_sphere` produces erroneous results. Somehow changing from a `Sphere` object to an `Hemisphere` object create the problem. I think the issue is in tracker.py, line 89.
- Running the example twice produces the same scores for prob and det, but not for ptt. There is a place where `RNG` is not used, probably. 

I will try looking into this next week, let me know if you find something.

skoudoro:
> doc/examples/tracking_fast_tracking.py, changing line 82 to sphere = default_sphere produces erroneous results. Somehow changing from a Sphere object to an Hemisphere object create the problem. I think the issue is in tracker.py, line 89.

I will try to play with this. not sure I get it 

> Running the example twice produces the same scores for prob and det, but not for ptt. There is a place where RNG is not used, probably.

`random_point_within_circle` in `propagator.pyx` is missing the rng

Thank you for the update

skoudoro:
ok, I just an updated to fix RNG in PTT. I checked and all good now. the last commit  does the following also:
- removing the propagator files
- moving the deterministic, probabilistic and parallel_transport propagators to the propspeed file
- renaming the fast_tracking files to tractogen
- renaming the test_propagation file to test_propspeed
- renaming the test_fast_tracking file to test_tractogen
- fixing rng for ptt

if you have time to look at the sphere issue @gabknight

Garyfallidis:
On windows there is a fetching error following the tutorial.

  File "C:\Users\Y\Devel\dipy\doc\examples\tracking_fast_tractography.py", line 49, in <module>
    disco1_fnames.index("DiSCo1_Connectivity_Matrix_Cross-Sectional_Area.txt")
ValueError: 'DiSCo1_Connectivity_Matrix_Cross-Sectional_Area.txt' is not in list

Garyfallidis:
Some of the  datasets take quite some time to download.

Garyfallidis:
Downloading the data of this tutorial takes way too long. More than 30' on my laptop. Hasn't finished yet.

Garyfallidis:
Congrats to @gabknight and @skoudoro !!! Let's update the rest of tutorials and increase the speed of data fetching.

PR #3092, opened 2024-03-01
Feature: volume synchronizing

maharshi-gor:
### Feature: Volume Synchronizing

This PR introduces the changes for a feature where the volume slider will be synchronized.

**When to synchronize**
- All the provided images have the same shape.
- All the images with volume should have the same volume samples.

### Behavior Expectation

**Images with the same shape and same volumes**
The slices will be synchronized and volumes will also be synchronized.

**Images with the same shape but different volumes**
The slices will be synchronized but volumes will not be synchronized.

**Images with different shapes but same volumes** and **Images with different shapes and different volumes**
Nothing will be synchronized and a warning will be visible.

### DIFF
--- a/dipy/viz/horizon/app.py
+++ b/dipy/viz/horizon/app.py
@@ -441,14 +441,14 @@ class Horizon:
                 self.__tabs.append(ClustersTab(
                     self.__clusters_visualizer, self.cluster_thr))
 
-        synchronize_slices = False
+        sync_slices = sync_vol = False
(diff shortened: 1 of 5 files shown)


skoudoro (approved):
Hi @maharshi-gor,

Looks good to me. I just need to do a quick local tests and then merging

skoudoro:
Thanks @maharshi-gor, merging

PR #3098, opened 2024-03-05
Removing tensorflow addon from DL models

pjsjongsung:
This commit is to remove the optional dependency of tensorflow addon due to its deprecation. I have also removed some repeated code in Synb0.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -17,6 +17,6 @@ jobs:
         runs-on: '["ubuntu-latest", ]'
         coverage: true
         enable-viz-tests: true
-        extra-depends: scikit-learn scipy statsmodels pandas tables fury tensorflow tensorflow-addons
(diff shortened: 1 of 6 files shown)


skoudoro:
Thanks @pjsjongsung. 

is tensorflow addon replaced by something else?

skoudoro:
Thank you for the clean up! 

Can you also remove it from the `pyproject.toml` and the `coverage.yml`  files? 

Then we can go ahead and merge this.

pjsjongsung:
> Thanks @pjsjongsung.
> 
> is tensorflow addon replaced by something else?

It is going fully out of support and most of the functions are now replaceable with the functions in Tensorflow.



> Thank you for the clean up!
> 
> Can you also remove it from the `pyproject.toml` and the `coverage.yml` files?
> 
> Then we can go ahead and merge this.

Will do!

skoudoro:
Thank you for the update @pjsjongsung, merging

PR #3102, opened 2024-03-06
Bugfix for peaks slices and synchronization.

maharshi-gor:
This PR introduces bug fixes for Peaks, making the slicer range till the start and end of the native coordinates.

It also introduces synchronization for the peak actors and images.


skoudoro:
Thank you @maharshi-gor for this work, merging

PR #3108, opened 2024-03-07
enabled trx support with correct header

maharshi-gor:


arokem:
Sorry - I am not sure what was wrong and why this is right. Any chance to get a test that demonstrates this?

Issue #2950, opened 2023-10-20
Clean up the non-local means denoising modules


jhlegarreta:
## Description

The `denoise` module contains two modules named `nlmeans` vs `non_local_means` whose meaning is exactly the same. The `nlmeans.py` module has a [commented deprecation warning](https://github.com/dipy/dipy/blob/master/dipy/denoise/nlmeans.py#L7). Some time should be invested in cleaning up/refactoring the code so we only have a single non-local means module.

## Way to reproduce
N/A

itsayopapi:
@jhlegarreta I want to work on this issue..can you please assign it to me

skoudoro:
Hi @itsayopapi,

You can go ahead, let us know if you have any specific question

PR #3113, opened 2024-03-10
[FIX] Nlmeans Algorithm Enhancement #2950

tanishka321:
## Fixes:
#2950 

## Description
Enhancement involves the utilization of nlmeans_block instead of nlmeans_3D approach.
The [code](
https://drive.google.com/file/d/1qeJv_CQeYHm7R52KwmrGpPUNJgHOKHfh/view?usp=sharing) used for the comparison of results on the standard T1 data from Stanford University

Figure_1

Dark regions indicate areas where the denoised image closely matches the noisy input, while brighter regions indicate areas where the denoising process resulted in significant alterations.
Ideally, the difference images should have minimal bright regions, indicating effective denoising without compromising image details.


pjsjongsung:
I think the deprecation message was there to keep non_local_means as the new function name instead of nlmeans so it should be the other way around? (non_local_means should stay) Also if we are moving towards completely replacing nlmeans_3d with nlmeans_block, we should remove that function as well. If not, we can have it as an option. (Like default is nlmeans_block but we can pass a parameter to use the nlmeans_3d version)

skoudoro:
Hi @tanishka321,

Thank you for working on this. We can not move forward to replace this. We need first some comparison of the 2 implementations because there are many issues reported from the second implementation. Look here: https://github.com/search?q=repo:dipy/dipy+non+local+means&type=issues

if you have time, can you look at those issues and compare the version? Thank you

tanishka321:
@skoudoro @pjsjongsung  
Upon going through the code-base and the GitHub issue search link sir have linked, I agree with your perspective that a straightforward replacement may not be the most prudent approach.
I will invest time to delve into the provided resources, to gain a deeper understanding of the challenges associated with nlmeans_block and compare them with the existing nlmeans_3D approach. By doing so, we can make a more informed decision regarding the best path forward for the enhancement.

skoudoro:
Closing for now. We can discuss the best approach directly on issue https://github.com/dipy/dipy/issues/2950

PR #3120, opened 2024-03-13
DOC - Removed unnecessary line from tracking example

gabknight:
- the tracking example included the response calculation, which is not required for csa odf fit.
- re-ordered imports

Thanks @ebrahimebrahim


skoudoro:
Thank you @gabknight, merging

PR #3126, opened 2024-03-15
Bugfix for ROI images updates

maharshi-gor:
### Bugfix of ROI option
This PR fixes the change of `--roi_images` option throwing an error.

### Fix Methodology
- Instead of checking entire dataset for unique data, we can pickup 3D box at the center of the 3D image, and check the box for binary values.
- For 4D images we are checking the values of the different 3D boxes.
- By default, the box is of size **(15, 15, 15)** and 15 volumes for 4D dataset.
- If the size of dimension less than the box the box will adjust to dimension.
Screenshot from 2024-03-15 16-12-09

### DIFF
--- a/dipy/viz/horizon/util.py
+++ b/dipy/viz/horizon/util.py
@@ -151,7 +151,7 @@ def _unpack_data(data, return_size=3):
     return result
 
 
-def is_binary_image(data, unique_points=100):
+def is_binary_image(data, sample_cube_size=(10, 10, 10)):
     """Check if an image is binary image.
 
     Parameters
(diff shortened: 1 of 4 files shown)


dipy/viz/horizon/util.py
```diff
@@ -165,14 +165,18 @@ def is_binary_image(data, unique_points=100):
...
+    for idx, dim in enumerate(data.shape):
+        if dim < sample_cube[idx]:
+            data = np.take(data, np.arange(stop=dim), axis=idx)
+        else:
+            start = int(dim/2) - int(sample_cube[idx]/2)
+            stop = int(dim/2) + int(sample_cube[idx]/2)
+            data = np.take(data, np.arange(start=start, stop=stop), axis=idx)
 
-    return np.unique(np.take(data, indices)).shape[0] <= 2
+    return np.unique(data).shape[0] <= 2
```
skoudoro, line 179: Why not `==2`?  

Not sure if having 1 values or 0 values can be considered binary

skoudoro (commented):
Hi @maharshi-gor,

Thank you for this. See below a small comment. After your feedback, I think this PR can go in

dipy/viz/horizon/util.py
maharshi-gor: This is in case of we do not find multiple values in the small box. This is how the old logic was, so made sure to keep it the same.

dipy/viz/horizon/util.py
```diff
@@ -151,7 +151,7 @@ def _unpack_data(data, return_size=3):
     return result
 
 
-def is_binary_image(data, unique_points=100):
+def is_binary_image(data, sample_cube_size=15):
```
skoudoro, line 154: `sample_cube_size` instead of a fix value, I would recommend to take a percentage of the volume.

`sample_cube_size=15` is not the same if the volume is 100x100x100 or 1000x1000x1000

dipy/viz/horizon/util.py
skoudoro: ok

skoudoro (commented):
Thank you for the feedback. See below additional comments

dipy/viz/horizon/util.py
maharshi-gor: Make sense, I have updated that!

skoudoro (approved):
ok, LGTM, merging, Thanks @maharshi-gor

PR #3127, opened 2024-03-15
Feature/viz interface tutorials

maharshi-gor:
Basic flow of the CLI tutorial shows a dipy_horizon using rgb image but it requires --rgb option. Thus, the tutorial needed an update


skoudoro:
Thank you for this fix @maharshi-gor , merging

Issue #3096, opened 2024-03-04
TissueClassifierHMRF has some argument logic error


pjsjongsung:
`dipy.segment.tissue.TissueClassifierHMRF` according to the document and code is supposed to not break the loop through `tolerance` in two scenarios
1. `max_iter` is given
2. `tolerance` is not given

However, in this case `tolerance` having a default value does not make any sense. It also does not have `max_iter` cancel out `tolerance`, given how they are completely independent parameters.

I believe the correct direction is not having the `allow_break` argument at all, and if someone does not want to use it, they can always just pass 0 to `tolerance`.

One additional thing we have to edit is  the different behavior of `allow_break` when` save_history` is given, and if both are True results in a duplicate append.

I (or someone else if they want) could work on this PR, but need some feedback on whether removing `allow_break` is the correct direction. As an alternative, we could just remove the default `tolerance` value, but I do not think that is the original intention.

Pavan-Satyam:
@pjsjongsung I want to look on this issue .Could you guide me what are the next steps where I could look at the code?

arokem:
Sorry I missed this befor. I think that removing `allow_break` and keeping the `tolerance` default is the right direction.

pjsjongsung:
Thanks @arokem ! @Pavan-Satyam if you want to work on this PR, now we seem to know the direction it should go.

skoudoro:
fixed by #3132

PR #3132, opened 2024-03-18
BF: Removed allow_break

deka27:
# Removed allow_break from TissueClassifierHMRF

## Overview
This PR introduces removal of 'allow_break' from the TissueClassifierHMRF as mentioned in issue #3096. 

## Changes
- Removed the `allow_break` variable to simplify loop exit logic.
- The `tolerance` parameter now explicitly disables the early stopping condition when set to `0`, allowing for clear differentiation between using and not using the tolerance check for loop control.
- Clarified documentation for the `tolerance` parameter to indicate how users can disable the tolerance check.

## Related Issues
* #3096 

## Screenshots
I ran this [tutorial](https://docs.dipy.org/stable/examples_built/segmentation/tissue_classification.html#sphx-glr-examples-built-segmentation-tissue-classification-py) to check if the changes were functional. Here are the results.

Old Code

Screenshot 2024-03-18 at 7 12 32 PM

New Code

Screenshot 2024-03-18 at 7 09 19 PM

## Additional Comments
* I also ran the test mentioned in `test_mrf.py` by running the function `test_classify`, the test was successful.
```python
pytest -svv -k 'test_classify'
```
* Based on my understanding I have put the changes, but please let me know if this is the right way or any other changes are required. Any feedbacks would be helpful.

## Checklist
- [x] I have reviewed my own code changes.
- [x] I have tested my changes locally.
- [x] My code follows the project's coding standards.

### DIFF
--- a/dipy/segment/tissue.py
+++ b/dipy/segment/tissue.py
@@ -5,13 +5,12 @@ from dipy.segment.mrf import (ConstantObservationModel,
 
 
 class TissueClassifierHMRF:
-    r"""
-    This class contains the methods for tissue classification using the Markov
-    Random Fields modeling approach
+    """
+    This class contains the methods for tissue classification using the
+    Markov Random Fields modeling approach.
     """
 
     def __init__(self, save_history=False, verbose=True):
-
         self.save_history = save_history
         self.segmentations = []
         self.pves = []
@@ -19,43 +18,45 @@ class TissueClassifierHMRF:
         self.energies_sum = []
         self.verbose = verbose
 
-    def classify(self, image, nclasses, beta, tolerance=None, max_iter=None):
-        r"""
+    def classify(self, image, nclasses, beta, tolerance=1e-05, max_iter=100):
+        """
         This method uses the Maximum a posteriori - Markov Random Field
-        approach for segmentation by using the Iterative Conditional Modes and
-        Expectation Maximization to estimate the parameters.
+        approach for segmentation by using the Iterative Conditional Modes
+        and Expectation Maximization to estimate the parameters.
 
         Parameters
         ----------
         image : ndarray,
-                3D structural image.
+            3D structural image.
         nclasses : int,
-                number of desired classes.
+            Number of desired classes.
         beta : float,
-                smoothing parameter, the higher this number the smoother the
-                output will be.
-        tolerance: float,
-                value that defines the percentage of change tolerated to
-                prevent the ICM loop to stop. Default is 1e-05.
-        max_iter : float,
-                fixed number of desired iterations. Default is 100.
-                If the user only specifies this parameter, the tolerance
-                value will not be considered. If none of these two
-                parameters
+            Smoothing parameter, the higher this number the smoother the
+            output will be.
+        tolerance: float, optional
+            Value that defines the percentage of change tolerated to
+            prevent the ICM loop to stop. Default is 1e-05.
+            If you want tolerance check to be disabled put 'tolerance = 0'.
+        max_iter : int, optional
+            Fixed number of desired iterations. Default is 100.
+            This parameter defines the maximum number of iterations the
+            algorithm will perform. The loop may terminate early if the
+            change in energy sum between iterations falls below the
+            threshold defined by `tolerance`. However, if `tolerance` is
+            explicitly set to 0, this early stopping mechanism is disabled,
+            and the algorithm will run for the specified number of
+            iterations unless another stopping criterion is met.
 
         Returns
         -------
         initial_segmentation : ndarray,
-                3D segmented image with all tissue types
-                specified in nclasses.
+            3D segmented image with all tissue types specified in nclasses.
         final_segmentation : ndarray,
-                3D final refined segmentation containing all
-                tissue types.
+            3D final refined segmentation containing all tissue types.
         PVE : ndarray,
-                3D probability map of each tissue type.
+            3D probability map of each tissue type.
         """
-
-        nclasses = nclasses + 1  # One extra class for the background
+        nclasses += 1  # One extra class for the background
         energy_sum = [1e-05]
 
         com = ConstantObservationModel()
@@ -67,7 +68,7 @@ class TissueClassifierHMRF:
         mu, sigmasq = com.initialize_param_uniform(image, nclasses)
         p = np.argsort(mu)
         mu = mu[p]
-        sigma = sigmasq[p]
+        sigmasq = sigmasq[p]
 
         neglogl = com.negloglikelihood(image, mu, sigmasq, nclasses)
         seg_init = icm.initialize_maximum_likelihood(neglogl)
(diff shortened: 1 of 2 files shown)


dipy/segment/tissue.py
```diff
@@ -5,57 +5,54 @@
...
-                prevent the ICM loop to stop. Default is 1e-05.
-        max_iter : float,
-                fixed number of desired iterations. Default is 100.
-                If the user only specifies this parameter, the tolerance
-                value will not be considered. If none of these two
-                parameters
+            Value that defines the percentage of change tolerated to
+            prevent the ICM loop to stop. Default is 1e-05.
+            If you want tolerance check to be disabled put 'tolerance = 0'.
+        max_iter : int,
```
skoudoro, line 40: Can you replace  `int,` by `int, optional`

dipy/segment/tissue.py
```diff
@@ -5,57 +5,54 @@
...
+            3D structural image.
         nclasses : int,
-                number of desired classes.
+            Number of desired classes.
         beta : float,
-                smoothing parameter, the higher this number the smoother the
-                output will be.
+            Smoothing parameter, the higher this number the smoother the
+            output will be.
         tolerance: float,
```
skoudoro, line 36: `float, optional` instead, can you add it?

dipy/segment/tissue.py
pjsjongsung, line 40: Could you also edit the max_iter description?

skoudoro (commented):
Hi @deka27,

Thank you for this, waiting for the CI's but this change make sense to me. 
see below few comments

deka27:
I will be squash the commits now.

deka27:
The squashing is done, Please let me know if things are correct now!

deka27:
I have done the squashing as required, please let me know if this is correct.

skoudoro (approved):
LGTM. 

is there any additional comment? @pjsjongsung ? @arokem ?

skoudoro:
Thanks @deka27 for this work.

deka27:
> Thanks @deka27 for this work.

My Pleasure.

dipy/segment/tissue.py
```diff
@@ -5,57 +5,58 @@
...
-        max_iter : float,
-                fixed number of desired iterations. Default is 100.
-                If the user only specifies this parameter, the tolerance
-                value will not be considered. If none of these two
-                parameters
+            Smoothing parameter, the higher this number the smoother the
+            output will be.
+        tolerance: float, optional
+            Value that defines the percentage of change tolerated to
+            prevent the ICM loop to stop. Default is 1e-05.
```
jhlegarreta, line 38: Remove defaults here: this is difficult to maintain.

dipy/segment/tissue.py
```diff
@@ -5,57 +5,58 @@
...
+            If you want tolerance check to be disabled put 'tolerance = 0'.
+        max_iter : int, optional
+            Fixed number of desired iterations. Default is 100.
```
jhlegarreta, line 41: Remove defaults here: this is difficult to maintain.

doc/doc-requirements.txt
```diff
@@ -15,7 +15,7 @@ numpydoc
 sphinx!=4.4
 sphinx-gallery>=0.10.0
 tomli>=2.0.1
-pydata-sphinx-theme==0.13.3
+grg-sphinx-theme>=0.3
```
jhlegarreta, line 18: Why was this changed? If this was required, should have been done on a separate PR: this is unrelated to the `allow_break` issue. Also, this was part of another commit: https://github.com/dipy/dipy/commit/c2c42c44569ec12a4cdf5ed8a97c45bb114d6b11

dipy/segment/tissue.py
```diff
@@ -81,51 +82,36 @@ def classify(self, image, nclasses, beta, tolerance=None, max_iter=None):
...
                 self.energies.append(energy)
-                if allow_break:
-                    self.energies_sum.append(energy_sum[-1])
-                else:
-                    self.energies_sum.append(energy[energy > -np.inf].sum())
-
-            if allow_break and i > 5:
+                self.energies_sum.append(energy_sum[-1])
 
+            if tolerance > 0 and i > 5:
```
jhlegarreta, line 110: For a separate PR: if 5 means something here, it should be put to a variable and explain the meaning, as it is also used in https://github.com/dipy/dipy/pull/3132/files#diff-540e9ecc7f5c0befbaff5a9def01d422445705275835db8a00e8c763792e6e12R114.

dipy/segment/tissue.py
```diff
@@ -5,57 +5,58 @@
...
+            explicitly set to 0, this early stopping mechanism is disabled,
+            and the algorithm will run for the specified number of
+            iterations unless another stopping criterion is met.
 
         Returns
         -------
         initial_segmentation : ndarray,
-                3D segmented image with all tissue types
-                specified in nclasses.
+            3D segmented image with all tissue types specified in nclasses.
```
jhlegarreta, line 53: Style changes (and there are a lot here) should be on their own commit. Otherwise the review gets a lot harder, the real issue being addressed getting lost.

jhlegarreta (commented):
@skoudoro @pjsjongsung Sorry for the late review: the scope of this was a little bit too broad IMO.

doc/doc-requirements.txt
pjsjongsung: Agree. It might have been from squashing some unrelated commit

dipy/segment/tissue.py
pjsjongsung: The default values were following what it already had. If we just remove the default values, it could result in breaking many user code in the next version. Plus, tolerance value is not something a user would change a lot, and is hard to understand how to select the appropriate value.

dipy/segment/tissue.py
pjsjongsung: Agree. We could give it a name like tolerance_window

dipy/segment/tissue.py
pjsjongsung: Agree. We need to double check if this change follows the styles of other code files as well.

dipy/segment/tissue.py
jhlegarreta: Only meant removing defaults in the documentation, same as in e.g.
https://github.com/dipy/dipy/pull/2707#discussion_r1071309787

If for some reason the default changes in the method signature, changes that the docstring goes out of sync are high.

pjsjongsung: Makes sense, especially since it is not a value user would have to understand most of the time.

dipy/segment/tissue.py
deka27: I can put up a separate PR for that. Thank you for the feedback.

doc/doc-requirements.txt
deka27: Yea, sorry about that. I updated the branch from upstream while working on the issue and then squashed it, because of that this also went into the PR.

PR #3138, opened 2024-03-19
[CI] update CI's script

skoudoro:
This PR:
- [x] Adds m1 matrix
- [x] Fix nightly user
- [x] Update pre to 3.12
- [x] Check codecov (might need an update)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,5 +20,5 @@ jobs:
   pip:
     uses: ./.github/workflows/test_template.yml
     with:
-      runs-on: '["ubuntu-latest", "macos-latest", "windows-latest"]'
-      python-version: '["3.9", "3.10", 3.11, 3.12]'
+      runs-on: '["ubuntu-latest", "macos-latest", "windows-latest", "macos-14"]'
+      python-version: '["3.9", "3.10", "3.11", "3.12"]'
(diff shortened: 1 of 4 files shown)


Get4nik:
Hi @skoudoro I checked the error in the CI script I think the there is error is using the initializer 'Orthogonal' so I tried changing it  to 'orthogonal' in the file 'dipy/nn/cnn_1d_denoising.py' I hope it should be fixed after it . I am unable to figure out if it is fixed or not as it is showing the same error
if it is not fixed we can use alternate initializers like 'glorot_uniform' or 'he_normal' 
I would really like to help fixing this issue. Can you please help me out in this

.github/workflows/test.yml
```diff
@@ -20,5 +20,5 @@ jobs:
   pip:
     uses: ./.github/workflows/test_template.yml
     with:
-      runs-on: '["ubuntu-latest", "macos-latest", "windows-latest"]'
+      runs-on: '["ubuntu-latest", "macos-latest", "windows-latest", "macos-14"]'
       python-version: '["3.9", "3.10", 3.11, 3.12]'
```
jhlegarreta, line 24: Looks like there are being captured correctly, but for the sake of consistency, it might be good to add double quotes for 3.11 an 3.12.

skoudoro:
Hi @Get4nik,

Thank you for helping. The creator of this algorithm is already taking care of this + some update. A pull request should appear today. 

After that, I will be able to check codecov status and see if we can move forward with this PR or not

.github/workflows/test.yml
skoudoro: done

skoudoro:
This PR is ready to go. Can someone have a look before merging? Thank you!

jhlegarreta (approved):

PR #3143, opened 2024-03-21
[NF] Add `dipy_gtable` cli

skoudoro:
This PR adds a new CLI named `dipy_gtable`. The main goal of this cli is to merged bvec and bval into one file named `*.gtab.csv`.  This new file will be required for the other CLI 's in a close future. 

to test it, just run one this command line: 
- `dipy_gtable --force dipy/data/files/small_64D.bval dipy/data/files/small_64D.bvec` 
- `dipy_gtable --force dipy/data/files/small_64D.bval dipy/data/files/small_64D.bvec --btens "LTE" `
- `dipy_gtable --force dipy/data/files/small_64D.bval dipy/data/files/small_64D.bvec --small_delta 0.01 --big_delta 0.03 --btens "LTE" `

then display the new file: `dipy_info gt_condensed.gtab.csv`


arokem:
Could we try to conform the parts that we can with https://mrtrix.readthedocs.io/en/latest/concepts/dw_scheme.html?

dipy/core/gradients.py
```diff
@@ -255,14 +255,11 @@ def tau(self):
 
     @auto_attr
     def qvals(self):
-        tau = self.big_delta - self.small_delta / 3.0
-        return np.sqrt(self.bvals / tau) / (2 * np.pi)
+        return np.sqrt(self.bvals / self.tau) / (2 * np.pi)
```
arokem, line 258: I don't think that `self.tau` is defined anywhere

skoudoro: self.tau is defined here: https://github.com/dipy/dipy/blob/85f62188ea53b3d1e4da31773b0202116d9470ae/dipy/core/gradients.py#L252-L254

skoudoro:
> Could we try to conform the parts that we can with https://mrtrix.readthedocs.io/en/latest/concepts/dw_scheme.html?

I was not aware, I will look into it and let you know.

skoudoro:
ok, I just look into @arokem,

Both are ASCII files (csv, txt, dat, etc...). So you mean:

- space separated instead of comma separated
- change the order (bvec, bval) instead of (bval bvec)
- not enforce the extension  instead of enforcing the extension
- do not insert header instead of inserting header ?

Also, we add more information like btensor, small_delta, big_delta, if available.

Can you tell me which one is important to respect. I personally would like to keep the header which help non-technical user to open the file in excel and understand it. The other point should be ok to update

dipy/core/gradients.py
arokem: Yep - you are right! Sorry for missing that.

arokem:
I just mean that we should spend some time thinking before we invent yet another file format. 

In terms of matching the `.b` files, I think that we get closer by using tab separated (rather than comma-separated; also more BIDS-friendly, if I am not mistaken), and inverting the order of b-values and b-vectors. 

One way to ask this is to see whether a file that is generated here could immediately be used by mrtrix, without any change. I am not sure what mrtrix does on i/o, so the added information (i.e., b-tensor) might throw this off, but it's worth trying.

skoudoro:
ok, thank you for the feedback! 

ok for tab separated and inverting b-values and b-vectors. I will update

Otherwise, I will play with other software to check what they do when the file contain extra information like `btensor` or `b0_threshold`.

Garyfallidis:
This is not the same purpose @arokem. Our version contains gradients, b-tensor encodings, small and big deltas etc. We will explain. And also provide a converter. Also we do not introduce a new file format.  This is just a csv file. We already have pandas as a dependency.

Garyfallidis:
Please see updated picture above.

Garyfallidis:
In addition, it makes sense for people to see b-values as the first column because this is what people look first. It is also part of our API we first provide b-values and then b-vectors.

PR #3150, opened 2024-03-25
BUG: Fix bug with nightly wheel build

larsoner:
Addresses https://github.com/dipy/dipy/issues/2979#issuecomment-2018508798

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -174,7 +174,7 @@ jobs:
         run: |
           set -eo pipefail
           ls -al ./dist/*.whl
-          python -m pip install ./dist/*.whl --extra-index-url "https://pypi.anaconda.org/scientific-python-nightly-wheels/simple" "numpy>=2.1.0.dev0"
+          python -m pip install ./dist/*.whl --only-binary="numpy,scipy,h5py,dipy" --index-url "https://pypi.anaconda.org/scientific-python-nightly-wheels/simple" "numpy>=2.1.0.dev0" "scipy>=1.14.0.dev0" "h5py"
           python -c "import dipy; print(dipy.__version__)"
           python -c "from dipy.align.imaffine import AffineMap"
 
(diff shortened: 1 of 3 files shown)


.github/workflows/nightly.yml
```diff
@@ -9,6 +9,8 @@ on:
     #        │ │ │ │ ┌───────────── day of the week (0 - 6 or SUN-SAT)
     #        │ │ │ │ │
     - cron: "0 0 * * 0,3"  # Every Sunday and Wednesday at midnight
+  pull_request:
+    branches: [master]
```
larsoner, line 13: Should probably revert this and the removals of the `if` clauses before merge, hence this PR has been changed to draft mode

larsoner:
Okay seems to be working!

https://github.com/dipy/dipy/actions/runs/8430465421/job/23086405171?pr=3150

I'll revert the TODO stuff then should be good to merge

.github/workflows/nightly.yml
```diff
@@ -142,6 +145,36 @@ jobs:
...
+      - name: Test wheel
+        run: |
+          set -eo pipefail
+          ls -al ./dist/*.whl
+          python -m pip install --only-binary="numpy,scipy,h5py" --index-url "https://pypi.anaconda.org/scientific-python-nightly-wheels/simple" "numpy>=2.1.0.dev0" "scipy>=1.14.0.dev0" h5py
+          python -m pip install ./dist/*.whl
+          python -c "import dipy; print(dipy.__version__)"
+          python -c "import numpy; assert int(numpy.__version__[0]) >= 2, numpy.__version__"
+          python -c "from dipy.align.imaffine import AffineMap"
+
```
skoudoro, line 177: I will merge this for now but I believe this section is not needed.

we should use `CIBW_TEST_command` instead to be sure to test all wheels and not only one wheel (the linux one).

I will update that on another PR. 

Thank you very much for this work.

skoudoro (approved):
Thank you a lot for this @larsoner!

I will go ahead and merge it but see my small comment below

Issue #3154, opened 2024-03-27
ENH: Add neighboring DWI correlation QC metric


mattcieslak:
## Description

We've found the Neighboring DWI Correlation to be very useful as an automated QC metric. It's currently only implemented in DSI Studio, but is easy to calculate and would be a great addition to DIPY. 

Some prior discussion [here](https://github.com/nipreps/mriqc/pull/1131).

skoudoro:
Hi @mattcieslak, 

Looking forward for this addition !

araikes:
@mattcieslak... Did you see the new DW contrast metric Frank just added? Ratio of NDC and orthogonal DWI correlation

arokem:
@araikes : that sounds interesting! Is there a paper that describes this metric? Is it also primarily used for QC? Any clear benefits over NDC?  Would it be easy to add to this module (in a subsequent PR)? How is "orthogonal DWI correlation" calculated?

mattcieslak:
> @mattcieslak... Did you see the new DW contrast metric Frank just added? Ratio of NDC and orthogonal DWI correlation

yes, I think these are really clever and am going to try them out as I recompute NDC on all my other data

araikes:
@arokem, honestly I know very little of the details. In theory it's a QC metric. I saw an X post from Frank about it and that's the extent (https://twitter.com/FangChengYeh/status/1773160062627783158). Still need to dig in but I knew that @mattcieslak was working on implementing the NDC metric here

PR #3156, opened 2024-03-27
ENH: Implement NDC from Yeh2019

mattcieslak:
Adds a function to address #3154.

What would you recommend to add for a test?


dipy/denoise/qc.py
```diff
@@ -0,0 +1,66 @@
...
+        vectorized_images = [
+            dwi_data[..., idx][mask > 0] for idx in dwi_indices]
+    else:
+        vectorized_images = [
+            dwi_data[..., idx].flatten() for idx in dwi_indices]
+
+    # Get a pseudo-qspace value for b>0s
+    qvecs = (np.sqrt(gtab.bvals)[:, np.newaxis] * gtab.bvecs)[dwi_mask]
+
+    neighbor_correlations = []
```
arokem, line 50: Couldn't this be preallocated for speed?

arokem (commented):
Looks good! Regarding testing, I'd suggest testing that correlation between a signal and itself is 1, maybe with different sets of b-vectors and b-values. Maybe also that this correlation doesn't change if one of the signals is scaled. For smoke testing, worth testing with different gradient schemes, with and without b=0, for example.

dipy/denoise/qc.py
```diff
@@ -0,0 +1,66 @@
+import numpy as np
+from dipy.core.geometry import cart_distance
+
+
+def neighboring_dwi_correlation(dwi_data, gtab, mask=None):
+    r"""Calculate the Neighboring DWI Correlation (NDC) from dMRI data.
```
skoudoro, line 6: `r` not needed. triple double quotes should be enough

dipy/denoise/qc.py
```diff
@@ -0,0 +1,66 @@
...
+    correlations. According to [Yeh2019], an NDC less than 0.4 indicates a
+    low quality image.
+
+    Parameters
+    ----------
+    dwi_data : 4D ndarray
+        dwi data on which to calculate NDC
+    gtab : dipy.core.gradients.GradientTable
+        Gradient table.
+    mask : 3D ndarray
```
skoudoro, line 18: `3D ndarray, optional` instead

skoudoro (commented):
Also there is no automatic module discovery.

So, this new file needs to be added in [`meson.build`](https://github.com/dipy/dipy/blob/master/dipy/denoise/meson.build).

Thank you for this. Personally, I need to read the paper

dipy/denoise/qc.py
mattcieslak: this is only written to and read once, so I'd favor a list for readability

arokem: Fair enough

mattcieslak:
should be good for review!

arokem:
LGTM!

ShreyasFadnavis:
Looks great! Thanks for this @mattcieslak ! -- Not sure if `denoise` is the right module for it though? 

I don't want people to miss this utility, just because they did not look in the right place.

Alternative modules could be `stats`?

skoudoro:
Is there any additional comment @arokem @ShreyasFadnavis or @araikes?? 

I think this PR is ready to get in.

skoudoro:
Thank you for this @mattcieslak !

Issue #3123, opened 2024-03-15
ENH: Add tutorial comparing vanilla DTI with warping a high-quality scan back to lower resolution dMRI data


alexrockhill:
With low resolution/direction dMRI data, you have limited options for tractography and will suffer from losing crossing fibers when using DTI. We could show this in a tutorial. You could also use a template with good labels and try and SDR morph it back, probably on the FA maps, that's a good question though. @pjsjongsung said he'd look around and see if this was done so making an issue for that info to persist.

alexrockhill:
A suggestion from Dianne Patterson was to use HCP data which seems like a no brainer.

PR #3157, opened 2024-03-28
ENH, WIP: Compare CSD on multishell data to DTI on downsampled data

alexrockhill:
Fixes https://github.com/dipy/dipy/issues/3123

### DIFF
--- a/doc/examples/reconst_compare.py
+++ b/doc/examples/reconst_compare.py
@@ -19,9 +19,7 @@ from dipy.denoise.localpca import mppca
 from dipy.core.gradients import gradient_table, unique_bvals_tolerance
 from dipy.io.gradients import read_bvals_bvecs
 from dipy.io.image import load_nifti
-from dipy.nn.evac import EVACPlus
 from dipy.segment.mask import median_otsu
-from dipy.align.imaffine import AffineMap
 
 from dipy.denoise.localpca import mppca
 from dipy.align import affine_registration, motion_correction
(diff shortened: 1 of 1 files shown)


doc/examples/reconst_compare.py
```diff
@@ -0,0 +1,150 @@
...
+data, affine = load_nifti(fraw)
+bvals, bvecs = read_bvals_bvecs(fbval, fbvec)
+gtab = gradient_table(bvals, bvecs)
+
+###############################################################################
+# For the sake of simplicity, we only select two non-zero b-values for this
+# example.
+
+bvals = gtab.bvals
+bvecs = gtab.bvecs
```
skoudoro, line 58: I do not think you need this, it is deffine above

doc/examples/reconst_compare.py
```diff
@@ -0,0 +1,150 @@
...
+###############################################################################
+# Load the dMRI data.
+
+data, affine = load_nifti(fraw)
+bvals, bvecs = read_bvals_bvecs(fbval, fbvec)
+gtab = gradient_table(bvals, bvecs)
+
+###############################################################################
+# For the sake of simplicity, we only select two non-zero b-values for this
+# example.
```
skoudoro, line 55: Can you add the number of b-value we have on this data and which one we select. it might not be obvious for some people when reading the code

doc/examples/reconst_compare.py
```diff
@@ -0,0 +1,150 @@
...
+
+bvals = gtab.bvals
+bvecs = gtab.bvecs
+
+sel_b = np.logical_or(np.logical_or(bvals == 0, bvals == 1000), bvals == 2000)
+data = data[..., sel_b]
+
+gtab = gradient_table(bvals[sel_b], bvecs[sel_b])
```
skoudoro, line 63: Note to myself: we might need an explicity function for that in DIPY

doc/examples/reconst_compare.py
```diff
@@ -0,0 +1,150 @@
...
+plt.imshow(np.rot90(maskdata[:, :, 9, 0]), cmap=plt.cm.bone)
+plt.show()
+
+denoised = mppca(data, mask=mask, patch_radius=2)
+
+axial_slice = 10
+plt.imshow(np.rot90(denoised[:, :, axial_slice, 0]), cmap=plt.cm.bone)
+plt.imshow(np.rot90(data[:, :, axial_slice, 0]), cmap=plt.cm.hot,
+           alpha=0.25)
+plt.show()
```
skoudoro, line 80: can you check if we do not have already a function in `dipy.viz.plotting`for that. if yes, use it, if no, create it and just call it in the tutorial

doc/examples/reconst_compare.py
```diff
@@ -0,0 +1,150 @@
...
+
+peaks = dp.peaks_from_model(model=qball_model, data=denoised,
+                            relative_peak_threshold=0.5,
+                            min_separation_angle=25,
+                            sphere=sphere, mask=mask)
+
+ap = shm.anisotropic_power(peaks.shm_coeff)
+
+plt.imshow(np.rot90(ap[:, :, axial_slice]), cmap=plt.cm.bone)
+plt.show()
```
skoudoro, line 92: same as above and explain what we see

doc/examples/reconst_compare.py
```diff
@@ -0,0 +1,150 @@
...
+wm = np.where(final_segmentation == 3, 1, 0)
+
+plt.imshow(np.rot90(denoised[:, :, axial_slice, 0]), cmap=plt.cm.bone)
+plt.imshow(np.rot90(np.where(gm[:, :, axial_slice], 0.5, np.nan)),
+           cmap=plt.cm.bone_r, vmin=0, vmax=1, alpha=0.5)
+plt.imshow(np.rot90(np.where(wm[:, :, axial_slice], 0.9, np.nan)),
+           cmap=plt.cm.bone, vmin=0, vmax=1, alpha=0.5)
+plt.imshow(np.rot90(np.where(csf[:, :, axial_slice], 1, np.nan)),
+           cmap=plt.cm.cool, alpha=0.5)
+plt.show()
```
skoudoro, line 109: same as above and explain what we see

skoudoro (commented):
Hi @alexrockhill,

Thank you for starting this. See below my comments.

Furthermore, the file need to be added in `doc/examples/_valid_examples.toml`.

I suppose the tutorial is not finished because I do not see any comparison.

EDIT: ok I just saw the WIP + draft, so it make sense.

alexrockhill:
I think I might just add this to the end of the existing tutorial instead

alexrockhill:
I'm having a bit of issues with this deciding what's the best way for comparison, I feel like just visualizing the corpus callosum like the other tutorials isn't really that telling about how much better having multishell data is. Is there phantom data to use to compare? Just thinking out loud and sharing in case anyone had advice, I haven't searched the tutorials but I will soon!

PR #3158, opened 2024-03-29
ENH: Remove filtering `UserWarning` warnings in test config file

jhlegarreta:
- STYLE: Define a unique `FURY` package absence import warning message
- ENH: Remove filtering `UserWarning` warnings in test config file

### DIFF
--- a/dipy/conftest.py
+++ b/dipy/conftest.py
@@ -18,7 +18,6 @@ https://github.com/nipy/nibabel/pull/556
 np.set_printoptions(legacy='1.13')
 
 warnings.simplefilter(action="default", category=FutureWarning)
-warnings.simplefilter("always", category=UserWarning)
 # List of files that pytest should ignore
 collect_ignore = ["testing/decorators.py", "bench*.py", "**/benchmarks/*"]
 
--- a/dipy/viz/__init__.py
+++ b/dipy/viz/__init__.py
@@ -4,12 +4,15 @@ import warnings
 from dipy.utils.optpkg import optional_package
 from dipy.viz.horizon.app import horizon
 
+fury_pckg_msg = (
+    "You do not have FURY installed. Some visualization functions might not "
+    "work for you. Please install or upgrade FURY using pip install -U fury. "
+    "For detailed installation instructions visit: https://fury.gl/"
+)
+
 # Allow import, but disable doctests if we don't have fury
 fury, has_fury, _ = optional_package(
-    'fury',
-    trip_msg="You do not have FURY installed. Some visualization functions"
-    "might not work for you. For installation instructions, please visit: "
-    "https://fury.gl/", min_version="0.10.0")
+    'fury', trip_msg=fury_pckg_msg, min_version="0.10.0")
 
 
 if has_fury:
@@ -17,11 +20,7 @@ if has_fury:
     from fury.data import DATA_DIR as FURY_DATA_DIR, fetch_viz_icons, read_viz_icons
 
 else:
-    warnings.warn(
-        "You do not have FURY installed. "
-        "Therefore, 3D visualization functions will not work for you. "
-        "Please install or upgrade FURY using pip install -U fury"
-        "For detailed installation instructions visit: https://fury.gl/")
+    warnings.warn(fury_pckg_msg)
 
 # We make the visualization requirements optional imports:
 _, has_mpl, _ = optional_package(


jhlegarreta:
Some thoughts about this:

- Briefly discussed in https://github.com/dipy/dipy/commit/5ea90189841bbe63d3339dd2e709fdbe53c9c30b#r135163729
- Despite `UserWarning` being the default category for a [`warning.warn`](https://docs.python.org/3/library/warnings.html#warnings.warn) message (i.e. when `None` is passed), maybe `None` is somehow being taken as the category and that is why the `pyproject.toml` filter rule is not working.
  Explicitly adding the warning category in https://github.com/dipy/dipy/commit/c304b3ccea85a30e1626b9ca5dab6a7e73375c11 did not fix this:
https://github.com/dipy/dipy/actions/runs/8485764336/job/23251115969#step:10:4401

 - Not sure if this can be related:
https://github.com/pytest-dev/pytest/issues/11483#issuecomment-1750306638
  Not sure whether we can somehow change the logic so that an exception is raised and caught immediately:

  ```
  try:
      from fury import actor, window, colormap, lib
      from fury import interactor, ui, utils, shaders
      from fury.data import (fetch_viz_icons, read_viz_icons,
                             DATA_DIR as FURY_DATA_DIR)
  except Exception:
      (...)
  ```

  Maybe it is necessary to define a custom warning class derived from `UserWarning` as the above `Exception` clause is maybe too broad/not a `UserWarning` one.

- The approach in 5a34815 has been to remove
https://github.com/dipy/dipy/blob/f765aa0d2cea8b594e366a50f581ab0b73913a47/dipy/viz/__init__.py#L22-L26

  altogether: I'd say that the behavior seems to match the expectations: no warning displayed in the summary when it is not present, `viz` tests are skipped, and tests pass when it is present.

  So:
  - If the user chooses not to install the `viz` dependencies, is it coherent to raise a warning in the module? They have made the choice not to install them, so I do not think raising the warning is justified as it is now.
   - For `matplotlib`, there is no `else` statement in https://github.com/dipy/dipy/blob/master/dipy/viz/__init__.py#L35 and `matplotlib` is also an optional dependency for the module. Maybe less extensively used than `FURY` but still a `viz` dependency if we only look at `pyproject.toml` and `viz.__init__.py`.

- An additional through about the `TripWire` message: the [`trip_msg`](https://github.com/dipy/dipy/blob/f765aa0d2cea8b594e366a50f581ab0b73913a47/dipy/viz/__init__.py#L10) is different from the warning message actually being displayed when `FURY` has not been installed, which comes from the [`warning.warn` statement](https://github.com/dipy/dipy/blob/f765aa0d2cea8b594e366a50f581ab0b73913a47/dipy/viz/__init__.py#L22). Not sure if the `trip_msg` is displayed in some circumstances.

  Is this expected? Is there a way to provide only one message for FURY e.g. by defining the message as string in the `__init__.py` file or accessing the `fury` `TripWire` instance `msg` attribute?

- I have not found an easy way to test this locally without testing all tests. So thoughts are welcome

skoudoro:
Hi @jhlegarreta,

Thank you for working on this.

> If the user chooses not to install the viz dependencies, is it coherent to raise a warning in the module? They have made the choice not to install them, so I do not think raising the warning is justified as it is now.

The warning appears only if you import `viz` module. So if the user do not install viz dependencies and do not import `viz` module, no warning.
In our case, pytest import all modules, that's why it appears.

> For matplotlib, there is no else statement 

in theory, something to update. Let see

> An additional through about the TripWire message

I agree, to update and maybe to just raise the TripWire message

> I have not found an easy way to test this locally without testing all tests

testing only 2 modules including viz does the job locally.

> Briefly discussed in https://github.com/dipy/dipy/commit/5ea90189841bbe63d3339dd2e709fdbe53c9c30b#r135163729

to follow-up with that, I tried to start a discussion in pytest repo : https://github.com/pytest-dev/pytest/discussions/12152.
Unfortunately, no answer, so either my request is not clear or either it is not possible.

Removing this warning is a temporary solution, we will have warning from other dependencies that we need to filter So we better figure out a way to solve this warning that we control if it make sense.

jhlegarreta:
> testing only 2 modules including viz does the job locally.

Not sure how to do this. I tried calling `pytest` on `test_streamline.py` and did not get the warning, only the tests being skipped.

Could only get it doing an `import dipy.viz` in an interactive python session.

> to follow-up with that, I tried to start a discussion in pytest repo : pytest-dev/pytest#12152.

Nice. Thanks. We can allow some more time to `pytest` maintainers to chime in.

> I agree, to update and maybe to just raise the TripWire message

OK.

> Removing this warning is a temporary solution, we will have warning from other dependencies that we need to filter So we better figure out a way to solve this warning that we control if it make sense.

OK. So keeping this as a draft for now. Will try to have another look as time permits.

jhlegarreta:
In 96faab6 I raised an exception, but looks like the filter still does not work:
https://github.com/dipy/dipy/actions/runs/8493607472/job/23267871007?pr=3158#step:10:4396

jhlegarreta:
@skoudoro any updates from the `pytest` crew and the discussion you opened?

skoudoro:
> @skoudoro any updates from the pytest crew and the discussion you opened?

Sorry, I did not take the time to follow up on this. I do not know when I will be able to look deeper on that.

skoudoro:
small note:  this old issue might be related to our problem https://github.com/pytest-dev/pytest/issues/3251.

skoudoro:
Hi @jhlegarreta,

So, I think, I have found the problem. 

it is coming from this line : https://github.com/dipy/dipy/blob/16d915346ce72e2c3b4ac77dfe19036444caa5dd/dipy/conftest.py#L21

Removing it solve the issue. I do not remember why I had to add this line for python2.7 but it is not needed anymore. 
Everything is much better managed

So, Can you update your PR and remove this line.

Thank you!

jhlegarreta:
:+1: Thanks Serge. I assume removing the statement indicated should be enough, so rolled back the `try/except` changes in https://github.com/dipy/dipy/commit/96faab66ca1ca9186cb57c17df715432a6dcdaa5 to see if is enough. We can add the `UserWarning` category as a style improvement in another commit, and the `try/except` back if CIs tell us that removing the statement is not enough.

skoudoro:
Thank you for pushing this @jhlegarreta!

jhlegarreta:
CI failures are unrelated.

> Thank you for pushing this @jhlegarreta!

This is team work. Thanks to you for having investigated further.

> It seems to work as expected!
> No warnings in many CI's !!! 🥳 🥳 🥳 (I did not check them all)

Yes, had a look at a few CIs and looks like your proposal does the job :fireworks:, e.g.

Stable:
https://github.com/dipy/dipy/actions/runs/8819001542/job/24209160676?pr=3158#step:10:4413
https://github.com/dipy/dipy/actions/runs/8819001542/job/24209161416?pr=3158#step:10:4440
https://github.com/dipy/dipy/actions/runs/8819001542/job/24209166099?pr=3158#step:10:4432

Optional deps:
https://github.com/dipy/dipy/actions/runs/8819001541/job/24209159273#step:10:3937

Parellelization:
https://github.com/dipy/dipy/actions/runs/8819001506/job/24209157470#step:10:4684

VIZ:
https://github.com/dipy/dipy/actions/runs/8819001509/job/24209159441#step:10:4778
https://github.com/dipy/dipy/actions/runs/8819001509/job/24209157791#step:10:4334

> can you create a follow-up PR to enable warnings as error with pytest? or you prefer that I do it?

I can do that. I assume that once this gets merged we want to call `pytest` as `pytest -W error::RuntimeWarning` or use the `"error"` flat in the warnings filter:
https://docs.pytest.org/en/7.1.x/how-to/capture-warnings.html#controlling-warnings
https://docs.python.org/3/library/warnings.html#the-warnings-filter

Probably the former, as the latter may make the code to be unable to proceed for users having different dependency versions that introduce warnings. Let me know what you think.

---

Unrelated to this PR: `PRE_WHEELS` is consistently reporting a pass, but in reality it is failing:
https://github.com/dipy/dipy/actions/runs/8819001502/job/24209158205#step:10:135

skoudoro (approved):
LGTM ! merging !

skoudoro:
> Unrelated to this PR: PRE_WHEELS is consistently reporting a pass, but in reality it is failing:

Yes,  I am aware, but not worried, I need to spend allocate time on this

> I can do that. I assume that once this gets merged we want to call pytest as pytest -W error::RuntimeWarning or use the "error" flat in the warnings filter:

I prefer error flat, let's be strict on the CI's with the latest version on all dependencies. if locally it pass, the contributor have to make sure that it also pass for new environment (or new contributor).

So -W error should only be at the CI level and not in `pyproject.toml`

jhlegarreta:
> I prefer error flat, let's be strict on the CI's with the latest version on all dependencies. if locally it pass, the contributor have to make sure that it also pass for new environment (or new contributor).
> So -W error should only be at the CI level and not in pyproject.toml

Yep, sorry, I meant `-W error`. PR #3201.

PR #3161, opened 2024-03-30
DOC: Fix `tri` parameter docstring in `viz.projections.sph_project`

jhlegarreta:
Fix `tri` parameter docstring in `viz.projections.sph_project`:
- Make the method signature parameter name match the corresponding name in the docstring.
- Rename the `matplotlib.tri` imported package name to avoid the `tri` argument name shadowing it. Prefer to rename the package instead of renaming the argument so as to avoid breaking the API.

Rename the returned variable name `ax` to `_ax` in the docstring example to avoid shadowing `ax` from outer scope.

Take advantage of the commit to make the docstring style consistent with the style used across the package.

### DIFF
--- a/dipy/viz/projections.py
+++ b/dipy/viz/projections.py
@@ -13,7 +13,7 @@ from dipy.testing.decorators import doctest_skip_parser
 
 matplotlib, has_mpl, setup_module = optional_package("matplotlib")
 plt, _, _ = optional_package("matplotlib.pyplot")
-tri, _, _ = optional_package("matplotlib.tri")
+mpl_tri, _, _ = optional_package("matplotlib.tri")
 bm, has_basemap, _ = optional_package("mpl_toolkits.basemap")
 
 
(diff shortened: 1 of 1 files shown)


dipy/viz/projections.py
```diff
@@ -13,7 +13,7 @@
 
 matplotlib, has_mpl, setup_module = optional_package("matplotlib")
 plt, _, _ = optional_package("matplotlib.pyplot")
-tri, _, _ = optional_package("matplotlib.tri")
+mpl_tri, _, _ = optional_package("matplotlib.tri")
```
skoudoro, line 16: is it use ? should we just remove it ?  I do not find any use of `mpl_tri`.

skoudoro (commented):
Thanks @jhlegarreta,

This PR is almost ready to go. See my small comment below

dipy/viz/projections.py
jhlegarreta: I know Serge, but I imagined that the package needs to be imported so that some sort of processing on the mesh can be completed. IIRC sometimes with `matplotlib` this happens, but I cannot point to an example out of my mind. Best would probably be to have a test and see if removing it makes any difference. Can be done in a separate PR.

skoudoro: ok, let's check that in a different PR. thank you for the feedback. merging

PR #3162, opened 2024-03-30
ENH: Fix variable potentially being referenced before assignment

jhlegarreta:
Fix variable potentially being referenced before assignment in `viz.projections.sph_project`.

Fixes:
```
Local variable 'fig' might be referenced before assignment
```


jhlegarreta:
Untested.

skoudoro (approved):
All good, merging

PR #3168, opened 2024-04-04
Feature/peaks tab revamp

maharshi-gor:
### Peaks Tab Revamp.

**Changes**

- This PR aims to make the peaks tab follow the newer architecture to create HorizonUIElement.
- It also reduces different methods of different slices and ranges to adapt to `partial` functionality to reduce code size.
- It introduces a new way to remove and add elements from the scene, from a tab using a callback approach.
- It also fixes the issues with synchronization of peaks, changing the view modes to cross-section. Now you can change the slice in the corresponding image and it will not change the view mode to cross-section but, it will be mindful of the currently selected slice value and when moved to cross-section view mode, will show a synchronized slice.

**Goals**

- With this PR we moved all the tabs to adapt to new methods of creating UI elements in Horizon.

- There will be a follow-up PR for clearing up all the old code.


skoudoro (approved):
Thank you for this work @maharshi-gor, merging

PR #3169, opened 2024-04-04
[RF] Add peaks generation to reconst workflows

skoudoro:
All in the title. 

Many of our reconstruction workflows were not saving the peaks which could be problematic if the user want to generate a tractogram. This PR fix mainly this issue.


skoudoro:
closing in favor of #1957 which was similar and already there

PR #3171, opened 2024-04-04
Clean up for tabs and tab manager

maharshi-gor:
### Clean up for tabs.

**Changes**

- This PR ensures that all the tabs do not have `tab_ui` while building the tab elements ensuring the **abstraction**.
- This PR ensures all the methods that are not used by any other class are kept private, ensuring the **encapsulation**.
- This PR replaces all the direct fury UI objects to the `HorizonUIElement` wrapper data class for maintaining the standard course of usage of the horizon.
- This PR moves methods common into the base `HorizonTab` class to ensure we are getting the benefits of **inheritance** and **polymorphism**.

**Goals**

- This PR aims to achieve intractability while having multiple visualizations in the window. The problem arises when you have an ROI, a Surface, or a Peak visualized. It creates problems in interacting with clusters or images. Both of them get placed behind the former objects and become non-interactable. Using show and hide methods in the toggler of the opacity makes it easier to remove and add the actors to the scene. If you use the toggle and turn the ROI off, the cluster or the image becomes intractable.

### DIFF
--- a/dipy/viz/horizon/app.py
+++ b/dipy/viz/horizon/app.py
@@ -454,13 +454,12 @@ class Horizon:
                 # It will be changed once all the elements wrapped in horizon
                 # elements.
                 text_block = build_label(HELP_MESSAGE, 18)
-                text_block.message = HELP_MESSAGE
 
                 self.help_panel = ui.Panel2D(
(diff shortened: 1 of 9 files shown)


skoudoro (approved):
All good, Thanks @maharshi-gor, merging

PR #3174, opened 2024-04-05
NF - initial directions from seed positions

gabknight:
This PR is related to #3089. It adds 2 functions:

- The function `dipy.direction.peaks.peaks_from_positions(.)` allows for extracting peaks at any 3D positions of an ODF image.
- The function `dipy.tracking.utils.seeds_directions_pairs(.)` allows matching seeds positions to the corresponding peaks. If more than 1 peak is available and selected at the position, the seed position is repeated for each peak. The output is 2 arrays of shape (N,3) corresponding to all selected peaks and their location.

Examples using those functions for tractography will be included in #3089.

### DIFF
--- a/dipy/direction/peaks.py
+++ b/dipy/direction/peaks.py
@@ -614,6 +614,6 @@ def peaks_from_positions(positions, odfs, sphere, relative_peak_threshold=.5,
         peaks, _, _ = peak_directions(odf, sphere, relative_peak_threshold,
                                       min_separation_angle, is_symmetric)
         n = min(npeaks, peaks.shape[0])
-        peaks_arr[i,:n,:] = peaks[:n, :]
+        peaks_arr[i, :n, :] = peaks[:n, :]
 
     return peaks_arr
--- a/dipy/direction/peaks.py
+++ b/dipy/direction/peaks.py
@@ -612,6 +612,9 @@ def peaks_from_positions(positions, odfs, sphere, *, relative_peak_threshold=.5,
 
     peaks_arr = np.zeros((len(positions), npeaks, 3))
 
+    if (positions.dtype not in [np.float64, float]):
+        positions = positions.astype(float)
+
     for i, s in enumerate(positions):
         odf = trilinear_interpolate4d(odfs, s)
         peaks, _, _ = peak_directions(odf, sphere, relative_peak_threshold,
--- a/dipy/direction/peaks.py
+++ b/dipy/direction/peaks.py
@@ -615,7 +615,7 @@ def peaks_from_positions(positions, odfs, sphere, *, relative_peak_threshold=.5,
         odf = trilinear_interpolate4d(odfs, s)
         peaks, _, _ = peak_directions(odf, sphere, relative_peak_threshold,
                                       min_separation_angle, is_symmetric)
-        n = min(npeaks, peaks.shape[0])
-        peaks_arr[i, :n, :] = peaks[:n, :]
+        nbr_peaks = min(npeaks, peaks.shape[0])
+        peaks_arr[i, :nbr_peaks, :] = peaks[:nbr_peaks, :]
 
     return peaks_arr
--- a/dipy/direction/peaks.py
+++ b/dipy/direction/peaks.py
@@ -604,6 +604,9 @@ def peaks_from_positions(positions, odfs, sphere, relative_peak_threshold=.5,
     -------
     peaks_arr : array (N, npeaks, 3)
     """
+    if positions.dtype is not np.float64:
+        positions = positions.astype(np.float64)
+
     peaks_arr = np.zeros((len(positions), npeaks, 3))
 
     for i, s in enumerate(positions):
--- a/dipy/direction/peaks.py
+++ b/dipy/direction/peaks.py
@@ -1,7 +1,7 @@
-import tempfile
 from itertools import repeat
 from multiprocessing import Pool
 from os import path
+import tempfile
 
 import numpy as np
 import scipy.optimize as opt
--- a/dipy/direction/peaks.py
+++ b/dipy/direction/peaks.py
@@ -1,21 +1,19 @@
+import tempfile
 from itertools import repeat
 from multiprocessing import Pool
 from os import path
-import tempfile
 
 import numpy as np
 import scipy.optimize as opt
 
+from dipy.core.interpolation import trilinear_interpolate4d
 from dipy.core.ndindex import ndindex
 from dipy.core.sphere import Sphere
 from dipy.data import default_sphere
 from dipy.reconst.eudx_direction_getter import EuDXDirectionGetter
 from dipy.reconst.odf import gfa
-from dipy.reconst.recspeed import (
-    local_maxima,
-    remove_similar_vertices,
-    search_descending,
-)
+from dipy.reconst.recspeed import (local_maxima, remove_similar_vertices,
+                                   search_descending)
 from dipy.reconst.shm import sh_to_sf_matrix
 from dipy.utils.deprecator import deprecated_params
 from dipy.utils.multiproc import determine_num_processes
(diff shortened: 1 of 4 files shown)


dipy/direction/peaks.py
```diff
@@ -572,3 +571,49 @@ def reshape_peaks_for_visualization(peaks):
         peaks = peaks.peak_dirs
 
     return peaks.reshape(np.append(peaks.shape[:-2], -1)).astype('float32')
+
+
+def peaks_from_positions(positions, odfs, sphere, relative_peak_threshold=.5,
+                         min_separation_angle=25, is_symmetric=True, npeaks=5):
```
skoudoro, line 577: We start to introduce keyword-only arguments in the codebase
Can you replace this by:

`def peaks_from_positions(positions, odfs, sphere, *, relative_peak_threshold=.5,
                         min_separation_angle=25, is_symmetric=True, npeaks=5):`

dipy/direction/peaks.py
```diff
@@ -572,3 +571,49 @@ def reshape_peaks_for_visualization(peaks):
...
+    ----------
+    position : array, (N, 3)
+        Voxel coordinates of the N positions.
+    odfs : array, (X, Y, Z, M)
+        Orientation distribution function (sphericla function) represented
+        on a sphere of M points.
+    sphere : Sphere
+        A discrete Sphere. The M points on the sphere correspond to the points
+        of the odfs.
+    relative_peak_threshold : float in [0., 1.]
```
skoudoro, line 591: `float, optional` instead

the interval should be in the description below

dipy/direction/peaks.py
```diff
@@ -572,3 +571,49 @@ def reshape_peaks_for_visualization(peaks):
...
+        Only peaks greater than ``min + relative_peak_threshold * scale`` are
+        kept, where ``min = max(0, odf.min())`` and
+        ``scale = odf.max() - min``.
+    min_separation_angle : float in [0, 90]
```
skoudoro, line 595: same as above. `float, optional`.

interval in the description below

dipy/direction/peaks.py
```diff
@@ -572,3 +571,49 @@ def reshape_peaks_for_visualization(peaks):
...
+        The minimum distance between directions. If two peaks are too close
+        only the larger of the two is returned.
+    is_symmetric : bool, optional
+        If True, v is considered equal to -v.
+    npeaks : int
```
skoudoro, line 600: `int, optional` instead

dipy/direction/peaks.py
```diff
@@ -572,3 +571,49 @@ def reshape_peaks_for_visualization(peaks):
...
+        The maximum number of peaks to extract at from each position.
+
+    Returns
+    -------
+    peaks_arr : array (N, npeaks, 3)
+    """
+    if positions.dtype is not np.float64:
+        positions = positions.astype(np.float64)
```
skoudoro, line 608: why does it need to be float64?  Do we need such a precision and take so much space ?

dipy/direction/peaks.py
```diff
@@ -572,3 +571,49 @@ def reshape_peaks_for_visualization(peaks):
...
+
+    peaks_arr = np.zeros((len(positions), npeaks, 3))
+
+    for i, s in enumerate(positions):
+        odf = trilinear_interpolate4d(odfs, s)
+        peaks, _, _ = peak_directions(odf, sphere, relative_peak_threshold,
+                                      min_separation_angle, is_symmetric)
+        n = min(npeaks, peaks.shape[0])
```
skoudoro, line 616: can you avoid one letter variable (`n`).  3 letters is the minimal

dipy/direction/tests/test_peaks.py
```diff
@@ -690,3 +688,49 @@ def test_reshape_peaks_for_visualization(rng):
...
+                       [0.0015, 0.0003, 0.0003]))
+    voxels = []
+    for i in range(27):
+        v, _ = multi_tensor(gtab, mevals, S0=100,
+                            angles=[(0, 0), (randint(0, 90), randint(0, 90))],
+                            fractions=[50, 50], snr=10)
+        voxels.append(v)
+    data = np.array(voxels).reshape((3, 3, 3, -1))
+
+    pam = peaks_from_model(model, data, default_sphere, return_odf=True,
```
skoudoro, line 713: When we run the tests, a warning appear. You need to filter this warning with
```
with warnings.catch_warnings():
            warnings.filterwarnings(
                "ignore", message=descoteaux07_legacy_msg,
                category=PendingDeprecationWarning)
```

See line 466 above, on the same file

dipy/tracking/utils.py
```diff
@@ -1048,3 +1047,47 @@ def min_radius_curvature_from_angle(max_angle, step_size):
         max_angle = np.pi / 2.0
     min_radius_curvature = step_size / 2 / np.sin(max_angle / 2)
     return min_radius_curvature
+
+
+def seeds_directions_pairs(positions, peaks, max_cross=-1):
```
skoudoro, line 1052: We start to introduce keyword-only arguments in the codebase
Can you replace this by:

`def seeds_directions_pairs(positions, peaks, *, max_cross=-1)`

dipy/tracking/utils.py
```diff
@@ -1048,3 +1047,47 @@ def min_radius_curvature_from_angle(max_angle, step_size):
...
+    Pair each seed to the corresponding peaks. If multiple peaks are available
+    the seed is repeated for each.
+
+    Parameters
+    ----------
+    positions : array, (N, 3)
+        Voxel coordinates of the N positions.
+    peaks : array (N, M, 3)
+        Peaks at each position
+    max_cross : int
```
skoudoro, line 1063: `int, optional instead`

dipy/tracking/utils.py
```diff
@@ -1048,3 +1047,47 @@ def min_radius_curvature_from_angle(max_angle, step_size):
...
+            or not len(peaks.shape) == 3
+            or not positions.shape[0] == peaks.shape[0]
+            or not positions.shape[1] == 3
+            or not peaks.shape[2] == 3
+            or not peaks.shape[1] > 0):
+        raise ValueError("The array shapes of the positions and peaks should"
+                         " be (N,3) and (N,M,3), respectively.")
+
+    seeds = []
+    directions = []
```
skoudoro, line 1083: Why not creating a np.array directly on the correct size. 

this should avoid the slow `append` operation. 

I think you should avoid to work with python list.

dipy/tracking/tests/test_utils.py
```diff
@@ -703,3 +702,28 @@ def test_curvature_angle():
     with pytest.warns(UserWarning):
         npt.assert_equal(min_radius_curvature_from_angle(0, 1),
                          min_radius_curvature_from_angle(np.pi/2, 1))
+
+
+def test_seeds_directions_pairs():
+    positions = np.random.random((10, 3))
```
skoudoro, line 708: np.random.random will be soon deprecated so we avoid it.

see abov,  in this file, the use of `set_random_number_generator` + `rng` instead.

skoudoro (changes requested):
Hi @gabknight,

Thank you for this. See below my comments. After fixing all of that, it should be good to go.

dipy/direction/peaks.py
gabknight: done.

dipy/direction/peaks.py
gabknight: thanks. done.

dipy/direction/peaks.py
gabknight: done.

dipy/direction/peaks.py
gabknight: `trilinear_interpolate4d` require `double`. Should I cast in the for loop instead, at the time of calling?

dipy/direction/peaks.py
skoudoro: it does not appear here as done, so might have done a mistake in your push.

dipy/direction/peaks.py
skoudoro: it does not appear here as done, so might have done a mistake in your push.

dipy/direction/peaks.py
skoudoro: it does not appear here as done, so might have done a mistake in your push.

dipy/direction/peaks.py
skoudoro: ok, Maybe we should update `trilinear_interpolate4d`  in a small PR to accept float and double

dipy/direction/tests/test_peaks.py
gabknight: There is an issue here. 

`Legacy=True` in my call to `peaks_from_model(.)`. However, the is no way to set the `Legacy=True` flag when creating a Dipy model (e.g. `model = CsaOdfModel(gtab, 4)`). Thus, when setting `Legacy=True` in `peaks_from_model`,  ODFs will always be wrongly scaled. I.e. the legacy option is set both in the model object and in `peaks_from_model`, if they are not equal, the results will be wrong.

TLDR: `peaks_from_model(Legacy=True)` will always output wrongly scaled ODFs.

Am I getting something wrong? @skoudoro @Garyfallidis

gabknight: To solve this, we could remove the `Legacy` parameter from `peaks_from_model`, and add it to all SH models. peaks_from_model should get the legacy option from the model parameter. SH models should set it correctly on object creation.

gabknight: fixed in this PR by setting `Legacy=True`

dipy/tracking/utils.py
gabknight: done.

dipy/tracking/utils.py
gabknight: I don't know the length of the list/array beforehand,  as the number peaks per voxel is variable. Is there a better way to do this with arrays?

dipy/tracking/utils.py
gabknight: fixed.

dipy/direction/peaks.py
gabknight: yes, fixed.

dipy/direction/tests/test_peaks.py
gabknight: My bad, the code is fine. The code go back to the ODF then to the selected SH basis.

gabknight: I created issue #3187 to add support for `Legacy=False` for SH basis model.

dipy/direction/peaks.py
skoudoro: ok, PR created in https://github.com/dipy/dipy/pull/3191

skoudoro:
Hi @gabknight, Do you have any update on this PR? it is almost ready to go. it would be great to be in before Friday

skoudoro:
All good! thank you @gabknight, merging

PR #3177, opened 2024-04-05
Loading Peaks faster with complete range and synchronization functionality.

maharshi-gor:
### Loading the peaks faster with a complete range.

**Changes**

- This PR changes the functionality back to FURY where we avoid calculating voxels that don't contain any peaks.
- Ensuring the sizes will still correspond to the actual image.

**Goals**

- This PR makes much faster than the current Horizon peaks loading time.

**Speed Up**

- Measuring the speedup on a Linux machine with below given specifications
Screenshot from 2024-04-10 11-51-18

- Measuring was done using peaks generated from stanford_hardi data.

| Before | After |
|--------|--------|
| 63s | 5s |

### DIFF
--- a/dipy/viz/horizon/visualizer/peak.py
+++ b/dipy/viz/horizon/visualizer/peak.py
@@ -99,7 +99,7 @@ class PeakActor(Actor):
 
         valid_dirs = directions[indices]
 
-        num_dirs = len(np.where(valid_dirs >= 0)[0])
+        num_dirs = len(np.nonzero(np.abs(valid_dirs).max(axis=-1) > 0)[0])
 
         pnts_per_line = 2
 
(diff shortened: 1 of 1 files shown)


skoudoro:
Hi @maharshi-gor,

Do you have benchmarks? to see how much faster

skoudoro:
Hi @maharshi-gor,

See my question above. Any ideas?

maharshi-gor:
Hello @skoudoro, I have updated the description to show the system info and the performance boost.

I opened up peaks using the master branch and current branch to compare.

skoudoro (approved):
Thank you for the description update.

Indeed, the performance improvement is significant.

merging

Issue #3175, opened 2024-04-05
Tract profiles in afq example look all wrong


arokem:
## Description
The tract profiles [here](https://docs.dipy.org/stable/examples_built/streamline_analysis/afq_tract_profiles.html#sphx-glr-examples-built-streamline-analysis-afq-tract-profiles-py) look all wrong. The values don't make sense, and the flat sections in the profile are also non-sense. I think that I've managed to narrow this down to the fact that the FA map that we are using, which is based on the bundle, does not work as expected.  

## Potential solution
In further exploring this, I just realized that *maybe* we are looking at data from two different subjects here (😱 ): one from the HCP and another from the "SNAIL dataset"? This is based on the docstring for the `fetch_bundles_2_subjects` docstring, but I don't know what the SNAIL dataset is. Does anyone here know? Ironically, I think that the code is probably correct otherwise, and would work when applied to other data, so my suggestion would be to redo this example with HBN data instead of the data we are currently using. Any objections to this path?

skoudoro:
Duplicate of #1985 and #2796 

So closing those issue and keeping this one.

> Any objections to this path?

No objection for this path

arokem:
Hahaha! Clearly I keep running into this and complaining, but never actually do anything about it, so there you go.

PR #3178, opened 2024-04-05
DOC: Fixes the AFQ tract profile tutorial.

arokem:
This also adds the capability to download AFQ derivatives for all of the HBN subjects from the FCP INDI bucket where those are stored as part of the HBN POD2 study.

Addresses #3175, hopefully once and for all.

### DIFF
--- a/dipy/data/fetcher.py
+++ b/dipy/data/fetcher.py
@@ -2398,7 +2398,83 @@ def fetch_hcp(subjects,
     return data_files, pjoin(my_path, study)
 
 
-def fetch_hbn(subjects, path=None):
+def _hbn_downloader(my_path, derivative, subjects, client):
+    base_dir = op.join(my_path, "HBN", 'derivatives', derivative)
+
+    if not os.path.exists(base_dir):
+        os.makedirs(base_dir, exist_ok=True)
+
+    data_files = {}
+
+    for subject in subjects:
+        initial_query = client.list_objects(
+            Bucket="fcp-indi",
+            Prefix=f"data/Projects/HBN/BIDS_curated/sub-{subject}/")
+        ses = initial_query.get('Contents', None)
+        if ses is None:
+            raise ValueError(f"Could not find data for subject {subject}")
+        else:
+            ses = ses[0]["Key"].split('/')[5]
+
+        query = client.list_objects(
+            Bucket="fcp-indi",
+            Prefix=f"data/Projects/HBN/BIDS_curated/derivatives/{derivative}/sub-{subject}/")  # noqa
+        query_content = query.get('Contents', None)
+        if query_content is None:
+            raise ValueError(
+                f"Could not find derivatives data for subject {subject}")
+        file_list = [kk["Key"] for kk in query["Contents"]]
+        sub_dir = op.join(base_dir, f'sub-{subject}')
+        ses_dir = op.join(sub_dir, ses)
+        if derivative == "qsiprep":
+            if not os.path.exists(sub_dir):
+                os.makedirs(os.path.join(sub_dir, 'anat'), exist_ok=True)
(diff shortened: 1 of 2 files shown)


skoudoro (commented):
Hi @arokem,

Thank you for this.

No sure the reason, but the download was quite long and TQDM did not show up. So I did not know if the terminal was freezing or something was happening. 

Can you look at this, and make sure that TQDM shows up.  Even when the data are already in place, we do not have any information like the other fetcher. 

Otherwise, Can you tell me if this result looks correct to you?
image

Apart from this 2 points, it looks good to go.

arokem:
The result looks correct. I'll look into the tqdm issue and report back.

dipy/data/fetcher.py
```diff
@@ -1980,7 +1980,83 @@ def fetch_hcp(subjects,
...
+
+    # Create the BIDS derivatives description file text
+    to_bids_description(base_dir,
+                        **{"Name": "HBN",
+                           "PipelineDescription": {'Name': 'qsiprep'}})
+
+    return data_files
+
+
+def fetch_hbn(subjects, path=None, include_afq=False):
```
skoudoro, line 2059: Also, We start to introduce keyword-only arguments in the codebase.

can you replace this by `def fetch_hbn(subjects, *, path=None, include_afq=False):`

arokem: Changed in [5123bcd](https://github.com/dipy/dipy/pull/3178/commits/5123bcd380b07225059cf8c7286c09678b831865). Just noting that I've noticed that this transition is breaking downstream code. I guess folks will have to update all their calls if they want to update dipy.

skoudoro: sorry for that

skoudoro:
Hi @arokem,

> No sure the reason, but the download was quite long and TQDM did not show up. So I did not know if the terminal was freezing or something was happening.

Do you have an update on this

skoudoro:
Hi @arokem, 

It would be great if you could finalize this PR.  It seems very close to go!

Thanks

arokem:
TBH, I am not sure what remains here. You mentioned that it takes a long time to download the data, and TQDM not showing up. The time is because this downloads a large amount of diffusion MRI and derivatives, and I can't replicate TQDM not showing up on my end.

arokem:
Would it be helpful if we download only the bits of the data that are directly used here maybe?  I can see pros and cons to that.

skoudoro:
> Would it be helpful if we download only the bits of the data that are directly used here maybe? I can see pros and cons to that.

I am ok with the long time to download. it was the TQDM issue. 

Let me try again to see if TQDM shows up.

skoudoro:
Can you also address the code format issue. See https://github.com/dipy/dipy/actions/runs/8715096640/job/23906448911?pr=3178#step:4:130

This is just spacing issue. Thanks !

doc/examples/afq_tract_profiles.py
```diff
@@ -31,44 +29,58 @@
...
-bundles_path = dpd.fetch_bundles_2_subjects()
-bundles_folder = bundles_path[1]
+subject = "NDARAA948VFH"
+session = "HBNsiteRU"
 
-cst_l_file = op.join(bundles_folder, "bundles_2_subjects", "subj_2", "bundles",
-                     "bundles_cst.left.trk")
-af_l_file = op.join(bundles_folder, "bundles_2_subjects", "subj_2", "bundles",
-                    "bundles_af.left.trk")
+fdict, path = fetch_hbn([subject], include_afq=True)
```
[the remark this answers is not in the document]
arokem: Fixed in last series of commits

arokem:
Now rebased on top of main and with ruff formatting applied.

skoudoro:
Thank you for the update @arokem. Merging since codespell and code format are ok, we do not have CI's for examples yet. 
something to fix in a close future.

Issue #3145, opened 2024-03-22
dki_fit errors from divide by zero


yizhwan:
dipy_fit_dki.sh ${dwi_path} ${bval_path} ${bvec_path} ${mask} gives the following error:

dear team, i am running dipy_fit_dk in a shell script, receiving the following error:
.py:497: RuntimeWarning: invalid value encountered in divide
  return 3 * np.sqrt(6) * determinant((A_squiggle / A_s_norm)) 

If this is due to brain mask edges or noisy fit in low kurtosis regions, can the error be handled in a better way?

Thanks

skoudoro:
Hi @yizhwan,

You can look at PR https://github.com/dipy/dipy/pull/3180 to see the fix. 
The next release on July will contains this fix.

Thanks to @RafaelNH !

PR #3180, opened 2024-04-09
BF: Fix bug in mode for isotropic tensors

RafaelNH:
Hi all.

This PR is to resolve an issue reported in #3145.

Although the issue was detected when running `dipy_fit_dki`, it appears to be a more general issue with the tensor mode calculation in `dti.py`. The calculation involves dividing by the norm of the deviatoric component of the diffusion tensor, leading to problems when the tensor is isotropic. To address this, I've adapted the code in `mode` to bypass the mode calculation for voxels where the norm of the deviatoric diffusion tensor is zero. In cases where the norm is zero, the mode value will be set to zero, which is appropriate since isotropic tensors should indeed have a mode of zero.

Let me know if there are any questions or comments.

### DIFF
--- a/dipy/reconst/dti.py
+++ b/dipy/reconst/dti.py
@@ -488,13 +488,20 @@ def mode(q_form):
         Images", Magnetic Resonance in Medicine, vol. 55, no. 1, pp. 136-146,
         2006.
     """
-
     A_squiggle = deviatoric(q_form)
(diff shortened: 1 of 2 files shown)


skoudoro (approved):
Hi @RafaelNH,

it makes sense.  I am going ahead and merge this PR.

Thank you for this update.

PR #3188, opened 2024-04-15
ENH: Add `pre-commit` to project `dev` dependencies

jhlegarreta:
Add `pre-commit` to project `dev` dependencies: will allow to run `ruff` as a pre-commit hook when installing DIPY in development mode.

Left behind in commit 91fd582.


skoudoro (approved):
Thanks @jhlegarreta , merging

PR #3189, opened 2024-04-16
Update affine_registration to clarify returns and make them consistent with docstring

FWijanto:
Changed name of return variable in the docstring to match the actual function return. Added a copy of starting_affine named final_affine after registrations to clarify the flow of data.


skoudoro (approved):
Thank you for this @FWijanto.

All good,  merging!

PR #3153, opened 2024-03-26
ENH: add save peaks to dipy_fit_dti, dki

alexrockhill:
Fixes https://github.com/dipy/dipy/issues/3122

Since I was there, I also added saving the peaks to the dki workflow and the generalized FA value to the dsi workflow for standardization.

Maybe you'd have a minute to review @skoudoro?

PR #3151, opened 2024-03-26
DKI Updates: (new radial tensor kurtosis metric, updated documentation and missing tests)

RafaelNH:
Hi all. 

I added a new DKI metric - the radial tensor kurtosis proposed by [Hansen et al., 2017](https://pubmed.ncbi.nlm.nih.gov/27539807/).

**Why another DKI metric?**
This radial tensor kurtosis metric is more robust to implausible negative kurtosis estimates than the standard radial kurtosis definition - see my simulation results in my repository [Radial_Tensor_Kurtosis](https://github.com/RafaelNH/Radial_Tensor_Kurtosis/blob/main/Radial_tensor_kurtosis_comparison_in_simulations.ipynb).

**Update in DOC/TESTS** I also used this opportunity to update some function documentation and tutorials of DKI (e.g. add reference to the paper reporting our DKI implementations). I also added tests for some lines of code that were not previously covered by tests - we have 100% test coverage for dki.py again!

Please let me know your comments/suggestions.

PR #3194, opened 2024-04-17
MAINT: fix warning

alexrockhill:
Fixes issue here https://github.com/dipy/dipy/pull/3153#issuecomment-2053689778.

I considered setting `legacy=False` in `peaks_from_model` but I only saw that in tests so I figured best just to capture and ignore the warning.

### DIFF
--- a/dipy/workflows/tests/test_reconst_dki.py
+++ b/dipy/workflows/tests/test_reconst_dki.py
@@ -1,5 +1,6 @@
 from os.path import join as pjoin
 from tempfile import TemporaryDirectory
+import warnings
 
 import numpy as np
 import numpy.testing as npt
@@ -10,7 +11,7 @@ from dipy.data import get_fnames
 from dipy.io.gradients import read_bvals_bvecs
 from dipy.io.image import load_nifti, load_nifti_data, save_nifti
 from dipy.io.peaks import load_peaks
-from dipy.reconst.shm import sph_harm_ind_list
+from dipy.reconst.shm import descoteaux07_legacy_msg, sph_harm_ind_list
 from dipy.workflows.reconst import ReconstDkiFlow
 
 
@@ -26,7 +27,11 @@ def test_reconst_dki():
 
         args = [data_path, bval_path, bvec_path, mask_path]
 
-        dki_flow.run(*args, out_dir=out_dir, extract_pam_values=True)
+        with warnings.catch_warnings():
+            warnings.filterwarnings(
+                "ignore", message=descoteaux07_legacy_msg,
+                category=PendingDeprecationWarning)
+            dki_flow.run(*args, out_dir=out_dir, extract_pam_values=True)
 
         fa_path = dki_flow.last_generated_outputs['out_fa']
         fa_data = load_nifti_data(fa_path)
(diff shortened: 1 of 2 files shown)


alexrockhill:
cc @jhlegarreta

alexrockhill:
Ah my bad, only saw and fixed the first one, will fix the other ones tomorrow.

alexrockhill:
Ok, oddly there's more warnings but I checked and none of them traced back to any of my recent changes https://github.com/dipy/dipy/actions/runs/8758110095/job/24038195569?pr=3194

jhlegarreta:
> Ok, oddly there's more warnings but I checked and none of them traced back to any of my recent changes

They come from PR #3151.

skoudoro:
> They come from PR https://github.com/dipy/dipy/pull/3151.

My bad, I should have check those warnings before merging.... 

Looking forward to put the flag to treat warning as error....

alexrockhill:
Ok to merge?

jhlegarreta (changes requested):
The way warnings are caught is the following:
https://github.com/dipy/dipy/blob/53c18e1997f70fc3ee1d742fd30429b208314e0c/dipy/workflows/tests/test_tracking.py#L75

An exact match to the message is sought.

doc/examples/reconst_dki.py
```diff
@@ -91,13 +91,13 @@
 
 from dipy.core.gradients import gradient_table
 from dipy.data import get_fnames
+from dipy.denoise.localpca import mppca
```
jhlegarreta, line 94: Done in https://github.com/dipy/dipy/pull/3197/files#diff-3883fbe25a03fa8e050608f8e3808888811b4e02f06c195b36a76083b331d4b7R94

We should document better how to set up/use the pre-commit hooks. Sorry for that.

In the meantime, the cleanest approach is to (assuming you have a python (virtual) environment set up to work on dipy):
1. Squash all your commits into a single commit (all these changes utitmately do a single thing, i.e. capturing the warnings properly) and follow the DIPY commit subject prefixes: https://github.com/dipy/dipy/blob/50ec2b60209e9945e9dedfd61beaa446f1054a2c/doc/devel/commit_codes.rst
1. Rebase on `master` (assumes `origin` is your remote fork, `upstream` is the dipy org repository):
    ```
    git checkout master
    git fetch upstream
    git merge upstream/master
    git push origin master
    git checkout {your_branch_name}
    git rebase origin/master
    ```
    You may need to solve the conflicts when rebasing on `master`.

    This way we do not introduce unnecessary merge commits.

1. Once that is done, in order to ensure that your local changes comply with the import sorting rules:
    ```
    pip install pre-commit
    pip install ruff
    ruff --fix
    ```
1. Add the changes if files are modified using and do a `git commit --amend --no-edit` if the feature/maint/bugfix/etc. change has been commited with an appropriate message (avoids introducing a new commit to tell about the style change, which is not informative of the feature/maint/bugfix/etc.) ; else, `git commit`.

Hope this helps.

alexrockhill:
Hopefully that should fix the out-of-sync-with-`master`-issues, thanks for the instructions

doc/examples/reconst_dki.py
jhlegarreta: Addressed in 3c2b823.

jhlegarreta (approved):
CI failures are unrelated.

I wish the commit message honored only the changes made, but approving to move forward.

Issue #3155, opened 2024-03-27
scipt.opt.approx_fprime not good enough


samcoveney:
I am checking through some of the tests for reconst/dti, and I am finding (after some changes made for an up and coming PR) that the checks on the non-linear-least-squares error function derivate using `scipy.opt.approx_fprime` are failing.

The tests assert that the analytical Jacobian and the numerical Jacobian are not equal.

This is not the first time I have encountered this issue (I came across it while doing this #2730 as well, and have found the same problem in other projects as well).

I decided to quickly check if approximation of the derivatives were the problem, so I installed some autograd code and got to work in the tests in `dipy\reconst\tests\test_dti.py` (look for `# === SEE HERE ===`):

```
    # === SEE HERE: import autograd libraries ===
    import autograd.numpy as npa  # Thinly-wrapped numpy
    from autograd import grad as grd    # The only autograd function you may ever need

    nlls = dti._NllsHelper()
    sigma = 1.4826 * np.median(np.abs(error - np.median(error)))
    for D in [D_orig, np.zeros_like(D_orig)]:

        # Test Jacobian at D
        args = [D, X, Y, 1/sigma**2]
        # NOTE: 1. call 'err_func', to set internal stuff in the class
        nlls.err_func(*args)
        # NOTE: 2. call 'jabobian_func', corresponds to last err_func call
        analytical = nlls.jacobian_func(*args)

        for i in range(len(X)):

            args = [X[i], Y[i], 1/sigma**2]

            # === SEE HERE: commented out ===
            #approx = opt.approx_fprime(D, nlls.err_func, 1e-8, *args)

            # === SEE HERE: use autodiff to approximate the derivative ===
            grad_err_func = grd(nlls.err_func) 
            approx = grad_err_func(D, X[i], Y[i], 1/sigma**2)

            print(approx - analytical[i])
            assert np.allclose(approx, analytical[i])

```
The tests now pass because, of course, the function `nlls.jacobian_func` really is the correct jacobian of `nlls.err_func`... I am printing the difference just to check, and the autograd perfectly performs the differentiation.

I suspect the problem is with `scipy.opt.approx_fprime` using only forward differences.

Suggestions welcome please, I am happy to attempt a fix as part of the PR I am finishing up. We could switch these tests off, but given that my next PR has involved making changes to this function, it would be good to have a working test.

(edit: I had to do `import autograd.numpy as np` in `dipy\reconst\dti.py` for this autograd to work)

[...]

PR #3196, opened 2024-04-20
Enhancing Gradient Approximation in DTI Tests #3155

tanishka321:
I've introduced a D_perturbed variable that is slightly different from the ground truth D_orig. By perturbing D away from the minimum, we avoid approximating the derivative at the minimum, as suggested. 

#3155

### DIFF
--- a/dipy/reconst/tests/test_dti.py
+++ b/dipy/reconst/tests/test_dti.py
@@ -5,8 +5,8 @@ import warnings
 import numpy as np
 import numpy.testing as npt
 import scipy.optimize as opt
-import autograd.numpy as npa  
-from autograd import grad as grd  
+import autograd.numpy as npa
+from autograd import grad as grd
 
 import dipy.core.gradients as grad
 import dipy.core.sphere as dps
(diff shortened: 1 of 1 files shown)


tanishka321:
Here is an outline I had been thinking of (soon will be addressing other issues), are these considerable?

skoudoro:
Thank you for doing this @tanishka321. Maybe @samcoveney can have a look since this is the initiator of the issue.

samcoveney:
Hi @tanishka321 thanks for the work. I think you may have misunderstood #3155 though - the original test with _non-perturbed_ D works with autograd already, meaning my suggestion in #3155 could be implemented mostly as is.

But note that this would (1) require to add `autograd` as a dependancy, which is why your tests have error `ERROR reconst/tests/test_dti.py - ModuleNotFoundError: No module named 'autograd'` and (2) would require using `autograd` in `dti.py` (as far as I am aware, or it is not possible to use `grad` on `nlls.err_func` - did your tests pass locally?)

I don't think this is the way to go if we can avoid it. Instead, I would suggest to check if you can get the original gradient estimation with `opt.approx_fprime` working with `D_perturbed`. If that works and passes tests, problem solved.

samcoveney:
any progress on this @tanishka321 ?

skoudoro:
I am closing this PR for now, but feel free to reopen it when you feel ready! 

Thank you for this work @tanishka321 and thank you for the review @samcoveney

PR #3201, opened 2024-04-24
TEST: Turn warnings into errors when calling `pytest` in CI testing

jhlegarreta:
Turn warnings into errors when calling `pytest` in CI testing.

### DIFF
--- a/tools/ci/run_tests.sh
+++ b/tools/ci/run_tests.sh
@@ -18,11 +18,11 @@ if [ "$COVERAGE" == "1" ] || [ "$COVERAGE" == true ]; then
     cp ../.coveragerc .;
     cp ../.codecov.yml .;
     # Run the tests and check for test coverage.
-    coverage run -m pytest -c pyproject.toml -svv --doctest-modules --verbose --durations=10 --pyargs dipy
+    coverage run -m pytest -c pyproject.toml -svv --doctest-modules --verbose --durations=10 --pyargs dipy -W error
     coverage report -m  # Generate test coverage report.
     coverage xml  # Generate coverage report in xml format for codecov upload.
 else
-    pytest -c pyproject.toml -svv --doctest-modules --verbose --durations=10 --pyargs dipy
+    pytest -c pyproject.toml -svv --doctest-modules --verbose --durations=10 --pyargs dipy -W error
 fi
 cd ..
 set +ex


skoudoro:
Thank you for this @jhlegarreta.

if all ok, it means:
- we have this flag for the cythonization 👍 
- we have this flag for pytest 👍 
- Next step, we need to set it up for the compilation, it seems there are some compilation warning
- Then, last but not least, doc generation

jhlegarreta:
Looks reasonable. Not sure where the compilation warnings can be checked (did not find them on a quick traversal to the CI logs, maybe they are silenced or maybe I did not look at the right place), but looks good.

Similar for the documentation. Setting up an e.g. weekly doc build may be useful so that errors/warnings can be inspected without requiring local builds, and thus be addressed as time permits. Unless the CI time limits prevent the build from completing. Maybe other platforms (CIrcleCI, etc.) allow for larger runs if we hit that limit (?).

skoudoro:
> Looks reasonable. Not sure where the compilation warnings can be checked (did not find them on a quick traversal to the CI logs, maybe they are silenced or maybe I did not look at the right place), but looks good.

here is an example: https://github.com/dipy/dipy/actions/runs/8822497881/job/24220827033#step:8:219
You find them in the install DIPY step in the CI's

> Similar for the documentation. Setting up an e.g. weekly doc build may be useful so that errors/warnings can be inspected without requiring local builds, and thus be addressed as time permits. Unless the CI time limits prevent the build from completing. Maybe other platforms (CIrcleCI, etc.) allow for larger runs if we hit that limit (?).

Yes I agree, it should be the first step before tracking the large amount of warnings

jhlegarreta:
OK, all CIs are now turning the `FURY` warning into an error; looks as if the `pyproject.toml` filter would be be overridden by the `-W` option

Have tried locally to remove the filter from `pyproject.toml` and adding it to the shell script call:
```
pytest -c pyproject.toml -svv --doctest-modules --verbose --durations=10 --pyargs dipy -W error,ignore:'.*You do not have FURY installed.*':UserWarning
```

or
```
pytest -c pyproject.toml -svv --doctest-modules --verbose --durations=10 --pyargs dipy -W error -W ignore:'.*You do not have FURY installed.*':UserWarning
```

but tests are not triggered due to some `conftest.py` loading error that I have not found how to solve.

Run out of ideas to try.

skoudoro:
> OK, all CIs are now turning the FURY warning into an error; looks as if the pyproject.toml filter would be be overridden by the -W option

I did not think about it but it make sense. So I propose another alternative.

> Run out of ideas to try.

look at #3202 


superseded by #3202, closing.

PR #3206, opened 2024-04-27
BF: Use `cnp.npy_intp` instead of `int` as counter

jhlegarreta:
Use `cnp.npy_intp` instead of `int` to as a counter/to iterate over arrays/as.


jhlegarreta:
Related to issue https://github.com/dipy/dipy/issues/423.

Not sure if compilation warnings are related to this.

skoudoro (approved):
LGTM, merging ! thanks @jhlegarreta

PR #3209, opened 2024-04-29
[CI] Move filterwarnings from pyproject to conftest 

skoudoro:
This PR is just a follow up of the discussion in #2593.

Since we adopted the no-warning policy, we needed a way to filterwarnings that was generated by optional dependencies during the import. 

Now, we handle that in the `conftest.py`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -180,5 +180,10 @@ Metrics = [
 [tool.pytest.ini_options]
 addopts = "--ignore=dipy/testing/decorators.py --ignore-glob='bench*.py' --ignore-glob=**/benchmarks/*"
 filterwarnings = [
-    'ignore:.*You do not have FURY installed.*:UserWarning',
+    # FURY
+    "ignore:.*You do not have FURY installed.*:UserWarning",
+    # pkg_resources (via Ray)
+    "ignore:Implementing implicit namespace packages.*:DeprecationWarning",
+    "ignore:Deprecated call to `pkg_resources.*:DeprecationWarning",
+    "ignore:pkg_resources is deprecated as an API.*:DeprecationWarning",
 ]


arokem (approved):
LGTM! Assuming the CI passes cleanly here, I think we can go ahead and merge this.

skoudoro:
I do not find an easy solution so I started a discussion on pytest: https://github.com/pytest-dev/pytest/discussions/12272

if no answer, we will have to stick to `pyproject.toml`

skoudoro:
I did not get any answer and not sure what to do here. Any recommandations @arokem or @jhlegarreta?

it seems I will have to go back to the `pyproject.toml`

jhlegarreta:
> I wonder if the MNE developers can provide any useful input. It looks like they managed to make this work on their project (?)

Have not looked into it, but sounds reasonable. If that ends up not leading us to a solution, I would host them in `pyproject.toml` for now and revise this at a later time.

skoudoro:
> I wonder if the MNE developers can provide any useful input. It looks like they managed to make this work on their project (?)

it seems they run pytest on the source. I do not see a package installation and then running tests on that package.
they do no use `--pyargs`.   When we do not use it, it works as well but I would prefer to keep the package installation step.

Another Solution is described here: https://stackoverflow.com/a/43747114/2149641. I think we were doing that with unittest before switching to pytest. Maybe we could implement back this and change our way to call pytest.

jhlegarreta:
Not sure if I'd go that way, especially if the above involves changing all test files. If we have the exceptions in the `pyproject.toml` file, at least there is only one block to change/have a look. And for now, the set of rules does not seem so convoluted (?).

skoudoro:
Ok, This PR is ready to go, feel free to merge it @arokem or @jhlegarreta

arokem:
Looks like there are a lot of warnings treated as errors on this build: https://github.com/dipy/dipy/actions/runs/9097752928/job/25006323276?pr=3209. 

Should we add these here as well?

jhlegarreta:
@arokem they were addressed in PR #3220.

jhlegarreta (approved):

PR #3210, opened 2024-05-02
Python 3.13: Fix tests for next Python release

penguinpee:
Python 3.13 will remove the deprecated alias `warn` from `logging`.

### DIFF
--- a/dipy/tracking/tests/test_tracking.py
+++ b/dipy/tracking/tests/test_tracking.py
@@ -1,4 +1,4 @@
-from logging import warn
+from logging import warning
 import warnings
 
 import nibabel as nib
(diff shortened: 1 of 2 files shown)


penguinpee:
A couple of docstring tests are failing as well with Python 3.13 since those will be stripped of indents. I'll push another commit shortly.

penguinpee:
Oh well. `sys.version_info <= (3, 12)` yields `False` when on Python 3.12.2. I changed it to `sys.version_info < (3, 13)`. I suppose I could also make it `sys.version_info.minor <= 12`.

Let me know if you prefer one over the other.

skoudoro (approved):
Thank you for this update @penguinpee,

merging

PR #3211, opened 2024-05-03
[RF] PMF Gen: from memoryview to pointer

skoudoro:
To help the parallelization of the tracking framework, we needed to convert some memoryview to pointer and make sure that `nogil` is allowed. 

This change has no impact to the current tracking framework. This is a simple redesign. 

In a close future, we mightnneed to change the `pmf` object


skoudoro:
This PR is ready to go @gabknight  or someone else !

Issue #3165, opened 2024-04-01
1.9.0 system test failures


mr-c:
## Description
In Debian, we are seeing two test failures for dipy 1.9.0 on 32-bit systems, including x86:
1.
```
______________________ test_shore_fitting_no_constrain_e0 ______________________

    def test_shore_fitting_no_constrain_e0():
>       asm = ShoreModel(data.gtab, radial_order=data.radial_order,
                         zeta=data.zeta, lambdaN=data.lambdaN,
                         lambdaL=data.lambdaL)
E       AttributeError: '_C' object has no attribute 'radial_order'

dipy/reconst/tests/test_shore.py:70: AttributeError
```
2.
```
____________________________ test_check_img_shapes _____________________________

pytestconfig = <_pytest.config.Config object at 0xf7369900>, args = ()
kwargs = {'rng': Generator(PCG64) at 0xE7C5A188}
rng = Generator(PCG64) at 0xE7C5A188, signature = <Signature (rng)>

    def _set_random_number_generator_wrapper(pytestconfig, *args, **kwargs):
        rng = np.random.default_rng(seed_v)
        kwargs['rng'] = rng
        signature = inspect.signature(func)
        if pytestconfig and 'pytestconfig' in signature.parameters.keys():
            output = func(pytestconfig, *args, **kwargs)
        else:
>           output = func(*args, **kwargs)

dipy/testing/decorators.py:95: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
dipy/viz/tests/test_util.py:52: in test_check_img_shapes
    data = 255 * rng.random((198, 233, 189, 14))
numpy/random/_generator.pyx:356: in numpy.random._generator.Generator.random
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   numpy.core._exceptions._ArrayMemoryError: Unable to allocate 931. MiB for an array with shape (198, 233, 189, 14) and data type float64

_common.pyx:307: MemoryError
```

Full i386 (really i686) build log: https://buildd.debian.org/status/fetch.php?pkg=dipy&arch=i386&ver=1.9.0-1&stamp=1711634908&raw=0

## Way to reproduce

[...]

PR #3215, opened 2024-05-10
tests: correct module-level setup

mr-c:
Fixes: https://github.com/dipy/dipy/issues/3165

See https://docs.pytest.org/en/7.4.x/how-to/xunit_setup.html

### DIFF
--- a/dipy/reconst/tests/test_shore.py
+++ b/dipy/reconst/tests/test_shore.py
@@ -23,7 +23,7 @@ class _C:
 data = _C()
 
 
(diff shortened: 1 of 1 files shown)


mr-c:
Build failures appear to be false positives due to other tests issuing warnings which are being treated as errors.

skoudoro (approved):
Hi @mr-c, 

Thank you for looking into this and apologize for taking the time to check this.

Looks good to me so merging despite the unrelated warnings as errors

PR #3216, opened 2024-05-11
STYLE: Format code using `ruff`

jhlegarreta:
- STYLE: Format code using `ruff`
- STYLE: Remove expression that has no effect
- STYLE: Do not derive `SimilarityMetric` from `abc.ABCMeta`
- STYLE: Remove cast to deprecated `long` type


jhlegarreta:
@skoudoro **Despite all CIs being failing, this is working**, tests are passing, e.g.:
https://github.com/dipy/dipy/actions/runs/9051840841/job/24868833931?pr=3216#step:10:4577

**Test failures** (warnings treated as errors) **are unrelated**:
https://github.com/dipy/dipy/actions/runs/9051840841/job/24868833931?pr=3216#step:10:4542

In fact, the `CVXPY` warning appeared yesterday on `master`:
https://github.com/dipy/dipy/actions/runs/9047369878/job/24859232082

So better to address it in a separate PR.

 Some comments:
- This addresses a large amount of style errors as reported in https://github.com/dipy/dipy/actions/runs/9045357134/job/24855013530
- Some minimal performance gains may be expected after removing unnecessary `dict` calls and rewritten as literals.
- The **codespell issues are unrelated**: they were not being detected because no GHA was running the checker:
  https://github.com/dipy/dipy/actions/runs/9051840842/job/24868829333?pr=3216#step:4:21
  Fixing that requires a deprecation cycle of `sagital`.
- The **remaining formatting issues** are indicative of potential bugs (`csdeconv`, and `combined_workflow_creation`) or require a parameter name deprecation cycle (`piesno`)
 https://github.com/dipy/dipy/actions/runs/9051840840/job/24868829334?pr=3216#step:4:122

  IMO, they should be done in separate PRs.

- There is a number of exclude rules that have been added to `ruff`:
  https://github.com/dipy/dipy/pull/3216/files#diff-26f436cb29eecc7bb4f9066747b43df8da81ac808955ce95c109f5d7aa778989R31

  Out of those, the following would need to be addressed:
  ```
  "dipy/align/__init__.py" = ["E402"]
  "dipy/direction/__init__.py" = ["F401", "F403"]
  "dipy/reconst/*.py" = ["B026"]
  "dipy/viz/__init__.py" = ["F401"]
  "doc/conf.py" = ["E402"]
  ```

  They require some more work and are not urgent IMO. Given the complexity of this PR, I'd advocate of merging this and working on the issues above separately.

jhlegarreta:
> Can you address the conflict and rebase or merge this PR with master since the cvxpy warning has been fixed. Thank you

Done. `code-format` issues are the expected ones: https://github.com/dipy/dipy/pull/3216#issuecomment-2106265738

dipy/align/cpd.py
```diff
@@ -184,48 +184,60 @@ class DeformableRegistration:
...
+        alpha=None,
+        beta=None,
+        low_rank=False,
+        num_eig=100,
+        max_iterations=None,
+        tolerance=None,
+        w=None,
+        *args,
+        **kwargs,
+    ):
```
skoudoro, line 201: can we change this rule for function ? This take too much space for no reason...

What do you think?  What is your opinion @arokem also ? 

I know that @Garyfallidis  prefer too maximize the line-length instead of 1 argument per line.

jhlegarreta: > can we change this rule for function ?

I'd say that this is determined by the line length: I am not aware of a setting that tells `ruff`: "add as many arguments as possible on the same line until the line length, and break the line only then". If you happen to find it, please let me know and I'll adapt it if Ariel and Elef are against the current style.

I know it is hard to read sometimes, and it requires effort to get used to it. This is one of prices to pay for the automation/consistency.

jhlegarreta: @skoudoro @Garyfallidis @arokem after rebasing the failures in `code-format` are again the expected ones (https://github.com/dipy/dipy/pull/3216#issuecomment-2106265738) so if CI's come green this should be considered for merging. Rebasing a PR with 4 commits changing 33K line is hard, also because changes need to be done to the first commit, and at times I forget. Also, note that I had to make >400 manual changes because `ruff` was finding statements that did not comply with PEP8 in the strict sense. So I'd be grateful if you could afford some time to consider this PR.

skoudoro: ok, I am going ahead and merge. but many formatting are still strange to me like my comment above and many array formatting. we might need use `fmt off and on` for the array.

Many new PRs will be needed to finalize/adapt this work

skoudoro: also, the correction for `SimilarityMetric` deriving from `abc.ABCMeta` seems incorrect following: https://docs.astral.sh/ruff/rules/meta-class-abc-meta/

skoudoro: Concerning import sorting, I got my answer in the links below:
https://github.com/astral-sh/ruff/issues/2600
https://docs.astral.sh/ruff/settings/#lint_isort_split-on-trailing-comma

jhlegarreta: > we might need use fmt off and on for the array.

Seems not very practical to me, and leads back to manual formatting/inconsistency. But feel free to have a go.

I will have a look at https://github.com/dipy/dipy/pull/3216#discussion_r1604243313 and https://github.com/dipy/dipy/pull/3216#discussion_r1604250350.

jhlegarreta: OK, had a look at the `ABC` inheritance. Unless other changes are made, a simple `class C(ABC):` will not do the job: you will then get
```
dipy/align/metrics.py:97:5: B027 `SimilarityMetric.use_static_image_dynamics` is an empty method in an abstract base class, but has no abstract decorator
dipy/align/metrics.py:136:5: B027 `SimilarityMetric.use_moving_image_dynamics` is an empty method in an abstract base class, but has no abstract decorator
```

and if you add the decorator to those functions, the CI's start failing, so either the methods need to be reimplemented in child classes with a simple `pass` or I do not see a way to inherit from `ABC`. @skoudoro feel free to have a stab at this.

PR #3221, opened 2024-05-16
Updating BundleWarp default value of alpha

BramshQamar:
This PR updates alpha from 0.3 to 0.5 in BundleWarp registration. Alpha controls the trade-off between regularizing the deformation and having points that match very closely: higher value allows minimal deformations.

This PR also corrects minor typos and mistakes in the documentation and code. For example, the labels "transforms" and "Gaussian kernel" were mistakenly swapped in the BundleWarp workflow.


doc/interfaces/bundlewarp_registration_flow.rst
```diff
@@ -34,22 +34,22 @@ moving bundle using a single regularization parameter, alpha (represented with
...
 Partially Deformable BundleWarp Registration
 --------------------------------------------
 
-Here, we partially deform/warp moving bundle to align it with static bundle.
+Here, we partially deform/warp the moving bundle to align it with the static bundle.
 partial deformations improve linear registration while preserving the anatomical
-shape and structures of the moving bundle. Here, we use relatively higher value
-of alpha=0.03. By default, BundleWarp partially deforms the bundle to preserve
+shape and structures of the moving bundle. Here, we use a relatively higher value
+of alpha=0.05. By default, BundleWarp partially deforms the bundle to preserve
```
arokem, line 47:
```suggestion
of alpha=0.5. By default, BundleWarp partially deforms the bundle to preserve
```

arokem (commented):
Can you say a bit more about why 0.5 is preferred as a default to 0.3? Is this based on findings in data that you have analyzed since? Is there a published comparison in one of your recent papers?

doc/interfaces/bundlewarp_registration_flow.rst
BramshQamar: Hi @arokem 

There is a subtle difference between registrations with alpha=0.3 and alpha=0.5, both resulting in fewer deformations. In our paper, we present results using both values (see [paper](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC9881938/)). However, please note that this preprint is not updated and does not include some recent experiments.

Here is one of the figures from paper demonstrating effect of different alpha values on registration:

image

And here is another example comparing results visually and using shape similarity score as quantitative metric with alpha=0.5 and alpha=0.3:
-------------------
**Static**
image
-------------------
**Affinely moved bundle** 
Shape similarity score between static and affinely moved bundle  =  0.513

-------------------
**Registered with alpha=0.5**
Shape similarity score between static and moved bundle  = 0.688

-------------------
**Registered with alpha=0.3**
Shape similarity score between static and moved bundle  = 0.697

-------------------

I've noticed some discrepancies between the partially deformable registration example in the API tutorial and the CLI tutorial. Specifically, the API tutorial uses an alpha value of 0.5 (see [API tutorial](https://docs.dipy.org/stable/examples_built/registration/bundlewarp_registration.html#sphx-glr-examples-built-registration-bundlewarp-registration-py)), while the CLI tutorial uses 0.3 but mistakenly lists it as 0.03 (see [CLI tutorial](https://docs.dipy.org/stable/interfaces/bundlewarp_registration_flow.html)). There are also a few typos and mistakes in the documentation.

I propose setting alpha to the highest recommended default value of 0.5 and correcting the typos in the examples. Using alpha=0.5 ensures that the registration process remains strict, avoiding heavy deformations. Users can always decrease the value if they wish to introduce more deformation to the bundles.

arokem: Yeah - that all sounds very reasonable to me. From the visuals at least, it seems that the difference between 0.3 and 0.5 is rather minor, but seems reasonable to default to 0.5. Thanks for all the details!

dipy/workflows/align.py
```diff
@@ -1042,7 +1043,7 @@ def run(
             new_tractogram, pjoin(out_dir, out_nonlinear_moved), header=moving_header
         )
 
-        df = pd.DataFrame(warp, columns=["transforms", "gaussian_kernel"])
+        df = pd.DataFrame(warp, columns=["gaussian_kernel", "transforms"])
```
arokem, line 1046: Was this in the wrong order before? Why wasn't it catastrophic?

BramshQamar: Yes, it was in the wrong order, but it was not catastrophic because we use the `df` data frame only to save the warp map and Gaussian kernel. We do not use it to apply any transforms to the bundle itself. Essentially, it was saving NumPy files of transforms and Gaussian kernel with incorrect names (their file names were interchanged).

arokem: Is there any test we could write to make sure that doesn't creep back in?

dipy/align/streamwarp.py
```diff
@@ -73,7 +76,7 @@ def bundlewarp(static, moving, dist=None, alpha=0.3, beta=20, max_iter=15, affin
     alpha : float, optional
         Represents the trade-off between regularizing the deformation and
         having points match very closely. Lower value of alpha means high
-        deformations (default 0.3)
+        deformations (default 0.5)
```
skoudoro, line 79: can you remove `(default 0.5)`. not needed in the docstring

dipy/workflows/align.py
```diff
@@ -973,7 +974,7 @@ def run(
             deformations. It is represented with λ in BundleWarp paper. NOTE:
             setting alpha<=0.01 will result in highly deformable registration
             that could extremely modify the original anatomy of the moving
-            bundle. (default 0.3)
+            bundle. (default 0.5)
```
skoudoro, line 977: can you remove `(default 0.5)`. not needed in the docstring

skoudoro (commented):
Hi @BramshQamar,

Thank you for this, can you address @arokem comments and the comments below.

It is almost ready to go and it would be great to be in the next release.

Garyfallidis:
Ping @BramshQamar :)

dipy/align/streamwarp.py
BramshQamar: Should I remove all the default values for different parameters (beta, max_iter, etc.) from the docstring? @skoudoro

skoudoro: Yes

dipy/workflows/align.py
BramshQamar: I made some changes. I moved this to BundleWarp function. Now, registration returns warp as a dataframe with "gaussian_kernel", "transforms" columns. No need to do it in the workflow (outside of the registration function). Also, added test function `test_transformation_dimensions()` that checks if column index and dimensions of transformation matrices are correct. Let me know if there's something else @arokem @skoudoro

BramshQamar:
@skoudoro ?

dipy/align/streamwarp.py
```diff
@@ -2,7 +2,7 @@
 
 import numpy as np
 from scipy.optimize import linear_sum_assignment
-
+import pandas as pd
```
skoudoro, line 5: pandas is an optional dependency

dipy/align/streamwarp.py
```diff
@@ -57,10 +57,11 @@ def find_missing(lst, cb):
     return [x for x in range(0, len(cb)) if x not in lst]
 
 
+
 @warning_for_keywords()
-def bundlewarp(
-    static, moving, *, dist=None, alpha=0.3, beta=20, max_iter=15, affine=True
-):
+def bundlewarp(static, moving, dist=None, alpha=0.5, beta=20, max_iter=15,
+               affine=True):
```
skoudoro, line 63: please keep the `*`. it is important.

dipy/workflows/align.py
skoudoro: ping on that

skoudoro (changes requested):
Hi @BramshQamar,


Can you address the conflict to fix.  Also, there are still some pep8 issues and see my comments below. 
Thank you for the future fixes

BramshQamar:
@skoudoro PR should be ready now

dipy/align/streamwarp.py
```diff
@@ -13,6 +12,7 @@
 from dipy.tracking.streamline import Streamlines, length, unlist_streamlines
 from dipy.viz.plotting import bundle_shape_profile
 
+pd, have_pd, _ = optional_package("pandas")
```
skoudoro, line 15: you did not import `optional_package`

dipy/align/streamwarp.py
```diff
@@ -57,10 +57,11 @@ def find_missing(lst, cb):
     return [x for x in range(0, len(cb)) if x not in lst]
 
 
+
 @warning_for_keywords()
-def bundlewarp(
-    static, moving, *, dist=None, alpha=0.3, beta=20, max_iter=15, affine=True
-):
+def bundlewarp(static, moving, *, dist=None, alpha=0.5, beta=20, max_iter=15,
+               affine=True):
```
skoudoro, line 63: formating issue, try to keep what is in red. just change the value

dipy/align/streamwarp.py
```diff
@@ -188,6 +188,8 @@ def bundlewarp(
         ty = ty.astype(float)
         deformed_bundle.append(ty)
         warp.append(pr)
+        
+    warp = pd.DataFrame(warp, columns=["gaussian_kernel", "transforms"])
```
skoudoro, line 192: if pandas is not installed, the function will fail. you might need to tests this

dipy/align/tests/test_streamwarp.py
```diff
@@ -81,3 +81,22 @@ def test_bundle_shape_profile():
 
     assert_equal(len(shape_profile), n)
     assert_equal(len(stdv), n)
+
+def test_transformation_dimensions():
```
skoudoro, line 85: this tests is failing when pandas is not installed since this is optional

skoudoro (commented):
> @skoudoro PR should be ready now

Hi @BramshQamar,

all CI's are failing, also, there are some pep8 issues. See below some additional comments to help you fix this PR. 

PR #3229, opened 2024-05-18
STYLE: Prefer using f-strings

jhlegarreta:
Prefer using f-strings.

Documentation:
https://peps.python.org/pep-0498/

Take advantage of the commit to simply implicitly concatenated strings
(e.g. "first part of string" + "second part of string") to a single line
(related to the `ISC001` `ruff` rule).

### DIFF
--- a/dipy/align/_public.py
+++ b/dipy/align/_public.py
@@ -248,8 +248,7 @@ def register_dwi_to_template(
         )
     else:
         raise ValueError(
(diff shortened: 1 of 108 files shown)


jhlegarreta:
Failures are unrelated.

skoudoro (approved):
Thanks for this @jhlegarreta, merging

PR #3232, opened 2024-05-18
STYLE: Simplify implicitly concatenated strings

jhlegarreta:
Simplify implicitly concatenated strings of the form:
```
msg = "first part of string " + "second part of string"
```

to a single string, i.e.
```
msg = "first part of string second part of string"
```

Related to the `ISC001` `ruff` rule.


skoudoro (approved):
LGTM! Thanks @jhlegarreta ! merging

PR #3233, opened 2024-05-18
STYLE: Set `stacklevel` argument explicitly to warning messages

jhlegarreta:
Set `stacklevel` argument explicitly to warning messages in
`testing.tests.test_testing.py` module. Set its value to 1, as opposed
to the `ruff` recommendation of using 2 [1], since otherwise the tests
do not pass: the `__warningregistry__` seems not to be created (and
cannot be queried) otherwise. This is related to the fact that the
`stacklevel` option to warnings causes the attribute to be set for the
module the warning was issued "in the name of", not necessarily the one
where `warn()` was called, e.g. if we specify a value larger than 1.

Fixes:
```
dipy/testing/tests/test_testing.py:52:9:
 B028 No explicit `stacklevel` keyword argument found
dipy/testing/tests/test_testing.py:56:9:
 B028 No explicit `stacklevel` keyword argument found
dipy/testing/tests/test_testing.py:60:9:
 B028 No explicit `stacklevel` keyword argument found
dipy/testing/tests/test_testing.py:65:9:
 B028 No explicit `stacklevel` keyword argument found
```

[1] https://docs.astral.sh/ruff/rules/no-explicit-stacklevel/

### DIFF
--- a/dipy/testing/tests/test_testing.py
+++ b/dipy/testing/tests/test_testing.py
@@ -49,19 +49,19 @@ def test_clear_and_catch_warnings():
     npt.assert_equal(getattr(my_mod, "__warningregistry__", {}), {})
     with dt.clear_and_catch_warnings(modules=[my_mod]):
         warnings.simplefilter("ignore")
-        warnings.warn("Some warning")
+        warnings.warn("Some warning", stacklevel=1)
     npt.assert_equal(my_mod.__warningregistry__, {})
     # Without specified modules, don't clear warnings during context
     with dt.clear_and_catch_warnings():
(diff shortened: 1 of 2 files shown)


jhlegarreta:
Failures are unrelated.

skoudoro:
it works but not sure to understand this. Especially, they said that ` stacklevel=1` by default, so why we need to precise it if it is already the default value.

I miss something. Do you have an idea @jhlegarreta ?

skoudoro (approved):
ok, Thanks @jhlegarreta,

Merging

PR #3234, opened 2024-05-18
STYLE: Make module level import be at top of file

jhlegarreta:
Make module level import be at top of file.

Fixes:
```
dipy/align/__init__.py:27:1: E402 Module level import not at top of file
```

Relocate the `np.float32` aliasing to an `align.utils.py` module to
avoid a circular import error.

Keep the list of semantically public names of the module (`__all__`) at
the `__init__.py` file, and avoid duplicating the list in the
`_public.py` file.

Relocate the `align.Bunch` class and the `align.VerbosityLevels` enum to
`utils` since they are general enough so as to be potentially used
elsewhere.

### DIFF
--- a/dipy/align/__init__.py
+++ b/dipy/align/__init__.py
@@ -1,29 +1,3 @@
-import numpy as np
-
-floating = np.float32
-
-
-class Bunch:
-    def __init__(self, **kwds):
-        r"""A 'bunch' of values (a replacement of Enum)
-
-        This is a temporary replacement of Enum, which is not available
-        on all versions of Python 2
-        """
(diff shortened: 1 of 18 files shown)


jhlegarreta:
Looks like the proposed way still creates a circular import on `floating`, even if I tried to move it to `align._public`, use a relative import in the `___init__.py` file, etc. so not quite sure how to fix this.

jhlegarreta:
Using a `utils.py` does the trick.

skoudoro:
closing in favor of #4037 

Thanks @jhlegarreta for this work

PR #3235, opened 2024-05-19
STYLE: Do not allow running unintended modules as scripts

jhlegarreta:
Do not allow running as scripts modules that are not intended for that: remove the `if __name__ == "__main__": pass` statements.


skoudoro (approved):
All Good! merging, thanks @jhlegarreta

PR #3237, opened 2024-05-21
Add support for tensor-valued spherical functions in `interp_rbf`

### LINKED ISSUE #3236: Allow `interp_rbf` to accept tensor-valued spherical functions
## Description
I am currently working with interpolation in the q-space. Let's say I have a series of DWIs of shape `(H, W, L, D)` where `D` is the number of observed diffusion directions. Given a set of new diffusion directions, I want to obtain the DWIs corresponding to these new diffusion directions. The output should be of shape `(H, W, L, new_D)` where `new_D` denotes the number of new diffusion directions.

I came across two methods for interpolating diffusion signals in the q-space, namely, `dipy.reconst.shm.sf_to_sh`, and `dipy.core.interpolation.interp_rbf`. 

- `sf_to_sh` supports **tensor-valued** spherical functions. Along with `sh_to_sf`, one can feed an `ndarray` of shape `(H, W, L, D)` into `sf_to_sh` and get an `ndarray` of shape `(H, W, L, new_D)` as the output of `sh_to_sf`.
- However, this is not the case for `interp_rbf`, which only support **scalar-valued** spherical functions. I noticed that `interp_rbf` is based on `scipy.interpolate.Rbf`, which support only **vector-valued** spherical functions when the kwarg `mode='N-D'` is supplied. But `interp_rbf` doesn't allow the user to supply this kwarg.
- Furthermore, SciPy has a new class `scipy.interpolate.RbfInterpolator`, which supports **tensor-valued** spherical functions of shape `(D, H, W, L)` (note that the diffusion direction dimension comes first here). Hence, it might be beneficial to change `interp_rbf` such that it wraps `RBFInterpolator` instead, which would allow users to interpolate **tensor-valued** spherical functions.

Support for tensor-valued spherical functions is useful since it allows one to vectorize the process of generating DWIs for new diffusion directions.

This should be a non-breaking change since `scipy.interpolate.RbfInterpolator` is fully compatible with **scalar-valued** spherical functions. The change should be very straight forward and would look something like the following:

```
def interp_rbf(data, sphere_origin, sphere_target, function='multiquadric', epsilon=None, smooth=0.1): 
    data = np.einsum('hwld -> dhwl', data)
    rbfi = RBFInterpolator(sphere_origin.vertices, data, kernel=function, smoothing=smooth, epsilon=epsilon)
    return np.einsum('dhwl -> hwld', rbfi(sphere_target.vertices))
```

kvttt:
Fixes #3236.

This PR adds support for tensor-valued spherical functions in `interp_rbf`. When the user calls this function, specifying `legacy=True` internally uses `scipy.interpolate.Rbf` which is the original behavior; specifying `legacy=False` internally uses `scipy.interpolate.RBFInterpolator`, which allows `data` to have arbitrary dimensions. 

It would be great if I could get some feedbacks on the following questions:
* I saw the kwarg `norm` is already pending deprecation and is not used in `scipy.interpolate.RBFInterpolator`. As of my initial commit, `norm` is ignored when `legacy=False`. Do we still want to keep `norm`?
* In the docstring, `function` is said to be one of `{'multiquadric', 'inverse', 'gaussian'}`, but both `scipy.interpolate.Rbf` and `scipy.interpolate.RBFInterpolator` support a little bit more than that (see [here](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.RBFInterpolator.html#scipy.interpolate.RBFInterpolator) and [here](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.Rbf.html#scipy-interpolate-rbf)). Should I change the docstring?
* In `Rbf`, when `epsilon=None`, it internally estimate the average distance between nodes. However, in `RBFInterpolator`, this behavior is gone. The user has to specify `epsilon` when `kernel` is not one of `{'linear', 'thin_plate_spline', 'cubic', or 'quintic'}`. Should I change the docstring to reflect this new behavior?

### DIFF
--- a/dipy/core/interpolation.pyx
+++ b/dipy/core/interpolation.pyx
@@ -16,8 +16,8 @@ from dipy.utils.deprecator import deprecate_with_version
     "dipy.core.interpolation.interp_rbf is deprecated, "
     "Please use "
     "dipy.core.interpolation.rbf_interpolation instead",
-    since="1.9",
-    until="2.0",
+    since="1.10.0",
+    until="1.12.0",
 )
 def interp_rbf(data, sphere_origin, sphere_target,
                function='multiquadric', epsilon=None, smooth=0.1,
@@ -93,7 +93,7 @@ def interp_rbf(data, sphere_origin, sphere_target,
     return rbfi(sphere_target.x, sphere_target.y, sphere_target.z)
 
 
-def rbf_interpolation(data, sphere_origin, sphere_target,
+def rbf_interpolation(data, sphere_origin, sphere_target, *,
                       function='multiquadric', epsilon=None, smoothing=0.1):
     """Interpolate `data` on the sphere, using radial basis functions,
     where `data` can be scalar- (1D), vector- (2D), or tensor-valued (3D and beyond).
@@ -107,14 +107,15 @@ def rbf_interpolation(data, sphere_origin, sphere_target,
     sphere_target : Sphere
         M positions on the unit sphere where the spherical function is interpolated.
 
-    function : {'linear', 'thin_plate_spline', 'cubic', 'quintic', 'multiquadric', 'inverse_multiquadric',
-    'inverse_quadratic', 'gaussian'}
-        Radial basis function. Default: `thin_plate_spline`.
-    epsilon : float
+    function: str, optional
+        Radial basis function.
+        Possible values: {'linear', 'thin_plate_spline', 'cubic', 'quintic',
+        'multiquadric', 'inverse_multiquadric', 'inverse_quadratic', 'gaussian'}.
+    epsilon : float, optional
         Radial basis function spread parameter.
         Defaults to 1 when `function` is 'linear', 'thin_plate_spline', 'cubic', or 'quintic'.
         Otherwise, `epsilon` must be specified.
-    smoothing : float
+    smoothing : float, optional
         Smoothing parameter. When `smoothing` is 0, the interpolation is exact.
         As `smoothing` increases, the interpolation approaches a least-squares fit of `data`
         using the supplied radial basis function. Default: 0.
@@ -131,7 +132,7 @@ def rbf_interpolation(data, sphere_origin, sphere_target,
     """
     from scipy.interpolate import RBFInterpolator
 
-    last_dim_idx = len(data.shape) - 1
+    last_dim_idx = data.ndim - 1
     if not data.shape[last_dim_idx] == sphere_origin.vertices.shape[0]:
         raise ValueError("The last dimension of `data` must be equal to the number of "
                          "vertices in `sphere_origin`.")
--- a/dipy/core/interpolation.pyx
+++ b/dipy/core/interpolation.pyx
@@ -13,12 +13,12 @@ from dipy.align.fused_types cimport floating, number
 
 def interp_rbf(data, sphere_origin, sphere_target,
                function='multiquadric', epsilon=None, smooth=0.1,
-               norm="angle"):
+               norm="angle", legacy=True):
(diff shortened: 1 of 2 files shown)


kvttt:
I am not sure how to fixed the two failing checks.

skoudoro:
Hi @kvttt,

Thank you for working on this! 

I would recommend to create a new function named `rbf_interpolation` and deprecate the `interp_rbf` by adding our decorator ([see an example here](https://github.com/dipy/dipy/blob/4a6bbf7d1de4f8ae110f3dee40715ee2a7bb5f16/dipy/tracking/metrics.py#L163C9-L163C25))

I do not see any benefit to keep all this different behaviors (if legacy or not )and add complexity to the maintenance. 1 function = 1 feature/task if possible. 

> Do we still want to keep norm?

no

>  Should I change the docstring?

Yes

> Should I change the docstring to reflect this new behavior?

No, since we plan to remove this version and keep only your new version

Can you update this?  Thank you !

skoudoro:
> I am not sure how to fix the two failing checks.

Concerning the checks, no worries concerning this 2 checks

dipy/core/interpolation.pyx
```diff
@@ -9,8 +9,16 @@ import warnings
...
 
 from dipy.align.fused_types cimport floating, number
+from dipy.utils.deprecator import deprecate_with_version
 
 
+@deprecate_with_version(
+    "dipy.core.interpolation.interp_rbf is deprecated, "
+    "Please use "
+    "dipy.core.interpolation.rbf_interpolation instead",
+    since="1.9",
```
skoudoro, line 19: `1.10.0` instead

dipy/core/interpolation.pyx
```diff
@@ -9,8 +9,16 @@ import warnings
...
+    until="2.0",
```
skoudoro, line 20: `1.12.0` instead

dipy/core/interpolation.pyx
```diff
@@ -85,6 +93,54 @@ def interp_rbf(data, sphere_origin, sphere_target,
     return rbfi(sphere_target.x, sphere_target.y, sphere_target.z)
 
 
+def rbf_interpolation(data, sphere_origin, sphere_target,
+                      function='multiquadric', epsilon=None, smoothing=0.1):
```
skoudoro, line 97: use keyword-only. 

`def rbf_interpolation(data, sphere_origin, sphere_target, *, function='multiquadric', epsilon=None, smoothing=0.1):` instead

skoudoro (commented):
Hi @kvttt,

Overall, it looks good. see below my last quick comments.  

@arokem or @jhlegarreta, can you have a look?

skoudoro:
Thank you for the quick update @kvttt! 

is there any additional comment ? @arokem ? @Garyfallidis ? someone else ?

dipy/core/interpolation.pyx
```diff
@@ -85,6 +93,54 @@ def interp_rbf(data, sphere_origin, sphere_target,
...
+    -------
+    v : (..., M) ndarray
+        Interpolated values of the spherical function at M positions specified by `sphere_target`.
+
+    See Also
+    --------
+    scipy.interpolate.RBFInterpolator
+
+    """
+    from scipy.interpolate import RBFInterpolator
```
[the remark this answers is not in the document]
arokem: I don't think there any good reason to put the import in here (there's a high likelihood I originally did this).

arokem (commented):
Overall, looks good to me!

skoudoro (approved):
Thank you for this @kvttt! 

merging

PR #3238, opened 2024-05-24
NF - add affine to peaks_from_position

gabknight:
`peaks_from_positions(.)` requires the positions to be in voxel coordinates. This is fine, but it doesn't play well with our seed generation functions that output seeds in world coordinates. This was an oversight in PR #3174.

This PR updates `peaks_from_positions(.)` to have an additional `affine` parameters, and require coordinates to be in world space.


dipy/direction/peaks.py
```diff
@@ -679,11 +680,13 @@ def peaks_from_positions(
     position : array, (N, 3)
         Voxel coordinates of the N positions.
```
skoudoro, line 681: This description seems to be wrong since we have the affine now. 

It should be `World coordinates`. Can you check and correct.

skoudoro (commented):
Hi @gabknight,

Sorry for the late review. Looks good to me. See below a small comment to fix before merging

dipy/direction/peaks.py
gabknight: Good catch. Fixed, thanks.

gabknight:
Thanks @skoudoro, I fixed the doc. If the tests pass, it will be ready to merge.

skoudoro (approved):
All good , thank you for this @gabknight!, merging

Issue #3029, opened 2024-01-05
Migrating to Keyword Only arguments (PEP 3102)


skoudoro:
## Description

Hi all, 

To simplify maintenance and avoid to messing around with function arguments order (it happens many times already), it would be great if we could switch our codebase to keyword-only arguments (see [pep 3102](https://peps.python.org/pep-3102/)).

To do a smooth transition, we might need to create a decorator that will warn the user about this feature. The decorator should stay in place for 3-4 releases (all this year (2024 and we remove it in 2025).

Please, Let me know what you think about this new feature or if you are against. 

I would like to start this work earliest as possible. 

ps: if we agree, all new function in DIPY codebase should follow this rules. 

Thank for the feedback.

pjsjongsung:
Sounds reasonable! I had some related troubles during a PR last time, especially when using/changing functions written by another contributor, and this could really help.

samcoveney:
Do you know of a way to check this using some tool? Would make the transition easier

skoudoro:
For now, no, I do not know any tool.  With might do like scikit-learn,  create a decorator to warn user of this change, and apply this decorator in every single function in the codebase....

skoudoro:
Closing since the decorator is merged #3239, and there is a whole code base update coming

PR #3239, opened 2024-05-27
NF: Decorator for keyword-only argument

deka27:
# Migrating to Keyword Only arguments

## Overview
This PR introduces new feature of decorator function for Keyword-Only arguments as mentioned in issue #3029.

## Changes
- Added Decorator function in decorators.py
- Added Testing function in test_decorators.py
- Added decorator function to functions in stats module

## Related Issues
- #3029 

## Modules Implementation

- [x] stats - #3239 
- [x] align - #3249 
- [x] core - #3251 
- [x] data - #3253 
- [x] denoise - #3252 
- [x] direction - #3254 
- [x] io - #3255 
- [x] nn - #3256 
- [x] reconst - #3257 
- [x] segment - #3258 
- [x] sims - #3259 
- [x] tracking - #3260 
- [x] utils - #3261 
- [x] viz - #3262 
- [x] workflows - #3263

### DIFF
--- a/dipy/stats/resampling.py
+++ b/dipy/stats/resampling.py
@@ -3,6 +3,8 @@
 import numpy as np
 import scipy as sp
 
+from dipy.testing.decorators import warning_for_keywords
+
 
 def bs_se(bs_pdf):
     """Calculate the bootstrap standard error estimate of a statistic."""
@@ -133,7 +135,7 @@ def abc(x, *, statistic=bs_se, alpha=0.05, eps=1e-5):
     # define helper variables -- w and l
     w = z_0 + __calc_z_alpha(1 - alpha)
     ell = w / (1 - a_hat * w) ** 2
-    return __tt(x, p_0 + ell * delta_hat / sigma_hat, statistic)
+    return __tt(x, p_0 + ell * delta_hat / sigma_hat, statistic=statistic)
 
 
 def __calc_z_alpha(alpha):
@@ -157,14 +159,23 @@ def __calc_z0(x, p_0, statistic, eps, a_hat, sigma_hat):
         tt_dot[i] = __tt_dot(i, x, p_0, statistic, eps)
     b_hat = b_hat / (2 * n**2)
     c_q_hat = (
-        __tt(x, ((1 - eps) * p_0 + eps * tt_dot / (n**2 * sigma_hat)), statistic)
-        + __tt(x, ((1 - eps) * p_0 - eps * tt_dot / (n**2 * sigma_hat)), statistic)
-        - 2 * __tt(x, p_0, statistic)
+        __tt(
+            x,
+            ((1 - eps) * p_0 + eps * tt_dot / (n**2 * sigma_hat)),
+            statistic=statistic,
+        )
+        + __tt(
+            x,
+            ((1 - eps) * p_0 - eps * tt_dot / (n**2 * sigma_hat)),
+            statistic=statistic,
+        )
+        - 2 * __tt(x, p_0, statistic=statistic)
     ) / eps**2
     return a_hat - (b_hat / sigma_hat - c_q_hat)
 
 
-def __tt(x, p_0, statistic=bs_se):
+@warning_for_keywords()
+def __tt(x, p_0, *, statistic=bs_se):
     """Calculate desired statistic from observable data and a
     given proportional weighting.
 
@@ -192,7 +203,8 @@ def __tt_dot(i, x, p_0, statistic, eps):
     e = np.zeros(x.shape)
     e[i] = 1
     return (
-        __tt(x, ((1 - eps) * p_0 + eps * e[i]), statistic) - __tt(x, p_0, statistic)
+        __tt(x, ((1 - eps) * p_0 + eps * e[i]), statistic=statistic)
+        - __tt(x, p_0, statistic=statistic)
     ) / eps
 
 
@@ -202,7 +214,10 @@ def __tt_dot_dot(i, x, p_0, statistic, eps):
     e[i] = 1
     return (
         __tt_dot(i, x, p_0, statistic, eps) / eps
-        + (__tt(x, ((1 - eps) * p_0 - eps * e[i]), statistic) - __tt(x, p_0, statistic))
+        + (
+            __tt(x, ((1 - eps) * p_0 - eps * e[i]), statistic=statistic)
+            - __tt(x, p_0, statistic=statistic)
+        )
         / eps**2
     )
 
--- a/dipy/testing/decorators.py
+++ b/dipy/testing/decorators.py
@@ -4,12 +4,18 @@
 Decorators for dipy tests
 """
 
+from functools import wraps
 import inspect
+from inspect import Parameter, signature
 import os
 import platform
 import re
+import warnings
 
 import numpy as np
+from packaging import version
+
+import dipy
 
 SKIP_RE = re.compile(r"(\s*>>>.*?)(\s*)#\s*skip\s+if\s+(.*)$")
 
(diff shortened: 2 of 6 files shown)


deka27:
Will correct all the format styling in one go at the end.

dipy/testing/decorators.py
```diff
@@ -103,3 +104,26 @@ def _set_random_number_generator_wrapper(pytestconfig, *args, **kwargs):
         return _set_random_number_generator_wrapper
 
     return _set_random_number_generator
+
+def warn_on_positional_args(valid=""):
```
itellaetxe, line 108: `valid` may be too vague to know what this argument means. If I understood correctly, this references the version in which the deprecation will start, so maybe the name could be something like: `deprecates_from_version`, `invalid_from_version`.

itellaetxe: The docstring is missing

itellaetxe (commented):
Overall it LGTM, I added a couple of suggestions regarding the decorator definition.

dipy/testing/tests/test_decorators.py
```diff
@@ -51,3 +51,36 @@ def f():
     del HAVE_AMODULE
     f.__doc__ = docstring
     assert_raises(NameError, doctest_skip_parser, f)
+
+def test_warn_on_positional_kwonly():
```
itellaetxe, line 55: Maybe it would be easier to test this concept as @WassCodeur did in [his implementation](https://github.com/fury-gl/fury/pull/888) of this same decorator for Fury. He defines only 1 function and tests it in all possible cases, instead of two functions. I think it is simpler.

dipy/testing/decorators.py
```diff
@@ -103,3 +104,26 @@ def _set_random_number_generator_wrapper(pytestconfig, *args, **kwargs):
...
+
+            # Check the number of positional arguments
+            num_positional_args = len(args)
+            for kwarg in kwonly_args:
+                if kwarg in param_names[:num_positional_args]:
+                    warning_msg = (f"Warning: Function '{func.__name__}' keyword-only argument '{kwarg}' is being "
+                                   f"called positionally. This will be deprecated from version '{valid}'.")
+                    warnings.warn(warning_msg, stacklevel=2)
+            
+            return func(*args, **kwargs)
```
itellaetxe, line 127: In the decorated function `func`, if there are N positional arguments but N+M (M > 0) are given, this will result in a `TypeError: func takes N positional arguments but N+M were given`, so either we `try` the function and catch the exception reorganizing the arguments and giving a warning about bad usage (could lead to unexpected function result for the user); or we update the tests to catch the `TypeError` when this happens (this is simpler).

dipy/testing/tests/test_decorators.py
```diff
@@ -51,3 +51,36 @@ def f():
...
+    # Test case: No warning expected when all arguments have default values
+    with warnings.catch_warnings(record=True) as w:
+        warnings.simplefilter("always")
+        func_with_default_args(a=1, b=2, c=3)
+        assert len(w) == 0, "Expected no warnings"
+
+    # Test case: Warning expected when keyword-only argument is used positionally
+    with warnings.catch_warnings(record=True) as w:
+        warnings.simplefilter("always")
+        func_with_kwonly_args(1, 2, 3)
```
itellaetxe, line 79: Test does not pass because a `TypeError` is raised. Catch it, or adapt the function as suggested in prior comment.

itellaetxe (commented):
Added some comments about catching `TypeError` when greater number of positional arguments than specified are given.

The CI is also complaining because there are a couple of functions in the tests that would need to be updated with the decorator.

dipy/testing/decorators.py
deka27: Noted. Thanks.

dipy/testing/decorators.py
deka27: I agree. Thanks for the feedback

deka27:
Codespell check issue is not related to this PR.
Also, I will squash all commits once PR is finalized.

skoudoro:
> Codespell check issue is not related to this PR. Also, I will squash all commits once PR is finalized.

ok, Can you check the CI's before reviewing. as you can see, they are all failing

deka27:
Optional Deps check failure is not relevant to this PR, i will check & try to fix that in another PR.

itellaetxe (approved):
Seems like the CI failures are not related to this PR.
LGTM, approving.

dipy/testing/decorators.py
```diff
@@ -103,3 +105,88 @@ def _set_random_number_generator_wrapper(pytestconfig, *args, **kwargs):
...
+                                kwargs[kwarg] = excess_args[i]
+                                positionally_passed_kwonly_args.append(kwarg)
+
+                        if positionally_passed_kwonly_args:
+                            warning_msg = (
+                                f"Warning: Function '{func.__name__}' "
+                                f"keyword-only arguments "
+                                f"{positionally_passed_kwonly_args} are being called "
+                                f"positionally. This will be deprecated from version "
+                                f"'{deprecates_from_version}'."
```
skoudoro, line 173: f"Pass {positionally_passed_kwonly_args} as keyword args. From version "
                    f"{deprecates_until_version} passing these as positional arguments "
                    "will result in an error"

dipy/testing/decorators.py
```diff
@@ -103,3 +105,88 @@ def _set_random_number_generator_wrapper(pytestconfig, *args, **kwargs):
         return _set_random_number_generator_wrapper
 
     return _set_random_number_generator
+
+
+def warn_on_positional_args(deprecates_from_version="1.10.0"):
```
skoudoro, line 110: add `until_version` as parameter also

skoudoro (commented):
Hi @deka27,

See below a review from what we talked previously. What is your current status ?

dipy/testing/decorators.py
```diff
@@ -103,3 +105,88 @@ def _set_random_number_generator_wrapper(pytestconfig, *args, **kwargs):
...
+
+            try:
+                bound_args = sig.bind(*args, **kwargs)
+                bound_args.apply_defaults()
+            except TypeError as e:
+                if "too many positional arguments" in str(e):
+                    num_positional_args = len(args)
+                    num_expected_positional = len(param_names) - len(kwonly_args)
+                    if num_positional_args > num_expected_positional:
+                        # Move excess positional arguments to keyword arguments
```
WassCodeur, line 156: Redundancy: The part where excess arguments are converted into keywords could be simplified.

dipy/testing/decorators.py
```diff
@@ -103,3 +110,111 @@ def _set_random_number_generator_wrapper(pytestconfig, *args, **kwargs):
...
+                max_positional_args = sum(
+                    1
+                    for param in params.values()
+                    if param.kind in
+                    (Parameter.POSITIONAL_ONLY, Parameter.POSITIONAL_OR_KEYWORD)
+                )
+
+                if len(args) > max_positional_args:
+                    # Split positional arguments into valid positional and 
+                    # those needingconversion
```
itellaetxe, line 174: I think that the CodeSpell CI job will give an error here.
`needingconversion` -> `needing conversion`

itellaetxe (approved):
Just a small typo to avoid the CodeSpell complaining.
I really like how you wrote the test cases, the comments helped me understand them.
LGTM, approving

dipy/testing/decorators.py
itellaetxe: Okay, the CodeSpell does not complain, sorry, I did not check the job status.

However, Ruff does complain about some code formatting. Check it when you can please.

Also, there is a problem with the 'Optional Deps' job for MacOS, Python version 3.10. I am not an expert in CI related problems, maybe @skoudoro can give you a hint.

deka27: understood. Thanks

deka27:
my CI is failing because of some other issue not related to this PR itself. I am looking into those.

itellaetxe (approved):
LGTM, approving
GJ Kaustav!

dipy/testing/decorators.py
```diff
@@ -103,3 +109,109 @@ def _set_random_number_generator_wrapper(pytestconfig, *args, **kwargs):
         return _set_random_number_generator_wrapper
 
     return _set_random_number_generator
+
+
+def warning_for_keywords(from_version="1.10.0", until_version="2.0.0"):
```
robinroy03, line 114:
```py
>>> dipy.__version__
'1.10.0.dev0+0.b46f850'
>>> @warning_for_keywords(from_version="1.10.0", until_version="2.0.0")
... def t(a, b, *, c, d):
...     return a+b+c+d
...
>>> t(1, 2, 3, 4)
10
>>> dipy.__version__
'1.10.0.dev0+0.b46f850'
>>> dipy.__version__ = "1.10.0"
>>> t(1, 2, 3, 4)
<stdin>:1: UserWarning: Pass ['c', 'd'] as keyword args. From version 2.0.0 passing these as positional arguments will result in an error.
10
```

Shouldn't the development version throw `warning` here?

dipy/stats/resampling.py
```diff
@@ -157,14 +157,22 @@ def __calc_z0(x, p_0, statistic, eps, a_hat, sigma_hat):
...
+            ((1 - eps) * p_0 - eps * tt_dot / (n**2 * sigma_hat)),
+            statistic=statistic,
+        )
+        - 2 * __tt(x, p_0, statistic=statistic)
     ) / eps**2
     return a_hat - (b_hat / sigma_hat - c_q_hat)
 
 
-def __tt(x, p_0, statistic=bs_se):
+def __tt(x, p_0, *, statistic=bs_se):
```
WassCodeur, line 175: There's no decorator here, is that normal?

WassCodeur, line 175: There's no decorator here, is that normal ?

dipy/testing/decorators.py
```diff
@@ -103,3 +109,109 @@ def _set_random_number_generator_wrapper(pytestconfig, *args, **kwargs):
...
+    """
+    Decorator to warn about keyword arguments passed as positional arguments
+    and handle version-based deprecation of functions.
+
+    Parameters
+    ----------
+    from_version : str
```
WassCodeur, line 121: from_version will be: Optional, str intead of str

dipy/testing/decorators.py
```diff
@@ -103,3 +109,109 @@ def _set_random_number_generator_wrapper(pytestconfig, *args, **kwargs):
...
+        The version from which the warning should start.
+    until_version : str
```
WassCodeur, line 123: until_version will be: Optional, str intead of str

dipy/testing/decorators.py
skoudoro: No @robinroy03, because dev version is before the release. so it looks good.

you can look at [Semantic Versioning](https://semver.org/)

Thank you for testing

dipy/stats/resampling.py
skoudoro: no, needs an update, good catch

skoudoro: same comment as above

dipy/testing/decorators.py
skoudoro: indeed,   `str, optional` instead

dipy/testing/decorators.py
skoudoro: indeed,   `str, optional` instead

dipy/stats/resampling.py
WassCodeur: Okay, Thank you

skoudoro:
Can you also rebase this PR? Thanks

dipy/testing/decorators.py
```diff
@@ -103,3 +109,109 @@ def _set_random_number_generator_wrapper(pytestconfig, *args, **kwargs):
...
+                    for param, arg in zip(
+                        list(params.values())[max_positional_args:],
+                        args[max_positional_args:],
+                    ):
+                        corrected_kwargs[param.name] = arg
+                        positionally_passed_kwonly_args.append(param.name)
+
+                    if positionally_passed_kwonly_args and version.parse(
+                        from_version
+                    ) <= version.parse(current_version) < version.parse(until_version):
```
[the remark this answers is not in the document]
deka27: will give default error.

dipy/testing/decorators.py
```diff
@@ -103,3 +109,109 @@ def _set_random_number_generator_wrapper(pytestconfig, *args, **kwargs):
...
+                version.parse(from_version)
+                <= version.parse(current_version)
+                < version.parse(until_version)
+            ):
+                # Convert positional to keyword arguments and issue a warning
+                return convert_positional_to_keyword(func, args, kwargs)
+
+            # If the version is greater than or equal to the until_version,
+            # pass the arguments as they are
+            elif version.parse(current_version) >= version.parse(until_version):
```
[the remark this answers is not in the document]
deka27: makes sense, i will make it inclusive.

dipy/testing/decorators.py
```diff
@@ -103,3 +109,109 @@ def _set_random_number_generator_wrapper(pytestconfig, *args, **kwargs):
...
+        The wrapped function that will issue a warning based
+        on the version.
+    """
+
+    def decorator(func):
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            current_version = dipy.__version__
+
+            def convert_positional_to_keyword(func, args, kwargs):
```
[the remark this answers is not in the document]
deka27: The `func` parameter in `convert_positional_to_keyword` is not redundant because this function needs to know which function it should call with the modified arguments. While the outer decorator function knows about `func` because it is its parameter, `convert_positional_to_keyword` also needs to know about it, hence it is passed as an argument when calling `convert_positional_to_keyword`.

It also allows the helper function to be generic and reusable for any function that the decorator is applied to.

dipy/stats/resampling.py
deka27: Fixed

skoudoro (approved):
Thank you for this work @deka27!

all good. merging

PR #3245, opened 2024-06-05
RF: Switch from using sparse *_matrix to *_array.

arokem:
See also:
- https://github.com/mne-tools/mne-python/issues/12642
- https://github.com/mne-tools/mne-python/pull/12646

### DIFF
--- a/dipy/core/optimize.py
+++ b/dipy/core/optimize.py
@@ -7,7 +7,6 @@ import numpy as np
 import scipy.optimize as opt
 from scipy.optimize import minimize
 
-from dipy.utils.deprecator import deprecate_with_version
 from dipy.utils.optpkg import optional_package
 
 cvxpy, have_cvxpy, _ = optional_package("cvxpy", min_version="1.4.1")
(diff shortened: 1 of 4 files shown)


dschult:
It looks like you need to make `optimize.spdot` use the `@` operator instead of `*`. And I believe the `view` part of that function should not be needed anymore as the result of `@` is sparse if at least one operand is sparse. But I'm not sure if there are other side effects you desire from the views that I may not be anticipating.

Ideally, The entire `spdot(A, B)` function could just be replaced by `A @ B`. But I would replace it inside the function first and test whether that works before changing each call of `spdot`. :)

Are there other places where you use `*` to mean dot/matmul?  I don't see any others causing test failures, but it might be worth checking unless you know pretty well where `*` is used on sparse. Hopefully it is encapsulated in `spdot` already.

dschult:
Looks like there is now a deprecation warning for line 132 in `segment/tests/test_fss.py`
Use `.toarray()` instead of `.A`

skoudoro (approved):
Thank you for this @arokem.

I think we should deprecate `spdot` function, it is useless now, but this can be done on a new PR. 

All good, merging

PR #3246, opened 2024-06-08
STYLE: Import explicitly `direction.peaks` symbols

jhlegarreta:
Import explicitly `direction.peaks` symbols.

Fixes:
```
dipy/direction/__init__.py:3:1: F403
 `from .peaks import *` used; unable to detect undefined names
```

raised for example in:
https://github.com/dipy/dipy/actions/runs/9045357134/job/24855013530#step:4:281


jhlegarreta:
`Optional Deps / pip / build (3.10, macos-latest, pip)` failure is unrelated.

PR #3250, opened 2024-06-14
DOC: Fix param order

larsoner:
Just fixes an issue where params were listed in the wrong order.

FYI if you haven't considered it, you could use [numpydoc's checks](https://numpydoc.readthedocs.io/en/latest/validation.html#built-in-validation-checks) in your doc build to catch this particular error with something like:
```
numpydoc_validate = True
numpydoc_validation_checks = {"PR03"}
```
But it looks like you use some custom vendored version (?) so you'd need to switch back to the official one.

As a numpydoc maintainer I'm also curious why you vendor your own parser. :thinking:

### DIFF
--- a/dipy/align/_public.py
+++ b/dipy/align/_public.py
@@ -334,14 +334,14 @@ def resample(
         Containing the data for the moving object, or full path to a nifti file
         with the moving data.
 
+    static : array, nifti image or str
+        Containing the data for the static object, or full path to a nifti file
+        with the moving data.
+
     moving_affine : 4x4 array, optional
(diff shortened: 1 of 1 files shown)


skoudoro:
Thank you for this @larsoner.

> As a numpydoc maintainer I'm also curious why you vendor your own parser.

Not sure about the history of this, it is at least 10years old decision, but it is on the TODO list to review the whole doc generation and remove the tons of warning generation.

I suppose, it was due to our command line interface which parse docstring to create command line. It should be easy to improve all of this and use the official version.

PR #3251, opened 2024-06-16
NF: Applying Decorators in Module (Core)

deka27:
## This PR is to apply keyword decorator to functions in core module.

## Main PR:
- #3239

### DIFF
--- a/dipy/core/profile.py
+++ b/dipy/core/profile.py
@@ -3,6 +3,7 @@
 import os
 import subprocess
 
+from dipy.testing.decorators import warning_for_keywords
 from dipy.utils.optpkg import optional_package
 
 cProfile, _, _ = optional_package("cProfile")
@@ -52,7 +53,7 @@ class Profiler:
 
     """
 
-    def __init__(self, call=None, *args):
+    def __init__(self, *args, call=None):
         # Delay import until use of class instance.  We were getting some very
         # odd build-as-we-go errors running tests and documentation otherwise
         import pyximport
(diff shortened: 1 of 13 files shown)


skoudoro:
Can you rebase and fix the format since #3239 is merged ? Thanks!

dipy/core/profile.py
```diff
@@ -52,7 +53,8 @@ class Profiler:
 
     """
 
-    def __init__(self, call=None, *args):
+    @warning_for_keywords()
+    def __init__(self, *args, call=None):
```
skoudoro, line 57: missing`*`.   I suppose, it should be `def __init__(self, *args, *, call=None):`

skoudoro (commented):
Hi @deka27,

Thank you for the update. See below my comment. Thanks!

dipy/core/profile.py
deka27: This will give syntax error, actually. I can't add separator after *args.
Now what will happen is whatever positional args are passed after self, will be stored in *args and will not even pass to call.
1. This will force user to call the keyword args properly, if they want to specifically use the keyword args.
2. If they don't input, we will have default value either-ways.

Also i can't keep `def __init__(self, *, call=None, *args)`, this will give same syntax error and not allowed.

The only other possible solution i see is `def __init__(self, args=(), *, call=None)`

deka27: both `*args` and `arg=()' will create tuple to store the extra pos arguments

skoudoro: Finally, here, you just need to remove the decorators

skoudoro (approved):
ok, all good, merging. thanks @deka27

PR #3259, opened 2024-06-17
NF: Applying Decorators in Module (Sims)

deka27:
## This PR is to apply keyword decorator to functions in sims module.

## Main PR:
- #3239


deka27:
This seems good to go.

PR #3261, opened 2024-06-17
NF: Applying Decorators in Module (Utils)

deka27:
## This PR is to apply keyword decorator to functions in Utils module.

## Main PR:
- #3239


skoudoro (approved):
All good, merging, thanks @deka27!

PR #3263, opened 2024-06-17
NF: Applying Decorators in Module (Workflow)

deka27:
## This PR is to apply keyword decorator to functions in workflow module.

## Main PR:
- #3239

### DIFF
--- a/dipy/workflows/mask.py
+++ b/dipy/workflows/mask.py
@@ -5,6 +5,7 @@ import logging
 import numpy as np
 
 from dipy.io.image import load_nifti, save_nifti
+from dipy.testing.decorators import warning_for_keywords
 from dipy.workflows.workflow import Workflow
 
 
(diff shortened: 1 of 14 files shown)


skoudoro:
Can you rebase and fix the format since https://github.com/dipy/dipy/pull/3239 is merged ? Thanks!

dipy/workflows/mask.py
```diff
@@ -13,6 +14,7 @@ class MaskFlow(Workflow):
     def get_short_name(cls):
         return "mask"
 
+    @warning_for_keywords()
     def run(self, input_files, lb, ub=np.inf, out_dir="", out_mask="mask.nii.gz"):
```
skoudoro, line 18: missing `*`. any reason ?

skoudoro (changes requested):
Hi @deka27, 

Can you check why it is failing? see the comment below also

itellaetxe (commented):
I checked all the changes. All the function definitions that you changed seem to be correctly updated in the argument definition section. OK

On the other hand, we should check why the CIs are failing

Good work

skoudoro (approved):
thank you for this @deka27, merging

PR #3264, opened 2024-06-18
TEST: avoid direct comparison of floating point numbers

mr-c:
We applied this patch in Debian to fix https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1053265


skoudoro (approved):
make sense, thanks for this fix @mr-c, merging

PR #3266, opened 2024-06-18
NF: Call `cnp.import_array()` explicitly to use the NumPy C API

jhlegarreta:
Call `cnp.import_array()` explicitly to use the NumPy C API: Cython modules (or anything using the NumPy C API) must call `cnp.import_array()` to import the NumPy C API.

Fixes:
```
  File "vector_fields.pyx", line 1, in init dipy.align.vector_fields
ImportError: numpy.core.multiarray failed to import
 (auto-generated because you didn't call 'numpy.import_array()' after
 cimporting numpy; use '<void>numpy._import_array' to disable if you
 are certain you don't need it).
```

reported by third-party tool (e.g. MNE) maintaners when testing their tools with NumPy 2.0.

Fixes #3265.

PR #3268, opened 2024-06-21
STYLE: Prefer using `np.asarray` to avoid copy while creating an array

jhlegarreta:
Prefer using `np.asarray` to avoid copy while creating an array in preparation for full NumPy 2.0 compatibility.

Fixes:
```
FAILED core/tests/test_interpolation.py::test_TriLinearInterpolator -
 ValueError: Unable to avoid copy while creating an array as requested.
If using `np.array(obj, copy=False)` replace it with `np.asarray(obj)`
 to allow a copy when needed (no behavior change in NumPy 1.x).
```

raised for example in:
https://github.com/dipy/dipy/actions/runs/9605741685/job/26493943578?pr=3266#step:10:7997

### DIFF
--- a/dipy/core/interpolation.pyx
+++ b/dipy/core/interpolation.pyx
@@ -1126,7 +1126,7 @@ class TriLinearInterpolator(Interpolator):
                              "of a 3d voxel")
 
     def __getitem__(self, index):
-        index = np.array(index, copy=False, dtype="float")
+        index = np.asarray(index, dtype="float")
         try:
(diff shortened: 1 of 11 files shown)


jhlegarreta:
Most builds are failing because they are installing NumPy 2.0.0, so depends on PR #3266. Anyways, this change should be safe, as these types and methods are used elsewhere in the code (`Compat` and `Optional Deps` passing proves this).

skoudoro (approved):
Thanks @jhlegarreta, merging

PR #3269, opened 2024-06-21
STYLE: Prefer using `isin` over `in1d`

jhlegarreta:
Prefer using `isin` over `in1d` in preparation for full NumPy 2.0 compatibility.

Fixes:
```
segment/tests/test_fss.py::test_fss_radius_search
  dipy/venv/lib/python3.12/site-packages/dipy/segment/tests/test_fss.py:96:
 DeprecationWarning: `in1d` is deprecated. Use `np.isin` instead.
    assert_true(np.all(np.in1d(rs_f1_sd.row, rs_f2_in_f1.row)))
```

reported for example in:
https://github.com/dipy/dipy/actions/runs/9605741685/job/26493943578?pr=3266#step:10:10205

### DIFF
--- a/benchmarks/benchmarks/bench_reconst.py
+++ b/benchmarks/benchmarks/bench_reconst.py
@@ -44,7 +44,7 @@ class BenchVecValSum:
 
 #         labels = labels_img.get_fdata()
 #         shape = labels.shape
-#         mask = np.in1d(labels, [1, 2])
+#         mask = np.isin(labels, [1, 2])
 #         mask.shape = shape
 
 #         center = (50, 40, 40)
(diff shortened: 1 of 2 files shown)


jhlegarreta:
Most builds are failing because they are installing NumPy 2.0.0, so depends on PR #3266. Anyways, this change should be safe, as these types and methods are used elsewhere in the code (`Compat` and `Optional Deps` passing proves this).

skoudoro:
>. Most builds are failing because they are installing NumPy 2.0.0

This is what I was talking about @jhlegarreta last time we see each other.  We might need to create specific environment with specific dependencies version to avoid this kind of version issue. Only one matrix needs to have the latest version of each dependencies.

Otherwise, it looks good to me. I am going ahead and merging

jhlegarreta:
> This is what I was talking about @jhlegarreta last time we see each other. We might need to create specific environment with specific dependencies version to avoid this kind of version issue. Only one matrix needs to have the latest version of each dependencies.

Yep, I remember. One thought is to limit the major version of the packages to the ones that are known to work for default builds; then add another set with that version +1, and another set with the latest versions.

PR #3267, opened 2024-06-20
STYLE: Avoid deprecated NumPy types and methods for NumPy 2.0 compat

jhlegarreta:
Avoid deprecated NumPy types and methods in preparation for full NumPy 2.0 compatibility.

Fixes:
```
dipy/align/imwarp.py:1585:13: NPY201 [*]
 `np.row_stack` will be removed in NumPy 2.0.
 Use `numpy.vstack` instead.
dipy/reconst/mcsd.py:411:18: NPY201 [*]
 `np.NaN` will be removed in NumPy 2.0.
 Use `numpy.nan` instead.
dipy/stats/tests/test_qc.py:69:13: NPY201 [*]
 `np.row_stack` will be removed in NumPy 2.0.
 Use `numpy.vstack` instead.
dipy/stats/tests/test_qc.py:70:32: NPY201 [*]
 `np.row_stack` will be removed in NumPy 2.0.
 Use `numpy.vstack` instead.
dipy/tracking/tests/test_tracking.py:1168:13:
 NPY201 [*] `np.row_stack` will be removed in NumPy 2.0.
 Use `numpy.vstack` instead.
doc/examples/linear_fascicle_evaluation.py:165:17:
 NPY201 [*] `np.row_stack` will be removed in NumPy 2.0.
 Use `numpy.vstack` instead.
```

reported when running:
ruff check . --select NPY201

All these features were already available in recent versions of NumPy 1.x, including the minimum required one for DIPY.

Documentation:
https://numpy.org/devdocs/numpy_2_0_migration_guide.html

PR #3271, opened 2024-06-25
RF: Do not use `np.any` for checking optional array parameters

jhlegarreta:
- DOC: Add missing optional label to optional parameter docstring
- RF: Do not use `np.any` for checking optional array parameters


jhlegarreta:
This should fix all remaining issues (see PRs #3266, #3267, #3268, #3269) related to NumPy except for the data size incompatibility, which will probably require pinning/limiting install require and dependencies versions.

jhlegarreta:
Same comment as in https://github.com/dipy/dipy/pull/3268#issuecomment-2181838710.

skoudoro (approved):
Thanks @jhlegarreta, LGTM

merging

PR #3183, opened 2024-04-13
ENH: Specify the solver for the MAP-MRI positivity constraint test

jhlegarreta:
Specify the solver for the MAP-MRI positivity constraint test: the solver picked up by `cvxpy` as the most appropriate one (`OSQP`) finds a solution that it suspects may be inaccurate, so set it to `CLARABEL`.

In either case, the optimal value found is 5.729e-01.

Fixes:
```
reconst/tests/test_mapmri.py::test_positivity_constraint
  /home/runner/work/dipy/dipy/venv/lib/python3.10/site-packages/cvxpy/problems/problem.py:1403:
UserWarning: Solution may be inaccurate.
Try another solver, adjusting the solver settings, or solve with verbose=True for more information.
    warnings.warn(
```

raised for example at:
https://github.com/dipy/dipy/actions/runs/8649253830/job/23715040118#step:10:4368

PR #3274, opened 2024-06-27
RF: Update pyproject.toml for numpy 2.0

skoudoro:
> A module that was compiled using NumPy 1.x cannot be run in
NumPy 2.0.0 as it may crash. To support both 1.x and 2.x
versions of NumPy, modules must be compiled with NumPy 2.0.
Some module may need to rebuild instead e.g. with 'pybind11>=2.12'.

make sure that we always build DIPY with `numpy>=2.0.0rc1`

### DIFF
--- a/dipy/reconst/tests/test_mapmri.py
+++ b/dipy/reconst/tests/test_mapmri.py
@@ -863,16 +863,23 @@ def test_positivity_constraint(radial_order=6, rng=None):
 
     # Set the cvxpy solver to CLARABEL as the one picked otherwise for this
     # problem (OSQP) triggers a `Solution may be inaccurate` UserWarning
-    mapmod_constraint = MapmriModel(
-        gtab,
-        radial_order=radial_order,
-        laplacian_regularization=False,
-        positivity_constraint=True,
-        pos_grid=gridsize,
-        pos_radius="adaptive",
-        cvxpy_solver=mapmri.cvxpy.CLARABEL,
-    )
-    mapfit_constraint = mapmod_constraint.fit(S_noise)
+    with warnings.catch_warnings():
+        warnings.filterwarnings(
+            "ignore",
+            message=".*Solution may be inaccurate.*",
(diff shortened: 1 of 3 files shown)


skoudoro:
ok, it works as expected, just 1-2 warnings to clean and it will be ready to go. 

here the warnings:
```
testing/__init__.py::dipy.testing.clear_and_catch_warnings
  <doctest dipy.testing.clear_and_catch_warnings[2]>:1: DeprecationWarning: numpy.core is deprecated and has been renamed to numpy._core. The numpy._core namespace contains private NumPy internals and its use is discouraged, as NumPy internals can change without warning in any release. In practice, most real-world usage of numpy.core is to access functionality in the public NumPy API. If that is the case, use the public NumPy API. If not, you are using NumPy internals. If you would still like to access an internal attribute, use numpy._core.fromnumeric.
  ```

I might close #3266 in favor of this PR.

skoudoro:
ok, all good. The first warning fixed also. 

I do not succeed to generate the second warning locally ( this warning is triggered only on macOS). 
if I do not succeed to fix it, I will just ignore this warning on `pyproject.toml`

skoudoro:
This PR is ready to go - @arokem and @jhlegarreta.   It finalizes the compatibility with Numpy 2.0 

I have corrected the 2 warnings above but it seems 2 others warnings are there. I will fix them in new PR.

Feel free to merge.  Thanks!

arokem (approved):

skoudoro:
Thanks @arokem !

Note: one of the failing matrix is waiting the the new release of `pytables` because current release does not have support of arm64. see https://github.com/PyTables/PyTables/issues/1165

jhlegarreta:
:+1: The cvxpy warning was present in some builds a while ago, but the following PR solved the issue at that time:
https://github.com/dipy/dipy/pull/3183

Looks like we are missing something else.

skoudoro:
I agree

PR #3170, opened 2024-04-04
Iteratively reweighted least squares for robust fitting

samcoveney:
Adds ability to do iteratively re-weighted least squares for DTI and DKI (with/without positivity constraints) by defining a function to return weights based on the residuals of the last fit.

Implementation is in:
- `iterative_fit_tensor` in `reconst/dti.py`
- `iterative_fit` in `reconst/dki.py`

The methods work by iterating over a standard fit method, such as WLS, NLLS, or CWLS, but between fitting iterations weights are calculated from a supplied function `weights_method`. Some examples are given in `reconst/weights_method.py`, but users can define their own functions outside of DIPY and pass the methods as arguments. Outlier rejection is naturally handled by the `weights_method` supplied.

Example:

```python
# use robust NLLS from dti.py
dkimodel = dki.DiffusionKurtosisModel(gtab, fit_method="RNLLS", return_S0_hat=True, num_iter=10)
dkifit = dkimodel.fit(data, mask=mask)
robust = dkimodel.extra["robust"]  # obtain mask of robust signals

# use robust CWLS from dki.py
from dipy.reconst.weights_method import weights_method_wls_gm as wc
dkimodel = dki.DiffusionKurtosisModel(gtab, fit_method="CWLS", return_S0_hat=True, weights_method=wc, num_iter=10)
dkifit = dkimodel.fit(data, mask=mask)
robust = dkimodel.extra["robust"]  # obtain mask of robust signals
```

Notes:
- It was simpler to implement without a stopping condition on individual voxels
- I re-implemented RESTORE and NLLS slightly, I think they are better now (we could actually implement RESTORE entirely within this new framework)
- By fitting to all voxels between iterations (rather than iterating over individual voxels, before moving on to the next voxel) it is possible to define weights based on multiple voxels. I will include my own weight functions for this, implemented in DiPy already, in a future PR

This is a fairly major extension that I think can be applied to other models using the same framework. I am sharing now to benefit everyone, ahead of my own publications, so please be kind!

Also, this has been a lot of work. Sorry I haven't done:
- tests
- tutorials / examples
- full pep8 compliance

but I would like to get feedback at this stage please

PR #3257, opened 2024-06-16
NF: Applying Decorators in Module (reconst)

deka27:
## This PR is to apply keyword decorator to functions in reconst module.

This PR is dependent on the main PR, which should be merged first.

## Main PR:
- #3239

PR #3286, opened 2024-07-05
BF: Allow passing kwargs in `fit` method, by moving parallelization kwargs elsewhere

arokem:
Trying to address issues raised in https://github.com/dipy/dipy/pull/3170/files#r1666792968

Assuming this is the issue, it still needs some testing to make sure that it works as expected.

### DIFF
--- a/dipy/reconst/multi_voxel.py
+++ b/dipy/reconst/multi_voxel.py
@@ -44,7 +44,8 @@ def multi_voxel_fit(single_voxel_fit):
         # The following are "standard" arguments:
         args.pop(args.index("self"))
         args.pop(args.index("data"))
-        args.pop(args.index("mask"))
+        if "mask" in args:
+            args.pop(args.index("mask"))
         # What remains are arguments that are defined separate from **kwargs
         # (kwargs are specifically reserved only for parallelization
         # arguments):
--- a/dipy/reconst/multi_voxel.py
+++ b/dipy/reconst/multi_voxel.py
@@ -1,7 +1,6 @@
 """Tools to easily make multi voxel models"""
 
 from functools import partial
-from inspect import getargspec
 import multiprocessing
 
 import numpy as np
@@ -36,26 +35,10 @@ def multi_voxel_fit(single_voxel_fit):
 
     def new_fit(self, data, mask=None, **kwargs):
         """Fit method for every voxel in data"""
-        # Analyze if the single voxel fit has any "non standard"
-        # key-word arguments:
-        arg_spec = getargspec(single_voxel_fit)
-        # List of the arguments to the function:
-        args = arg_spec.args
-        # The following are "standard" arguments:
-        args.pop(args.index("self"))
-        args.pop(args.index("data"))
-        args.pop(args.index("mask"))
-        # What remains are arguments that are defined separate from **kwargs
-        # (which we reserve for parallelization arguments):
-        func_kwargs = {}
-        for kwarg in args:
-            func_kwargs.update({kwarg: kwargs[kwarg]})
-            kwargs.pop(kwarg)
-
         # If only one voxel just return a standard fit, passing through
         # the functions key-word arguments:
         if data.ndim == 1:
-            return single_voxel_fit(self, data, **func_kwargs)
+            return single_voxel_fit(self, data, **kwargs)
 
         # Make a mask if mask is None
         if mask is None:
@@ -67,15 +50,14 @@ def multi_voxel_fit(single_voxel_fit):
         # Fit data where mask is True
         fit_array = np.empty(data.shape[:-1], dtype=object)
         # Default to serial execution:
-        engine = kwargs.get("engine", "serial")
+        engine = self.parallel_kwargs.get("engine", "serial")
+        verbose = self.parallel_kwargs.get("verbose", True)
         if engine == "serial":
-            bar = tqdm(
-                total=np.sum(mask), position=0, disable=kwargs.get("verbose", True)
-            )
+            bar = tqdm(total=np.sum(mask), position=0, disable=verbose)
             bar.set_description("Fitting reconstruction model using serial execution")
             for ijk in ndindex(data.shape[:-1]):
                 if mask[ijk]:
-                    fit_array[ijk] = single_voxel_fit(self, data[ijk], **func_kwargs)
+                    fit_array[ijk] = single_voxel_fit(self, data[ijk], **kwargs)
                 bar.update()
             bar.close()
         else:
@@ -95,8 +77,8 @@ def multi_voxel_fit(single_voxel_fit):
                         _parallel_fit_worker,
                         chunks,
                         func_args=[single_voxel_with_self],
-                        func_kwargs=func_kwargs,
-                        **kwargs,
+                        func_kwargs=kwargs,
+                        **self.parallel_kwargs,
                     )
                 )
             )
--- a/dipy/reconst/multi_voxel.py
+++ b/dipy/reconst/multi_voxel.py
@@ -36,20 +36,24 @@ def multi_voxel_fit(single_voxel_fit):
 
     def new_fit(self, data, mask=None, **kwargs):
         """Fit method for every voxel in data"""
-        # Analyze if the single voxel fit has any additiona key-word arguments:
+        # Analyze if the single voxel fit has any "non standard"
+        # key-word arguments:
         arg_spec = getargspec(single_voxel_fit)
         # List of the arguments to the function:
         args = arg_spec.args
+        # The following are "standard" arguments:
         args.pop(args.index("self"))
         args.pop(args.index("data"))
         args.pop(args.index("mask"))
-        # What remains are arguments that are not in **kwargs:
+        # What remains are arguments that are defined separate from **kwargs
+        # (which we reserve for parallelization arguments):
         func_kwargs = {}
         for kwarg in args:
             func_kwargs.update({kwarg: kwargs[kwarg]})
             kwargs.pop(kwarg)
 
-        # If only one voxel just return a normal fit
+        # If only one voxel just return a standard fit, passing through
+        # the functions key-word arguments:
         if data.ndim == 1:
             return single_voxel_fit(self, data, **func_kwargs)
 
--- a/dipy/reconst/multi_voxel.py
+++ b/dipy/reconst/multi_voxel.py
@@ -1,7 +1,7 @@
 """Tools to easily make multi voxel models"""
 
 from functools import partial
-from inspect import getargspec
+from inspect import getfullargspec as getargspec
 import multiprocessing
 
 import numpy as np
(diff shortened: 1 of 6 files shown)


arokem:
OK - now with testing also implemented.

arokem:
This now works locally. Let's see if the CI also works.

arokem:
On a bit of further thought, I don't know if this PR is a good idea, even if I can make it work. I think that we should, as a rule, never pass anything to a `ReconstModel`'s `fit` method except data, mask and parallelization kwargs. I think that I can make this work, but it's awfully brittle, and easy to avoid by passing everything else that we need (i.e., hyperparameters) to the initialization of the model object.

samcoveney:
> On a bit of further thought, I don't know if this PR is a good idea, even if I can make it work. I think that we should, as a rule, never pass anything to a `ReconstModel`'s `fit` method except data, mask and parallelization kwargs. I think that I can make this work, but it's awfully brittle, and easy to avoid by passing everything else that we need (i.e., hyperparameters) to the initialization of the model object.

if we cannot pass additional things to `fit` then there is no way to implement my PR request on iteratively reweighted least squares in https://github.com/dipy/dipy/pull/3170 as I have explained in that PR. If we need to do a fit where the weights can vary but need to be calculated outside of fit, then there is no choice but to pass in different weights for each chunk of data.

One possible solution would be to have weights be a non-optional input to fit, which would avoid the keyword issue, but that seems like a lot of work. However, if you did want DiPy to be able to generalize robust fitting to everything (which was the point of my PR - we could generalize this to all fitting routines), then that could be a way to go. Otherwise, I am surprised that parallelization cannot handle the keywords to be different for each parallel process... (and I was surprised to see that the keywords were not being passed _at all_ in the original problem code...)

samcoveney:
can I also ask: why is this solution here "brittle"?
Does this solution allow for each parallel worker to receive a _different_ value for the same kwarg, as would be required to allow weights to be passed in that differed for each process?

arokem:
Sorry for the break here - I was on vacation and returned yesterday. I actually worked up this recent commit on the ✈️ home! What about the solution implemented in the recent commit? The idea behind this is that on initialization of a `ReconstModel` and derived classes, the user can pass instructions about how parallelization should proceed. This frees up the possibility to pass any key-word arguments that we want to the `fit` method. I think it's a bit more robust than the solution I had previously proposed. But interested to hear if folks have any concerns. This would have to be implemented also at the level of derived classes, so I am setting this PR to WIP for now.

samcoveney:
Thanks for the work.

I'm afraid I can't tell from the PR easily, whether this makes it possible to pass a different keyword to every parallel process. Can you tell whether this is going to work with my PR #3170 , in which every chunk of data to process is going to receive a `weights` keyword that consist of a different chunk of an array (in much that same way as data is divided up into chunks, but of course `data` is not a keyword).

Thanks

skoudoro:
After looking at both PR, I am sure that both of your solution work @arokem with @samcoveney PR. You just need to make sure to pass kwargs `everywhere`

Actually, I prefer your first solution than this one. it was easier to understand and straight forward. it does not conflict wit Sam work. The new solution is a bit hidden to me. I started to fix your PR with all models and I see that it will become quickly a mess. 

When you make sure that `kwargs` is added and updated everywhere, @samcoveney will have to just update `kwargs` here: https://github.com/dipy/dipy/blob/71d9e7f69adb55567b1f925b3771bfb9196e2e61/dipy/reconst/multi_voxel.py#L80  which he is already doing when you look at his PR : https://github.com/dipy/dipy/blob/6fb312aac3d6d5350248af6c96c07692d808baeb/dipy/reconst/multi_voxel.py#L77-L86

So, I would recommend to revert your last commit, make sure that `kwargs` is passed correctly everywhere. Then we can merge and help @samcoveney for its rebase.

I might miss something but it looks quite straight forward

arokem:
Thanks for taking a look @skoudoro! Just making sure that I understand your comment: do you prefer the solution that was implemented in this commit? https://github.com/dipy/dipy/pull/3286/commits/4e316474d32bf682a4f649ac025930e6eafc7a63. 

That is, you'd rather I fiddle with the args/kwargs within the decorator and then pass things through? If you think that's more straightforward, happy to revert to that and continue to make everything work harmoniously.

skoudoro:
> do you prefer the solution that was implemented in this commit? https://github.com/dipy/dipy/commit/4e316474d32bf682a4f649ac025930e6eafc7a63

Yes !

arokem:
I've reverted the commit that moved parallelization setting to the init and restored the use of getargspec to manipulate inputs to the decorator. I wonder if we want to do something to verify that implementations of a `fit` method always comply with this "spec", so that future implementations continue to be compliant with this.

skoudoro:
>  I wonder if we want to do something to verify that implementations of a fit method always comply with this "spec", so that future implementations continue to be compliant with this.

Not sure to understand, can you be more explicit?

dipy/reconst/multi_voxel.py
```diff
@@ -35,9 +36,26 @@ def multi_voxel_fit(single_voxel_fit):
...
+        func_kwargs = {}
+        for kwarg in args:
+            func_kwargs.update({kwarg: kwargs[kwarg]})
+            kwargs.pop(kwarg)
+
+        # If only one voxel just return a standard fit, passing through
+        # the functions key-word arguments:
         if data.ndim == 1:
-            return single_voxel_fit(self, data)
+            return single_voxel_fit(self, data, **func_kwargs)
```
skoudoro, line 58: The mask seems to be missing here

dipy/reconst/multi_voxel.py
```diff
@@ -57,7 +75,7 @@ def new_fit(self, data, mask=None, **kwargs):
             bar.set_description("Fitting reconstruction model using serial execution")
             for ijk in ndindex(data.shape[:-1]):
                 if mask[ijk]:
-                    fit_array[ijk] = single_voxel_fit(self, data[ijk])
+                    fit_array[ijk] = single_voxel_fit(self, data[ijk], **func_kwargs)
```
skoudoro, line 78: same as above

dipy/reconst/multi_voxel.py
```diff
@@ -77,6 +95,7 @@ def new_fit(self, data, mask=None, **kwargs):
                         _parallel_fit_worker,
                         chunks,
                         func_args=[single_voxel_with_self],
+                        func_kwargs=func_kwargs,
                         **kwargs,
```
skoudoro, line 99: I think the problem come from here.

Each chunk has its own `kwargs`  **OR** Each chunk has its common kwargs.  

- `paramap` should check if there is a list of `kwargs` or just one `kwargs` for all chuncks
- Currently, `paramap` do not handle the case when each chunk have its own `kwargs`. Without that, Sam PR can not work. it was possible before with the multivoxel fit.

skoudoro (commented):
Hi @arokem,

I did not check why the PR is failing but see below some important comments. Tell me what you think.

skoudoro:
Hi @arokem, 

Do you have an ETA for this? I know that you are busy so let me know if I can help to move forward.

arokem:
I will try to get to it this week, but NeuroHackademy hackathon is also taking place this week, so it's particularly busy. If not this week, I should be able to get to it by end of next week.

samcoveney:
Much thanks for the work on this so far @arokem and @skoudoro

dipy/reconst/multi_voxel.py
arokem: The single voxel fit version doesn't need a mask. I'll add that in a comment.

dipy/reconst/multi_voxel.py
arokem: Same here.

dipy/reconst/multi_voxel.py
arokem: OK - thanks for this. Let me refactor paramap to accomodate sequences of kwargs as well.

arokem: Implemented in 934b3818240015bb3cb54ab2db9c75433eaed84d

arokem: Comment on this: the current implementation supports sequences of func_kwargs, but not sequences of func_args, which need to be identical across calls. Also, the number of items in func_kwargs needs to be identical to the number of items in the data sequence (i.e., one set of func_kwargs per voxel). Does that still work for what you need in your PR, @samcoveney? Adding sequence support for func_args is possible and we could come up with ways to support more complicated combinations (e.g., sets of func_kwargs to map to particular indices in the data), but that will complexify the implementation.

skoudoro: > the current implementation supports sequences of func_kwargs, but not sequences of func_args

make sense to me and we should keep it like that. 

> Also, the number of items in func_kwargs needs to be identical to the number of items in the data sequence (i.e., one set of func_kwargs per voxel).

make sense also and I suppose you meant one set of func_kwargs per chunck and not voxel

> Does that still work for what you need in your PR, @samcoveney? 

I think it is good like this but let me try @arokem. Thank you for updating this. I will review it tomorrow or thursday.

> Adding sequence support for func_args is possible

we should avoid that

arokem:
Failures seem unrelated. Looks like something about some of our downloaded data that it isn't validating properly on Windows: https://github.com/dipy/dipy/actions/runs/10378863121/job/28735933908#step:11:4075

skoudoro:
> Looks like something about some of our downloaded data that it isn't validating properly

yes, recently, we have this issue by intermittence. I do not know why... yet....

dipy/reconst/multi_voxel.py
skoudoro: I got confused, it makes sense

dipy/reconst/multi_voxel.py
```diff
@@ -35,9 +36,29 @@ def multi_voxel_fit(single_voxel_fit):
...
+        if "mask" in this_args:
+            this_args.pop(this_args.index("mask"))
+        # What remains are arguments that are defined separate from **kwargs
+        # (kwargs are specifically reserved only for parallelization
+        # arguments):
+        func_kwargs = {}
+        for kwarg in this_args:
+            if kwarg in kwargs:
+                func_kwargs.update({kwarg: kwargs[kwarg]})
+                kwargs.pop(kwarg)
```
skoudoro, line 56: We are switching the code bad to keyword-only arguments ( [PEP 3102 ](https://peps.python.org/pep-3102/)).

Many PRs has been merged and the reconst module will be merge soon. https://github.com/dipy/dipy/pull/3257

I believe this code will not work with keyword only arguments since they are not in `arg_spec.args`.

I need to test it, but can you handle this case (which will become the general case)?

skoudoro (commented):
Hi @arokem,

Thank you very much for this work. Overall , it looks good and should work with sam PR.

I just have a small comment below that needs to be addressed and it will be good to go.

Thank you

arokem:
Maybe best to make a new pr rebasing on top of #3257? I’ll do that later
today.

dipy/reconst/multi_voxel.py
samcoveney: Yes, I would need one set of func_kwargs per chunk, thanks

skoudoro:
superseded by #3323, closing

PR #3287, opened 2024-07-06
NF: Add `__len__` to `GradientTable`

### LINKED ISSUE #3282: Define ``__len__`` within ``GradientTable``?
It would be very useful to implement ``__len__`` on the ``GradientTable`` object so that bvals do not need to be accessed explicitly:

```
def __len__(self):
    return len(self.bvals)
```

That way the length can be easily checked:

```
>>> len(gtab)
16
```

jhlegarreta:
Add the `__len__` method to `GradientTable`: adds the ability to query the number of gradients using `len()`.

Add the corresponding tests.

Add a doctest example.

Demonstrate it in the `Getting started with DIPY` `Quick Start` documentation section.

Resolves #3282.

### DIFF
--- a/dipy/core/gradients.py
+++ b/dipy/core/gradients.py
@@ -341,6 +341,19 @@ class GradientTable:
         msg += f"          max {self.bvecs.max():f}\n"
         return msg
 
+    def __len__(self):
+        """Get the number of gradients. Includes the b0s.
+
+        Examples
+        --------
+        >>> bvals = np.array([1000, 1000, 1000, 1000, 2000, 2000, 2000, 2000, 0])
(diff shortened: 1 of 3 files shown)


arokem:
LGTM! Does this need any additional documentation? Is there any example where this could be smoothly integrated? Maybe in the quickstart?

jhlegarreta:
> LGTM! Does this need any additional documentation? Is there any example where this could be smoothly integrated? Maybe in the quickstart?

Done in 4d137c1.

skoudoro (approved):
LGTM! thanks @jhlegarreta, merging

PR #3295, opened 2024-07-17
ENH: Add a GHA workflow file to build docs

jhlegarreta:
Add a GHA workflow file to build the documentation (excluding the examples).

### DIFF
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -1,0 +1,39 @@
+name: Build docs
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        shell: bash
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ['3.10']
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v5
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install Dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install .[extra]
+        pip install .[doc]
+    - name: Build documentation
+      run: |
+        cd doc
+        make -C . html-no-examples


jhlegarreta:
@skoudoro You may want to check whether the deployment section is appropriate/accurate/enough. Otherwise, I can remove it from this PR and only leave the documentation build, which is already a step forward to openly see any potential warning/error.

Also,
- Deploying the changes introduced by a PR before it gets merged maybe useful to check the changes any PR introduces to the documentation prior to merging it.
- We may want to make the artefacts available as well in the future.

Best would be to do these in separate PRs.

jhlegarreta:
A number of warnings there:
https://github.com/dipy/dipy/actions/runs/9982200633/job/27587486843?pr=3295#step:5:2618

Looks like it did not continue to the deployment step. Not sure if the warnings are a show stopper for that, as the build of the documentation itself is marked as successful. They are not in FURY's builds at least.

.github/workflows/build_docs.yml
```diff
@@ -0,0 +1,56 @@
...
+        pip install .[doc]
+    - name: Build documentation
+      run: |
+        cd doc
+        make -C . html-no-examples
+    - name: Deploy Devs
+      if: success() && github.ref == 'refs/heads/master'
+      uses: JamesIves/github-pages-deploy-action@v4
+      with:
+        ssh-key: ${{ secrets.ACTIONS_DEPLOY_KEY }}
```
skoudoro, line 44: Did you have added this secret key ?

.github/workflows/build_docs.yml
```diff
@@ -0,0 +1,56 @@
...
+        repository-name: dipy/preview-html
+        folder: ./doc/_build/html
+        target-folder: dev
```
skoudoro, line 47: every PR or merge on master will override the last documentation build. is it the intended goal ?

.github/workflows/build_docs.yml
```diff
@@ -0,0 +1,56 @@
...
+    - name: Update last release
+      if: success() && github.ref == 'refs/heads/master'
```
skoudoro, line 49: this seems to be the same as above so no difference between dev and release. I think I misunderstood something.

skoudoro (commented):
Hi @jhlegarreta, Thank you for this. 

Indeed, this is needed, then we need to go down to zero warning and last step will be to include example generation.

> ou may want to check whether the deployment section is appropriate/accurate/enough

Sorry, I do not understand what you mean

> Deploying the changes introduced by a PR before it gets merged maybe useful to check the changes any PR introduces to the documentation prior to merging it.

I agree

> We may want to make the artefacts available as well in the future.

I do not think it is necessary. I think it will be enough to visualize it online.

> Looks like it did not continue to the deployment step. Not sure if the warnings are a show stopper for that, as the build of the documentation itself is marked as successful.

Currently, warnings are NOT a stopper. This is something to modify when we have fix all warnings.

your github action allow deployment only if merge on master as you can see in your if condition. So the behavior is normal apart if it was not the intended goal.

.github/workflows/build_docs.yml
jhlegarreta: I did not. I am not all that familiar with the keys; I assume you can do that if we want the deployment to happen.

.github/workflows/build_docs.yml
jhlegarreta: IMO, it should be. There is a `dev` API documentation version:
https://docs.dipy.org/dev/reference/index.html

So I assume that after every PR and as long as the doc build is successful, we want the `dev` version to be updated, otherwise, unless a manual deployment is done, the `dev` version will be the same as the last release, right?

Now, one thing to consider is that the examples are not being checked in this action. So maybe the deployment should be deferred to a later PR once we figure out whether there is a way to build the documentation with the examples within a reasonable time.

.github/workflows/build_docs.yml
jhlegarreta: The target folder here is the one corresponding to the last release. It is hard-coded, though:
https://github.com/dipy/dipy/pull/3295/files#diff-4dfe935b4852393622c2b414783d1d11e1a868fd8485709124d90208fa61ec8bR55

However, I now see what you mean, and I agree, it does not make sense here; a separate action should take care of the deployment when a release event is triggered.

jhlegarreta:
> Sorry, I do not understand what you mean

So, do we want the website documentation to be updated through the action after every PR? This is linked to

> your github action allow deployment only if merge on master as you can see in your if condition. So the behavior is normal apart if it was not the intended goal.

But I now see that following https://github.com/dipy/dipy/pull/3295#discussion_r1701031737 given that examples are not build, we should leave the deployment out of this PR. Do you agree @skoudoro?

jhlegarreta:
Push forced, removed the deployment. Will be done in a separate PR. This is already a step forward as is.

skoudoro (approved):
Agreed and it looks good. Merging.

Next steps are:
- Fixing all warnings from doc generation.
- automatic deployment for PR and master
- build with examples for PR and master

PR #3300, opened 2024-07-29
BF: Title Fix

deka27:
In line with the fix in `dipy.org` repo.

https://github.com/dipy/dipy.org/issues/54

Changing the title here too.


deka27:
@skoudoro Hi Serge, same changes here. As we discussed.

PR #3308, opened 2024-08-10
DOC: Rework the BibTeX bibliography file

jhlegarreta:
Rework the BibTeX bibliography file:
- Remove existing references after discussion with other maintainers: the existing bibliography file was created with an bibliography collection tool (Pybliographer), and it will be hard to use it and maintain it as the source of the references for the toolkit: even if the collection is shared somewhere, every time a change is needed, the file should be exported, and replace the existing file. The current formatting is hard to read, references are not sorted, they contain a lot of unnecessary fields, contain non-ASCII characters, and some references are not used in the toolkit. Thus, it is agreed to remove them and to enter references manually following some conventions.
- Add references manually using a set of fields that provide complete information, without adding unnecessary or non-standard keys.
- Fix some references that had inaccurate information (venue, year, etc.).
- Rename the file to the more meaningful `references.bib`.
- Add a section to the developer's guide explaining how the file is organized/maintained.


skoudoro (approved):
Thank you again for this huge work @jhlegarreta! very valuable. 

Let's link all the reference to this now!

PR #3310, opened 2024-08-11
DOC: Cite `nn` references through `sphinxcontrib-bibtex`

jhlegarreta:
- DOC: Cite `nn` references through `sphinxcontrib-bibtex`
- DOC: Cite `theory` references through `sphinxcontrib-bibtex`

### DIFF
--- a/dipy/nn/histo_resdnn.py
+++ b/dipy/nn/histo_resdnn.py
@@ -37,6 +37,17 @@ logger = logging.getLogger("histo_resdnn")
 class HistoResDNN:
     """
     This class is intended for the ResDNN Histology Network model.
+
+    ResDNN :footcite:p:`Nath2019` is a deep neural network that employs residual
+    blocks deep neural network to predict ground truth SH coefficients from SH
+    coefficients computed using DWI data. To this end, authors considered
+    histology FOD-computed SH coefficients (obtained from ex vivo non-human
+    primate acquisitions) as their ground truth, and the DWI-computed SH
+    coefficients as their target.
+
+    References
+    ----------
+    .. footbibliography::
     """
 
     @deprecated_params("sh_order", "sh_order_max", since="1.9", until="2.0")
@@ -45,7 +56,7 @@ class HistoResDNN:
     def __init__(self, *, sh_order_max=8, basis_type="tournier07", verbose=False):
         r"""
         The model was re-trained for usage with a different basis function
-        ('tournier07') like the proposed model in [1]_, [2]_.
+        ('tournier07') like the proposed model in :footcite:p:`Nath2019`.
 
         To obtain the pre-trained model, use::
         >>> resdnn_model = HistoResDNN() # skip if not have_tf
@@ -72,16 +83,7 @@ class HistoResDNN:
 
         References
         ----------
-        ..  [1] Nath, V., Schilling, K. G., Parvathaneni, P., Hansen,
-            C. B., Hainline, A. E., Huo, Y., ... & Stepniewska, I. (2019).
-            Deep learning reveals untapped information for local white-matter
-            fiber reconstruction in diffusion-weighted MRI.
-            Magnetic resonance imaging, 62, 220-227.
-        ..  [2] Nath, V., Schilling, K. G., Hansen, C. B., Parvathaneni,
-            P., Hainline, A. E., Bermudez, C., ... & Stępniewska, I. (2019).
-            Deep learning captures more accurate diffusion fiber orientations
-            distributions than constrained spherical deconvolution.
-            arXiv preprint arXiv:1911.07927.
+        .. footbibliography::
         """  # noqa: E501
 
         if not have_tf:
--- a/dipy/workflows/nn.py
+++ b/dipy/workflows/nn.py
@@ -23,7 +23,7 @@ class EVACPlusFlow(Workflow):
         out_mask="brain_mask.nii.gz",
         out_masked="dwi_masked.nii.gz",
     ):
-        """Extract brain using EVAC+ [Park2022]_.
+        """Extract brain using EVAC+ :footcite:p:`Park2024`.
 
         Parameters
         ----------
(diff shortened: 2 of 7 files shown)


skoudoro:
#2810 has been merged. Can you rebase this one to be able to test it locally. Thanks !

dipy/nn/histo_resdnn.py
```diff
@@ -69,19 +74,6 @@ def __init__(self, *, sh_order_max=8, basis_type="tournier07", verbose=False):
...
-        ..  [1] Nath, V., Schilling, K. G., Parvathaneni, P., Hansen,
-            C. B., Hainline, A. E., Huo, Y., ... & Stepniewska, I. (2019).
-            Deep learning reveals untapped information for local white-matter
-            fiber reconstruction in diffusion-weighted MRI.
-            Magnetic resonance imaging, 62, 220-227.
-        ..  [2] Nath, V., Schilling, K. G., Hansen, C. B., Parvathaneni,
-            P., Hainline, A. E., Bermudez, C., ... & Stępniewska, I. (2019).
-            Deep learning captures more accurate diffusion fiber orientations
-            distributions than constrained spherical deconvolution.
-            arXiv preprint arXiv:1911.07927.
```
skoudoro, line 84: I did not check but it seems 2 differents papers. why you did not use `Nath2019a` and b?

jhlegarreta:
Getting some errors in here
```
/home/runner/work/dipy/dipy/doc/reference/dipy.nn.rst:36:<autosummary>:1: ERROR: Too many autonumbered footnote references: only 0 corresponding footnotes available.
/home/runner/work/dipy/dipy/doc/reference/dipy.nn.rst:36:<autosummary>:1: ERROR: Unknown target name: "footcite-kanakaraj2024".
/home/runner/work/dipy/dipy/doc/reference/dipy.nn.rst:52:<autosummary>:1: ERROR: Unknown target name: "footcite-park2024".
```

Not sure what's missing here. Have you encountered this in your local tests @skoudoro ?

dipy/nn/histo_resdnn.py
jhlegarreta: arXiv papers are not reviewed, so I looked for the equivalent peer-reviewed version and added that:
https://github.com/dipy/dipy/blob/master/doc/references.bib#L1807

skoudoro:
Never encountered this error. But I never added citation in the class docstring. I can reproduce it now locally.  Not sure what is going on.

it looks like class docstring is not parsed correctly.  I will need to dig later this week to understand.

For now, I do not have a solution

jhlegarreta:
>  But I never added citation in the class docstring

Good point. When I did my local attempt a few days ago and reported it in https://github.com/dipy/dipy/pull/2810#issuecomment-2276838712, I tried with the `EVAC` class citation and I got this:



So not sure what is happening; also, every time I build the documentation, and despite cleaning the build directory, Sphinx somehow catches old files, so unless I re-create my virtual env I am not being able to check locally the refs.

The `theory` file is not triggering warnings (it did not before, either; have not checked the result locally).

jhlegarreta:
OK, last push force using `cite` and `bibliography` for the citation in `EVAC` dooes not raise warnings/errors and thus I suspect would give the result I got in https://github.com/dipy/dipy/pull/3310#issuecomment-2287405183. However, as discussed in https://github.com/dipy/dipy/pull/2810#issuecomment-2278382956, using `cite`/`bibliography` apparently requires a single call.

skoudoro:
When I look at this: https://github.com/mne-tools/mne-python/blob/ac863e63e6233dd3ddb60e1058731f1839e55b02/mne/decoding/ssd.py#L29-L36

it seems to work. So I suppose, `footcite` can not be in the short description. it can be in the other part of the docstring (long description, notes, parameters, etc...

This need to be confirmed

skoudoro:
I confirm that it seems to work when the references are not in the short description. The warning also disappear. So, for now, you should add a "long" description if you want to add the citation. See here:

image

skoudoro:
This is confirmed from `sphinxcontrib-bibtex` documentation:

> Numpydoc will sometimes duplicate the short description (i.e. the first line of the docstring) of some python objects such as member functions. If it does that, and you have a footnote citation in the short description, Sphinx may not be able to properly resolve the footnote target. If this happens, the workaround is not to have footnote citations in the first line of your docstrings. Instead, put them in the long description. Alternatively, set numpydoc_class_members_toctree to False in your conf.py file. This will cause numpydoc not to duplicate the short descriptions for class members.

ref: https://sphinxcontrib-bibtex.readthedocs.io/en/latest/usage.html#unknown-target-name-when-using-footnote-citations-with-numpydoc

jhlegarreta:
Build failures are unrelated.

dipy/nn/histo_resdnn.py
```diff
@@ -69,19 +80,6 @@ def __init__(self, *, sh_order_max=8, basis_type="tournier07", verbose=False):
...
```
skoudoro, line 84:
```
    References
    ----------
    .. footbibliography::
 ```

is needed here

File: dipy/workflows/nn.py [hunk 3]
```diff
@@ -38,13 +38,6 @@ def run(
             Name of the mask volume to be saved.
         out_masked : string, optional
             Name of the masked volume to be saved.
-
-        References
-        ----------
-        ..  [Park2022] Park, J.S., Fadnavis, S., & Garyfallidis, E. (2022).
-            EVAC+: Multi-scale V-net with Deep Feature CRF Layers for Brain
-            Extraction.
```
skoudoro, line 46:
```   
    References
    ----------
    .. footbibliography::
```

is needed here

skoudoro (commented):
Hi @jhlegarreta,

Thank you for the update see below some comments, it is almost ready to go.

Small side Note, we need to change the way we parse the docstring for the command line interface. Documentation OK but CLI help incorrect with this new method.

dipy/nn/histo_resdnn.py
jhlegarreta: Sorry, my bad, fixed in c98c3d7.

dipy/workflows/nn.py
jhlegarreta: Sorry, my bad, fixed in c98c3d7.

skoudoro (approved):
All good, LGTM. merging

PR #3312, opened 2024-08-12
DOC: Use `misc` for other types of BibTeX entries

jhlegarreta:
Use `misc` for other types of BibTeX entries: `dataset` and `electronic` entries are supported by `biblatex` but not by BibTeX, so use `misc` for such types.


skoudoro (approved):
Thank you for this update. merging

PR #3321, opened 2024-08-17
BF: Set the superclass `fit_method` param value to the one provided

jhlegarreta:
Set the superclass initialization method `fit_method` parameter value to the one provided to the child's initialization method.


dipy/reconst/dki_micro.py
```diff
@@ -349,7 +349,9 @@ def __init__(self, gtab, fit_method="WLS", *args, **kwargs):
                Characterization with Diffusion Kurtosis Imaging. Neuroimage
                58(1): 177-188. doi:10.1016/j.neuroimage.2011.06.006
         """
-        DiffusionKurtosisModel.__init__(self, gtab, fit_method="WLS", *args, **kwargs)
+        DiffusionKurtosisModel.__init__(
+            self, gtab, fit_method=fit_method, *args, **kwargs
```
skoudoro, line 353: Interesting that we never catch this issue before. We might need to review the tests. Can you update the tests to catch this? Thank you !

jhlegarreta: This might be more complicate that it looks; `fit_method` is already an ivar (should have had a leading underscore to avoid the situation), and thus we cannot add a `fit_method` `@property`. Testing each fit method and checking that they give different results (?) does not look like the best option. I cannot spare time to rework all this unless you see a very easy solution @skoudoro, so the path forward IMO would be to merge this and kindly ask you to open an issue to keep track of this so that it can be addressed whenever somebody has time.

skoudoro: here is an example of an easy test to add in `test_dki_micro.py`:

```python
def test_fit_selection():
    from dipy.reconst.dki import common_fit_methods
    for model in [dki_micro.KurtosisMicrostructureModel,
                  dki_micro.DiffusionKurtosisModel]:
        for name, method in common_fit_methods.items():
            model_instance = model(gtab_2s, fit_method=name)

            assert model_instance.fit_method == method
```

Can you add the test, then I can go ahead and merge. Thank you

jhlegarreta: :+1: Done in 1ab5dfe.

skoudoro: Thanks, all good, just ruff format to update. see here: https://github.com/dipy/dipy/actions/runs/10586658198/job/29335884633?pr=3321#step:4:128

my bad

jhlegarreta: :+1: Done.

skoudoro: I was going to merge but code format still complaining https://github.com/dipy/dipy/actions/runs/10587077477/job/29337117326?pr=3321#step:4:126

jhlegarreta: My bad. Fixed.

dipy/reconst/tests/test_dki_micro.py
```diff
@@ -91,6 +92,17 @@ def teardown_module():
     FIE, RDI, ADI, ADE, Tor, RDE = None, None, None, None, None, None
 
 
+def test_fit_selection():
```
skoudoro, line 95: Sorry, it is missing 

```python
# below the import
needs_cvxpy = pytest.mark.skipif(not have_cvxpy, reason="Requires CVXPY")

# then
@needs_cvxpy
def test_fit_selection():
```

skoudoro (approved):
Thank you for this and adding the tests. all good, merging

PR #3322, opened 2024-08-19
NF: DIPY_DOCKER

deka27:
# Add Dockerfile and Build/Push Scripts

## Description
This PR introduces a Dockerfile for our project and adds two shell scripts to simplify the Docker image building and pushing process.

## Changes
1. Added a `Dockerfile` to define our project's Docker image.
2. Created `docker-build.sh` script to automate the Docker image building process.
3. Created `docker-push.sh` script to automate the Docker image pushing process to a registry.

## Dockerfile Details
- Base image: `python:3.10`
- Installs system dependencies: `libhdf5-dev` and `gcc`
- Installs Python packages: 
  - numpy==1.26.4
  - dipy
  - nibabel
  - scipy
  - matplotlib

### docker-build.sh
- Builds the Docker image with a specified name and tag.
- Provides feedback on build success or failure.

### docker-push.sh
- Tags the built image for the specified registry.
- Pushes the image to the registry.
- Provides feedback on push success or failure.

## Checklist
- [X] Dockerfile builds successfully
- [X] docker-build.sh script runs without errors
- [ ] docker-push.sh script runs without errors (requires registry credentials) - Need to confirm with Serge
- [ ] Resulting Docker image runs and contains all required dependencies (Runs good locally at the moment)
- [ ] Create Tutorial for docker

## Notes for Reviewers
- Please check if the Dockerfile includes all necessary dependencies for our project.
- Verify if the Python package versions are appropriate and up-to-date.
- Suggest any additional error handling or features for the shell scripts.

### DIFF
--- a/dipy-docker/Dockerfile
+++ b/dipy-docker/Dockerfile
@@ -1,0 +1,19 @@
+# Use an official Python runtime as a parent image
+# We are not using slim due to hdf5 dependency
+FROM python:3.10
+
+# Install system dependencies and clean up in one RUN command
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    libhdf5-dev \
+    gcc \
+    && rm -rf /var/lib/apt/lists/* \
+    && apt-get clean
+
+# Install python dependencies
+# We are installing numpy < 2 because some dependencies are complied with 1.x.x and they aren't working with 2.x.x
+RUN pip install --no-cache-dir \
+    numpy==1.26.4 \
+    dipy \
+    nibabel \
+    scipy \
+    matplotlib
--- a/dipy-docker/docker-build.sh
+++ b/dipy-docker/docker-build.sh
@@ -1,0 +1,14 @@
+# Set variables
+IMAGE_NAME="dipy" # Change after confirmation from Serge
+TAG="latest"
+
+# Build the Docker image
+docker build -t ${IMAGE_NAME}:${TAG} .
+
+# Check if the build was successful
+if [ $? -eq 0 ]; then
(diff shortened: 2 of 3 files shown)


arokem:
How do you anticipate the docker image would be used? Would it make sense to install all of the optional dependencies, or is this intentionally kept to the minimal dependencies?

deka27:
> How do you anticipate the docker image would be used? Would it make sense to install all of the optional dependencies, or is this intentionally kept to the minimal dependencies?

The Docker image was created to enable users to work with dipy easily, providing a ready-to-use environment where they can run dipy commands and tools directly from the shell. The decision to go with a minimal setup was based on the tutorials and start guides available for dipy, which typically focus on these core dependencies. I felt that this setup would be sufficient to get users up and running with dipy efficiently. By keeping the image minimal, it ensures quicker builds and a leaner environment, which should be more than adequate for most introductory and standard use cases.

However, I understand that some users might require additional optional dependencies for extended functionality. If there's a need to support a broader range of use cases, we could either:

1. Provide a secondary Dockerfile that includes the optional dependencies, allowing users to choose between a minimal and a full-featured image.
2. Offer guidance on how users can extend the provided Dockerfile to install additional packages based on their specific needs.

Maybe @skoudoro can add more to it.

arokem:
I like the idea of two different images: `dipy-minimal` and `dipy-full` or something like that.

deka27:
> I like the idea of two different images: `dipy-minimal` and `dipy-full` or something like that.

Hmm. I agree as well.

PR #3329, opened 2024-08-20
STYLE: Use a leading underscore to name private methods

jhlegarreta:
Use a leading underscore to name private methods: using a trailing underscore is translated by Sphinx into the same method name as its underscoreless counterpart.

Fixes:
```
/dipy/dipy/doc/reference/dipy.workflows.rst:651:
 ERROR: Duplicate target name, cannot be used as a unique reference: "io_iterator".
```

and similar errors raised for example in:
https://github.com/dipy/dipy/actions/runs/10472634043/job/29002561651#step:5:1146

Documentation:
https://peps.python.org/pep-0008/#method-names-and-instance-variables

### DIFF
--- a/dipy/reconst/qtdmri.py
+++ b/dipy/reconst/qtdmri.py
@@ -287,7 +287,7 @@ class QtdmriModel(Cache):
                 ut /= tau_scaling
                 us = np.clip(us, self.eigenvalue_threshold, np.inf)
                 q = np.dot(bvecs, R) * qvals[:, None]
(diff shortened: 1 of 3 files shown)


jhlegarreta:
I assume they were intended to be private and that this is not an API change.

skoudoro (approved):
> I assume they were intended to be private and that this is not an API change.

same assumption and it looks good to me so I am going ahead and merge this PR

Thanks @jhlegarreta!

PR #3332, opened 2024-08-20
BF: Set the `Diso` parameter value to the one provided 

jhlegarreta:
Set the `Diso` parameter value to the one provided in `wls_fit_tensor`.

### DIFF
--- a/dipy/reconst/fwdti.py
+++ b/dipy/reconst/fwdti.py
@@ -420,7 +420,7 @@ def wls_fit_tensor(
                 data[v],
                 S0[v],
                 min_signal=min_signal,
-                Diso=3e-3,
+                Diso=Diso,
                 piterations=piterations,
                 mdreg=mdreg,
             )


jhlegarreta:
Have not investigated why that was there.

samcoveney:
This might be there reasons related to instability of the numerics - I have found that constrained DKI fails if you don't scale the units, and this might have been a setup to fixing that problem more generally, but I am not sure...

dipy/reconst/fwdti.py
```diff
@@ -344,7 +344,7 @@ def wls_iter(
 
 
 def wls_fit_tensor(
-    gtab, data, Diso=3e-3, mask=None, min_signal=1.0e-6, piterations=3, mdreg=2.7e-3
+    gtab, data, mask=None, min_signal=1.0e-6, piterations=3, mdreg=2.7e-3
```
skoudoro, line 347: Hi @jhlegarreta,

Diso should not be removed but it should be used here:

https://github.com/dipy/dipy/blob/8be51800309373135cdda26bbb7bfd9d310c1d41/dipy/reconst/fwdti.py#L423

Can you update this, then it is ready to go. Thanks!

skoudoro (changes requested):
Hi @jhlegarreta,

See my comment below. Can you update this when you have time. Thanks!

dipy/reconst/fwdti.py
jhlegarreta: :+1: Nice catch. Done.

skoudoro (approved):
Thank you for the update @jhlegarreta , merging !

PR #3333, opened 2024-08-26
RF: Decorator fix

deka27:
### Description:
This PR refactors the version handling logic in the warning_for_keywords decorator by using the packaging module for parsing and comparing versions.

- Added packaging.version.parse for accurate version parsing.
- Improved version comparisons by using the base_version attribute, which removes any pre-release or development suffixes.

This change improves the reliability of version-based warnings and argument handling in the decorator.

### DIFF
--- a/dipy/core/onetime.py
+++ b/dipy/core/onetime.py
@@ -54,7 +54,6 @@ Hettinger. http://users.rcn.com/python/download/Descriptor.htm
 [2] Python data model, https://docs.python.org/reference/datamodel.html
 """
 
-from dipy.testing.decorators import warning_for_keywords
 
 # ----------------------------------------------------------------------------
 # Classes and Functions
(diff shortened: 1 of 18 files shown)


arokem (changes requested):
This requires testing with a large range of input options. I think that a more robust solution may be implemented with https://packaging.pypa.io/en/latest/version.html#packaging.version.Version

deka27:
> This requires testing with a large range of input options. I think that a more robust solution may be implemented with https://packaging.pypa.io/en/latest/version.html#packaging.version.Version

Looking into this.

deka27:
@arokem I have updated the code. Is this implementation better?
Also, thank you for the feedback.

arokem:
Yes - much better! Could you please add some tests?

skoudoro (approved):
LGTM. 

Any additional comment @arokem ?

arokem:
LGTM. Thanks for adding tests!

PR #3345, opened 2024-08-31
DOC: Skip element in documentation generation

jhlegarreta:
Skip element in documentation generation: add a function to the Sphinx setup in the configuration file so that the element at issue is not rendered when generating the documentation.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -54,6 +54,10 @@ numpydoc_show_inherited_class_members = {
     "dipy.viz.horizon.visualizer.peak.PeakActor": False
 }
 
+autodoc_skip_members = [
+    "docstring_addendum",
+]
+
 # Sphinx extension for BibTeX style citations.
 # https://github.com/mcmtroffaes/sphinxcontrib-bibtex
 bibtex_bibfiles = ["references.bib"]
@@ -559,3 +563,12 @@ sphinx_gallery_conf = {
 
 # Example configuration for intersphinx: refer to the Python standard library.
 intersphinx_mapping = {"python": ("https://docs.python.org/3/", None)}
+
+
+def setup(app):
+    def skip_doc_element(app, what, name, obj, would_skip, options):
+        if name in autodoc_skip_members:
+            return True
+        return would_skip
+
+    app.connect("autodoc-skip-member", skip_doc_element)


jhlegarreta:
Some notes:
- Makes the docstring_addendum to be skipped and be only added to the `__init__` docstring. It looks like this in dipy.org:




- The snippet is borrowed from an example; not sure if all the unused parameters have a purpose in our case; also considering that we will need to add other names to the skip list, e.g.
https://github.com/dipy/dipy/actions/runs/10635388112/job/29484847069#step:5:723

jhlegarreta:
Have tried doing something similar locally for one of the many global variables that are raising warnings: e.g. for `octahedron_vertices`:

- Adding it to the skipping method:
  ```
  def setup(app):
      def skip_doc_element(app, what, name, obj, would_skip, options):
          if name == "docstring_addendum":
              return True
          elif name == "octahedron_vertices":
              return True
  
          if what == "dipy.core.sphere" and isinstance(obj, dipy.core.sphere.octahedron_vertices):
              return True
          return would_skip
  
      app.connect("autodoc-skip-member", skip_doc_element)
  ```

- Adding a new case to the above method to skip objects that are not callable:
  ```
     if not callable(obj):
        return True
  ```
- Explicitly listing it to exclude from the autodoc generation:
  ```
   autodoc_default_options = {'exclude-members': 'octahedron_vertices'}
  ```
- Adding `#:` after the definition of `octahedron_vertices`.

None of the above attempts have worked, so I think this should be merged as is, and to try to address those other warnings in a separate PR. Maybe #3348 addresses them.

skoudoro (commented):
Hi @jhlegarreta,

Thank you for looking into this. 

I will not merge this PR, I think this need to be done at autodoc level or api generation script level. 

I agree that variable need to be skipped, like private functions.

I will look into it today and update https://github.com/dipy/dipy/pull/3348 accordingly

skoudoro:
When #3348 is merged, this PR needs to be updated following https://github.com/dipy/dipy/pull/3348/commits/9c2cc0ddfc6df7d5a675ec8edc715f584eb1c63a#diff-c27bbdb8753faa2bf17aa40da2bc0f1dbd01ea82be31bb009a81a9696a8897cfR95

jhlegarreta:
After rebasing and adopting the solution in 9c2cc0d does not fix the problem for me locally:



Have tried with the plain regex ` r'.*docstring_addendum.*',` and something a little bit more sophisticated (`r'\bdocstring_addendum = .*',`) does not work either.

With the proposed fix I get:



NB: In any case, I do not get the `__init__` docstring (even in the `master` branch, so I assume that can be disregarded in the screenshots)

doc/conf.py
```diff
@@ -551,3 +551,12 @@
 
 # Example configuration for intersphinx: refer to the Python standard library.
 intersphinx_mapping = {"python": ("https://docs.python.org/3/", None)}
+
+
+def setup(app):
+    def skip_doc_element(app, what, name, obj, would_skip, options):
+        if name == "docstring_addendum":
```
skoudoro, line 558: `if name in autodoc-skip-members:` instead

please create a new variable just after `numpydoc_show_inherited_class_members` named

```python
autodoc-skip-members = [
    "docstring_addendum",
]
```

skoudoro (changes requested):
Hi @jhlegarreta,

I have made some research and it seems to be the only viable and fastest solution for class members.
I proposed to create a variable to handle that. See my comment below. 

After this change, it will be ready to go. 

Thank you

doc/conf.py
skoudoro: feel free to change this variable name if you have better naming

jhlegarreta: :+1: Done in 0697bf9.

skoudoro (approved):
Thank you @jhlegarreta! merging

PR #3346, opened 2024-08-31
RF: Removal of keyword form Cython files

deka27:
#### This PR removes keyword decorator from the cython files.


skoudoro (approved):
thank you for cleaning up this @deka27, merging

PR #3348, opened 2024-08-31
DOC: Address remaining some warnings

skoudoro:
- [doc: fix Inline emphasis start-string without end-string](https://github.com/dipy/dipy/commit/7183d6f563cc3e644c86b614ce3de3ce354e0024)
- [doc: remove link_names](https://github.com/dipy/dipy/commit/f3b1318c886702727fb2b2f132e128fb7f507879)
- fix #3326
- bump python version for `build_docs` CI

PR #3349, opened 2024-09-02
DOC: Fix footbibliography-related errors in workflow help doc

jhlegarreta:
Fix footbibliography-related errors in workflow help documentation: remove the `sphinxcontrib-bibtex` `.. footbibliography::` directive string from the generated reStructuredText files corresponding to the workflow files help documentation.

The directive is used to introduce automatically the bibliography in the Python code, and is kept when generating the workflow file reStructuredText documentation: it needs to be removed since it has already served its purpose to introduce the references and generates warnings.

Fixes:
```
/dipy/dipy/doc/reference_cmd/dipy_buan_profiles.rst:38:
 WARNING: Explicit markup ends without a blank line; unexpected unindent.
```

and other similar errors raised for example at:
https://github.com/dipy/dipy/actions/runs/10647886977/job/29516329417#step:5:849

### DIFF
--- a/doc/tools/docgen_cmd.py
+++ b/doc/tools/docgen_cmd.py
@@ -163,6 +163,7 @@ if __name__ == '__main__':
                 format_title("output arguments(optional)"))
             help_txt = help_txt.replace("References:",
                                         format_title("References"))
+            help_txt = help_txt.replace(".. footbibliography::", "")
             help_txt = help_txt.rstrip()
             fp.write(help_txt)
 


jhlegarreta:
@skoudoro Not sure this is the most appropriate fix: the directive is printed when calling the workflow files with `-h` in teh command line. If there is a way, we should remove it from there, which would avoid the need for this workaround. Have not found how to interfere in the process of generating that help, though.

skoudoro:
> Not sure this is the most appropriate fix:

Agree, I do not think I will merge it.

> the directive is printed when calling the workflow files with -h in the command line.

yes, I saw that, we need to check in `sphinxcontrib-bibtex`  how we can render this part via code, then we can introduce it to the help. 

> Have not found how to interfere in the process of generating that help

it is via argparse, numpydoc, and our custom class `IntrospectiveArgumentParser``


In general, we need to update the workflow documentation generator script

skoudoro:
closing in favor of #3348 and specifically 95eb63ed90b96c1d0401781cba07150ab3415800

PR #3355, opened 2024-09-13
FIX: missing keyword only arguments on example

skoudoro:
This short PR fix some warnings on our tutorials.

All warnings were coming from keyword-only arguments feature recently added

### DIFF
--- a/doc/examples/syn_registration_2d.py
+++ b/doc/examples/syn_registration_2d.py
@@ -91,13 +91,13 @@ regtools.plot_2d_diffeomorphic_map(mapping, delta=10, fname="diffeomorphic_map.p
 # Now let's warp the moving image and see if it gets similar to the static
 # image
 
-warped_moving = mapping.transform(moving, "linear")
+warped_moving = mapping.transform(moving, interpolation="linear")
 regtools.overlay_images(
     static,
     warped_moving,
     title0="Static",
     title_mid="Overlay",
-    title_1="Warped moving",
+    title1="Warped moving",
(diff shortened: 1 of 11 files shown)


doc/examples/syn_registration_2d.py
```diff
@@ -91,13 +91,13 @@
...
 
-warped_moving = mapping.transform(moving, "linear")
+warped_moving = mapping.transform(moving, interpolation="linear")
 regtools.overlay_images(
     static,
     warped_moving,
     title0="Static",
     title_mid="Overlay",
-    title_1="Warped moving",
+    title1="Warped moving",
```
arokem, line 100: Is this a recent change to the api of this function?

arokem (approved):

arokem:
Overall looks good to me but just one small question

doc/examples/syn_registration_2d.py
skoudoro: no, it was a typo not detected on a previous PR

jhlegarreta:
BTW, how come these have not been captured by the warning decorators @deka27 ?

jhlegarreta:
> they are capture, we just do not run the examples in our CI. I try to run each one of them to prepare the reease

If the reason is that they take too long, maybe we could run them only once a week.

skoudoro:
> If the reason is that they take too long, maybe we could run them only once a week.

github action limit is 6h. We need to try but I believe it take more than 6h to run all examples. A strategy could be to split them in different build. We just need to block time to put that in place

PR #3323, opened 2024-08-19
BF: Allow passing kwargs in fit method, by moving parallelization kwargs elsewhere, including PEP 3102

arokem:
Replaces #3286

Issue #3222, opened 2024-05-16
Issue with CWLS in dki


samcoveney:
I need some help figuring out if there is a problem with constrained WLS in the DKI module. See `dipy/reconst/dki.py`

https://github.com/dipy/dipy/blob/e5cc120bf6c5c271f32cced62150157abd15a321/dipy/reconst/dki.py#L2611-L2619

So here we see that, in order to solve a weighted least squares problem, we define W as "signal" (such that W is that square root of the weight matrix that would be applied to the squared residuals, which is "signal squared"). Then we redefine `y` and `A` so that we can solve the weight least squares problem. This is fine as a way to solve WLS, see https://en.wikipedia.org/wiki/Weighted_least_squares

The problem is that we are performing constrained weighted least squares. The positivity constraints are set up here:

https://github.com/dipy/dipy/blob/e5cc120bf6c5c271f32cced62150157abd15a321/dipy/reconst/dki.py#L1795-L1814

But my understanding is that these constraints are set up with the understanding that `y` is log(signal). But we redefine `y` in the constrained weighted least squares.

As detailed in #3170 I am having problems writing the tests for CWLS (several of which were missing...) using noisy data - the tests keep failing. No problem with non-noisy data, probably because the solver detects that the problem is _already solved_ (weighting zero residuals results in zero-residuals i.e. no effect).

For example:

```
    # testing return of S0 when mask is inputted
    mask_test = dki_S0F.fa > 0
    for fit_method in tested_methods:
        kwargs = {}
        if fit_method in ["CLS", "CWLS"]:
            kwargs = {"cvxpy_solver": cvxpy.CLARABEL}
        dki_S0M = dki.DiffusionKurtosisModel(gtab_2s, fit_method=fit_method,
                                             return_S0_hat=True, **kwargs)
        dki_S0F = dki_S0M.fit(DWI + np.random.normal(size=DWI.shape), mask=mask_test) # NOTE: added noise to track down the problem - if noise added, CWLS solve fails here! Weighted problem not working!
        dki_S0F_S0 = dki_S0F.model_S0

```

Resulting in:

```
_solver_opts = {}

    def solve(_solver_opts):
    
        _settings = CLARABEL.parse_solver_opts(verbose, _solver_opts)
        _solver = clarabel.DefaultSolver(P, c, A, b, cones, _settings)
>       _results = _solver.solve()
E       pyo3_runtime.PanicException: Eigval error: Eigen(1)
```
[...]

PR #3358, opened 2024-09-17
robust algorithm REBASE

samcoveney:
Attempt to replace #3170 after rebasing on changes made in #3323

Probably need help from @arokem and @skoudoro on this

### DIFF
--- a/dipy/reconst/dki.py
+++ b/dipy/reconst/dki.py
@@ -38,8 +38,8 @@ from dipy.reconst.utils import dki_design_matrix as design_matrix
 from dipy.reconst.vec_val_sum import vec_val_vect
 from dipy.testing.decorators import warning_for_keywords
 from dipy.reconst.weights_method import (
-    weights_method_nlls_gm,
-    weights_method_wls_gm,
+    weights_method_nlls_m_est,
+    weights_method_wls_m_est,
 )
 
 
@@ -1824,10 +1824,7 @@ class DiffusionKurtosisModel(ReconstModel):
                                               'LS', 'LLS', 'OLS', 'OLLS',
                                               'CLS', 'CWLS']
 
-        if "weights_method" in self.kwargs:
-            self.is_iter_method = True
-        else:
-            self.is_iter_method = False
+        self.is_iter_method = "weights_method" in self.kwargs
 
         if "num_iter" not in self.kwargs and self.is_iter_method:
             self.kwargs["num_iter"] = 4
@@ -1957,7 +1954,7 @@ class DiffusionKurtosisModel(ReconstModel):
         return DiffusionKurtosisFit(self, params, model_S0=S0_params), extra
 
     def iterative_fit(self, data_thres, *, mask=None, num_iter=4,
-                      weights_method=weights_method_wls_gm):
+                      weights_method=weights_method_wls_m_est):
         """ Iteratively Reweighted fitting for the DKI model.
 
         Parameters
@@ -1969,7 +1966,7 @@ class DiffusionKurtosisModel(ReconstModel):
             should be analyzed that has the shape data.shape[-1]
         num_iter : int, optional
             Number of times to iterate.
-        weights_method : function, optional
+        weights_method : callable, optional
             A function with args and returns as follows:
 
             (weights, robust) =
@@ -1989,17 +1986,18 @@ class DiffusionKurtosisModel(ReconstModel):
         if num_iter < 2:  # otherwise, weights_method will not be utilized
             raise ValueError("num_iter must be 2+")
 
+        w, robust = True, None  # w = True ensures WLS (not OLS)
+        tmp, extra, leverages = None, None, None  # initialize, for clarity
+
         TDX = num_iter
         for rdx in range(1, TDX + 1):
 
-            if rdx == 1:
-                w, robust = True, None  # w = True ensures WLS (not OLS)
-            else:
+            if rdx > 1:  #  after first iteration, update weights
                 # if using mask, define full-sized arrays for w and robust
                 if rdx == 2 and mask is not None:
                     cond = (mask == 1)
                     w = np.ones_like(data_thres, dtype=float)
-                    robust = np.zeros_like(data_thres)
+                    robust = np.zeros_like(data_thres, dtype=bool)
                     robust[cond] = 1
 
                 # make prediction of the signal
--- a/dipy/reconst/dki.py
+++ b/dipy/reconst/dki.py
@@ -1722,7 +1722,8 @@ class DiffusionKurtosisModel(ReconstModel):
     """Class for the Diffusion Kurtosis Model"""
 
     @warning_for_keywords()
-    def __init__(self, gtab, *args, fit_method="WLS", return_S0_hat=False, **kwargs):
+    def __init__(self, gtab, *args,
+                 fit_method="WLS", return_S0_hat=False, **kwargs):
         """Diffusion Kurtosis Tensor Model.
 
         See :footcite:p:`Jensen2005` and :footcite:p:`NetoHenriques2021a` for
@@ -1857,10 +1858,11 @@ class DiffusionKurtosisModel(ReconstModel):
             return fit_result
 
         if self.is_multi_method and self.is_iter_method:
-            fit_result, extra =\
-              self.iterative_fit(data_thres, mask=mask,
-                                 num_iter=self.kwargs["num_iter"],
-                                 weights_method=self.kwargs["weights_method"])
+            fit_result, extra = self.iterative_fit(
+                data_thres, mask=mask,
+                num_iter=self.kwargs["num_iter"],
+                weights_method=self.kwargs["weights_method"]
+            )
             if extra is not None:
                 self.extra = extra
             return fit_result
@@ -1926,8 +1928,7 @@ class DiffusionKurtosisModel(ReconstModel):
 
     @multi_voxel_fit
     @warning_for_keywords()
-#    def multi_fit(self, data, *, mask=None):  # FIXME: this was current master
-    def multi_fit(self, data, *, mask=None, **kwargs):  # FIXME: I think I need 'kwargs'... else how to pass to fit_method below? 
+    def multi_fit(self, data, *, mask=None, **kwargs):
         """ Convenience function for fitting multiple voxels.
         """
         extra_args = (
@@ -2018,7 +2019,6 @@ class DiffusionKurtosisModel(ReconstModel):
                                      self.design_matrix, leverages,
                                      rdx, TDX, robust)
 
-            #print(rdx, TDX + 1)  # FIXME: leverages is not getting returned
             tmp, extra = self.multi_fit(data_thres, mask=mask,
                                         weights=w, return_leverages=True)
             leverages = extra["leverages"]
@@ -2797,8 +2797,6 @@ def cls_fit_dki(
     ----------
     .. footbibliography::
     """
-    #print("running cls")
-    #return_leverages = True  # NOTE: forcing this, because multi-voxel not passing it # FIXME: needs removing
     # Set up least squares problem
     A = design_matrix
     y = np.log(data)
@@ -2808,7 +2806,6 @@ def cls_fit_dki(
     # e.g. weights for WLS are diag(fn**2), so W are diag(fn)
     if weights is not False:
         if type(weights) is np.ndarray:  # use supplied weights
-            #print(A.shape, y.shape, weights.shape) # FIXME: remove...
             W = np.diag(np.sqrt(weights))
         else:  # Define weights as diag(fn**2) (fn = fitted signal from OLS)
             result = np.dot(inverse_design_matrix, y)
@@ -2824,7 +2821,6 @@ def cls_fit_dki(
     # Solve sdp
     result = sdp.solve(A, y, check=True, solver=cvxpy_solver)
 
-    print("return_leverages:", return_leverages) # FIXME: multi_voxel parallel should be passing this
     if return_leverages:
         leverages = {"leverages": leverages}
     else:
--- a/dipy/reconst/dki.py
+++ b/dipy/reconst/dki.py
@@ -1955,7 +1955,7 @@ class DiffusionKurtosisModel(ReconstModel):
 
     def iterative_fit(self, data_thres, *, mask=None, num_iter=4,
                       weights_method=weights_method_wls_m_est):
-        """ Iteratively Reweighted fitting for the DKI model.
+        """Iteratively Reweighted fitting for the DKI model.
 
         Parameters
         ----------
--- a/dipy/reconst/dki.py
+++ b/dipy/reconst/dki.py
@@ -2007,17 +2007,17 @@ class DiffusionKurtosisModel(ReconstModel):
 
                 # define weights for next fit
                 if mask is not None:
-                    w[cond], robust_mask =\
-                      weights_method(data_thres[cond], pred_sig[cond],
-                                     self.design_matrix, leverages[cond],
-                                     rdx, TDX, robust[cond])
+                    w[cond], robust_mask = weights_method(
+                      data_thres[cond], pred_sig[cond], self.design_matrix,
+                      leverages[cond], rdx, TDX, robust[cond]
+                    )
                     if robust_mask is not None:
                         robust[cond] = robust_mask
                 else:
-                    w, robust =\
-                      weights_method(data_thres, pred_sig,
-                                     self.design_matrix, leverages,
-                                     rdx, TDX, robust)
+                    w, robust = weights_method(
+                      data_thres, pred_sig, self.design_matrix,
+                      leverages, rdx, TDX, robust
+                    )
 
             tmp, extra = self.multi_fit(data_thres, mask=mask,
                                         weights=w, return_leverages=True)
--- a/dipy/reconst/dki.py
+++ b/dipy/reconst/dki.py
@@ -8,7 +8,10 @@ import warnings
 import numpy as np
 import scipy.optimize as opt
 
-from dipy.core.geometry import cart2sphere, perpendicular_directions, sphere2cart
+from dipy.core.geometry import (
+    cart2sphere,
+    perpendicular_directions,
+    sphere2cart)
 from dipy.core.gradients import check_multi_b
 from dipy.core.ndindex import ndindex
 from dipy.core.optimize import PositiveDefiniteLeastSquares
@@ -20,17 +23,24 @@ from dipy.reconst.dti import (
     TensorFit,
     decompose_tensor,
     from_lower_triangular,
+    iterative_fit_tensor,
     lower_triangular,
     mean_diffusivity,
     nlls_fit_tensor,
     radial_diffusivity,
     restore_fit_tensor,
+    robust_fit_tensor_nlls,
+    robust_fit_tensor_wls,
 )
 from dipy.reconst.multi_voxel import multi_voxel_fit
 from dipy.reconst.recspeed import local_maxima
 from dipy.reconst.utils import dki_design_matrix as design_matrix
 from dipy.reconst.vec_val_sum import vec_val_vect
 from dipy.testing.decorators import warning_for_keywords
+from dipy.reconst.weights_method import (
+    weights_method_nlls_gm,
+    weights_method_wls_gm,
+)
 
 
 @warning_for_keywords()
@@ -1759,7 +1769,7 @@ class DiffusionKurtosisModel(ReconstModel):
         self.common_fit_method = not callable(fit_method)
         if self.common_fit_method:
             try:
-                self.fit_method = common_fit_methods[fit_method]
+                self.fit_method = common_fit_methods[fit_method.upper()]
             except KeyError as e:
                 msg = '"' + str(fit_method) + '" is not a known fit method. '
                 msg += " The fit method should either be a function or one of "
@@ -1807,17 +1817,21 @@ class DiffusionKurtosisModel(ReconstModel):
                 )
             self.sdp = PositiveDefiniteLeastSquares(22, A=self.sdp_constraints)
 
-        self.weights = fit_method in {"WLS", "WLLS", "UWLLS", "CWLS"}
-        self.is_multi_method = fit_method in [
-            "WLS",
-            "OLS",
-            "UWLLS",
-            "ULLS",
-            "WLLS",
-            "OLLS",
-            "CLS",
-            "CWLS",
-        ]
+        self.weights = fit_method in {'WLS', 'WLLS', 'CWLS'}
+
+        self.is_multi_method = fit_method in ['WLS', 'WLLS',
+                                              'LS', 'LLS', 'OLS', 'OLLS',
+                                              'CLS', 'CWLS']
+
+        if "weights_method" in self.kwargs:
+            self.is_iter_method = True
+        else:
+            self.is_iter_method = False
+
+        if "num_iter" not in self.kwargs and self.is_iter_method:
+            self.kwargs["num_iter"] = 4
+
+        self.extra = {}
 
     @warning_for_keywords()
     def fit(self, data, *, mask=None):
@@ -1826,18 +1840,34 @@ class DiffusionKurtosisModel(ReconstModel):
         Parameters
         ----------
         data : array
-            The measured signal from one voxel.
+            The measured signal.
         mask : array
             A boolean array used to mark the coordinates in the data that
             should be analyzed that has the shape data.shape[-1]
 
         """
         data_thres = np.maximum(data, self.min_signal)
-        if self.is_multi_method:
-            return self.multi_fit(data_thres, mask=mask)
+
+        if self.is_multi_method and not self.is_iter_method:
+            fit_result, extra = self.multi_fit(data_thres, mask=mask,
+                                               weights=self.weights,
+                                               **self.kwargs)
+            if extra is not None:
+                self.extra = extra
+            return fit_result
+
+        if self.is_multi_method and self.is_iter_method:
+            fit_result, extra =\
+              self.iterative_fit(data_thres, mask=mask,
+                                 num_iter=self.kwargs["num_iter"],
+                                 weights_method=self.kwargs["weights_method"])
+            if extra is not None:
+                self.extra = extra
+            return fit_result
 
         S0_params = None
-        if data.ndim == 1:
+
+        if data.ndim == 1:  # NOTE: doesn't use mask...
             params, extra = self.fit_method(
                 self.design_matrix,
                 data_thres,
@@ -1847,6 +1877,9 @@ class DiffusionKurtosisModel(ReconstModel):
             )
             if self.return_S0_hat:
                 params, S0_params = params
+            if extra is not None:
+                for key in extra:
+                    self.extra[key] = extra[key]
             return DiffusionKurtosisFit(self, params, model_S0=S0_params)
 
         if mask is not None:
@@ -1873,20 +1906,30 @@ class DiffusionKurtosisModel(ReconstModel):
             out_shape = data.shape[:-1] + (-1,)
             dki_params = params.reshape(out_shape)
             if self.return_S0_hat:
-                S0_params = S0_params.reshape(out_shape).squeeze()
+                S0_params = S0_params.reshape(out_shape).squeeze(axis=-1)
+            if extra is not None:
+                for key in extra:
+                    self.extra[key] = extra[key].reshape(data.shape)
         else:
             dki_params = np.zeros(data.shape[:-1] + (27,))
             dki_params[mask, :] = params
             if self.return_S0_hat:
                 S0_params_in_mask = np.zeros(data.shape[:-1])
-                S0_params_in_mask[mask] = S0_params.squeeze()
+                S0_params_in_mask[mask] = S0_params.squeeze(axis=-1)
                 S0_params = S0_params_in_mask
+            if extra is not None:
+                for key in extra:
+                    self.extra[key] = np.zeros(data.shape)
+                    self.extra[key][mask, :] = extra[key]
 
         return DiffusionKurtosisFit(self, dki_params, model_S0=S0_params)
 
     @multi_voxel_fit
     @warning_for_keywords()
-    def multi_fit(self, data, *, mask=None):
+#    def multi_fit(self, data, *, mask=None):  # FIXME: this was current master
+    def multi_fit(self, data, *, mask=None, **kwargs):  # FIXME: I think I need 'kwargs'... else how to pass to fit_method below? 
+        """ Convenience function for fitting multiple voxels.
+        """
         extra_args = (
             {}
             if not self.convexity_constraint
@@ -1895,21 +1938,93 @@ class DiffusionKurtosisModel(ReconstModel):
                 "sdp": self.sdp,
             }
         )
+
         params, extra = self.fit_method(
             self.design_matrix,
             data,
             self.inverse_design_matrix,
             return_S0_hat=self.return_S0_hat,
-            weights=self.weights,
             min_diffusivity=self.min_diffusivity,
             **extra_args,
+            **kwargs,
         )
 
         S0_params = None
         if self.return_S0_hat:
             params, S0_params = params
 
-        return DiffusionKurtosisFit(self, params, model_S0=S0_params)
+        return DiffusionKurtosisFit(self, params, model_S0=S0_params), extra
+
+    def iterative_fit(self, data_thres, *, mask=None, num_iter=4,
+                      weights_method=weights_method_wls_gm):
+        """ Iteratively Reweighted fitting for the DKI model.
+
+        Parameters
+        ----------
+        data_thres : array
+            The measured signal.
+        mask : array, optional
+            A boolean array used to mark the coordinates in the data that
+            should be analyzed that has the shape data.shape[-1]
+        num_iter : int, optional
+            Number of times to iterate.
+        weights_method : function, optional
+            A function with args and returns as follows:
+
+            (weights, robust) =
+              weights_method(data, pred_sig,
+                             design_matrix, leverages,
+                             idx, num_iter,
+                             robust)
+
+        Notes
+        -----
+        On the first iteration, (C)WLS fit is performed. Subsequent iterations
+        will be weighted according to weights_method.  Outlier rejection should
+        be handled within weights_method by setting the corresponding weights
+        to zero (if weights_method implements outlier rejection).
+
+        """
+        if num_iter < 2:  # otherwise, weights_method will not be utilized
+            raise ValueError("num_iter must be 2+")
+
+        TDX = num_iter
+        for rdx in range(1, TDX + 1):
+
+            if rdx == 1:
+                w, robust = True, None  # w = True ensures WLS (not OLS)
+            else:
+                # if using mask, define full-sized arrays for w and robust
+                if rdx == 2 and mask is not None:
+                    cond = (mask == 1)
+                    w = np.ones_like(data_thres, dtype=float)
+                    robust = np.zeros_like(data_thres)
+                    robust[cond] = 1
+
+                # make prediction of the signal
+                pred_sig = self.predict(tmp.model_params, S0=tmp.model_S0)
+
+                # define weights for next fit
+                if mask is not None:
+                    w[cond], robust_mask =\
+                      weights_method(data_thres[cond], pred_sig[cond],
+                                     self.design_matrix, leverages[cond],
+                                     rdx, TDX, robust[cond])
+                    if robust_mask is not None:
+                        robust[cond] = robust_mask
+                else:
+                    w, robust =\
+                      weights_method(data_thres, pred_sig,
+                                     self.design_matrix, leverages,
+                                     rdx, TDX, robust)
+
+            #print(rdx, TDX + 1)  # FIXME: leverages is not getting returned
+            tmp, extra = self.multi_fit(data_thres, mask=mask,
+                                        weights=w, return_leverages=True)
+            leverages = extra["leverages"]
+
+        extra = {"robust": robust}
+        return tmp, extra
 
     @warning_for_keywords()
     def predict(self, dki_params, *, S0=1.0):
@@ -2513,6 +2628,8 @@ def ls_fit_dki(
     return_S0_hat=False,
     weights=True,
     min_diffusivity=0,
+    return_lower_triangular=False,
+    return_leverages=False,
 ):
     r"""Compute the diffusion and kurtosis tensors using an ordinary or
     weighted linear least squares approach.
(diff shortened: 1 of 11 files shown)


samcoveney:
I'll tidy the code up, but I think a lot of tests are failing due to an import error for ray? Did I accidentally remove some code, or is this a problem at your end?

skoudoro:
> Did I accidentally remove some code, or is this a problem at your end?

I do not think it is at our end. 

Should we start to review this PR to help?

samcoveney:
Sure please start to review. Bear in mind it already had a lot of review before.

The testing errors I thought were about not finding ray. I wasn't aware that I had removed the corresponding imports. Maybe I am wrong or I forgot what the errors were

samcoveney:
question: I run "pycodestyle" (up to date) but it doesn't find pep8 issues that then get flagged here. This is annoying. What am I apparently doing wrong? Surely, there has to be a better way than pushing in order to see the (supposed) pep8 issues

samcoveney:
I don't know how to get pep8speaks to shut up about some things, as it seems I am stuck with one complaint or another, so please recommend, what exactly to do here?

https://github.com/dipy/dipy/blob/80b50a6eec518c0061e4881da371808e682395e1/dipy/reconst/dki.py#L2009C1-L2013C22

skoudoro:
Hi @samcoveney,

No worries, this PR will not miss the next release. 

Personally, the review has started, but get interrupted multiple times. All current deadlines are done, next ones are in the beginning of November so I should be able to push this and other PRs this week and next week.

samcoveney:
Thanks @skoudoro - like last time, let me know if there are any things in other PR I can help with

arokem:
Hello! I am in the midst of reviewing the code and it looks overall quite good. Could you please say a bit more about how RESTORE was reimplemented here? I see the removal of the sigma input and the Geman McClure estimator, but not entirely sure how to think about this change, and how we might be sure this is an improvement over the previous implementation.

arokem:
Also,  is there a reference for the iterative reweighting in the context of dMRI?

arokem:
I think the answer to my latter question is Collier, Q., Veraart, J., Jeurissen, B., den Dekker, A., Sijbers, J., 2015. Iterative reweighted linear least squares for accurate, fast, and robust estimation of diffusion magnetic resonance parameters: IRLLS for Estimation of Diffusion MR Parameters. Magnetic Resonance in Medicine 73, 2174-2184. Correct?

dipy/reconst/weights_method.py
```diff
@@ -0,0 +1,248 @@
...
+    usual form of the WLS problem is accounted for when defining these new
+    weights. On the second-to-last iteration, OLS is performed without
+    outliers. On the last iteration, WLS is performed without outliers.
+
+    """
+    # check if M-estimator is valid (defined in this function)
+    if m_est not in ["gm", "cauchy"]:
+        raise ValueError("unknown M-estimator choice")
+
+    # min 4 iters: (1) WLS (2) WLS with M-weights (3) clean OLS (3) clean WLS
```
arokem, line 80: Not sure I understand this comment. These methods all require min 4 iterations?

dipy/reconst/weights_method.py
```diff
@@ -0,0 +1,248 @@
...
+        w[robust == 0] = 0.0
+        w[robust == 1] = 1.0
+
+    w[np.isinf(w)] = 0
+    w[np.isnan(w)] = 0
+
+    return w, robust
+
+
+# define specific M-estimators
```
arokem, line 220: Why are these here? If they are supposed to be used as a public API, please add proper docstrings.

dipy/utils/parallel.py
```diff
@@ -29,16 +29,19 @@ def paramap(
     func_kwargs=None,
     **kwargs,
 ):
+    # FIXME: but several fitting functions return "extra" as well
```
arokem, line 32: Is this still an issue? Does any of your tests address this?

arokem (commented):
Looks great. I think that weights_method needs a test suite.

dipy/reconst/dti.py
```diff
@@ -772,7 +775,7 @@ def __init__(self, gtab, *args, fit_method="WLS", return_S0_hat=False, **kwargs)
         self.extra = {}
 
     @warning_for_keywords()
-    def fit(self, data, *, mask=None, adjacency=False):
+    def fit(self, data, *, mask=None):
```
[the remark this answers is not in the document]
skoudoro: I do not think it is necessary, it was added only on the previous release. I do not think it is used a lot

dipy/reconst/dti.py
```diff
@@ -1377,7 +1359,7 @@ def wrapped_fit_tensor(
                             extra[key] = np.empty(data.shape)
                         extra[key][i : i + step] = extra_i[key]
 
-            if extra is not None:
+            if extra:  # if extra is not just {}
```
skoudoro, line 1362: comment not needed

dipy/reconst/dti.py
```diff
@@ -1411,8 +1394,17 @@ def wls_fit_tensor(design_matrix, data, *, return_S0_hat=False):
     data : array ([X, Y, Z, ...], g)
         Data or response variables holding the data. Note that the last
         dimension should contain the data. It makes no copies of data.
+    weights : array ([X, Y, Z, ...], g), optional
```
skoudoro, line 1397: here, `weights=None`, you should follow the same logic above

dipy/reconst/dti.py
```diff
@@ -1634,71 +1623,44 @@ def err_func(self, tensor, design_matrix, data, weighting=None, sigma=None):
...
+            The voxel signal in all gradient directions
+
+        weights : array ([X, Y, Z, ...], g), optional
+            Weights to apply for fitting. These weights must correspond to the
+            squared residuals such that $S = \sum_i w_i r_i^2$.
 
         Notes
         -----
-        This is an implementation of equation 14 in :footcite:p:`Koay2006c`.
+        This is an implementation of equation 14 in [1]_, but also
```
skoudoro, line 1662: please, keep the `:footcite:p:`

dipy/reconst/dti.py
```diff
@@ -1900,12 +1894,19 @@ def nlls_fit_tensor(
...
+    fail_is_nan=False
 ):
     """
-    Use the RESTORE algorithm to calculate a robust tensor fit.
-
-    See :footcite:p:`Chang2005` for further details about the method.
+    Use the RESTORE algorithm [1]_ to calculate a robust tensor fit.
+    Note that [1]_ does not define Geman–McClure M-estimator weights as
+    claimed (instead, Cauchy M-estimator weights are defined), but this
+    function does define these weights correctly.
```
skoudoro, line 1909: please, keep the `footcite`

dipy/reconst/dti.py
```diff
@@ -2124,6 +2111,133 @@ def restore_fit_tensor(
...
+
+    design_matrix : array of shape (g, ...)
+        Design matrix holding the covariants used to solve for the regression
+        coefficients.
+
+    data : array of shape ([X, Y, Z, n_directions], g)
+        Data or response variables holding the data. Note that the last
+        dimension should contain the data. It makes no copies of data.
+
+    return_S0_hat : bool
```
skoudoro, line 2132: `, optional` to add

dipy/reconst/dti.py
```diff
@@ -2124,6 +2111,133 @@ def restore_fit_tensor(
...
+        Boolean to return (True) or not (False) the S0 values for the fit.
+
+    num_iter : int, optional
+        Number of times to iterate.
+
+    weights_method : function, optional
```
skoudoro, line 2138: `Callable` instead of function

dipy/reconst/multi_voxel.py
```diff
@@ -72,22 +110,61 @@ def new_fit(self, data, *, mask=None, **kwargs):
...
+                extra = {}
+                for key in extra_mask:
+                    extra[key] = np.zeros(data.shape)
+                    extra[key][mask == 1] = extra_mask[key]
+            else:
+                extra = None
+
+        # If fit method does not return extra, assume we cannot return extra
+        if return_extra:
+            return MultiVoxelFit(self, fit_array, mask), extra
```
skoudoro, line 165: I start to wonder if extra should not be part of `Fit` object

dipy/reconst/weights_method.py
```diff
@@ -0,0 +1,248 @@
...
+    leverages : array
+        The leverages (diagonal of the 'hat matrix') of the fit.
+    idx : int
+        The current iteration number.
+    total_idx : int
+        The total number of iterations.
+    last_robust : array (int),
+        Array of 1s (inlier) and 0s (outlier).
+        Has the same shape as data.
+    m_est : which M-estimator weighting scheme to use. Currently,
```
skoudoro, line 63: `str, optional` then description

dipy/reconst/weights_method.py
```diff
@@ -0,0 +1,248 @@
...
+            'gm' (Geman-McClure) and 'cauchy' are provided.
+    cut_off : float
```
skoudoro, line 65: `float, optional`instead

dipy/reconst/weights_method.py
```diff
@@ -0,0 +1,248 @@
...
+        Outliers are defined by residuals > |cut_off * standard_deviation|
+
```
skoudoro, line 67: missing one arguments

dipy/reconst/weights_method.py
```diff
@@ -0,0 +1,248 @@
...
+        The current iteration number.
+    total_idx : int
+        The total number of iterations.
+    last_robust : array (int),
+        Array of 1s (inlier) and 0s (outlier).
+        Has the same shape as data.
+    m_est : which M-estimator weighting scheme to use. Currently,
+            'gm' (Geman-McClure) and 'cauchy' are provided.
+    cut_off : float
+        Outliers are defined by residuals > |cut_off * standard_deviation|
```
skoudoro, line 161: same as above

dipy/reconst/weights_method.py
```diff
@@ -0,0 +1,248 @@
...
+
+    Notes
+    -----
+    Robust weights are calculated specifically for the NLLS problem.
+    On the last iteration, NLLS is performed without outliers.
+
+    """
+    # check if M-estimator is valid (defined in this function)
+    if m_est not in ["gm", "cauchy"]:
+        raise ValueError("unknown M-estimator choice")
```
skoudoro, line 171: Can you add what are the possible choices

dipy/reconst/weights_method.py
skoudoro: +1

skoudoro (commented):
Hi @samcoveney,

See below initial review, I am still testing your code and trying to understand some part of it but you can start cleaning.

Thank you!

dipy/reconst/dti.py
jhlegarreta: No blank lines needed between parameter docstrings. We are trying to move away from that so that docs are consistent. Applies to the whole PR.

dipy/reconst/weights_method.py
```diff
@@ -0,0 +1,248 @@
...
+
+import numpy as np
+
+MIN_POSITIVE_SIGNAL = 0.0001
+
+
+def simple_cutoff(residuals, log_residuals, pred_sig,
+                  design_matrix, leverages, C, cutoff):
+    """ Define outliers based on signal.
+    """
```
jhlegarreta, line 12: Parameter docstrings. Applies to the entire PR.

dipy/reconst/weights_method.py
```diff
@@ -0,0 +1,248 @@
...
+    leverages[np.isclose(leverages, 1.0)] = 0.99  # avoids rare issues
+    HAT_factor = np.sqrt(1 - leverages)
+    cond = (np.abs(residuals) > + cutoff * C * HAT_factor)
+    return cond
+
+
+def two_eyes_cutoff(residuals, log_residuals, pred_sig,
+                    design_matrix, leverages, C, cutoff):
+    """ Two-eyes approach to define outliers, see [1]_.
```
jhlegarreta, line 21: We have adopted [sphinxcontrib-bibtex](https://pypi.org/project/sphinxcontrib-bibtex/) for references.

dipy/reconst/weights_method.py
jhlegarreta: Revise `last_robust` parameter type: if it 1s and 0s, then why not a boolean? and docstring: use ndarray, remove the type, and remove the comma. Make "Has the same shape as data" dwell on the same line as the previous line (and have not checked the doc, but if it needs to have the same shape, is this checked in the code?).

The `m_est` docstring will most likely be rendered incorrectly: use only the type (missing as it stands now) on the same line as the parameter name, add the description on a separate line.

These apply to the entire PR.

dipy/reconst/weights_method.py
jhlegarreta: Check how this renders:
- Every time Sphinx finds an asterisk it thinks that an emphasized word starts (cannot say out the top of my head if a white space before the following word avoids this)
- May yield an `Undefined substitution referenced: "cut_off * standard_deviation".`: https://docutils.sourceforge.io/docs/ref/rst/restructuredtext.html#substitution-references

dipy/reconst/weights_method.py
jhlegarreta: Start sentence with capitals.

jhlegarreta:
Check also the `Build doc` warnings that are being introduced: https://github.com/dipy/dipy/actions/runs/11141206898/job/30961605431?pr=3358#step:5:632

dipy/reconst/tests/test_dki.py
```diff
@@ -271,16 +317,86 @@ def test_dki_fits():
...
     for fit_method in tested_methods:
         kwargs = {}
-        if fit_method == "CLS":
+        if fit_method in ["CLS", "CWLS"]:
             kwargs = {"cvxpy_solver": cvxpy.CLARABEL}
         dki_S0M = dki.DiffusionKurtosisModel(
             gtab_2s, fit_method=fit_method, return_S0_hat=True, **kwargs
         )
-        dki_S0F = dki_S0M.fit(DWI, mask=mask_test)
+        # FIXME: CLS and CWLS generally fail with noisy data - needs fixing
```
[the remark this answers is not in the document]
samcoveney: I have found the source of the problem with #3222 ; this test is broken - the scaling of the design matrix, determined by the units of bval, intrinsically breaks the constrained fitting, the only reason it 'worked' on noiseless data was that the constrained fitting never really ran at all, and the check on the solution was passed immediately.

This is something to look into, but I am not sure what the best answer is. In my own code, I simply scale the bvals first by dividing by 1000, then unscale the corresponding quantities after the solver. This fixes the CWLS and CLS problems. Happy to put that here, but also happy to leave for a future change, with #3222

samcoveney:
> Hello! I am in the midst of reviewing the code and it looks overall quite good. Could you please say a bit more about how RESTORE was reimplemented here? I see the removal of the sigma input and the Geman McClure estimator, but not entirely sure how to think about this change, and how we might be sure this is an improvement over the previous implementation.

The change to RESTORE removed the previous spatially variable sigma option, which I had put in the code in the first place. The reason for this is simplicity, and merging with iRESTORE and more modern implementations of robust fitting that estimate the noise level as part of the problem i.e. sigma is not an input. Of course, sigma can still be given as an input.

For the change regarding the GMM estimator, it's simply the case that the original RESTORE paper was wrong. They did IRLS, but did not use the weights for the GMM estimator, due to misunderstanding how to derive the IRLS weights from the robust cost function. The problem is: there isn't really a 'canonical' RESTORE algorithm, because what they say they do is different from what they actually do. At least, now, we are consistent with the intent of the algoroithm

dipy/reconst/tests/test_dki.py
```diff
@@ -271,16 +317,86 @@ def test_dki_fits():
...
         dki_S0F_S0 = dki_S0F.model_S0
 
         assert_array_almost_equal(dki_S0F_S0, np.full(dki_S0F_S0.shape, S0))
 
+    # testing robust fitting by with WLS and CWLS (needs noisy data)
+    # --------==============----------------------==================
+    YN = DWI + np.random.normal(size=DWI.shape)
+    if have_cvxpy:
+        tested_methods = ['WLS']
+        # FIXME: CWLS generally failing with noisy data, needs fixing
```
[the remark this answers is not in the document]
samcoveney: see reply above. Given delays so far, I would rather fix these 'issues' later.

dipy/reconst/dti.py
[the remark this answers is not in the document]
samcoveney: yeah not used much. In future, I will demonstrate how to build adjacency into a custom weight function using partials or lambda. It seems better to handle in this way, than to give every fitting function the adjacency keyword. The only reason to consider what voxels are adjacent, is weighting or outlier detection, therefore it falls under the IRLS umbrella

dipy/reconst/dti.py
```diff
@@ -1888,7 +1874,15 @@ def nlls_fit_tensor(
             params[vox, 12:] = this_param[6:-1] / md2
 
     if resort_to_OLS:
-        warnings.warn(ols_resort_msg, UserWarning, stacklevel=2)
+        warnings.warn(ols_resort_msg, UserWarning)
```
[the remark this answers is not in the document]
samcoveney: Did I remove this? Seems odd if I did, I can put back

dipy/reconst/dti.py
```diff
@@ -1844,36 +1828,38 @@ def nlls_fit_tensor(
...
+            evals, evecs = decompose_tensor(
+                from_lower_triangular(this_param[:6]),
+                min_diffusivity=tol / -design_matrix.min())
             params[vox, :3] = evals
             params[vox, 3:12] = evecs.ravel()
 
         # If leastsq failed to converge and produced nans, we'll resort to the
         # OLS solution in this voxel:
-        except (np.linalg.LinAlgError, TypeError):
+        except (np.linalg.LinAlgError, TypeError) as e:
```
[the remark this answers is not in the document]
samcoveney: Are you asking, why have I caught the error? It was probably for printing during testing... you suggest to remove the `as e` part?

dipy/reconst/dti.py
```diff
@@ -1985,60 +1992,49 @@ def restore_fit_tensor(
...
+                                                   flat_data[vox]),
+                                             Dfun=jac_func)
 
             # Get the residuals:
             pred_sig = np.exp(np.dot(design_matrix, this_param))
             residuals = flat_data[vox] - pred_sig
 
+            # estimate or set sigma
+            if sigma is not None:
+                C = sigma
```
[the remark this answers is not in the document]
samcoveney: `C` is the usual letter in dMRI literature for the MAD estimate of noise sigma. I would rather leave as is for this reason.

dipy/reconst/dti.py
```diff
@@ -1985,60 +1992,49 @@ def restore_fit_tensor(
...
-                        C**2,
-                        denominator,
-                        out=np.zeros_like(denominator),
-                        where=denominator != 0,
+                    # GM weights (original Restore paper used Cauchy weights)
+                    C = factor * np.median(
+                        np.abs(residuals - np.median(residuals))
                     )
+                    denominator = (C**2 + residuals**2)**2
+                    gmm = np.divide(C**2, denominator,
```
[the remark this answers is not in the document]
samcoveney: but i fixed it.... it now IS the GM M-estimator... and this is consistent with the new docstring

dipy/reconst/dti.py
```diff
@@ -1985,60 +1992,49 @@ def restore_fit_tensor(
...
+                                    out=np.zeros_like(denominator),
+                                    where=denominator != 0)
 
                     # Do nlls with GMM-weighting:
```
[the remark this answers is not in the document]
samcoveney: it is, see above:
` ``
    Use the RESTORE algorithm [1]_ to calculate a robust tensor fit.
    Note that [1]_ does not define Geman–McClure M-estimator weights as
    claimed (instead, Cauchy M-estimator weights are defined), but this
    function does define these weights correctly.
```

File: dipy/reconst/tests/test_dti.py [hunk 26]
```diff
@@ -559,36 +720,29 @@ def test_nnls_jacobian_func(rng):
...
-                if weighting == "gmm":
-                    args = [X[i], Y[i], weighting, sigma]
+
+                args = [X[i], Y[i], weights]
+
                 approx = opt.approx_fprime(D, nlls.err_func, 1e-8, *args)
 
-                assert np.allclose(approx, analytical[i])
+                # assert np.allclose(approx, analytical[i])
+                assert True  # NOTE: approx_fprime not accurate enough
```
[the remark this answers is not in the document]
samcoveney: See https://github.com/dipy/dipy/issues/3155

dipy/reconst/weights_method.py
samcoveney: yes, because you implement robust IRLS for the WLS problem (the robust form of WLS) in this way, without 4 iterations, because of the flexible way IRLS implement works here. First, one must do WLS as usual, then one must fit with robust weights, then fitting with weights as either 1 or 0 based on outlier detection for OLS, then the final WLS step.

dipy/reconst/weights_method.py
samcoveney: Good point, will do.

As for why they are here, it is because they implement specific M-estimators; it's supposed to make life easier for the user. If you disagree, I can remove them?

dipy/utils/parallel.py
samcoveney: I added tests on `extra` in the past and it was committed... and then this parallel stuff was added afterwards without seeming to consider it... so I don't really think I should be the one to add these tests, especially after finding errors in that original parallelization implementation that were precisely caused by not testing properly.

dipy/reconst/dki.py
```diff
@@ -1895,21 +1939,92 @@ def multi_fit(self, data, *, mask=None):
...
+                    robust[cond] = 1
+
+                # make prediction of the signal
+                pred_sig = self.predict(tmp.model_params, S0=tmp.model_S0)
+
+                # define weights for next fit
+                if mask is not None:
+                    w[cond], robust_mask = weights_method(
+                        data_thres[cond], pred_sig[cond], self.design_matrix,
+                        leverages[cond], rdx, TDX, robust[cond]
```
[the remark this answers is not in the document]
samcoveney: no this all works as designed, leverages *does* get initialized on the first iteration `rdx == 1`, there is no way to reach this part of the code before that.
(This has all basically been working for 6 months ;) )

dipy/reconst/dki.py
```diff
@@ -1895,21 +1939,92 @@ def multi_fit(self, data, *, mask=None):
...
+            else:
+                # if using mask, define full-sized arrays for w and robust
+                if rdx == 2 and mask is not None:
+                    cond = (mask == 1)
+                    w = np.ones_like(data_thres, dtype=float)
+                    robust = np.zeros_like(data_thres)
+                    robust[cond] = 1
+
+                # make prediction of the signal
+                pred_sig = self.predict(tmp.model_params, S0=tmp.model_S0)
```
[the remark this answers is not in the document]
samcoveney: again, this is a problem with not reading the iteration `if else` - I will add some comments, but you can see that `tmp` gets defined in the first iteration of IRLS, so it will always already be defined here.

dipy/reconst/dki.py
```diff
@@ -1895,21 +1939,92 @@ def multi_fit(self, data, *, mask=None):
...
+                        robust[cond] = robust_mask
+                else:
+                    w, robust = weights_method(
+                        data_thres, pred_sig, self.design_matrix,
+                        leverages, rdx, TDX, robust
+                    )
+
+            tmp, extra = self.multi_fit(data_thres, mask=mask,
+                                        weights=w, return_leverages=True)
+            leverages = extra["leverages"]
```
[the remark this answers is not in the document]
samcoveney: I'm not sure I understand the problem

dipy/reconst/dki.py
```diff
@@ -1895,21 +1939,92 @@ def multi_fit(self, data, *, mask=None):
...
+
+        """
+        if num_iter < 2:  # otherwise, weights_method will not be utilized
+            raise ValueError("num_iter must be 2+")
+
+        TDX = num_iter
+        for rdx in range(1, TDX + 1):
+
+            if rdx == 1:
+                w, robust = True, None  # w = True ensures WLS (not OLS)
```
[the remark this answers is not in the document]
samcoveney: No sorry, it should be `True` not `None`. If it's `None` what would it mean? To not use WLS and to use OLS instead? Or to calculate some weights automatically? It's totally ambiguous if it's `None`. I just don't think, it larger scheme of this PR, something like this really matters.

dipy/reconst/dki.py
```diff
@@ -2530,13 +2647,20 @@ def ls_fit_dki(
         Inverse of the design matrix.
     return_S0_hat : bool, optional
         Boolean to return (True) or not (False) the S0 values for the fit.
-    weights : bool, optional
-        Parameter indicating whether weights are used.
+    weights : array ([X, Y, Z, ...], g), optional
```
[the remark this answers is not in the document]
samcoveney: But I don't want the default to be None, I want it to be False, so that when weights is False it will calculate OLS instead of WLS. Is it really necessary to risk breaking this PR just so it can be None instead of False?

dipy/reconst/dti.py
samcoveney: pretty sure this was all caused by rebase issues.... there were simply so many rebases, that I must have lost track. I will put them back

dipy/reconst/weights_method.py
samcoveney: well, it's a boolean array. Are you only saying that the docstring is wrong?

dipy/reconst/dti.py
```diff
@@ -2124,6 +2111,133 @@ def restore_fit_tensor(
...
+
+    extra = {"robust": robust}
+    if return_S0_hat:
+        model_S0.shape = data.shape[:-1] + (1,)
+        return [params, model_S0], extra
+    else:
+        return params, extra
+
+
+# define robust WLS and NLLS functions using GM M-estimators
```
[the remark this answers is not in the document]
samcoveney: it is GMM _now_ and that's what counts. In case you are wondering, I did check with the RESTORE authors.

dipy/reconst/dki.py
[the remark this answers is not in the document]
samcoveney: And more to the point, things are currently working, and I have extensively used this code in all kinds of ways in my own work. While I have written good tests for this PR, if we start changing things for (frankly) aesthetic reasons, things could break in subtle ways that get missed

samcoveney:
Thanks for checking.

Many things, I will clean up.

Some things though, I think at this point just need to not be fiddled with, such as aesthetic concern over types, and aesthetic worries about when things are declared the iterations loops. I emphasize aesthetic because these things literally pose no possible risk to the code as it is, it is not possible for the code to reach places without the correct variables being defined, and there is really no reason not to use mixed types in a few key locations where this makes the code more understandable.

samcoveney:
Last comment: I have a lot of other work to get on with, I will get to this when I can. When are you planning the release?

dipy/reconst/weights_method.py
jhlegarreta: If it's a boolean array, then the docstring should read something like:
```
last_robust : ndarray 
    True for inlier indices and False for outlier indices. Must have the same shape as data.
```

The use of ndarray applies to all array docstrings.

The use of booleans is reaffirmed by this line here using a boolean NumPy function:
https://github.com/dipy/dipy/pull/3358/files#diff-a7d1f52350ccb4a2b7ff4decac1456106cda4471627ec532985e7e52c2cfc2b2R117

Assuming that these will be made true booleans, these blocks:
https://github.com/dipy/dipy/pull/3358/files#diff-a7d1f52350ccb4a2b7ff4decac1456106cda4471627ec532985e7e52c2cfc2b2R119-R120
and
https://github.com/dipy/dipy/pull/3358/files#diff-a7d1f52350ccb4a2b7ff4decac1456106cda4471627ec532985e7e52c2cfc2b2R124-R125

can then be simplified to use true boolean indexing:
```
        w[~robust] = 0.0
        w[robust] = 1.0
```

and
```
        w[~robust] = 0.0
        w[robust] = pred_sig[robust == 1]**2
````

Please add a check on the dimensionality of the array at the beginning of the method.

Thanks for considering this and for all the endeavor and persistence Sam.

File: dipy/reconst/weights_method.py [hunk 33]
```diff
@@ -0,0 +1,248 @@
...
+
+    w[np.isinf(w)] = 0
+    w[np.isnan(w)] = 0
+
+    return w, robust
+
+
+# define specific M-estimators
+def weights_method_wls_gm(*args):
+    """ return weights_method_wls_m_est(*args, m_est="gm", cutoff=3),
```
jhlegarreta, line 222: A few more notes about docstrings applying to the entire PR.
- Remove the whitespace before the beginning of the short description.
- Start the sentences with capitals.
- Use a short description/summary (one line), and then start a new paragraph (blank line between them) with the long description so that the documentation can be rendered correctly.

Thanks.

dipy/reconst/multi_voxel.py
skoudoro: > The problem is, that DiPy wsa not written in a clever enough way to assume that other things may need returning from the fitting functions, in general.

DIPY is 15years old.... back then, we agree with scikit-learn to follow the same logic... and the need was not the same as today.

and yes, something to consider in the future, always looking for the best solution for the next 15-20years...

skoudoro:
> Last comment: I have a lot of other work to get on with, I will get to this when I can. When are you planning the release?

I was hoping this week or next week but for sure before November 1rst.

>  I emphasize aesthetic because these things literally pose no possible risk to the code as it is, it is not possible for the code to reach places without the correct variables being defined, and there is really no reason not to use mixed types in a few key locations where this makes the code more understandable.

I get your point but I think you over estimate the level of users 😅.  the key word is `the correct variables being defined`. By experience, users often DO NOT put the correct variable unfortunately. Also, a unique choice make long term maintenance easier. After working with multiple devs with DIPY or other libraries, I do not think I agree with this sentence: `where this makes the code more understandable`. 

But that's fine, it can be updated later, carefully and with more thinking. I agree, let's not break something that is working for now. You can ignore this part of my review.

samcoveney:
On the matter of the mixed types for the weights argument, I still disagree. The provided weights need to be either:
- an array (use these weights)
- True - calculate W = signal
- False - don't use weights
There really isn't another way to write this in a straightforward manner. If weights is None, what does this mean? I suppose None could mean "don''t use weights"  but then you would still need the option of True.

Taking the philosophy that the understandability of code should be maximized even without comments, I'm not sure I can think of a better way at present. I did think quite hard about this when writing the code, and generalizing the `wls_fit_tensor` function

Anyway, for another time.

It will be difficult for me to hit this deadline tbh, and given that I opened this PR a month ago, can you give me some extra time if required? The ISMRM 2025 abstract deadline is 5 Nov and I have a lot of work to do before now and then.

dipy/reconst/dti.py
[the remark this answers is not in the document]
samcoveney: done

dipy/reconst/dti.py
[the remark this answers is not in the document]
samcoveney: done

dipy/reconst/dti.py
```diff
@@ -1918,13 +1919,13 @@ def restore_fit_tensor(
         Data or response variables holding the data. Note that the last
         dimension should contain the data. It makes no copies of data.
 
-    sigma : float, array of shape [n_directions], array of shape [X, Y, Z]
-        An estimate of the variance. :footcite:t:`Chang2005` recommend to use
+    sigma : float, optional
+        An estimate of the variance. [1]_ recommend to use
```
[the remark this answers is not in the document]
samcoveney: thanks, done, I didn't mean to undo this, must have been during another rebase

dipy/reconst/dti.py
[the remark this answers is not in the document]
samcoveney: I have updated the docstring to be even more explicit

dipy/reconst/weights_method.py
samcoveney: I think I'll just remove these, they seem a bit unecessary

dipy/reconst/dti.py
samcoveney: Sorted, these references were left over anyway, not actually needed.

dipy/reconst/dti.py
samcoveney: sorted, added Chang2012 to the references as well

dipy/reconst/dti.py
samcoveney: Okay, I will change for the stuff specific to my PR

dipy/reconst/multi_voxel.py
samcoveney: Happy to help in the future if needs be. It is sort of nice that it's separate from the Fit object, in the sense that it contains info on the fitting but not the results... I find that this isn't a terrible solution. It makes it easy to have the fitting potential return `extra` while leaving the fit objects alone.

dipy/reconst/weights_method.py
samcoveney: Gotcha, this makes sense, I'll make the changes.

dipy/reconst/weights_method.py
samcoveney: replaced "*" with "x" for multiplication, looks better anyway, thanks for the spot

dipy/reconst/weights_method.py
samcoveney: the choices are in the docstring

samcoveney:
> I think the answer to my latter question is Collier, Q., Veraart, J., Jeurissen, B., den Dekker, A., Sijbers, J., 2015. Iterative reweighted linear least squares for accurate, fast, and robust estimation of diffusion magnetic resonance parameters: IRLLS for Estimation of Diffusion MR Parameters. Magnetic Resonance in Medicine 73, 2174-2184. Correct?

I have added this reference into docs/references.bib thanks for the spot

samcoveney:
I think I've addressed everything, except for any pep8 issues I've accidentally made in the process of fixing. I will check tomorrow.

One other thing: I need to finish off a suite of tests for the functions in `weights_method.py` as suggested by @arokem, but I've done as much as I can today so I am committing as is (since I know you're aiming at a release date of 01 Nov). I'll try to fix up everything else tomorrow, including finishing these tests.

I am aware that I still need to write some tutorials for this, but this is something I am committed to doing later. I want these tutorials to exist, to explain how to use my added functionality, so it will definitely get done...

arokem:
Thanks! FWIW, I think that implementing the tutorial on a separate PR would be fine.

samcoveney:
Some of these tests are failing here, because they are complaining about the test `test_nnls_jacobian_func` specifically that `nlls.err_func` should return a scaler variable. For DiPy, due to `scip.opt.leastsq`, we do NOT want `nlls.err_func` to return a scalar variable, we need to return an array of weighted squared residuals, all fine. But more confusingly, is that as far as I can see in the tests `test_nnls_jacobian_func` where we call `nlls.err_func`, we are in fact returning a scalar variable. No idea what's going on here, and why it doesn't fail for me, and only fails on some tests here.

Since we have issues with testing the gradient anyway, see #3155 , I would suggest to switch off this test for now by commenting out that line `approx = opt.approx_fprime(D, nlls.err_func, 1e-8, *args)`, since we are not currently going on to use the result anyway.

skoudoro:
Thank you for pushing this @samcoveney. 

I will be able to look tomorrow morning. it would be great if @arokem and @jhlegarreta could look at this PR also tomorrow. 

Thank you all

jhlegarreta:
code formatting (introduction paragraph):
https://github.com/dipy/dipy/blob/master/doc/devel/coding_style_guideline.rst#coding-style

`pre-commit run` if you want to run the formatter as a standalone command.

skoudoro:
Thank you for finalizing this @samcoveney.

Overall, I am ok with this PR. Multiple follow up PRs will be needed. 
- Tutorial on a new PR
- parallel backend currently not available. We need a specific PR to focus on this.
- Check all the FIXME tagged by @samcoveney (thank you for this).

After fixing the pep8 issues, I will be +1 for merging this ASAP. To fix those:

```
pip install pre-commit
pre-commit install
pre-commit run --files  my_files.py
```

We have disabled pep8speak to avoid conflicting message with ruff. Can you fix those issues tagged by @jhlegarreta ?

@arokem, any objection ?

skoudoro:
@arokem and @samcoveney ? When do you think you will have time to finalize this? 

I am waiting for this PR to finalize the release. if not possible I can push to next release which I plan in end of december/beginning of january.

Please let me know.

Thanks

arokem:
All good on my end! I'll make a formatting PR right now, so we can move ahead.

arokem:
Here it is with ruff formatting (and elimination of several unused variables in tests): #3401

dipy/reconst/weights_method.py
```diff
@@ -0,0 +1,276 @@
...
+
+    # handle potential zeros
+    pred_sig[pred_sig < MIN_POSITIVE_SIGNAL] = MIN_POSITIVE_SIGNAL
+
+    # calculate quantities needed for C and w
+    log_pred_sig = np.log(pred_sig)
+    residuals = data - pred_sig
+    log_data = np.log(data)
+    log_residuals = log_data - log_pred_sig
+    z = pred_sig * log_residuals
```
arokem, line 249: Just flagging that ruff has identified this as a variable that is not being used in the code. Sorry for not noticing this earlier, but seems like that might point to something that doesn't make sense?

skoudoro: Not sure why ruff say that because it seems to be used the line just after. Can it be confuse by the one letter variable ?

arokem: Wait, which line? I don't see `z` being used on line 251. It is used on line 150 in similar context, so maybe it was erroneously copy-pasted here?

skoudoro: ok I was confused with the line 148-151. You are right, we need a clarification here.

arokem:
For now, I have removed it in #3401

dipy/reconst/weights_method.py
samcoveney: There is a function where this is no longer used, it seems. It's used in the WLS weight function, but not in the NLLS weight function. It was copy-pasted _on purpose_ but left in by accident.

skoudoro:
superseded by #3403 which has been merged, So closing 

Thank you for this work @samcoveney

PR #3363, opened 2024-09-22
Avoid SyntaxWarnings due to embedded LaTeX

mr-c:


mr-c:
Pardon me, I see this was handled by explicitly escaping the `\` instead of using raw strings.

PR #3366, opened 2024-09-26
DOC: Make `rng` optional parameter docstrings consistent

jhlegarreta:
Make `rng` optional parameter docstrings consistent:
- Improve the description of the parameter and the behavior in the default case.
- Remove the default value from the parameter docstring.
- Remove the mention to "class" in the parameter type.
- Add the parameter docstring to the methods that were missing it.


skoudoro (approved):
All good, Thanks @jhlegarreta, merging

PR #3372, opened 2024-10-09
BugFix: New Atlas OMM not working with Horizon

maharshi-gor:
This PR aims to support the new Atlas (Images with intensities from 0 to 0.9).

**Underlying issue:**
The slice visualizer in Horizon supported the intensity percentile from (2, 98), due to underlying errors from earlier versions of FURY when trying to move the slider to min or max values.

**Solution:**
The current minimum required version of the FURY does not have that issue. So, updating the percentile range will open the image properly.

**Additional Changes**
Fixed a method call where the `visibility` argument was not passed as a keyword argument. Resulting in user warnings.

### DIFF
--- a/dipy/viz/horizon/app.py
+++ b/dipy/viz/horizon/app.py
@@ -545,7 +545,7 @@ class Horizon:
                         data,
                         affine=affine,
                         world_coords=self.world_coords,
-                        rgb=self.rgb
+                        rgb=self.rgb,
                     )
                     self.__tabs.append(
                         SlicesTab(
--- a/dipy/viz/horizon/app.py
+++ b/dipy/viz/horizon/app.py
@@ -545,8 +545,7 @@ class Horizon:
                         data,
                         affine=affine,
                         world_coords=self.world_coords,
-                        rgb=self.rgb,
-                        is_binary=binary_image,
+                        rgb=self.rgb
                     )
                     self.__tabs.append(
                         SlicesTab(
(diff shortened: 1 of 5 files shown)


skoudoro (commented):
Hi @maharshi-gor,

Do we still need the function `is_binary_image`. It seems that more code needs to be clean up. thank you for the feedback

maharshi-gor:
> Hi @maharshi-gor,
> 
> Do we still need the function `is_binary_image`. It seems that more code needs to be clean up. thank you for the feedback

We need the function as it is the sanitary check on the rois.
If someone is using a non-binary image with `--rois_image` option in the command line or pass it to python method in ROIs. The current system will fail without the check on this line: https://github.com/dipy/dipy/blob/d30b75657952dd7f29bb75709c0af307578934a2/dipy/viz/horizon/app.py#L533

Let me know if you need further clarification.

skoudoro:
> If someone is using a non-binary image with --rois_image option in the command line

following here the function `contour_from_roi`, the data is binarized in fury and does not need `is_binary` function. Check here https://github.com/fury-gl/fury/blob/a09d2ce2021d61c9c53b3311833200177dccdd62/fury/actor.py#L442

So still not sure if `is_binary_image`is needed

maharshi-gor:
> > If someone is using a non-binary image with --rois_image option in the command line
> 
> following the function `contour_from_roi` in FURY, the data is binarized in fury and does not need `is_binary` function. Check here https://github.com/fury-gl/fury/blob/a09d2ce2021d61c9c53b3311833200177dccdd62/fury/actor.py#L442
> 
> So still not sure if `is_binary_image`is needed

Understood. I was not aware of this functionality. Let me do the clean-up and update the PR.

skoudoro (approved):
Thank you for cleaning @maharshi-gor. Merging

Issue #3347, opened 2024-08-31
Reconstruction model list not linked in documentation since it cannot be located


jhlegarreta:
## Description

The table of reconstruction methods located in
https://github.com/dipy/dipy/blob/master/doc/reconstruction_models_list.rst

cannot be located properly when generating the documentation:

```
/home/runner/work/dipy/dipy/doc/examples_built/reconstruction/index.rst:13: CRITICAL: Problems with "include" directive path:
InputError: [Errno 2] No such file or directory: 'examples_built/reconstruction_models_list.rst'.
```

https://github.com/dipy/dipy/actions/runs/10647467841/job/29515431777#step:5:698

The script expects it to be in `examples_built` but it is not found there; it is originally located in `examples`.

Probably the `prepare_gallery.py` script should copy the file to the `examples_built` folder.

## Way to reproduce
- Clone DIPY.
- Follow the installation in https://docs.dipy.org/stable/devel/installation_from_source.html#getting-the-source to install DIPY
- Generate the documentation following https://docs.dipy.org/stable/devel/installation_from_source.html#documentation.
- Check the error.

skoudoro:
or maybe we can do a `..include:  :ref:{reconstruction_models_list_name}` 

something to check. Thank you for the report

skoudoro:
> or maybe we can do a ..include:  :ref:{reconstruction_models_list_name}

not possible, I checked

jhlegarreta:
Thanks for having given this a try. @skoudoro So is `cp examples/reconstruction_models_list.rst examples_built/reconstruction_models_list.rst' the only option left? Do we need to copy this into each of the conditional possibilities in `docs/make.bat`?

jhlegarreta:
I think I tried copying the file to `examples_built` manually prior to generation and that did not make the trick, so not sure how to fix this.

jhlegarreta:
So the table appears in
https://docs.dipy.org/stable/examples_built/index.html
but not in
https://docs.dipy.org/stable/examples_built/reconstruction/index.html

I assume that this issue is related, or that we'd like the table appear in the `Reconstruction` section index.

I'll look into improving the references to adopt the new bibtex convention once we manage to fix this issue and as time permits. A slightly related thought would be to have all bibtex references (including those that exist across the docstrings) listed in a section of the website.

Issue #3270, opened 2024-06-23
Use numpydoc check


skoudoro:
See this comment for additional information: https://github.com/dipy/dipy/pull/3250#issue-2353365028

skoudoro:
I experimented it with the following in the `conf.py`
```
error_ignores = {
    # These we do not live by:
    "GL01",  # Docstring should start in the line immediately after the quotes
    "EX01",  # No examples section found
    "EX02",  # examples failed (we test them separately)
    "ES01",  # no extended summary
    "SA01",  # no see also
    "YD01",  # no yields section
    "SA04",  # no description in See Also
    "PR04",  # Parameter "shape (n_channels" has no type
    "RT02",  # The first line of the Returns section should contain only the type, unless multiple values are being returned  # noqa
    # XXX should also verify that | is used rather than , to separate params
    # XXX should maybe also restore the parameter-desc-length < 800 char check
}

numpydoc_validate = True
numpydoc_validation_checks = {"all"} | set(error_ignores)
```

and I get this:

<img width="445" alt="image" src="https://github.com/user-attachments/assets/70657617-e4b4-403e-87d6-186a27c143d9">


Most of them are easy fixes and:
- it will improve consistency a lot
- Avoid this task during review

mrinalchaturvedi27:
Hi @skoudoro! I am interested in doing this issue. Could you please assign me.

skoudoro:
Hi @mrinalchaturvedi27,

You can go ahead

aman-coder03:
Hi @skoudoro 
I would like to help with this issue by fixing `numpydoc` validation warnings and standardizing docstrings across modules.
I have already started looking into docstring inconsistencies and would be happy to contribute fixes following `numpydoc` conventions. Please let me know if there are specific modules or validation errors I should prioritize

Sunilkakde17:
Hello @skoudoro 
I would like to work on this issue as my first contribution to DIPY.

I am currently setting up the project locally and would appreciate any guidance if there are specific requirements for this task.

Thank you!

skoudoro:
ok, no problem @Sunilkakde17, go ahead. 

you can not do it all in once so focus on 1 doc issue only.

jofreyn7:
Hi @skoudoro, I am Jofrey Nasson, a GSoC 2026 contributor applicant. I would like to contribute to this issue by fixing numpydoc warnings in one module. Could you please suggest a specific file or module I should focus on? Thank you.

[...]

PR #3373, opened 2024-10-11
DOC: Consider warnings as errors in documentation CI build

jhlegarreta:
Consider warnings as errors in documentation CI build.

### DIFF
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -36,4 +36,4 @@ jobs:
     - name: Build documentation
       run: |
         cd doc
-        make -C . html-no-examples
+        make -C . html-no-examples SPHINXOPTS="-W --keep-going"


jhlegarreta:
2 of the `Build docs` warnings should be fixed by merging #3374 and #3375, and we will be left with issue #3347.

jhlegarreta:
I'll rebase on `master` once #3379 gets merged. Builds should succeed without warnings after that.

skoudoro:
> I'll rebase on master once https://github.com/dipy/dipy/pull/3379 gets merged. Builds should succeed without warnings after that.

done 👍 -> can be rebased

jhlegarreta:
Documentation builds without warnings. Beautiful. Ready to be merged.

skoudoro (approved):
Great! Thank you for all this hard work @jhlegarreta! merging

Now, next step is #3270 with numpydoc (for later or a GSoC task). 

I experimented it with the following in the `conf.py`
```
error_ignores = {
    # These we do not live by:
    "GL01",  # Docstring should start in the line immediately after the quotes
    "EX01",  # No examples section found
    "EX02",  # examples failed (we test them separately)
    "ES01",  # no extended summary
    "SA01",  # no see also
    "YD01",  # no yields section
    "SA04",  # no description in See Also
    "PR04",  # Parameter "shape (n_channels" has no type
    "RT02",  # The first line of the Returns section should contain only the type, unless multiple values are being returned  # noqa
    # XXX should also verify that | is used rather than , to separate params
    # XXX should maybe also restore the parameter-desc-length < 800 char check
}

numpydoc_validate = True
numpydoc_validation_checks = {"all"} | set(error_ignores)
```

and I get this:

image

jhlegarreta:
@skoudoro thanks for unlocking the last bit to make this possible.

As for

> and I get this:

:pensive:

I need a break from docstring tasks.

Having a build that reports these 3790 warnings would already be good so that we can see them openly and such that any members of the GRG or the community can fix them as time permits. GSoD might be good to address these as well.

A separate step or build would be required to report these warnings so that we do not take steps back in considering doc warnings as errors.

jhlegarreta:
Additionally, it may be useful to have a preview of the documentation builds for each PR, as it would enable seeing how the docs are rendered and would enable spotting any formatting issue early with any new code documentation.

PR #3374, opened 2024-10-11
DOC: Remove `tracking_introduction_eudx` from quick start

jhlegarreta:
Remove `tracking_introduction_eudx` from the quick start guide:
- The file is already included in the tracking group.
- Directly adding it to the start guide with no other context may be counterproductive, as it skips many of the other steps needed before streamline tracking.
- Although tracking is a very unique capability of DIPY, DIPY has many more capabilities that are not highlighted (and may go unnoticed) if we only add the tracking to the quick start guide: the quick start guide would be more useful if it contained some documentation about the workflow for streamline tracking or how each module is related to each other for the purposes of dMRI analysis.
- It eases the documentation configuration as we do not need additional tinkering to avoid the label duplication warning.

Fixes:
```
/dipy/doc/examples_built/quick_start/tracking_introduction_eudx.rst:25:
 WARNING: duplicate label intro_basic_tracking, other instance in
/dipy/doc/examples_built/fiber_tracking/tracking_introduction_eudx_1.rst
```

raised for example at:
https://github.com/dipy/dipy/actions/runs/11284383070/job/31385471672?pr=3373#step:5:6836

### DIFF
--- a/doc/examples/_valid_examples.toml
+++ b/doc/examples/_valid_examples.toml
@@ -40,7 +40,6 @@ position = 1
 enable = true
 files = [
     "quick_start.py",
-    "tracking_introduction_eudx.py",
 ]
 
 


skoudoro (changes requested):
Hi @jhlegarreta,

I am sorry but I disagree with this PR. In general, we should be able to add some tutorial in multiple section and this PR does not fix this initial issue. 

Here, we should focus on the script that convert python file to rst, and make sure to attribute a unique label.

jhlegarreta:
Let me know if you have a clear path on how to do that.

skoudoro:
> Let me know if you have a clear path on how to do that.

The best way is to remove the label that you see here: https://github.com/dipy/dipy/blob/54b75dfcb92a7bb413345a6c302da85f879b7eee/doc/examples/tracking_introduction_eudx.py#L3

1. this label is never used
2. sphinx gallery create a new and unique label to use for each file/tutorials

So this label is useless. Actually, all labels in our tutorials are useless and need to be removed. I think, most of them have been removed but some are remaining

jhlegarreta:
Looks like some of them are indeed used #3379:
https://github.com/dipy/dipy/actions/runs/11305245523/job/31444539398?pr=3379#step:5:738

But `_intro_basic_tracking` is not. Do we look for a better solution or are we happy with removing the unused ones for now?

jhlegarreta:
Closing in favor of #3379.

PR #3375, opened 2024-10-11
DOC: Reference footnote in `streamline_tools`

jhlegarreta:
Reference footnote in `streamline_tools`.

Fixes:
```
/dipy/doc/examples_built/streamline_analysis/streamline_tools.rst:336:
 WARNING: Footnote [#] is not referenced. [ref.footnote]
```

raised for example at:
https://github.com/dipy/dipy/actions/runs/11284383070/job/31385471672?pr=3373#step:5:686

PR #3376, opened 2024-10-11
DOC: Use placeholder for unused variable in `streamline_tools`

jhlegarreta:
Use `_` as a placeholder for the unused `affine` variable in `streamline_tools` example when loading the Stanford dataset HARDI data. This reduces confusion as a few lines later the same variable is set to the identity matrix.

Follow-up to commit e6f5ab0.

### DIFF
--- a/doc/examples/streamline_tools.py
+++ b/doc/examples/streamline_tools.py
@@ -48,7 +48,7 @@ hardi_fname, hardi_bval_fname, hardi_bvec_fname = get_fnames(name="stanford_hard
 label_fname = get_fnames(name="stanford_labels")
 t1_fname = get_fnames(name="stanford_t1")
 
-data, affine, hardi_img = load_nifti(hardi_fname, return_img=True)
+data, _, hardi_img = load_nifti(hardi_fname, return_img=True)
 labels = load_nifti_data(label_fname)
 t1_data = load_nifti_data(t1_fname)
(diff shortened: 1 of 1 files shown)


doc/examples/streamline_tools.py
```diff
@@ -48,7 +48,7 @@
 label_fname = get_fnames(name="stanford_labels")
 t1_fname = get_fnames(name="stanford_t1")
 
-data, affine, hardi_img = load_nifti(hardi_fname, return_img=True)
+data, _, hardi_img = load_nifti(hardi_fname, return_img=True)
```
skoudoro, line 51: LGTM, but isnot it strange that we do not use the affine in this tutorial? why we use `np.eye` instead ?

skoudoro (commented):
Hi @jhlegarreta,

Thank you for this, see below my comment

doc/examples/streamline_tools.py
jhlegarreta: It is Serge. The only thing I investigated is whether the dataset affine is the identity matrix, which is not. I agree that the issue deserves further investigation.

jhlegarreta: If we end up finding that this is a bug and that this very same affine needs to be used, we'll need to move and adapt the text about the affine added in #3375.

skoudoro:
merging this, since it make sense with what we have. However, we might have to review this tutorial. 
In general, we need to review all tutorials, most of them needs an update. Thanks @jhlegarreta

PR #3380, opened 2024-10-14
ensure all calls to a python executable are to `python3`

mr-c:
This patch is from the Debian package for `dipy`.


skoudoro (approved):
LGTM, thanks @mr-c, merging

PR #3391, opened 2024-10-27
STYLE: removing pep8speaks conf file in favor of pre-commit action

skoudoro:
All in the title.

This PR removes pep8speaks which is redundant with pre-commit + ruff and many rules conflicting also.

it avoids confusion


arokem:
Yeah - good call!

jhlegarreta (approved):

PR #3392, opened 2024-10-27
[WIP] NF: Adding pytorch versions

pjsjongsung:
This PR is to create pytorch versions of our tensorflow model here, and possibly in the future delete tensorflow from requirements. The PR creates a new directory dipy.nn.torch, which has the torch versions of evac and histo_resdnn. Necessary changes were made in the tensorflow model code as well, to avoid confusion.

PR #3395, opened 2024-10-31
RF: Refactor `nn` module and deprecate `tensorflow` module

skoudoro:
This PR is a follow up of #3392. it manages Tensorflow and Pytorch backend from `nn`module. 

Pytorch becomes the preferred backend and Tensorflow a deprecated backend.


pjsjongsung:
I'm guessing this let's us do `dipy.nn.evac.EVACPlus` instead of `dipy.nn.torch.evac.EVACPlus` now. I think you want to have a similar system in `histo_resdnn_torch_test.py` I believe?

Also I am not sure how modules work in python but I'm assuming, as the description says, it will be called before and after the tests in the file?

skoudoro:
> I think you want to have a similar system in histo_resdnn_torch_test.py I believe?

it is the same system

> I'm assuming, as the description says, it will be called before and after the tests in the file?

not sure to understand your point here

skoudoro:
> I'm guessing this let's us do dipy.nn.evac.EVACPlus instead of dipy.nn.torch.evac.EVACPlus

yes, both are possible, this is for backward compatibility

by default, this is "torch" backend but it can be modified by using an environment variable `DIPY_NN_BACKEND`

pjsjongsung:
> > I think you want to have a similar system in histo_resdnn_torch_test.py I believe?
> 
> it is the same system
> 
> > I'm assuming, as the description says, it will be called before and after the tests in the file?
> 
> not sure to understand your point here

I meant like I do not see where the modules are called. Is it something that is automatically called by pytest?
Also I mentioned `histo_resdnn_torch_test.py` because I see it is not in the list of files changed.

skoudoro:
Yes, `pytest` import automatically all modules and use the default backend which is pytorch. 
I need to reload the import when I want to use Tensorflow.  Otherwise, we will need to run twice `pytest`, one for each backend

skoudoro:
is there any additional comments concerning this PR?

Garyfallidis:
Pytorch wins.

Garyfallidis:
We will need to run more benchmarks for the next stages. But good job for this release. Tensorflow to be deprecated ...

PR #3399, opened 2024-11-03
STYLE: Remove empty quoted paragraph in developer guide index

jhlegarreta:
Remove empty quoted paragraph in developer guide index: remove the whitespace before the document label.


skoudoro (approved):
Good catch! merging, thanks @jhlegarreta

PR #3400, opened 2024-11-04
CI: Add python 3.13

skoudoro:
This PR add python.3.13 to our CI's list.

Currently checking the freed-threaded version that's why it is in WIP.

ps: Python 3.9 will be removed after the DIPY 1.10.0 release in a new PR

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -38,7 +38,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        cibw_python: ["cp39-*", "cp310-*", "cp311-*", "cp312-*"]
+        cibw_python: ["cp310-*", "cp311-*", "cp312-*", "cp313-*"]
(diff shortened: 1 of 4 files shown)


skoudoro:
Some of our dependencies (like `cvxpy`) are not ready for python 3.13.

I will let this open and check next month if there is any update

skoudoro:
ok, python 3.13 seems to be ok for linux and mac. 

it fails on Windows, I need to ask the cvxpy people or scs particularly.

python 3.13 free-threaded is failing but it is normal. it needs much more work, and our dependencies compatible with python3.13t are not released yet. So Pypi is not enough.  I will remove the 3.13 free-threaded naive approach and it will be on a specific PR later on.

skoudoro:
> it fails on Windows, I need to ask the cvxpy people or scs particularly.

ok, cool, SCS created a wheel for windows. So all good, I need to clean up this PR and it will be ready to go.

PR #3404, opened 2024-11-05
RF: Improve and Simplify version management

### LINKED ISSUE #3115: Fix `get_info` for release package
It seems that `dipy.get_info` fails with the release but not with a dev version. here the error:

```python
OSError: Missing commit info file /home/kostas/anaconda3/envs/dipy_env/lib/python3.11/site-packages/dipy/COMMIT_INFO.txt
```
We need to look at commit number management

skoudoro:
The goal of this PR is to simplify and improve DIPY versionning during release process.

it also fix #3115. a small bug introduced previously during meson migration. Unnecessary files has been removed also


arokem:
This looks fine, but wondering whether you'd like to use https://github.com/pypa/setuptools-scm in the long run?

skoudoro:
I would like to avoid mixing meson and setuptools-scm. 

Also, I saw a ton of controversy about setuptools-scm during the meson migration. That's why I prefer to avoid it for projects with C extension.  For pure python package, I agree, this is the way to go I believe.

arokem:
OK - LGTM. Codecov failure is spurious.

skoudoro:
Thank you 👍

PR #3408, opened 2024-11-06
RF: fix step argument in DTI NLLS model

### LINKED ISSUE #3030: I do not see a way to change step as used by reconst.dti.TensorModel.fit()
## Description
DTI fitting seems to have gotten slower between dipy 1.6.0 and 1.8.0. Raising the value of the optional step parameter in iter_fit_tensor() is suggested, e.g. in https://docs.dipy.org/stable/reference/dipy.reconst.html#dipy.reconst.dti.TensorModel , but if I change
```
dtm = dti.TensorModel(gtab, fit_method=fit_meth, **fit_kwargs)
dt = dtm.fit(data, mask)
```
to 
```
dtm = dti.TensorModel(gtab, fit_method=fit_meth, step=1e5, **fit_kwargs)
dt = dtm.fit(data, mask)
```
I get `TypeError: nlls_fit_tensor() got an unexpected keyword argument 'step'`.

Stepping through the code, it appears that specifying step in either `TensorModel.__init__ ` or  iter_fit_tensor() would be fine, but the intermediate layers to abstract the kind of fit method do not accept the step parameter. nlls_fit_tensor() even calls ols_fit_tensor() internally without the step parameter, so that would need to be caught too.

I understand how step tries to find a happy medium between voxel-by-voxel looping and whole-image-out-of-memory problems, but what is the right target for the amount of memory? The CPU RAM or the GPU memory?

Am I missing something?

Thanks

skoudoro:
this PR fixes #3030

It warns user that `step` can not be used for NLLS.

This is a problem found by @captainnova. Can you have a quick look into it. Thanks!

### DIFF
--- a/dipy/reconst/dti.py
+++ b/dipy/reconst/dti.py
@@ -756,6 +756,14 @@ class TensorModel(ReconstModel):
 
         if not callable(fit_method):
             try:
+                if fit_method.upper() in ["NLS", "NLLS"] and "step" in kwargs:
+                    _ = kwargs.pop("step")
+                    warnings.warn(
+                        "The 'step' parameter can not be used in the "
+                        f"{fit_method.upper()} method. It will be ignored.",
+                        UserWarning,
+                        stacklevel=2,
+                    )
                 fit_method = common_fit_methods[fit_method.upper()]
             except KeyError as e:
                 e_s = '"' + str(fit_method) + '" is not a known fit '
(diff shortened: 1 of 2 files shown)


arokem:
LGTM, CI failures unrelated.

captainnova:
> This is a problem found by @captainnova. Can you have a quick look into it. Thanks!

Hi @skoudoro, I looked at the commit but haven't yet run it and am not sure when I will get to it. The code looks perfect. I admit I was a bit intimidated by the request for a test case, and didn't realize it would be that simple (if I'd known how).

Thanks!

PR #3409, opened 2024-11-06
[RF] Cleanup deprecated functions and remove some old legacy scripts

skoudoro:
Quick PR to cleanup the codebase from
- deprecated function
- legacy script not used anymore


skoudoro:
This PR is ready to go! Thanks in advance @arokem or @jhlegarreta !

jhlegarreta (approved):
As always, I would very much appreciate commit messages with a body in them where an explanation is given, even if the rationale of the changes is very clear as in this case.

But do as you feel @skoudoro.

skoudoro:
> As always, I would very much appreciate commit messages with a body in them where an explanation is given, even if the rationale of the changes is very clear as in this case.


Not sure to understand, can you provide an example

jhlegarreta:
> Not sure to understand, can you provide an example

I've understood the changes, but when one reviews the history without the file diffs it is hard to know what the changes are with a one-liner message. So if the message is not self-contained the reviewer/maintainer has to go back to the diff, and that takes time from them.

I keep reminding ourselves or projects that I am involved in that self-contained messages should be self-contained. This goes a long way when reviewing changes. But it's fine, I know it is time-consuming.

skoudoro:
>  I know it is time-consuming.

Yes, it is, but well-noted and make sense. I will do my best for next time. For now, I would like to move forward with this. 

Thank you for reminding me those important principle

PR #3417, opened 2024-11-28
MAINT: Address changes to API with new (1.6) version of cvxpy.

arokem:


skoudoro:
Hi @arokem,

Thank you for this. Can you say more before merging this? What is sparsity and what happens if we have an old version of CVXPY ? Should we update our requirement to be safe ?

arokem:
Yes, I think that we should require the newest version of CVXPY to be on the safe side. 

I am not sure what this parameter was, to be honest, but looks like the tests pass without it so 🤷

dipy/core/optimize.py
```diff
@@ -440,10 +440,8 @@ def __init__(self, m, *, A=None, L=None):
         t = len(A) if A else 0
         k = t - m - 1
 
-        sparsity = [(i, j) for i in range(m) for j in range(i, m)]
```
skoudoro, line 443:
```python
if Version(cvxpy.__version__) < Version('1.6'):
    sparsity = [(i, j) for i in range(m) for j in range(i, m)]
else:
    sparsity =  [range(m), range(m)]
```

dipy/core/optimize.py
```diff
@@ -440,10 +440,8 @@ def __init__(self, m, *, A=None, L=None):
...
-
         # Unknowns
-        self._X = cvxpy.Parameter((m, m), sparsity=sparsity)  # Design matrix
```
skoudoro, line 446: to keep like before

skoudoro (changes requested):
Back to DIPY 😅, 

So I checked, and sparsity just changed shape.
instead of `sparsity=[(0, 0), (0,1), (0,2), ...]` it is `sparsity=[(0,0,0, ...), (0, 1, 2, ...)]`

it would be great to keep it because it impact performance on some model. We just to check `cvxpy` version and plug in the right shape.  See below.

arokem:
Thanks for looking into this. Changes made in 72897943a

dipy/core/optimize.py
```diff
@@ -1,6 +1,7 @@
 """A unified interface for performing and debugging optimization problems."""
 
 import abc
+from distutils.version import Version
```
skoudoro, line 4:
```suggestion
from packaging.version import Version
```

arokem:
Seems like these recent changes are now making the tests fail. Are you sure my original proposal was incorrect? The tests all passed with that change.

skoudoro:
> Seems like these recent changes are now making the tests fail. Are you sure my original proposal was incorrect? The tests all passed with that change.

It was not incorrect, just making thing slower.  Let me go a bit deeper into that issue and if no easy solution we revert back to what you did. Sorry about that.

arokem:
Oh gotcha. Yeah, worth a bit more digging...

skoudoro:
ok, I think it is due to [`load_sdp_constraints`](https://github.com/dipy/dipy/blob/927246108f6aa67283003a46b63f2deeced9becb/dipy/data/__init__.py#L410) but I am not 100% sure.

I did a bit more digging but I did not find an easy solution. I do not want to spend more time on this.

Furthermore, this feature was available but not active/used in reality following this issue https://github.com/cvxpy/cvxpy/issues/2038.

So, we can safely remove it since it was already ignored.

Sorry for all of this @arokem. you can get back to what you did initially, remove sparsity, then we can merge.

thanks!

skoudoro:
ps: just remove the 2 last commits

arokem:
OK - done! Also rebased on master for good measure.

PR #3423, opened 2024-12-15
Make the docs more reproducible

mr-c:
When building the docs and determining the current year for the copyright entry: If the environment variable [`$SOURCE_DATE_EPOCH`](https://reproducible-builds.org/docs/source-date-epoch/) is defined, use that instead of `time.time()`.

Without this patch, building the dipy docs using a source release from a prior year would result in different output;  hampering reproducible builds.

This issue was discovered by the Debian reproducibility test for dipy: https://tests.reproducible-builds.org/debian/rb-pkg/unstable/amd64/diffoscope-results/dipy.html (there are other reproducibility issues, but they are `cython`'s issues, not the fault of dipy)

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -9,10 +9,11 @@
 # All configuration values have a default; values that are commented out
 # serve to show the default.
 
-from datetime import datetime
+import datetime
 import os
 import re
 import sys
+import time
 
 # Doc generation depends on being able to import dipy
 try:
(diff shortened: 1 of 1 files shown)


skoudoro (approved):
Thank you for this @mr-c.

just learned something about `SOURCE_DATE_EPOCH`. 

merging

PR #3424, opened 2024-12-16
NF: Automate requirement files

skoudoro:
Problem: the codebase had multiple requirement files not maintained. 

Goal: Centralized all the requirements in `pyproject.toml` and automate the creation of requirement files  via a script and precommit.

This PR is achieving this goal to simplify the maintenance of requirement files.


skoudoro:
this PR is ready to go. Is there any additional comment?

arokem:
This is cool. A question I have is whether the auto-generated files should even be committed to the repo. Is there some way to trigger `generate_requirements` when the files are needed? (i.e., at build time?)

skoudoro:
Are they really needed at build time ? 

I see your point. This need some discussion.

In fact, only `build.txt`  and `pyproject.toml` are needed when you use Pypi because you can do `pip install "dipy[doc]"`.

This command is not possible when you use conda if I am not wrong, so the requirements files help a lot. 

Maybe we can provide just some guidance to the user via the doc on how to use `generate_requirements.py` with conda.

Not sure what is the best strategy

arokem:
After some discussion with @skoudoro, I believe that the answer to my question above is "no". The auto-generated files need to be tracked by version control. The analogy I have in mind is that ruff edits files as part of pre-commit hook, and the files get committed into git. This system does something similar, but only for requirements files. The nice thing is that it automates what is otherwise a manual and error prone process. So, I am 👍 for merging this (once conflicts are resolved).

skoudoro:
This PR is ready to go @arokem . Failing CI's are not related. 

Thanks

PR #3426, opened 2024-12-17
RF: from relative import to absolute import

skoudoro:
Small PR to fix a relative import.  Our codebase use only absolute import so it was surprising to see that.

I use the opportunity to update the doc string also in this specific file.


arokem:
Failures are unrelated and this is pretty uncontroversial, so I will go ahead and merge.

PR #3427, opened 2024-12-18
NF: add dipy_math workflow

skoudoro:
this PR adds a new CLI named `dipy_math` which allows to do basic operations on volume (nifti images).

suppose you have 1, 2 or 3 nifti volumes in a folder, you could do:
- `dipy_math "vol1 + vol2" nifti_a.nii.gz nifti_b.nii.gz`
- `dipy_math "vol1 - vol2 * vol3" nifti_a.nii.gz nifti_b.nii.gz nifti_c.nii.gz`
- `dipy_math "vol1 / vol2" nifti_a.nii.gz nifti_b.nii.gz -o my_result.nii.gz` 
- `dipy_math  "sin(vol1) + arcsinh(vol2/vol3)" nifti_a.nii.gz nifti_b.nii.gz nifti_c.nii.gz -o my_result.nii.gz`
- etc.


maharshi-gor (commented):
Hey @skoudoro , cool feature!

I have tested this on my local, LGTM! 

Let's add few more corner cases in test to match our criteria for test in the PR.

Thanks!

skoudoro:
> Let's add few more corner cases in test to match our criteria for test in the PR.

Can you be more explicit @maharshi-gor? example of a corner case ?

arokem:
Interesting! I would like to see a bit more error handling. Like when inputs are not matched in terms of their shape, I think it would be good to raise an informative error. I think also the affine needs to be identical, right? Otherwise how would you decide which affine to use?

skoudoro:
> I would like to see a bit more error handling. Like when inputs are not matched in terms of their shape, I think it would be good to raise an informative error. I think also the affine needs to be identical, right? Otherwise how would you decide which affine to use?

ok, I see. I will add that.

Thank you both for the feedback.

maharshi-gor:
> > Let's add few more corner cases in test to match our criteria for test in the PR.
> 
> Can you be more explicit @maharshi-gor? example of a corner case ?

Please check the codecov report it is missing some of the branch coverage, like a few if conditions. For example `dtype` needs to be passed to cover the `if dtype:` line in the coverage.

skoudoro:
> Please check the codecov report it is missing some of the branch coverage, like a few if conditions. For example dtype needs to be passed to cover the if dtype: line in the coverage.

I see, thanks for the feedback @maharshi-gor

skoudoro:
Any additional comment @arokem or @maharshi-gor ?

I think this PR is ready to go and address your last review

dipy/workflows/io.py
```diff
@@ -775,3 +778,125 @@ def run(
...
+            info_msg += f"{fname}:\n- vol index: {i}\n- shape: {data.shape}"
+            info_msg += f"\n- affine:\n{affine}\n"
+            if ref_affine is None:
+                ref_affine = affine
+                ref_shape = data.shape
+                continue
+
+            have_errors = (
+                have_errors
+                or not np.array_equal(ref_affine, affine)
```
arokem, line 869: Do you want to allow for optional tolerance input here in case of affines that differ on a very small number?

dipy/workflows/tests/test_io.py
```diff
@@ -342,3 +349,73 @@ def test_pam_to_niftis_flow():
...
+def test_math():
+    with TemporaryDirectory() as out_dir:
+        data_path, _, _ = get_fnames(name="small_101D")
+        data_path_a = pjoin(out_dir, "data_a.nii.gz")
+        data_path_b = pjoin(out_dir, "data_b.nii.gz")
+        shutil.copy(data_path, data_path_a)
+        shutil.copy(data_path, data_path_b)
+
+        data, _ = load_nifti(data_path)
+        operations = ["vol1*3", "vol1+vol2+vol3", "5*vol1-vol2-vol3", "vol3*2 + vol2"]
```
arokem, line 363: Will this elegantly handle extraneous white space like "vol1 + vol2 "?

arokem (commented):
A couple more thoughts about this. I think we want this to be pretty bullet proof because it's intended for users who don't program so everything needs to be really clear

dipy/workflows/tests/test_io.py
skoudoro: yes, it handles it without problem. I did not put space because 88 line limit 😅

dipy/workflows/io.py
skoudoro: sounds like a good idea, let me update that

dipy/workflows/tests/test_io.py
skoudoro: I just realized that the last one on the list has a space between operator

skoudoro:
ok, all has been addressed @arokem and @maharshi-gor. Let me know if I missing something

Garyfallidis:
@arokem is there anything else?
 @deka27 can you check this PR too?

maharshi-gor (approved):

deka27:
I dont see any issues, nice addition to CLI.

PR #3430, opened 2024-12-21
ENH: Add type annotation information

jhlegarreta:
Add type annotation information: add the `py.typed` file and distribute the type information with the package by adding the corresponding entry in the `pyproject.toml` file  so that third-party tools can use the DIPY type annotation information.

Documentation:
https://typing.readthedocs.io/en/latest/spec/distributing.html#packaging-typed-libraries

### DIFF
--- a/dipy/meson.build
+++ b/dipy/meson.build
@@ -325,7 +325,8 @@ endif
 # ------------------------------------------------------------------------
 python_sources = [
   '__init__.py',
-  'conftest.py'
+  'conftest.py',
+  'py.typed'
 ]
 
 py3.install_sources(


jhlegarreta:
Have not checked the provided type information.
Have not checked/fixed all issues in typing that DIPY may potentially have.

pyproject.toml
```diff
@@ -181,6 +181,9 @@ Metrics = [
 ]
 # TODO: Add custom commands
 
+[tool.setuptools.package-data]
+"dipy" = ["py.typed"]
```
skoudoro, line 185: you might need to add it in meson also

jhlegarreta: Done in e7ef4de.

skoudoro, line 185: `py.typed` and `meson.build` are enough.

I do not think we need this 2 lines in `pyproject.toml` since we do not use Setuptools at all. So I recommend to remove those lines. Let me know if I am missing something.

skoudoro (commented):
Hi @jhlegarreta,

See my comment below, after that. I believe it can be merged

pyproject.toml
jhlegarreta: :+1: Removed in 0ad2235. I do not know ways to test it, so I trust what your intuition is here.

skoudoro:
Thanks @jhlegarreta, merging

PR #3432, opened 2024-12-24
RF: bump tensorflow minimal version to 2.18.0.

skoudoro:
This PR just bumps TensorFlow minimal version to 2.18.0. This is the first version compatible to numpy 2.0 and we need to enforce that to avoid weird error.


skoudoro:
this PR is ready to go if anyone has time to look at it. Thanks !

PR #3438, opened 2025-01-10
NF: Add 7 new `dipy_fit_*` workflows

skoudoro:
This PR just adds new reconstruction workflows:
- dipy_fit_qball
- dipy_fit_opdt
- dipy_fit_dsid
- dipy_fit_forecast
- dipy_fit_sdt
- dipy_fit_sfm
- dipy_fit_gqi

`test_reconst_rumba.py` has been renamed `test_reconst.py`


dipy/workflows/reconst.py
```diff
@@ -1902,3 +1963,862 @@ def run(
...
+        Parameters
+        ----------
+        input_files : string
+            Path to the input volumes. This path may contain wildcards to
+            process multiple inputs at once.
+        bvalues_files : string
+            Path to the bvalues files. This path may contain wildcards to use
+            multiple bvalues files at once.
+        bvectors_files : string
+            Path to the bvalues files. This path may contain wildcards to use
```
azizasaber, line 2667: Path to the bvector files.

Atharva-Shah-2298:
Used the dipy_fit_qball function, works correctly with the intended results.

Atharva-Shah-2298 (commented):
Used the dipy_fit_qball function, works correctly with the intended results.

maharshi-gor:
Used the dipy_fit_gqi and worked perfectly fine. Used the `cfin` data.

dipy/workflows/reconst.py
```diff
@@ -1902,3 +1963,862 @@ def run(
...
+            multiple bvalues files at once.
+        mask_files : string
+            Path to the input masks. This path may contain wildcards to use
+            multiple masks at once. (default: No mask used)
```
maharshi-gor, line 2463: @skoudoro It says default No mask used but there is no way to leave it empty from the command line.

dipy/workflows/reconst.py
```diff
@@ -1902,3 +1963,862 @@ def run(
...
+            ogfa,
+            osphere,
+            ob,
+            oqa,
+        ) in io_it:
+            logging.info(f"Loading {dwi}")
+            data, affine = load_nifti(dwi)
+            bvals, bvecs = read_bvals_bvecs(bval, bvec)
+
+            # If all b-values are smaller or equal to the b0 threshold, it is
```
maharshi-gor, line 2535: Are we planning to put comments on the code? Or is it just a miss?

dipy/workflows/reconst.py
```diff
@@ -1902,3 +1963,862 @@ def run(
...
+            peaks_gqi = peaks_from_model(
+                model=gqi_model,
+                data=data,
+                sphere=peaks_sphere,
+                relative_peak_threshold=relative_peak_threshold,
+                min_separation_angle=min_separation_angle,
+                mask=mask_vol,
+                return_sh=True,
+                sh_order_max=sh_order_max,
+                normalize_peaks=True,
```
maharshi-gor, line 2579: `normalize_peaks` needs to be passed with the value of `normalize_peaks` instead of default `True` value.

dipy/workflows/reconst.py
```diff
@@ -1902,3 +1963,862 @@ def run(
...
+            peaks_forecast = peaks_from_model(
+                model=forecast_model,
+                data=data,
+                sphere=peaks_sphere,
+                relative_peak_threshold=relative_peak_threshold,
+                min_separation_angle=min_separation_angle,
+                mask=mask_vol,
+                return_sh=True,
+                sh_order_max=sh_order_max,
+                normalize_peaks=True,
```
maharshi-gor, line 2792: `normalize_peaks` Should be provided to the user as we can control it with a variable.

maharshi-gor (changes requested):
All the methods have the option of normalize_peaks which is only exposed in the `gqi` method which is also hardcoded as `True` in the method invocation.

m-agour:
Hi @skoudoro the dsid command line code seems good enough but I am having problems with the functionality itself I can't get it to work on any dataset. for `DiB_70_lte_pte_ste` dataset i am using ` dipy_fit_dsid.exe .\DiB_70_lte_pte_ste.nii.gz .\bval_DiB_70_lte_pte_ste.bval .\bvec_DiB_70_lte_pte_ste.bvec .\DiB_mask.nii.gz ` command. this takes forever and never finished and also not using any resources. what am I doing wrong here?

skoudoro:
Hi @m-agour,

Can you try with the ``taiwan_ntu_dsi`` dataset instead. 

Thank you

dipy/workflows/reconst.py
skoudoro: Good catch, removing

dipy/workflows/reconst.py
skoudoro: For now, I do not think it is necessary to provide access to this variable. 

Later on, if requested, we could.

dipy/workflows/reconst.py
skoudoro: Good catch !

dipy/workflows/reconst.py
skoudoro: same comment as above:

For now, I do not think it is necessary to provide access to this variable.

Later on, if requested, we could.

Garyfallidis:
Merging, a followup PR is needed.

PR #3443, opened 2025-01-31
NF: Replace urllib by requests to improve fetcher stability. 

skoudoro:
It has been at least  2 months that some of our tests are failing due error during fetching test data. 

This PR attempt to stabilize this

- I replaced the standard library `urllib` by 3rd party library named `requests`
- I added an extra step to retry download where it stops when it fails. It will retry 3 times.


skoudoro:
It does not work, still the same issue, closing

PR #3444, opened 2025-02-06
NF: Add 3 new ``dipy_extract_*`` workflows

skoudoro:
This PR adds 3 new workflows:

- dipy_extract_b0
- dipy_extract_shell
- dipy_extract_volume

TODO
------
- [x] tests for the command line


jhlegarreta (commented):
The code for `dipy_extract_volume` is missing.

skoudoro:
> The code for dipy_extract_volume is missing.

this is `class ExtractVolumeFlow(Workflow)`, there is no unique function since this is basic in python (slicing)

dipy/core/gradients.py
```diff
@@ -1343,3 +1343,131 @@ def _check_ornt(ornt):
...
+    """Extract diffusion-weighted imaging (DWI) volumes based on b-value shells.
+
+    Parameters
+    ----------
+    dwi : ndarray
+        4D diffusion-weighted imaging data (X, Y, Z, N), where N is the number of
+        volumes.
+    gtab : GradientTable
+        The gradient table associated with the DWI dataset.
+    bvals_to_extract : list of int, optional
```
pjsjongsung, line 1430: This does not seem to be optional

dipy/workflows/io.py
```diff
@@ -324,6 +332,259 @@ def run(self, input_files, vol_idx=0, out_dir="", out_split="split.nii.gz"):
...
+            process multiple inputs at once.
+        bvalues_files : string
+            Path to the bvalues files. This path may contain wildcards to use
+            multiple bvalues files at once.
+        bvectors_files : string
+            Path to the bvectors files. This path may contain wildcards to use
+            multiple bvectors files at once.
+        bvals_to_extract : variable int, optional
+            List of b-values to extract.
+        b0_threshold : float, optional
```
pjsjongsung, line 469: Some of the parameters here, b0_threshold, bvecs_tol seem to change the behavior a lot from the original function. If it is an important behavior shouldn't it be in the actual function not just on the workflow?

dipy/workflows/io.py
```diff
@@ -324,6 +332,259 @@ def run(self, input_files, vol_idx=0, out_dir="", out_split="split.nii.gz"):
...
+        io_it = self.get_io_iterator()
+        for dwi, bval, bvec, ob0 in io_it:
+            logging.info("Extracting b0 from {0}".format(dwi))
+            data, affine, image = load_nifti(dwi, return_img=True)
+
+            bvals, bvecs = read_bvals_bvecs(bval, bvec)
+            # If all b-values are smaller or equal to the b0 threshold, it is
+            # assumed that no thresholding is requested
+            if any(mask_non_weighted_bvals(bvals, b0_threshold)):
+                if b0_threshold < bvals.min():
```
pjsjongsung, line 399: This warning is applicable to the original function as well. Maybe it should be in the base function and not the workflow?

dipy/core/gradients.py
skoudoro: good catch

dipy/workflows/io.py
skoudoro: the original function expect b0_mask as an input which can be a simple boolean array.  No assumption and does not check the bvals. 

So I think, it is better here. I do not want to force the user to pass `bvals` in the original function.

dipy/workflows/io.py
skoudoro: since the original function expect a `gtab` (gradient table), `b0_threshold` and `bvecs_tol` are implicitly needed when the user will create his gradient table.  The original function assumes that this work has been done before

skoudoro:
Thank you for the feedback @pjsjongsung. your comment has been addressed

dipy/core/gradients.py
```diff
@@ -1343,3 +1343,131 @@ def _check_ornt(ornt):
...
+    if chunk_size:
+        b0_data = []
+        for idx in b0_indices:
+            b0_data.append(dwi[..., idx])
+        b0_data = np.stack(b0_data, axis=-1)
+    else:
+        b0_data = dwi[..., b0_mask]
+
+    if group_contiguous_b0:
+        groups = np.split(b0_indices, np.where(np.diff(b0_indices) > 1)[0] + 1)
```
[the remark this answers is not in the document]
skoudoro: Not sure to understand, can you provide an example ?

dipy/core/gradients.py
```diff
@@ -1343,3 +1343,131 @@ def _check_ornt(ornt):
...
+        4D diffusion-weighted imaging data (X, Y, Z, N), where N is the number of
+        volumes.
+    gtab : GradientTable
+        The gradient table associated with the DWI dataset.
+    bvals_to_extract : list of int, optional
+        List of b-values to extract. If None, all b-values are included.
+    tol : int, optional
+        Tolerance range for b-value selection. A value of 20 means volumes with b-values
+        within ±20 units of the specified b-values will be extracted.
+    chunk_size : int, optional
```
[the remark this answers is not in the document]
skoudoro: done and removed, `chunk_size` was not well used, removed for now

dipy/core/gradients.py
```diff
@@ -1343,3 +1343,131 @@ def _check_ornt(ornt):
...
+        The b-values of the selected volumes.
+    output_bvecs : ndarray
+        The b-vectors of the selected volumes.
+    """
+    bvals = gtab.bvals
+    bvecs = gtab.bvecs
+
+    mask = (
+        np.ones_like(bvals, dtype=bool)
+        if bvals_to_extract is None
```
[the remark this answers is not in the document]
skoudoro: done

dipy/core/gradients.py
```diff
@@ -1343,3 +1343,131 @@ def _check_ornt(ornt):
...
+        else np.any(np.abs(bvals[:, None] - np.array(bvals_to_extract)) <= tol, axis=1)
```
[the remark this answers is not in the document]
skoudoro: done

dipy/core/tests/test_gradients.py
```diff
@@ -1008,3 +1010,68 @@ def test_gradient_table_len():
         obt_grad_count1 = gtab.__len__()
         obt_grad_count2 = len(gtab)
         assert exp_grad_count == obt_grad_count1 == obt_grad_count2
+
+
+def test_extract_b0():
```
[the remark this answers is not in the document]
skoudoro: fixed

dipy/core/gradients.py
```diff
@@ -1343,3 +1343,133 @@ def _check_ornt(ornt):
...
+        b0_indices = np.where(b0_mask)[0]
+        groups = np.split(b0_indices, np.where(np.diff(b0_indices) > 1)[0] + 1)
+        grouped_b0 = []
+        for group in groups:
+            group_data = dwi[..., group - b0_indices[0]]
+            if strategy == "first":
+                grouped_b0.append(group_data[..., 0])
+            elif strategy == "mean":
+                grouped_b0.append(np.mean(group_data, axis=-1))
+            else:  # "all"
```
pjsjongsung, line 1402: The behavior of "all" here confuses me. This would return a 5D array, which does not seem ideal. Shouldn't it concatenate instead of stack in this case?

pjsjongsung: I understand then it would be the same behavior as using "all" without `group_contiguous_b0`. We could even then just change the code to check if strategy is "all" and return `b0_data` in that case regardless of what `group_contiguous_b0`

dipy/core/tests/test_gradients.py
```diff
@@ -1008,3 +1010,74 @@ def test_gradient_table_len():
...
+    npt.assert_array_equal(result_first, expected_first)
+
+    # Test extracting mean b0 volume
+    result_mean = extract_b0(dwi, b0_mask, strategy="mean")
+    npt.assert_equal(result_mean.shape, dwi.shape[:-1])
+    npt.assert_almost_equal(result_mean, expected_mean)
+
+    # Test contiguous grouping
+    result_grouped = extract_b0(dwi, b0_mask, group_contiguous_b0=True, strategy="mean")
+    assert len(result_grouped.shape) >= 3  # Should be at least 3D
```
pjsjongsung, line 1039: Why are we only testing shape here instead of the expected values like the one above?

dipy/core/gradients.py
```diff
@@ -1343,3 +1343,133 @@ def _check_ornt(ornt):
...
+    Parameters
+    ----------
+    dwi: ndarray
+        4D diffusion-weighted imaging data (X, Y, Z, N), where N is the number of
+        volumes.
+    b0_mask: ndarray of bool
+        Mask over the time dimension (4th) identifying b0 volumes.
+    group_contiguous_b0: bool, optional
+        If True, each contiguous b0 volumes are grouped together.
+    strategy: str, optional
```
maharshi-gor, line 1360: Please mention what is default value

dipy/core/gradients.py
maharshi-gor: @pjsjongsung If we do that we will be loosing the grouping information of the b0s. Having said that we also need to tell the user explicitly as the use-case of `strategy="all"` with `group_contiguous_bo=True` is not possible as it is an inconsistent behavior from what is happening with other strategies.

dipy/core/gradients.py
```diff
@@ -1343,3 +1343,133 @@ def _check_ornt(ornt):
...
+        4D diffusion-weighted imaging data (X, Y, Z, N), where N is the number of
+        volumes.
+    gtab : GradientTable
+        The gradient table associated with the DWI dataset.
+    bvals_to_extract : list of int
+        List of b-values to extract. If None, all b-values are included.
+    tol : int, optional
+        Tolerance range for b-value selection. A value of 20 means volumes with b-values
+        within ±20 units of the specified b-values will be extracted.
+    group_shells : bool, optional
```
maharshi-gor, line 1430: Provide the default value details.

dipy/core/gradients.py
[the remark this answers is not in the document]
azizasaber: What I meant is imagine b0_indices is an empty array or it only has one value inside (not sure if this can happen practically or not). Then using np.diff will return an empty array I believe which will then raise an error for np.split since no value is being passed to split.

maharshi-gor (changes requested):
We can provide information of the default values of the optional parameters in the docstrings for easier access of values

dipy/core/gradients.py
jhlegarreta: No, please don't: @maharshi-gor default values are documented in the function signature, and adding them to the docstrings only increases the maintenance burden and the risk of going out of sync if the default value changes. So unless the default values triggers some secondary computation that needs to be documented, please do not add default values to docstrings. This has been mentioned multiple times.

What may be relevant is to mention the different possible strategies, either here or in the function kong description.

dipy/core/gradients.py
jhlegarreta: Please don't. X-ref: https://github.com/dipy/dipy/pull/3444#discussion_r1966687825.

jhlegarreta:
> We can provide information of the default values of the optional parameters in the docstrings for easier access of values

Please, don't: https://github.com/dipy/dipy/pull/3444#discussion_r1966687825

dipy/core/gradients.py
skoudoro: Agree with @maharshi-gor here. let me know if you want more detailed explanation.

Garyfallidis:
Extract_b0 should require only bvals.
Extract_shell has some issue with cfin_multib
Extract_volume works well and it is nice how it gets ranges as strings. Perhaps use the same for extract_b0.

dipy/core/gradients.py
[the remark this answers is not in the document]
skoudoro: ok, fixed, we check if `b0_mask` is emtpy. if it is not, this case should not happen

skoudoro:
Failing CI's are unrelated. This PR is ready to go

PR #3446, opened 2025-02-12
Dki constraints

samcoveney:
More generalized constraint matrices, including more basis matrices satisfying e.L_i.e = 0

See https://github.com/dipy/dipy/discussions/3419


samcoveney:
Sorry looks like in my second 'tidy' commit, I removed the code pointing to the _correct_ constraint matrix, let me put that in place and update...

samcoveney:
everything seems to pass okay, that can pass okay. Let me know

dipy/reconst/dki.py
```diff
@@ -1799,6 +1799,7 @@ def __init__(self, gtab, *args, fit_method="WLS", return_S0_hat=False, **kwargs)
             if isinstance(self.convexity_level, str):
                 if self.convexity_level == "full":
                     self.sdp_constraints = load_sdp_constraints("dki")
+                    # NOTE: load my constraint matrices instead, see if they work!
```
skoudoro, line 1802: Not sure to understand this comment. 

it seems that we always use your constraint matrices by default. Right ?

maybe comment to remove

skoudoro (commented):
Hi @samcoveney,

Apart from my question below, all good, ready to be merged for me.

dipy/core/optimize.py
```diff
@@ -433,12 +433,12 @@ def __init__(self, m, *, A=None, L=None):
         .. footbibliography::
         """
         # Input
-        self.A = A
-        self.L = L
+        self.A = A  # list: [zeros, H(omega), L(alpha)] --- see eq 22
```
arokem, line 436: This is eq 22 in DelaHaije2020?

dipy/data/files/meson.build
```diff
@@ -14,6 +14,7 @@ data_files = [
     'dki_constraint_2.npz',
     'dki_constraint_4.npz',
     'dki_constraint.npz',
+    'dki_constraint_SC.npz',
```
arokem, line 17: Maybe instead of your initials, we could name this with a word that explains how these are different?

arokem (commented):
Small comments from me

pjsjongsung:
Not sure if it is intended but the documentation is inside __init__ preventing it from appearing in the docs.

dipy/reconst/dki.py
samcoveney: yes you're right, needs to be removed, I will do it

dipy/core/optimize.py
samcoveney: Yes, I will put a comment in somewhere that makes this clear

dipy/data/files/meson.build
samcoveney: Could do, I don't think it matters very much though. I though having a second file would be better, so that the old one wasn't deleted.

I have a paper under review that actually shows and explains these equations, so my initials are not the worst choice. The main difference with the existing ones are just extra slack variables and different form of the main basis matrix - it appears to do exactly the same as before, but this one is more general because is it fully parameterized with all the possible slack matrices.

What would you suggest? Can I just leave a comment instead?

samcoveney:
Thanks all, I am on paternity leave now, and this isn't urgent PR, so we may have to wait to get it over the line.

dipy/data/files/meson.build
samcoveney: If this isn't that important, can we leave until I get a chance to write the tutorial?

skoudoro:
> Thanks all, I am on paternity leave now

Congratulations ! 🎈 🎉

skoudoro:
if ok for @arokem, I will go ahead and merge this.

arokem:
Yes you can go ahead

PR #3449, opened 2025-02-18
Adding DeepN4 PyTorch model

pjsjongsung:
This PR adds a pytorch version of the DeepN4 model. Files, tests and data fetchers have been changed accordingly. Some pep8 and doc errors has been addressed in test_evac.py and evac.py as well.


dipy/nn/torch/deepn4.py
```diff
@@ -0,0 +1,375 @@
...
+
+    The DeepN4 model :footcite:p:`Kanakaraj2024` predicts the bias field for
+    magnetic field inhomogeneity correction on T1-weighted images.
+
+    References
+    ----------
+    .. footbibliography::
+    """
+
+    @warning_for_keywords()
```
skoudoro, line 177: no need of this

dipy/nn/torch/deepn4.py
```diff
@@ -0,0 +1,375 @@
...
+            img.shape,
+            offset_array,
+            voxsize,
+            scale,
+            crop_vs,
+            pad_vs,
+        )
+
+        # Correct the image
+        THRESHOLD = 0.5
```
skoudoro, line 368: could this be an argument of the predict function ?

dipy/nn/torch/deepn4.py
```diff
@@ -0,0 +1,375 @@
...
+    def predict(self, img, img_affine, *, voxsize=(1, 1, 1)):
+        """Wrapper function to facilitate prediction of larger dataset.
+        The function will mask, normalize, split, predict and 're-assemble'
+        the data as a volume.
+
+        Parameters
+        ----------
+        input_file : string
+            Path to the T1 scan
+
```
skoudoro, line 327: docstring to update

skoudoro (commented):
Hi @pjsjongsung,

Overall, it looks good, See below few comments before merging

dipy/nn/tf/deepn4.py
```diff
@@ -323,8 +329,8 @@ def predict(self, img, img_affine, *, voxsize=(1, 1, 1)):
         )
 
         # Correct the image
-        THRESHOLD = 0.5
-        below_threshold_mask = np.abs(upsample_final_field) < THRESHOLD
+        threshold = 0.5
```
skoudoro, line 332: to remove or `threshold = threshold`

dipy/nn/torch/deepn4.py
```diff
@@ -0,0 +1,380 @@
...
+            img.shape,
+            offset_array,
+            voxsize,
+            scale,
+            crop_vs,
+            pad_vs,
+        )
+
+        # Correct the image
+        threshold = 0.5
```
skoudoro, line 373: same here

dipy/nn/torch/deepn4.py
```diff
@@ -0,0 +1,380 @@
...
+        This model is designed to take as input file T1 signal and predict
+        bias field. Effectively, this model is mimicking bias correction.
+
+        Parameters
+        ----------
+        verbose : bool, optional
+            Whether to show information about the processing.
+        """
+
+        if not have_torch:
```
m-agour, line 194: Maybe `raise ImportError("PyTorch is required to use DeepN4.")`

dipy/nn/torch/deepn4.py
```diff
@@ -0,0 +1,380 @@
...
+        x_test : np.ndarray
+            Image should match the required shape of the model.
+
+        Returns
+        -------
+        np.ndarray (batch, ...)
+            Predicted bias field
+        """
+
+        return self.model(x_test)[:, 0].detach().numpy()
```
m-agour, line 257: Here the model could be on the GPU or CPU since it's being handled but the input data on the other hand is not wich could trigger an error so it's better to match the device between data and model you can get model device using  `device = any_model_parameter.device`  or directly `device = next(self.model.parameters()).decive` or maybe set it as a global variable.

dipy/nn/torch/deepn4.py
```diff
@@ -0,0 +1,380 @@
...
+        Load the custom pre-training weights to use for the fitting.
+        get_fnames('deepn4_default_torch_weights').
+
+        Parameters
+        ----------
+        weights_path : str
+            Path to the file containing the weights (hdf5, saved by tensorflow)
+        """
+        try:
+            if torch.cuda.is_available():
```
m-agour, line 222: this could be simplified by getting the device first then

 `device = torch.device('cuda' if torch.cuda.is_available() else 'cpu')
    state_dict = torch.load(weights_path, map_location=device)
    self.model.load_state_dict(state_dict['model_state_dict'])
    self.model.to(device)` and even better the device could bee a global variable to be used for data later or be stored somewhere.

m-agour: Also `try: above except: invalid model weights`

m-agour (commented):
Hi @pjsjongsung, thanks for this PR it looks good. Please see my comments below.

dipy/nn/torch/deepn4.py
pjsjongsung: That was possible to say but now that we have both TF and torch models it is hard to say that statement. This behavior is also the case for every other nn model for now we can keep it like this and later think of a error message to use

m-agour: got it

jhlegarreta (changes requested):
@pjsjongsung Add `, optional` to the docstrings of all optional parameters that you have added in this PR.

dipy/nn/torch/deepn4.py
```diff
@@ -0,0 +1,375 @@
...
+            rx,
+            rX,
+            ry,
+            rY,
+            rz,
+            rZ,
+            in_max,
+        )
+
+    def predict(self, img, img_affine, *, voxsize=(1, 1, 1)):
```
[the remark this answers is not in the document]
pjsjongsung: I looked at the code but since there are many parameters shared between pre and post processing I don't think there is a easy simpler way in my mind right now

pjsjongsung:
It seems there is a small problem with histo_resdnn and I do not know what to do since I do not know the history.

Basically histo_resdnn's model uses a default `sh_order_max=8` when initializing the model, but it uses a default weight of `sh_order_max=6`. This problem seems to be intentional in some way, given how `test_histo_resdnn.py` works in `test_wrong_sh_order_input`. It assumes an error when using the default value.

Should we fix this? This throws an error in line 165 but is skipped (I think) by  catching `RuntimeError` instead of `ValueError`.

dipy/nn/torch/deepn4.py
```diff
@@ -0,0 +1,377 @@
...
+    magnetic field inhomogeneity correction on T1-weighted images.
+
+    References
+    ----------
+    .. footbibliography::
+    """
+
+    @doctest_skip_parser
+    def __init__(self, *, verbose=False):
+        r"""
```
jhlegarreta, line 179: Do not use raw strings unless necessary.

Garyfallidis:
This PR lacks a tutorial and a CLI interface.

Garyfallidis:
I tried the method with some clinical data and it worked correctly.

skoudoro:
> This PR lacks a tutorial and a CLI interface.

a follow up PR will be done for that

pjsjongsung:
> It seems there is a small problem with histo_resdnn and I do not know what to do since I do not know the history.
> 
> Basically histo_resdnn's model uses a default `sh_order_max=8` when initializing the model, but it uses a default weight of `sh_order_max=6`. This problem seems to be intentional in some way, given how `test_histo_resdnn.py` works in `test_wrong_sh_order_input`. It assumes an error when using the default value.
> 
> Should we fix this? This throws an error in line 165 but is skipped (I think) by catching `RuntimeError` instead of `ValueError`.

@skoudoro @maharshi-gor any suggestions on the comment above?
If not, I can make the small edit Jon wants and take care of that later in a separate PR

skoudoro:
@frheault, can you provide an opinion for above ? Thanks !

pjsjongsung:
The error was not the wrong default value but the fact that torch and tf were throwing different error types for the same problem. Hopefully this PR takes care of it.

pjsjongsung:
@jhlegarreta Could you elaborate on what you mean by 'avoid adding a first blank line'? Do you mean in the docstring?

jhlegarreta:
> @jhlegarreta Could you elaborate on what you mean by 'avoid adding a first blank line'? Do you mean in the docstring?

Docstring, yes, e.g.:
https://github.com/dipy/dipy/pull/3449/files#diff-bafd744ee6cb6473dc6ee141325fcbb688fec7c3cfbdca17d9b10a01f5a33e07R179-R180

Garyfallidis:
Any other issues? Should I go ahead and merge? @pjsjongsung @jhlegarreta and @skoudoro ?

jhlegarreta:
> Any other issues? Should I go ahead and merge? @pjsjongsung @jhlegarreta and @skoudoro ?

Well, the code format has to pass before thinking about merging, and
- Not marked as optional
https://github.com/dipy/dipy/pull/3449/files#diff-7d0c988772bb79422dcbc6bfdbc06b4480302ae72b09cd9df421fd4fa1636c9bR288
- Unnecessary blank line: https://github.com/dipy/dipy/pull/3449/files#diff-bafd744ee6cb6473dc6ee141325fcbb688fec7c3cfbdca17d9b10a01f5a33e07R166. Also, not sure if the class doscstrings (i.e. those that do not belong to a method are displayed in the docs). Please check.\
- More blank lines, https://github.com/dipy/dipy/pull/3449/files#diff-bafd744ee6cb6473dc6ee141325fcbb688fec7c3cfbdca17d9b10a01f5a33e07R207. Please check all. There are other instances in the added code.

skoudoro:
@pjsjongsung, can you fix the code format issue: https://github.com/dipy/dipy/actions/runs/13703142391/job/38321790573?pr=3449#step:4:118

skoudoro:
Thank you for the update.

Not sure why, you always miss my comment here: https://github.com/dipy/dipy/pull/3449/files#r1962005301

can you update and we are good

pjsjongsung:
> Thank you for the update.
> 
> Not sure why, you always miss my comment here: https://github.com/dipy/dipy/pull/3449/files#r1962005301
> 
> can you update and we are good

I have no idea either I am so sorry. I addressed it this time.

skoudoro:
Thank you for this work @pjsjongsung  

I am going ahead and merge this PR.

Issue #3453, opened 2025-02-27
Inconistancy between documentation and implementation on dipy.reconst.dti.nlls_fit_tensor


tito21:
In version 1.10 the `dipy.reconst.dti.nlls_fit_tensor` loss the option to pass sigma (noise level of each image) to the reconstruction. Instead, we can only pass the weights argument. I'm calculating the weights as 1/sigma^2. The [documentation](https://docs.dipy.org/1.10.0/reference/dipy.reconst.html#dipy.reconst.dti.nlls_fit_tensor) specifies weights should have shape (X, Y, Z, g), however when I tried this, I got the following error:

```
Traceback (most recent call last):
  File "/Users/adibiase/debug_dipy/dipy/debug.py", line 65, in <module>
    tensor_fit_weighted = TensorModel(gtab, fit_method="NLLS", weights=weights).fit(signal)
  File "/Users/adibiase/debug_dipy/dipy/dipy/testing/decorators.py", line 201, in wrapper
    return convert_positional_to_keyword(func, args, kwargs)
  File "/Users/adibiase/debug_dipy/dipy/dipy/testing/decorators.py", line 192, in convert_positional_to_keyword
    return func(*args, **kwargs)
  File "/Users/adibiase/debug_dipy/dipy/dipy/reconst/dti.py", line 816, in fit
    params_in_mask, extra = self.fit_method(
                            ~~~~~~~~~~~~~~~^
        self.design_matrix,
        ^^^^^^^^^^^^^^^^^^^
    ...<3 lines>...
        **self.kwargs,
        ^^^^^^^^^^^^^^
    )
    ^
  File "/Users/adibiase/debug_dipy/dipy/dipy/testing/decorators.py", line 201, in wrapper
    return convert_positional_to_keyword(func, args, kwargs)
  File "/Users/adibiase/debug_dipy/dipy/dipy/testing/decorators.py", line 192, in convert_positional_to_keyword
    return func(*args, **kwargs)
  File "/Users/adibiase/debug_dipy/dipy/dipy/reconst/dti.py", line 1860, in nlls_fit_tensor
    this_param, status = opt.leastsq(
                         ~~~~~~~~~~~^
        err_func,
        ^^^^^^^^^
    ...<2 lines>...
        Dfun=jac_func,
        ^^^^^^^^^^^^^^
    )
    ^
  File "/Users/adibiase/.local/lib/python3.13/site-packages/scipy/optimize/_minpack_py.py", line 444, in leastsq
[...]

PR #3455, opened 2025-02-28
Fix Bugs in #3453: Ensure Correct Weight Reshaping & Consistent Extra Output in iter_fit_tensor

tito21:
Fix both bugs mentioned in #3453:

 1. Reshapes the weights passed to `nlls_fit_tensor` to have the same shape as the flattened data
 2. Make sure the extra output is always None when doing the fitting in chunks with the `iter_fit_tensor` decorator

### DIFF
--- a/dipy/reconst/dti.py
+++ b/dipy/reconst/dti.py
@@ -1369,6 +1369,9 @@ def iter_fit_tensor(*, step=1e4):
             if extra:
                 for key in extra:
                     extra[key] = extra[key].reshape(shape + (-1,))
+            else:
+                # Make sure extra is None if there is no extra
+                extra = None
 
             if return_S0_hat:
                 return (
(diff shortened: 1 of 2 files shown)


arokem:
Thanks for your work on this. It would be great to add a test that fails without this fix, just to make sure that this doesn't creep back in.

tito21:
Thanks for your feedback. Looking at the tests for the different tensor fitting methods, they only fit on one voxel. I guess this is done to keep the tests fasts. However, the second problem only happens with more than 1e4 voxels.

The first problem is easier to test, just give incorrectly shaped weights.

tito21:
OK, there are some tests that force the `iter_fit_tensor` and used that as a template for adding a test for the second case. 
I checked and it fails without this fix.

skoudoro:
Thank you for doing this @tito21!

it would be great if @samcoveney, could have a very quick look here before moving forward

skoudoro (commented):
Looks good to me.

Thank you @tito21 and sorry for the delay. 

I fixed some pep8 issues. Waiting for the CI's and then I will go ahead and merge this PR.

samcoveney:
Hi all, late to this, but I've tried to follow what has happened and the original issue.

I think it seems solved, hopefully nothing go broken in the process. I think my use of nlls_fit_tensor always had the data and weights passed in pre-flattened, so I must have missed this. If it is fixed now, thanks!

PR #3460, opened 2025-03-05
[WIP] RF: force use of header for download

skoudoro:
Small experiment to see if having an header when download data can fix our fetching issue that we have the past 3-4months.


skoudoro:
I got my answer, closing....

Issue #3371, opened 2024-10-07
Horizon miscalculating contrast when returning to previous volume


pjsjongsung:
When returning to a binary mask volume from after moving, it miscalculates the contrast and throws an error like below

`C:\Work\Repos\dipy\dipy\viz\horizon\tab\slice.py:286: UserWarning: Volume N°0 does not have any contrast. Please, check the value ranges of your data. Returning to previous volume.`

This happens even though the volume it was moving back to was ok with the contrast (horizon did not have any problems visualizing it)

DIPY is on it's master branch
numpy 1.24.4
scipy 1.9.0
nibabel 5.1.0
h5py 3.10.0
cvxpy 1.4.2
fury 0.10.0

[real_test_pred5.nii.gz](https://github.com/user-attachments/files/17283412/real_test_pred5.nii.gz)

this is an example file that reproduces the problem

Issue #3112, opened 2024-03-09
Horizon fails with a single channel 4D volume


pjsjongsung:
## Description
The recent change of code for detecting unique values uses image shape for sampling random integers. This will throw an error when dealing with single channel 4D volumes since it would try to sample from a range of 0 to 0.

## Way to reproduce
```
from dipy.viz.horizon.app import horizon
from dipy.io.image import load_nifti

image, affine = load_nifti(some_file_path_with_single_channel)

horizon(images=[(image, affine)])
```

Garyfallidis:
You mean with data that have the following shape? For example (128, 128, 65, 1)?

Garyfallidis:
People can currently use dipy_split to get the "first" volume. But indeed this should be fixed.

PR #3461, opened 2025-03-05
BF: Handling binary and small intensity value range and Volume slider not be shown if single channel provided.

maharshi-gor:
### BF: Handling binary and small intensity value range.

**Resolves**
#3371 Horizon miscalculating contrast when returning to previous volume.
#3112 Horizon fails with a single channel 4D volume.

**PR Contents**

- Replaces the percentile based placement of the sliders from previous volume to new volume. It provides support for the smaller range of the intensities.
- Ignores the selection of the intensity range with a warning. If the intensity range selection does not have any contrast in it while changing to newer volume.
- Ignores the volume slider if only single channel provided in volume.

**Local Testing**

- The changes have been tested on the `provided example`, `stanford hardi` and `cenir_multib` datasets.
- To get the warning from the _adaptive_percentile method please move both the sliders of the intensities super close in the example given, which will lead to only single value in the selection and will result in to the warning but as the entire data has more intensity value it will not result into warning as it will fallback to entire intensity range and update the slider.

### DIFF
--- a/dipy/viz/horizon/visualizer/slice.py
+++ b/dipy/viz/horizon/visualizer/slice.py
@@ -35,7 +35,7 @@ class SlicesVisualizer:
 
         self._slice_actors = [None] * 3
 
-        if len(self._data.shape) == 4 and self._data.shape[-1]:
+        if len(self._data.shape) == 4 and self._data.shape[-1] == 1:
             self._data = self._data[:, :, :, 0]
 
         self._data_ndim = self._data.ndim
@@ -174,8 +174,9 @@ class SlicesVisualizer:
         self._picked_voxel_actor = actor.dot(pnt, colors=(0.9, 0.4, 0.0), dot_size=10)
         self._scene.add(self._picked_voxel_actor)
 
-    def _adaptive_percentile(self, vol_data, percentile, idx):
-        value_range = np.percentile(vol_data, percentile)
+    def _adaptive_percentile(self, vol_data, intensity_ratios, idx):
+        value_range = np.percentile(np.ravel(vol_data), intensity_ratios * 100)
+        default_range = False
 
         if np.sum(np.diff(value_range)) == 0:
             warnings.warn(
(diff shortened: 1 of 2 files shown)


dipy/viz/horizon/visualizer/slice.py
```diff
@@ -36,8 +35,11 @@ def __init__(
 
         self._slice_actors = [None] * 3
 
-        self._data_ndim = data.ndim
-        self._data_shape = data.shape
+        if len(self._data.shape) == 4 and self._data.shape[-1]:
```
pjsjongsung, line 38: Is there a reason we check for self._data.shape[-1]? Are there cases we get a shape something like (256, 256, 256, None) ?

dipy/viz/horizon/visualizer/slice.py
```diff
@@ -172,19 +174,29 @@ def _replace_picked_voxel_actor(self, x, y, z):
         self._picked_voxel_actor = actor.dot(pnt, colors=(0.9, 0.4, 0.0), dot_size=10)
         self._scene.add(self._picked_voxel_actor)
 
+    def _adaptive_percentile(self, vol_data, percentile, idx):
+        value_range = np.percentile(vol_data, percentile)
+
+        if np.sum(np.diff(value_range)) == 0:
+            warnings.warn(
+                f"The selected intensity range have no contrast for Volume N°{idx}."
+                "The selection intensities will be ignored.",
```
pjsjongsung, line 183: We are technically not ignoring them but changing them to default value.

dipy/viz/horizon/visualizer/slice.py
```diff
@@ -172,19 +174,29 @@ def _replace_picked_voxel_actor(self, x, y, z):
...
+                stacklevel=2,
+            )
+            value_range = np.percentile(vol_data, [0, 100])
```
pjsjongsung, line 186: Also minor issue. I think this is equal to value_range = (np.min(vol_data), np.max(vol_data)). From what I've tested, using min and max is much faster. I understand from a code interpretability standpoint this is better though.

dipy/viz/horizon/visualizer/slice.py
maharshi-gor: No that is a typo we are just checking for the volumes with single channel. Where we convert it to a normal 3D image instead of 4D image.

maharshi-gor:
Hello @pjsjongsung and @skoudoro 

I worked around a bit of ratios of the slider and able to find an approach which can work for both 4D volumes with contrast and without contrast.

Garyfallidis:
@pjsjongsung can you confirm that this resolves your issue?

pjsjongsung:
I can confirm that it resolves the issues! Thanks @maharshi-gor

skoudoro (approved):
Looks good to me too.

Thanks for the review @pjsjongsung and thanks for this PR @maharshi-gor, 

merging

Issue #3164, opened 2024-03-30
Simplify or remove the use of mpl_tri in viz module


skoudoro:
I know Serge, but I imagined that the package needs to be imported so that some sort of processing on the mesh can be completed. IIRC sometimes with `matplotlib` this happens, but I cannot point to an example out of my mind. Best would probably be to have a test and see if removing it makes any difference. Can be done in a separate PR.

_Originally posted by @jhlegarreta in https://github.com/dipy/dipy/pull/3161#discussion_r1545463229_

PR #3465, opened 2025-03-06
BF: Import fixes

maharshi-gor:
### BF: Import fixes

**Resolves**
#3164 Simplify or remove the use of mpl_tri in viz module.

**PR Contents**
- mpl_toolkits.basemap was not used using the `optional_package` and had a separate import.
- `mpl_tri` is not required to be install separately as it is already included in the basemap import.

### DIFF
--- a/dipy/viz/projections.py
+++ b/dipy/viz/projections.py
@@ -14,7 +14,6 @@ from dipy.utils.optpkg import optional_package
 
 matplotlib, has_mpl, setup_module = optional_package("matplotlib")
 plt, _, _ = optional_package("matplotlib.pyplot")
-mpl_tri, _, _ = optional_package("matplotlib.tri")
 bm, has_basemap, _ = optional_package("mpl_toolkits.basemap")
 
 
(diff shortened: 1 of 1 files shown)


jhlegarreta (commented):
I am fine with this change, but what bug does this fix, i.e. where in the logs or executions is the bug encountered? If there is such a bug, what is the error trace.

maharshi-gor:
> I am fine with this change, but what bug does this fix, i.e. where in the logs or executions is the bug encountered? If there is such a bug, what is the error trace.

Hi @jhlegarreta , Thanks for the feedback.

This technically is not a bugfix. It is more of refractor of the existing code. The method `sph_projection` was using the `matplotlib` before the `basemap` import so Tripwire message will be initiated anyways. But, from our coding practice we do use the optional dependencies using `optional_package`.

Also, the `mpl_tri` import was not used anywhere in the code.

Let me know if you have any other concerns.

jhlegarreta (approved):
CI errors are unrelated.

skoudoro (approved):
Thank you for this @maharshi-gor 

merging

PR #3470, opened 2025-03-06
RF: Enforce bvecs for the cli extract_b0 

skoudoro:
In some case, this cli was failing because a `bvecs` were missing. 

However, bvecs are not really needed in this case so we generate a random one to avoid any issue.

Can you try @iprajwalreddy?


iprajwalreddy:
@skoudoro I just checked it now and it seems to be working properly.

PR #3476, opened 2025-03-10
BF - generate_tractogram needs seed coordinate in image space

gabknight:
The tractography is fully performed in image space. However, seed coordinates are generated and passed in world space. This fixes the issue by applying the inverse affine to seed coordinates before calling `generate_tractogram(.)`.

I think it would be better to do it inside the  `generate_tractogram(.)` function, but I'm not sure how to do it properly in Cython. @skoudoro can you have a look?

### DIFF
--- a/dipy/tracking/tests/test_tracker.py
+++ b/dipy/tracking/tests/test_tracker.py
@@ -48,7 +48,7 @@ def track(method, **kwargs):
         "return_all": True,
         "sf": fODFs if use_sf else None,
         "sh": sh if not use_sf else None,
-        "seeds_directions": directions if use_directions else None,
+        "seed_directions": directions if use_directions else None,
         "sphere": sphere,
     }
     stream_gen = method(seeds, sc, affine, **params)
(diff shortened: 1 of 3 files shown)


skoudoro:
Hi @gabknight,

Good Catch !

in generate_tractogram, it would be the same code. see here: https://github.com/dipy/dipy/blob/3289096989f68d217dd5a6399037b6feb0551af4/dipy/tracking/tractogen.pyx#L109

you might need to create a temporary variable.

Can you update ?

gabknight:
Ok, should be all good now. @skoudoro 
The last commit uniformized `seed_positions/seed_directions` variable names across functions.

skoudoro:
Hi @gabknight,

All good, Thanks!

Can you add in the docstring of all the trackers in which space the seeds need to be. Thanks!

Then, I can merge

gabknight:
Done. Thanks @skoudoro !

skoudoro (approved):
All good, merging.

Thanks @gabknight !

PR #3477, opened 2025-03-10
NF: Add dipy_classify_tissue CLI

skoudoro:
This PR introduce a new CLI for Tissue classification on T1w images (using HMRF) and DWI images (using DAM).

Test has been added


dipy/workflows/segment.py
```diff
@@ -368,3 +372,130 @@ def run(
...
+            process multiple inputs at once.
+        bvals_file : string, optional
+            Path to the b-values file. Required for 'dam' method.
+        method : string, optional
+            Method to use for tissue extraction. Options are:
+                - 'hmrf': Markov Random Fields modeling approach.
+                - 'dam': Directional Average Maps, proposed by :footcite:p:`Cheng2020`.
+
+            'hmrf' method is recommended for T1w images, while 'dam' method is
+            recommended for DWI images.
```
maharshi-gor, line 413: Should we also add that the DWI image should be multi shell?

maharshi-gor (commented):
Have a small suggestion otherwise, the CLI works.

dipy/workflows/segment.py
skoudoro: Thanks, added!

dipy/workflows/segment.py
```diff
@@ -368,3 +372,144 @@ def run(
...
+                    "tissue_classified_pve.nii.gz",
+                    f"{prefix}_tissue_classified_pve.nii.gz",
+                )
+
+        self.update_flat_outputs(self.flat_outputs, io_it)
+
+        for fpath, tissue_out_path, opve in io_it:
+            logging.info(f"Extracting tissue from {fpath}")
+
+            data, affine, img = load_nifti(fpath, return_img=True)
```
pjsjongsung, line 481: We are loading the img here but we do not use it

dipy/workflows/segment.py
```diff
@@ -368,3 +372,144 @@ def run(
...
+                )
+                result = np.zeros(wm_mask.shape)
+                result[wm_mask] = 1
+                result[gm_mask] = 2
+                save_nifti(tissue_out_path, result, affine)
+                save_nifti(
+                    opve, np.stack([wm_mask, gm_mask], axis=-1).astype(np.int32), affine
+                )
+
+            logging.info(f"Tissue saved as {tissue_out_path}")
```
pjsjongsung, line 513: Since we always save PVE, should we also mention that PVE is saved at this path?

dipy/workflows/segment.py
skoudoro: Thanks, fixed!

dipy/workflows/segment.py
skoudoro: thanks, updated!

Garyfallidis:
There is no default for --nclass 
Merging... but fix in another PR (if does not get auto-magically fixed).

PR #3478, opened 2025-03-10
RF: Small variable mismatch in the tutorial for DAM classifier

Atharva-Shah-2298:
Earlier version had a mismatch of the variable wrongly used. Results remain identical but skull stripped data should have been passed to denoiser instead of raw data. 

dam_updated


skoudoro (approved):
Make sense. Good catch. Thank you for this!

merging

PR #3486, opened 2025-03-12
RF: Fixed Warnings of Patch2self3

iprajwalreddy:
- changed the alpha value from 0 to 1e-10 for performing ridge_regression which is faster than using ols. This is done when the model parameter is 'ols'.


skoudoro (approved):
Thank you for this update @iprajwalreddy!

merging

PR #3489, opened 2025-03-14
DOC - Renamed fast tracking example

gabknight:
- Change default PTT angle to `max_angle=10`. It my test it provide better results on human data. `max_angle=20` is better on disco to allow sharper turns.
- Renamed the fast tracking example to `tracking_disco_phantom.py`. Since we are now using the fast tracking module in other examples, I think it is better to focus this example on the disco data.

The example should be extented to include other tracking methods later.


skoudoro (approved):
I just tested it and all good, merging

PR #3490, opened 2025-03-14
ENH: Updated the logging info

maharshi-gor:
- Replaced all the print statements with logging in viz.
- Updated the warnings.warn to logging.warning for mild warnings.
- Test cases fixed for the same.

image


arokem:
+1 to replacing the print statement, but I think that the warnings should remain as warnings, if they are to be caught by our (and other software's) tests, no?

dipy/viz/horizon/tab/base.py
```diff
@@ -162,7 +163,7 @@ def __init__(
                 "Images are of different dimensions, "
                 + "synchronization of slices will not work"
             )
-            warnings.warn(msg, stacklevel=2)
+            logging.warning(msg, stacklevel=2)
```
skoudoro, line 166: stacklevel ?

maharshi-gor:
> +1 to replacing the print statement, but I think that the warnings should remain as warnings, if they are to be caught by our (and other software's) tests, no?

Hey @arokem , Thank you for the quick feedback. We changed these warnings as they do not define any change in the workflow they are mild in sense do not affect the visualization. Also, these changes are tested on the `test_util.py` file which make sure that the function is behaving in the expected manner.

Discussed with @skoudoro as those warning were looking scary for non-tech savvy users.

dipy/viz/horizon/tab/base.py
maharshi-gor: @skoudoro updated the stacklevel as per the discussion.

PR #3492, opened 2025-03-15
BF: Peaks working with world coordinates.

maharshi-gor:
- The peaks were not working with the world coordinates after the changes for the fast peaks.
Screenshot 2025-03-14 at 10 44 53 PM


skoudoro:
it would be great to have test for this @maharshi-gor, Horizon have less and less test which is not good.

PR #3493, opened 2025-03-16
NF: allow broadcasting in dipy_math

skoudoro:
This is a small PR to allow unsafe use of dipy_math via a flag, and also, to allow broadcasting. 

Array with different dimension can be used. a `None` dimension will be added to the smallest one automatically. 

all of this is possible only by using the flag `--disable_check`

### DIFF
--- a/dipy/workflows/io.py
+++ b/dipy/workflows/io.py
@@ -2,6 +2,7 @@ import importlib
 from inspect import getmembers, isfunction
 import logging
 import os
+import re
 import sys
 import warnings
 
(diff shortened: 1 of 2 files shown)


maharshi-gor:
Hi @sreekarchigurupati can you check this PR? It would be a great help as you were asking for the broadcasting in the `dipy_math` so we can test the usecase you were trying as well.

Thanks!

maharshi-gor (approved):
LGTM!

Working on local as expected!

PR #3495, opened 2025-03-17
BF: Fixing minor issue in dipy_classify_tissue dam option

pjsjongsung:
This PR fixes an issue where dipy_classify_tissue workflow with the dam option returns a float type array as output.
Instead, it returns it as np.int32 array now.
This solves problems when visualizing the labels on the image the labels had floating points, due to interpolation.

Some documentation errors in patch2self has been addressed as well.


skoudoro (approved):
LGTM, thank you for that and the cleaning. merging

PR #3513, opened 2025-03-20
STYLE: Log message instead of printing orphan string

jhlegarreta:
Log message instead of printing an orphan string to the standard output in `FetchFlow`.

### DIFF
--- a/dipy/workflows/io.py
+++ b/dipy/workflows/io.py
@@ -269,7 +269,7 @@ class FetchFlow(Workflow):
                 available_data[data_name]()
 
             nb_success = len(data_names) - len(skipped_names)
-            print("\n")
+            logging.info("\n")
             logging.info(f"Fetched {nb_success} / {len(data_names)} Files ")
             if skipped_names:
                 logging.warn(f"Skipped data name(s): {' '.join(skipped_names)}")


jhlegarreta:
CI failures are unrelated, e.g.
https://github.com/dipy/dipy/actions/runs/13980725193/job/39145181598?pr=3513#step:11:4341
https://github.com/dipy/dipy/actions/runs/13980725185/job/39145180420?pr=3513#step:11:6890

dipy/workflows/io.py
```diff
@@ -269,7 +269,7 @@ def run(self, data_names, out_dir=""):
                 available_data[data_name]()
 
             nb_success = len(data_names) - len(skipped_names)
-            print("\n")
```
skoudoro, line 272: I believe this was done to have an empty line.

using `logging.info` will show `INFO:`

Not sure what is the good way to do it, but I do not think it is equivalent

skoudoro (commented):
Hi @jhlegarreta,

Not sure I can merge this, see my comment below

dipy/workflows/io.py
skoudoro: I do not think so, but it needs to be tested. 
let's find a good solution instead of a temporary one to fix this issue.

skoudoro: it seems you need a custom handler. 

```
class CustomHandler(logging.Handler):
    def emit(self, record):
        # Check if the log message is an empty string
        if record.getMessage() == "":
            # Directly write an empty line to the log stream
            self.stream.write("\n")
        else:
            # Use the default formatter for non-empty messages
            msg = self.format(record)
            self.stream.write(msg + "\n")

# Create a custom handler
custom_handler = CustomHandler()
custom_handler.setLevel(logging.INFO)
# Add the custom handler to the root logger
logging.getLogger().addHandler(custom_handler)
```

then `logging.info("")` should do the job. 

What do you think?

jhlegarreta: I am fine with the solution you propose. Should this apply to the entire code base? Where do you propose to put this?

skoudoro: After talking to Ariel, this should apply to the entire codebase. We can have a logging module in `dipy.utils` or `dipy.core` where all the custom logging function can live.

Thank you again @jhlegarreta

jhlegarreta: PR #3592. @skoudoro Please change it as you believe it is more convenient. I cannot devote more time to it.

jhlegarreta:
Closing in favor of PR #3592.

Issue #3521, opened 2025-03-22
Division by zero when machine has a single CPU


sanvila:
Hello. While building the Debian package for dipy on a single-CPU system I got a division by zero error while running the tests:

```
            data_to_fit = data[np.where(mask)]
            if weights_is_array:
                weights_to_fit = weights[np.where(mask)]
            single_voxel_with_self = partial(single_voxel_fit, self)
            n_jobs = kwargs.get("n_jobs", multiprocessing.cpu_count() - 1)
            vox_per_chunk = kwargs.get(
>               "vox_per_chunk", np.max([data_to_fit.shape[0] // n_jobs, 1])
            )
E           ZeroDivisionError: integer division or modulo by zero

dipy/reconst/multi_voxel.py:105: ZeroDivisionError
```

The following patch was applied in Debian:

```
--- a/dipy/reconst/multi_voxel.py
+++ b/dipy/reconst/multi_voxel.py
@@ -100,7 +100,7 @@
             if weights_is_array:
                 weights_to_fit = weights[np.where(mask)]
             single_voxel_with_self = partial(single_voxel_fit, self)
-            n_jobs = kwargs.get("n_jobs", multiprocessing.cpu_count() - 1)
+            n_jobs = kwargs.get("n_jobs", max(multiprocessing.cpu_count() - 1, 1))
             vox_per_chunk = kwargs.get(
                 "vox_per_chunk", np.max([data_to_fit.shape[0] // n_jobs, 1])
             )

```

Thanks.

PR #3522, opened 2025-03-22
FIX: Avoid division by zero on single-CPU systems (issue #3521)

sanvila:
(Explanation in issue #3521)


skoudoro (approved):
Good catch, thank you for this @sanvila, merging

PR #3527, opened 2025-03-31
BF: Search bar should come bigger in the center.

maharshi-gor:
- Removed search-field.html from navbar_end.
- Added search-button-field.html from navbar_end.
- It provides a button which looks like the search field and opens up the search field in the screen.
- The shortcut ctrl + k would do the same.
Screenshot from 2025-03-31 16-57-57

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -170,7 +170,11 @@ html_theme_options = {
     },
     "check_switcher": False,
     "show_version_warning_banner": True,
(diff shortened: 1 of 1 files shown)


jhlegarreta (approved):
Build failures are unrelated. Have not tried, but the screenshot looks good and it is an improvement over the current tiny search box.

skoudoro (approved):
All good, I just tried it, and works as expected.

Merging. Thanks @maharshi-gor

PR #3531, opened 2025-04-07
CI: Ignore specific cvxpy warnings to avoid CI failure.

skoudoro:
CVXPY generates warning that we can not control. this PR ignore them.  they should be fixed on a new CVXPY release. 
we might need to keep this until our minimal version of CVXPY contain this update.

See the warnings below not due to our codebase:
```
reconst/tests/test_forecast.py: 1 warning
reconst/tests/test_ivim.py: 3 warnings
reconst/tests/test_mapmri.py: 3 warnings
reconst/tests/test_mcsd.py: 3 warnings
reconst/tests/test_shore.py: 2 warnings
workflows/tests/test_reconst_mapmri.py: 320 warnings
  /Users/runner/work/dipy/dipy/venv/lib/python3.10/site-packages/osqp/interface.py:290: DeprecationWarning: "polish" is deprecated. Please use "polishing" instead.
    warnings.warn(

reconst/tests/test_forecast.py: 1 warning
reconst/tests/test_ivim.py: 3 warnings
reconst/tests/test_mapmri.py: 3 warnings
reconst/tests/test_mcsd.py: 3 warnings
reconst/tests/test_shore.py: 2 warnings
workflows/tests/test_reconst_mapmri.py: 320 warnings
  /Users/runner/work/dipy/dipy/venv/lib/python3.10/site-packages/osqp/interface.py:405: PendingDeprecationWarning: The default value of raise_error will change to True in the future.
    warnings.warn(

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=================================== Werrors ====================================
Warnings as errors: Activated.
664 warnings were raised and treated as errors.
```

see also https://github.com/cvxpy/cvxpy/issues/2770

PR #3533, opened 2025-04-08
CI: introduce cached Data.

skoudoro:
This PR introduce the following feature:

- Before running any test, we download test data and cache them. Then they are restored when we need them.
- Merge multiple workflows into one. Not necessary to split them anymore since they can be restarted individually.

This should reduce the number of download so speed up the whole process and avoid most of our 404 error.

this PR should failed until #3531 is merged. When #3531, I will rebase and all should be ok.


skoudoro:
TODO:  All data are not needed, check which one are use for our tests to speedup the process and reduce the size.

skoudoro:
Ok, this PR is ready to go!  the failure CI's are not related. 

I will look at the warning below in a new PR:
```
  /Users/runner/work/dipy/dipy/venv/lib/python3.13/site-packages/joblib/externals/loky/backend/fork_exec.py:38: DeprecationWarning: This process (pid=4517) is multi-threaded, use of fork() may lead to deadlocks in the child.
    pid = os.fork()
```

Issue #3536, opened 2025-04-11
FA latex formula does not show correctly on DTI tutorial


Garyfallidis:
https://docs.dipy.org/stable/examples_built/reconstruction/reconst_dti.html#sphx-glr-examples-built-reconstruction-reconst-dti-py

Please also explain that the eigenvalues of the tansor are $\lambda_1, \lambda_2, \lambda_3$.

skoudoro:
fixed by #3538, closing

PR #3538, opened 2025-04-11
RF: Fixed the latex

deka27:
This fix is solving issue with latex mentioned here. https://github.com/dipy/dipy/issues/3536

image

to

image


skoudoro:
Thank you @deka27. Looks good, merged

PR #3541, opened 2025-04-16
RF: Deprecate autocrop in median_otsu

### LINKED ISSUE #3537: Remove autocrop from median_otsu
Autocrop should be its own command.

skoudoro:
This PR deprecate `autocrop` from `median_otsu` function. 

this feature brings a lot of errors and confusion when the  voxel size is not isotropic and size (1,1,1). You endup with a image with a different size of the original image and the mask.  it exists a crop function if the user wishes to do so.

fixes  #3537

### DIFF
--- a/dipy/segment/mask.py
+++ b/dipy/segment/mask.py
@@ -11,6 +11,7 @@ except Exception:
 from dipy.reconst.dti import color_fa, fractional_anisotropy
 from dipy.segment.utils import remove_holes_and_islands
 from dipy.testing.decorators import warning_for_keywords
+from dipy.utils.deprecator import deprecated_params
 
 
 def multi_median(data, median_radius, numpass):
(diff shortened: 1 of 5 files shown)


arokem:
Looks good, but the warning probably needs to be added to the filterwarnings in pyproject.toml?

skoudoro:
> Looks good, but the warning probably needs to be added to the filterwarnings in pyproject.toml?

Thank you for the feedback, I will clean that tomorrow. I do not think it needs to be in pyproject.toml since we can control it

pjsjongsung:
Overall looks good. I'm just wondering if there should be some explanation in the functions doc on what the behavior would be after deprecation of the `out_masked` and `save_masked` parameter? Like, is the function never going to save the masked output anymore or is it always going to save? If saved by default, what would be the name from now on then?

skoudoro:
After checking more in details, the deprecation of the `out_masked` and `save_masked` parameters is not needed.

only `autocrop` parameter is needed and we do not need to add a note. 

All updated and ready to go

PR #3542, opened 2025-04-18
RF: Editing tranform_img in dipy.nn.utils

pjsjongsung:
The transform_img function did not have an option to make an image isotropic and have a fixed size.
This adds a way of doing this by padding to a multiple of that shape and resizing. Recover_img has been changed to compensate for the difference as well.
The parameters returned by transform_img, which was only used in recover_img, was simplified to make the code cleaner. 
Deepn4 strangely had different results between the old transform_img and new, but the output when checked did not have meaningful differences. Hence, test output was edited to make the test functional.

### DIFF
--- a/dipy/nn/tf/evac.py
+++ b/dipy/nn/tf/evac.py
@@ -446,28 +446,22 @@ class EVACPlus:
             batch_size = 1
 
         input_data = np.zeros((128, 128, 128, len(T1)))
-        affines = np.zeros((len(T1), 4, 4))
-        mid_shapes = np.zeros((len(T1), 3)).astype(int)
-        offset_arrays = np.zeros((len(T1), 4, 4)).astype(int)
-        scales = np.zeros(len(T1))
-        crop_vss = np.zeros((len(T1), 3, 2))
-        pad_vss = np.zeros((len(T1), 3, 2))
+        params_list = []
 
         # Normalize the data.
         n_T1 = np.zeros(T1.shape)
         for i, T1_img in enumerate(T1):
             n_T1[i] = normalize(T1_img, new_min=0, new_max=1)
-            t_img, t_affine, mid_shape, offset_array, scale, crop_vs, pad_vs = (
-                transform_img(n_T1[i], affine[i], voxsize=voxsize[i])
+            t_img, params = transform_img(
+                n_T1[i],
+                affine[i],
+                voxsize=voxsize[i],
+                ratio=2,
+                init_shape=(256, 256, 256),
+                target_voxsize=(1, 1, 1),
             )
             input_data[..., i] = t_img
-            affines[i] = t_affine
-            mid_shapes[i] = mid_shape
-            offset_arrays[i] = offset_array
-            scales[i] = scale
-            crop_vss[i] = crop_vs
-            pad_vss[i] = pad_vs
-
+            params_list.append(params)
         # Prediction stage
         prediction = np.zeros((len(T1), 128, 128, 128), dtype=np.float32)
         for batch_idx in range(batch_size, len(T1) + 1, batch_size):
@@ -483,17 +477,7 @@ class EVACPlus:
 
         output_mask = []
         for i in range(len(T1)):
-            output = recover_img(
-                prediction[i],
-                affines[i],
-                mid_shapes[i],
-                n_T1[i].shape,
-                offset_arrays[i],
-                voxsize=voxsize[i],
-                scale=scales[i],
-                crop_vs=crop_vss[i],
-                pad_vs=pad_vss[i],
-            )
+            output, _ = recover_img(prediction[i], params_list[i])
             if not return_prob:
                 output = np.where(output >= 0.5, 1, 0)
                 if finalize_mask:
--- a/dipy/nn/torch/deepn4.py
+++ b/dipy/nn/torch/deepn4.py
@@ -330,8 +330,13 @@ class DeepN4:
             The volume has matching shape to the input data
         """
         # Preprocess input data (resample, normalize, and pad)
-        resampled_T1, inv_affine, mid_shape, offset_array, scale, crop_vs, pad_vs = (
-            transform_img(img, img_affine, voxsize=voxsize)
+        resampled_T1, params = transform_img(
+            img,
+            img_affine,
+            voxsize=voxsize,
+            ratio=2,
+            init_shape=(256, 256, 256),
+            target_voxsize=(1, 1, 1),
         )
         (in_features, lx, lX, ly, lY, lz, lZ, rx, rX, ry, rY, rz, rZ, in_max) = (
             self.load_resample(resampled_T1)
@@ -349,18 +354,7 @@ class DeepN4:
         )
         final_field[rx:rX, ry:rY, rz:rZ] = field[lx:lX, ly:lY, lz:lZ]
         final_fields = gaussian_filter(final_field, sigma=3)
-        upsample_final_field = recover_img(
-            final_fields,
-            inv_affine,
-            mid_shape,
-            img.shape,
-            offset_array,
-            voxsize,
-            scale,
-            crop_vs,
-            pad_vs,
-        )
-
+        upsample_final_field, _ = recover_img(final_fields, params)
         # Correct the image
         below_threshold_mask = np.abs(upsample_final_field) < threshold
         with np.errstate(divide="ignore", invalid="ignore"):
--- a/dipy/nn/torch/evac.py
+++ b/dipy/nn/torch/evac.py
@@ -484,28 +484,22 @@ class EVACPlus:
             batch_size = 1
 
         input_data = np.zeros((128, 128, 128, len(T1)))
-        affines = np.zeros((len(T1), 4, 4))
-        mid_shapes = np.zeros((len(T1), 3)).astype(int)
-        offset_arrays = np.zeros((len(T1), 4, 4)).astype(int)
-        scales = np.zeros(len(T1))
-        crop_vss = np.zeros((len(T1), 3, 2))
-        pad_vss = np.zeros((len(T1), 3, 2))
+        params_list = []
 
         # Normalize the data.
         n_T1 = np.zeros(T1.shape)
         for i, T1_img in enumerate(T1):
             n_T1[i] = normalize(T1_img, new_min=0, new_max=1)
-            t_img, t_affine, mid_shape, offset_array, scale, crop_vs, pad_vs = (
-                transform_img(n_T1[i], affine[i], voxsize=voxsize[i])
+            t_img, params = transform_img(
+                n_T1[i],
+                affine[i],
+                voxsize=voxsize[i],
+                ratio=2,
+                init_shape=(256, 256, 256),
+                target_voxsize=(1, 1, 1),
             )
             input_data[..., i] = t_img
-            affines[i] = t_affine
-            mid_shapes[i] = mid_shape
-            offset_arrays[i] = offset_array
-            scales[i] = scale
-            crop_vss[i] = crop_vs
-            pad_vss[i] = pad_vs
-
+            params_list.append(params)
         # Prediction stage
         prediction = np.zeros((len(T1), 128, 128, 128), dtype=np.float32)
         for batch_idx in range(batch_size, len(T1) + 1, batch_size):
@@ -521,17 +515,7 @@ class EVACPlus:
 
         output_mask = []
         for i in range(len(T1)):
-            output = recover_img(
-                prediction[i],
-                affines[i],
-                mid_shapes[i],
-                n_T1[i].shape,
-                offset_arrays[i],
-                voxsize=voxsize[i],
-                scale=scales[i],
-                crop_vs=crop_vss[i],
-                pad_vs=pad_vss[i],
-            )
+            output, _ = recover_img(prediction[i], params_list[i])
             if not return_prob:
                 output = np.where(output >= 0.5, 1, 0)
                 if finalize_mask:
--- a/dipy/nn/utils.py
+++ b/dipy/nn/utils.py
@@ -3,6 +3,7 @@ from scipy.ndimage import affine_transform
 
 from dipy.align.reslice import reslice
 from dipy.testing.decorators import warning_for_keywords
+from dipy.utils.deprecator import deprecated_params
 
 
 @warning_for_keywords()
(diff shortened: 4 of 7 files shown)


dipy/nn/utils.py
```diff
@@ -80,45 +81,66 @@ def set_logger_level(log_level, logger):
...
-    considered_points : str, optional
-        which points to consider when calculating
-        the boundary of the image. If there is shearing
-        in the affine, 'all' might be more accurate
-    init_shape : list, tuple or numpy array (3,), optional
-        Initial shape to transform the image to
-    scale : float, optional
-        How much we want to scale the image
+        Affine matrix provided by the image file.
+    voxsize : tuple (3,)
```
skoudoro, line 106: `, optional`

dipy/nn/utils.py
```diff
@@ -80,45 +81,66 @@ def set_logger_level(log_level, logger):
...
+        Voxel size provided by the image file.
+    target_voxsize : tuple (3,)
```
skoudoro, line 108: `, optional`

dipy/nn/utils.py
```diff
@@ -80,45 +81,66 @@ def set_logger_level(log_level, logger):
...
+        The voxel size we want to start from.
+        Ignored if need_isotropic is True.
+    considered_points : str
```
skoudoro, line 111: `, optional`

dipy/nn/utils.py
```diff
@@ -80,45 +81,66 @@ def set_logger_level(log_level, logger):
...
+        Considered points when calculating the transformed shape.
+        \"corners\" will consider only corners of the image shape.
+        \"all\" will consider all voxels. Might be needed when shearing is applied.
+    init_shape : tuple (3,)
```
skoudoro, line 115: `, optional`

dipy/nn/utils.py
```diff
@@ -80,45 +81,66 @@ def set_logger_level(log_level, logger):
...
+        What we want the initial shape to be before last resizing step.
+        Ignored if need_isotropic is True.
+    ratio : float
```
skoudoro, line 118: `, optional`

dipy/nn/utils.py
```diff
@@ -80,45 +81,66 @@ def set_logger_level(log_level, logger):
...
+        The ratio of change in the last resizing step.
+        Ignored if need_isotropic is True.
+    set_size : tuple (3,)
```
skoudoro, line 121: `, optional`
weird variable name...

dipy/nn/utils.py
```diff
@@ -80,45 +81,66 @@ def set_logger_level(log_level, logger):
...
+        The final size of the image array.
+    need_isotropic : bool
```
skoudoro, line 123: `, optional`

dipy/nn/utils.py
```diff
@@ -165,82 +189,132 @@ def transform_img(
...
+        offset_array,
+        crop_vs,
+        pad_vs,
+        ratio,
+        voxsize,
+        target_voxsize,
+        need_isotropic,
+        set_size,
+        ori_shape,
+    ) = params
```
skoudoro, line 271: this unpack will fail is params have a wrong size. Can you check that?

skoudoro (commented):
Hi @pjsjongsung,

Thank you for this, I need to look deeper, a lot of conditional directive meaning a lot of corner case so I need to understand.

I will keep you updated

pjsjongsung:
I'll make the necessary changes thank you for the review @skoudoro . We can have a quick zoom later if you need help understanding what I wrote here. I had a lot of corner cases because I had to make it work with the already implemented deep learning models but also work on the new one.

skoudoro:
Hi @pjsjongsung,

Will you have time to come back on this ? What is the update ?

dipy/nn/utils.py
```diff
@@ -80,45 +81,71 @@ def set_logger_level(log_level, logger):
...
+
+
 
     Returns
     -------
-    tuple
-        Tuple with variables for recover_img
+    new_image : np.ndarray
+        Transformed image to be used in the Deep Learning model.
+    params : tuple
```
maharshi-gor: can we return params as dict. It is much safer comparison to tuple which can be really hard to remember the order of the params.

dipy/nn/utils.py
```diff
@@ -80,45 +81,71 @@ def set_logger_level(log_level, logger):
...
+        Parameters that are used when recovering the original image space.
     """
-    if voxsize is not None and np.any(voxsize != np.ones(3)):
-        image2, affine2 = reslice(image, affine, voxsize, (1, 1, 1))
+    ori_shape = image.shape
+    if need_isotropic:
+        target_voxsize = tuple(np.max(voxsize) * np.ones(3))
+    if target_voxsize is not None and np.any(target_voxsize != np.ones(3)):
+        image2, affine2 = reslice(image, affine, voxsize, target_voxsize)
```
maharshi-gor: `resliced_img`,` resliced_affine` can be better names of the variables.

dipy/nn/utils.py
```diff
@@ -80,45 +81,71 @@ def set_logger_level(log_level, logger):
...
     else:
         image2 = image.copy()
         affine2 = affine.copy()
+
     shape = image2.shape
 
     if considered_points == "corners":
```
maharshi-gor, line 102: Let's create a method that only focuses on returning corners. and consider better naming convention for the inner variables.

maharshi-gor (commented):
Thank you for the work @pjsjongsung.

Please check the comments mentioned. I feel we should find a way to break down the large method to be more manageable chunks.

pjsjongsung:
Changed! I don't think the errors are from my side, let me know if there's more change needed

maharshi-gor:
Thank you for the update. I will try and let you know if I find something.

dipy/nn/utils.py
```diff
@@ -135,112 +117,264 @@ def transform_img(
...
+        output_shape=tuple(new_shape),
+        output=new_image,
+        order=order,
     )
 
-    new_image, pad_vs, crop_vs = pad_crop(new_image, init_shape)
-
-    if scale != 1:
-        new_image, _ = reslice(new_image, np.eye(4), (1, 1, 1), (scale, scale, scale))
+    mid_image = new_image.copy()
```
sukrutakolkar: `mid_image` is used only in the outer-most else block below with no changes to `new_image` which makes this copy redundant, perhaps we could use `new_image` as is?

sukrutakolkar:
Tested it with some T1s and compared outputs against the previous version. Looks good.

dipy/nn/tf/evac.py
```diff
@@ -442,28 +442,22 @@ def predict(
...
         for i, T1_img in enumerate(T1):
             n_T1[i] = normalize(T1_img, new_min=0, new_max=1)
-            t_img, t_affine, mid_shape, offset_array, scale, crop_vs, pad_vs = (
-                transform_img(n_T1[i], affine[i], voxsize=voxsize[i])
+            t_img, params = transform_img(
+                n_T1[i],
+                affine[i],
+                voxsize=voxsize[i],
+                ratio=2,
+                init_shape=(256, 256, 256),
```
skoudoro: why `init_shape` is static ?  why (256, 256, 256) ?  what happens if we want another input_size ?

dipy/nn/torch/deepn4.py
```diff
@@ -326,8 +326,13 @@ def predict(self, img, img_affine, *, voxsize=(1, 1, 1), threshold=0.5):
             The volume has matching shape to the input data
         """
         # Preprocess input data (resample, normalize, and pad)
-        resampled_T1, inv_affine, mid_shape, offset_array, scale, crop_vs, pad_vs = (
-            transform_img(img, img_affine, voxsize=voxsize)
+        resampled_T1, params = transform_img(
+            img,
+            img_affine,
+            voxsize=voxsize,
```
skoudoro: not needed, same comment as above

dipy/nn/torch/evac.py
```diff
@@ -480,28 +480,22 @@ def predict(
...
         # Normalize the data.
         n_T1 = np.zeros(T1.shape)
         for i, T1_img in enumerate(T1):
             n_T1[i] = normalize(T1_img, new_min=0, new_max=1)
-            t_img, t_affine, mid_shape, offset_array, scale, crop_vs, pad_vs = (
-                transform_img(n_T1[i], affine[i], voxsize=voxsize[i])
+            t_img, params = transform_img(
+                n_T1[i],
+                affine[i],
+                voxsize=voxsize[i],
```
skoudoro: not needed, same comment as above

dipy/nn/utils.py
```diff
@@ -135,112 +117,262 @@ def transform_img(
...
+    return min_bounds, max_bounds
+
+
+@warning_for_keywords()
+@deprecated_params("scale", new_name="ratio", since="1.12", until="1.14")
+def transform_img(
+    image,
+    affine,
+    *,
+    voxsize=(1, 1, 1),
```
skoudoro: same as above, not needed

dipy/nn/utils.py
```diff
@@ -135,112 +117,262 @@ def transform_img(
...
+    Returns
+    -------
+    new_image : np.ndarray
+        Transformed image to be used in the Deep Learning model.
+    params : dict
+        Parameters that are used when recovering the original image space.
+    """
+    ori_shape = image.shape
+    if need_isotropic:
+        target_voxsize = tuple(np.max(voxsize) * np.ones(3))
```
skoudoro: if you have a `target_voxsize= (2, 2, 2)` and `need_isotropic=True`,  it means that my request is ignored. Something wrong here, it does not make sense.
Yes, in both case, isotropic, but not my desired voxel size for whatever reason

skoudoro:
@sukrutakolkar and @sreekarchigurupati , Can you share the data to do the test. thanks !

dipy/nn/tf/evac.py
pjsjongsung: I have it in default value because in deep learning most of the time we want a 128 128 128 image which is scaled down from 256 256 256. But if we, let's say, gave 260, 260, 260 with ratio of 2, it will return 130, 130, 130.

pjsjongsung: If we do direct resizing, the image will not be resized using the ratio on equally on each axis.

dipy/nn/utils.py
pjsjongsung: Those are two different arguments, because need_isotropic means it will auto calculate an isotropic voxel size that loses the minimum information. Those look the same because we are passing an isotropic voxsize for target_voxsize, but if we didn't, they will have very different behaviors.

dipy/nn/torch/deepn4.py
pjsjongsung: Yes, but one can also not pass voxsize (not care which voxsize the image is). If we do it that way, by calculating voxsize from affine, we are always assuming the user wants to use the voxsize information.

sukrutakolkar:
@skoudoro I used T1s from dipy.data - stanford_t1, tissue_data (2 T1s) and syn_data

pjsjongsung:
I will need to change the reslice function a bit to make the shape correspond after recover_img function.
My plan is to have an optional parameter in reslice that can throw the exact fractional shape (instead of rounded integer shape), which can be accurately used in recovering the original image translation.

pjsjongsung:
I will move onto synthseg once this PR is merged. I think it is ready for review

dipy/nn/tf/evac.py
```diff
@@ -92,7 +92,7 @@ def prepare_img(image):
 
 
 class Block(Layer):
-    @warning_for_keywords()
+    # @warning_for_keywords()
```
skoudoro: to put it back

dipy/nn/utils.py
```diff
@@ -135,112 +112,204 @@ def transform_img(
...
-    transformed_corners = (affine2 @ corners.T).T
+    transformed_corners = (affine @ points.T).T
     min_bounds = transformed_corners.min(axis=0)[:3]
     max_bounds = transformed_corners.max(axis=0)[:3]
 
+    return min_bounds, max_bounds
+
+
+@warning_for_keywords()
+# @deprecated_params("scale", new_name="ratio", since="1.12", until="1.14")
```
skoudoro: to remove

dipy/nn/utils.py
```diff
@@ -135,112 +112,204 @@ def transform_img(
...
-    ori_shape,
-    offset_array,
-    voxsize,
-    scale,
-    crop_vs,
-    pad_vs,
-):
+    # Calculate how much crop and pad is needed
+    # to get to target_voxsize and final_size
+    # with the current new_shape and voxsize
```
skoudoro: to remove

dipy/nn/utils.py
```diff
@@ -292,3 +361,43 @@ def pad_crop(image, target_shape):
...
+    """
+    Function to figure out pad and crop range
+    to fit the target shape with the image
+
+    Parameters
+    ----------
+    image : np.ndarray
+        Target image
+    target_shape : (3,)
+        Target shape
```
skoudoro: misalignment of parameter

dipy/nn/utils.py
```diff
@@ -79,50 +75,31 @@ def set_logger_level(log_level, logger):
...
-def transform_img(
-    image,
-    affine,
-    *,
-    voxsize=None,
-    considered_points="corners",
-    init_shape=(256, 256, 256),
-    scale=2,
-):
+def get_bounds(shape, affine, considered_points="corners"):
```
skoudoro: keywords only `*`

dipy/nn/utils.py
```diff
@@ -135,112 +112,204 @@ def transform_img(
...
+    considered_points : str, optional
+        Considered points when calculating the transformed shape.
+        \"corners\" will consider only corners of the image shape.
+        \"all\" will consider all voxels. Might be needed when shearing is applied.
+    init_shape : tuple (3,), optional
+        What we want the initial shape to be before last resizing step.
+        Ignored if need_isotropic is True.
+    ratio : float, optional
+        The ratio of change in the last resizing step.
+        Ignored if need_isotropic is True.
```
skoudoro: to remove

dipy/nn/utils.py
```diff
@@ -135,112 +112,204 @@ def transform_img(
...
+    affine : np.ndarray
+        Affine matrix provided by the image file.
+    target_voxsize : tuple (3,), optional
+        The voxel size we want to start from.
+        If none is provided, we calculate the closest isotropic voxel size.
+    considered_points : str, optional
+        Considered points when calculating the transformed shape.
+        \"corners\" will consider only corners of the image shape.
+        \"all\" will consider all voxels. Might be needed when shearing is applied.
+    init_shape : tuple (3,), optional
```
skoudoro: to remove

dipy/nn/utils.py
```diff
@@ -135,112 +112,204 @@ def transform_img(
...
-    scale : float
-        Scale used in transform_img
-    crop_vs : np.ndarray (3,2)
-        crop range used in transform_img
-    pad_vs : np.ndarray (3,2)
-        pad range used in transform_img
+        Image to recover.
+    params : tuple
+        Parameters for recover_img function.
+        Returned from transform_img.
```
skoudoro, line 248: missing order

skoudoro:
All good, thank you @pjsjongsung ! merging

Issue #3540, opened 2025-04-16
Tractography with GFA as stopping criterion is not reproducible across runs


jhlegarreta:
While playing with a very simple script based on the example:
https://docs.dipy.org/stable/examples_built/fiber_tracking/tracking_probabilistic.html#sphx-glr-examples-built-fiber-tracking-tracking-probabilistic-py

I realize that the tractography output is 100% reproducible across runs when using the binary stopping criterion,.

However, when I switch to the GFA-based stopping criterion, the result (tractogram) is not reproducible.

The script is simply this part of the example:
https://github.com/dipy/dipy/blob/b27a47b7743c830af8d9be0ea3dd1abb4c317a6e/doc/examples/tracking_probabilistic.py#L23-L81

and the use of the GFA-based criterion just changes:
https://github.com/dipy/dipy/blob/b27a47b7743c830af8d9be0ea3dd1abb4c317a6e/doc/examples/tracking_probabilistic.py#L52

to:
```
csa_model = CsaOdfModel(gtab, sh_order_max=sh_order_max)
gfa = csa_model.fit(dwi_data, mask=white_matter).gfa
sc = ThresholdStoppingCriterion(gfa, 0.25)
```

jhlegarreta:
An offline discussion with @gabknight and @skoudoro has cast light in this: as found out by @gabknight, results are reproducible across runs in single-thread mode when using GFA as a stopping criterion, but are not in multi-thread. @skoudoro suggested and @gabknight confirmed that the issue lies in the `self.interp_out_view` array being modified in parallel by the multiple threads:

![Image](https://github.com/user-attachments/assets/8b73bd8a-ceac-41ad-9543-754a0cddaec8)

So any stopping criterion inheriting from this class will be affected.

Thanks both for the investigation.

Issue #3439, opened 2025-01-14
data fetching: problem with UW server ?


skoudoro:
Recently, we have many issues with data fetching. it seems that data in the `https://digital.lib.washington.edu/researchworks` are not anymore accessible. we got often `404 Not found`. Sometimes, they reappear few hours.

Here some link to test:
- https://digital.lib.washington.edu/researchworks/bitstream/handle/1773/38475/HARDI193.nii.gz
- https://digital.lib.washington.edu/researchworks/bitstream/handle/1773/38476/t1.nii.gz

That's why we have this kind of error below in the CI's:
```python
FAILED align/tests/test_api.py::test_affine_registration - urllib.error.HTTPError: HTTP Error 404: Not Found
FAILED align/tests/test_api.py::test_register_dwi_series_multi_b0 - urllib.error.HTTPError: HTTP Error 404: Not Found
```

Do you have any clue @arokem?

 We might have to review our data storage strategy.

arokem:
I am looking into this

arokem:
It seems that the data is still there, but ResearchWorks changed the way that the data is represented in their database. Maybe we should move this all to figshare? And them make a quick bug-fix release with that change?

arokem:
To be clear, we can still get the data through the browser, but the URL location has changed: https://digital.lib.washington.edu/researchworks/items/a8b2acb0-54d3-4f2f-99b8-d6e0f5a8a71e, and it is not clear whether the individual files have a URL anymore.

skoudoro:
figshare seems to be the only reliable long term solution. (I hope).

So I agree with this plan. Apart if you see any other alternatives.

It seems we have 8 datasets that needs migration.

skoudoro:
it seems we can download them individually when I look at this page: 

https://digital.lib.washington.edu/researchworks/items/a8b2acb0-54d3-4f2f-99b8-d6e0f5a8a71e/full

Maybe we should just update our links

arokem:
Update: UW libraries implementated a configuration change such that these files are now available through the URLs in DIPY. I verified this locally - and also running one of pyAFQ's nightly tests that returned a 404 yesterday: https://github.com/tractometry/pyAFQ/actions/runs/12824407639, so if that one comes back green this probably works. I've requested some more information about long-term support for these links, and I'll follow up here regarding whether we should move the data elsewhere for the long run, or if we can keep using these links.

skoudoro:
Great ! Thank you for the update and following up with that!

[...]

PR #3545, opened 2025-04-22
ENH: Add `StopingCriterion` reproducibility test

jhlegarreta:
Add a test to demonstrate the `StoppingCriterion` reproducibility across runs.

### DIFF
--- a/dipy/tracking/tests/test_stopping_criterion.py
+++ b/dipy/tracking/tests/test_stopping_criterion.py
@@ -127,6 +127,55 @@ def test_threshold_stopping_criterion(rng):
 
 
 @set_random_number_generator()
+def test_threshold_stopping_criterion_reproducibility(rng):
+    tissue_map = rng.random((4, 4, 4))
+
+    # Instantiate once
+    ttc1 = ThresholdStoppingCriterion(tissue_map.astype("float32"), 0.5)
+
+    # Instantiate again
+    ttc2 = ThresholdStoppingCriterion(tissue_map.astype("float32"), 0.5)
+
+    # Assert identical outputs
+
+    # Test voxel center
+    for ind in ndindex(tissue_map.shape):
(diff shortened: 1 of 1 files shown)


jhlegarreta:
@skoudoro the data fetching workflow has failed 3 times in a row, and hence, none of the testing CIs were triggered. Data fetching failures are unrelated to this patch set. Not sure what can be done.

skoudoro:
Sorry for that @jhlegarreta.  we have this issue https://github.com/dipy/dipy/issues/3439 concerning the data fetching. UW is trying to fix their server. 
we might need to migrate those data on figshare. but for now, I do not have a solution apart from retry.

jhlegarreta:
So it looks like the test is not demonstrating the issue:
https://github.com/dipy/dipy/actions/runs/14596139356/job/41048360402#step:12:3031
```
tracking/tests/test_stopping_criterion.py::test_threshold_stopping_criterion_reproducibility <- ../venv/lib/python3.10/site-packages/dipy/testing/decorators.py PASSED
```

I assume GHA CIs are multi-threaded. Have not found any related documentation, though. When I used a `ThresholdStoppingCriterion` in the #3540 case, I did not use any flag to signal that I wanted a multi-threaded execution. Any guess @skoudoro ?

skoudoro:
I think github action have 2 cores. We have other test using 1 core or 2 cores then compare.

Your test does use the multithreaded so it can not fail. Also, you create 2 Stopping criterion objects which is not the problem. 

The current problem is 1 stopping criterion object and multithread hitting this same object.

skoudoro:
after talking with @gabknight and @jhlegarreta,  simulating 2 threads that writes at the same time in the memory seems tricky with Github Action (not enough cores ?). For now, I recommend that we close this test.

It should be a way to do that, we need to search independently a good way to simulate threads writing in the same buffer.

PR #3546, opened 2025-04-22
BF: Make `StoppingCriterion` reproducible for multi-thread execution

### LINKED ISSUE #3540: Tractography with GFA as stopping criterion is not reproducible across runs

jhlegarreta:
Make `StoppingCriterion` reproducible for multi-thread execution: instantiate the resulting interpolation array at each call.

In single-thread mode the maps created by the criterion are reproducible; however, in multi-thread execution the `self.interp_out_view` is modified in parallel by the multiple threads, and since it is instantiated only once, the maps are not reproducible across runs.

This effect is only observed in criterion objects that use the `self.interp_out_view` array: i.e. it is not observed in the `BinaryStoppingCriterion` objects.

Fixes #3540.

### DIFF
--- a/dipy/tracking/stopping_criterion.pxd
+++ b/dipy/tracking/stopping_criterion.pxd
@@ -11,9 +11,6 @@ cpdef enum StreamlineStatus:
 
 
 cdef class StoppingCriterion:
-    cdef:
-        double interp_out_double[1]
-        double[::1] interp_out_view
 
     cpdef StreamlineStatus check_point(self, double[::1] point)
     cdef StreamlineStatus check_point_c(self, double* point, RNGState* rng=*) noexcept nogil
--- a/dipy/tracking/stopping_criterion.pyx
+++ b/dipy/tracking/stopping_criterion.pyx
@@ -29,7 +29,6 @@ cdef class BinaryStoppingCriterion(StoppingCriterion):
     """
 
     def __cinit__(self, mask):
-        self.interp_out_view = self.interp_out_double
         self.mask = (mask > 0).astype('uint8')
 
     cdef StreamlineStatus check_point_c(self, double* point, RNGState* rng=NULL) noexcept nogil:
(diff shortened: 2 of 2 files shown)


jhlegarreta:
Related to PR #3545: the latter should demonstrate the issue, and have it resolved using the fix in this PR.

skoudoro:
Hi @jhlegarreta,

Thank you for moving forward with this. it was on my todo list for this friday. 

Let me know if you need any help. You might have to update much more places.

jhlegarreta:
> Thank you for moving forward with this. it was on my todo list for this friday.

:+1:

> Let me know if you need any help. You might have to update much more places.

Tests are passing, though: i.e. the reproducibility issue is not demonstrated (re PR #3545), but the changes seem to be fine. So suggestions are welcome, as I am not very skilled in Cython yet.

Also, I first want to make sure that the test in PR #3545 does not pass.

skoudoro:
> Tests are passing, though ....

Indeed!  after a deeper look, your update is enough and cleanup the logic! 
Actually, I wonder why it was a class variable when it could have just been local. Thanks again!

skoudoro (approved):
LGTM. thank you for fixing this big issue @jhlegarreta, 

merging

PR #3553, opened 2025-05-26
ENH: Implement a predict() function for the ``GeneralizedQSamplingModel``

oesteban:
Uses a calculated ODF to simulate data.

cc/ @arokem, @jhlegarreta

X-Ref: nipreps/nifreeze#143

### DIFF
--- a/dipy/reconst/gqi.py
+++ b/dipy/reconst/gqi.py
@@ -81,15 +81,14 @@ class GeneralizedQSamplingModel(OdfModel, Cache):
     def fit(self, data, **kwargs):
         return GeneralizedQSamplingFit(self, data)
 
-    @warning_for_keywords()
     def predict(self, odf, gtab, sphere):
         """
-        Predict a signal for this TensorModel class instance given parameters.
+        Predict a signal for this GeneralizedQSamplingModel class instance given parameters.
 
         Parameters
         ----------
         odf : ndarray
-            Map of ODFs
+            Map of ODFs.
         gtab : ndarray
             Orientations where signal will be simulated
         sphere : :obj:`~dipy.core.sphere.Sphere`
@@ -137,7 +136,6 @@ class GeneralizedQSamplingFit(OdfFit):
         return np.dot(self.data, self.gqi_vector)
 
 
-@warning_for_keywords()
 def gqi_kernel(gtab, param_lambda, sphere, method="gqi2"):
     # 0.01506 = 6*D where D is the free water diffusion coefficient
     # l_values sqrt(6 D tau) D free water diffusion coefficient and
@@ -341,7 +339,7 @@ def odf_prediction(odf, gtab, param_lambda, sphere, method="gqi2"):
     odf : ndarray
         ODF parameters.
 
-    gtab : a GradientTable class instance
+    gtab : GradientTable
         The gradient table for this prediction
 
     Notes
@@ -352,7 +350,7 @@ def odf_prediction(odf, gtab, param_lambda, sphere, method="gqi2"):
 
         S(\theta, b) = K_{ii}^{-1} \cdot ODF
 
-    where $K_{ii}^{-1}$, is the inverse of the gQI kernels for the direction(s) $ii$
+    where $K_{ii}^{-1}$, is the inverse of the GQI kernels for the direction(s) $ii$
     given by ``gtab``.
 
     """
--- a/dipy/reconst/gqi.py
+++ b/dipy/reconst/gqi.py
@@ -4,13 +4,14 @@ import warnings
 
 import numpy as np
 
+from dipy.core.subdivide_octahedron import create_unit_sphere
 from dipy.reconst.cache import Cache
-from dipy.reconst.multi_voxel import multi_voxel_fit
 from dipy.reconst.odf import OdfFit, OdfModel
 from dipy.testing.decorators import warning_for_keywords
 from dipy.utils.logging import logger
 
 INVERSE_LAMBDA = 1e-6
+DEFAULT_SPHERE_RECURSION_LEVEL = 5
 
 
 class GeneralizedQSamplingModel(OdfModel, Cache):
@@ -18,7 +19,13 @@ class GeneralizedQSamplingModel(OdfModel, Cache):
 
     @warning_for_keywords()
     def __init__(
-        self, gtab, *, method="gqi2", sampling_length=1.2, normalize_peaks=False
+        self,
+        gtab,
+        *,
+        method="standard",
+        sampling_length=1.2,
+        normalize_peaks=False,
+        sphere=None,
     ):
         r"""Generalized Q-Sampling Imaging.
 
@@ -41,6 +48,9 @@ class GeneralizedQSamplingModel(OdfModel, Cache):
             diffusion sampling length (lambda in eq. 2.14 and 2.16)
         normalize_peaks : bool, optional
             True to normalize peaks.
+        sphere : :obj:`~dipy.core.sphere.Sphere`, optional
+            A sphere object, must be the same used for calculating the ODFs.
+            If not provided, a default sphere will be created.
 
         Notes
         -----
@@ -64,7 +74,7 @@ class GeneralizedQSamplingModel(OdfModel, Cache):
         >>> from dipy.core.subdivide_octahedron import create_unit_sphere
         >>> sphere = create_unit_sphere(recursion_level=5)
         >>> from dipy.reconst.gqi import GeneralizedQSamplingModel
-        >>> gq = GeneralizedQSamplingModel(gtab, method='gqi2', sampling_length=1.1)
+        >>> gq = GeneralizedQSamplingModel(gtab, method='standard', sampling_length=1.1)
         >>> voxel_signal = data[0, 0, 0]
         >>> odf = gq.fit(voxel_signal).odf(sphere)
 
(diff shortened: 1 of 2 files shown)


jhlegarreta (commented):
Thanks for doing this. Can tests be added, please?

dipy/reconst/gqi.py
```diff
@@ -299,3 +321,32 @@ def triple_odf_maxima(vertices, odf, width):
...
+
+    Notes
+    -----
+    The predicted signal is given by:
+
+    .. math::
+
+        S(\theta, b) = K_{ii}^{-1} \cdot ODF
+
+    where $K_{ii}^{-1}$, is the inverse of the gQI kernels for the direction(s) $ii$
```
arokem, line 346:
```suggestion
    where $K_{ii}^{-1}$, is the inverse of the GQI kernels for the direction(s) $ii$
```

arokem:
Overall, looks good! Does it fit the data well? For testing, maybe you could use `dipy.data.dsi_voxels`, which has a dsi acquisition and another non-Cartesian acquisition scheme with high b-values (the CENIR data set in DIPY or data from HBN should be good, as both have b=3,000 and relatively low SNR at that b-value).

oesteban:
> Does it fit the data well?

That's my next step. I wanted to ensure it looked reasonable overall, and it seems to do so. I'm not particularly worried about "fitting the data" because there's really no data fitting here: it's a matter of properly calculating and inverting the kernel. Other than that, it's only a matter of dot product with the ODF.



> For testing, maybe you could use `dipy.data.dsi_voxels`

👍 thanks!

dipy/reconst/gqi.py
```diff
@@ -72,19 +72,30 @@ def __init__(
...
-        gradsT = self.gtab.bvecs.T
-        b_vector = gradsT * tmp  # element-wise product
-        self.b_vector = b_vector.T
+        self.gtab = gtab
 
     @multi_voxel_fit
     def fit(self, data, **kwargs):
         return GeneralizedQSamplingFit(self, data)
 
+    @warning_for_keywords()
```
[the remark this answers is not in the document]
oesteban:
```suggestion
```

dipy/reconst/gqi.py
```diff
@@ -107,30 +118,41 @@ def __init__(self, model, data):
...
+                self.model.Lambda,
+                self.model.gtab,
+                method=self.model.method,
+            )
             self.model.cache_set("gqi_vector", sphere, self.gqi_vector)
 
         return np.dot(self.data, self.gqi_vector)
 
 
+@warning_for_keywords()
```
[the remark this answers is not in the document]
oesteban:
```suggestion
```

dipy/reconst/gqi.py
```diff
@@ -72,19 +72,30 @@ def __init__(
...
 
+    @warning_for_keywords()
+    def predict(self, odf, gtab, sphere):
+        """
+        Predict a signal for this TensorModel class instance given parameters.
+
+        Parameters
+        ----------
+        odf : ndarray
+            Map of ODFs
```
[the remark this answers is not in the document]
oesteban:
```suggestion
            Map of ODFs.
```

dipy/reconst/gqi.py
```diff
@@ -72,19 +74,29 @@ def __init__(
...
-        gradsT = self.gtab.bvecs.T
-        b_vector = gradsT * tmp  # element-wise product
-        self.b_vector = b_vector.T
+        self.gtab = gtab
 
     @multi_voxel_fit
     def fit(self, data, **kwargs):
         return GeneralizedQSamplingFit(self, data)
 
+    def predict(self, odf, gtab, sphere):
```
arokem: I think that for the model object, we want to provide parameters, but use the gtab that the model already has (see: https://github.com/dipy/dipy/blob/master/dipy/reconst/dti.py#L852) and then for the Fit object, we want the prediction API to be: 

```
def predict(self, gtab, *, sphere=None): 

```

(along the lines of https://github.com/dipy/dipy/blob/master/dipy/reconst/dti.py#L1194)

Where new measurement parameters are provided and the parameters are taken from the object itself (i.e., the odf used is the one that was calculated when the model was fit.) Does that make sense here as well?

oesteban: > the odf used is the one that was calculated when the model was fit.) Does that make sense here as well?

Absolutely. In fact, this is what I have been moving toward. The thing is that this model does not calculate ODFs until the method is explicitly called with a specific sphere.

To make it more generic, I think *the number of parameters* to encode the model fit should be defined at model instantiation time (meaning, the model should be created with a default sphere or take it as argument and keep that all along).

I'm having issues with the multi_voxel_fit for this too.

Would it make sense that models whose parameters are directly encoded with estimated ODFs should use those as the internal storage of the model fit? As it stands in master, this model does not actually fit any data if you call fit().

WDYT?

oesteban:
@arokem @jhlegarreta Okay, this is generating visually-acceptable images. How should I move next? I was thinking of adding an example to https://docs.dipy.org/stable/theory/gqi.html and then try to write some tests. DYT?

jhlegarreta:
> @arokem @jhlegarreta Okay, this is generating visually-acceptable images. 

:100:.

> How should I move next? I was thinking of adding an example to https://docs.dipy.org/stable/theory/gqi.html and then try to write some tests. DYT?

Examples go in here https://github.com/dipy/dipy/tree/master/doc/examples

Unless you believe the theory needs more content
https://github.com/dipy/dipy/tree/master/doc/theory

I'd leave it as is and add an example if you already have one.

Tests go into
https://github.com/dipy/dipy/blob/master/dipy/reconst/tests/test_gqi.py

dipy/reconst/gqi.py
```diff
@@ -145,14 +174,43 @@ def gqi_kernel(gtab, param_lambda, sphere, method="gqi2"):
...
+    r"""
+    Predict a signal given the ODF.
+
+    Parameters
+    ----------
+    odf : ndarray
+        ODF parameters.
+
+    gtab : GradientTable
+        The gradient table for this prediction
```
jhlegarreta, line 201: Re-work the parameter docstrings: some do not apply (`odf`) and some are missing.

jhlegarreta (commented):
Thanks for doing this.

dipy/reconst/gqi.py
```diff
@@ -68,25 +79,49 @@ def __init__(
...
+    def predict(self, odf, *, S0=None):
+        """
+        Predict a signal for this GeneralizedQSamplingModel instance given parameters.
 
-class GeneralizedQSamplingFit(OdfFit):
+        Parameters
+        ----------
+        odf : ndarray
+            Map of ODFs.
+        gtab : ndarray
```
jhlegarreta, line 116: `gtab` is not in the function signature and it is not used.

dipy/reconst/gqi.py
```diff
@@ -68,25 +79,49 @@ def __init__(
...
+            Orientations where signal will be simulated
+        sphere : :obj:`~dipy.core.sphere.Sphere`
```
jhlegarreta, line 118: `sphere` is not in the function signature and it is not used.

dipy/reconst/gqi.py
```diff
@@ -68,25 +79,49 @@ def __init__(
...
         return GeneralizedQSamplingFit(self, data)
 
+    def predict(self, odf, *, S0=None):
+        """
+        Predict a signal for this GeneralizedQSamplingModel instance given parameters.
 
-class GeneralizedQSamplingFit(OdfFit):
+        Parameters
+        ----------
+        odf : ndarray
```
jhlegarreta, line 114: `odf` is not used by the function. Should be removed from the function signature as well, much like `S0`.

dipy/reconst/gqi.py
```diff
@@ -98,37 +133,84 @@ def __init__(self, model, data):
...
+            @ self.model.kernel
+        )
+
+        return (K @ self.data.T).T
+
+
+def gqi_kernel(gtab, param_lambda, sphere, method="standard"):
+    # 0.01506 = 6*D where D is the free water diffusion coefficient
+    # l_values sqrt(6 D tau) D free water diffusion coefficient and
+    # tau included in the b-value
```
jhlegarreta, line 176:
```
    r"""Compute the GQI kernel.
     
    The scaling factor for the b-values is computed as $0.01506 = 6*D$ where
    $D$ is the free water diffusion coefficient, and $l_{values} = \sqrt(6 D_{\tau})$,
    $D$ being the free water diffusion coefficient and $\tau$ being included in the
    b-value.

    Parameters
    ----------
    gtab : GradientTable
        Gradient table.
    param_lambda : float
        Diffusion sampling length.
    sphere : :obj:`~dipy.core.sphere.Sphere`
        A sphere object.
    method : str
        Method to compute the kernel.

    Returns
    -------
    np.array
        GQI kernel.

    """"
```

dipy/reconst/gqi.py
```diff
@@ -98,37 +133,84 @@ def __init__(self, model, data):
...
+    scaling = np.sqrt(gtab.bvals * 0.01506)
```
jhlegarreta, line 177: 0.01506 should probably be computed as a function of D (which should be a parameter of the function) using the formula in the docstring. I have not found these specific values in the paper, though. This is related to the docstring, where it was not clear to me what the formula to compute 0.01506 was from the existing comment and the paper.

skoudoro:
Hi @oesteban,

do you have any update on this work?

PR #3557, opened 2025-06-05
DOC: fix streamline-tools tutorial by avoiding the use of identity affine

### LINKED ISSUE #3382: Check use of affine in `streamline_tools` example
## Description

The `streamline_tools` example reads some DWI data and discards its affine:
https://github.com/dipy/dipy/blob/master/doc/examples/streamline_tools.py#L49

The seeding mask is built with an identity matrix:
https://github.com/dipy/dipy/blob/master/doc/examples/streamline_tools.py#L82

This seems suspicious. See https://github.com/dipy/dipy/pull/3376/files#r1797507989.

The affine was returned but overwritten prior to PR #3376, so the underlying issue still applied.

## Way to reproduce

- [x] Code example: `streamline_tools.py` example.
- [x] Operating system and version: any
- [x] Python version: any
- [x] dipy version: `master` as of 2024-10-15 (commit 98675b7)
- [x] dependency version: any

skoudoro:
This PR fixes #3382

Not sure why the identity affine was used in this tutorial when it was not necessary. 

This update should make more sense for the user.

### DIFF
--- a/doc/examples/streamline_tools.py
+++ b/doc/examples/streamline_tools.py
@@ -26,7 +26,7 @@ from dipy.core.gradients import gradient_table
 from dipy.data import get_fnames
 from dipy.direction import peaks
 from dipy.io.gradients import read_bvals_bvecs
-from dipy.io.image import load_nifti, load_nifti_data, save_nifti
+from dipy.io.image import load_nifti, save_nifti
 from dipy.io.stateful_tractogram import Space, StatefulTractogram
 from dipy.io.streamline import save_trk
 from dipy.reconst import shm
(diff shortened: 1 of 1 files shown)


skoudoro:
> Please, change this text as well

Good catch, will do

> How do results change? 

Result do not change which make sense. Either you use the affine everywhere and save in the correct space or you do not use affine at all (via identity).  I think, people mix both, using affine at some place but not always everywhere. Especially in the visualization... if some actors have an affine and or other not, for sure the result will not be aligned.

skoudoro:
ok, update done in 24e7baead8d10d57b41969464c62d30d2407637b. Ready for review or to be merged

jhlegarreta (approved):

PR #3561, opened 2025-06-12
STYLE: Make affine variable naming consistent

jhlegarreta:
Make affine variable naming consistent in streamline tools example.


jhlegarreta:
No need to wait for CIs. Merging.

PR #3565, opened 2025-06-13
RF: use multi_voxel_fit for rumba

### LINKED ISSUE #3352: Rumba and multivoxel fit decorator
## Description

For Whatever reason, Rumba model does not `multivoxel fit` decorator. it has its own voxelwise function which is less efficient and have less capabilities. 

This needs to follow the standard and be updated

skoudoro:
This PR fixes #3352 by applying `multi_voxil_fit` decorator to the fitting function.

The goal is to match the current reconstruction model design. Also, it gives access to other optimization or tools like ray, dask, etc...


arokem:
This looks good to me. Relatedly: Do we have a mechanism in place to test all the models that are decorated with `multi_voxel_fit` across multiple different parallelization settings? Might be worth doing something like that?

skoudoro:
> Do we have a mechanism in place to test all the models that are decorated with multi_voxel_fit across multiple different parallelization settings?

No

> Might be worth doing something like that?

Absolutely, we need to write it somewhere to not forget

arokem:
Let's get this one merged and follow up with an issue. Is this ready as far as you are concerned? Looks like it to me.

skoudoro:
yes it is

PR #3582, opened 2025-06-27
NF: Add PR labeler workflow

jhlegarreta:
Add PR labeler workflow.

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,149 @@
+# Commit prefix regexes
+type:bugfix:
+  title: "^BF:.*"
+
+type:backward compatibility:
+  title: "^BW:.*"
+
+type:continuous integration:
+  title: "^CI:.*"
+
+type:documentation:
+  title: "^DOC:.*"
+
+type:maintenance:
+  title: "^MNT:.*"
+
+type:new feature:
+  title: "^NF:.*"
+
+type:optimization:
+  title: "^OPT:.*"
+
+type:refactoring:
+  title: "^RF:.*"
+
+type:style:
+  title: "^STYLE:.*"
+
+type:test:
+  title: "^TEST:.*"
+
+type:coverage:
+  title: "coverage"
+
+type:design:
+  title: "design"
+
+# Filename regexes
+area:align:
+  files:
+    - "dipy/align/*"
+
+area:benchmarking:
+  files:
+    - "benchmarks/*"
+
+area:boots:
+  files:
+    - "dipy/boots/*"
+
+area:core:
+  files:
+    - "dipy/core/*"
+
+area:data:
+  files:
+    - "dipy/data/*"
+
+area:denoise:
+  files:
+    - "dipy/denoise/*"
+
+area:direction:
+  files:
+    - "dipy/direction/*"
+
+area:documentation:
+  files:
+    - "doc/*"
+    - "README.rst"
+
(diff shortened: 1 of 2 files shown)


jhlegarreta:
@skoudoro A few comments:
- https://docs.dipy.org/stable/devel/commit_codes.html: `BK` is unclear to me; `PL` we do not use `pylint`.
- `type:*` and `area:*` are labels that have to be defined in GitHub. I have not done that yet until we agree what we want. There is some labels that are already in place; I'd stick to all lowercases, and adapt them to what is proposed in this PR or what we end up agreeing to (e.g. I'd change `github_actions` to `type:infrastructure`).

skoudoro:
I would remove `BK` and `PL`, we never use them. I propose to replace them by.
- `CI` -> everything related to github action, travis, etc
- `MTN` -> for maintenance tasks, sometime in a middle of `BF` and `RF` like release preparation

concerning labels, ok for lowercase for `type` and ok for adding `area` which would be diffusion area like `area:denoising` or `area:reconstruction`. Does it make sense?

jhlegarreta:
> I would remove BK and PL, we never use them. I propose to replace them by.

>    CI -> everything related to github action, travis, etc
>    MTN -> for maintenance tasks, sometime in a middle of BF and RF like release preparation

https://github.com/dipy/dipy/pull/3583 and made the corresponding changes here.

> which would be diffusion area like area:denoising or area:reconstruction. Does it make sense?

That is the aim.

jhlegarreta:
@skoudoro there are labels for "new feature" and for "enhancement" but there is not commit prefixes to distinguish these two.

skoudoro:
> there are labels for "new feature" and for "enhancement" but there is not commit prefixes to distinguish these

To me "New Feature" is clear : NF.  "Enhancement" is equivalent to RF -> refactoring.  we can remove enhancement and use refactoring

jhlegarreta:
> To me "New Feature" is clear : NF. "Enhancement" is equivalent to RF -> refactoring. we can remove enhancement and use refactoring

OK. Changing "enhancement" to "refactoring".

jhlegarreta:
This is ready then.

Related PR: https://github.com/dipy/dipy/pull/3584

skoudoro:
Question before merging: 

how does the labeller react if someone use lowercase `test` or `[TEST]` instead of `TEST`  ? 
if we rename the title afterwards, is it ok? will it update the labelling ?

jhlegarreta:
> how does the labeller react if someone use lowercase test or [TEST] instead of TEST ?

Not sure. In ITK we do not such a problem since we enforce programmatically commit subjects to have a given format :v:, so I have not looked into that. But it is a regex so as long as your regex includes both cases, it should be OK.

IMO, testing that or enforcing given commit subjects should be left for a separate PR, once we see that this is working.

> if we rename the title afterwards, is it ok? will it update the labelling ?

If you mean changing the regex, it will: the only thing this does is to look for the specified patterns in the commit and serve as the mapping for the GH action to associate the corresponding GH label, which needs to be defined. That is, the labels need to be created manually.

Not sure if that answers the question.

skoudoro:
> Not sure if that answers the question.

it does not, but I got the idea

> IMO, testing that or enforcing given commit subjects should be left for a separate PR, once we see that this is working.

ok, I am fine with that, Can you create this issue, I am going to merge this PR

PR #3583, opened 2025-06-27
DOC: Change unused commit prefixes to some other more useful ones

jhlegarreta:
Change unused commit prefixes to some other more useful ones.

Original proposal: https://github.com/dipy/dipy/pull/3582#issuecomment-3011959757


skoudoro (approved):
All good, thanks @jhlegarreta, merging

PR #3584, opened 2025-06-28
MNT: Add GitHub issue templates

jhlegarreta:
Add GitHub issue templates.

These templates were heavily inspired by the ones contributed to ITK in commit
https://github.com/InsightSoftwareConsortium/ITK/commit/72eae4bc21aae0ef82995ad2ff0313da4a4fac89.


jhlegarreta:
@skoudoro What is the template where you want to redirect to discussions? Or should we add a new one e.g. "help" that does this?

skoudoro:
This is cool but too complex, it will afraid most users. I like what `uv` does:

see here: https://github.com/astral-sh/uv/tree/main/.github/ISSUE_TEMPLATE

we can mix both knowing that Support & Help and question should be first in the list and redirect to discussion

jhlegarreta:
> This is cool but too complex, it will afraid most users.

Maybe. But the less time the user wants to invest in providing information, the more time the maintainers have to invest in getting that information, and the worse is towards the solution of the issue.

Note that the complexity or the very, very detailed and long explanations (long enough so that most users will not even bother reading them) are hidden in their multiple links, e.g. "Before posting in the issue tracker", or their nice "Reproducible examples" documentation. And we do not have that for now.

> I like what uv does:

Please, feel free to edit the templates and amend the commit.

jhlegarreta:
@skoudoro See if commit b990c01 together with https://github.com/dipy/dipy/discussions/3585 are good enough.

skoudoro (approved):
Thanks @jhlegarreta , merging

jhlegarreta:
Looks like there is some syntax issue as they are not being displayed.

PR #3589, opened 2025-07-02
MNT: Add commit message hook to ensure message adheres to standards

jhlegarreta:
Add commit message hook to ensure message adheres to standards:
- The commit message subject has some reasonable size to contain meaningful content.
- The commit message subject does not have leading or trailing spaces and it does not end with a period.
- The commit message has a maximum line length.
- The commit message subject has one of the pre-defined prefixes.

Add the corresponding command to the documentation showing how to install the `commit-msg` hook using `pre-commit`.

The hook script was borrowed with minor modifications from the equivalent one in ITK: https://github.com/InsightSoftwareConsortium/ITK

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -36,3 +36,8 @@ repos:
         entry: python tools/generate_requirements.py
         files: "pyproject.toml|requirements/.*\\.txt|tools/generate_requirements.py"
         additional_dependencies: ["dependency-groups"]
+      - id: dipy-commit-msg
+        name: dipy-commit-msg
+        entry: tools/hooks/commit-msg.py
+        language: python
+        stages: [commit-msg]
--- a/.github/workflows/check_format.yml
+++ b/.github/workflows/check_format.yml
@@ -30,3 +30,27 @@ jobs:
 
     - name: Install and run pre-commit hooks
       uses: pre-commit/action@v3.0.1
+
+  commit-messages:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    if: github.event_name == 'pull_request'
+
+    steps:
+    - name: Check out repository
+      uses: actions/checkout@v6.0.2
+      with:
+        fetch-depth: 0
+
+    - name: Set up Python 3.12
+      uses: actions/setup-python@v6
+      with:
+        python-version: '3.12'
+
+    - name: Check commit messages
+      run: |
+        for sha in $(git log origin/${{ github.base_ref }}..HEAD --format=%H); do
+          git log -1 --format=%B "$sha" > /tmp/commit_msg
+          echo "Checking: $(git log -1 --format=%s $sha)"
+          python tools/hooks/commit-msg.py /tmp/commit_msg || exit 1
+        done
--- a/doc/devel/coding_style_guideline.rst
+++ b/doc/devel/coding_style_guideline.rst
@@ -54,6 +54,7 @@ Setting up pre-commit
 2. Install the git hook scripts::
 
      pre-commit install
+     pre-commit install --hook-type commit-msg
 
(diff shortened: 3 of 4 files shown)


jhlegarreta:
Re: https://github.com/dipy/dipy/pull/3582#issuecomment-3020990254 here it is.

jhlegarreta:
If this needs changes, I'd be grateful that whoever has suggestions applied them as I've spent quite some time trying to figure out some things that did not seem to work at first, and cannot devote more time to it. My last tests seems to indicate that this is working, though.

jhlegarreta:
@skoudoro The hook is not being run by the pre-commit action:
https://github.com/dipy/dipy/actions/runs/16013005246/job/45174358635?pr=3589#step:4:110

So there is probably some edit to be made related to this
https://github.com/dipy/dipy/pull/3589/files#diff-af6ba5d06db44a158ab9fbf060eee8ff157ad48d02f2f643240da3f58257e5d9R44

Leaving this to you.

skoudoro:
Thank you for this @jhlegarreta.

I will look into it tomorrow or Monday next week.

jhlegarreta:
Pinging @skoudoro.

.pre-commit-config.yaml
```diff
@@ -16,3 +16,10 @@ repos:
       - id: codespell
         additional_dependencies:
           - tomli
+  - repo: local
+    hooks:
+    - id: dipy-commit-msg
+      name: 'dipy-commit-msg'
+      entry: 'tools/hooks/commit-msg.py'
+      language: python
+      stages: [commit-msg]
```
skoudoro: `[commit-msg]` not sure if this tell pre-commit to check only locally at the commit moment. At the CI's level, the commit is already done and push so it would make sense to me to not show up here. 

Soorrry, this decrease in priority but I will look at it at some point.

jhlegarreta: Thanks for the review @skoudoro. Makes sense. Will remove, force-push and see if the pre-commit job reports the status.

jhlegarreta: Now, it is being run,
https://github.com/dipy/dipy/actions/runs/16331696031/job/46135555136?pr=3589#step:4:119

But it does not get it right. I see it passing locally.

jhlegarreta: The issue may be that the commit message file does not actually exist in the CI runner.

jhlegarreta: And since removing `stages: [commit-msg]` the thing gets messed locally, so `stages: [commit-msg]` is maybe telling the entry point to take the commit message as an argument. Not sure.

jhlegarreta: Ran out of ideas:
https://stackoverflow.com/questions/79704142/make-pre-commit-work-to-check-the-commit-message-on-github

jhlegarreta: Re https://github.com/dipy/dipy/pull/3589#discussion_r2211862915 seems like the SO question was not very popular, but digging more into this I found this:
https://jorisroovers.com/gitlint/latest/

which is supported by `pre-commit`. It seems to support what we'd like here, e.g.
https://github.com/jorisroovers/gitlint/blob/4d9119760056492eabc201bfad5de2f9e660b85f/examples/gitlint#L26
and
https://github.com/jorisroovers/gitlint/blob/4d9119760056492eabc201bfad5de2f9e660b85f/examples/gitlint#L35C1-L35C20

skoudoro:
Seems to work. Before merging, can you try it @jhlegarreta ?

skoudoro (approved):
ok, I just tried and it fails as expected if wrong commit message on PRs.

Concerning dev environment, you have to explicitly install this pre-commit hook, otherwise, nothing happens. 

Thanks @jhlegarreta for initiating this! 

merging

PR #3593, opened 2025-07-04
STYLE: Adopt `pathlib` for path manipulation

jhlegarreta:
Adopt `pathlib` for path manipulation:
- Use `pathlib` for path manipulation in all internal operations, except workflows.
- All I/O functions except workflows now accept `Path` instances for paths besides `str` instances.
- Simplifies the code and eases the manipulation of paths.
- The local files returned by the fetcher are now `Path` instances.

The modules dealing with the `workflow` class I/O argument iteration, namely `workflows/io.py` and `workflows/multi_io.py` are not changed in this patch set. Thus, atlhough workflow tests have been transitioned to create `Path` instances for the I/O arguments, they are casted into `str` instances.

### DIFF
--- a/dipy/__init__.py
+++ b/dipy/__init__.py
@@ -39,7 +39,7 @@ from dipy.version import version as __version__
 
 
 def get_info():
-    from os.path import dirname
+    from pathlib import Path
     import sys
 
     import numpy
@@ -47,7 +47,7 @@ def get_info():
     import dipy
 
     return {
-        "pkg_path": dirname(__file__),
+        "pkg_path": Path(__file__).resolve().parent,
         "commit_hash": dipy.version.git_revision,
         "sys_version": sys.version,
         "sys_executable": sys.executable,
--- a/dipy/align/_public.py
+++ b/dipy/align/_public.py
@@ -9,6 +9,7 @@ streamlines.
 import collections.abc
 from functools import partial
 import numbers
+from pathlib import Path
 import re
 from warnings import warn
 
@@ -186,16 +187,16 @@ def register_dwi_to_template(
 
     Parameters
     ----------
-    dwi : 4D array, nifti image or str
+    dwi : 4D array, nifti image, str or Path
         Containing the DWI data, or full path to a nifti file with DWI.
-    gtab : GradientTable or sequence of strings
+    gtab : GradientTable or sequence of strings or Paths
         The gradients associated with the DWI data, or a sequence with
         (fbval, fbvec), full paths to bvals and bvecs files.
     dwi_affine : 4x4 array, optional
         An affine transformation associated with the DWI. Required if data
         is provided as an array. If provided together with nifti/path,
         will over-ride the affine that is in the nifti.
-    template : 3D array, nifti image or str
+    template : 3D array, nifti image, str or Path
         Containing the data for the template, or full path to a nifti file
         with the template data.
     template_affine : 4x4 array, optional
@@ -296,13 +297,13 @@ def read_mapping(disp, domain_img, codomain_img, *, prealign=None):
 
     Parameters
     ----------
-    disp : str or Nifti1Image
+    disp : str, Path or Nifti1Image
         A file of image containing the mapping displacement field in each voxel
         Shape (x, y, z, 3, 2)
 
-    domain_img : str or Nifti1Image
+    domain_img : str, Path or Nifti1Image
 
-    codomain_img : str or Nifti1Image
+    codomain_img : str, Path or Nifti1Image
 
     Returns
     -------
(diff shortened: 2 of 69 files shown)


jhlegarreta:
There maybe other minor helper functions that are now gone (i.e. changes the API): I've only recorded `split_name_with_gz`. We can probably ask Copilot to tell us which ones are gone.

jhlegarreta:
OK, this is working. Unfortunately, I had to give up on adding the `Path` capabilities to the workflows for now, as it does not seem so immediate. I know they will mostly take strings from the command line, but maybe there is scenarios where they will be called from other Python code. https://github.com/dipy/dipy/commit/e27ae6a3394d0652fb2b0d06d47e16e29fdbfed2 contains some relevant changes. Main points:
- Key files for this to work are `dipy/workflows/io.py` and `dipy/workflows/multi_io.py`.
- The globbing over string vs `Path` instances needs some work:
https://github.com/dipy/dipy/commit/bfde4b42ce15c4ea62a9ef76758489e05af446c0#diff-061244c2af96e3b6c590236903412f79e87ecf21d1737dc5620143da1f8782d1L255-L258
https://github.com/dipy/dipy/commit/bfde4b42ce15c4ea62a9ef76758489e05af446c0#diff-061244c2af96e3b6c590236903412f79e87ecf21d1737dc5620143da1f8782d1L310
- This block
https://github.com/dipy/dipy/blob/master/dipy/workflows/tests/test_iap.py#L217-L227
poses problems. If `Path` instances are kept as they are, `argparse` starts failing for the calls `args = parser.get_flow_args()` in tests in that file. That was the decisive reason to give up on this now, as I do not have time to investigate. But I am sure that it can be done as `argparse` is happy if one gives arguments that are of type `Path`.
- Had to convert manually all input paths to strings in tests, as I had prepared them to use only `Path` instances, e.g.
https://github.com/dipy/dipy/pull/3593/files#diff-c5275d38822ecb0b718339d9fae7812ff074cc44c7c2aa548af85a098d2645aeR42
https://github.com/dipy/dipy/pull/3593/files#diff-c5275d38822ecb0b718339d9fae7812ff074cc44c7c2aa548af85a098d2645aeR99
https://github.com/dipy/dipy/pull/3593/files#diff-2caf09b292dfb9e041d4e31371727fa9c26a2e205f8c1212e8a133e817c83523R134

etc. Those will need to be reverted once `workflows/multi_io.py` adn `workflows.io.py` are working.

- `fury` does not support `Path` instances, so all calls to `fury` methods had to use strings.
- There are some `NiBabel` functions that do not support `Path` instances for now. So same as above.

Will open an issue when this gets merged to have a go as time permits.

jhlegarreta:
CI failure is unrelated to patch set:
```
reconst/tests/test_sfm.py::test_sfm_stick
  /Users/runner/work/dipy/dipy/venv/lib/python3.12/site-packages/dipy/reconst/sfm.py:652: RuntimeWarning: invalid value encountered in dot
    return np.dot(
```

https://github.com/dipy/dipy/actions/runs/16101655522/job/45431900183?pr=3593#step:12:3161

skoudoro:
Thank you @jhlegarreta, I have hard time to see the benefit of this, However, I completely fine with this change. 

I will look into it after we merge logging PR. More conflict will appear, So I prefer to fix these conflicts all in once

jhlegarreta:
>  have hard time to see the benefit of this, 

The benefit is that you accept both `str` and `Path`, the latter being a class that is adapted for path manipulation. For example, you can easily get the dirnames, file basenames, rootnames, extensions, etc. without having to deal with splitting the parts manually, figuring out whether it has multiple extensions, etc. You can call any numpy or matplotlib function with `Path` and they will work without issues, even some `NiBabel` functions, so time to evolve.

Edit: When we get workflows to work, we will be able to remove all casts in their tests, and they will look a lot cleaner.

skoudoro:
As expected, more conflicts after merging logging PR

jhlegarreta:
> As expected, more conflicts after merging logging PR

Resolved.

jhlegarreta:
RE https://github.com/dipy/dipy/pull/3599#issuecomment-3050360037: no need to rebase as there are no conflicts. This is ready to be merged.

skoudoro:
ok, all good, merging. Thanks @jhlegarreta.

PS1: this will break many PR and need rebase. We will have to be careful during reviews.
PS2: Can you update https://github.com/dipy/dipy/blob/master/doc/api_changes.rst  on a new PR please @jhlegarreta . Precise that `split_name_with_gz` has been removed without deprecation cycle. Thanks

jhlegarreta:
> PS1: this will break many PR and need rebase. We will have to be careful during reviews.

Yes, but changes are very clear and localized to path manipulation, so people should not get into trouble. And these changes will make maintaining things easier.

> PS2: Can you update https://github.com/dipy/dipy/blob/master/doc/api_changes.rst on a new PR please @jhlegarreta . Precise that split_name_with_gz has been removed without deprecation cycle. Thanks

https://github.com/dipy/dipy/pull/3601, but we can bring it back if you want, even if DIPY itself will no longer use it. Anyways, it was an `io.utils` function, so not a core function.

PR #3595, opened 2025-07-06
DOC: Remove default arguments from docstrings

jhlegarreta:
Remove default arguments from docstrings: their values are clearly visible in the function signatures and reduces maintenance burden. Also, the modified file is not part of the API.

Take advantage of the commit to fix a parameter name.


skoudoro (approved):
All good, merging, thanks @jhlegarreta

PR #3598, opened 2025-07-07
BF: Miscellaneous fixes to surfaces

jhlegarreta:
- RF: Remove unreachable code in tractogram saving function
- BF: Do not re-write the destination file extension with the references'
- DOC: Make the docstring parameters match the names in the signature


skoudoro (approved):
Those fixes make sense and are straight forward. Merging

Thanks @jhlegarreta

PR #3429, opened 2024-12-21
StatefulSurface - Class to handle surfaces

frheault:
**Summary**
This PR introduces functionality inspired by the `StatefulTractogram`, serving as its first iteration. Like the `StatefulTractogram`, this implementation requires surfaces to be loaded alongside a matching NIfTI file to allow seamless switching between coordinate spaces (`LPSMM, RASMM, VOXMM, VOX`).

**Surface Loading:**
- The existing `load_pial()`, `load_polydata()`, and `load_gifti()` functions remain available.
- A new `load_surface()` function wraps these loading methods, ensuring adherence to conventions:
  - GIFTI files are treated as being in `RASMM`.
  - PIAL files are treated as being in `RASMM`, but may be offset by half the reference volume.
  - VTK/PLY files can be saved in various formats, typically `RASMM` or `LPSMM`.
  - Since VTK/PLY files can be saved in different formats, they must be loaded with knowledge of the saving format. The `load_surface() `function simplifies exploration when this information is uncertain.

**Surface Saving:**

- New `save_pial()` and `save_gifti()` functions provide basic saving functionality.
- A `save_surface()` function wraps these methods, streamlining the saving process and ensuring convention compliance for formats such as GIFTI and PIAL.

**Testing and Future Work**
The Sherbrooke team plans to submit a PR to Scilpy in January, using this branch in an external library to validate the implementation and identify potential issues.
We need to expand the `scil_surface_convert.py` and `scil_surface_flip.py` scripts, ensuring compatibility with tools like:
- 3D Slicer
- DSI Studio
- SurfIce
- ITK-Snap
- Compatibility with MI-Brain (in LPSMM) has already been confirmed.
- Compatibility with output of Freesurfer and GIFTI datasets

**Related PR**
This PR complements the following (enabling simpler cross-software testing, as many tools still lack support for the new VTK format): [https://github.com/fury-gl/fury/pull/957 ](https://github.com/fury-gl/fury/pull/957 )

PR #3599, opened 2025-07-08
TEST: Restore removed test case in io workflows

jhlegarreta:
Restore removed test case in io workflows. The test case was added in commit 5478aad and removed without any proper reason when PR https://github.com/dipy/dipy/pull/3429 was merged.

That PR started from a `master` branch that was more than 1400 commits behind (commit 75b490b from Dec 29 2023) the current DIPY `master` branch, and the commit where that test case was added was dated Jun 3, 2025, and thus way more recent. The line was removed in commit 9bcd2b3, across the many conflict fix and `master` merge commits that were done in https://github.com/dipy/dipy/pull/3429.

### DIFF
--- a/dipy/workflows/tests/test_io.py
+++ b/dipy/workflows/tests/test_io.py
@@ -79,6 +79,9 @@ def test_io_info():
     npt.assert_raises(SystemExit, io_info_flow.run, filepath_dix["gs_streamlines.tck"])
 
     io_info_flow = IoInfoFlow()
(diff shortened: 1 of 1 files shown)


jhlegarreta:
I think it is safe to merge this. I will then rebase PR #3593 for the last time so that we have it merged soon, before more rebases, waiting times and retriggers of the test CI, etc. are needed.

PR #3603, opened 2025-07-09
DOC: Fix workflows test utils parameter name in docstring

jhlegarreta:
Fix workflows test utils parameter name in docstring.

Fixes:
```
Missing parameter positional_int in docstring
```

and
```
Function 'run' does not have a parameter 'positional_variable_int'
```

raised by the IDE.

### DIFF
--- a/dipy/workflows/tests/workflow_tests_utils.py
+++ b/dipy/workflows/tests/workflow_tests_utils.py
@@ -166,7 +166,7 @@ class DummyVariableTypeWorkflow(Workflow):
         ----------
         positional_variable_str : variable string
             fake input string param
-        positional_variable_int : int
+        positional_int : int
             fake positional param
         out_dir : string
             fake output directory


jhlegarreta:
@skoudoro The labeler is not applying the labels here, despite finding the right labels:
```
2025/07/10 00:02:22 Final set of labels: `["area:workflows" "type:documentation"]`
```

https://github.com/dipy/dipy/actions/runs/16182745511/job/45682725854?pr=3603#step:3:771

and despite having write access to the PR. You may want to look into it as time permits.

skoudoro:
it might be due to `type:documentation` since this label do not exist, the existing one is with emoticon. 

Maybe we should remove all emoticon in the labels

skoudoro (approved):
All good, merging

jhlegarreta:
> Maybe we should remove all emoticon in the labels

Removed them from `type:documentation`.

PR #3605, opened 2025-07-15
RF: Parallel `quantize_evecs`

skoudoro:
This PR supersedes #1164 and #1097

- Added a multiprocessing option by using `paramap`. Parallel engine can be choosen between Ray, Dask, Joblib, etc..
- Added the ability to choose which eigenvector to use as primary vector.
- This function was not tested so many tests added. Actually, this function had a bug which has been fixed.
- Some variable renaming to clarify the function

### DIFF
--- a/dipy/reconst/dti.py
+++ b/dipy/reconst/dti.py
@@ -20,6 +20,7 @@ from dipy.reconst.weights_method import (
     weights_method_wls_m_est,
 )
 from dipy.testing.decorators import warning_for_keywords
+from dipy.utils.parallel import paramap
 
 MIN_POSITIVE_SIGNAL = 0.0001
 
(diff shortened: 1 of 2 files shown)


dipy/reconst/dti.py
```diff
@@ -2484,30 +2485,61 @@ def design_matrix(gtab, *, dtype=None):
...
     ----------
     evecs : ndarray
-        Eigenvectors.
+        Eigenvectors of shape (..., 3, n_evecs)
     odf_vertices : ndarray, optional
         If None, then set vertices from symmetric362 sphere.  Otherwise use
         passed ndarray as vertices
+    idx : int, optional
+        Use idx-th eigenvector as the primary eigenvector for fiber tracking.
+        If 0, uses the largest eigenvalue.
```
jhlegarreta, line 2500: Using the 0 index for that seems counter-intuitive, e.g.
https://github.com/dipy/dipy/pull/3605/files#diff-8475f43b33a3b4095f0eb13e515f92b79430c558a4ac22b42cc9875e1d40276bR1184

Why not -1 for the largest eigenvalue?

arokem (approved):
I think this looks fine. Not sure that I understand the use-case for `idx`, though.

dipy/reconst/dti.py
skoudoro: > Why not -1 for the largest eigenvalue?

This is a convention in the current codebase. Changing it means update all tractography algorithm and  peaks related function like `peak_from_model`

skoudoro:
> I think this looks fine. Not sure that I understand the use-case for idx, though.

@tangwei asked me this once. She needed to track in a very small region using the second peak to find a very specific bundle in the litterature. 
Also, since we want to generalize tracking to other field / body part, it might be useful. I do not have a specific use-case @arokem

PR #3606, opened 2025-07-16
RF: Address cvxpy 1.7.0 warnings.

skoudoro:
Currently, our CI's are failing due to the new `cvxpy`release.  Here is a quick fix, but I am not the problem is in our level, we shoudl check with `cvxpy` and introspect deeper.

```
reconst/tests/test_mapmri.py::test_plus_constraint
reconst/tests/test_qti.py::test_ls_sdp_fits
reconst/tests/test_qti.py::test_qti_fit
  /Users/runner/work/dipy/dipy/venv/lib/python3.10/site-packages/scs/__init__.py:83: UserWarning: Converting A to a CSC (compressed sparse column) matrix; may take a while.
    warn(
```

### DIFF
--- a/dipy/core/optimize.py
+++ b/dipy/core/optimize.py
@@ -549,7 +549,10 @@ class PositiveDefiniteLeastSquares:
                     return np.asarray(self._h.value).squeeze()
 
             # Solve constrained problem
-            self.problem.solve(**kwargs)
+            with warnings.catch_warnings():
+                if self._y.value.shape[0] < 1000:
+                    warnings.filterwarnings("ignore", message="Converting A to a CSC")
+                self.problem.solve(**kwargs)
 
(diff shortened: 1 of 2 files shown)


skoudoro:
> Looks like this could be a pretty serious performance regression...

I agree

> Should we post an issue on cvxpy's GitHub?

Please, go ahead ! I did not found any particular issue on that point. I was trying to find an alternative.

arokem:
OK! Issue raised here: https://github.com/cvxpy/cvxpy/issues/2875

arokem:
Looks like this will eventually be fixed upstream when a new release of scs comes along. IIUC, it's also not an indication of a regression. I think we can merge this for now but I'll also post an issue to come back and revert these changes when there's a new release of scs

Issue #3608, opened 2025-07-18
Predicting from a fitted DTI model is broken?


samcoveney:
### Summary

I am encountering a weird issue where setting up a DTI model,

e.g.

```
tenmodel = Model(gtab, fit_method=method, return_S0_hat=True)
```
and then fitting it, and trying to subsequently predict:

```
data_f = tenfit.predict(gtab_other, S0=tenfit.model_S0)
```

is giving my infinities and other nonsense. Absolutely nothing wrong with the fitted parameters though.

This is bizarre, as I never had a problem with my workflow before, so I think something has changed in the prediction function of tensor models, but I can't figure this out. Has my computer got a ghost in it, or has something changed in DiPy that is causing this issue?

### Platform

linux

### DIPY version

1.12.0.dev0

### Python version

3.12.8

arokem:
Can you provide a full reprex? I.e., code that includes the gradient tables and data from one single voxel where you are seeing this?

samcoveney:
https://github.com/dipy/dipy/blob/124310c90c32ed6bbe463c5146de3fce57125b5e/dipy/reconst/dti.py#L2380C1-L2381C23

What you can have is this. Someone has changed the lower_triangular code to average the provided (fitted!) S0 values for some unknown reason

samcoveney:
The parameters are per voxel. If one wished to make a prediction, we need to use the fitted S0 (NOT a dummy variable) per voxel. How can we check prediction and data unless scaling properly...

Beer time

arokem:
Oh dear. Good catch! 🍻

samcoveney:
There are two beers in that emoji, so you have to also have one.

Alright, we'll sort this somehow, I don't have the chance right now. It would be good to know when this got in here.

arokem:
I am having a coffee for now, but will proceed to beer in a short while... 

It hasn't been there very long: https://github.com/dipy/dipy/pull/3563

samcoveney:
It was a very interesting solution to solve some negative b0 values in an array by averaging the array.

skoudoro:
My bad, Not sure and do not remember what I had in mind when I did this. Maybe many beer before 😅. PR to fix it coming...

thank you for catching it

jhlegarreta:
Taking part of this blame as well. The tests should capture the faulty behavior.

samcoveney:
There is a general problem with many tests only focussing on a single voxel.

jhlegarreta:
[...]

PR #3563, opened 2025-06-12
RF: Address some Zero division warnings

skoudoro:
this PR fixes #3418.

it addresses some zero division warnings and update some tests.

PR #3612, opened 2025-07-24
BF: fix lower triangular bug introduced in #3563

skoudoro:
All the title. 

Bug that I introduced in the `lower_triangular` function in #3563.

This is a quick fix but we have to do a follow-up PR to improve our DTI testing by using multiple voxel instead of a single one.

See #3608 for the discussion.

Can you give a quick look @samcoveney or @arokem?


samcoveney:
Just tested it, seems to fix the issue in #3608 

I will say, that maybe it is a bit odd that if b is less than 0, that it is set to 1? Would it be better to set it to ```MIN_POSITIVE_SIGNAL``` instead?

i.e. in dti.py we have
```
MIN_POSITIVE_SIGNAL = 0.0001
```

skoudoro:
> Would it be better to set it to MIN_POSITIVE_SIGNAL instead?

"Better" I do not know, but it will clearly make more sense to me. I just updated it in [aec352b](https://github.com/dipy/dipy/pull/3612/commits/aec352bc0b43e6c7d42a7c6a89a6059bbd01dc9a)

Thanks

samcoveney:
Okay that's good. I suppose if we were really fussy, we might check that the signal is above MIN_POSITIVE_SIGNAL instead of above zero... But probably neither here nor there

samcoveney:
How has the test passed on line 625 of test_dti.py ? It looks like the last entry of D ought to be -log(MIN_POSITIVE_SIGNAL) and yet it's tested against an entry of "9", how are the tests passing here?

```- log(0.0001) = 9.21``` (definitely not 9)

**Comment by skoudoro** [comment]
> How has the test passed on line 625 of test_dti.py ?

This is due to the tensor type in the test, see below

image

So the int truncate the float which give 9 (``np.in64(9.21) = 9``) so the test make sense since we output the same dtype as the input.

Maybe this information can be added in the docstring

**Comment by samcoveney** [comment]
Ah okay thanks for that. All seems good then. Docstring or comment, sure, but as long as it's behaving correctly.

**Comment by samcoveney** [comment]
any movement on this one?

**Comment by skoudoro** [comment]
@arokem ? @jhlegarreta ? @Garyfallidis ? Any additional comments ? this is ready to go

**Review by jhlegarreta — APPROVED** [review]
Looks good. The only thing I was thinking was whether this could be refactored into a function so that this can be reused if necessary or tested better.

**Comment by skoudoro** [comment]
> I was thinking was whether this could be refactored into a function

Can you tell me more what do you have in mind?  Also, since we have a critical bug in master, I believe this should be merged and then we can do a follow-up PR for refactoring.

**Comment by jhlegarreta** [comment]
> Can you tell me more what do you have in mind? 

So the statement
```
np.where(b0 > MIN_POSITIVE_SIGNAL, b0, MIN_POSITIVE_SIGNAL)
```

seems like it could be applied to all those places where you will be willing to avoid negative values in the `b0` signal. I do not know if there are other places, but IMO the way the signal is clipped should be made apparent in a function. Feel free to merge as is and follow with a separate PR (that is why I approved the PR :+1:).

PR #3613, opened 2025-07-24
BF: Allow `length` to work with float16 streamlines

skoudoro:
This PR is a follow up of https://github.com/dipy/dipy/discussions/3611.

Saving streamlines as `np.float16` can be useful for saving disk space.

Unfortunately, it seems, `length` was not handling that correctly. This PR fix this issue.

In general, this is something to track. Maybe other cython function do not handle correctly this datatype.

### DIFF
--- a/dipy/tracking/streamlinespeed.pyx
+++ b/dipy/tracking/streamlinespeed.pyx
@@ -98,16 +98,23 @@ def length(streamlines):
             return 0.0
 
         arclengths = np.zeros(len(streamlines), dtype=np.float64)
-
-        if as_native_array(streamlines._data).dtype == np.float32:
+        streamlines_data = as_native_array(streamlines._data)
+        if streamlines_data.dtype == np.float32:
             c_arclengths_from_arraysequence[float2d](
-                                    as_native_array(streamlines._data),
+                                    streamlines_data,
                                     streamlines._offsets.astype(np.intp),
                                     streamlines._lengths.astype(np.intp),
                                     arclengths)
-        else:
+        elif streamlines_data.dtype == np.float64:
             c_arclengths_from_arraysequence[double2d](
-                                      as_native_array(streamlines._data),
+                                      streamlines_data,
+                                      streamlines._offsets.astype(np.intp),
+                                      streamlines._lengths.astype(np.intp),
+                                      arclengths)
+        else:
+            streamlines_data = streamlines_data.astype(np.float64)
+            c_arclengths_from_arraysequence[double2d](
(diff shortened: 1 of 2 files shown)


dipy/tracking/tests/test_streamline.py
```diff
@@ -161,6 +161,8 @@
 )
 
 streamline_64bit = streamline.astype(np.float64)
+streamline_16bit = streamline.astype(np.float16)
+streamline_16bit = streamline.astype(np.float16)
```
jhlegarreta:
```suggestion
```

jhlegarreta (commented):
A duplicate ?

dipy/tracking/tests/test_streamline.py
skoudoro: Thanks!

jhlegarreta: The duplicate is back.

skoudoro: sorry for that, during my rebase, I forgot to git pull before....

dipy/tracking/streamlinespeed.pyx
```diff
@@ -98,16 +98,23 @@ def length(streamlines):
...
                                     streamlines._lengths.astype(np.intp),
                                     arclengths)
+        elif streamlines_data.dtype == np.float64:
+            c_arclengths_from_arraysequence[double2d](
+                                      streamlines_data,
+                                      streamlines._offsets.astype(np.intp),
+                                      streamlines._lengths.astype(np.intp),
+                                      arclengths)
         else:
+            streamlines_data = streamlines_data.astype(np.float64)
```
jhlegarreta, line 115: So for `np.float16` you effectively convert to `np.float64` I guess because there is no other equivalent type in Cython? Then where is the inverse conversion made when writing? i.e. referring to this comment in the commit message:
```
Saving streamlines as np.float16 can be useful for saving disk space.
```

skoudoro: you are right, I should revert back to the original dtype. Let me look at it.

skoudoro: Finally, I just looked at the code and this cast in not important since we return `arclength` and not `streamlines_data`. So, the original streamlines are not modified. 

We can keep it as it is

skoudoro: thank you for the feedback

jhlegarreta (approved):

PR #3618, opened 2025-08-06
BF: manage empty file in dipy_slr cli to avoid enigmatic crash

skoudoro:
Small PR to improve the error message when using ``dipy_slr`` CLI.

The error message was quite enigmatic and it was easy to waste time to understand when it should be obvious.

Associated tests has been added


dipy/workflows/tests/test_align.py
```diff
@@ -62,6 +62,30 @@ def test_slr_flow():
 
         npt.assert_equal(Path(out_path).is_file(), True)
 
+        sft = sft.from_sft(np.array([]), sft)
+        # sft.streamlines._data =
+        # sft2 = StatefulTractogram([], 'same', Space.RASMM)
```
jhlegarreta: The above two lines should be removed.

dipy/workflows/tests/test_align.py
```diff
@@ -62,6 +62,30 @@ def test_slr_flow():
...
+            bbox_valid_check=False,
+        )
+        npt.assert_raises(
+            SystemExit,
+            slr_flow.run,
+            data_path,
+            empty_path,
+            out_dir=out_dir,
+            bbox_valid_check=False,
+        )
```
jhlegarreta: I'd suggest making the tests check exactly the exception message.

skoudoro: We test `sys.exit(1)` here.  So, not sure how to check that. 
before the `sys.exit`, we have `logger.error`  so, I will look how to make it more beautiful

skoudoro: fixed in 939fe704a46077a349801193f904a2f54bf2a73f

dipy/workflows/tests/test_align.py
skoudoro: fixed in 939fe704a46077a349801193f904a2f54bf2a73f

skoudoro:
Thank you for your comments @jhlegarreta. 

everything has been fixed in 939fe704a46077a349801193f904a2f54bf2a73f

jhlegarreta (approved):

jhlegarreta:
Looks like the failing build is related to the patch set:
https://github.com/dipy/dipy/actions/runs/16833031622/job/47687618592?pr=3618#step:12:3419
```
assert f"Static file {empty_path} is empty" in caplog.text
```

skoudoro:
interesting.... why this one and not the others CI's..... I will look into it, something seems to interfere

dipy/workflows/align.py
```diff
@@ -231,6 +232,13 @@ def run(
                 moving_file, "same", bbox_valid_check=bbox_valid_check
             )
 
+            if not len(static_obj.streamlines):
+                logger.warning(f"Static file {static_file} is empty")
```
maharshi-gor: this logger should be an error. As the warning should be used when you can continue after showing the warning, or a case where we are deviating from the ideal parameters but it can work in the process.

Here, we are stopping the execution on very next-line. I suppose this needs to be an error. To communicate better to the audience.

maharshi-gor:
Hi @skoudoro one of the CIs seem to be error on the slr flow itself. Can you check if this related? Otherwise I can merge this PR.

skoudoro:
yes, I will look into it after nlmeans PR

skoudoro:
the log is still not catch in the CI coverage. Not sure why, still investigating...

jhlegarreta:
Thanks for all this effort, Serge.

maharshi-gor:
LGTM!

Issue #1259, opened 2017-06-16
nlmeans does not respect boundaries


samuelstjean:
this one is pretty important is deserves another note instead of the fixing pr in case I forgot.
Turns out the cython part disables wrapping around with negative indexes, but the code is using it and the test, well, does not test the correct array. I am wondering if it might be reading garbage sometimes, but for now not sure the test does as intended.

Right now (copied from https://github.com/nipy/dipy/blob/master/dipy/denoise/tests/test_non_local_means.py#L45) it does 

~~~python
import numpy as np
from numpy.testing import (run_module_suite,
                           assert_,
                           assert_equal,
                           assert_array_almost_equal,
                           assert_raises)
from dipy.denoise.non_local_means import non_local_means


def test_nlmeans_boundary():
    # nlmeans preserves boundaries
    S0 = 100 + np.zeros((20, 20, 20))
    noise = 2 * np.random.standard_normal((20, 20, 20))
    S0 += noise
    S0[:10, :10, :10] = 300 + noise[:10, :10, :10]
    S0n = non_local_means(S0, sigma=np.std(noise), rician=False)
    assert_(S0[9, 9, 9] > 290)
    assert_(S0[10, 10, 10] < 110)
~~~

But notice how it checks the original array instead of doing
~~~python
def test_nlmeans_boundary():
    # nlmeans preserves boundaries
    S0 = 100 + np.zeros((20, 20, 20))
    noise = 2 * np.random.standard_normal((20, 20, 20))
    S0 += noise
    S0[:10, :10, :10] = 300 + noise[:10, :10, :10]
    S0n = non_local_means(S0, sigma=np.std(noise), rician=False)
    assert_(S0n[9, 9, 9] > 290)
    assert_(S0n[10, 10, 10] < 110)
~~~

And that does not work as of now.

Issue #1224, opened 2017-04-16
Memory issue


RafaelNH:
Just to report a memory issue that raises when trying to denoise the CENIR multi-shell data using the nlmeans function. It  will be important to address this issue since it happens when trying to run the DKI's example of usage 'reconst_dki.py'  (I am using an 8GB RAM machine).

**Code:**

fetch_cenir_multib(with_raw=False)
bvals = [200, 400, 1000, 2000]
img, gtab = read_cenir_multib(bvals)
data = img.get_data()
affine = img.affine

axial_slice = 40
sigma = estimate_sigma(data, N=4)
mask_roi = np.zeros(data.shape[:-1], dtype=bool)
mask_roi[:, :, axial_slice] = mask[:, :, axial_slice]

den = nlmeans(data, sigma=sigma, mask=mask_roi)
den = den[:, :, axial_slice, :]
den = nlmeans(data, sigma=sigma, mask=mask_roi)

**Error**

  File "<ipython-input-14-70957b8625e9>", line 8, in <module>
    den = nlmeans(data, sigma=sigma, mask=mask_roi)

  File "C:\Users\Rafael\dipy\dipy\denoise\nlmeans.py", line 67, in nlmeans
    sigma = np.ones(arr.shape, dtype=np.float64) * sigm

MemoryErro

arokem:
Maybe use `dipy.denoise.non_local_means.non_local_means` instead? I suspect it might be less memory-demanding.

PR #3623, opened 2025-08-15
RF: refactoring of ``nlmeans`` to allow the selection our classic version and the blockwise version

### LINKED ISSUE #1178: Recheck indexing with nlmeans as some of it is indexing wrong positions

skoudoro:
This PR fixes #1178, #1259, #1224, #2950, #1131

This PR refactors the `nlmeans` module and allow the use of our both version. Currently, the blockwise version was not used. Instead of keeping dead code, this small correction has been done to allow a faster version for users.

Here are the list of the differents task done in this PR

1. Merged both Implementations into one 1 function to keep backward compatibility. 
2. Added a variable named `method` which let you select between both methods.
3. Update of all tests
4. Deletion of some files in favor of the unification into one unique module (`nlmeans.py` and `denspeed.pyx`)
5. fixed indexing issues and memory issue, and making sure that parallelization is working.

So we have now 3 files instead of 6.

  - nlmeans.py - Unified interface with method selection
  - denspeed.pyx - Contains both classic and blockwise implementations
  - test_nlmeans.py - Consolidated tests for both methods

After some tests, the "blockwise" version seems to be much faster and provide similar results.
```python
# test on stanford T1
Sigma value ([5320.9854]), Sigma estimation time: 0.009128332138061523
Time taken for nlmeans 0.3900949954986572
Time taken for non_local_means 0.057128190994262695
```

image

### DIFF
--- a/dipy/denoise/nlmeans.py
+++ b/dipy/denoise/nlmeans.py
@@ -103,6 +103,10 @@ def nlmeans(
     >>> noise_levels = np.linspace(50, 100, 30)  # Varying noise
     >>> denoised_dwi = nlmeans(dwi_data, sigma=noise_levels)
     """
+    method = method.lower()
+    if method not in ["classic", "blockwise"]:
+        raise ValueError(f"Unknown method '{method}'. Use 'classic' or 'blockwise'.")
+
     if block_radius is None:
         if method == "classic":
             block_radius = 5
@@ -119,10 +123,8 @@ def nlmeans(
         if not np.issubdtype(sigma.dtype, np.number):
             raise ValueError("sigma should be an array of floats", sigma)
 
-        # Validation depends on both array dimensions and method
         if arr.ndim == 3:
             if method == "classic":
-                # Classic method for 3D: sigma should be 3D array or scalar
                 if sigma.ndim not in [0, 3] and not (
                     sigma.ndim == 3 and sigma.shape == arr.shape
                 ):
@@ -131,7 +133,6 @@ def nlmeans(
                         f" 3D array, got shape {sigma.shape}"
                     )
             elif method == "blockwise":
-                # Blockwise method: more flexible, accepts scalar, 1D, or 3D
                 if sigma.ndim > 3:
                     raise ValueError(
                         "For blockwise method, sigma should be at most 3D, got "
@@ -162,11 +163,9 @@ def nlmeans(
 
     if arr.ndim == 3:
         if method == "classic":
-            # Classic method needs 3D sigma array
             if not isinstance(sigma, np.ndarray):
                 sigma_3d = np.full(arr.shape, sigma, dtype="f8")
             else:
-                # Ensure sigma is 3D and same shape as arr
                 if sigma.shape != arr.shape:
                     if sigma.ndim == 0 or (sigma.ndim == 1 and len(sigma) == 1):
                         sigma_3d = np.full(arr.shape, float(sigma), dtype="f8")
@@ -178,7 +177,6 @@ def nlmeans(
                 else:
                     sigma_3d = np.ascontiguousarray(sigma, dtype="f8")
         else:
-            # For blockwise method, sigma can be scalar or array
             sigma_3d = sigma
 
         result = nlmeans_3d(
@@ -196,14 +194,12 @@ def nlmeans(
     elif arr.ndim == 4:
         denoised_arr = np.zeros_like(arr)
         for i in range(arr.shape[-1]):
-            # For classic method, we need to convert scalar sigma to 3D array
             if method == "classic":
                 if isinstance(sigma, np.ndarray):
                     sigma_vol = np.full(arr[..., i].shape, sigma[i], dtype="f8")
                 else:
                     sigma_vol = np.full(arr[..., i].shape, sigma, dtype="f8")
             else:
-                # For blockwise method, use scalar sigma
                 if isinstance(sigma, np.ndarray):
                     sigma_vol = sigma[i]
                 else:
--- a/dipy/workflows/denoise.py
+++ b/dipy/workflows/denoise.py
@@ -1,3 +1,4 @@
+from collections.abc import Sequence
 import shutil
 
 import numpy as np
@@ -129,10 +130,12 @@ class NLMeansFlow(Workflow):
     def run(
         self,
         input_files,
-        sigma=0,
+        sigma=(0,),
         patch_radius=1,
         block_radius=5,
         rician=True,
+        num_threads=None,
+        method="blockwise",
         out_dir="",
         out_denoised="dwi_nlmeans.nii.gz",
     ):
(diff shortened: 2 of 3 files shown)


skoudoro:
Yes, there, I might have not use the best parameters since we still see some structure, it was just to give an idea during this PR.

dipy/denoise/nlmeans.py
```diff
@@ -18,83 +13,219 @@ def nlmeans(
...
+        Number of OpenMP threads to use for parallel processing. If None,
+        uses all available CPU threads. Set to 1 to disable parallel processing.
+    method : str, optional
+        Algorithm method to use:
+
+        - 'blockwise': Improved algorithm with better coordinate handling,
+          memory efficiency, and statistical pre-filtering
+        - 'classic': Original algorithm with traditional implementation
+
+        .. versionadded:: 1.12.0
```
maharshi-gor, line 71: To be removed?

dipy/workflows/denoise.py
maharshi-gor: Please mention here if it is 0, we estimate a sigma. Which threw me off in the help. That is 0.

dipy/denoise/nlmeans.py
```diff
@@ -18,83 +13,219 @@ def nlmeans(
...
-            block_radius=block_radius,
-            rician=rician,
-            num_threads=num_threads,
-        ).astype(arr.dtype)
+        if method == "classic":
+            # Classic method needs 3D sigma array
+            if not isinstance(sigma, np.ndarray):
+                sigma_3d = np.full(arr.shape, sigma, dtype="f8")
+            else:
+                # Ensure sigma is 3D and same shape as arr
```
maharshi-gor: To remove

dipy/denoise/nlmeans.py
```diff
@@ -18,83 +13,219 @@ def nlmeans(
...
+                        sigma_3d = np.full(arr.shape, float(sigma), dtype="f8")
+                    else:
+                        raise ValueError(
+                            f"sigma shape {sigma.shape} incompatible "
+                            f"with arr shape {arr.shape}"
+                        )
+                else:
+                    sigma_3d = np.ascontiguousarray(sigma, dtype="f8")
+        else:
+            # For blockwise method, sigma can be scalar or array
```
maharshi-gor: To remove

dipy/denoise/nlmeans.py
```diff
@@ -18,83 +13,219 @@ def nlmeans(
...
-            denoised_arr[..., i] = nlmeans_3d(
-                arr[..., i],
-                mask=mask,
-                sigma=sigma[..., i],
-                patch_radius=patch_radius,
-                block_radius=block_radius,
-                rician=rician,
-                num_threads=num_threads,
-            ).astype(arr.dtype)
+            # For classic method, we need to convert scalar sigma to 3D array
```
maharshi-gor: To remove

dipy/denoise/nlmeans.py
```diff
@@ -18,83 +13,219 @@ def nlmeans(
...
+            if method == "classic":
+                if isinstance(sigma, np.ndarray):
+                    sigma_vol = np.full(arr[..., i].shape, sigma[i], dtype="f8")
+                else:
+                    sigma_vol = np.full(arr[..., i].shape, sigma, dtype="f8")
+            else:
+                # For blockwise method, use scalar sigma
```
maharshi-gor: To remove

dipy/denoise/nlmeans.py
```diff
@@ -18,83 +13,219 @@ def nlmeans(
...
-        return nlmeans_3d(
-            arr,
-            mask=mask,
-            sigma=sigma,
-            patch_radius=patch_radius,
-            block_radius=block_radius,
-            rician=rician,
-            num_threads=num_threads,
-        ).astype(arr.dtype)
+        if method == "classic":
```
maharshi-gor, line 165: Can we create a method for 3 dim and use the same for each volume in the 4 dim?

dipy/denoise/nlmeans.py
```diff
@@ -18,83 +13,219 @@ def nlmeans(
...
+                f"Unknown method '{method}'. Use 'classic' or 'blockwise'."
+            )
+
+    if isinstance(sigma, np.ndarray) and sigma.size == 1:
+        sigma = sigma.item()
+    if isinstance(sigma, np.ndarray):
+        if not np.issubdtype(sigma.dtype, np.number):
+            raise ValueError("sigma should be an array of floats", sigma)
+
+        # Validation depends on both array dimensions and method
```
maharshi-gor: Remove comments from the PR.

maharshi-gor (changes requested):
Hello @skoudoro ,

Very cool results with the block_wise methods.

I tried it with Stanford_hardi DWI and T1, also I tried with syn_data. Showed the results as expected. num_threads was working fine for me.

Please look at the comments in the review and remove all the comments present in the nlmeans method.

Otherwise, LGTM!

dipy/denoise/nlmeans.py
skoudoro: this is done below, not sure to understand

dipy/denoise/nlmeans.py
skoudoro: No, actually, we need to use this more...

dipy/denoise/nlmeans.py
maharshi-gor: I was thinking writing it so that we do not need to check the dimensionality multiple times to decide how should we treat it. But, I think that will be an over kill.

maharshi-gor:
Thank you for the work @skoudoro . The results are good and the out-of-the-box parallelization helps the user while using it.

PR #3631, opened 2025-09-16
Fixed the bug that caused weird volumes instead of denoised volumes.

iprajwalreddy:
This PR fixes a bug where the denoising step intermittently produced volumes that rendered as nearly all black/white (saturated) instead of properly denoised images.

Fixed by:
Keep processing in floating point throughout the denoise pipeline; remove unintended early dtype casts.

### DIFF
--- a/dipy/denoise/patch2self.py
+++ b/dipy/denoise/patch2self.py
@@ -581,7 +581,7 @@ def _patch2self_version1(
             )
 
         if verbose is True:
-            logger.info("Denoised b0 Volume: ", vol_idx)
+            logger.info("Denoised b0 Volume: %d", vol_idx)
     # Separate denoising for DWI volumes
     train_dwi = _extract_3d_patches(
         np.pad(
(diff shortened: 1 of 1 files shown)


jhlegarreta (changes requested):
Use f-strings when logging messages.

skoudoro (changes requested):
Hi @iprajwalreddy 

Please, create a test that demonstrates the bug and then the fix. 

That is an important rule when you do a bug fix. 

Here, you do a change and we do not know if it really fix the bug that you announced...

iprajwalreddy:
@skoudoro and @jhlegarreta. Thank you for the review, I have added a test case for the bug fixes and also used f-strings when logging the messages. Please let me know in case any changes or updates are required.

Atharva-Shah-2298:
This still has the same issues. I checked I am on master for this, The merged changes are already present my CLI and I am using version 3. The issues are presenting differently. Earlier it was all white, now its a masked white for some reason. I am attaching the output here. 

Screenshot 2025-09-23 at 12 01 00 PM

This is on sherbrook 3 shell and the volumes are the ones on the bval=3000 shell. Version one doesn't have this issue. Only version 3. @iprajwalreddy @skoudoro

Atharva-Shah-2298:
@skoudoro . The same data on @iprajwalreddy 's computer looks fine. We figured out the issue, while installing dipy's optional dependencies, even with --no-cache-dir, because it has dipy[viz,ml]. it installs dipy 1.11 which is the older implementation.

https://github.com/dipy/dipy/blob/master/requirements/optional.txt

skoudoro:
Not sure what you mean @Atharva-Shah-2298 but great that it works as expected!

Atharva-Shah-2298:
When installing dev version, I created a fresh environment, cloned a new dipy repo, and installed it. dipy_denoise had a tripwire error so I installed optional dependencies. But during optional dependency install, the dev version of 1.12 was downggraded to 1.11 because in optional dependency, there is dipy[viz, ml] which installed prebuilt pypi dipy package which is 1.11.

skoudoro:
ok, I see, thanks for the detailed info.  I need to update this script : tools/generate_requirements.py

PR #3634, opened 2025-09-24
BF: expand requirements during its generation

skoudoro:
This address the bug found by @Atharva-Shah-2298 which was the following:

> When installing dev version, I created a fresh environment, cloned a new dipy repo, and installed it. dipy_denoise had a tripwire error so I installed optional dependencies. But during optional dependency install, the dev version of 1.12 was downggraded to 1.11 because in optional dependency, there is dipy[viz, ml] which installed prebuilt pypi dipy package which is 1.11.

Now, the generation tools expand the implicit import.


jhlegarreta:
It is confusing to have the txt files being tracked. Maybe I am missing the reason why the txt files are needed in the first place, but even if they are needed by a given process, I am wondering if they can be generated on the fly by calling the python script before whichever process requires requirements to be in txt format.

skoudoro:
Thank you for your feedback @jhlegarreta. 

To get an answer, I will forward you to the previous discussion: https://github.com/dipy/dipy/pull/3424

jhlegarreta:
I see, thanks, Serge. This need may vanish eventually: https://github.com/conda/conda/issues/10633. As mentioned in that thread, I've seen other large projects move to [pixi](https://pixi.sh/latest/). I have not used it myself, but what I've heard from it was positive.

jhlegarreta (approved):
Have not tested, but thanks for doing this.

skoudoro:
Thanks for the feedback.  I will look into `pixi`. I also need to check the difference with `uv`.

Garyfallidis:
Thank you @skoudoro !
You may need to rebase things across after this.

PR #3635, opened 2025-09-26
BF: fix CVXPY warning

deka27:
Fixing the CVXPY warning in CI

### DIFF
--- a/dipy/workflows/tests/test_reconst_mapmri.py
+++ b/dipy/workflows/tests/test_reconst_mapmri.py
@@ -40,8 +40,10 @@ def reconst_mmri_core(flow, lap, pos):
         mmri_flow = flow()
 
         msg = "Optimization did not find a solution"
+        cvxpy_msg = "Solution may be inaccurate"
         with warnings.catch_warnings():
             warnings.filterwarnings("ignore", message=msg, category=UserWarning)
+            warnings.filterwarnings("ignore", message=cvxpy_msg, category=UserWarning)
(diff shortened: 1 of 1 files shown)


pjsjongsung:
I did a bit of backtracking, and the warning is no there until cvxpy version of 1.7.1. It starts at 1.7.2. A brief search seems to find the problem in line 537 of `dipy.core.optimize.py` where cvxpy's problem solver is called. My guess is it is getting `optimal_inaccurate` or `user_limit` as it's status. (Referring to cvxpy package's [settings.py](https://github.com/cvxpy/cvxpy/blob/master/cvxpy/settings.py) line 65). I have looked at their [release notes](https://github.com/cvxpy/cvxpy/releases), but could not find what suddenly started to create this warning.

deka27:
Based off on @pjsjongsung comment . Did some looking . Starting with cvxpy 1.7.2, the library throws warnings when a solution has an `optimal_inaccurate` status. The old code only accepted `status == "optimal"`, which meant it was unnecessarily rejecting solutions that were actually fine.

So I updated `dipy/core/optimize.py` to check `status in cvxpy.settings.SOLUTION_PRESENT`. This covers `OPTIMAL`, `OPTIMAL_INACCURATE`, and `USER_LIMIT`. The commit i pushed now properly distinguishes between cases with no solution (returns zeros) and those with an inaccurate one (uses it but logs a warning). I also added filters to silence any leftover cvxpy warnings in the test suite.

Checking if this works.

Issue #3638, opened 2025-10-11
UnboundLocalError: local variable 'fits' referenced before assignment


Mengyao-Li:
### Summary

This is code:
```
import numpy as np
from dipy.core.sphere import unit_icosahedron
from dipy.io.image import load_nifti, save_nifti
from dipy.reconst.bingham import sh_to_bingham

sphere = unit_icosahedron.subdivide(n=5)
max_search_angle = 45

def calculate_bingham_metrics_from_sh(sh_file, sphere, max_angle):
    sh_coeff, affine = load_nifti(sh_file)

    if sh_coeff.ndim != 4:
        raise ValueError("Input Nifiti file must be a 4D image of SH coefficients.")
    
    print('Fitting Bingham functions to SH coefficients...')

    bingham_metrics = sh_to_bingham(sh_coeff, sphere, max_angle)

    return bingham_metrics, affine

def main():
    fodf_fixed_path = 'wm_fod.nii.gz'
    mask_path = 'nodif_brain_mask.nii.gz'

    metrics_fixed, affine_fixed = calculate_bingham_metrics_from_sh(fodf_fixed_path, sphere, max_search_angle)
   
main()
```

This is error:

> Traceback (most recent call last):
>   File "/home/lmy/mri/ODI/test.py", line 27, in <module>
>     main()
>   File "/home/lmy/mri/ODI/test.py", line 25, in main
>     metrics_fixed, affine_fixed = calculate_bingham_metrics_from_sh(fodf_fixed_path, sphere, max_search_angle)
>   File "/home/lmy/mri/ODI/test.py", line 17, in calculate_bingham_metrics_from_sh
>     bingham_metrics = sh_to_bingham(sh_coeff, sphere, max_angle)
>   File "/home/lmy/.local/lib/python3.10/site-packages/dipy/reconst/bingham.py", line 781, in sh_to_bingham
>     [fits, npeaks_final] = _single_sf_to_bingham(
>   File "/home/lmy/.local/lib/python3.10/site-packages/dipy/reconst/bingham.py", line 177, in _single_sf_to_bingham
>     return fits, n
> UnboundLocalError: local variable 'fits' referenced before assignment

This error occurs when the input SH/ODF volume includes voxels with empty data (e.g., containing zeros).

### Platform

ubuntu22.04

### DIPY version

1.11.0

### Python version

Python 3.10

### Additional information

I’m preparing a pull request to fix this issue

Mengyao-Li:
[...]

PR #3639, opened 2025-10-11
fix: BUG: fix UnboundLocalError in _single_sf_to_bingham when no valid peaks found

Mengyao-Li:
#3638

### DIFF
--- a/dipy/reconst/tests/test_bingham.py
+++ b/dipy/reconst/tests/test_bingham.py
@@ -5,7 +5,10 @@ from numpy.testing import (
     assert_array_less,
 )
 
-from dipy.data import get_sphere, get_fnames
+from dipy.core.gradients import gradient_table
+from dipy.data import get_fnames, get_sphere
+from dipy.io.gradients import read_bvals_bvecs
+from dipy.io.image import load_nifti_data
 from dipy.reconst.bingham import (
     _bingham_fit_peak,
     _convert_bingham_pars,
@@ -18,11 +21,7 @@ from dipy.reconst.bingham import (
     sf_to_bingham,
     sh_to_bingham,
 )
-from dipy.reconst.shm import sf_to_sh
-from dipy.io.image import load_nifti_data
-from dipy.io.gradients import read_bvals_bvecs
-from dipy.core.gradients import gradient_table
-from dipy.reconst.shm import CsaOdfModel
+from dipy.reconst.shm import CsaOdfModel, sf_to_sh
 
 
 def setup_module():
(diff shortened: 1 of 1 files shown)


arokem:
Thanks for catching this. Is there any chance you could add a test that would catch this error? Something like the code that you reported in #3638, but that works with the example data that ships with DIPY, should work.

Mengyao-Li:
Thanks for the feedback! I have added a regression test using CSA ODF on small_64D to ensure sh_to_bingham handles empty voxels.
https://github.com/dipy/dipy/pull/3639/commits/5efb7e1708cbd5eefb0a202ac4beb0f2ebd5223a

dipy/reconst/tests/test_bingham.py
```diff
@@ -221,3 +226,37 @@ def test_bingham_from_sh():
...
+
+    Uses DIPY's built-in small DWI dataset to fit SH coefficients via CSA ODF,
+    then verifies that sh_to_bingham handles empty voxels correctly.
+    """
+    # Load small diffusion dataset
+    dwi_fname, bval_fname, bvec_fname = get_fnames(name="small_64D")
+    data = load_nifti_data(dwi_fname)
+    bvals, bvecs = read_bvals_bvecs(bval_fname, bvec_fname)
+
+    if data.ndim != 4:
```
arokem: I don't think this check is needed

Mengyao-Li:
Thanks! I’ve removed the unnecessary ndim check as suggested.

dipy/reconst/tests/test_bingham.py
```diff
@@ -19,6 +19,11 @@
     sh_to_bingham,
 )
 from dipy.reconst.shm import sf_to_sh
+from dipy.io.image import load_nifti_data
+from dipy.io.gradients import read_bvals_bvecs
+from dipy.core.gradients import gradient_table
+from dipy.reconst.shm import CsaOdfModel
+from unittest import SkipTest
```
arokem: I believe this import is not needed

```suggestion
```

Mengyao-Li:
Thank you for pointing that out — I’ve removed the unnecessary SkipTest import.

maharshi-gor:
Hi @Mengyao-Li 

Thanks for the work!

The code format is failing due to the code not meeting the specified format.

You can use pre-commit for that.

That you can by installing dev dependencies:
```
pip install -r requirements/style.txt
```

You can run then

```
pre-commit install
```

This will make sure by default every commit is automatically gets the format.

You can automatically fix all the current issue by running

```
pre-commit run --all-files
```

Welcome to DIPY community! 🚀 

Best,
Maharshi

Mengyao-Li:
Thanks! I’ve installed pre-commit and ran pre-commit run --all-files. The code has been formatted and pushed again.

maharshi-gor:
The CIs are failing outside of her work @skoudoro can you check?

Issue #3621, opened 2025-08-12
filename for statefulsurface and statefultractogram cannot have period in them anymore


frheault:
### Summary

Due to the PR #3593 (The `utils` function `split_name_with_gz` was removed)
The line with extension = "".join(filename.suffixes) make it impossible to load or save tractogram and surface with SFT and SFS if they contain a period in them.

It would require to compare the N last caracter to an extension of len N (requiring a for loop over the extension.

### Platform

all

### DIPY version

1.12

### Python version

3.12

skoudoro:
Hi @frheault,

Thank you for the feedback. We should have done a deprecation cycle. 

Also, it is strange that no tests detect that. Can you provide extensions example? it will be helpful for future tests to add

frheault:
> Hi @frheault,
> 
> Thank you for the feedback. We should have done a deprecation cycle. 
> 
> Also, it is strange that no tests detect that. Can you provide extensions example? it will be helpful for future tests to add

I was trying to save a tractogram with my step size in the filename for some testing "repulsion_sz_1.0.trk" and I wasn't understanding the crash.

I believe this was made to catch .nii.gz as if they were the only filename with period. I believe the previous function was very specific to gz, but it could be extended to any double extension we routinely encountered?

skoudoro:
closed by #3648

PR #3648, opened 2025-10-27
Revert split extension (for SFT and SFS)

frheault:
Fix problem as described in #3621 

Since our field routinely see period in filename and only one extension suffixe exists (.gz) this handle correctly filenames like smoothwm.L.surf.gii or smoothwm.L.surf.gii.gz or my.brain_2025.nii or my.brain_2025.nii.gz

Also in our field the extension is always JUST the last period (except .gz), this would have to be customize if one day we invent new double or triple extension.

(This is close to what it was before)

### DIFF
--- a/dipy/io/tests/test_utils.py
+++ b/dipy/io/tests/test_utils.py
@@ -1,3 +1,4 @@
+from pathlib import Path
 import tempfile
 from urllib.error import HTTPError, URLError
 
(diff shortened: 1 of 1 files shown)


dipy/io/surface.py
```diff
@@ -78,9 +83,9 @@ def load_surface(
     vtk_ext = [".vtk", ".vtp", ".obj", ".stl", ".ply"]
     freesurfer_ext = [".gii", ".gii.gz", ".pial", ".nofix", ".orig", ".smoothwm", ".T1"]
 
-    ext = "".join(Path(fname).suffixes).lower()
-
-    if not any(ext.endswith(_ext) for _ext in freesurfer_ext + vtk_ext):
+    name, ext = split_filename_extension(fname)
+    print(name, ext)
```
skoudoro: print to remove

dipy/io/utils.py
```diff
@@ -504,3 +504,36 @@ def recursive_compare(d1, d2, level="root"):
     else:
         if np.dtype(d1).itemsize != np.dtype(d2).itemsize:
             raise ValueError(f"Values {d1}, {d2} do not match at level {level}")
+
+
+def split_filename_extension(filename):
+    """Splits the filename and its extension(s).
```
skoudoro: `Split` remove the `s`

dipy/io/utils.py
```diff
@@ -504,3 +504,36 @@ def recursive_compare(d1, d2, level="root"):
...
+    In our field filename can have period in it (e.g. smoothwm.L.surf.gii)
```
skoudoro, line 512: missing empty line between short and long description.

dipy/io/utils.py
```diff
@@ -504,3 +504,36 @@ def recursive_compare(d1, d2, level="root"):
     else:
         if np.dtype(d1).itemsize != np.dtype(d2).itemsize:
             raise ValueError(f"Values {d1}, {d2} do not match at level {level}")
+
+
+def split_filename_extension(filename):
```
skoudoro, line 509: can you add the test of this function in test_utils

skoudoro (approved):
Thank you for the update, looks good to me. 

Doing some tests locally and if all ok, I will go ahead and merge

frheault:
I agree my.brain_2025.nii is a user choice (and a bad one), but  things like tractogram_step_size_1.0.trk or brain_bet_v_1.0.nii.gz are still very common. And even if it can be consider a bad practice it is still very common to simply append ''extension'' to filename (like DSI studio: 100206.src.gz.gqi.1.7.fib.gz). I find it annoying myself when people do fodf.nii.gz.trk or dwi.nii.gz.nii.gz

I do not know what is the best way here, but this could be a longer term goal to push people to stop using period in filename except for extensions (And do a transition with warning when a period is detected before the final valid extension)

dipy/io/tests/test_utils.py
```diff
@@ -256,3 +257,35 @@ def test_read_img_arr_or_path(rng):
...
+        ("test.nii.nii", "test.nii", ".nii"),
+        ("my_brain_2025.nii", "my_brain_2025", ".nii"),
+        ("my_brain_2025.nii.gz", "my_brain_2025", ".nii.gz"),
+        (".my.file-name.gii.gz", ".my.file-name", ".gii.gz"),
+    ],
+)
+def test_split_filename_extension(filename, expected_name, expected_extension):
+    name, ext = split_filename_extension(filename)
+    assert name == expected_name
+    assert ext == expected_extension
```
skoudoro, line 289: can you address @jhlegarreta comment and add a test with `Path` object instead of `string` only

skoudoro:
Thank you for this @frheault, merging

PR #3650, opened 2025-10-28
BF: update first interaction Github action parameters 

skoudoro:
This PR fixes the following error/warnings:

Unexpected input(s) 'repo-token', 'issue-message', 'pr-message', valid inputs are ['issue_message', 'pr_message', 'repo_token']


skoudoro:
It seems to not take into account the new change in the PR. 

So I will go ahead and merge it. 

Then, check if it works as expected on a future new PR

PR #3652, opened 2025-10-28
NF:  Reduce default iteration counts and make level_iters configurable in motion correction

deka27:
Reduced default `level_iters` from `[10000, 1000, 100]` to `[1000, 500, 100]` for faster motion correction.
Added `level_iters` parameter to `dipy_correct_motion`.

### DIFF
--- a/dipy/workflows/align.py
+++ b/dipy/workflows/align.py
@@ -394,7 +394,7 @@ class ImageRegistrationFlow(Workflow):
         out_quality : string, optional
             Name of the file containing the saved quality metric.
         """
-        if level_iters == (1000, 500, 100):
+        if tuple(level_iters) == (1000, 500, 100):
             logger.info(
                 "Default level_iters have been updated to [1000, 500, 100] for "
                 "performance improvement. Identical results are expected. In case "
@@ -899,7 +899,7 @@ class MotionCorrectionFlow(Workflow):
         out_affine : string, optional
             Name for the saved affine matrix.
         """
-        if level_iters == (1000, 500, 100):
+        if tuple(level_iters) == (1000, 500, 100):
             logger.info(
                 "Default level_iters have been updated to [1000, 500, 100] for "
                 "performance improvement. Identical results are expected. In case "


dipy/align/imaffine.py
```diff
@@ -953,7 +953,7 @@ def __init__(
             self.metric = MutualInformationMetric()
 
         if level_iters is None:
-            level_iters = [10000, 1000, 100]
+            level_iters = [1000, 500, 100]
```
skoudoro, line 956: same comment as above. or we should add a warnings

dipy/align/_public.py
```diff
@@ -515,7 +515,7 @@ def affine_registration(
 
     """
     pipeline = pipeline or ["center_of_mass", "translation", "rigid", "affine"]
-    level_iters = level_iters or [10000, 1000, 100]
+    level_iters = level_iters or [1000, 500, 100]
```
skoudoro: For backward compatibility, this can not be changed. 

Also, Are you sure this change does not impact the result?

dipy/workflows/align.py
```diff
@@ -859,13 +860,19 @@ def run(
         bvecs_tol : float, optional
             Threshold used to check that norm(bvec) = 1 +/- bvecs_tol
             b-vectors are unit vectors
+        level_iters : list of int, optional
```
skoudoro: `variable int` instead. otherwise, it will not work

skoudoro (changes requested):
Hi @deka27,

Thank you for starting this.

See below some comments

pjsjongsung:
I agree with Serge. To test whether if reducing the level_iters default can work, we need thorough testing in multiple images. I think this should be a separate PR, and just adding level_iters parameter be one PR.

Garyfallidis:
@skoudoro what type of warning would be appropriate in this case?

skoudoro:
> @skoudoro what type of warning would be appropriate in this case?

I would import logger, and do a `logger.info` saying that "levels of iter has been updated for performance improvement. Identical results are expected. In case of change you can revert this change by defining level_iters=[1000, 100, 10]"

or something similar.

skoudoro:
I would keep this message for 2 releases before removing it @Garyfallidis

dipy/workflows/align.py
```diff
@@ -859,13 +867,22 @@ def run(
...
+        level_iters : variable int, optional
+            The number of iterations at each level of the Gaussian pyramid.
         out_dir : string or Path, optional
             Directory to save the transformed image and the affine matrix.
         out_moved : string, optional
             Name for the saved transformed image.
         out_affine : string, optional
             Name for the saved affine matrix.
         """
+        if level_iters == (1000, 500, 100):
```
skoudoro: this is not working, you need to do `if tuple(level_iters) == (1000, 500, 100)`

deka27:
Done. Should i squash?

deka27:
Getting HTTP error for fetch data. not related to this PR

Garyfallidis:
@skoudoro this looks ready to go. What do you think?

skoudoro:
> @skoudoro this looks ready to go. What do you think?

doing a break with merging. I need to fix CI's first make sure there are not hidden issue. otherwise it looks good indeed

maharshi-gor:
Hello @deka27 Please rebase the PRs for the CIs to run

PR #3653, opened 2025-10-28
RF: Fix imports in reconst

deka27:
This fixes the bug, we get in the CI for undefined OS and Logging.

### DIFF
--- a/dipy/workflows/reconst.py
+++ b/dipy/workflows/reconst.py
@@ -1,6 +1,6 @@
+from ast import literal_eval
 import logging
 import os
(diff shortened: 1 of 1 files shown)


deka27:
very odd. i ran pre-commit, Ruff automatically did stage changes, but now the same changes done by ruff on local are getting rejected by ruff on CI.

skoudoro:
Incorrect fix and already done by #3646 which has been merged. Closing

deka27:
Ah, nvm, i understood why. the call was wrong. logging vs logger

PR #3651, opened 2025-10-28
RF: handle bingham tests warnings

skoudoro:
CI's are currently failing due to warnings in bingham tests. 

Unfortunately, we could not catching it before due to other CI's error. 

Clean up continues....

PR #3655, opened 2025-10-28
CI: remove fetch-data job

skoudoro:
This CI job was there to download and cache all the datasets in one step.
 
Unfortunately, it does not work. it regularly fails and bring no benefit, only drawback.
So,

- I removed this job
- Make sure we cache the datasets after each test job completes
- make sure to check cache when we start a test job
- Also a exponential backoff has been added for the diverse http errors.  
- A new mirror server with all the data has been created. If data failed during download in zenodo for example, we try on our mirror server which has less restriction.
- Also, it seems we had issue with downloading txt file. It was causing the failure of md5 comparison. Zenodo was intermittently serving a gzip-compressed version of `txt` file. The compression was happening without proper Content-Encoding headers. So, our fetcher was saving the compressed bytes directly, causing MD5 mismatches (This one was hard to find). In future, we should not allow `txt`. better to have a zipped package.
- Fix Tensorflow issue which was not compatible with the decorator `@warning_for_keywords()`
- Update simulation files which was not compatible with numpy 2.x
- removed python 3.10, added python 3.14


maharshi-gor:
Hi @skoudoro please rebase this PR, as some of the CIs are failing with warning issue of the `sh_to_bingham` which we just merge in #3651

skoudoro:
I am currently satisfy with the CI's update. this PR can be merged

Failing CI's are related to #3623

one Conda is very slow, I will introspect this issue in a new PR

PR #3658, opened 2025-10-29
Added interpolation parameter to dipy_apply_transform

azizasaber:
looks like dipy_apply_transform did not have an argument to add interpolation type. so by default it uses linear interpolation. added the two types, linear and nearest to ApplyTransformFlow and the corresponding tests.


dipy/workflows/align.py
```diff
@@ -582,11 +593,16 @@ def run(
 
             # Transforming the image
             if moving_image_full is None:
-                transformed = mapping.transform(moving_image)
+                print(f"Using interpolation: {interpolation}")
```
skoudoro: print not necessary

dipy/workflows/tests/test_align.py
```diff
@@ -429,6 +442,29 @@ def test_apply_affine_transform():
...
 
+        apply_trans.run(
+            static_image_file,
+            images,
+            out_dir=temp_out_dir,
+            transform_map_file=out_affine,
+            out_file="transformed_linear.nii.gz",
+        )
+
+        # Checking for the transformed file.
```
skoudoro: coment to remove

dipy/workflows/tests/test_align.py
```diff
@@ -429,6 +442,29 @@ def test_apply_affine_transform():
...
+        apply_trans.run(
+            static_image_file,
+            images,
+            out_dir=temp_out_dir,
+            transform_map_file=out_affine,
+            interpolation="nearest",
+            out_file="transformed_nearest.nii.gz",
+        )
+
+        # Checking for the transformed file.
```
skoudoro: comment to remove

azizasaber:
> Hi @azizasaber,
> 
> Thank you for this. See below some comments. Also, Can you avoid to edit the requirements files. Thank you

Hi, thank you for the review. Made the requested changes. Apparently hook automatically changed requirements files, now it should be fine.

skoudoro (approved):
Thank you for this, all good, merging

PR #3665, opened 2025-11-28
FIX: Correct path construction logic in fetcher data loader, Resolve path construction errors in read_bundles_2_subjects

nishantkai11:
Description: Fixed a regression in read_bundles_2_subjects where dname was treated as a string instead of a tuple. Changes:

Unpacked dname tuple to construct base_path correctly.

Fixed missing path separators (replaced commas with slashes) in load_nifti and load_tractogram calls.

Corrected loop indentation and variable scope for streamlines.

Verification: Verified locally by running doc/examples/viz_advanced.py, which now executes without crashing.

### DIFF
--- a/dipy/data/fetcher.py
+++ b/dipy/data/fetcher.py
@@ -2771,34 +2771,30 @@ def read_bundles_2_subjects(
     .. footbibliography::
 
     """
-    dname = Path(dipy_home) / "exp_bundles_and_maps", "bundles_2_subjects"
+    dname = Path(dipy_home) / "exp_bundles_and_maps" / "bundles_2_subjects"
 
     from dipy.tracking.streamline import Streamlines
 
     res = {}
 
     if "t1" in metrics:
-        data, affine = load_nifti(Path(dname) / subj_id, "t1_warped.nii.gz")
+        data, affine = load_nifti(Path(dname) / subj_id / "t1_warped.nii.gz")
         res["t1"] = data
 
     if "fa" in metrics:
-        fa, affine = load_nifti(Path(dname) / subj_id, "fa_1x1x1.nii.gz")
+        fa, affine = load_nifti(Path(dname) / subj_id / "fa_1x1x1.nii.gz")
         res["fa"] = fa
 
     res["affine"] = affine
-
     for bun in bundles:
         streams = load_tractogram(
-            Path(dname) / subj_id,
-            "bundles",
-            f"bundles_{bun}.trk",
+            Path(dname) / subj_id / "bundles" / f"bundles_{bun}.trk",
             "same",
             bbox_valid_check=False,
         ).streamlines
 
         streamlines = Streamlines(streams)
         res[bun] = streamlines
-
     return res
 
 


dipy/data/fetcher.py
```diff
@@ -2778,27 +2778,26 @@ def read_bundles_2_subjects(
     res = {}
 
     if "t1" in metrics:
-        data, affine = load_nifti(Path(dname) / subj_id, "t1_warped.nii.gz")
-        res["t1"] = data
+       
+       base_path = dname[0] / dname[1]
+       data, affine = load_nifti(base_path / subj_id/ "t1_warped.nii.gz")
```
skoudoro: I do not think it is needed, you just need to replace `,` by `/` above. So,

`data, affine = load_nifti(Path(dname) / subj_id / "t1_warped.nii.gz")`

dipy/data/fetcher.py
```diff
@@ -2778,27 +2778,26 @@ def read_bundles_2_subjects(
...
+        
+    res["t1"] = data
 
     if "fa" in metrics:
-        fa, affine = load_nifti(Path(dname) / subj_id, "fa_1x1x1.nii.gz")
+        fa, affine = load_nifti(base_path / subj_id / "fa_1x1x1.nii.gz")
```
skoudoro: same, keep `Path(dname)`, the other change is good

dipy/data/fetcher.py
```diff
@@ -2778,27 +2778,26 @@ def read_bundles_2_subjects(
...
         res["fa"] = fa
 
     res["affine"] = affine
-
     for bun in bundles:
         streams = load_tractogram(
-            Path(dname) / subj_id,
-            "bundles",
-            f"bundles_{bun}.trk",
+            base_path / subj_id / "bundles" / f"bundles_{bun}.trk",
```
skoudoro: `Path(dname)` instead of `base_path`

dipy/data/fetcher.py
```diff
@@ -2778,27 +2778,26 @@ def read_bundles_2_subjects(
...
             "same",
             bbox_valid_check=False,
         ).streamlines
-
+
```
skoudoro: empty space to remove

skoudoro:
> Why was this missed by the CI?

because CI's do not run any tutorials.

jhlegarreta:
> because CI's do not run any tutorials.

I vaguely recall having discussed or mentioned this, but unless they exceed the max runtime, they should be run e.g. once a week. If they exceed runtime, maybe there is a not very costly way to split them across separate runs.

skoudoro:
> I vaguely recall having discussed or mentioned this, but unless they exceed the max runtime, they should be run e.g. once a week. If they exceed runtime, maybe there is a not very costly way to split them across separate runs.

yes, we have discussed this and agree with that. 

It is on the TODO. I would like to do it once for all and not a quick and dirty way, that's why it is not yet there. Also, tons of other priorities...

skoudoro:
Hi @nishantkai11,

will you find time to finalize this work or can I update it in your PR?

Please, let me know what you prefer. Thank you for your feedback

nishantkai11:
> Hi @nishantkai11,
> 
> will you find time to finalize this work or can I update it in your PR?
> 
> Please, let me know what you prefer. Thank you for your feedback

  hey i will do it today itself

nishantkai11:
> Hi @nishantkai11,
> 
> Good catch! thank you for this!
> 
> See below some comments, then we can go ahead and merge (I believe, the update can be simpler)
> 
> Thank you

Hi @skoudoro , thanks for the review!

I tested `Path(dname)` locally, but it raises a `TypeError` because `dname` is passed as a tuple: `(path_object, 'folder_name')` rather than a single path string.

Here is the traceback I encountered when using `Path(dname)`:
`TypeError: expected str, bytes or os.PathLike object, not tuple`

That is why I unpacked it explicitly using `dname[0] / dname[1]`. If there is a preferred helper function to handle this tuple automatically, please let me know and I will update it!

I will remove the empty space immediately.

maharshi-gor:
Hi @nishantkai11 

What @skoudoro  meant is to keep the Path() wrapper around the variable so it works where the absolute path are not provided. I think you can adapt it like `Path(*dname)`, which will unpack the path properly rather than creating a variable.

skoudoro:
Hi @nishantkai11,

> I tested Path(dname) locally, but it raises a TypeError because dname is passed as a tuple: (path_object, 'folder_name') rather than a single path string.

You get this error because you need to update the `dname`. 

```diff
- dname = Path(dipy_home) / "exp_bundles_and_maps", "bundles_2_subjects"
+ dname = Path(dipy_home) / "exp_bundles_and_maps" / "bundles_2_subjects"
```

then all the `Path(dname)` will work normally. 

Thank you for your future update

skoudoro (approved):
All good, merging. 

Thanks @nishantkai11 for catching this!

nishantkai11:
> All good, merging.
> 
> Thanks @nishantkai11 for catching this!

Thank you so much @skoudoro .

PR #3669, opened 2025-12-12
DOC: Prefer BibTeX reference in GQSI example documentation

jhlegarreta:
Prefer using BibTeX reference in GQSI example documentation for automatic reference linking.


skoudoro (approved):
good catch! thanks merging

PR #3671, opened 2025-12-15
Increased speed and lisibility of connectivity matrix, added weights

DelinteNicolas:
**Changes:**
- improved lisibility, reduced number of lines
- added weights parameter to adapt to weighted streamline connectivity tools such as SIFT2. The weight of each streamline is added to the connectivity matrix instead of 1. The weights should be stored in a 1D array of size n, for n streamlines.
- new version is faster than current implementation when removing small streamlines and 'inclusive' is set to False

**Suggestions for further code clean up:**
- remove 'ndbincount' function which was used in a older version of the connectivity matrix function
- move small streamlines removal out of connectivity matrix function
- remove the 'if not isinstance(streamlines, Streamlines)' condition

### DIFF
--- a/dipy/tracking/utils.py
+++ b/dipy/tracking/utils.py
@@ -184,11 +184,11 @@ def connectivity_matrix(
 
     if weights is None:
         weights = np.ones(len(streamlines))
-        matrix = np.zeros((np.max(label_volume) + 1, np.max(label_volume) + 1))
-    else:
         matrix = np.zeros(
             (np.max(label_volume) + 1, np.max(label_volume) + 1), dtype=np.int64
         )
+    else:
+        matrix = np.zeros((np.max(label_volume) + 1, np.max(label_volume) + 1))
 
     if discard_stream_size > 0:
         (keep_idx,) = np.where(streamlines._lengths > discard_stream_size)


dipy/tracking/utils.py
```diff
@@ -139,6 +140,8 @@ def connectivity_matrix(
     symmetric : bool, optional
         Symmetric means we don't distinguish between start and end points. If
         symmetric is True, ``matrix[i, j] == matrix[j, i]``.
+    weights : ndarray
```
skoudoro: `ndarray, optional`, instead

dipy/tracking/utils.py
```diff
@@ -139,6 +140,8 @@ def connectivity_matrix(
...
+        A 1D array containing the weights of each streamline.
```
skoudoro: > The weights should be stored in a 1D array of size n, for n streamlines.

add this information here and check it in the code please

dipy/tracking/utils.py
```diff
@@ -172,17 +175,24 @@ def connectivity_matrix(
...
-    matrix = np.zeros(
-        (np.max(label_volume) + 1, np.max(label_volume) + 1), dtype=np.int64
-    )
+    matrix = np.zeros((np.max(label_volume) + 1, np.max(label_volume) + 1),
+                      dtype=np.int64)
 
     mapping = defaultdict(list)
     lin_T, offset = _mapping_to_voxel(affine)
 
+    # Setting weights to 1 if not specified
```
skoudoro: unnecessary comment

dipy/tracking/utils.py
```diff
@@ -172,17 +175,24 @@ def connectivity_matrix(
...
+    if weights is None:
+        weights = np.ones(len(streamlines))
+
+    # Removing small streamlines
```
skoudoro: unnecessary comment

DelinteNicolas:
The changes have been made and the tests seem to have passed for the most part. For the remaining failing tests, they are either due to a "No space left on device" error or, for the four tests failing on macos and ubuntu, do not seem to be linked to functions updated in this pull request.

skoudoro:
Thank you for the update and feedback @DelinteNicolas.

Can you quickly check why those tests failed (it seems related)

```
FAILED tracking/tests/test_propspeed.cpython-311-darwin.so::test_deterministic_performances
FAILED tracking/tests/test_propspeed.cpython-311-darwin.so::test_probabilistic_performances
FAILED tracking/tests/test_propspeed.cpython-311-darwin.so::test_ptt_performances
FAILED tracking/tests/test_tractogen.cpython-311-darwin.so::test_tractogram_reproducibility
```

DelinteNicolas:
The failing tests were indeed related, I have modified the code to include other types of generator in the 'if' condition. The macos and ubuntu tests used cpython generators which were not caught with the previous 'if' statement. This issue should be fixed now.

dipy/tracking/utils.py
```diff
@@ -179,11 +183,19 @@ def connectivity_matrix(
     mapping = defaultdict(list)
     lin_T, offset = _mapping_to_voxel(affine)
 
+    if type(streamlines).__name__ == "generator":
+        streamlines = Streamlines(streamlines)
```
skoudoro, line 183: small reminder here: Do we want to release the generator or return a new generator for memory management reason. 
But this is a general question and not specific for this PR so not a blocker.

skoudoro (changes requested):
Hi @DelinteNicolas,

Overall, Looks good to me, nice clean up.

It would be great if you could add a basic test with different weight values in test_connectivity_matrix. 

After that, it should be ready to be merged. 

Does anyone has any other comments?

arokem:
Yeah, definitely need a test for the new key-word argument. Otherwise, looks good!

DelinteNicolas:
The tests for the weighted connectivity matrix have been added.

PR #3675, opened 2025-12-23
Adding synthseg pytorch model

pjsjongsung:
This PR adds a pytorch version of the synthseg model. ([paper](https://www.sciencedirect.com/science/article/pii/S1361841523000506)) Major changes include synthseg in nn, tests and data connected to the model, and utils.py in segment to mimic a postprocessing behavior that is in the original model pipeline.


skoudoro (commented):
Overall, looks good to me. Let me run it locally

arokem:
Should we include the license of the original work  (https://github.com/BBillot/SynthSeg/blob/master/LICENSE.txt) together with this adapted code? I assume you used their code as a reference?

Garyfallidis:
That or consider switching to Apache license. Many projects did that in the past. Nonetheless AFAIK both licenses are compatible.

arokem:
Yep - the licenses are definitely compatible! I just think we need to add the text of the license and particularly the copyright of the original authors. I think we've done things like that before (e.g., for code we took from skimage).

pjsjongsung:
Made the necessary changes for fixing bugs. Need some advice on where to add the text of the license @skoudoro @arokem . Should it be a separate file in the same folder? I assume we are talking about this above? [dipy/doc/tools/LICENSE.txt](https://github.com/dipy/dipy/blob/0e3155a8df201d0ab1259a0c54f4d772e206ae8d/doc/tools/LICENSE.txt)

skoudoro:
put a link, do not copy the entire license, You can do like this @pjsjongsung : https://github.com/dipy/dipy/blob/0e3155a8df201d0ab1259a0c54f4d772e206ae8d/dipy/align/cpd.py#L1-L16

skoudoro:
Thank you for the update @pjsjongsung. 

For now, the first tests are giving me an empty image. Can you check why? I will also try on HCP to see

image

skoudoro:
After some feedback from @pjsjongsung, it works on my laptop, see below on our famous stanford_hardi t1 data:

image

We just need to return all the labels as a dict, then, I will go deeper through the code and it should be ready to be merged.

Getting closer!

skoudoro:
Also, make sure to run pre-commit to fix all format error. currently code format CI is failing @pjsjongsung

dipy/nn/torch/synthseg.py
```diff
@@ -0,0 +1,633 @@
...
+else:
+
+    class Module:
+        pass
+
+    logger.warning(
+        "This model requires Pytorch.\
+                    Please install these packages using \
+                    pip."
+    )
```
skoudoro: Warning to remove and not necessary. `SynthSeg`  init function should do the job

dipy/nn/torch/synthseg.py
```diff
@@ -0,0 +1,633 @@
...
+        """
+        if not have_torch:
+            raise torch()
+
+        log_level = "INFO" if verbose else "CRITICAL"
+        set_logger_level(log_level, logger)
+
+        self.model = self.init_model()
+        self.model.eval()
+        if use_cuda and torch.cuda.is_available():
```
skoudoro: I would split this `if` in 2, just to explain to the user what is going on:

```
if use_cuda:
  if torch.cuda.is_available():
     self.device = torch.device("cuda")
   else:
      logger.warning("CUDA Requested but Not found, switching to CPU")
      self.device = torch.device("cpu")
```

dipy/nn/torch/synthseg.py
```diff
@@ -0,0 +1,633 @@
...
+        }
+
+    def init_model(self, model_scale=24, n_levels=5, output_channels=33):
+        return Model(
+            model_scale=model_scale, n_levels=n_levels, output_channels=output_channels
+        )
+
+    def fetch_default_weights(self):
+        """Load the model pre-training weights to use for the fitting.
+        While the user can load different weights, the function
```
skoudoro, line 404: add empty line between short and long description

pjsjongsung:
Pushed with fixes and workflow. What I found was that when running tests, if there's a file with the name that is uses in workflow tests, it will create errors due to the output file name already existing. I ran into this issue and added a force flag in the segment workflows tests because it made more sense (pytest should run regardless of what file you have in the working directory). I personally think this force=True should happen in all workflows test, but I can also remove the force flag back if needed.

pjsjongsung:
Does anyone know why I'm getting permission denied when saving csv files in the tests here?

dipy/workflows/segment.py
```diff
@@ -499,6 +497,13 @@ def run(
 
                 save_nifti(tissue_out_path, segmentation_final, affine)
                 save_nifti(opve, PVE, affine)
+                csv_object.write("0,Background\n")
+                for i in range(1, nclass + 1):
+                    csv_object.write(f"{i},Tissue_{i}\n")
+                csv_object.write(
+                    "Due to the nature of the HMRF algorithm"
```
skoudoro: missing a leading `#` character to be labelled as a comment

dipy/workflows/tests/test_segment.py
```diff
@@ -186,7 +191,7 @@ def test_classify_tissue_flow():
                 "wm_threshold": 0.5,
                 "out_dir": out_dir,
             }
-            flow = ClassifyTissueFlow()
+            flow = ClassifyTissueFlow(force=True)
```
skoudoro: why do you need force, it does not make sense to me , knowing that `out_dir` should be empty here

pjsjongsung: Maybe it's just an issue with pytest. When I was running on the luddy server, it had an issue if force was False,(io_it was not fetching anything due to this) which I suspected it to be from having the corresponding file name in the local repo

dipy/nn/utils.py
```diff
@@ -135,15 +135,13 @@ def get_bounds(shape, affine, *, considered_points="corners"):
 def transform_img(
     image,
     affine,
-    *,
```
skoudoro, line 138: to keep

dipy/nn/utils.py
```diff
@@ -135,15 +135,13 @@ def get_bounds(shape, affine, *, considered_points="corners"):
...
     target_voxsize=None,
     final_size=None,
     order=3,
     considered_points="corners",
 ):
     """
     Function to transform images for Deep Learning models
-
```
skoudoro, line 146: to keep

dipy/nn/utils.py
```diff
@@ -163,82 +161,64 @@ def transform_img(
         Considered points when calculating the transformed shape.
         \"corners\" will consider only corners of the image shape.
         \"all\" will consider all voxels. Might be needed when shearing is applied.
-
```
skoudoro, line 166: to keep

dipy/nn/utils.py
```diff
@@ -163,82 +161,64 @@ def transform_img(
...
-        target_voxsize = np.array(target_voxsize)
-
-    if not np.allclose(voxsize, target_voxsize):
-        image, resliced_affine = reslice(
-            image, affine, tuple(voxsize), tuple(target_voxsize)
-        )
-    else:
-        resliced_affine = None
-
-    init_shape = image.shape
```
skoudoro, line 190: I do not understand those change since you did it 3 weeks ago. 

is it just a wrong rebase, intentional ? Can you check the 2 `utils` files that you updated.

pjsjongsung: This is intentional. While fixing the bug in the function, I found it easier and simpler to remove the `reslice` function and just have voxel resizing as part of affine transform as well.

Regarding utils from segment, I added options to the `remove_holes_and_islands` function so that it would be able to accommodate some of the post processing steps used in synthseg.

PR #3676, opened 2026-01-06
RF: Allow reslice to automatically determines isotropic resolution

skoudoro:
This pull request enhances the `dipy_reslice` by adding support for automatic voxel size calculation. 
Tests has been added to validate these features.

During previous workshop, I realize that users with anisotropic data add hard time to make it isotropic and define a voxel size.


maharshi-gor:
I tested it on local, with different edge cases. Works for me on anisotropic data very nicely.

PR #3678, opened 2026-01-07
Build 3.14 nightly wheels

larsoner:
Moving MNE-Python to testing 3.14 there are no 3.14 dipy wheels in scientific-python-nightly-wheels. This should fix it I think! :crossed_fingers:

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -38,7 +38,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        cibw_python: ["cp310-*", "cp311-*", "cp312-*", "cp313-*"]
+        cibw_python: ["cp310-*", "cp311-*", "cp312-*", "cp313-*", "cp314-*"]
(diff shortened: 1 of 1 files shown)


skoudoro:
Yes, it should fix it. 

Thanks!

Let me merge and trigger it manually

larsoner:
Thanks for starting https://github.com/dipy/dipy/actions/runs/20787837330, I'll keep an eye on it! :eyes:

skoudoro:
triggered here : https://github.com/dipy/dipy/actions/runs/20787837330

so they should appear soon

larsoner:
Failed

https://github.com/dipy/dipy/actions/runs/20787837330/job/59702194831

Can I add a `on: pull_request` but that only runs if `[wheels]` is in the commit message in #3679? That would allow testing that it actually works without needing to merge and manually trigger every time

skoudoro:
Compilation and wheels seems ok,

only few tests seems to failed. We need to fix those tests for python 3.14 and numpy pre-release.

PR #3679, opened 2026-01-07
Build aarch64 natively

larsoner:
Watching https://github.com/dipy/dipy/actions/runs/20787837330/job/59702194950 I noticed aarch64 was taking a while... might as well build these natively instead. Should be much faster.

I also made some opinionated changes about how to simplify and unify some vars and steps now that the matrix is just building native on each image. Happy to roll these back if they're not desired.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -1,5 +1,8 @@
 name: Build and upload nightly wheels
 on:
+  pull_request:
+    branches:
+      - master
   workflow_dispatch:
     inputs:
       branch_or_tag:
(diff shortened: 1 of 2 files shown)


skoudoro:
Fine with this. Thanks !

Can you remove python 3.10 since we are at it. The next release will have python 3.11 as a minimum version

larsoner:
Okay macos fails:

https://github.com/dipy/dipy/actions/runs/20791985210/job/59716103590?pr=3679

I wasn't sure what all the `MACOS_COMPILER_ENV` gymnastics were for so I ensured they applied to all runs rather than just maybe some of them... seems like that's problematic. I can bump to 15.0 but I might try removing the `brew` stuff entirely first to see if it helps and the native / default compliers can be used?

larsoner:
> I wasn't sure what all the MACOS_COMPILER_ENV gymnastics were for

Okay it's because of the `brew install`... but this is problematic because it will always build+install for the current macOS version. So if we build the wheels on `macos-15`, it'll build `libomp` compat with 15.0+ which is problematic. I'll see if it works without these compilers, hopefully it does. In another project I recently hit this issue and adapting https://mac.r-project.org/openmp/ I was able to get it to work. I'll try it here, too

larsoner:
Bummer on macOS:
```
    Run-time dependency OpenMP found: NO  (tried system)
```
but it seems intentional on the part of the `meson` folks. I think at least for nightly it's worth disabling for now. But @skoudoro if you know how to get it working properly without pushing the `MACOSX_DEPLOYMENT_TARGET` so high then feel free to push commits...

*EDIT:*
xrefs:
- https://mac.r-project.org/openmp/ (what I wanted to use)
- https://github.com/mesonbuild/meson/issues/7435
- https://github.com/mesonbuild/meson/issues/13145

skoudoro:
ok, all good, Thanks @larsoner !!!

I use https://mac.r-project.org/openmp. Unfortunately, their website block Github Action for download, So I downloaded it manually and store it in our github dipy_data repo. 

it seems to works like a charm now. 

I will squash some commit  now then merge.

skoudoro:
ok, I messed it up, rebase/squash,rewrite history, 

sorry @larsoner ...

larsoner:
No problem! If you squash+merge you can have a `Co-authored by` line in there and we'll both get credit. But even if you don't do that it's okay by me!

skoudoro:
Yes, this is what I did the `Co-authored by`

Issue #3693, opened 2026-01-28
QTI issues - lack of MIN_POSITIVE_SIGNAL and issue with constrained solver and WLS is wrong anyway


samcoveney:
### Summary

When fitting QTI data, DiPy does not enforce ```MIN_POSITIVE_SIGNAL```, which I think it should since the fitting is linear.

Secondly, I get this warning when trying to fit with the constrained method ```fit_method=SDPdc```:

> UserWarning: Your problem has too many parameters for efficient DPP compilation. We suggest setting 'ignore_dpp = True'.

Anyone seen this before?

### Platform

Ubuntu

### DIPY version

 968e833bd545fe7a108bfd320ec67cb1c98c3b7c

### Python version

3.12.8

samcoveney:
Additionally, as for DKI in https://github.com/dipy/dipy/issues/3222, we almost certainly have the same issue for QTI

samcoveney:
And it seem https://github.com/dipy/dipy/blob/968e833bd545fe7a108bfd320ec67cb1c98c3b7c/dipy/reconst/qti.py#L587  shows that the weights for WLS fitting are the raw (observed) signals, rather than the fitted signal, which is inconsistent with our approach for DTI and DKI.

samcoveney:
And now I think I have found that the QTI code was wrong anyway.

If you look at https://github.com/dipy/dipy/blob/9438ca129e0ba5ef778e03fe1b813cd178ebf8e7/dipy/reconst/qti.py#L623-L627

you see that the weights `C` are defined as the signal (ignore the "observed vs predicted"  issue, that's not the current point).

In order to go from OLS to WLS, you effectively need to incorporate the weights into the design matrix and observation vector. So there may have been confusion over the weights definition. 

If you define the weights to be the signal, as in eq (24) of "Formal characterization and extension of the linearized diffusion tensor model" (Salvador, 2005), then you arrive at their equation (25)

$$
\beta = (X^T W^2 X)^{-1} (X^T W^2 Y)
$$

However, as far as I can see, in order to do this you would have to define `C` (the weights) to be signal squared.

This is why I would usually solve such a problem as:

```
fit_result, _ = ols_fit_tensor(
    design_matrix, data, return_lower_triangular=True
)
w = np.exp(fit_result @ design_matrix.T)

fit_result = np.einsum(
    "...ij,...j", np.linalg.pinv(design_matrix * w[..., None]), w * log_s
)
```


[...]

PR #3699, opened 2026-02-05
Qti robust

samcoveney:
Robust fitting (and robust and constrained fitting) for QTI, in a similar style to the implementation in DKI.

I also fixed some bugs with QTI WLS weights, see #3693

Tests updated - these show how to use the new functionality. The API for before should remain the same.

(edit: I'll try to tidy all the PEP issues later, but please tell me if documentation is wrong / insufficient)

### DIFF
--- a/dipy/reconst/qti.py
+++ b/dipy/reconst/qti.py
@@ -664,7 +664,7 @@ def _wls_fit(X, data_masked, *, weights=None, step=int(1e4), return_leverages=Fa
     return params_masked, extra
 
 
-def _sdpdc_fit(X, data_masked, cvxpy_solver, weights=None, return_leverages=False):
+def _sdpdc_fit(X, data_masked, cvxpy_solver, *, weights=None, return_leverages=False):
     r"""Estimate the model parameters using Semidefinite Programming (SDP),
     while enforcing positivity constraints on the D and C tensors (SDPdc).
 
@@ -677,8 +677,8 @@ def _sdpdc_fit(X, data_masked, cvxpy_solver, weights=None, return_leverages=Fals
         coefficients.
     data_masked : array (N, g)
         The measured signal, already masked.
-    cvxpy_solver: string, required
-        The name of the SDP solver to be used. Default: 'SCS'
+    cvxpy_solver: str
+        The name of the SDP solver to be used.
     weights : array (N, g), optional
         Weights to apply for fitting. These weights must correspond to the
         squared residuals such that $S = \sum_i w_i r_i^2$. If not provided,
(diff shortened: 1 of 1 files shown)


samcoveney:
Any chance of a quick look in here please?

dipy/reconst/dti.py
```diff
@@ -1357,6 +1359,9 @@ def wrapped_fit_tensor(
                         )
                     )
                 else:
+                    D, E = fit_tensor(
+                        design_matrix, data[i : i + step], *args, **kwargs
+                    )
```
skoudoro: I do not understand this call, since it is call the line just after and you do not use `D, E`

Can you explain ?

dipy/reconst/qti.py
```diff
@@ -719,14 +780,20 @@ def _sdpdc_fit(data, mask, X, cvxpy_solver):
...
+        extra = None
+
+    return params_masked, extra
 
 
 class QtiModel(ReconstModel):
     @warning_for_keywords()
-    def __init__(self, gtab, *, fit_method="WLS", cvxpy_solver="SCS"):
+    def __init__(
+        self, gtab, *args, fit_method="WLS", cvxpy_solver="SCS", **kwargs
```
skoudoro: `*args` is unclear here. Can we be more explicit and use keyword arguments and document them (`weights_method and num_iter`) ? or use `**kwargs`. in your test, you actually use kwargs.

Also, the error message still lists only 'OLS', 'WLS', 'SDPdc' but now RWLS is also valid. Please update

dipy/reconst/qti.py
```diff
@@ -608,44 +630,70 @@ def _wls_fit(data, mask, X, *, step=int(1e4)):
...
+    if return_leverages:
+        extra = {"leverages": leverages}
+    else:
+        extra = None
+
+    return params_masked, extra
+
+
+def _sdpdc_fit(
+    X, data_masked, cvxpy_solver, weights=None, return_leverages=True
```
skoudoro: Why `return_leverages=True` here when it is `False` for `_ols_fit/_wls_fit`

dipy/reconst/dti.py
samcoveney: Sorry about this, I must have copied the code from somewhere else, before copying it and editing the copy. I will delete this!

dipy/reconst/qti.py
samcoveney: I've added docstrings to match what's in `dki.DiffusionKurtosisModel` - I hope this will be okay for now.

Fixed the docstring to states that `RWLS` is valid too.

dipy/reconst/qti.py
samcoveney: Not sure why I did this, probably a mistake, have changed back to False

dipy/reconst/qti.py
skoudoro: Thanks for the update.

Sorry, I still have a issue with `*args`. I think we should remove it. `**kwgars` should be enough.

I might be missing something but why `*args` is so important?

skoudoro (changes requested):
Thank you for the update. 

Can you fix the code format also ? See below my last comments. then, I believe it is ready to go

samcoveney:
I don't think `*args` was important, I was copying other `DiPy` code for consistency.

On the code format, are you talking about the fussy code format / pre-commit stuff seems like I can never get my local pep8 to agree with github about?

samcoveney:
I have removed args, will look at this PR later tonight or tomorrow.

samcoveney:
Alright, fingers crossed the other tests pass!

skoudoro:
All good !

Thank you for this work @samcoveney 

merging

PR #3704, opened 2026-02-10
Fix get simplified backward transform

36000:
Typo in get simplified code in diffeomorphic mapping

### DIFF
--- a/dipy/align/tests/test_imwarp.py
+++ b/dipy/align/tests/test_imwarp.py
@@ -348,8 +348,8 @@ def test_diffeomorphic_map_simplification_3d():
     domain_sphere = vfu.create_sphere(
         domain_shape[0], domain_shape[1], domain_shape[2], radius
     )
-    expected = diff_map.transform(domain_sphere, interpolation="linear")
-    warped = simplified.transform(domain_sphere, interpolation="linear")
+    expected = diff_map.transform_inverse(domain_sphere, interpolation="linear")
+    warped = simplified.transform_inverse(domain_sphere, interpolation="linear")
 
     assert_array_almost_equal(warped, expected)
 


skoudoro (commented):
Good catch! thanks @36000!

I agree with copilot saying that it would be great to have a test catching that.

then we can merge

dipy/align/imwarp.py
```diff
@@ -1017,7 +1017,7 @@ def get_simplified_transform(self):
         )
 
         # Simplify the backward transform
-        C = self.codomain_world2grid
+        C = self.codomain_grid2world
```
arokem, line 1020: I agree that an added test would be great to have. I will also add that I think that the code below could be simplified a lot by not allocating all these temporary variables (`C`, `Rinv`, etc.). While you're here, maybe you can clean that up a bit?

36000: I like how readable the temporary variables are, it keeps lines shorter. Also, I think nothing new is allocated, its just a reassignment. So, I would vote to keep them.

dipy/align/imwarp.py
```diff
@@ -1017,7 +1017,7 @@ def get_simplified_transform(self):
...
         Pinv = self.prealign_inv
         Dinv = self.domain_world2grid
```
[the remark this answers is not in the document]
36000: @copilot open a new pull request to apply changes based on [this feedback](https://github.com/dipy/dipy/pull/3704#discussion_r2786770662)

36000:
@skoudoro
Thank you, I added that test of the backwards transform. 
I think this is ready to merge now?

dipy/align/tests/test_imwarp.py
```diff
@@ -344,6 +344,15 @@ def test_diffeomorphic_map_simplification_3d():
...
     assert_equal(simplified.disp_world2grid, None)
 
+    # Now do the same test for the backwards transform
+    domain_sphere = vfu.create_sphere(
+        domain_shape[0], domain_shape[1], domain_shape[2], radius
+    )
+    expected = diff_map.transform(domain_sphere, interpolation="linear")
+    warped = simplified.transform(domain_sphere, interpolation="linear")
+
+    assert_array_almost_equal(warped, expected)
```
skoudoro, line 354: Hi @36000,

The tests passed in both cases meaning `C = self.codomain_world2grid` or `C = self.codomain_grid2world`.

I was hoping that it will fail with the old version and pass with the update. For now, this test just bring more maintenance. 

Can you look into it ?

36000: Good catch I will modify the test to be more extensive

36000: After about 1-2 hours of spinning my wheels I realized it was a simple typo in my addition to the test. I have fixed it in the recent commit, and now the test fails with the old code and passes with the new!

skoudoro: Thank you for fixing that !

PR #3707, opened 2026-02-12
Adding Cholesky parameterization in DTI 

marianaabv:
Hello DIPY team! 
I am Mariana and I am working under supervision of @RafaelNH. I am currently exploring constrained fitting strategies for DTI/DKI/CTI. 
The first thing that I noticed is that we only had the possibility of using Cholesky decomposition in fwDTI (which was not even working properly).
So in this PR I am purposing a fix of Cholesky parameterization in fwDTI and adding it on DTI. At the moment this functionalities should be already working well for a fwDTI, DTI and DKI. The only thing that is missing is to create some unit tests. 
Please let me know if u have any comments or suggestions :)

### DIFF
--- a/dipy/reconst/dti.py
+++ b/dipy/reconst/dti.py
@@ -22,6 +22,32 @@ from dipy.reconst.weights_method import (
 from dipy.testing.decorators import warning_for_keywords
 from dipy.utils.parallel import paramap
 
+
+def lower_triangular_to_cholesky(tensor_elements):
+    """Performs Cholesky decomposition of the diffusion tensor
+    """
+    R0 = np.sqrt(tensor_elements[0])
+    R3 = tensor_elements[1] / R0
+    R1 = np.sqrt(tensor_elements[2] - R3**2)
+    R5 = tensor_elements[3] / R0
+    R4 = (tensor_elements[4] - R3 * R5) / R1
+    R2 = np.sqrt(tensor_elements[5] - R4**2 - R5**2)
+
+    return np.array([R0, R1, R2, R3, R4, R5])
+
+
+def cholesky_to_lower_triangular(R):
+    """Convert Cholesky decomposition elements to the diffusion tensor elements
+    """
+    Dxx = R[0] ** 2
+    Dxy = R[0] * R[3]
+    Dyy = R[1] ** 2 + R[3] ** 2
+    Dxz = R[0] * R[5]
+    Dyz = R[1] * R[4] + R[3] * R[5]
+    Dzz = R[2] ** 2 + R[4] ** 2 + R[5] ** 2
+    return np.array([Dxx, Dxy, Dyy, Dxz, Dyz, Dzz])
+
+
 MIN_POSITIVE_SIGNAL = 0.0001
 
 ols_resort_msg = "Resorted to OLS solution in some voxels"
@@ -108,7 +134,8 @@ def geodesic_anisotropy(evals, *, axis=-1):
     evals : array-like
         Eigenvalues of a diffusion tensor.
     axis : int, optional
-        Axis of `evals` which contains 3 eigenvalues.
+        Axis of `
+        ` which contains 3 eigenvalues.
 
     Returns
     -------
@@ -1624,7 +1651,7 @@ def _ols_fit_matrix(design_matrix):
 class _NllsHelper:
     r"""Class with member functions to return nlls error and derivative."""
 
-    def err_func(self, tensor, design_matrix, data, weights=None):
+    def err_func(self, tensor, design_matrix, data, weights=None, cholesky=False):
         r"""
         Error function for the non-linear least-squares fit of the tensor.
 
@@ -1644,6 +1671,15 @@ class _NllsHelper:
             squared residuals such that $S = \sum_i w_i r_i^2$.
 
         """
+        #Cholesky decomposition if requested ---------------------------
+        if cholesky:
+            r_params = tensor[:6]
+            s0_param = tensor[6:]
+            d_params = cholesky_to_lower_triangular(r_params)
+            tensor = np.concatenate((d_params, s0_param))
+        
+        #----------------------------------------------------------------
+        
         # This is the predicted signal given the params:
         y = np.exp(np.dot(design_matrix, tensor))
         self.y = y  # cache the results
@@ -1762,6 +1798,7 @@ def nlls_fit_tensor(
     return_lower_triangular=False,
     return_leverages=False,
     init_params=None,
+    cholesky=False
 ):
     r"""
     Fit the cumulant expansion params (e.g. DTI, DKI) using non-linear
@@ -1815,14 +1852,16 @@ def nlls_fit_tensor(
     # Detect if number of parameters corresponds to dti
     dti = npa == 12
 
-    # Flatten for the iteration over voxels:
+    # Flatten data for the iteration over voxels:
     flat_data = data.reshape((-1, data.shape[-1]))
+    
+    # Flatten weights for the iteration over voxels:
     weights = weights.reshape((-1, weights.shape[-1])) if weights is not None else None
     if weights is not None:
         assert weights.shape == flat_data.shape, "Weights shape mismatch"
 
+    # Use the OLS method parameters as the starting point for nlls
     if init_params is None:
-        # Use the OLS method parameters as the starting point for nlls
         D, extra = ols_fit_tensor(
             design_matrix,
             flat_data,
@@ -1854,11 +1893,28 @@ def nlls_fit_tensor(
 
     if return_S0_hat:
         model_S0 = np.empty((flat_data.shape[0], 1))
+
+    # Start processing voxels
     for vox in range(flat_data.shape[0]):
+
+        # Check if voxel only contains zeros 
         if np.all(flat_data[vox] == 0):
             raise ValueError("The data in this voxel contains only zeros")
 
         start_params = ols_params[vox]
+
+        if cholesky:
+            tensor_ols = start_params[:6]
+            s0_param = start_params[6:]
+
+            # remove possible zero eigen-value in starting pars for cholesky
+            evals, evecs = decompose_tensor(from_lower_triangular(tensor_ols))
+            evals = np.maximum(evals, 2e-7)
+
+            dt_clean = lower_triangular(vec_val_vect(evecs, evals))
+
+            r_params = lower_triangular_to_cholesky(dt_clean)
+            start_params = np.concatenate((r_params, s0_param))
 
         weights_vox = weights[vox] if weights is not None else None
 
@@ -1867,9 +1923,14 @@ def nlls_fit_tensor(
             this_param, status = opt.leastsq(
                 err_func,
                 start_params,
-                args=(design_matrix, flat_data[vox], weights_vox),
-                Dfun=jac_func,
+                args=(design_matrix, flat_data[vox], weights_vox, cholesky),
+                Dfun=jac_func if not cholesky else None,
             )
+
+            if cholesky:
+                r_result = this_param[:6]
+                d_result = cholesky_to_lower_triangular(r_result)
+                this_param = np.concatenate((d_result, this_param[6:]))
 
             flat_params[vox] = this_param
 
@@ -1878,6 +1939,7 @@ def nlls_fit_tensor(
                 from_lower_triangular(this_param[:6]),
                 min_diffusivity=tol / -design_matrix.min(),
             )
+
             params[vox, :3] = evals
             params[vox, 3:12] = evecs.ravel()
 
--- a/dipy/reconst/dti.py
+++ b/dipy/reconst/dti.py
@@ -1673,10 +1673,10 @@ class _NllsHelper:
         """
 
         if cholesky:
-            r_params = tensor[:6]
-            s0_param = tensor[6:]
+            r_params = tensor[:6] 
+            neg_log_s0 = tensor[6:]
             d_params = cholesky_to_lower_triangular(r_params)
-            tensor = np.concatenate((d_params, s0_param))
+            tensor = np.concatenate((d_params, neg_log_s0))
 
         # This is the predicted signal given the params:
         y = np.exp(np.dot(design_matrix, tensor))
@@ -1903,7 +1903,7 @@ def nlls_fit_tensor(
 
         if cholesky:
             tensor_ols = start_params[:6]
-            s0_param = start_params[6:]
+            neg_log_s0 = start_params[6:]
 
             # remove possible zero eigen-value in starting pars for cholesky
             evals, evecs = decompose_tensor(from_lower_triangular(tensor_ols))
@@ -1912,7 +1912,7 @@ def nlls_fit_tensor(
             dt_clean = lower_triangular(vec_val_vect(evecs, evals))
 
             r_params = lower_triangular_to_cholesky(dt_clean)
-            start_params = np.concatenate((r_params, s0_param))
+            start_params = np.concatenate((r_params, neg_log_s0))
 
         weights_vox = weights[vox] if weights is not None else None
 
--- a/dipy/reconst/dti.py
+++ b/dipy/reconst/dti.py
@@ -92,8 +92,8 @@ def fractional_anisotropy(evals, *, axis=-1):
     ----------
     evals : array-like
         Eigenvalues of a diffusion tensor.
-    axis : int, optional
-        Axis of `evals` which contains 3 eigenvalues.
+        axis : int, optional
+        Axis of evals which contains 3 eigenvalues.
 
     Returns
     -------
--- a/dipy/reconst/dti.py
+++ b/dipy/reconst/dti.py
@@ -1889,13 +1889,21 @@ def nlls_fit_tensor(
     err_func = nlls.err_func
     jac_func = nlls.jacobian_func if jac else None
 
+    if cholesky and jac:
+        warnings.warn(
+            "The analytical Jacobian is not currently implemented"
+            " for Cholesky parameterization. ",
+            UserWarning,
+            stacklevel=2,
+        )
+
     if return_S0_hat:
         model_S0 = np.empty((flat_data.shape[0], 1))
 
     # Start processing voxels
     for vox in range(flat_data.shape[0]):
 
-        # Check if voxel only contains zeros 
+        # Check if voxel only contains zeros
         if np.all(flat_data[vox] == 0):
             raise ValueError("The data in this voxel contains only zeros")
 
--- a/dipy/reconst/dti.py
+++ b/dipy/reconst/dti.py
@@ -1671,15 +1671,13 @@ class _NllsHelper:
             squared residuals such that $S = \sum_i w_i r_i^2$.
 
         """
-        #Cholesky decomposition if requested ---------------------------
+
         if cholesky:
             r_params = tensor[:6]
             s0_param = tensor[6:]
             d_params = cholesky_to_lower_triangular(r_params)
             tensor = np.concatenate((d_params, s0_param))
-        
-        #----------------------------------------------------------------
-        
+
         # This is the predicted signal given the params:
         y = np.exp(np.dot(design_matrix, tensor))
         self.y = y  # cache the results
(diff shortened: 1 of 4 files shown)


dipy/reconst/dti.py
```diff
@@ -1644,6 +1671,15 @@ def err_func(self, tensor, design_matrix, data, weights=None):
             squared residuals such that $S = \sum_i w_i r_i^2$.
 
         """
+        #Cholesky decomposition if requested ---------------------------
+        if cholesky:
+            r_params = tensor[:6]
+            s0_param = tensor[6:]
```
samcoveney: Isn't this either log(s0) or -log(s0) in DiPy ? Probably should be named better if true

samcoveney:
The gradients calculated with the error function are not going to be correct, as is, if passing in the Cholesky form. This is because the gradient of the error function wrt the Cholesky form of the parameters is different. While I see that you've set the Jacobian function as none if Cholesky is used, it is a shame as explicit gradients usually speed up and improve fitting.

marianaabv:
> The gradients calculated with the error function are not going to be correct, as is, if passing in the Cholesky form. This is because the gradient of the error function wrt the Cholesky form of the parameters is different. While I see that you've set the Jacobian function as none if Cholesky is used, it is a shame as explicit gradients usually speed up and improve fitting.

Thank you for the feedback, Sam! I completely agree. The analytical Jacobian for the Cholesky parameterization requires applying the chain rule to the standard tensor Jacobian, which is not implemented yet. As in my project objectives we are not planning to derive the Jacobian for Cholesky, the safest approach for now is to disable the analytical Jacobian (Dfun=None) when cholesky=True, adding a warning when this combination of options is used.

dipy/reconst/dti.py
marianaabv: Good catch! I renamed the variable to neg_log_s0 to make the code mathematically clearer. Thanks :)

dipy/reconst/tests/test_dti.py
```diff
@@ -1224,3 +1230,32 @@ def test_quantize_evecs_parallel_engines():
...
+    dt_recovered = cholesky_to_lower_triangular(r_elements)
+    npt.assert_array_almost_equal(dt_gt, dt_recovered)
+
+
+def test_dti_nlls_cholesky_positivity():
+    """Test if NLLS with Cholesky parameterization enforces positive eigenvalues."""
+    evals_gt = np.array([0.0017, 0.0003, 0.0003])
+    evecs_gt = np.eye(3)
+
+    _, fbvals, fbvecs = get_fnames(name="small_64D")
```
RafaelNH: Can you load "small_25" instead?

dipy/reconst/tests/test_dti.py
```diff
@@ -1224,3 +1230,32 @@ def test_quantize_evecs_parallel_engines():
...
+    bvals, bvecs = read_bvals_bvecs(fbvals, fbvecs)
+    gtab = grad.gradient_table(bvals, bvecs=bvecs)
+    S_noisy = single_tensor(gtab, S0=100, evals=evals_gt, evecs=evecs_gt, snr=5)
```
RafaelNH: Don't make this simulation random (i.e. snr=None). 

My suggestion to test Choslesky constraints are the following:
1) Make sure that when Choslesky is used it gives the right solution on non-problematic cases.
2) Make sure that all diffusivities are larger than zero in cases that you give a corrupted diffusion tensor with a small negative eigenvalue. (No noise is needed)

dipy/reconst/tests/test_dti.py
```diff
@@ -1224,3 +1230,32 @@ def test_quantize_evecs_parallel_engines():
...
+    X = design_matrix(gtab)
+
+    params_nls, _ = nlls_fit_tensor(X, S_noisy, cholesky=True)
```
RafaelNH: Test this instead using:
dtim = dti.TensorModel(gtab, fit_method="NLS", cholesky=True)
dtif = dtim.fit(Spred)
evals_nls = dtif.evals

dipy/reconst/tests/test_dti.py
```diff
@@ -1224,3 +1230,32 @@ def test_quantize_evecs_parallel_engines():
...
+    evals_nls = params_nls[..., :3]
+ 
+    # Assert all eigenvalues are non-negative (within float precision)
+    npt.assert_(np.all(evals_nls >= -1e-8))
```
RafaelNH: Adapt this according to my previous comment about the two tests to perform for checking Choslesky decomposition.

dipy/reconst/dti.py
```diff
@@ -108,7 +134,8 @@ def geodesic_anisotropy(evals, *, axis=-1):
     evals : array-like
         Eigenvalues of a diffusion tensor.
     axis : int, optional
-        Axis of `evals` which contains 3 eigenvalues.
+        Axis of `
```
RafaelNH: What happened here? Please fix the missing "evals" and added space

RafaelNH (changes requested):
Thanks @marianaabv for the work done here. I left some initial comments.

dipy/reconst/dti.py
marianaabv: Thanks and done! All the previous suggestions have been addressed and the tests are updated as we discussed 🚀

skoudoro:
Hi @marianaabv,

what is the status of this PR? 

Thank you for your feedback

marianaabv:
> Hi @marianaabv,
> 
> what is the status of this PR?
> 
> Thank you for your feedback

Hi @skoudoro ,
I just made the final addition today, which was the Jacobian warning when the Cholesky parameterization is used. I have also removed the "WIP" tag from the title.
All requested changes are now complete, and the PR is ready for final review. 
Thank you for your feedback!

dipy/reconst/tests/test_dti.py
```diff
@@ -1224,3 +1230,32 @@ def test_quantize_evecs_parallel_engines():
...
+def test_cholesky_transformations():
+    """Test Cholesky decomposition and its inverse for DTI."""
+    dt_gt = np.array([0.0017, 0, 0.0003, 0, 0, 0.0003])
+    r_elements = lower_triangular_to_cholesky(dt_gt)
+    dt_recovered = cholesky_to_lower_triangular(r_elements)
+    npt.assert_array_almost_equal(dt_gt, dt_recovered)
+
+
+def test_dti_nlls_cholesky_positivity():
+    """Test if NLLS with Cholesky enforces positive eigenvalues."""
```
RafaelNH, line 1244: Reading the code - here you are testing if Cholesky decomposition gives the right ground truth parameters in non-problematic voxels. Change the comments to reflect that.

dipy/reconst/tests/test_dti.py
```diff
@@ -1224,3 +1230,32 @@ def test_quantize_evecs_parallel_engines():
...
+    evals_gt = np.array([0.0017, 0.0003, 0.0003])
+    evecs_gt = np.eye(3)
+
+    _, fbvals, fbvecs = get_fnames(name="small_25")
+    bvals, bvecs = read_bvals_bvecs(fbvals, fbvecs)
+    gtab = grad.gradient_table(bvals, bvecs=bvecs)
+    Spred = single_tensor(gtab, S0=100, evals=evals_gt, evecs=evecs_gt)
+    X = design_matrix(gtab)
+
+    params_nls, _ = nlls_fit_tensor(X, Spred, cholesky=True)
```
RafaelNH, line 1254: Instead of calling the nlls_fit_tensor, it is better to run the test by calling the TensorModel class object, i.e.
dtim=dti.TensorModel(gtab, fit_method="NLS", cholesky=True)
dtif = dtim.fit(Spred)

dipy/reconst/tests/test_dti.py
```diff
@@ -1224,3 +1230,32 @@ def test_quantize_evecs_parallel_engines():
...
+    evals_nls = params_nls[..., :3]
+
+    # Assert all eigenvalues are non-negative (within float precision)
+    npt.assert_(np.all(evals_nls >= -1e-8))
```
RafaelNH, line 1258: In this case, since you are checking if the code will retrieve the right ground truth values when Cholesky parameterization is used in non-problematic voxels, a better way to assert the output is:
npt.assert_array_almost_equal(dtif.evals, evals_gt)

dipy/reconst/tests/test_dti.py
```diff
@@ -1224,3 +1230,32 @@ def test_quantize_evecs_parallel_engines():
...
+
+    fa_nls = fractional_anisotropy(evals_nls)
+    npt.assert_(np.all((fa_nls >= 0) & (fa_nls <= 1)))
```
RafaelNH, line 1261: These lines of code are not necessary for this test.

RafaelNH: I was reviewing my previous comments, and, indeed - one test that you skipped is to test that all diffusivities are larger than zero in cases that you give a corrupted diffusion tensor with a small negative eigenvalue.

For that just run a simulation for:

evals_gt = np.array([0.0017, 0.0003, -0.0001])

Spred_corrupted = single_tensor(gtab, S0=100, evals=evals_gt, evecs=evecs_gt)

dtif = dtim.fit(Spred_corrupted)

npt.assert_(np.all(dtif.evals >= -1e-8))
npt.assert_((dtif.fa >= 0) & (dtif.fa <= 1))

dipy/reconst/dti.py
RafaelNH: Hi @marianaabv! You mentioned that this was resolved but I am still seeing a missing `evals` and a line break. Please fix this line of code to become: "Axis of `evals` which contains 3 eigenvalues."

dipy/reconst/dti.py
```diff
@@ -1856,32 +1889,63 @@ def nlls_fit_tensor(
     err_func = nlls.err_func
     jac_func = nlls.jacobian_func if jac else None
 
+    if cholesky and jac:
+        warnings.warn(
+            "The analytical Jacobian is not currently implemented"
```
RafaelNH, line 1894: Can we add a test in "test_dti.py" to check if this warning properly appears?

RafaelNH (changes requested):
Hi @marianaabv! Thanks for having addressed my first round of review. I am now posting some follow up comments.

skoudoro:
Hi @marianaabv,

Do you have an update on this ?

dipy/reconst/tests/test_dti.py
marianaabv: Updated. The docstring now explicitly states that this test verifies the retrieval of ground truth parameters in non-problematic voxels

dipy/reconst/tests/test_dti.py
marianaabv: Done. I've refactored the test to use the TensorModel class interface and the fit method as suggested, ensuring the test aligns with the public API

dipy/reconst/tests/test_dti.py
marianaabv: Done. I have replaced the inequality check, since this specific test now focuses on the accuracy of retrieving ground truth values in ideal conditions

dipy/reconst/tests/test_dti.py
marianaabv: Done. I have refactored this part to split the validation into two distinct tests as suggested in your previous comments: one for accuracy in ideal conditions and another for positivity in corrupted voxels

dipy/reconst/tests/test_dti.py
marianaabv: You're right, these lines were redundant for the accuracy test, so I've removed them.

Regarding the missing case: I have now implemented a second test, test_dti_nlls_cholesky_positivity, using the parameters you provided. This new test specifically validates that the Cholesky parameterization enforces non-negative eigenvalues and valid FA values even when the underlying signal corresponds to a wrong tensor

marianaabv:
> Hi @marianaabv,
> 
> Do you have an update on this ?

Hi @skoudoro,
Yes, I was finishing the final requested tests. Everything is now finalized and updated in the latest commits. Ready for review!

dipy/reconst/tests/test_dti.py
```diff
@@ -1224,3 +1230,48 @@ def test_quantize_evecs_parallel_engines():
...
+    bvals, bvecs = read_bvals_bvecs(fbvals, fbvecs)
+    gtab = grad.gradient_table(bvals, bvecs=bvecs)
+    
+    Spred_corrupted = single_tensor(gtab, S0=100, evals=evals_gt, evecs=evecs_gt)
+
+    dtim = dti.TensorModel(gtab, fit_method="NLS", cholesky=True)
+    dtif = dtim.fit(Spred_corrupted)
+
+    npt.assert_(np.all(dtif.evals >= -1e-8))
+    npt.assert_(np.all((dtif.fa >= 0) & (dtif.fa <= 1)))
```
RafaelNH: Remove tab on the last line

dipy/reconst/dti.py
```diff
@@ -66,8 +92,8 @@ def fractional_anisotropy(evals, *, axis=-1):
     ----------
     evals : array-like
         Eigenvalues of a diffusion tensor.
-    axis : int, optional
-        Axis of `evals` which contains 3 eigenvalues.
+        axis : int, optional
```
RafaelNH: Remove the extra added space.

dipy/reconst/dti.py
RafaelNH: This still has to be fixed

dipy/reconst/dti.py
RafaelNH: This is still missing

dipy/reconst/dti.py
```diff
@@ -1856,32 +1889,63 @@ def nlls_fit_tensor(
     err_func = nlls.err_func
     jac_func = nlls.jacobian_func if jac else None
 
+    if cholesky and jac:
+        warnings.warn(
+            "The analytical Jacobian is not currently implemented"
+            " for Cholesky parameterization. Running with jac=False.",
+            UserWarning,
+            stacklevel=2,
+        )
```
RafaelNH, line 1897: As you are saying that the code sets jac=False, add that line inside this if statement.

dipy/reconst/dti.py
```diff
@@ -1856,32 +1889,63 @@ def nlls_fit_tensor(
...
 
         try:
             # Do the optimization in this voxel:
             this_param, status = opt.leastsq(
                 err_func,
                 start_params,
-                args=(design_matrix, flat_data[vox], weights_vox),
-                Dfun=jac_func,
+                args=(design_matrix, flat_data[vox], weights_vox, cholesky),
+                Dfun=jac_func if not cholesky else None,
```
RafaelNH: If you add the line "jac_func=False" inside the if statement "if cholesky and jac" (see my comment above), you can remove "if not cholesky else None".

RafaelNH (changes requested):
Another round of comments - hopefully the last one at my side :)

dipy/reconst/tests/test_dti.py
marianaabv: Fixed, thanks for catching this

dipy/reconst/dti.py
marianaabv: Fixed

dipy/reconst/dti.py
marianaabv: I had fixed the wording but missed the indentation, which was causing numpydoc to render `axis` as part of the `evals` description instead of a separate parameter. Fixed now

dipy/reconst/dti.py
marianaabv: Added `test_cholesky_jac_warning` in test_dti.py to check the warning is raised when `jac=True` and `cholesky=True` are combined

dipy/reconst/dti.py
marianaabv: Done. `jac_func` is now set to `None` inside the `if cholesky and jac:` block, and I removed the redundant `if not cholesky else None` from the `Dfun` argument below

marianaabv (commented):
I made all the changes requested. It's ready for review

dipy/reconst/dti.py
RafaelNH: This problem have not been fixed yet. Please replace the lines 137-138 with "Axis of `evals` which contains 3 eigenvalues." in a single line.

RafaelNH (commented):
Only one minor aspect missing: in **dti.py** replace the lines **137-138** with "**Axis of evals which contains 3 eigenvalues**." in a single line.

marianaabv:
> Only one minor aspect missing: in **dti.py** replace the lines **137-138** with "**Axis of evals which contains 3 eigenvalues**." in a single line.

Done, thanks

RafaelNH (dismissed):
All good for me! Many thanks @marianaabv

RafaelNH:
This PR seem ready to be merged- does anyone have some final comments?

Atharva-Shah-2298 (approved):
LGTM

skoudoro:
I will have a last look this weekend and then merge on Monday if CIs ok and no issue.

Thanks form this work !

Atharva-Shah-2298:
Hello @marianaabv , Can you please take a look at the Failing CIs

marianaabv:
Hi @Atharva-Shah-2298, thanks for flagging this!

I’ve addressed the failing CI checks and force-pushed the updated branch. The formatting/lint issues are now fixed locally and all pre-commit checks pass

The workflows are currently awaiting maintainer approval to run again

dipy/reconst/dti.py
```diff
@@ -1632,8 +1656,10 @@ def _ols_fit_matrix(design_matrix):
 class _NllsHelper:
     r"""Class with member functions to return nlls error and derivative."""
 
-    # scipy.optimize.leastsq calls this with args=(design_matrix, data, weights)
-    def err_func(self, tensor, design_matrix, data, weights=None):  # pep3102: ignore
+    # scipy.optimize.leastsq passes these arguments positionally.
+    def err_func(
+        self, tensor, design_matrix, data, weights=None, cholesky=False
+    ):  # pep3102: ignore
```
Atharva-Shah-2298, line 1662: "#pep3102..." Not needed as a comment. 

Always use the precommit hook which will style your code according to the PEP specs for dipy.

dipy/reconst/dti.py
```diff
@@ -1864,32 +1900,63 @@ def nlls_fit_tensor(
...
+            r_params = lower_triangular_to_cholesky(dt_clean)
+            start_params = np.concatenate((r_params, neg_log_s0))
+
         weights_vox = weights[vox] if weights is not None else None
 
         try:
             # Do the optimization in this voxel:
             this_param, status = opt.leastsq(
                 err_func,
                 start_params,
```
Atharva-Shah-2298, line 1949: It still says TypeError: _NllsHelper.jacobian_func() takes from 4 to 5 positional arguments but 6 were given  "Resorted to OLS" when jac is true. Should be a better way of handling if jacobian is not implemented right? Its not an error but just thinking..

Atharva-Shah-2298 (commented):
Because of jac=True, 2 of the tests are failing and also please rebase on top of current master and push so that CI s absorb the latest changes with the new CI defs. Also one of the CI which was correct before if I remember correctly...was that cholesky=True returned correct values on poor decomposition but it is wrong in this pass....When a fit fails with cholesky=True

Also it would be nice if we had DKI test with cholesky and testing nlls accuracy with some noise as without noise, I dont think it moved the estimates as OLS was already exact

dipy/reconst/dti.py
marianaabv: I kept this marker because these methods are SciPy least-squares callbacks: `scipy.optimize.leastsq` passes the callback arguments positionally. Making the default arguments keyword-only would therefore break the callback interface.

The current DIPY `enforce-pep3102-defaults` pre-commit hook explicitly requires either keyword-only arguments or a `# pep3102: ignore` exemption for positional third-party callbacks. The hook passes with this explicit exemption.

dipy/reconst/dti.py
marianaabv: Thanks! 

Addressed. When `cholesky=True` and `jac=True`, the fit now emits a warning and explicitly switches to `jac=False` before calling `scipy.optimize.leastsq`. This prevents the analytical Jacobian from receiving the incompatible Cholesky argument and avoids the resulting `TypeError`/OLS fallback.

The analytical Jacobian remains enabled for the standard parameterization. I also added coverage for the warning, noisy DKI Cholesky fitting, and the corrected Cholesky fallback. After rebasing on the current master, all pre-commit hooks pass and the seven Cholesky-focused tests pass.

PR #3746, opened 2026-02-16
TST: Replace legacy nose-style `yield assert_raises` with `pytest.raises`

### LINKED ISSUE #3745: Replace legacy nose-style `yield assert_raises` with `pytest.raises`
### Description

`dipy/reconst/tests/test_reco_utils.py` contains a nose-style `yield assert_raises(...)` pattern that is incompatible with modern pytest. The tests were wrapped in a docstring (effectively dead code) due to historical segfault issues with NumPy circa 2010.

### Current behavior

4 error-condition tests in `test_argmax_from_countarrs` are disabled — hidden inside a triple-quoted docstring using the legacy `yield assert_raises` pattern that nose used to support.

### Expected behavior

Tests should use `pytest.raises()` context managers and be actively running.

### Scope

- **1 file**: `dipy/reconst/tests/test_reco_utils.py`
- **4 test cases** converted and re-enabled

### References

- [pytest documentation — pytest.raises](https://docs.pytest.org/en/stable/reference/reference.html#pytest-raises)

satyambitm:
Fixes #3745 

### Summary

Converts the nose-style `yield assert_raises(...)` pattern in `reconst/tests/test_reco_utils.py` to modern `pytest.raises` context managers.

The 4 error-condition tests were previously disabled inside a docstring due to segfault issues with NumPy SVN circa 2010. This PR re-enables them using the standard pytest pattern.

### Changes

- **1 file** modified (14 insertions, 30 deletions)
- Replaced 4 `yield assert_raises(Exception, func, *args)` → `with pytest.raises(Exception): func(*args)`
- Added `import pytest`
- Removed docstring wrapper and outdated 2010-era comment

### Checklist

- [x] No API or behavior changes
- [x] Follows DIPY commit message convention (`TST:` prefix)
- [x] File compiles without errors

### DIFF
--- a/dipy/reconst/tests/test_reco_utils.py
+++ b/dipy/reconst/tests/test_reco_utils.py
@@ -1,7 +1,7 @@
 """Testing reconstruction utilities."""
 
 import numpy as np
-from numpy.testing import assert_array_equal, assert_equal
+from numpy.testing import assert_array_equal, assert_equal, assert_raises
 
 from dipy.reconst.recspeed import adj_to_countarrs, argmax_from_countarrs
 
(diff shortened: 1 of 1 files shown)


skoudoro:
Can you explain this choice knowing that we use numpy testing ? Thank you for your feedback

satyambitm:
Good point, thank you! I chose `pytest.raises` because the original code used the nose-specific `yield assert_raises(...)` generator pattern (which doesn't work in pytest), and `pytest.raises` is the standard modern replacement for exception testing. However, you're right that DIPY primarily uses `numpy.testing.assert_raises` throughout the test suite for consistency.

I'll update the PR to use `numpy.testing.assert_raises` as a context manager instead:

```python
from numpy.testing import assert_raises

with assert_raises(ValueError):
    argmax_from_countarrs(vals, vertinds, adj_counts, adj_inds_raw)
```

This keeps the tests consistent with the rest of the codebase while still re-enabling the previously dead test code. I'll push the update shortly.

PR #3752, opened 2026-02-18
CI: Update secrets management

mrinalchaturvedi27:
I discovered 10+ issues when I ran **actionlint** on the workflows in which I found 3 of them to be critical.

```
.github/workflows/nightly.yml:128:52: property "cibw_manylinux" is not defined in object type {cibw_arch: string; cibw_python: string; compiler_env: string; os: string} [expression]
    |
128 |             name: wheels-${{ env.PY_VERSION }}-${{ matrix.cibw_manylinux }}-${{ matrix.cibw_arch }} 
```
---

```
.github/workflows/test_template.yml:150:20: property "codecov_token" is not defined in object type {actions_runner_debug: string; actions_step_debug: string; codecov-token: string; github_token: string} [expression]
    |
150 |         token: ${{ secrets.CODECOV_TOKEN }} 
```
---

```
.github/workflows/nightly.yml:208:11: if: condition "${{ always() }} && github.repository_owner == 'dipy' && github.ref == 'refs/heads/master'" is always evaluated to true because extra characters are around ${{ }} [if-cond]
    |
208 |       if: ${{ always() }} && github.repository_owner == 'dipy' && github.ref == 'refs/heads/master' 
```
---
I have fixed these critical issues.

### You can reproduce these bugs by using the command: ###
**`actionlint .github/workflows/*.yml`**

I referred this [for the third issue mentioned above](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idif).

**Additional Information:**
https://github.com/rhysd/actionlint

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -205,7 +205,7 @@ jobs:
         contents: write # for softprops/action-gh-release to create GitHub release
       name: Upload to Anaconda
       needs: [build_linux_wheels, build_osx_wheels, build_windows_wheels]
-      if: ${{ always() }} && github.repository_owner == 'dipy' && github.ref == 'refs/heads/master'
+      if: always() && github.repository_owner == 'dipy' && github.ref == 'refs/heads/master'
       runs-on: ubuntu-latest
       timeout-minutes: 10
       steps:


.github/workflows/nightly.yml
```diff
@@ -125,7 +125,7 @@ jobs:
         run: echo "PY_VERSION=$(echo ${{ matrix.cibw_python }} | cut -d- -f1)" >> $GITHUB_ENV
       - uses: actions/upload-artifact@v6
         with:
-            name: wheels-${{ env.PY_VERSION }}-${{ matrix.cibw_manylinux }}-${{ matrix.cibw_arch }}
+            name: wheels-${{ env.PY_VERSION }}-${{ matrix.cibw_arch }}
```
skoudoro: can you remove this change. someone work already on it in another PR. Thanks

.github/workflows/nightly.yml
```diff
@@ -205,7 +205,7 @@ jobs:
         contents: write # for softprops/action-gh-release to create GitHub release
       name: Upload to Anaconda
       needs: [build_linux_wheels, build_osx_wheels, build_windows_wheels]
-      if: ${{ always() }} && github.repository_owner == 'dipy' && github.ref == 'refs/heads/master'
+      if:  always() && github.repository_owner == 'dipy' && github.ref == 'refs/heads/master'
```
skoudoro: make sense

.github/workflows/test_template.yml
```diff
@@ -150,7 +147,7 @@ jobs:
       uses: codecov/codecov-action@v5
       if: ${{ fromJSON(env.COVERAGE) }}
       with:
-        token: ${{ secrets.CODECOV_TOKEN }}
+        token: ${{ secrets.codecov-token }}
```
skoudoro: this one does not make sense. since we did not write the env vars like this in settings. Can you explain ?

mrinalchaturvedi27: Hi @skoudoro. Thanks for the review.
This is documented in the GitHub Actions docs under ["Using inputs and secrets in a reusable workflow"](https://docs.github.com/en/actions/sharing-automations/reusing-workflows#using-inputs-and-secrets-in-a-reusable-workflow). 
 We can only reference secrets by the name declared in the on.workflow_call.secrets. So inside test_template.yml, I referenced it as secrets.codecov-token.
 ```
workflow_call:
  secrets:
      codecov-token:
        description: "codecov token to send the report"
        required: false
```

File: .github/workflows/nightly.yml (hunk 1 above)
Comment (mrinalchaturvedi27) [reply .github/workflows/nightly.yml]: Sure!

File: .github/workflows/test_template.yml (hunk 3 above)
Comment (skoudoro) [reply .github/workflows/test_template.yml]:
```
workflow_call:
 secrets:
     codecov-token:
       description: "codecov token to send the report"
       required: false
```

this is not used. we are referring to the secrets from the general settings. if you change it, it will stop to work. `workflow_call` is use only for manual triggering.  and the code above is use only for manual triggering which we do only for debug.
we could add an `&&`to handle both like if one is empty, try the other one. that the best case scenario, but currently, this change is wrong.

mrinalchaturvedi27: Thanks for the clarification!
I'll implement || as the fallback  `${{ secrets.CODECOV_TOKEN || inputs.codecov-token }} `
which tries the repo secret first and falls back to the input if it's empty. Let me know if that matches what you had in mind.

skoudoro: yes, you can go ahead. but input secret first then repo secret

skoudoro: also, make sure to rebase this branch and resolve conflict. Thanks

mrinalchaturvedi27: Done!

PR #3781, opened 2026-02-21
Fix epsilon docstring mismatch

### LINKED ISSUE #3764: epsilon docstring mismatch
While going through Dipy Docs, I found that in parameter description of [remove_invalid_streamlines](https://docs.dipy.org/dev/reference/dipy.io.html#:~:text=verification.%20Default%20is-,1e%2D6) the default mentioned is **1e-6** but the function signature has **epsilon=1e-3**. Could you please clarify which value is correct? If this is a documentation inconsistency, I would be happy to submit a fix.

[Source code doc](https://docs.dipy.org/dev/_modules/dipy/io/stateful_tractogram.html#StatefulTractogram.remove_invalid_streamlines)

mrinalchaturvedi27:
This PR fixes the inconsistency in between the documented default value of `epsilon` in **remove_invalid_streamlines** and the actual function signature. 
Fixes #3764.

### DIFF
--- a/dipy/io/stateful_tractogram.py
+++ b/dipy/io/stateful_tractogram.py
@@ -620,7 +620,6 @@ class StatefulTractogram:
         ----------
         epsilon : float, optional
             Epsilon value for the bounding box verification.
-            Default is 1e-6.
 
         Returns
         -------


skoudoro:
Thanks @mrinalchaturvedi27 , merging

PR #3804, opened 2026-03-02
adding dipy.nn.hub, pre-trained model hub for dMRI

### LINKED ISSUE #3787: [FEATURE] `dipy.nn.hub` pre trained model hub for `dMRI` deep learning models
### Summary

DIPY has no unified way to discover, download, and run community trained model checkpoints. Researchers wanting to use models like `TractCloud`, `TractSeg`, or `BundleSeg` must clone separate repos, manage separate environments, and manually handle data format conversion. This proposal adds `dipy.nn.hub`, a lightweight registry + download layer that makes pre trained model inference a one liner, similar to what `PyTorch` Hub did for the `PyTorch` ecosystem.

### DIPY version

dipy 1.10.0

### Additional information

I have been looking at the `dipy.nn` module and noticed there's no way to just load and run a pre-trained model without going and cloning a completely separate repository. if I want to use `TractCloud` for bundle segmentation, I have to clone their repo, set up a separate environment, figure out their data format, and then manually map the output back to DIPY types. It is a lot of friction for something that should be straightforward
i would like to propose a `dipy.nn.hub` submodule that works similarly to `PyTorch` Hub, you point it at a model name, it downloads the weights once, caches them locally, and returns something you can call `.predict()` on with normal DIPY inputs like `ArraySequence` and `NIfTI` images
the idea is to keep a `models_registry.yaml` file versioned right inside the DIPY repo. Adding a new model would just mean opening a PR to update that file. Each model would have a thin adapter class responsible for converting DIPY's native data types to whatever the model expects internally, and converting the output back. The hub itself wouldn't depend on `PyTorch` or `TensorFlow` that would be the adapter's concern, so users only need whatever framework that specific model requires
i am happy to implement this myself, starting with the core download/cache infrastructure, a `BaseHubAdapter` class, and a full `TractCloud` adapter as the reference implementation plus tests
i have already reviewed `evac.py` and `dipy.nn.__init__` and plan to follow the same `optional_package` + weight fetching patterns already used in the codebase.
would love to get early feedback on the registry schema design before going too far

aman-coder03:
## Description

Adds `dipy.nn.hub`, a lightweight registry and adapter layer that lets researchers load and run pre-trained dMRI deep learning models in one line.
```python
model = hub.load('TractCloud', version='1.0')
bundles = model.predict(streamlines, affine=affine)
```
This is opened to get early feedback on the architecture before full implementation. No model weights or full adapters yet

## Motivation and Context
Closes #3787
Researchers wanting to use models like TractCloud, TractSeg, or BundleSeg must currently clone separate repos, manage separate environments, and manually handle data format conversion back to DIPY types. This adds unnecessary friction for something that should be straightforward.`dipy.nn.hub` solves this the same way PyTorch Hub solved it for the PyTorch ecosystem, a central registry, automatic weight download and caching, and a consistent `.predict()` interface using DIPY-native types like `ArraySequence` and NIfTI images.

The implementation follows existing DIPY patterns exactly:
- `optional_package()` for soft dependencies
- `_make_fetcher()` in `dipy.data.fetcher` for weight downloads
- Same model class structure as `evac.py` and `synthseg.py`

## How Has This Been Tested?

Not tested yet, this is currently for architecture feedback only, Before marking ready for review, the following will be added:
- Unit tests with mocked HTTP for the download/cache pipeline
- Adapter I/O tests for the `TractCloud` reference implementation
- Integration test on a small `tractogram`

## Questions for Reviewers

1. Is `models_registry.yaml` the right approach or do you prefer a Python dict inside the module?
2. Should hub weights go through the existing `_make_fetcher()` system in `fetcher.py` or have a separate lightweight downloader?
3. Any naming preferences maybe `hub.load()` vs `hub.get_model()`?
@skoudoro 

## Checklist

## Type of Change

- [ ] Documentation update

### DIFF
--- a/dipy/nn/hub/__init__.py
+++ b/dipy/nn/hub/__init__.py
@@ -46,7 +46,7 @@ def load(name, version=None, use_cuda=False):
     Parameters
     ----------
     name : str
-        Name of the model e.g. 'TractCloud'.
+        Name of the model e.g. 'SynthSeg'.
     version : str, optional
         Version string. Uses registry default if None.
     use_cuda : bool, optional
@@ -73,20 +73,6 @@ def load(name, version=None, use_cuda=False):
 
     logger.info(f"Loading model '{name}' from DIPY hub...")
 
-    # Weights fetching — plugs into existing dipy.data fetcher system
-    from dipy.data import fetcher
-
-    fetch_fn_name = f"fetch_{name.lower()}_torch_weights"
-
-    if not hasattr(fetcher, fetch_fn_name):
-        raise NotImplementedError(
-            f"No fetcher found for '{name}'. "
-            f"Expected 'dipy.data.fetcher.{fetch_fn_name}' to exist."
-        )
-
-    fetch_fn = getattr(fetcher, fetch_fn_name)
-    weights_path = fetch_fn()
-
     # Load adapter
     adapter_path = f"dipy.nn.hub.adapters.{name.lower()}"
     adapter_class_name = f"{name}Adapter"
@@ -100,4 +86,4 @@ def load(name, version=None, use_cuda=False):
             f"Expected '{adapter_path}.{adapter_class_name}'."
         ) from e
 
-    return adapter_class(weights_path, use_cuda=use_cuda)
+    return adapter_class(use_cuda=use_cuda)
--- a/dipy/nn/hub/adapters/synthseg.py
+++ b/dipy/nn/hub/adapters/synthseg.py
@@ -19,13 +19,12 @@ class SynthSegAdapter(BaseHubAdapter):
         Whether to use GPU if available.
     """
 
-    def __init__(self, weights_path, use_cuda=False):
+    def __init__(self, weights_path=None, use_cuda=False):
         if not have_torch:
             raise ImportError("PyTorch is required. Install with: pip install torch")
         from dipy.nn.torch.synthseg import SynthSeg
 
         self._model = SynthSeg(use_cuda=use_cuda)
-        self._model.load_model_weights(weights_path)
 
     def init_model(self):
         """Not used directly — SynthSeg initializes its own model."""


skoudoro:
Thanks @aman-coder03  for suggesting this idea.

@pjsjongsung and @arokem, Opinions ?

pjsjongsung:
Sounds reasonable. Would love to see how this can be implemented

aman-coder03:
thank you @skoudoro and @pjsjongsung for the feedback
happy to proceed with the full implementation. Before i do, a few quick architecture questions to make sure i am aligned with the codebase...

1. **Registry as Python dict** i have already moved away from `models_registry.yaml` to a plain Python dict inside `__init__.py` to avoid adding a `yaml` dependency. Does that approach work for you?

2. **Weight fetching** my plan is to register new model weights via `_make_fetcher()` in `dipy/data/fetcher.py`, exactly like `fetch_evac_torch_weights`. Does that work, or would you prefer a separate download mechanism for hub models?

3. **First model** i would like to use `TractCloud` as the reference adapter implementation. Are there any licensing or hosting concerns with distributing their weights via figshare?

once I have answers to these i'll move forward

pjsjongsung:
I can answer 2: yes what we do for other nn models would work. and 3: We do not have a list of licenses of different models, so if you are sourcing the weights from their database (e.g. github) you would have to check the license mentioned there.

aman-coder03:
after reviewing TractCloud's inference pipeline in detail, it has heavy dependencies(`whitematteranalysis`,`pytorch3d`) and uses VTK PolyData as input format, which makes it difficult to integrate cleanly as the first reference adapter

i would like to propose using a simpler model as the first reference implementation to establish the hub pattern, then add TractCloud as a follow up once the infrastructure is in place

@pjsjongsung would it be possible to use one of the existing DIPY models(e.g. a future tractography model) as the first adapter, or do you have a suggestion for a simpler dMRI model with minimal dependencies?

pjsjongsung:
Ok I am a little lost here now. If we have to design the adapter in the end, and all we can fetch are weights, could you explain in more detail how this change is different from the current approach? Are you suggesting it as a more organizing the models approach, so that for example instead of loading with `dipy.nn.evac` and `dipy.nn.deepn4` we would be loading `dipy.nn.hub` and `hub.load('EVAC+')` and `hub.load('DeepN4')`?

aman-coder03:
yes @pjsjongsung the main value is discoverability and consistent interface
instead of knowing that `EVACPlus` lives in `dipy.nn.evac` and `DeepN4` in `dipy.nn.deepn4`, users can just do..
```
hub.list_models()  #see everything available
hub.load('EVACPlus')   #consistent interface for all models
```
secondary goal is making it easier to add external models like TractCloud in the future...they would just appear in `list_models()` alongside existing DIPY models

happy to start by wrapping the existing DIPY models first(`EVACPlus`,`DeepN4`,`SynthSeg`,`HistoResDNN`)...that would 
demonstrate the value immediately. Does that direction work?

pjsjongsung:
Ok, but in that case do we need `hub`? Could we just load with `dipy.nn.load_model` and so on? Since almost everything else that is loaded through nn would be inside hub anyways.

aman-coder03:
that's a fair point! the main reason for keeping hub as a separate namespace is future extensibility, when external models like `TractCloud` are added, they wouldn't naturally live in `dipy.nn.torch` since their architecture code isn't bundled with DIPY. hub would act as the entry point for those external weights only models.

pjsjongsung:
We would have to think about that as well. We write from scratch to reduce the dependencies. If the current hub structure adds too much dependencies when adding external models, it is not so desirable.

aman-coder03:
to keep dependencies minimal, the hub would not bundle any external model code, it would only provide the registry, the consistent `predict()` interface, and fetch the weights. External model's architecture code would be an optional install, similar to how `dipy` handles torch and `tensorflow` today with `optional_package()`

so adding `TractCloud` would look like...
```
    pip install dipy[tractcloud]  # installs only what's needed
```
hub adapter would guard it with `optional_package()` so users without it installed get a clear error message, not a hard import failure
this way the hub itself adds zero new hard dependencies, does that address the concern @pjsjongsung?

skoudoro:
more code = more maintenance = potentially more bug.

less is more 😄

pjsjongsung:
It will add too much optional dependencies. For example, users can install all dependencies to maximize what they can do with DIPY, and having a hub designed like this would make them download all the packages available. Also in requirements, we keep it minimal. We do not want to add `model_name.txt` everytime we create a new model.

aman-coder03:
that's a valid concern. Given that the hub adds maintenance overhead and potentially encourages dependency bloat, I think the simpler approach makes more sense, adding `list_models()` and `load_model()` 
directly to `dipy.nn` without a separate `hub` submodule

this would give users discoverability and consistent interface without adding new structure/dependencies
would you like me to pivot the PR in that direction instead?

pjsjongsung:
Sounds reasonable to me. Thank you

skoudoro:
Ok, After more thinking, and talking to other maintainer,  I do not see the value of this PR. I do not think we will integrate stuff like bundleSeg of TractCloud. It does not seems to be the direction of the library.

Thank you for this proposal and this work @aman-coder03. We appreciate it. 

Closing.

Issue #3193, opened 2024-04-17
Improving MPPCA performance


skoudoro:
When you run the tutorials (dki tutorial or denoising tutorial), `mppca` is very slow.

Even if we have better denoising algorithm, I think we should look into it and improve the performance of this algorithm.

samcoveney:
It is very slow. Please let me know if I can help here.

arokem:
+1 for cython. At least we could try to use prange to parallelize across iterations in those loops.

skoudoro:
see here last interesting comment concerning the cythonize version of mppca: https://github.com/dipy/dipy/pull/1934#issuecomment-542451401

samcoveney:
Have taken a look at #1934 and left a comment

Wasif-Shahzad:
Hello, all. I'd like to work on this but I can't understand the issue in depth. Is it just converting the `mppca` function in `localpca.py` to cpython? If yes, then what is the standard method followed here at dipy for converting python to cpython? Any help would be appreciated.

NekoAsakura:
Hi @skoudoro,

Hope you had a relaxing spring break.

While investigating compute-bound algorithms in DIPY for GPU acceleration profiling, I identified a significant performance regression where MPPCA experiences severe multi-core slowdown on certain chip.

**Setting A (i9-13900K 24C/32T, numpy 2.3.5, scipy 1.17.1, OpenBLAS 0.3.28, Linux 6.17 bare metal)**:
Running CFIN multi-b dataset ($96 \times 96 \times 19 \times 67$, $b \le 400$) with patch_radius=2, the 8-thread MPPCA runtime was ~296s, while the single-thread runtime was ~51s — a 5.8x degradation from enabling multi-threading.

To isolate the cause, I wrote a minimal reproducer that mimics the genpca() inner loop pattern (50K iterations of `scipy.linalg.eigh` interleaved with numpy array allocations). The results were striking:

```
TEST CASE                               | 1 THREAD | 8 THREADS | SLOWDOWN
----------------------------------------|----------|-----------|----------
scipy.eigh loop (no alloc)              |   12.0s  |    12.2s  |    1.0x
scipy.eigh + alloc (genpca pattern)     |   17.2s  |   364.8s  |   21.2x
```

The slowdown only appears when `scipy.linalg.eigh` is combined with interleaved `numpy` allocations — the exact access pattern in genpca(), and swapping to `numpy.linalg.eigh` eliminates it entirely.

[...]

PR #3806, opened 2026-03-03
Cython implementation for the Python nested triple loop on localpca 'eig' path.

TomasGuija:
## Description

The PR introduces a Cython implementation of the default 'eig' path in genpca ([dipy/dipy/denoise/localpca.py](https://github.com/dipy/dipy/blob/master/dipy/denoise/localpca.py)), replacing the Python triple loop. The 'svd' path remains as it was and still uses the existing Python implementation. 

Parallelization using prange was considered, but overlapping patch accumulation into theta and thetax would introduce problems.

## Motivation and Context

The fix was motivated by #3193. The current implementation of the default 'eig' path relies on a heavy Python triple loop. The new Cython core moves that algorithm into compiled code while providing the same behavior.

## How Has This Been Tested?

The provided implementation passes all existing tests in ([dipy/dipy/denoise/tests/test_lpca.py](https://github.com/dipy/dipy/blob/master/dipy/denoise/tests/test_lpca.py)).
 It also proves to be faster than current python version when running [dipy/doc/examples/denoise_localpca.py](https://github.com/dipy/dipy/blob/master/doc/examples/denoise_localpca.py) on Windows. On average, it was 26% faster across 5 runs. Some variability between runs was observed (likely due to system-level factors). A more controlled benchmarking setup could provide a more precise estimate of the performance gain.

## Questions for Reviewers

- Does this design align with DIPY's expectations?
- Would it be valuable to extend the PR to include a Cython 'svd' implementation? I did not yet dive into it, so far I focused on the default path, but from what I've read the 'svd' path is quite heavier on Python and it could benefit from a refactor too.
- Do you have any suggestions for a preferred, reproducible benchmarking setup?

## Checklist

## Type of Change

- [ ] Documentation update


skoudoro:
Thank you for this @TomasGuija. 

Can you address the conflicts in this PR. 

Otherwise we will look deeper in this a bit later. Thanks !

TomasGuija:
> Thank you for this @TomasGuija.
> 
> Can you address the conflicts in this PR.
> 
> Otherwise we will look deeper in this a bit later. Thanks !

Sure, I'll do so tomorrow as soon as I can.

dipy/denoise/eig_localpca.pyx
```diff
@@ -0,0 +1,637 @@
...
+from numpy cimport ndarray
+
+from libc.stdlib cimport malloc, free
+from libc.math cimport sqrt
+
+from scipy.linalg.cython_lapack cimport ssyev, dsyev
+from scipy.linalg.cython_blas cimport sgemm, dgemm
+
+ctypedef cnp.float32_t f32
+ctypedef cnp.float64_t f64
```
skoudoro, line 26: let's use `cython.floating`

dipy/denoise/eig_localpca.pyx
```diff
@@ -0,0 +1,637 @@
...
+                    pca_classifier_f32(d, N, n_samples, &this_var, &ncomps)
+                else:
+                    # Predefined variance
+                    this_var = var_map[i, j, k]
+
+                # Threshold by tau
+                tau = (tau_factor * tau_factor) * this_var
+
+                # Update ncomps according to tau_factor
+                ncomps = 0
```
skoudoro, line 561: ncomps, overwrites the result above. So what was the goal of ncomps. never use it seems

dipy/denoise/eig_localpca.pyx
```diff
@@ -0,0 +1,637 @@
...
+
+    cdef f32[:, ::1] X
+    cdef f32[::1] M
+    cdef f32[::1, :] C
+    cdef f32[::1] d
+    cdef f32[:, ::1] proj_buf
+    cdef f32[::1] work
+
+    cdef int info
+    cdef int lwork = 3 * N   # sufficient workspace for ssyev
```
skoudoro: very safe, maybe need a better formula for better performance

skoudoro (changes requested):
Hi @TomasGuija,

Thank you for looking into this. Before a deeper review:

- please add tests.
- Make sure fast and current implementation provide similar result
- use cython.floating fused type because a lot of duplication and maintenance burden.
- What happens when non-convergence ?
- More docstring needed, currently very hard to review.

Can you look at all of this before we go further

dipy/denoise/eig_localpca.pyx
TomasGuija: Just refactored the code as much as possible to avoid duplication. Didn't manage to remove the `run_core_f32` and `run_core_f64` methods though. I think they might be needed as wrappers at the Python/Cython boundary. I'll explore further though, I'm still learning about Cython.

dipy/denoise/eig_localpca.pyx
TomasGuija: You're right. I replicated the `localpca::_pca_classifier` method which also computes ncomps, but didn't realize it's returned value isn't used. I removed it the unsued ncomps now.

TomasGuija:
> Hi @TomasGuija,
> 
> Thank you for looking into this. Before a deeper review:
> 
> * please add tests.
> * Make sure fast and current implementation provide similar result
> * use cython.floating fused type because a lot of duplication and maintenance burden.
> * What happens when non-convergence ?
> * More docstring needed, currently very hard to review.
> 
> Can you look at all of this before we go further

Thank you very much for your time @skoudoro. I already looked into some of the details you pointed out. When I'm finished with everything I'll let you know here.

dipy/denoise/eig_localpca.pyx
TomasGuija: Looking into the [`LAPACK documentation`](https://netlib.org/lapack/explore-html/d8/d1c/group__heev_gab3474da7a9f45653c4b7bc7d523ebc61.html) I found that we can call `ssyev` and `dsyev` with `lwork=-1` to make a query for the optimal workspace to be allocated. I changed the code to make a call before the triple loop to manage the needed workspace.

TomasGuija:
> Hi @TomasGuija,
> 
> Thank you for looking into this. Before a deeper review:
> 
> * please add tests.
> * Make sure fast and current implementation provide similar result
> * use cython.floating fused type because a lot of duplication and maintenance burden.
> * What happens when non-convergence ?
> * More docstring needed, currently very hard to review.
> 
> Can you look at all of this before we go further

Hello @skoudoro,  

While I may still make a few adjustments, I believe most of the requested changes have now been addressed:

1. We now use `cython.floating` to avoid duplicated implementations. I believe `run_core_f32` and `run_core_f64` are still necessary as a bridge between Cython and Python, though I’m not sure if we can avoid those.   

2. I added a couple of tests under `test_lpca`:  
   - one to ensure consistent outputs across both typing paths  
   - another to verify that the fast (Cython) and original (Python) implementations produce similar results  

3. I refactored the code following a pattern similar to the one used in `reconst/dki.py`, introducing a `fast` flag to select between the Python and Cython backends. This keeps the original implementation available within this PR, while allowing `fast=True` (default) to use the optimized path if it proves reliable.  

4. For non-convergence handling, I reviewed the current implementation based on `scipy.linalg.eigh`. Since SciPy raises different errors depending on the `info` flag, I attempted to mimic that behavior accordingly.  

5. Finally, I added docstrings. I may revisit them to further refine clarity and accuracy, but they should already be in a good state.

skoudoro:
Can you rebase this PR @TomasGuija ?

Thank you

TomasGuija:
> Can you rebase this PR @TomasGuija ?
> 
> Thank you

Hello, I just did it. 

I checked the tests, and I believe the new complex values behaviour doesn't work with the Cython version. I will check them out as soon as I can to see exactly what's going on. I will leave a new comment too and mention you to let you know.

TomasGuija:
Hi @skoudoro, quick update on the latest changes: 

I added the complex-valued Cython paths for the eigenvalue-based local PCA core, including separate `complex64` and `complex128` implementations.

One implementation detail worth mentioning: the complex loops are intentionally duplicated for the two complex dtypes. I initially tried to make this more generic with fused types, but using both the real fused type and the complex fused type in the same implementation failed at compile time.

At the moment, all tests passing locally. I may still add a focused follow-up test to check that complex-valued eig-based denoising remains consistent between `fast=True` and `fast=False`, similar to the existing real-valued fast/reference comparison.

PR #3808, opened 2026-03-04
RF: improve trx support for recobundles CLI

skoudoro:
This pull request updates the bundle recognition workflow in `dipy/workflows/segment.py` to output all recognized bundles as named groups in a single TRX file, instead of saving each bundle as a separate tractogram file.

### DIFF
--- a/dipy/workflows/nn.py
+++ b/dipy/workflows/nn.py
@@ -181,7 +181,15 @@ class BiasFieldCorrectionFlow(Workflow):
                     "biasfield_corrected.nii.gz"
                 ).with_name(f"{prefix}_biasfield_corrected.nii.gz")
 
-        self.update_flat_outputs(self.flat_outputs, io_it)
+        if io_it:
+            self.update_flat_outputs(self.flat_outputs, io_it)
+        else:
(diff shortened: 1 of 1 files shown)


skoudoro:
Waiting for https://github.com/tee-ar-ex/trx-python/issues/108 which should fix the current failing CI's on windows

arokem:
I just retriggered all the CI runs here. Let's hope the GHA outage is over...

arokem:
LGTM! Let's see how the CI fares.

skoudoro:
Thanks! I will update the PR because `StatefulTractogram` do not manage correctly group. This is problematic for visualization tools.

For Now, I will go back to one file per bundle instead of one file with everything.

Do you use `StatefulTracgram` in pyafq or only trx directly @arokem ?

arokem:
We do rely on the SFT class in pyAFQ. We also implemented a [SegmentedSFT](https://github.com/tractometry/pyAFQ/blob/main/AFQ/utils/streamlines.py#L17) class, which may be interesting to bring into DIPY at some point for this kind of use-case.

PR #3816, opened 2026-03-04
RF: relax precision for affine_registration / motion_correction

skoudoro:
The main change is the addition of the `optimizer_options` parameter to key registration functions, allowing users to customize optimizer behavior. Default optimizer options have also been updated to provide more robust optimization. Additionally, informative logging statements have been added to clarify the registration process steps.

### DIFF
--- a/dipy/align/_public.py
+++ b/dipy/align/_public.py
@@ -415,6 +415,7 @@ def affine_registration(
     ret_metric=False,
     moving_mask=None,
     static_mask=None,
+    optimizer_options=None,
     **metric_kwargs,
 ):
     """
@@ -485,6 +486,10 @@ def affine_registration(
         static image mask that defines which pixels in the static image
         are used to calculate the mutual information.
 
+    optimizer_options : dict, optional
+        AffineRegistration key-word argument: options to be passed to the
+        optimizer. See `scipy.optimize.minimize` documentation for details.
+
     nbins : int, optional
         MutualInformationMetric key-word argument: the number of bins to be
         used for computing the intensity histograms. The default is 32.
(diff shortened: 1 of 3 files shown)


maharshi-gor (approved):
Thank you @skoudoro for the work!

LGTM. Merging.

raj-magesh:
FYI, this change in default precision resulted in poor CT-to-MRI registration on some data I'm working with (specifically, on the `rigid` step after `center_of_mass` and `translation`).

(It took me a while to troubleshoot too.)

Might I suggest reverting the change to the default (i.e. leaving `ftol` unset) but keeping the `optimizer_options` parameter for those who want to fiddle with it?

Or if not, adding a warning in the log analogous to `Default level_iters have been updated to [1000, 500, 100] for performance improvement. Identical results are expected. In case of any discrepancy, you can revert to the previous default by setting level_iters=[10000, 1000, 100].` so that folks aren't caught off-guard like I was.

skoudoro:
Thank you for the feedback @raj-magesh.

- the change of `level_iters` is another PR and a warning has been added already
- I agree that we should have add a warning concerning the change of precision. Apologize for that.

So I think (what I have in mind, open to other ideas 😅 ),

- The core class `AffineRegistration` will return back to the previous values (`level_iters` and `precision`)
-  the helper function `affine_registration` will go back to the previous values  (`level_iters` and `precision`)
- the helper function `motion_correction` or `register_dwi_*`, will keep its current values with more warnings.

We will try to cut a release next month, when we get more feedback from the release 1.12.0.

raj-magesh:
No worries, thanks for the quick response!

In my case, `level_iters` didn't appear to make a difference, it was just the change in default `ftol` that led to poor alignment.

But yeah, I think defaulting to high-fidelity alignment is probably best. Folks who want faster alignment can always optimize the parameters on their own data. Perhaps someone could evaluate these parameters on a large test suite of real data to find better defaults.

skoudoro:
1.12.1 is released @raj-magesh.  the fix is there. thanks again for the feedback

PR #3827, opened 2026-03-07
Enh asv continuous benchmarks

Kaushalt2004:
Adds an opt-in ASV “continuous” benchmark workflow to CI.
- Runs only when you manually dispatch it or add the run-benchmark label to a PR (so it won’t slow down normal CI).
- Compares PR HEAD vs master`using the existing ASV config/template and posts the asv compare summary as a PR comment.
- Uses cache + caching and pins thread env vars for more stable numbers.
- Tightens permissions by keeping benchmark execution read-only and doing PR commenting in a separate job.
- Also fixes asv_compare.conf.json.tpl to include the missing `build` dependency, and makes the old always-on benchmark workflow manual-only.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -1,6 +1,7 @@
 name: Benchmarks
 
-on: [push, pull_request]
+on:
+  workflow_dispatch:
 
 permissions:
   contents: read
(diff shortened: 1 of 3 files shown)


skoudoro:
Hi @Kaushalt2004, 

this PR is currently very messy and mix a lot of stuff.

Unfortunately it can not be reviewed. Closing

PR #3839, opened 2026-03-10
BUG: Add missing int() conversion in shore_indices function

Kaushalt2004:
I noticed that shore_indices()` in `dipy/reconst/shore.py computes n_c without converting it to an integer:

python
n_c = np.round(1 / 6.0 * (F + 1) * (F + 2) * (4 * F + 3))

This gives back a numpy.float64 instead of a plain int. The same calculation appears in 5 other places in the file, and all of them correctly wrap it in int() — so this one was likely just missed.
The fix

One-line change — just adding `int()` around the result to match the rest of the file:

```diff
- n_c = np.round(1 / 6.0 * (F + 1) * (F + 2) * (4 * F + 3))
+ n_c = int(np.round(1 / 6.0 * (F + 1) * (F + 2) * (4 * F + 3)))
```

### DIFF
--- a/dipy/reconst/tests/test_shore.py
+++ b/dipy/reconst/tests/test_shore.py
@@ -111,9 +111,9 @@ def test_shore_fitting_constrain_e0():
 
 def test_shore_indices_returns_int():
     # n_c should be computed as int, not numpy.float64
-    n, l, m = shore_indices(4, 0)
+    n, ell, m = shore_indices(4, 0)
     assert isinstance(n, int)
-    assert isinstance(l, int)
+    assert isinstance(ell, int)
     assert isinstance(m, int)
 
 


Rishav23av:
Hi @Kaushalt2004, great catch! I was looking at this issue too.

The fix looks correct to me  matches the pattern used in all 5 other 
places in the file.

One thing I noticed while looking at the code: there are no existing tests 
for `shore_indices` that check the return type. It might be worth adding 
a small test to prevent this from regressing in the future:
```python
def test_shore_indices_type():
    from dipy.reconst.shore import shore_indices
    # n_c should be int, not numpy.float64
    result = shore_indices(4, 0)
    assert isinstance(result, int)
```

Just a suggestion  the fix itself is solid!

Kaushalt2004:
Yes thanks for the comment i will look into it.Thanks

Kaushalt2004:
> Hi @Kaushalt2004, great catch! I was looking at this issue too.
> 
> The fix looks correct to me matches the pattern used in all 5 other places in the file.
> 
> One thing I noticed while looking at the code: there are no existing tests for `shore_indices` that check the return type. It might be worth adding a small test to prevent this from regressing in the future:
> 
> ```python
> def test_shore_indices_type():
>     from dipy.reconst.shore import shore_indices
>     # n_c should be int, not numpy.float64
>     result = shore_indices(4, 0)
>     assert isinstance(result, int)
> ```
> 
> Just a suggestion the fix itself is solid!




This is the result after the test

FIXED (with int()):
  n_c = 22, type = int
  isinstance(n_c, int) = True

OLD (without int()):
  n_c = 22.0, type = float64
  isinstance(n_c, int) = False

Rishav23av:
Thanks for adding the test @Kaushalt2004! 

The test looks solid checking all three return values `(n, l, m)` is 
even better than what I suggested 

@skoudoro would you be able to take a look when you get a chance

skoudoro (approved):
LGTM, thanks @Kaushalt2004, 

merging

PR #3863, opened 2026-03-14
BF: fix synthseg indexing and reduce batch size for test to reduce me…

skoudoro:
1. `dipy/nn/torch/synthseg.py` — Fixed a slicing bug in the batch prediction loop: prediction[:batch_idx] →
  prediction[batch_start:batch_idx]. The original code would overwrite previously computed results in multi-iteration batches (when batch_size < len(T1)), producing incorrect output.
 2. `dipy/nn/tests/test_synthseg.py` — Changed batch_size=2 → batch_size=1. This processes 2 images sequentially instead of simultaneously, halving the peak tensor size during inference (~1.87 GB → ~0.93 GB peak for temp_pred), keeping total memory under 7 GB on CI runners. The batch code path is still fully tested since multiple images are still passed as a list.

### DIFF
--- a/dipy/nn/tests/test_tf.py
+++ b/dipy/nn/tests/test_tf.py
@@ -59,7 +59,7 @@ def test_default_mnist_sequential():
 
     hist = model.fit(x_train, y_train, epochs=epochs)
     model.evaluate(x_test, y_test, verbose=2)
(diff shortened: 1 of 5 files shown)


maharshi-gor:
Thank you for the work @skoudoro 

LGTM Merging

Issue #3837, opened 2026-03-10
Latest DIPY broken on headless environments


36000:
### Summary

Hello all, 

We have run into an issue with DIPY after updating to Python 3.13 and the latest Fury. Overall we are very happy with Fury 2.0, however we have one issue. In PyAFQ, we have some tests which require graphics libraries and run visualization code, and these work well. But, we also have some tests where no graphics libraries are installed and we are just running normal dipy code (load_tractogram, save_tractogram, etc.). In these, we do not want to depend on graphics libraries as we do not assume pyAFQ users need to install graphics libraries unless they are actually making graphics.

So here is the bug. In python 3.13 and the latest DIPY 1.11.0, if you do some normal code in a headless environment:

```
from dipy.io.streamline import save_tractogram
```

You will get an error:

```
E   RuntimeError: Request adapter failed (3): Validation Error
E   
E   Caused by:
E     No suitable graphics adapter found; noop support not compiled in, vulkan drivers/libraries could not be loaded, metal support not compiled in, dx12 support not compiled in, gl drivers/libraries could not be loaded, webgpu support not compiled in
```

This is because dipy.io pulls in visualization libraries to save/load VTK, even if you are just grabbing save_tractogram to load a trk or trx file:

```
/opt/hostedtoolcache/Python/3.11.14/x64/lib/python3.11/site-packages/AFQ/recognition/utils.py:10: in <module>
    from dipy.io.streamline import save_tractogram
/opt/hostedtoolcache/Python/3.11.14/x64/lib/python3.11/site-packages/dipy/io/streamline.py:15: in <module>
    from dipy.io.vtk import load_vtk_streamlines, save_vtk_streamlines
/opt/hostedtoolcache/Python/3.11.14/x64/lib/python3.11/site-packages/dipy/io/vtk.py:10: in <module>
    import fury.io
/opt/hostedtoolcache/Python/3.11.14/x64/lib/python3.11/site-packages/fury/io.py:13: in <module>
    from fury.lib import Texture, wgpu
/opt/hostedtoolcache/Python/3.11.14/x64/lib/python3.11/site-packages/fury/lib.py:135: in <module>
    wgpu_device = gfx.renderers.wgpu.get_shared().device
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/opt/hostedtoolcache/Python/3.11.14/x64/lib/python3.11/site-packages/pygfx/renderers/wgpu/engine/shared.py:267: in get_shared
[...]

PR #3865, opened 2026-03-15
FIX: Move VTK imports to lazy loading to support headless environments

Sayeem3051:
## Checklist

- [ ] I have read the [CONTRIBUTING](https://github.com/dipy/dipy/blob/master/.github/CONTRIBUTING.md) guidelines.

## Type of Change

- [ ] Documentation update

### DIFF
--- a/dipy/io/streamline.py
+++ b/dipy/io/streamline.py
@@ -99,7 +99,7 @@ def save_tractogram(
         nib.streamlines.save(fileobj, str(filename))
 
     elif extension in [".vtk", ".vtp", ".fib"]:
-                    from dipy.io.vtk import save_vtk_streamlines
+                            from dipy.io.vtk import save_vtk_streamlines
         binary = extension in [".vtk", ".fib"]
         save_vtk_streamlines(sft.streamlines, filename, binary=binary, to_lps=False)
         logger.warning(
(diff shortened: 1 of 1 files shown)


arokem:
I believe this might adequately address #3837

Sayeem3051:
Fixed: The indentation issue in the lazy-loaded VTK imports has been corrected.

The imports on lines 102 and 221 now have the proper indentation (8 spaces) to match the elif blocks they belong to. This should resolve the `ruff-format` pre-commit check failure.

The changes in this PR successfully address issue #3837 by moving VTK imports to lazy loading, which allows DIPY to be imported in headless environments without graphics libraries installed.

skoudoro:
> I believe this might adequately address https://github.com/dipy/dipy/issues/3837

I don't think so @arokem. It just delay the issue...

skoudoro:
I strongly believe that is needs to be fixed in fury

Sayeem3051:
`I understand your concern, @skoudoro. You're right that the root cause should be fixed in Fury library. However, this PR still provides value as an intermediate solution:

**Benefits of this PR (lazy loading approach):**
1. Allows users to import `save_tractogram`/`load_tractogram` in headless environments for non-VTK formats (trk, trx, etc.)
2. Provides relief for existing DIPY users with Fury v2.0 until Fury releases a fix
3. Minimal code change with zero breaking changes
4. Can coexist alongside the Fury fix

**Recommended next steps:**
1. **Keep this PR merged** - It helps the current situation while Fury is being fixed
2. **File issue in Fury repository** - Request that Fury make graphics initialization optional or lazy
3. **Track Fury fix** - Once Fury releases a version that doesn't require graphics on import, that will provide the complete solution

Would this be an acceptable approach? We can move forward with merging this as a temporary mitigation while the proper fix is pursued upstream in Fury.`

skoudoro:
I prefer to avoid intermediate solution and prefer long term solution.

Furthermore, indentation seems wrong. 

Thank you for spending the time to try to fix this @Sayeem3051.

But closing for now

PR #3888, opened 2026-03-18
[Draft] NF: add wgpu-accelerated mppca backend

NekoAsakura:
## Description

GPU-accelerated MPPCA via WebGPU.
Adds `backend='gpu'` parameter to `mppca()` to enable this feature.
This is a prototype for GSoC 2026 Project 3.
Related: #3193, #3806, #1934

## Motivation and Context

Profiling `genpca()` on the CFIN multi-b dataset (96×96×19×67, b ≤ 400) shows `eigh` takes 58% of per-patch runtime. The remaining goes to Python dispatch, dot products, and classification. Each patch is ~29 KB — fits in L1 cache — so batched vectorization may actually hurt by thrashing the cache. The real bottleneck is that the eigendecompositions are sequential and cannot be parallelized on CPU, and GPU solve both.

  <details>
  <summary>Line profile of genpca() on CFIN multi-b (96×96×19×67)</summary>

  lineprof_heatmap

  </details>

## How Has This Been Tested?
**System Info**
```
  CPU: i9-13900K
  GPU: RTX 4090
  OS:  Linux 6.17.0-14-generic
```
  CFIN multi-b dataset (b ≤ 400), shape 96×96×19×67, patch_radius=2:
```
  === Timing ===
  CPU:     51.1s
  GPU:     4.0s

  === Denoised output (GPU vs CPU) ===
  Mean absolute error:  0.0007
  RMSE:                 0.0034
  Mean relative error:  0.0041%
  99th pctile rel err:  0.0900%
  Voxels > 1% rel err:  1142 / 11495385 (0.0099%)

=== Sigma estimate (GPU vs CPU) ===
  Mean absolute error:  0.0002
  RMSE:                 0.0007
  Mean relative error:  0.0070%
  99th pctile rel err:  0.1283%
  Voxels > 1% rel err:  18 / 175104 (0.0103%)
```

Most error comes from float32 precision rounding errors. A float64 path can be added if necessary.

<details>
<summary>cProfile breakdown: CPU vs GPU</summary>

cprofile_breakdown

</details>

## Next steps
**Reduce polling overhead**: cProfile shows ~2.9s of GPU wall time is `wgpu-py` synchronous polling (`time.sleep`), async readback (`buffer.map_async`) can eliminate this.

This prototype hard-wires the GPU pipeline inside `_gpu_mppca.py`. For GSoC, I would redesign this into a proper GPU compute layer.

## Checklist

## Type of Change

- [ ] Documentation update

### DIFF
--- a/dipy/denoise/patch2self.py
+++ b/dipy/denoise/patch2self.py
@@ -83,7 +83,7 @@ def _extract_3d_patches(arr, patch_radius):
     return np.array(all_patches)
 
 
-def _fit_denoising_model(train, vol_idx, model, alpha):
+def _fit_denoising_model(train, vol_idx, model, alpha, *, copy_X=True):
     """Fit a single 3D volume using a train and test phase.
 
     Parameters
@@ -104,8 +104,8 @@ def _fit_denoising_model(train, vol_idx, model, alpha):
         and other objects that inherit from `sklearn.base.RegressorMixin`.
     alpha : float
         Regularization parameter only for ridge and lasso regression models.
-    version : int
-        Version 1 or 3 of Patch2Self to use.
+    copy_X : bool, optional
+        If True, the design matrix X will be copied; else, it may be overwritten.
 
     Returns
     -------
@@ -117,21 +117,21 @@ def _fit_denoising_model(train, vol_idx, model, alpha):
     """
     if isinstance(model, str):
         if model.lower() == "ols":
-            model_instance = linear_model.Ridge(copy_X=False, alpha=1e-10)
+            model_instance = linear_model.Ridge(copy_X=copy_X, alpha=1e-10)
         elif model.lower() == "ridge":
-            model_instance = linear_model.Ridge(copy_X=False, alpha=alpha)
+            model_instance = linear_model.Ridge(copy_X=copy_X, alpha=alpha)
         elif model.lower() == "lasso":
-            model_instance = linear_model.Lasso(copy_X=False, max_iter=50, alpha=alpha)
+            model_instance = linear_model.Lasso(copy_X=copy_X, max_iter=50, alpha=alpha)
         else:
             raise ValueError(
                 f"Invalid model string: {model}. Should be 'ols', 'ridge', or 'lasso'."
             )
-    elif isinstance(model, linear_model.BaseEstimator):
+    elif isinstance(model, sklearn.base.BaseEstimator):
         model_instance = model
     else:
         raise ValueError(
             "Model should either be a string or \
-                an instance of sklearn.linear_model BaseEstimator."
+                an instance of sklearn.base.BaseEstimator."
         )
     cur_x, y = _vol_split(train, vol_idx)
     model_instance.fit(cur_x.T, y.T)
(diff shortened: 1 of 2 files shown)


skoudoro:
Thank you for exploring this @NekoAsakura,

Currently CI's do not have GPU, SO I will have to look at this more deeper.

For now, I believe:

- code can be simplified, it seems there are a lot of duplication or function that are very close....
- Why there is still a triple python loop?

it is a good start but a lot of work to finalize

NekoAsakura:
Thank you for the review! Now this pr is a working prototype, not a polished one.

> Why there is still a triple python loop?

Yes, I can replace it with `np.mgrid` + mask indexing to vectorise the patch index precomputation. I intentionally left it as-is for now so that the benchmark numbers reflect pure GPU acceleration gains rather than mixed optimisations. I'll get this cleaned up within the week.

> code can be simplified, it seems there are a lot of duplication or function that are very close....

Agreed — the shared/global shader variants should be unified via runtime string templating instead of maintaining near-identical copies. I've been tied up with lab work lately, so I'll tackle this cleanup after I submit the GSoC application.

I'd also really appreciate a response on [DIPY discussion #3889](https://github.com/dipy/dipy/discussions/3889) when you get a chance — the GSoC proposal deadline is fast approaching on March 31, and the answers would help me finalise the timeline.

NekoAsakura:
Hi @skoudoro, I have vectorised `_precompute_patch_indices` with `np.mgrid` — the triple Python loop is gone. The output is element-wise identical to the original loop (verified on CFIN).

No measurable runtime difference on this dataset (GPU: 3.97s vs 4.0s, CPU: 51.85s vs 51.1s — within noise). May benefit on larger volumes.

NekoAsakura:
Hi @skoudoro, I've spent some time optimising the GPU pipeline, and runtime is now down to 3.37s.

Here's what changed:
- **Single compute pass per tile** - all 4 stages dispatched in one encoder, no more cpu-gpu sync between passes
- **Native f32 atomic support** - auto probe whether the gpu adapter supports `shader-float32-atomic` extension
- **Unified shader variants** - applied conditional templating and got rid of wgsl duplication
- **Eliminated tile-local patch index buffer** - accumulate reads from the full buffer with `tile_offset` now

I've also done a fair bit of reading on eigendecomposition algorithm choices. There's room for discussion on alternatives, but the current Jacobi is already a well-optimised version. Happy to have that chat publicly after the GSoC proposal deadline.

**Next steps**: I'll get the proposal updated, then I'm waiting on your feedback for both the code and the proposal. Cheers!

NekoAsakura:
New benchmark(5 times avg ± std)
CFIN (96, 96, 19, D), Sherbrooke(128, 128, 60, 193), r=2.
| dataset | D | GPU | CPU |
|---|---:|---:|---:|
| CFIN | 34 | 0.310 ± 0.014 s | 18.808 s |
| CFIN | 67 | 0.695 ± 0.013 s | 44.953 s |
| CFIN | 100 | 1.713 ± 0.030 s | 85.747 s |
| CFIN | 133 | 3.751 ± 0.019 s | 140.869 s |
| Sherbrooke | 193 | 19.592 ± 0.045 s | 1322.771 s |

skoudoro:
Hi @NekoAsakura,

Can you share your benchmark file ? 

Also, address the code format CI's please

dipy/denoise/_gpu_mppca.py
```diff
@@ -0,0 +1,499 @@
...
+
+    if use_economy:
+        shader_cov = _compile_shader(device, "extract_economy_cov.wgsl", D=D)
+    else:
+        b_max = (max_shared - D * 4) // (D * 4)
+        b_max = max(1, min(b_max, num_samples))
+        shader_cov = _compile_shader(device, "extract_covariance.wgsl", D=D, B_MAX=b_max)
+
+    shader_tridiag = _compile_shader(
+        device, "householder_tridiag.wgsl", sizes={"1111": eco_dim, "2222": eco_dim}
```
skoudoro: I do not understand this `1111` or `2222`. Can you explain ?

NekoAsakura:
Thanks for the feedback, @skoudoro. Given the additional time this would require and my current schedule, I've decided not to continue with the prototype. I'd rather focus on refactoring and bug fixes that deliver value to users sooner (e.g. #4005). Thanks again for the time you spent reviewing.

skoudoro:
Thank you for your feedback and working on this!

PR #3904, opened 2026-03-24
ENH: reuse prefactorized CSD normal matrix in csdeconv

### LINKED ISSUE #3903: Avoid repeated Cholesky factorization of static normal matrix in csdeconv
### Summary

While profiling the csdeconv implementation in DIPY, I observed that the base normal matrix P = XᵀX is constant for a given ConstrainedSphericalDeconvModel configuration, but its Cholesky factorization is recomputed repeatedly during voxel-wise solves.

Specifically, the initial unconstrained solve inside csdeconv factorizes P for each voxel, even though P does not change across the model fit. This introduces avoidable overhead in a hot path that is executed many times in typical whole-volume processing.

Why this matters:
- CSD is commonly applied across large datasets (many voxels), so small per-voxel inefficiencies accumulate
- Profiling indicates that linear algebra operations (including Cholesky factorization) contribute significantly to runtime
- The repeated factorization of a constant matrix is redundant and can be safely avoided

Proposed direction:
- Precompute P = XᵀX and its Cholesky factorization once during model initialization
- Reuse the prefactorized matrix during voxel-wise solves in csdeconv
- Maintain backward compatibility by keeping the current behavior as a fallback when no prefactorization is provided

Expected impact:
- Small but consistent reduction in per-voxel computation time
- Improved scalability for large-volume processing
- No change in numerical results or convergence behavior

I’d be happy to work on this and open a PR implementing the optimization along with tests and benchmarks.

### Platform

windows 11

### DIPY version

1.13.0.dev0

### Environment

_No response_

### Python version

3.11

### Additional information

_No response_

Kaushalt2004:
Description:

While profiling csdeconv, I noticed that the normal matrix P = XᵀX is constant for a given model, but its Cholesky factorization is recomputed for every voxel. This adds unnecessary overhead in a frequently used code path.

This PR optimizes that by computing the factorization once during model initialization and reusing it during voxel-wise solves.

Changes:
- Precompute P and its Cholesky factor once per model
- Reuse it in csdeconv (with fallback to existing behavior)

Why this matters:
- CSD is run on many voxels, so even small savings add up
- Removes redundant work without changing the math

Validation:
- Outputs match the existing implementation
- Added a regression test

Benchmark (micro):
- ~2–5% speedup per call

This is a small, safe optimization with no API changes.

Optional follow-up:
- Explore caching for repeated constraint sets in iterations

### DIFF
--- a/dipy/reconst/csdeconv.py
+++ b/dipy/reconst/csdeconv.py
@@ -67,8 +67,6 @@ class AxSymShResponse:
 
 
 class ConstrainedSphericalDeconvModel(SphHarmModel):
-    """Constrained Spherical Deconvolution (CSD) model."""
-
     @deprecated_params("sh_order", new_name="sh_order_max", since="1.9", until="2.0")
     @warning_for_keywords()
     def __init__(
@@ -190,7 +188,7 @@ class ConstrainedSphericalDeconvModel(SphHarmModel):
         self.tau = tau
         self.convergence = convergence
         self._X = X = self.R.diagonal() * self.B_dwi
-        self._P = np.dot(X.T, X)
+        self._P, self._P_chol = _regularized_cholesky(np.dot(X.T, X), 1e-5)
 
     @multi_voxel_fit
     def fit(self, data, **kwargs):
@@ -202,6 +200,7 @@ class ConstrainedSphericalDeconvModel(SphHarmModel):
             tau=self.tau,
             convergence=self.convergence,
             P=self._P,
+            P_chol=self._P_chol,
         )
         return SphHarmFit(self, shm_coeff, None)
 
@@ -251,8 +250,6 @@ class ConstrainedSphericalDeconvModel(SphHarmModel):
 
 
 class ConstrainedSDTModel(SphHarmModel):
-    """Constrained Spherical Deconvolution Transform (SDT) model."""
-
     @deprecated_params("sh_order", new_name="sh_order_max", since="1.9", until="2.0")
     @warning_for_keywords()
     def __init__(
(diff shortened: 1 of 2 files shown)


calebjubal:
Closes #3903

Kaushalt2004:
Please review @skoudoro

skoudoro:
Sorry for the delays, I will get back to you on Monday or Tuesday

Kaushalt2004:
Please review @skoudoro

skoudoro:
@Atharva-Shah-2298, can you look at this since you are looking at CSD this past period ?

Atharva-Shah-2298:
@Kaushalt2004 @skoudoro Yes I will. Lot on plate right now but I will find the time ASAP.

dipy/reconst/csdeconv.py
Atharva-Shah-2298, line 1: The error now will happen at init stage with the try except block gone.  I say its better to fail fast but it is changing signature of the failure. Do we want that @skoudoro

Atharva-Shah-2298 (commented):
I tested this implementation on 3 datasets and the SH are identical with less than 10^-12 difference in SH coefs. The speed gain is roughly 2-5% on average (3.4% , 4.8%, 2.1%) is what I got. The math is sound and this follows closely to the regularized scheme of the original implementers. 

The only caveat I feel is that the benchmarking is added in this PR. I am not sure about the scope of that to be honest. What do you think @skoudoro

skoudoro:
Thank you for the feedback @Atharva-Shah-2298 .

ok, I will clean up this PR because there is unrelated  change (`test_template.yml`) to be removed then we can merge.

Thank you for this work @Kaushalt2004

skoudoro:
Thanks for this work @Kaushalt2004 and sorry for the late review. Merging

PR #3908, opened 2026-03-25
BF: add mask input for biasfield correction cli

skoudoro:
This pull request enhances the bias field correction workflow in `dipy` by adding support for an optional brain mask input, allowing users to provide a mask file for more precise bias correction. The workflow and its tests are updated accordingly to handle this new feature.

It was an issue when mask was generated with another method. Without this option, biasfield was generating its own mask and we got a misalignment between both mask. 

ps: Using this PR to test backport functionality also for the future release 1.12.1


maharshi-gor (approved):
Thank you for the work @skoudoro

PR #3920, opened 2026-03-29
FIX: Terminate workflow when output files exist without force flag (#3496)

### LINKED ISSUE #3496: Not having the force option when output file exists does not stop the process
Despite the error message when not using the force option on CLI

```
"The following output files already exist, the "
"workflow will not continue processing any "
"further. Add the --force flag to allow output "
"files overwrite."
```

the workflow continues, throwing an unexpected error later in the workflow due to `io_it` being an empty list.

From what I understand from the error message, the warning message should be more of an error message, or terminate the workflow at the very least.

williamcancodee:
## Description

Makes a hard stop (`ValueError`) in `workflow.py` if output files already exist without using the `--force` flag. Also, it changes the associated file logging to `logger.error` and adds a test for this exception.

## Motivation and Context

The current implementation does not handle the case where users forget to use the `--force` flag. As a result, the CLI will run with an empty output list, causing unnecessary memory consumption for downstream tools, such as `dipy_recobundles`, which attempt to load massive datasets into memory but ultimately crash.

A hard stop would eliminate this memory and performance bottleneck.

Closes #3496

## How Has This Been Tested?

- Locally verified the hard stop when files exist without `--force`.
- I updated `dipy/workflows/tests/test_workflow.py` to correctly test for this exception.
- Locally run the test suite to verify that all assertions pass.

## Checklist

## Type of Change

- [ ] Documentation update

### DIFF
--- a/dipy/workflows/tests/test_workflow.py
+++ b/dipy/workflows/tests/test_workflow.py
@@ -20,8 +20,9 @@ def test_force_overwrite():
         mask_file = mo_flow.last_generated_outputs["out_mask"]
         first_time = os.path.getmtime(mask_file)
 
-        # re-run with no force overwrite, modified time should not change
-        mo_flow.run(data_path, out_dir=out_dir)
+        # re-run with no force overwrite should terminate early with a clear
+        # error, and therefore not modify existing outputs.
+        npt.assert_raises(ValueError, mo_flow.run, data_path, out_dir=out_dir)
         mask_file = mo_flow.last_generated_outputs["out_mask"]
         second_time = os.path.getmtime(mask_file)
         assert first_time == second_time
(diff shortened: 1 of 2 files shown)


skoudoro:
Hi @williamcancodee,

Thank you for looking into this.

unfortunately, having a hard stop is not the way to go. we should be able to continue without processing which is the case with the other CLI. I recommended you to try the other CLI to understand the behavior but I will have to closes this PR because it does not go to the right direction

feel free to open a new one when you understand how the other CLI works.

williamcancodee:
Hi @skoudoro,

Thank you for taking the time to review this and for providing clarity.
I see now why a hard stop isn't the right approach here, and that the process needs to continue

I'll take a step back, study how the other CLIs handle this as you suggested, and work on a new implementation. I'll open a fresh PR once I have it properly aligned with the project's standards.

Thank you again. I really appreciate your guidance

PR #3941, opened 2026-04-02
ENH: Add PeakPmfGen support for peaks based tracking

mrinalchaturvedi27:
## Description

This PR adds support for using pre computed peaks in Dipy's modern tracking algorithms `probabilistic_tracking,deterministic_tracking, and ptt_tracking`.
I found this issue while looking through the `dipy/tracking code`. I noticed that if a user tries to pass `peaks=` to `generic_tracking()` (around line 65 in `tracker.py`), the code automatically crashes and says: `NotImplementedError("Peaks are not yet implemented.")`.
This PR fixes that error by adding a new `PeakPmfGen` class that actually processes the peaks so the tracker can use them.

## Motivation and Context

Many users calculate their fiber peaks early in their workflow to save time, or because they are using custom peak finding methods.
Even though the tracking functions allow users to pass peaks as an argument, the backend never actually knew what to do with them. This left users stuck.
To fix this, I created a way to convert the sparse peak data into a continuous probability tracking sphere. It does this by finding the nearest voxel to the tracking point and using those peak directions.

## How Has This Been Tested?

Added `test_peaks_tracking()` in `dipy/tracking/tests/test_tracker.py` to prove the tracking algorithms can now use `peaks=` without crashing.
Added `test_pmf_from_peaks()` in `dipy/direction/tests/test_pmf.py` to ensure the new `PeakPmfGen` class correctly maps peak directions to the tracking sphere.
Double checked that the code compiles correctly (spin build) and has no memory leaks.
Made sure passing an invalid object to `peaks=` now throws a helpful `TypeError` instead of the old `NotImplementedError`.

## Checklist

## Type of Change

- [ ] Documentation update

### DIFF
--- a/dipy/direction/tests/test_pmf.py
+++ b/dipy/direction/tests/test_pmf.py
@@ -5,8 +5,7 @@ import numpy.testing as npt
 
 from dipy.core.sphere import HemiSphere, unit_octahedron
 from dipy.data import default_sphere, get_sphere
-from dipy.direction.pmf import SHCoeffPmfGen, SimplePmfGen, PeakPmfGen
-from dipy.direction.pmf import SHCoeffPmfGen, SimplePeakGen, SimplePmfGen
+from dipy.direction.pmf import PeakPmfGen, SHCoeffPmfGen, SimplePeakGen, SimplePmfGen
 from dipy.reconst import shm
 from dipy.testing.decorators import set_random_number_generator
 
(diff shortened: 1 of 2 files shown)


skoudoro:
I am not sure about the goal of this PR since we already have peak management with `SimplePeakPMF`. 

It does not make sense to me for now so closing

mrinalchaturvedi27:
> I am not sure about the goal of this PR since we already have peak management with `SimplePeakPMF`.
> 
> It does not make sense to me for now so closing

Hi @skoudoro, thanks for the review.
Actually, `PeakPmfGen` is different from `SimplePeakGen`:
`SimplePeakGen` needs discrete indices.
`PeakPmfGen` (this PR) supports continuous (x, y, z) vectors directly.
This fixes the `NotImplementedError` for users with raw vector data.

PR #3944, opened 2026-04-02
BF: Fix PTT initialization threshold and probe position, adjust defau…

deka27:
## Description                                                                                                                                                                                               
                                                            
  Fix two bugs in PTT initialization (`initialize_ptt_candidate`) and adjust the default `max_angle` for PTT in the workflow.                                                                                  
                                                                                                                                                                                                               
  ## Motivation and Context                                                                                                                                                                                    
                                                            
  1. **Missing pmf_threshold in initialization**: `compute_ptt_likelihood` thresholds FOD amplitudes below `pmf_threshold` to zero, but `initialize_ptt_candidate` did not. This caused inconsistent likelihood scoring between initialization and propagation, where noisy/weak FOD lobes could inflate the initial likelihood.
                                                                                                                                                                                                               
  2. **Wrong probe position in multi-probe path**: When `probe_count > 1`, the code computed offset probe positions into `position` but then passed `&stream_data[19]` (the center point) to `get_pmf_value_c`. All probes were sampling the same center point, defeating the purpose of multi-probe sampling.
                                                                                                                                                                                                               
  3. **Default max_angle too loose for PTT**: The workflow passed `max_angle=30.0` (the global default) to PTT, overriding PTT's own default of 10 degrees. Now the workflow defaults `max_angle` to `None` and resolves it per-method: 10 degrees for PTT, 30 degrees for all others. If the user explicitly sets any value, it is respected. A logger info message is emitted when the default is applied, informing the user which value was chosen.                                                                                                                                                                                 
                                                            
  ## How Has This Been Tested?

  Tested locally with PTT tracking on existing data. Verified that the threshold is applied during initialization, probes sample at correct positions, and max_angle defaults/overrides work as expected.

  ## Checklist

  ## Type of Change

  - [ ] Documentation update

### DIFF
--- a/dipy/tracking/propspeed.pyx
+++ b/dipy/tracking/propspeed.pyx
@@ -502,6 +502,7 @@ cdef void initialize_ptt_candidate(TrackerParameters params,
 
     """
     cdef double[3] position
+    cdef double fod_amp
     cdef int count
     cdef int i
     cdef double* pmf
@@ -522,8 +523,10 @@ cdef void initialize_ptt_candidate(TrackerParameters params,
     stream_data[22] = 0  # last_val
 
     if params.ptt.probe_count == 1:
-        stream_data[22] = pmf_gen.get_pmf_value_c(&stream_data[19],
-                                                  &stream_data[1])  # position, frame[0]
+        fod_amp = pmf_gen.get_pmf_value_c(&stream_data[19],
+                                          &stream_data[1])  # position, frame[0]
+        fod_amp = fod_amp if fod_amp > params.sh.pmf_threshold else 0
+        stream_data[22] = fod_amp
     else:
         for count in range(params.ptt.probe_count):
             for i in range(3):
@@ -537,8 +540,10 @@ cdef void initialize_ptt_candidate(TrackerParameters params,
                               * sin(count * params.ptt.angular_separation)
                               * params.inv_voxel_size[i])
 
-            stream_data[22] += pmf_gen.get_pmf_value_c(&stream_data[19],
-                                                       &stream_data[1])
+            fod_amp = pmf_gen.get_pmf_value_c(position,
+                                              &stream_data[1])
+            fod_amp = fod_amp if fod_amp > params.sh.pmf_threshold else 0
+            stream_data[22] += fod_amp
 
 
 cdef void prepare_ptt_propagator(TrackerParameters params,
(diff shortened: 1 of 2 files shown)


dipy/tracking/propspeed.pyx
```diff
@@ -502,6 +502,7 @@ cdef void initialize_ptt_candidate(TrackerParameters params,
 
     """
     cdef double[3] position
+    cdef double fod_amp
```
skoudoro, line 505: Why do you use a temporary variable ?

dipy/workflows/tracking.py
```diff
@@ -155,6 +156,10 @@ def run(
                 seed_mask, affine, density=[seed_density, seed_density, seed_density]
             )
 
+            if max_angle is None:
+                max_angle = 10.0 if tracking_method in ["ptt"] else 30.0
```
skoudoro, line 160: in `pft_tracking` function, we use 20, why 10 here ? it would be great to have both the same.

dipy/tracking/propspeed.pyx
deka27: Used a temporary variable because I need to check the value before using it. Without it, I'd have to call the function twice — once to check and once to use — which would be wasteful since it's an expensive interpolation call. The temp variable lets me grab the result once, check if it's above the threshold, and then use it.

dipy/workflows/tracking.py
deka27: for ptt its 10, and pft its 20. This workflow doesnt have pft, we have different workflow for pft, no.

dipy/tracking/propspeed.pyx
skoudoro: you mean, you can not do that ?

`stream_data[22] = stream_data[22] if stream_data[22] > params.sh.pmf_threshold else 0`

Why ? that strange

dipy/workflows/tracking.py
skoudoro: oh yes, sorry for the confusion 👍

dipy/tracking/propspeed.pyx
deka27: That would work for the single-probe case. I just used the temp variable to keep it consistent with the multi-probe loop below, where I actually need it since I'm doing +=. But yeah, I can change it if you want.

skoudoro: ok, I see, fine, thanks for the feedback, merging

skoudoro:
All good, Merging

PR #3951, opened 2026-04-03
BF: ODFs working with different affine

maharshi-gor:
## Description

- Works with both LAS and RAS affines.
- ODF snaps to the images.
- pam5 without sh_coeffs will not error out.
- test cases intiailized for the same.

### DIFF
--- a/dipy/viz/skyline/app.py
+++ b/dipy/viz/skyline/app.py
@@ -3,6 +3,8 @@
 import os
 import time
 
+import numpy as np
+
 from dipy.io.utils import split_filename_extension
 from dipy.utils.logging import logger
 from dipy.utils.optpkg import optional_package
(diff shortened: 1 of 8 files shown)


JiwaniZakir (commented):
The removal of the bounds-clipping logic in `SHSlicer.set_slice` (the `np.clip(idx, 0, self.shape[dim] - 1)` lines) deserves attention. Previously this protected against out-of-range slice indices; now that `self.state` (world-space floats) is passed directly instead of clamped voxel indices, there's no guard if the world-space coordinates fall outside the volume extent. It's worth verifying the shader handles out-of-bounds float slice values gracefully, or re-introducing a clamp in world-space terms.

The `active_slice_x/y/z` uniform type change from `i4` to `f4` in `SlicedSphGlyphMaterial` is consistent with passing float world-space coordinates, but any WGSL shader code consuming these uniforms should be audited to confirm it was updated in lockstep — the diff doesn't show the shader side, so it's unclear if this is complete.

On the `is_left_handed` removal: the simplification is clean, but the original comment said left-handed affines required sign flips for `m > 0` coefficients. If that was genuinely wrong (i.e., the flip was a bug), a brief note in the commit or a unit test asserting correct ODF orientation under a negative-determinant affine (like LAS) would make the behavioral change verifiable rather than assumed.

skoudoro (approved):

PR #3955, opened 2026-04-05
RF: mask update on auto pipeline

skoudoro:
Fix some feedback from Workshop attendees


JiwaniZakir (commented):
In `reconst.py`, the `return sys.exit(1)` pattern on the error path is incorrect — `sys.exit(1)` raises `SystemExit`, so the `return` is unreachable dead code. It should either be just `sys.exit(1)` or, more consistently with the rest of the codebase, raise a `ValueError` or return early without calling `sys.exit`.

The `while` loop following `order_from_ncoef` also appears redundant: `order_from_ncoef` should already return the largest valid order for the given number of coefficients, so the subsequent loop that decrements by 2 suggests either `order_from_ncoef` isn't being used as intended, or there's an off-by-one assumption worth documenting explicitly with a comment.

In `templates.py`, changing the `tracking` step to consume `force_fit.out_pam` and `force_fit.out_fa` (instead of `csd_fit.out_pam` and `dti_fit.out_fa`) is a significant behavioral change for the auto pipeline — CSD is typically the preferred model for tractography. If FORCE is intentionally the new default, this warrants a note in the template or changelog explaining the rationale.

Finally, `engine = "ray"` is hardcoded in the `force_fit` pipeline block, which will silently fail or error in environments without Ray installed. A default of `"serial"` or a comment flagging the Ray dependency would be safer.

dipy/workflows/templates.py
```diff
@@ -559,8 +571,8 @@
 [[pipeline]]
 name = "tracking"
 cli = "dipy_track"
-pam_files = "${csd_fit.out_pam}"
-stopping_files = "${dti_fit.out_fa}"
+pam_files = "${force_fit.out_pam}"
+stopping_files = "${force_fit.out_fa}"
 seeding_files = "${brain_mask.out_mask}"
 seed_density = 2
```
maharshi-gor, line 577: This @skoudoro 

Let's keep it CSD for now.

maharshi-gor (approved):
Thank you for the change @skoudoro

skoudoro:
Superseded by #4045 (dipy-auto-clean) — clean 14-commit history on top of workshop26.

PR #3961, opened 2026-04-08
NF: Extensive colormap support for Skyline

maharshi-gor:
## Description

This PR introduces multiple changes

- It provides extensive colormap support.
- It provides de-bouncing for chocked up processes.
- It provides test cases for proposed changes.

## Motivation and Context

This change is required to introduce more colormaps for visualization of images. Like distinct and divergent.

## How Has This Been Tested?

You can load and 3D/4D image and change the colormaps and provide details about it.

## Type of Change

- [ ] Documentation update

### DIFF
--- a/dipy/viz/skyline/UI/elements.py
+++ b/dipy/viz/skyline/UI/elements.py
@@ -970,7 +970,7 @@ def segmented_switch(label, options, value, *, width=0, height=28):
 
     imgui.push_id(label)
 
-    value_options = [option.lower() for option in options]
+    value_options = [option.title() for option in options]
     current_value = value if value in value_options else options[0]
     label_color = THEME["text"]
     imgui.push_style_var(imgui.StyleVar_.frame_padding, (12.0, 6.0))
--- a/dipy/viz/skyline/render/image.py
+++ b/dipy/viz/skyline/render/image.py
@@ -30,6 +30,7 @@ if has_fury_v2:
         show_slices,
         volume_slicer,
     )
+    from fury.colormap import distinguishable_colormap
     from fury.lib import gfx
 
 imgui_bundle, has_imgui, _ = optional_package("imgui_bundle", min_version="1.92.600")
@@ -199,7 +200,26 @@ class Image3D(Visualization):
         self._volume_idx = 0
         self.interpolation = interpolation or "linear"
         self._value_percentiles = value_percentiles
-        self._colormap_options = ("Gray", "Inferno", "Magma", "Plasma", "Viridis")
+        self._colormap_options = (
+            "Gray",
+            "Inferno",
+            "Magma",
+            "Plasma",
+            "Viridis",
+            "Cool",
+            "Hot",
+            "Bone",
+            "Copper",
+            "Pink",
+            "Spring",
+            "Summer",
+            "Autumn",
+            "Winter",
+            "Jet",
+            "Cividis",
+            "Distinct",
+            "Divergent",
+        )
         self.colormap = colormap
         self._picked_voxel = None
         self._picked_intensity = None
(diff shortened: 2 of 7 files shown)


dipy/viz/skyline/render/image.py
```diff
@@ -225,16 +246,12 @@ def _pick_voxel(self, event):
         self._picked_intensity = self.active_volume[voxel]
 
     def _create_slicer_actor(self):
-        """Handle  create slicer actor for ``Image3D``.
-        None
-        """
+        """Handle  create slicer actor for ``Image3D``."""
         volume = self.active_volume
-        self.value_range = self._value_range_from_percentile(volume)
```
skoudoro, line 232: you remove this but when do you initialize `self.value_range` ? Are you always calling `_apply_colormap` to fix this value whatever the situation?

dipy/viz/skyline/render/image.py
```diff
@@ -272,12 +309,34 @@ def _apply_colormap(self, colormap):
...
+        elif self.colormap.lower() == "divergent":
+            map_colors = np.array([[0, 0, 1], [1, 0, 0]], dtype=np.float32)
+            map = gfx.cm.create_colormap(map_colors, n=2)
+            max_abs = np.max(np.abs(self.active_volume))
+            if max_abs == 0:
+                max_abs = 1.0
+            for actor in self._slicer.children:
+                actor.material.map = map
+                actor.material.clim = (-float(max_abs), float(max_abs))
+                self._set_interpolation("nearest")
```
skoudoro: why `self` ? why not `actor. _set_interpolation("nearest")`? why calling the same function 20-100 times ?

dipy/viz/skyline/render/image.py
```diff
@@ -272,12 +309,34 @@ def _apply_colormap(self, colormap):
...
+        elif self.colormap.lower() == "distinct":
+            map_colors = np.asarray(
+                distinguishable_colormap(nb_colors=256), dtype=np.float32
+            )
+            map = gfx.cm.create_colormap(map_colors, n=256)
+            for actor in self._slicer.children:
+                actor.material.map = map
+                self._set_interpolation("nearest")
```
skoudoro: why `self` ? why not `actor. _set_interpolation("nearest")`? why calling the same function 20-100 times ?

dipy/viz/skyline/render/image.py
```diff
@@ -544,10 +624,11 @@ def _axis_slider_bounds(axis):
         imgui.spacing()
         imgui.spacing()
         changed, new = segmented_switch(
-            "Interpolation", ["Linear", "Nearest"], self.interpolation
+            "Interpolation",
+            ["Linear", "Nearest"],
+            self.interpolation.capitalize(),
```
skoudoro: title() instead of capitalize()

dipy/viz/skyline/UI/elements.py
```diff
@@ -970,7 +970,7 @@ def segmented_switch(label, options, value, *, width=0, height=28):
 
     imgui.push_id(label)
 
-    value_options = [option.lower() for option in options]
+    value_options = [option.capitalize() for option in options]
```
skoudoro: title() instead of capitalize()

skoudoro (changes requested):
Hi @maharshi-gor,

Overall looks good. See the small comments below

iprajwalreddy:
@maharshi-gor  Thank you for the colormaps. All seems to be working for me when I tested it out.

dipy/viz/skyline/render/image.py
maharshi-gor: why `self`? and why not `actor.` 

Just to make sure that all the interpolation based changes goes through the same function ==> single source of truth. No magics that one flow break other works.

why calling function 20-100 times?

We need to understand how the structure is working.
If you move between direction you are required to recreated the actor. That is why you see it.
Also we need to understand how imgui works. It is stateless so it will recalculates everything that on everyrun which is on constant loop.
so you might see it 2-3 times when you change the colormaps.

@skoudoro  Let me know if that answers the questions.

dipy/viz/skyline/render/image.py
maharshi-gor: Yes _apply_colormap has to be called when the create actor is going to be called. As it is a mandatory step in the creation of the actor with all the attributes accounted for from the control panel.

self.value_range will surely be set. I can set default value (-1, 1) if you want. But rest assured it is necessary step.

maharshi-gor:
Thank you for your review @skoudoro Please check if all the explanation looks good. I can fix the rest of the stuff.

I think I should remove the warning as it annoys the hell out of the person.

skoudoro (commented):
Thanks for the explanation. LGTM , merging

PR #3962, opened 2026-04-08
FIX: Tissue classification of heavily masked data

### LINKED ISSUE #3943: `TissueClassifierHMRF` with heavily masked data
### Summary

I've been using `TissueClassifierHMRF` to segment the tissues within kidneys and encountered a bug. 

Once the kidneys are masked, they occupy relatively little of the field of view, as such when [`seg_stats` is called](https://github.com/dipy/dipy/blob/2bd1ed5b2366e2c289784c3924cbe712adf16828/dipy/segment/tissue.py#L84)  it often returns zero which results in NaNs/infs propagating through the optimisation process and no masks being produced. My solution to this has been to set `sigmasq` and `sigmasq_upd` to a minimum of 1e-6 after its been calculated which seems to fix things and result in nice tissue masks.

I'm happy to work on this in a PR but thought it was worth opening an issue first to check maintainers are happy with my solution before I invest time in this (especially as this would be my first contribution to dipy). Happy to provide more information if that's helpful!

### Platform

Windows 11

### DIPY version

1.11.0

### Environment

_No response_

### Python version

3.12.12

### Additional information

_No response_

alexdaniel654:
## Description

There's a bug where the HMRF tissue classification doesn't converge if the data is heavily masked.

## Motivation and Context

This is linked to the issue raised in #3943.

## How Has This Been Tested?

I've added a test that masks out more of the sample data then tries to segment the tissue. Without my fix, this test fails; with the fix, the test passes. The new lines are covered by the tests.

## Checklist

## Type of Change

- [ ] Documentation update

### DIFF
--- a/dipy/segment/tissue.py
+++ b/dipy/segment/tissue.py
@@ -26,7 +26,8 @@ class TissueClassifierHMRF:
         self.verbose = verbose
 
     @warning_for_keywords()
-    def classify(self, image, nclasses, beta, *, tolerance=1e-05, max_iter=100):
+    def classify(self, image, nclasses, beta, *, tolerance=1e-05, max_iter=100,
+                 min_var=1e-6):
         """
         This method uses the Maximum a posteriori - Markov Random Field
         approach for segmentation by using the Iterative Conditional Modes
@@ -54,6 +55,9 @@ class TissueClassifierHMRF:
             explicitly set to 0, this early stopping mechanism is disabled,
             and the algorithm will run for the specified number of
             iterations unless another stopping criterion is met.
+        min_var : float, optional
+            Minimum variance within each tissue class to prevent division by 
+            zero. Default is 1e-6.
 
         Returns
         -------
@@ -81,9 +85,8 @@ class TissueClassifierHMRF:
         neglogl = com.negloglikelihood(image, mu, sigmasq, nclasses)
         seg_init = icm.initialize_maximum_likelihood(neglogl)
 
-        min_sigmasq = 1e-6
         mu, sigmasq = com.seg_stats(image, seg_init, nclasses)
-        sigmasq = np.maximum(sigmasq, min_sigmasq)
+        sigmasq = np.maximum(sigmasq, min_var)
 
         zero = np.zeros_like(image) + 0.001
         zero_noise = add_noise(zero, 10000, 1, noise_type="gaussian")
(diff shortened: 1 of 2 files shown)


dipy/segment/tissue.py
```diff
@@ -81,7 +81,9 @@ def classify(self, image, nclasses, beta, *, tolerance=1e-05, max_iter=100):
         neglogl = com.negloglikelihood(image, mu, sigmasq, nclasses)
         seg_init = icm.initialize_maximum_likelihood(neglogl)
 
+        min_sigmasq = 1e-6
```
skoudoro: I think, it can be a function argument, to avoid a magic number

dipy/segment/tissue.py
```diff
@@ -81,7 +81,9 @@ def classify(self, image, nclasses, beta, *, tolerance=1e-05, max_iter=100):
...
         mu, sigmasq = com.seg_stats(image, seg_init, nclasses)
+        sigmasq = np.maximum(sigmasq, min_sigmasq)
```
skoudoro: To be consistent, and make sure that all previous functions use the same value, I would move that up.

skoudoro (changes requested):
Hi @alexdaniel654 ,

Thank you so much for doing this contribution. 

Overall, looks good but see below some comments. 

Thanks for the future update

alexdaniel654:
Thanks @skoudoro. I've made `min_sigmasq` into an input argument for the `classify` method.
There's a bit of a discrepancy in nomenclature. The `seg_stats` method refers to variance whereas the `classify` method uses standard deviation squared (`sigmasq`). While mathematically they're the same, I think variance is clearer as a variable name so have followed that convention for the input argument (named it `min_var` rather than `min_sigmasq`) as I think its easier for end users. 
Looking at the commit history (47274eacf7e2e7bc74cb7ef9bc9853b565141dcb), I think the `classify` method probably started off by calculating the standard deviation then squaring it which makes sense to call the variable `sigmasq` as at one point there was probably a `sigma`. It looks like there was a refactor where the variance was calculated directly but the variable stayed as `sigmasq`. I can change all mentions of `sigmasq` to `var` but am aware that's a bit beyond the scope of my original issue/PR.

skoudoro:
Thank you for the update and the feedback @alexdaniel654!

I agree with your observation, So let's just do it here and update PR title and description.

Thank you in advance for the update and sorry for the late reply

alexdaniel654:
No problem, that's changed now.

alexdaniel654:
Are you happy with the changes @skoudoro?

skoudoro:
> Are you happy with the changes @skoudoro?

yes all good @alexdaniel654. I just push a commit to fix the styling issue. 

After CI's result, It will be ready to be merged.

Thanks for this work

PR #3975, opened 2026-04-13
 Generalizing localpca to be compatible with complex data

catarinagfreitas:
Hi DIPY team!
I am Catarina and I am working under the supervision of @RafaelNH. I am exploring the preprocessing pipeline for CTI and currently I'm focusing on denoising.

In this PR, I am generalizing the localpca implementation to be compatible with complex data. I have added logic blocks to handle complex-valued inputs and expanded the underlying matrix theory to ensure the denoising works correctly in the universal spectrum (for both real and complex data).

The implementation is currently working well, and I have addeda few unit tests for complex data to verify the denoising of random complex noise, RMSE improvement, and the preservation of the signal phase. Please let me know if u have any comments or suggestions!

### DIFF
--- a/dipy/denoise/localpca.py
+++ b/dipy/denoise/localpca.py
@@ -253,8 +253,8 @@ def genpca(
     # We retain float64 precision, iff the input is in this precision:
     if arr.dtype == np.float64:
         calc_dtype = np.float64
-    elif arr.dtype == np.complex128:
-        calc_dtype = np.complex128
+    elif np.issubdtype(arr.dtype, np.complexfloating):
+        calc_dtype = arr.dtype
     # Otherwise, we'll calculate things in float32 (saving memory)
     else:
         calc_dtype = np.float32
@@ -335,7 +335,7 @@ def genpca(
 
                 if is_svd:
                     # PCA using an SVD
-                    if calc_dtype == np.complex128:
+                    if np.issubdtype(calc_dtype, np.complexfloating):
                         U, S, Vt = np.linalg.svd(X, full_matrices=False)
                     else:
                         svd_args = [1, 0]
@@ -382,7 +382,7 @@ def genpca(
 
     denoised_arr = thetax / theta
 
-    if calc_dtype != np.complex128:
+    if not np.issubdtype(calc_dtype, np.complexfloating):
         denoised_arr.clip(min=0, out=denoised_arr)
 
     denoised_arr[mask == 0] = 0
--- a/dipy/denoise/localpca.py
+++ b/dipy/denoise/localpca.py
@@ -8,7 +8,6 @@ from scipy.linalg.lapack import dgesvd as svd
 from dipy.denoise.pca_noise_estimate import pca_noise_estimate
 from dipy.testing.decorators import warning_for_keywords
 
-
 def dimensionality_problem_message(arr, num_samples, spr):
     """Message about the number of samples being smaller than one less the
     dimensionality of the data to be denoised.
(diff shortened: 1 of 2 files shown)


dipy/denoise/localpca.py
```diff
@@ -104,7 +103,7 @@ def create_patch_radius_arr(arr, pr):
 
     if isinstance(patch_radius, int):
         patch_radius = np.ones(3, dtype=int) * patch_radius
-    if len(patch_radius) != 3:
+    if len(patch_radius) != 3:
```
arokem:
```suggestion
    if len(patch_radius) != 3:
```

arokem (commented):
Looks good to me!

dipy/denoise/localpca.py
```diff
@@ -333,24 +335,28 @@ def genpca(
 
                 if is_svd:
                     # PCA using an SVD
-                    svd_args = [1, 0]
-                    U, S, Vt = svd(X, *svd_args)[:3]
-                    # Items in S are the eigenvalues, but in ascending order
-                    # We invert the order (=> descending), square and normalize
+                    if calc_dtype == np.complex128:
```
skoudoro: Why only complex128 and not the other cases?

Image

dipy/denoise/localpca.py
catarinagfreitas: Thank you for the feedback!

dipy/denoise/localpca.py
catarinagfreitas: Thanks for the suggestion, @skoudoro! I used np.complex128 as a placeholder, but you're right that it would fail for complex64 or other precisions. I've updated the code to use np.issubdtype(..., np.complexfloating) throughout localpca.py .

RafaelNH:
@skoudoro - thanks for spotting that this was not completely general for other complex types.
Apart from that, I agree with @arokem - this looks good!
@catarinagfreitas - nice job!

Thanks everyone!

I am removing the "WIP" tab.

dipy/denoise/localpca.py
```diff
@@ -333,24 +335,28 @@ def genpca(
 
                 if is_svd:
                     # PCA using an SVD
-                    svd_args = [1, 0]
-                    U, S, Vt = svd(X, *svd_args)[:3]
-                    # Items in S are the eigenvalues, but in ascending order
-                    # We invert the order (=> descending), square and normalize
+                    if np.issubdtype(calc_dtype, np.complexfloating):
+                        U, S, Vt = np.linalg.svd(X, full_matrices=False)
```
skoudoro, line 339: Why you use the numpy `svd` and not the scipy `svd`? is there a difference ? 

Now it is strange to use 2 differents svd depending on the datatypes. Can you clarify that?

we use the scipy version for speed (faster)

I believe this can be simplified, to simplify maintenance 😅

skoudoro (changes requested):
Hi @catarinagfreitas and @RafaelNH ,

I have 2 last comments, then we can go ahead and merge. 
- The first one is below.
- the second is about the docstring. Can you add this info in `genpca` and `localpca` that we support complex data ? Also, we say that `clipped to non-negative values` which is now only true for real data. So this needs to be precised. 

thanks again for this work!

dipy/denoise/localpca.py
catarinagfreitas: Thank you @skoudoro for the final comments! I have updated the PR to address these points.

I have consolidated the SVD implementation to use direct LAPACK drivers from scipy.linalg.lapack. Specifically, I am now using dgesvd for real data and zgesvd for complex data; this ensures phase information is preserved while maintaining the performance standards of the existing implementation.

Additionally, I have revised the docstrings for genpca, localpca and mppca to explicitly state that complex data is now supported and that output clipping to non-negative values is only applied to real-valued data.

I have also included a new unit test to verify the complex-valued denoising path and ensure long-term stability.

skoudoro (approved):
Thank you for this work @catarinagfreitas. 

Merging

PR #3980, opened 2026-04-14
Modularize CTI non-linear fit

marianaabv:
Hello DIPY team!

Continuing my work under the supervision of @RafaelNH on constrained fitting strategies, this PR focuses on the modularization of the CTI.

The goal is to integrate standard fitting procedures into CTI. By leveraging the core tensor engine, the work I previously implemented for the Cholesky parameterization (PR #3707)will now be available for CTI as well.

Transitioning to a stable NLLS framework is a key step in enhancing the robustness of CTI parameter estimation.

Please let me know if you have any comments or suggestions :)


dipy/reconst/dti.py
```diff
@@ -1910,8 +1906,19 @@ def nlls_fit_tensor(
             model_S0[vox] = np.exp(-this_param[-1])
         if not dti:
             md2 = evals.mean(0) ** 2
-            params[vox, 12:] = this_param[6:-1] / md2
-
+            if md2 > 0:
+                if npa == 27:  # DKI (12 DTI + 15 DKI)
+                    params[vox, 12:] = this_param[6:-1] / md2
+                elif npa == 48:  # CTI (12 DTI + 15 DKI + 21 CTI)
```
arokem, line 1916: Any chance we could add a test for this?

dipy/reconst/cti.py
```diff
@@ -502,7 +517,7 @@ def K_total(self):
                     + D[..., 2, 2] ** 2
                     + 2 * D[..., 0, 1] ** 2
                     + 2 * D[..., 0, 2] ** 2
-                    + D[..., 1, 2] ** 2
+                    + 2 * D[..., 1, 2] ** 2
```
arokem, line 520: Was this wrong before?

dipy/reconst/cti.py
```diff
@@ -269,15 +270,29 @@ def fit(self, data, *, mask=None):
...
-        )
+
+        if self.fit_method is nlls_fit_tensor:
+            result = self.fit_method(
+                self.design_matrix,
+                data_thres,
+                *self.args,
+                **self.kwargs,
+            )
+            if isinstance(result,tuple):
```
arokem:
```suggestion
            if isinstance(result, tuple):
```

dipy/reconst/dti.py
marianaabv: Yes! I have added a unit test in _test_dti.py_ to ensure that _tensor fit_ correctly handles models with 48 parameters (CTI), verifying that no regressions are introduced for the standard DTI metrics

RafaelNH: No sure if we need that test in test_dti.py. This line of code will be covered in the test that was already added in test_cti.py. @arokem - what do you think? If the test_cti.py is already covering these lines - should we add an extra check here?

dipy/reconst/tests/test_dti.py
```diff
@@ -1224,3 +1224,30 @@ def test_quantize_evecs_parallel_engines():
         except Exception as e:
             # If an engine fails, that's okay - just skip it
             print(f"Warning: Could not test engine {engine}: {e}")
+
+
+def test_cti_params_in_dti_fit():
```
RafaelNH, line 1229: In addition to my comment about not having tests in test_dti.py about CTI, I really don't understand the logic behind this proposed test. 

Please remove this test for now (I think it is really incorrect). If at the end we decide to add some test about a number of parameters different of 7 at dti, we have to think of something different.

dipy/reconst/cti.py
```diff
@@ -269,15 +270,29 @@ def fit(self, data, *, mask=None):
...
-        )
+
+        if self.fit_method is nlls_fit_tensor:
+            result = self.fit_method(
+                self.design_matrix,
+                data_thres,
+                *self.args,
+                **self.kwargs,
+            )
+            if isinstance(result, tuple):
```
RafaelNH, line 283: Why is this line of code required?

Can you give an example where you found that the results was an instance?

dipy/reconst/cti.py
RafaelNH: Yes it was :(. We did not detected that before because our two previous tests used simulations of random oriented structures in which D[..., 1, 2]=0.

@marianaabv - to improve the quality of CTI tests - add an test corresponding to a well aligned structure - for that you can use the following ground truth CTI parameters:

[1.075      0.3255     0.3255     1.         0.         0.
 0.         0.         1.         0.         1.         0.
 1.59927944 1.66389033 1.66389033 0.         0.         0.
 0.         0.         0.         0.14829682 0.14829682 0.55463011
 0.         0.         0.         0.013125   0.18358725 0.18358725
 0.18358725 0.0490875  0.0490875  0.         0.         0.
 0.         0.         0.         0.         0.         0.
 0.         0.         0.         0.         0.         0.        ] 

Grouth truth values computed from metrics definition should be:

Kiso_gt = 0.5319401049184376
Kaniso_gt = 0.9504530935095312
Kmicro_gt = 0.2960649543684015
Ktotal_gt = 1.7784581527963703

RafaelNH (changes requested):
@marianaabv - thanks for this PR. I am submitting my first round of review for this.

dipy/reconst/tests/test_dti.py
marianaabv: Understood. I have removed the proposed test from test_dti.py. If we decide to implement generic parameter-length checks in the future, we can discuss a different approach.

skoudoro:
Do you have an update on this @marianaabv ?

dipy/reconst/cti.py
marianaabv: Thanks @arokem and @RafaelNH for the feedback. 
I've added the regression test `test_cti_metrics_fixed` using the provided ground truth values to ensure this remains fixed

dipy/reconst/dti.py
marianaabv: Hi @arokem and @RafaelNH,
Since we've just added the regression test for CTI metrics and we already have `test_cti_fits `which uses the NLS method, this specific logic for the 48-parameter case I think is already being covered. 

Given this, do you still think a separate unit test in `test_dti.py` is necessary, or can we consider this covered by the CTI suite?

dipy/reconst/cti.py
marianaabv: Hi @RafaelNH,
I think this line is required because `nlls_fit_tensor` returns a tuple containing both the parameters and the codirection when fitting a single voxel. However, when the model's fit method is called on a 4D data array, the `@multi_voxel_fit` decorator handles the output differently and might return only the parameters array

An example where this is necessary is when calling `cti_model.fit(data)` on a single voxel: without this check, CorrelationTensorFit would receive the full tuple instead of just the parameter array, causing the initialization to fail

marianaabv:
> Do you have an update on this @marianaabv ?
Hi @skoudoro,
Yes, I have just pushed the requested updates. Everything is now finalized and ready for review. Thanks for following up!

dipy/reconst/tests/test_cti.py
```diff
@@ -410,6 +426,37 @@ def test_cti_design_matrix():
     A1 = design_matrix(gtab1, gtab2)
     A2 = design_matrix(gtab2, gtab1)
     # Check if the two matrices are the same
-    assert np.allclose(A1, A2), (
-        "The design matrices are not symmetric for different gradientdirections order."
-    )
+    assert np.allclose(
+        A1, A2
```
RafaelNH: Revert code lines 429-431 to their previous version.

dipy/reconst/tests/test_cti.py
```diff
@@ -410,6 +426,37 @@ def test_cti_design_matrix():
...
+    ), "The design matrices are not symmetric for different gradientdirections order."
+
+
+def test_cti_metrics():
+    """Test CTI metrics with a well-aligned structure to verify formula.
```
RafaelNH: Move these lines of code to be inside "test_cti_fit" - no need to create a different function if the objective is also to check the metrics at different conditions.

RafaelNH (commented):
Here are my last couple of comments - after addressing these I don't have anything to add, so at my side this can be merged. Make sure that this branch is rebased.

marianaabv (commented):
I made all the changes requested. It's ready for review

dipy/reconst/tests/test_dti.py
RafaelNH: Please remove this function in test_dti, since we no longer want to test CTI here.

dipy/reconst/tests/test_dti.py
```diff
@@ -1224,3 +1224,30 @@ def test_quantize_evecs_parallel_engines():
...
+    params[..., 0] = 1.5e-3
+    params[..., 1] = 0.5e-3
+    params[..., 2] = 0.5e-3
+    params[..., 3] = 1
+    params[..., 7] = 1
+    params[..., 11] = 1
+
+    fit = dti.TensorFit(dti.TensorModel(gtab), params)
+
+    # 4. Verificar se o DTI extrai as métricas básicas corretamente
```
RafaelNH: All comments have to be written in english. Just remove this function in test_dti.py and all will be sorted.

RafaelNH (commented):
Hi @marianaabv! There is still one minor alteration that is required. Removing the alterations done in test_dti.py, specifically the function test_cti_params_in_dti_fit.

Please do this as soon as possible. After that, I recommend that this PR is merged as soon as possible as it includes a important bug fix in cti.py - the proper calculations of total and microscopic kurtosis.

dipy/reconst/tests/test_dti.py
marianaabv: Done, removed the function. Should be good to merge now

marianaabv:
Thank you @arokem and @RafaelNH for your reviews and feedback!

I opened #4170 to replace this PR after having difficulty updating the commit messages on the original branch. The replacement PR carries over the changes made in response to your reviews and links back here for the earlier review discussions.

I’m closing this PR so that the work can continue in #4170

PR #3984, opened 2026-04-17
BF: Colorpicker applies color after closing the dialog.

### LINKED ISSUE #3983: DIPY Skyline colorpicker should only update color when the dialog box is closed.
### Summary

The colorpicker currently changes the color immediately which introduces huge performance load for large visualizations like tractograms. 

The colorpicker should update only when the dialog is closed.

### Platform

all

### DIPY version

1.13,0.dev

### Environment

_No response_

### Python version

all

### Additional information

_No response_

maharshi-gor:
## Description

This PR tries to make the colorpicker efficient. It forces the color to apply only when it closes the picker dialog. It is helpful if a large data/tractogram is visualized. Saving multiple runs of color change.

## Motivation and Context

- We are tracking a new variable called `is_open` for the colorpicker that will be consumed by the downstream api to determine whether to apply the color.
- The downstream APIs will be keeping a variable called `_draft_color` which will help us track the selected changes while the dialog is open. It will also have a variable called `_color_picker_open` to which will track the state on every tick of the imgui rendering. 
- when the `_color_picker_open` is `False` and `is_open` from colorpicker is `True` ==> Dialog is opened.  ==> Start tracking the draft color.
- when the `_color_picker_open` is `True` and `is_open` from colorpicker is `False` ==> Dialog is closed. ==> Apply the latest draft color.
Closes #3983 

## How Has This Been Tested?

- Load a ROI (brain_mask), surface or tractogram and choose colorpicker.
- Change the color in the dialog.
- It should not apply the color.
- Now, close the dialog.
- The last selected color should be applied.
- The same should be applicable for background color.

## Checklist

## Type of Change

- [ ] Documentation update

### DIFF
--- a/dipy/viz/skyline/UI/elements.py
+++ b/dipy/viz/skyline/UI/elements.py
@@ -3,6 +3,8 @@
 import math
 from pathlib import Path
 
+import numpy as np
+
 from dipy.utils.logging import logger
 from dipy.utils.optpkg import optional_package
 from dipy.viz.skyline.UI.theme import (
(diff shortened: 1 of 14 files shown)


pjsjongsung:
I confirm the function works as expected. However, there seems to be several unrelated changes added with the PR. Do we want to add description about them here as well?

maharshi-gor:
> I confirm the function works as expected. However, there seems to be several unrelated changes added with the PR. Do we want to add description about them here as well?

Thanks I should mention them. Those are going together as they are touching the same functionality.

dipy/viz/skyline/UI/elements.py
```diff
@@ -23,10 +25,53 @@
...
+    color_a_arr = np.asarray(color_a)
+    color_b_arr = np.asarray(color_b)
+    if color_a_arr.ndim != 1 or color_b_arr.ndim != 1:
+        return False
+
+    return np.array_equal(color_a_arr[:3], color_b_arr[:3])
+
+
+def normalize_picker_color(color, *, fallback=(1.0, 0.0, 0.0)):
+    """Return an RGB tuple suitable for ImGui color picker widgets."""
```
skoudoro: Can you add `Parameters`and `returns` in the docstring

dipy/viz/skyline/render/streamline.py
```diff
@@ -552,20 +556,27 @@ def render_widgets(self):
...
-        changed, new_color = color_picker(
-            selected_color=self._color_picker,
+        selected_color = normalize_picker_color(self._draft_color)
+        changed, new_color, is_open = color_picker(
+            selected_color=selected_color,
             tooltip="Pick a color for the streamlines.",
+            popup_id=self._color_picker_popup_id,
         )
+        if is_open and not self._color_picker_open:
+            self._draft_color = normalize_picker_color(self._original_color)
```
skoudoro: this might be a small issue: 

If an user previously changed the streamline color, then opens the picker again, the draft resets to the original color rather than the current color. So, the picker show the wrong starting color 

ok it is just inconvenient not a blocking issue for merging

skoudoro (changes requested):
Overall looks good @maharshi-gor,

See the few comments below. 

Also,  After the picker closes, `roi.py` and `surface.py` reset `_draft_color` but `streamline.py`do not do it. maybe something to fix.

dipy/viz/skyline/render/streamline.py
maharshi-gor: I understand but I really need to find a way to store previous state in stateless UI.

maharshi-gor:
Hello @skoudoro 

Thank you so much for the review.
I have solved the issues you mentioned and should be able to work as expected.

skoudoro:
Thank you @maharshi-gor, merging

PR #3991, opened 2026-04-21
BUG: Decouple VTK/FURY optional imports and add dependency guards in vtk.py

Kaushalt2004:
Fixes #3932 — Optional dependency coupling in vtk.py causes NameError when FURY is unavailable.
 Changes
This PR makes three targeted changes to dipy/io/vtk.py (33 insertions, 1 deletion):
1. Decouple VTK and FURY optional imports
Previously, vtk, vtk.util.numpy_support, and DATATYPE_DICT were all defined inside if have_fury:, coupling VTK availability to FURY. They are now imported independently via a separate optional_package("vtk") call:
vtk, have_vtk, _ = optional_package("vtk")

### DIFF
--- a/dipy/io/vtk.py
+++ b/dipy/io/vtk.py
@@ -8,9 +8,13 @@ fury, have_fury, setup_module = optional_package(
     "fury", min_version="0.10.0", max_version="1.0.0"
 )
 
+vtk, have_vtk, _ = optional_package("vtk")
+
 if have_fury:
     import fury.io
     import fury.utils
+
+if have_vtk:
     import vtk
     import vtk.util.numpy_support as ns
 


dipy/io/vtk.py
```diff
@@ -28,6 +33,22 @@
...
+            f"{func_name} requires FURY. Please install it with: pip install fury"
+        )
+
+
+def _require_vtk(func_name):
+    """Raise ImportError if VTK is not available."""
+    if not have_vtk:
+        raise ImportError(
+            f"{func_name} requires VTK. Please install it with: pip install vtk"
+        )
```
skoudoro: this is not needed since `optional_package` does it already. Please, check how `optional_package` works

dipy/io/vtk.py
```diff
@@ -69,9 +91,11 @@ def save_polydata(
         Use legacy VTK file format. Only applied when fury >= 2.0
         is installed.
     """
+    _require_fury("save_polydata")
     # use kwargs for backward compatibility with fury < 2.0
     kwargs = {}
-    if fury.__version__.split(".")[0] >= "2":
+    fury_version = getattr(fury, "__version__", "0.0.0")
```
skoudoro: `fury.__version__` is enough. Not sure why so much complexity ?

skoudoro (changes requested):
Hi @Kaushalt2004,

I understand the idea which is correct, but the implementation is totally wrong and redundant. `ImportError`  will be called in all those cases without your code. there is a mis understanding of `optional_package`.

Also, we can see a big decrease of code coverage. 

Looking forward for your update.

Thanks !

dipy/io/vtk.py
Kaushalt2004: Thanks @skoudoro, you're right — I've removed the _require_fury / _require_vtk guards since optional_package already returns a TripWire that raises on access. The updated PR is now just 4 lines (+1 optional_package("vtk") call + splitting the if have_fury: block into separate if have_fury: and if have_vtk: blocks). No redundant guards, no version complexity, no test.yml changes.

dipy/io/vtk.py
Kaushalt2004: Agreed, reverted to the original fury.__version__ usage. The getattr / packaging.version complexity was unnecessary since optional_package ensures fury is either a real module (with __version__) or a TripWire (which would error before reaching that line). Fixed in the latest push.

Kaushalt2004:
Thanks for the review @skoudoro! I've updated the PR:

Removed _require_fury / _require_vtk guards — you're right, optional_package already handles this via TripWire.
Reverted fury.__version__ to the original — no need for getattr / packaging.version complexity.
Reverted test.yml changes — those were unrelated.
The PR is now a minimal 4-line change: just decoupling the VTK imports from the if have_fury: block into their own if have_vtk: block with a separate optional_package("vtk") call. CI re-triggered.

skoudoro:
all good,  Thanks @Kaushalt2004,

merging

PR #4000, opened 2026-04-29
TEST: speed up RUMBA tests

skoudoro:
This PR speed up RUMBA tests (800s -> 15s) by reducing the number of processed voxels during the tests (600 voxels -> 16voxels).

The tests were too long and I do not things it is necessary to tests with so many voxels in anycase.


maharshi-gor (approved):
Thank you @skoudoro for the reduction it should be helpful on making CIs faster.

Issue #4002, opened 2026-05-01
dipy_info CLI to summarize peaks.pam5 file


Garyfallidis:
### Summary

Hi team!,

We need to allow dipy_info to summarize the contents of pam5 files.

Thank you,
Elef

### DIPY version

dipy 1.12.1

Atharva-Shah-2298:
Already addressed in #4003 . Closing

PR #4003, opened 2026-05-01
NF: Added PAM info in dipy_info CLI

deka27:
## Description

Adds `.pam5` (PeaksAndMetrics HDF5) support to the `dipy_info` CLI so users can quickly summarize the full contents of a peaks file without writing a script. The new dispatch branch in `IoInfoFlow.run` loads the file via `load_pam`, reads the format version directly from the HDF5 attributes, and prints a structured summary covering volume geometry, peak statistics, SH metadata, and all auxiliary metrics (GFA, QA, ODF, total weight, angular threshold, affine).

Reported fields:
- PAM5 file format version
- Volume dimensions, voxel size, voxel order (derived from the affine)
- Peaks-per-voxel (max), peak coverage (fraction of voxels with ≥1 non-zero peak), and mean peaks per non-empty voxel
- Shapes of `peak_dirs`, `peak_values`, `peak_indices`
- Sphere vertex count
- Peak-value stats (min/max/median/mean/std)
- SH coefficient shape and derived SH order (via `order_from_ncoef`)
- B matrix shape
- GFA shape + stats
- QA shape + stats
- ODF shape
- `total_weight`, `ang_thr`
- Full affine matrix

## Motivation and Context

Issue - #4002 

`dipy_info` summarizes NIfTI, bval, bvec, and tractography files. This PR adds `.pam5` to the supported types so users can inspect a PAM file's contents directly from the CLI.

## How Has This Been Tested?

- Extended `test_io_info` in `dipy/workflows/tests/test_io.py` with a `.pam5` case: builds a `PeaksAndMetrics` via the existing `generate_random_pam()` helper, writes it to a `TemporaryDirectory` with `save_pam`, and runs `IoInfoFlow` on the resulting file.
- Test command:
```bash
pytest dipy/workflows/tests/test_io.py::test_io_info -xvs
```
- Manually verified on a `.pam5` file produced by a CSD reconstruction on the Stanford HARDI dataset. All reported fields matched expected values.
image

## Checklist

## Type of Change

- [ ] Documentation update

### DIFF
--- a/dipy/workflows/io.py
+++ b/dipy/workflows/io.py
@@ -8,6 +8,7 @@ import shutil
 import sys
 import warnings
 
+import nibabel as nib
 import numpy as np
 import trx.trx_file_memmap as tmm
 
@@ -29,7 +30,7 @@ from dipy.io.peaks import (
 )
 from dipy.io.streamline import load_tractogram, save_tractogram
 from dipy.io.utils import split_filename_extension
-from dipy.reconst.shm import convert_sh_descoteaux_tournier
+from dipy.reconst.shm import convert_sh_descoteaux_tournier, order_from_ncoef
 from dipy.reconst.utils import convert_tensors
 from dipy.tracking.streamlinespeed import length
 from dipy.utils.logging import logger
@@ -99,6 +100,31 @@ class TractographyPropertyName(enum.Enum):
     ORIGIN = "Origin"
     SPACE = "Space"
     STEP_SIZE = "Step size (mm)"
+
+
+class PamPropertyName(enum.Enum):
+    """PAM5 (PeaksAndMetrics) data property names."""
+
+    VERSION = "PAM5 version"
+    DIMENSIONS = "Volume dimensions"
+    VOXEL_SIZE = "Voxel size"
+    VOXEL_ORDER = "Voxel order"
+    NUM_PEAKS = "Peaks per voxel (max)"
+    PEAK_COVERAGE = "Peak coverage (fraction)"
+    MEAN_PEAKS = "Mean peaks per non-empty voxel"
+    PEAK_DIRS_SHAPE = "Peak dirs shape"
+    PEAK_VALUES_SHAPE = "Peak values shape"
+    PEAK_INDICES_SHAPE = "Peak indices shape"
+    SPHERE_VERTICES = "Sphere vertices"
+    SHM_COEFF_SHAPE = "SH coefficients shape"
+    SH_ORDER = "SH order"
+    B_SHAPE = "B matrix shape"
+    GFA_SHAPE = "GFA shape"
+    QA_SHAPE = "QA shape"
+    ODF_SHAPE = "ODF shape"
+    TOTAL_WEIGHT = "Total weight"
+    ANG_THR = "Angular threshold"
+    AFFINE = "Affine matrix"
 
 
 def _print_property_information(prop, val, alignment_space):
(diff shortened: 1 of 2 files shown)


dipy/workflows/io.py
```diff
@@ -352,6 +378,135 @@ def _print_tractography_information(
...
+            alignment_space,
+        )
+        try:
+            sh_order = order_from_ncoef(pam.shm_coeff.shape[-1])
+            _print_property_information(
+                PamPropertyName.SH_ORDER.value, sh_order, alignment_space
+            )
+        except Exception:
+            pass
+    if pam.B is not None:
```
skoudoro: for all of them, I recommend `hasattribute(pam, 'B') and pam.B is not None`.

Sometimes, the attribute is missing

dipy/workflows/io.py
```diff
@@ -561,6 +717,20 @@ def run(
                     tab,
                 )
 
+            if extension == ".pam5":
+                pam_property_length = (
+                    len(max([item.value for item in PamPropertyName], key=len)) + 2
+                )
+                with h5py.File(input_path, "r") as f:
+                    version = f.attrs["version"]
+                pam = load_pam(input_path)
```
skoudoro: Why you open it twice ? `load_pam` has already a version verification .

dipy/workflows/io.py
```diff
@@ -352,6 +378,135 @@ def _print_tractography_information(
...
+            PamPropertyName.SHM_COEFF_SHAPE.value,
+            pam.shm_coeff.shape,
+            alignment_space,
+        )
+        try:
+            sh_order = order_from_ncoef(pam.shm_coeff.shape[-1])
+            _print_property_information(
+                PamPropertyName.SH_ORDER.value, sh_order, alignment_space
+            )
+        except Exception:
```
skoudoro: use `ValueError` instead on the general `Exception`. 
Also, log a warning for the user. Now, it is too silent

dipy/workflows/io.py
```diff
@@ -8,6 +8,8 @@
 import sys
 import warnings
 
+import h5py
```
skoudoro: import not needed I believe

skoudoro (commented):
Thank you for this @deka27,

See below the review

dipy/workflows/tests/test_io.py
```diff
@@ -91,6 +91,13 @@ def test_io_info():
         reference=str(filepath_dix["gs_volume.nii"]),
     )
 
+    pam = generate_random_pam()
+    with TemporaryDirectory() as out_dir:
+        pam_fname = Path(out_dir) / "test_info.pam5"
+        save_pam(pam_fname, pam)
+        io_info_flow = IoInfoFlow()
+        io_info_flow.run(pam_fname)
```
skoudoro: what is the goal of this test if there is no assertions?

skoudoro (changes requested):
Thanks for the update @deka27. 

See below the last forgotten comment. Otherwise, all good

dipy/workflows/tests/test_io.py
deka27: My bad, missed that. Pushed the fix.

dipy/workflows/io.py
```diff
@@ -352,6 +377,138 @@ def _print_tractography_information(
...
+            sh_order = order_from_ncoef(pam.shm_coeff.shape[-1])
+            _print_property_information(
+                PamPropertyName.SH_ORDER.value, sh_order, alignment_space
+            )
+        except ValueError:
+            logger.warning(
+                f"Could not derive SH order from shm_coeff with "
+                f"{pam.shm_coeff.shape[-1]} coefficients."
+            )
+    if hasattr(pam, "B") and pam.B is not None:
```
skoudoro, line 498: After some thought, my problem with this is:

The user will never know that it is missing since with print nothing and skip. 

I believe we should print something like `B:None` or `B:Not available`

idem with the one below.

skoudoro: Can you update ?

deka27: Done. Added a test too.

skoudoro:
Hi @deka27, 

Sorry for that, can you rebase? and fix conflict? Thanks

skoudoro:
Thank you for this @deka27, merging !

Issue #3619, opened 2025-08-08
Unexpected Patch2Self behavior when number of diffusion directions is small


kvttt:
### Summary

The behavior of Patch2Self is unexpected when used to denoise a dataset consisting of a small number of diffusion directions, e.g., 20 b1000s. There are two main issues: 
1. Applying Patch2Self shifts the intensity histogram to the right by a lot.
2. I tried to fix the issue above by shifting the intensity back, i.e., making the mean of the noisy data equal the mean of the denoised data. But when I inspect the images, the DWIs no longer have direction-specific contrasts and look more like a mean DWI. When I inspect the absolute difference, there are residual structures.

On the contrary, MPPCA gives consistent performance even when the number of diffusion directions is small.

Considering that 20 b1000s is quite typical for a clinical DTI protocol, I would like to know the reason why Patch2Self has this behavior and if there is any way to address this issue. 

Here is a minimal reproducible example.

```
import matplotlib.pyplot as plt
import numpy as np

from dipy.data import get_fnames
from dipy.denoise.localpca import mppca
from dipy.denoise.patch2self import patch2self
from dipy.io.image import load_nifti, save_nifti
from dipy.core.gradients import gradient_table


fname, bval_fname, bvec_fname = get_fnames(name="sherbrooke_3shell")
data, affine = load_nifti(fname)
bvals = np.loadtxt(bval_fname)
bvecs = np.loadtxt(bvec_fname)
gtab = gradient_table(bvals=bvals, bvecs=bvecs)

# we are selecting 20 b1000s here to simulate a typical clinical protocol
data = data[..., bvals == 1000]
bvals = bvals[bvals == 1000]
ndirs = data.shape[-1]
choice = np.random.choice(ndirs, size=20, replace=False)
data = data[..., choice]
bvals = bvals[choice]

# for plotting
vmin = np.percentile(data[:, :, 30, 0].T, 0.5)
vmax = np.percentile(data[:, :, 30, 0].T, 99.5)

# denoise
mppca_arr = mppca(data, patch_radius=1)
p2s_shift_arr = patch2self(data, bvals, model="ols", shift_intensity=True)
p2s_no_shift_arr = patch2self(data, bvals, model="ols", shift_intensity=False)

# for plotting
p2s_shift_vmax = np.percentile(p2s_shift_arr.ravel(), 99.5)
p2s_no_shift_vmax = np.percentile(p2s_no_shift_arr.ravel(), 99.5)

# histogram
[...]

PR #4005, opened 2026-05-03
RF: speed up patch2self v1 via shared Gram matrix

NekoAsakura:
https://github.com/dipy/dipy/issues/3619#issuecomment-4227179753
cc: @skoudoro, @iprajwalreddy

### DIFF
--- a/dipy/denoise/patch2self.py
+++ b/dipy/denoise/patch2self.py
@@ -138,6 +138,60 @@ def _fit_denoising_model(train, vol_idx, model, alpha, *, copy_X=True):
     return model_instance, cur_x
 
 
+def _fit_denoising_models(train, model, alpha):
+    """Fit every leave-one-out 3D volume from a shared Gram matrix.
+
+    Parameters
+    ----------
+    train : ndarray
+        Array of all 3D patches flattened out to be 2D.
+    model : {'ols', 'ridge'}
+        Closed-form linear models only. ``'ols'`` is mapped to
+        ``Ridge(alpha=1e-10)`` to match ``_fit_denoising_model``.
+    alpha : float
+        Regularization parameter for ridge regression.
+
+    Returns
+    -------
+    fits : list of (coef, intercept)
+        One entry per held-out volume, in vol_idx order.
+
+    """
+    if not (isinstance(model, str) and model.lower() in ("ols", "ridge")):
+        raise ValueError(
+            f"_fit_denoising_models only supports 'ols' or 'ridge', got: {model!r}"
+        )
+    ridge_alpha = 1e-10 if model.lower() == "ols" else alpha
+
+    n_vols, n_patch, n_voxels = train.shape
+    train_flat = train.reshape(n_vols * n_patch, n_voxels)
+
+    row_means = train_flat.mean(axis=1)
+    G_cent = train_flat @ train_flat.T - n_voxels * np.outer(row_means, row_means)
+
+    fits = []
+    center_offset = n_patch // 2
+    for vol_idx in range(n_vols):
+        held_start = vol_idx * n_patch
+        held_stop = held_start + n_patch
+        target_row = held_start + center_offset
+
+        keep = np.ones(n_vols * n_patch, dtype=bool)
+        keep[held_start:held_stop] = False
+
+        G_xx = G_cent[np.ix_(keep, keep)]
+        b = G_cent[keep, target_row]
+
+        coef = np.linalg.solve(
+            G_xx + ridge_alpha * np.eye(G_xx.shape[0], dtype=G_xx.dtype),
+            b,
+        )
+        intercept = row_means[target_row] - row_means[keep] @ coef
+        fits.append((coef, float(intercept)))
+
+    return fits
+
+
 def vol_denoise(
     data_dict, b0_idx, dwi_idx, model, alpha, b0_denoising, verbose, tmp_dir
 ):
@@ -544,6 +598,8 @@ def _patch2self_version1(
         position = data.shape.index(1)
         data = np.concatenate((data, data, data), position)
 
+    use_gram_solver = isinstance(model, str) and model.lower() in ("ols", "ridge")
+
     # Segregates volumes by b0 threshold
     b0_idx = np.argwhere(bvals <= b0_threshold)
     dwi_idx = np.argwhere(bvals > b0_threshold)
@@ -581,14 +637,21 @@ def _patch2self_version1(
             patch_radius=patch_radius,
         )
 
-        for vol_idx in range(0, data_b0s.shape[3]):
-            b0_model, cur_x = _fit_denoising_model(
-                train_b0, vol_idx, model, alpha=alpha
-            )
-
-            denoised_b0s[..., vol_idx] = b0_model.predict(cur_x.T).reshape(
-                data_b0s.shape[0], data_b0s.shape[1], data_b0s.shape[2]
-            )
+        if use_gram_solver:
+            fits = _fit_denoising_models(train_b0, model, alpha)
+            for vol_idx, (coef, intercept) in enumerate(fits):
+                cur_x, _ = _vol_split(train_b0, vol_idx)
+                denoised_b0s[..., vol_idx] = (cur_x.T @ coef + intercept).reshape(
+                    data_b0s.shape[0], data_b0s.shape[1], data_b0s.shape[2]
+                )
+        else:
+            for vol_idx in range(0, data_b0s.shape[3]):
+                b0_model, cur_x = _fit_denoising_model(
+                    train_b0, vol_idx, model, alpha=alpha
+                )
+                denoised_b0s[..., vol_idx] = b0_model.predict(cur_x.T).reshape(
+                    data_b0s.shape[0], data_b0s.shape[1], data_b0s.shape[2]
+                )
 
         if verbose is True:
             logger.info(f"Denoised b0 Volume: {vol_idx}")
@@ -608,14 +671,25 @@ def _patch2self_version1(
     )
 
     # Insert the separately denoised arrays into the respective empty arrays
-    for vol_idx in range(0, data_dwi.shape[3]):
-        dwi_model, cur_x = _fit_denoising_model(train_dwi, vol_idx, model, alpha=alpha)
-        denoised_dwi[..., vol_idx] = dwi_model.predict(cur_x.T).reshape(
-            data_dwi.shape[0], data_dwi.shape[1], data_dwi.shape[2]
-        )
-
-        if verbose is True:
-            logger.info(f"Denoised DWI Volume: {vol_idx}")
+    if use_gram_solver:
+        fits = _fit_denoising_models(train_dwi, model, alpha)
+        for vol_idx, (coef, intercept) in enumerate(fits):
+            cur_x, _ = _vol_split(train_dwi, vol_idx)
+            denoised_dwi[..., vol_idx] = (cur_x.T @ coef + intercept).reshape(
+                data_dwi.shape[0], data_dwi.shape[1], data_dwi.shape[2]
+            )
+            if verbose is True:
+                logger.info(f"Denoised DWI Volume: {vol_idx}")
+    else:
+        for vol_idx in range(0, data_dwi.shape[3]):
+            dwi_model, cur_x = _fit_denoising_model(
+                train_dwi, vol_idx, model, alpha=alpha
+            )
+            denoised_dwi[..., vol_idx] = dwi_model.predict(cur_x.T).reshape(
+                data_dwi.shape[0], data_dwi.shape[1], data_dwi.shape[2]
+            )
+            if verbose is True:
+                logger.info(f"Denoised DWI Volume: {vol_idx}")
 
     if verbose is True:
         t2 = time.time()
--- a/dipy/denoise/patch2self.py
+++ b/dipy/denoise/patch2self.py
@@ -157,10 +157,6 @@ def _fit_denoising_models(train, model, alpha):
         One entry per held-out volume, in vol_idx order.
 
     """
-    if not (isinstance(model, str) and model.lower() in ("ols", "ridge")):
-        raise ValueError(
-            f"_fit_denoising_models only supports 'ols' or 'ridge', got: {model!r}"
-        )
     ridge_alpha = 1e-10 if model.lower() == "ols" else alpha
 
     n_vols, n_patch, n_voxels = train.shape
--- a/dipy/denoise/patch2self.py
+++ b/dipy/denoise/patch2self.py
@@ -167,7 +167,8 @@ def _fit_denoising_models(train, model, alpha):
     train_flat = train.reshape(n_vols * n_patch, n_voxels)
 
     row_means = train_flat.mean(axis=1)
-    G_cent = train_flat @ train_flat.T - n_voxels * np.outer(row_means, row_means)
+    train_centered = train_flat - row_means[:, None]
+    G_cent = train_centered @ train_centered.T
     reg = ridge_alpha * np.eye((n_vols - 1) * n_patch, dtype=G_cent.dtype)
 
     fits = []
--- a/dipy/denoise/patch2self.py
+++ b/dipy/denoise/patch2self.py
@@ -353,6 +353,7 @@ def patch2self(
     shift_intensity=True,
     tmp_dir=None,
     version=3,
+    gram=True,
 ):
     """Patch2Self Denoiser.
 
@@ -399,6 +400,10 @@ def patch2self(
         files are saved in the system's default temporary directory.
     version : int, optional
         Version 1 or 3 of Patch2Self to use.
+    gram : bool, optional
+        Solve leave-one-out fits from a shared Gram matrix, reducing runtime
+        and memory. If False, perform per-volume regression as in the
+        original publication. Applies only to `version=1`.
 
     Returns
     -------
@@ -429,6 +434,7 @@ def patch2self(
             b0_denoising,
             clip_negative_vals,
             shift_intensity,
+            gram,
         )
     return _patch2self_version3(
         data,
@@ -538,6 +544,7 @@ def _patch2self_version1(
     b0_denoising,
     clip_negative_vals,
     shift_intensity,
+    gram=True,
 ):
     """Patch2Self Denoiser.
 
@@ -576,6 +583,10 @@ def _patch2self_version1(
     shift_intensity : bool
         Shifts the distribution of intensities per volume to give
         non-negative values.
+    gram : bool
+        Solve leave-one-out fits from a shared Gram matrix, reducing runtime
+        and memory. If False, perform per-volume regression as in the
+        original publication. Applies only to `version=1`.
 
     Returns
     -------
@@ -597,7 +608,9 @@ def _patch2self_version1(
         position = data.shape.index(1)
         data = np.concatenate((data, data, data), position)
 
-    use_gram_solver = isinstance(model, str) and model.lower() in ("ols", "ridge")
+    use_gram_solver = (
+        gram and isinstance(model, str) and model.lower() in ("ols", "ridge")
+    )
 
     # Segregates volumes by b0 threshold
     b0_idx = np.argwhere(bvals <= b0_threshold)
--- a/dipy/denoise/patch2self.py
+++ b/dipy/denoise/patch2self.py
@@ -6,6 +6,7 @@ from warnings import warn
 
 import numpy as np
 from numpy.lib.stride_tricks import sliding_window_view
+from scipy.linalg import solve
 from tqdm import tqdm
 
 from dipy.stats.sketching import count_sketch
@@ -160,10 +161,14 @@ def _fit_denoising_models(train, model, alpha):
     ridge_alpha = 1e-10 if model.lower() == "ols" else alpha
 
     n_vols, n_patch, n_voxels = train.shape
+    if n_vols == 0:
+        return []
+
     train_flat = train.reshape(n_vols * n_patch, n_voxels)
 
     row_means = train_flat.mean(axis=1)
     G_cent = train_flat @ train_flat.T - n_voxels * np.outer(row_means, row_means)
+    reg = ridge_alpha * np.eye((n_vols - 1) * n_patch, dtype=G_cent.dtype)
 
     fits = []
     center_offset = n_patch // 2
@@ -178,10 +183,7 @@ def _fit_denoising_models(train, model, alpha):
         G_xx = G_cent[np.ix_(keep, keep)]
         b = G_cent[keep, target_row]
 
-        coef = np.linalg.solve(
-            G_xx + ridge_alpha * np.eye(G_xx.shape[0], dtype=G_xx.dtype),
-            b,
-        )
+        coef = solve(G_xx + reg, b, assume_a="pos")
         intercept = row_means[target_row] - row_means[keep] @ coef
         fits.append((coef, float(intercept)))
 
--- a/dipy/denoise/tests/test_patch2self.py
+++ b/dipy/denoise/tests/test_patch2self.py
@@ -456,6 +456,24 @@ def test_patch_radius_accepts_numpy_array():
 
 @needs_sklearn
 @set_random_number_generator(2026)
+def test_patch2self_v1_lasso_and_verbose(rng):
+    # Lasso routes through the non-gram fallback in _patch2self_version1
+    data = rng.random((5, 5, 5, 12))
+    bvals = np.concatenate([np.zeros(2), np.full(10, 1000.0)])
+
+    out_lasso = p2s.patch2self(
+        data, bvals, model="lasso", version=1, alpha=0.5, verbose=True
+    )
+    assert_equal(out_lasso.shape, data.shape)
+
+    out_ridge = p2s.patch2self(
+        data, bvals, model="ridge", version=1, alpha=1.0, verbose=True
+    )
+    assert_equal(out_ridge.shape, data.shape)
+
+
+@needs_sklearn
+@set_random_number_generator(2026)
 def test_patch2self_v3_b0_passthrough_when_b0_denoising_disabled(rng):
     # Regression test: when b0_denoising=False, the previous code copied
     # data_tmp[..., b0_counter] which indexed the wrong volume whenever the
--- a/dipy/denoise/tests/test_patch2self.py
+++ b/dipy/denoise/tests/test_patch2self.py
@@ -455,21 +455,14 @@ def test_patch_radius_accepts_numpy_array():
 
 
 @needs_sklearn
-@set_random_number_generator(2026)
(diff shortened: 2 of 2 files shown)


dipy/denoise/patch2self.py
```diff
@@ -138,6 +138,56 @@ def _fit_denoising_model(train, vol_idx, model, alpha, *, copy_X=True):
...
+        One entry per held-out volume, in vol_idx order.
+
+    """
+    ridge_alpha = 1e-10 if model.lower() == "ols" else alpha
+
+    n_vols, n_patch, n_voxels = train.shape
+    train_flat = train.reshape(n_vols * n_patch, n_voxels)
+
+    row_means = train_flat.mean(axis=1)
+    G_cent = train_flat @ train_flat.T - n_voxels * np.outer(row_means, row_means)
```
skoudoro: The biggest problem here is memory.  `G_cent shape = (n_vols * n_patch, n_vols * n_patch)` 
 With `patch_radius=3` → `n_patch=343, n_vols=60`.   => G_cent = 20580 × 20580 × 8 bytes ≈ 3.4 GB.

With classical clinical data (60 volumes, radius=2) = ~6 GB,

Overall, it make sense, faster because it use the memory but this huge increase of memory is problematic for the majority of laptop.

dipy/denoise/patch2self.py
```diff
@@ -138,6 +138,56 @@ def _fit_denoising_model(train, vol_idx, model, alpha, *, copy_X=True):
...
+
+        keep = np.ones(n_vols * n_patch, dtype=bool)
+        keep[held_start:held_stop] = False
+
+        G_xx = G_cent[np.ix_(keep, keep)]
+        b = G_cent[keep, target_row]
+
+        coef = np.linalg.solve(
+            G_xx + ridge_alpha * np.eye(G_xx.shape[0], dtype=G_xx.dtype),
+            b,
```
skoudoro: Can you add `, assume_a='pos'` to make sure we have the same stability as sklearn Ridge, So similar result.

dipy/denoise/tests/test_patch2self.py
```diff
@@ -454,6 +454,24 @@ def test_patch_radius_accepts_numpy_array():
...
+
+    out_lasso = p2s.patch2self(
+        data, bvals, model="lasso", version=1, alpha=0.5, verbose=True
+    )
+    assert_equal(out_lasso.shape, data.shape)
+
+    out_ridge = p2s.patch2self(
+        data, bvals, model="ridge", version=1, alpha=1.0, verbose=True
+    )
+    assert_equal(out_ridge.shape, data.shape)
```
skoudoro: Here, it is missing a comparaison between your version and the previous one to make sure we have similar results and make sure we have performance improvement?

dipy/denoise/patch2self.py
```diff
@@ -544,6 +594,8 @@ def _patch2self_version1(
         position = data.shape.index(1)
         data = np.concatenate((data, data, data), position)
 
+    use_gram_solver = isinstance(model, str) and model.lower() in ("ols", "ridge")
```
skoudoro: Due to the memory issue, I think we should have an option in the main function name `use_gram` explaining advantage and drawback. 

Every user will be able to choose between memory and speed.

dipy/denoise/tests/test_patch2self.py
```diff
@@ -454,6 +454,24 @@ def test_patch_radius_accepts_numpy_array():
         p2s.patch2self(data, bvals, patch_radius=np.array([1, 1]), version=1)
 
 
+@needs_sklearn
+@set_random_number_generator(2026)
+def test_patch2self_v1_lasso_and_verbose(rng):
```
skoudoro: weird name since ridge is also tested. to rename

dipy/denoise/tests/test_patch2self.py
```diff
@@ -454,6 +454,24 @@ def test_patch_radius_accepts_numpy_array():
...
+    out_lasso = p2s.patch2self(
+        data, bvals, model="lasso", version=1, alpha=0.5, verbose=True
+    )
+    assert_equal(out_lasso.shape, data.shape)
+
+    out_ridge = p2s.patch2self(
+        data, bvals, model="ridge", version=1, alpha=1.0, verbose=True
+    )
+    assert_equal(out_ridge.shape, data.shape)
+
```
skoudoro, line 466: missing test for `ols`

skoudoro (changes requested):
Hi @NekoAsakura,

Thank you for this work, See below my review. It is important to let the user choose between both version depending of its need (speed vs memory)

@iprajwalreddy Can you look and review this PR also.

dipy/denoise/patch2self.py
NekoAsakura: `G_cent` isn't net-new memory, it just consolidates one persistent matrix in place of the n_vols transient `XᵀX` allocations sklearn's solver builds on every call. The real cost is `cur_x`, and the gram path doesn't materialise it during the fit phase, so that's gone. It's two wins for gram, you can reproduce on your side.

dipy/denoise/tests/test_patch2self.py
NekoAsakura: I don't think such a compare should be a regression test in CI. I can provide the e2e test I used.
```
import time, numpy as np, nibabel as nib
from sklearn.linear_model import Ridge
from dipy.denoise.patch2self import patch2self

f, b, _ = get_fnames(name="sherbrooke_3shell")
d = nib.load(f).get_fdata()[..., :60]
b = np.loadtxt(b)[:60]

def run(m):
    t = time.time()
    o = patch2self(d, b, model=m, version=1)
    return time.time()-t, o

tg, og = run("ols")
tn, on = run(Ridge(alpha=1e-10))
print(f"shape {d.shape} | gram {tg:.1f}s | nogram {tn:.1f}s (x{tn/tg:.1f}) | diff {abs(og-on).max():.1e}")
```
Results:
```
shape (128, 128, 60, 60) | gram 22.5s | nogram 39.0s (x1.7) | diff 5.6e-11
```

dipy/denoise/patch2self.py
NekoAsakura: Use e2e scripts below:
```
shape [32, 32, 32, 20] r=3 | gram 51.7s / 7.72GB | nogram 138.7s / 8.57GB | x2.7 speed, -0.85GB RAM | diff 0.0e+00
```

iprajwalreddy:
@NekoAsakura Sorry for the delays on the review. I am currently checking the fragility regarding the way gram matrix is being computed. While the formula is mathematically valid, I want to make sure that there is not too much loss of information when this is being used.

NekoAsakura:
@iprajwalreddy No worries on the timing, thanks for taking a look.

Quick note on the math in case it helps the audit: the Gram path is an algebraically exact rewrite, not an approximation. Any difference should only come from floating-point reordering (rounding error).

Can you compare it with v3?

iprajwalreddy:
Gotcha @NekoAsakura. Yup, I am worried on the floating point reordering. 
Yup, sure I can compare it with v3 as well.

NekoAsakura:
@iprajwalreddy Any feedback or insights?

iprajwalreddy:
Hello @NekoAsakura! Sorry on all the delays. Here is what I got when I ran the comparisons:
image

We can see the issue when using float32 wherein the relative error increases to approximately 50% when the baseline is 5000. The condition number also purely inflates due to rounding off when the baseline is at 5000.

Let me know what you think.

NekoAsakura:
Thanks for running those @iprajwalreddy , really helpful.
Looking at the numbers, the Gram path now edges out v3 on both fronts: faster and lower error. Reckon we could make the case for switching the default over to v1? Happy with that? @skoudoro 
On the condition number bit, I've moved the centring step to happen before forming the Gram, which lines up with how sklearn handles it in [`_preprocess_data`](https://github.com/scikit-learn/scikit-learn/blob/1.8.0/sklearn/linear_model/_base.py#L190-L193). I think float32 precision itself was fine once that's out the way.
Any chance you could re-run your tests against this, or share the script? And if it's not too much hassle, would you be able to add the existing v1 and v3 curves to the first two plots for comparison? That'd make it proper clear where each path lands.

iprajwalreddy:
@NekoAsakura Sure thing! Here is the new plot:
image
Also, while v1 would be faster and low error than v3, it does have a higher memory requirement than v3 and hence, I wouldn't recommend it yet to be the default option. 

Also, here is the script for the plots @NekoAsakura :

```
import warnings

import matplotlib
matplotlib.use("Agg")
import matplotlib.pyplot as plt
import numpy as np
from sklearn.linear_model import Ridge

import sys
sys.path.insert(0, r"f:\dipy_pr\dipy")

import dipy.denoise.patch2self as _p2s_mod
from dipy.denoise.patch2self import patch2self

rng = np.random.default_rng(0)

def gram_current(X):
    row_means = X.mean(axis=1)
    return X @ X.T - X.shape[1] * np.outer(row_means, row_means)

def relative_error(A, B):
    return np.linalg.norm(A - B, "fro") / (np.linalg.norm(B, "fro") + 1e-30)

baselines = [0, 50, 100, 500, 1000, 2000, 5000]

from dipy.denoise.patch2self import _fit_denoising_model, _fit_denoising_models, _vol_split

N_VOLS, N_PATCH, N_VOXELS = 8, 1, 300
BVALS_BL = np.array([0, 0] + [1000] * (N_VOLS - 2), dtype=float)
V3_SHAPE  = (6, 6, 6, N_VOLS)   # tiny spatial dims so v3 finishes fast


def _predictions_gram(train):
    fits = _fit_denoising_models(train, "ols", alpha=1e-10)
    preds = []
    for vi, (coef, intercept) in enumerate(fits):
        cur_x, _ = _vol_split(train, vi)
        preds.append(cur_x.T @ coef + intercept)
    return np.stack(preds, axis=-1)


def _predictions_sklearn(train):
    preds = []
    for vi in range(train.shape[0]):
        m, cur_x = _fit_denoising_model(train, vi, "ols", alpha=1e-10)
        preds.append(m.predict(cur_x.T))
    return np.stack(preds, axis=-1)


mae_gram_f32, mae_gram_f64 = [], []
mae_sklearn_f32, mae_sklearn_f64 = [], []
mae_v3_f32, mae_v3_f64 = [], []
cond_gram_f32, cond_gram_f64 = [], []
cond_sklearn_f32, cond_sklearn_f64 = [], []

print("Computing per-baseline metrics ...")
for baseline in baselines:
    train32 = (baseline + rng.standard_normal((N_VOLS, N_PATCH, N_VOXELS)) * 5
               ).astype(np.float32)
    train64 = train32.astype(np.float64)

    ref = _predictions_sklearn(train64)

    mae_gram_f32.append(float(np.mean(np.abs(_predictions_gram(train32) - ref))))
    mae_gram_f64.append(float(np.mean(np.abs(_predictions_gram(train64) - ref))))

    mae_sklearn_f32.append(float(np.mean(np.abs(_predictions_sklearn(train32) - ref))))
    mae_sklearn_f64.append(0.0)

    data_bl32 = (baseline + rng.standard_normal(V3_SHAPE) * 5).astype(np.float32)
    data_bl64 = data_bl32.astype(np.float64)
    with warnings.catch_warnings():
        warnings.simplefilter("ignore")
        out_sk32 = patch2self(
            data_bl32, BVALS_BL, model=Ridge(alpha=1e-10, copy_X=True),
            patch_radius=(0, 0, 0), shift_intensity=False,
            clip_negative_vals=False, version=1,
        )
        out_v3_32 = patch2self(
            data_bl32, BVALS_BL, model="ols",
            shift_intensity=False, clip_negative_vals=False, version=3,
        )
        out_sk64 = patch2self(
            data_bl64, BVALS_BL, model=Ridge(alpha=1e-10, copy_X=True),
            patch_radius=(0, 0, 0), shift_intensity=False,
            clip_negative_vals=False, version=1,
        )
        out_v3_64 = patch2self(
            data_bl64, BVALS_BL, model="ols",
            shift_intensity=False, clip_negative_vals=False, version=3,
        )
    mae_v3_f32.append(float(np.mean(np.abs(out_v3_32 - out_sk32))))
    mae_v3_f64.append(float(np.mean(np.abs(out_v3_64 - out_sk64))))

    for train, cond_gram_list, cond_sk_list in [
        (train32, cond_gram_f32, cond_sklearn_f32),
        (train64, cond_gram_f64, cond_sklearn_f64),
    ]:
        tf  = train.reshape(N_VOLS * N_PATCH, N_VOXELS)
        rm  = tf.mean(axis=1)
        G_b = tf @ tf.T - N_VOXELS * np.outer(rm, rm)
        keep = np.ones(N_VOLS * N_PATCH, dtype=bool)
        keep[:N_PATCH] = False
        cond_gram_list.append(
            np.linalg.cond(G_b[np.ix_(keep, keep)].astype(np.float64))
        )
        cur_x_sub = tf[keep]
        X_train_  = cur_x_sub.T
        X_c       = X_train_ - X_train_.mean(axis=0)
        cond_sk_list.append(
            np.linalg.cond((X_c.T @ X_c).astype(np.float64))
        )

print("Done.\n")

SHAPE  = (20, 20, 20, 20)
bvals  = np.array([0, 0] + [1000] * 18, dtype=float)

xx, yy, zz = np.meshgrid(
    np.linspace(0, 1, SHAPE[0]),
    np.linspace(0, 1, SHAPE[1]),
    np.linspace(0, 1, SHAPE[2]),
    indexing="ij",
)
gt_b0  = 1000 + 20 * np.sin(np.pi * xx) * np.cos(np.pi * yy)
gt_dwi =  600 + 15 * np.cos(np.pi * yy) * np.sin(np.pi * zz)

data = np.zeros(SHAPE)
for i in range(SHAPE[-1]):
    base = gt_b0 if bvals[i] == 0 else gt_dwi
    data[..., i] = base + rng.standard_normal(SHAPE[:3]) * 5.0
data = data.astype(np.float32)

print("Running three solvers (MRI baseline ~1000, float32) ...")

kw_v1 = dict(patch_radius=(0, 0, 0), shift_intensity=False,
             clip_negative_vals=False, version=1)

with warnings.catch_warnings():
    warnings.simplefilter("ignore")

    out_gram = patch2self(data, bvals, model="ols", **kw_v1)

    out_sklearn = patch2self(
        data, bvals, model=Ridge(alpha=1e-10, copy_X=True), **kw_v1
    )

    out_v3 = patch2self(
        data, bvals, model="ols",
        shift_intensity=False, clip_negative_vals=False, version=3,
    )

print("Done.\n")

def pmae(a, b):
    return np.array([np.mean(np.abs(a[..., i] - b[..., i]))
                     for i in range(a.shape[-1])])

def pcorr(a, b):
    return np.array([np.corrcoef(a[..., i].ravel(), b[..., i].ravel())[0, 1]
                     for i in range(a.shape[-1])])

vols           = np.arange(SHAPE[-1])
mae_gram_sk    = pmae(out_gram,    out_sklearn)
mae_v3_sk      = pmae(out_v3,      out_sklearn)
corr_gram_sk   = pcorr(out_gram,   out_sklearn)
corr_v3_sk     = pcorr(out_v3,     out_sklearn)

print(f"gram    vs sklearn : mean MAE={mae_gram_sk.mean():.4f}  max={mae_gram_sk.max():.4f}  "
      f"mean r={corr_gram_sk.mean():.6f}")
print(f"v3      vs sklearn : mean MAE={mae_v3_sk.mean():.4f}  max={mae_v3_sk.max():.4f}  "
      f"mean r={corr_v3_sk.mean():.6f}")

GRAM_COL  = "crimson"
V3_COL    = "steelblue"
SK_COL    = "black"

fig = plt.figure(figsize=(18, 20))
fig.suptitle(
    "patch2self solver comparison  (float32, MRI baseline ≈ 1000)\n"
    "gram = current PR   |   sklearn = original v1   |   v3",
    fontsize=13,
)
gs = fig.add_gridspec(4, 4, hspace=0.45, wspace=0.35)

ax = fig.add_subplot(gs[0, :2])
ax.semilogy(baselines, mae_gram_f32,    "o-",  color=GRAM_COL,      label="gram    float32")
ax.semilogy(baselines, mae_gram_f64,    "o--", color=GRAM_COL,      label="gram    float64", alpha=0.5)
ax.semilogy(baselines, mae_sklearn_f32, "s-",  color=SK_COL,        label="sklearn float32")
ax.semilogy(baselines, mae_sklearn_f64, "s--", color=SK_COL,        label="sklearn float64", alpha=0.5)
ax.semilogy(baselines, mae_v3_f32,      "^-",  color=V3_COL,        label="v3      float32")
ax.semilogy(baselines, mae_v3_f64,      "^--", color=V3_COL,        label="v3      float64", alpha=0.5)
ax.set_xlabel("MRI baseline intensity")
ax.set_ylabel("Mean prediction MAE vs float64 reference")
ax.set_title("Prediction accuracy across baselines\n(solid=float32, dashed=float64)")
ax.legend(fontsize=8)
ax.grid(True, alpha=0.3)

ax = fig.add_subplot(gs[0, 2:])
ax.semilogy(baselines, cond_gram_f32,    "o-",  color=GRAM_COL, label="gram    float32")
ax.semilogy(baselines, cond_gram_f64,    "o--", color=GRAM_COL, label="gram    float64", alpha=0.5)
ax.semilogy(baselines, cond_sklearn_f32, "s-",  color=SK_COL,   label="sklearn float32")
ax.semilogy(baselines, cond_sklearn_f64, "s--", color=SK_COL,   label="sklearn float64", alpha=0.5)
ax.set_xlabel("MRI baseline intensity")
ax.set_ylabel("Condition number  κ")
ax.set_title("Condition number of effective solve matrix\n(solid=float32, dashed=float64)")
ax.legend(fontsize=8)
ax.grid(True, alpha=0.3)

ax = fig.add_subplot(gs[1, :2])
w = 0.35
ax.bar(vols - w/2, mae_gram_sk, width=w, color=GRAM_COL, alpha=0.8, label="gram vs sklearn")
ax.bar(vols + w/2, mae_v3_sk,   width=w, color=V3_COL,   alpha=0.8, label="v3   vs sklearn")
ax.axvspan(-0.5, 1.5, alpha=0.08, color="gray")
ax.set_xlabel("Volume index  (gray = b0)")
ax.set_ylabel("Mean Absolute Error")
ax.set_title("Per-volume MAE vs sklearn reference")
ax.legend(fontsize=9)
ax.grid(axis="y", alpha=0.3)

ax = fig.add_subplot(gs[1, 2:])
ax.plot(vols, corr_gram_sk, "o-",  color=GRAM_COL, label="gram vs sklearn")
ax.plot(vols, corr_v3_sk,   "s--", color=V3_COL,   label="v3   vs sklearn")
ax.axvspan(-0.5, 1.5, alpha=0.08, color="gray")
ax.set_xlabel("Volume index  (gray = b0)")
ax.set_ylabel("Pearson r")
ax.set_title("Per-volume Pearson correlation with sklearn")
ax.legend(fontsize=9)
ax.grid(alpha=0.3)

idx = rng.choice(out_sklearn.size, 5000, replace=False)
flat_sk   = out_sklearn.ravel()
flat_gram = out_gram.ravel()
flat_v3   = out_v3.ravel()

ax = fig.add_subplot(gs[2, :2])
ax.scatter(flat_sk[idx], flat_gram[idx], s=2, alpha=0.4, color=GRAM_COL,
           label=f"gram  r={np.corrcoef(flat_sk, flat_gram)[0,1]:.5f}")
lims = [flat_sk.min(), flat_sk.max()]
ax.plot(lims, lims, "k--", lw=1)
ax.set_xlabel("sklearn (reference)")
ax.set_ylabel("gram")
ax.set_title("Voxel scatter: gram vs sklearn")
ax.legend(fontsize=9)

ax = fig.add_subplot(gs[2, 2:])
ax.scatter(flat_sk[idx], flat_v3[idx], s=2, alpha=0.4, color=V3_COL,
           label=f"v3  r={np.corrcoef(flat_sk, flat_v3)[0,1]:.5f}")
ax.plot(lims, lims, "k--", lw=1)
ax.set_xlabel("sklearn (reference)")
ax.set_ylabel("v3")
ax.set_title("Voxel scatter: v3 vs sklearn")
ax.legend(fontsize=9)

diff_gram = (out_gram - out_sklearn).ravel()
diff_v3   = (out_v3   - out_sklearn).ravel()

ax = fig.add_subplot(gs[3, 0])
ax.hist(diff_gram, bins=80, color=GRAM_COL, alpha=0.8)
ax.axvline(0, color="k", lw=1)
ax.set_xlabel("gram − sklearn")
ax.set_title(f"Diff dist: gram\nmean={diff_gram.mean():.3f}  std={diff_gram.std():.3f}")

ax = fig.add_subplot(gs[3, 1])
ax.hist(diff_v3, bins=80, color=V3_COL, alpha=0.8)
ax.axvline(0, color="k", lw=1)
ax.set_xlabel("v3 − sklearn")
ax.set_title(f"Diff dist: v3\nmean={diff_v3.mean():.3f}  std={diff_v3.std():.3f}")

mid_z   = SHAPE[2] // 2
dwi_vol = int(np.argwhere(bvals > 0)[0][0])
sl_gram = out_gram[..., mid_z, dwi_vol] - out_sklearn[..., mid_z, dwi_vol]
sl_v3   = out_v3[...,  mid_z, dwi_vol] - out_sklearn[..., mid_z, dwi_vol]
absmax  = max(np.abs(sl_gram).max(), np.abs(sl_v3).max(), 1e-9)

ax = fig.add_subplot(gs[3, 2:])
im = ax.imshow(
    np.hstack([sl_gram, sl_v3]),
    cmap="bwr", vmin=-absmax, vmax=absmax,
)
ax.axvline(SHAPE[0] - 0.5, color="k", lw=1)
ax.set_title(
    f"Mid-slice diff vs sklearn  (vol {dwi_vol}, z={mid_z})\n"
    f"left = gram [max|Δ|={np.abs(sl_gram).max():.2f}]   "
    f"right = v3  [max|Δ|={np.abs(sl_v3).max():.2f}]",
    fontsize=9,
)
plt.colorbar(im, ax=ax)

out_path = r"f:\dipy_pr\dipy\gram_stability_comparison.png"
plt.savefig(out_path, dpi=150, bbox_inches="tight")
print(f"Plot saved to: {out_path}")
```

NekoAsakura:
> `G_b = tf @ tf.T - N_VOXELS * np.outer(rm, rm)`

@iprajwalreddy Can you update it with my latest commit? Thanks. 
The conditioning question seems has been resolved on my end. Regarding the default version, let’s keep it as is if you truly believe RAM is such a deciding constraint. Since the gram path beats the original v1 on every axis, I don't think it worth adding a flag to toggle in v1. Anything else you'd like me to address before this can go in?

iprajwalreddy:
@NekoAsakura I think its better now. 
image.

NekoAsakura:
@iprajwalreddy Can you double check graph of conditional number it seems still out of date.

iprajwalreddy:
@NekoAsakura Oh, I just updated the first plot and a few others as that was the one which was showing the accuracy. I'll update all the plots now.

iprajwalreddy:
@NekoAsakura Here's the updated plots:
image

NekoAsakura:
> @NekoAsakura Here's the updated plots: image

@skoudoro I think it's good enough. Any concern?

iprajwalreddy:
@NekoAsakura and @skoudoro, I ran a couple more tests for checking memory usage and time taken. The results are:
* Stanford_hardi:
  * Version 1: Memory : 2994.05 MB, Total time taken: 95.5 seconds.
  * Version 3: Memory : 323 MB, Total time taken: 27.88 seconds.
  * Gram Matrix: Memory: 2021.78 MB, Total time taken: 42.62 seconds.

* Another Nifti data which is 16gb:
  * Version 1: Memory: 57223.06 MB, Total time taken: 3510.098 seconds.
  * Version 3: Memory: 3022.625 MB, Total time taken: 1038.65 seconds.
  * Gram Matrix: Memory: 49279.04 MB, Total time taken: 2071.545 seconds.


While there is a definite speed-up using gram matrix, I think we should still use a flag to toggle whether to use gram matrix or not, as it is not mentioned in the original paper that there is usage of gram matrix and it would be different than the original method, if we implement only gram matrix. 
What do you think @skoudoro ?

NekoAsakura:
> @NekoAsakura and @skoudoro, I ran a couple more tests for checking memory usage and time taken. The results are:
> 
> * Stanford_hardi:
>   
>   * Version 1: Memory : 2994.05 MB, Total time taken: 95.5 seconds.
>   * Version 3: Memory : 323 MB, Total time taken: 27.88 seconds.
>   * Gram Matrix: Memory: 2021.78 MB, Total time taken: 42.62 seconds.
> * Another Nifti data which is 16gb:
>   
>   * Version 1: Memory: 57223.06 MB, Total time taken: 3510.098 seconds.
>   * Version 3: Memory: 3022.625 MB, Total time taken: 1038.65 seconds.
>   * Gram Matrix: Memory: 49279.04 MB, Total time taken: 2071.545 seconds.
> 
> While there is a definite speed-up using gram matrix, I think we should still use a flag to toggle whether to use gram matrix or not, as it is not mentioned in the original paper that there is usage of gram matrix and it would be different than the original method, if we implement only gram matrix. What do you think @skoudoro ?

Thanks for extra testing. Personally, I believe this is just an engineering optimization and does not involve modifying the patch2self algorithm itself. We also have confirmed that the results are sufficiently close. But I respect your final decision, just let me know whether any changes are needed.

skoudoro:
patch2self v3 is an optimization  (speed and memory) and we created an option to go back to the original v1

like you said @NekoAsakura, the gram matrix is another optimization, we need the option to let the user to go back to the original paper

Gram matrix, can be the default behavior for v1. But we need the option to go back to the original paper

When done, we can merged. 

Thanks @NekoAsakura and thanks @iprajwalreddy for working in details into this.

NekoAsakura:
Done. Pre-commit checks passed. Thanks for the review @skoudoro @iprajwalreddy.

skoudoro (approved):
Thanks for the update @NekoAsakura.

I will go ahead and merge. 

@iprajwalreddy, can you check in a new PR while v3 is failing sometimes. See below

```
FAILED denoise/tests/test_patch2self.py::test_patch2self_v3_preserves_out_dtype_and_precision_invariance - AssertionError: 
Not equal to tolerance rtol=0.02, atol=1.25

Mismatched elements: 1 / 13 (7.69%)
Max absolute difference among violations: 1.8789718500969208
Max relative difference among violations: 0.063662832871781
 ACTUAL: array([ 27.624342,  26.959814,  26.168653,  26.125479,  25.42737 ,
        27.030502,  24.055204,  26.89164 ,  25.138487,  27.61835 ,
        27.635452,  26.855309,  25.57368 ], dtype=float32)
 DESIRED: array([ 26.98437 ,  26.87689 ,  26.025953,  26.941496,  26.325948,
        27.49642 ,  24.502018,  27.153825,  25.822155,  26.85549 ,
        29.514424,  28.111983,  26.329036])
=========== 1 failed, 1314 passed, 182 skipped in 1534.58s (0:25:34) ===========
```

Thanks

PR #4009, opened 2026-05-06
STY: Apply Ruff SIM rules

DimitriPapadopoulos:
## Description

Apply subset of `SIM` rules, without enforcing them for now, as they're usually those that maintainers are not fond of, together with some `C4` or `PERF` rules.

## Motivation and Context

Applying `SIM222` helped identify and fix an actual bug.

~~Applying `SIM103` help identify the use of `&` instead of a logical `and`. Both work, but the logical conjunction operator is `and` in Python.~~ **Edit:** Actually `&` appears to be the only option for arrays, although using `&` feels like an abuse of Python syntax.

## How Has This Been Tested?

Pass CI tests.

## Checklist

## Type of Change

- [X] Bug fix (non-breaking change which fixes an issue)
- [ ] Documentation update

### DIFF
--- a/dipy/core/tests/test_gradients.py
+++ b/dipy/core/tests/test_gradients.py
@@ -160,7 +160,7 @@ def test_GradientTable_btensor_calculation():
             # Check that the b tensor magnitude is correct
             npt.assert_almost_equal(np.trace(bten), bval)
             # Check that the b tensor orientation is correct
-            if btens == ("LTE" or "CTE"):
+            if btens in ("LTE", "CTE"):
                 if bval != 0:
                     evals, evecs = np.linalg.eig(bten)
                     dot_prod = np.dot(np.real(evecs[:, np.argmax(evals)]), bvec)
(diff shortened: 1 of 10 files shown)


skoudoro:
Rules are fine here, thanks for this work @DimitriPapadopoulos.

However, all CI's are failing so a condition have a wrong modifications. Can you update?

dipy/core/tests/test_gradients.py
```diff
@@ -160,7 +160,7 @@ def test_GradientTable_btensor_calculation():
             # Check that the b tensor magnitude is correct
             npt.assert_almost_equal(np.trace(bten), bval)
             # Check that the b tensor orientation is correct
-            if btens == ("LTE" or "CTE"):
+            if btens in ("LTE", "CTE"):
```
DimitriPapadopoulos, line 163: I hope I got the intent right.

skoudoro: all good here. Thanks !

dipy/reconst/multi_voxel.py
```diff
@@ -187,7 +187,7 @@ def new_fit(self, data, *, mask=None, **kwargs):
 
             # Get weights from kwargs if provided
             weights = kwargs["weights"] if "weights" in kwargs else None
-            weights_is_array = True if type(weights) is np.ndarray else False
+            weights_is_array = type(weights) is np.ndarray
```
skoudoro: I would prefer `instance()` instead of `type`

skoudoro (commented):
Hi @DimitriPapadopoulos,

Overall, looks good. See below my tiny comment to fix. Thanks !

skoudoro (approved):
All good, merging

PR #4016, opened 2026-05-07
MNT: No need to specify `wheel` in build dependencies

DimitriPapadopoulos:
## Description

No need to add `wheel` to build tools.

## Motivation and Context

I noticed the issues here:
https://learn.scientific-python.org/development/guides/repo-review/?repo=dipy%2Fdipy&ref=HEAD&refType=branch

> [PP003](https://learn.scientific-python.org/development/guides/packaging-classic#PP003): Does not list wheel as a build-dep
> 
> Do not include `"wheel"` in your `build-system.requires`, setuptools does this via PEP 517 already. Setuptools will also only require this for actual wheel builds, and might have version limits.

## How Has This Been Tested?

CI tests pass.

## Checklist

## Type of Change

- [ ] Documentation update


skoudoro:
Hi @DimitriPapadopoulos,

Thanks !  Make sense, but I need to check locally in a new environment.

DimitriPapadopoulos:
It shouldn't be needed, unless `wheel` is called directly from the command line. I couldn't identify such a use case.

skoudoro:
ok, it is not needed. We are using `meson-python` and not `setuptools` to build wheels and it seems they do not need this package.

Thanks @DimitriPapadopoulos, merging

DimitriPapadopoulos:
Indeed, recent build backends do not need `wheels` any more.

PR #4023, opened 2026-05-08
MNT: `project.optional-dependencies` → `dependency-groups`

DimitriPapadopoulos:
## Description

Apply [PEP 735 – Dependency Groups in pyproject.toml](https://peps.python.org/pep-0735/).

## Motivation and Context

See:
https://learn.scientific-python.org/development/guides/repo-review/?repo=dipy%2Fdipy&ref=HEAD&refType=branch

> #### [PP006](https://learn.scientific-python.org/development/guides/packaging-simple#PP006): The dev dependency group should be defined
> The `dev` dependency group should be defined so tools like uv will work. These are better than the old `extras` system for tests, docs, and other dependencies that are not needed for PyPI installs.

## How Has This Been Tested?

Pass CI tests.

## Checklist

## Type of Change

- [ ] Documentation update


skoudoro:
Hi @DimitriPapadopoulos,

Not sure about this one. 

People are not used to this syntax, it will confuse them more than help I think. Personally, I never see anyone using this syntax for now despite being from 2023 ! 

Furthermore, All the files in requirements are very useful for conda users.

Personally not ready and comfortable to use [PEP 735](https://peps.python.org/pep-0735/)[PEP 735](https://peps.python.org/pep-0735/).

So I can not approve this PR. 

@arokem @Garyfallidis or other devs, any opinions ?

DimitriPapadopoulos:
I have seen lots of other projects moving to dependency groups. I believe the migration is slow across the ecosystem, because changes are often not straightforward.  All changes are initially confusing. Yet this the way forward since PEP 735 has been accepted.

It's not just a change of syntax. It also clarifies which dependencies are useful for end-users (and appear in PyPI) and which dependencies are for developers.

This also removes the need for regenerating the requirements files.

I can't comment for Conda and its use of requirements files, I was never a fan of Anaconda, from the start.

.github/workflows/nightly.yml
```diff
@@ -24,7 +24,7 @@ env:
   CIBW_BUILD: "cp3{12,13,14}-*"
   CIBW_BUILD_FRONTEND: 'pip; args: --pre --extra-index-url "https://pypi.anaconda.org/scientific-python-nightly-wheels/simple"'
   CIBW_TEST_SKIP: "*"
-  CIBW_TEST_REQUIRES: "-r requirements/build.txt pytest==8.0.0"
```
DimitriPapadopoulos: It's probably worth changing `pytest==8.0.0"` to the `test` group, whether this PR is accepetd or not.

tools/ci/install_dependencies.sh
```diff
@@ -7,7 +7,7 @@ set -ex
 
 echo "Display Python version"
 python -c "import sys; print(sys.version)"
-python -m pip install -U pip 'setuptools~=69.5' wheel
```
DimitriPapadopoulos: Note that that I have removed _wheel_, in the spirit of #4016. I believe this need to be applied whether this PR is accepetd or not.

.github/workflows/benchmark.yml
```diff
@@ -34,7 +34,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install .[dev,test,ml,extra,benchmark]
```
DimitriPapadopoulos: I think `extra` was an error that should be fixed whether this PR is accepted or not. Perhaps you meant `optional`?

.github/workflows/nightly.yml
DimitriPapadopoulos: See #4024.

tools/ci/install_dependencies.sh
DimitriPapadopoulos: See #4028.

.github/workflows/benchmark.yml
DimitriPapadopoulos: See #4028.

DimitriPapadopoulos:
Note that you don't **have** to use uv. However, it's a good idea to support all tools, whether it's `pip` or `uv`.

skoudoro:
> I can't comment for Conda and its use of requirements files, I was never a fan of Anaconda, from the start.
> Note that you don't have to use uv. However, it's a good idea to support all tools, whether it's pip or uv.

😄 that's why requirements files needed, a good idea to support all tools. Conda and others. They are still widely use.


Can you fix all the conflicts.

After a second thought, it is fine to have this syntax for pip. So it is fine to update our documentation for pip. However, I insist on keeping all requirements files. Also, we need to keep a system to auto-generate the requirements files if a group changes. 

Any thoughts ? anyone ?

DimitriPapadopoulos:
I could indeed try to adapt the script that generates requirements.

Out of curiosity, what are the requirements files needed for? Is Conda not compliant with PEP 735 yet? Projects that are also distributed via Conda have already transitioned to dependency groups.

**Edit:** Just found https://github.com/conda/conda/issues/10633. But then, how does Conda handle projects that have already transitioned to dependency groups?

skoudoro:
Out of curiosity, which big project has fully transitioned to dependency groups (PEP 735) ?

DimitriPapadopoulos:
You're right, I don't see large "famous" scientific projects having moved to PEP 735 dependency groups. I was mostly aware of tools such as the [Python Packaging Authority](https://github.com/pypa) tools.

However, quite a few smaller, less-known scientific projects have moved to PEP 735. Here is a more-or-less random list: [astropy](https://github.com/astropy/astropy), [pyuvdata](https://github.com/RadioAstronomySoftwareGroup/pyuvdata), [nibabel](https://github.com/nipy/nibabel), [pydicom](https://github.com/pydicom/pydicom), [pynetdicom](https://github.com/pydicom/pynetdicom), [fmriprep](https://github.com/nipreps/fmriprep), [Euphonic](https://github.com/pace-neutrons/Euphonic), [pyhf](https://github.com/scikit-hep/pyhf), ...

I understand `conda` or `mamba` understand neither `project.optional-dependencies` nor `dependency-groups`, only requirements files. But then how do other projects that don't have requirements files end up in Conda repositories (again I don't know much about Conda packaging)? is this perhaps about `conda-lock`, which does understand `project.optional-dependencies`?

DimitriPapadopoulos:
In any case, and without having investigated feasibility in any detail, wouldn't it be worthwhile rewriting the existing script and generate compatibility files for Conda using modern tools? For example:
* **uv** (would only be used for that purpose)
  ```
  uv export --group dev -o requirements/dev.txt
  ```
* **pip-compile**

skoudoro:
> wouldn't it be worthwhile rewriting the existing script and generate compatibility files for Conda using modern tools?

Yes, which one ?

> uv

I would prefer to avoid, even for this case...

> pip-compile

jazz-band is sunsetting -> https://jazzband.co/news/2026/03/14/sunsetting-jazzband

So not sure about the future of `pip-tools` package so `pip-compile`

---

That's why the custom script but open to other solution

DimitriPapadopoulos:
It turns out that the [dependency-groups](https://github.com/pypa/dependency-groups) package, under the [Python Packaging Authority](https://github.com/pypa) umbrella, resolves dependencies under `[dependency-groups]`and flattens a group to a list of PEP 508 strings. I can give it a try.

DimitriPapadopoulos:
The CI error seems unrelated:
```
dipy.data.fetcher.FetcherError: Failed to download https://zenodo.org/record/14538513/files/gs_streamlines.trk after 5 attempts. Got 403 Forbidden - likely rate limiting. Try reducing parallel downloads or wait before retrying. Last error: HTTP Error 403: Forbidden
```

pyproject.toml
```diff
@@ -150,14 +149,24 @@ doc = [
...
+[project.optional-dependencies]
+viz = ["fury>=0.12.0", "matplotlib"]
+ml = ["scikit_learn", "pandas", "statsmodels>=0.14.0", "tables", "torch"]
+optional = [
+    "dipy[viz, ml]",
+    "cvxpy",
+    "scikit-image",
+    "boto3",
+    "numexpr",
+]
```
skoudoro, line 169: why do we still need this ? 

Sorry for the naive question, I will look deeper.

doc/devel/installation_from_source.rst
```diff
@@ -272,8 +272,8 @@ To build the documentation in HTML in your computer you will need to do::
 
 You will also need to install the requirements for the documentation::
 
-    pip install -r requirements/doc.txt
-    pip install -r requirements/optional.txt
+    pip install --group doc
+    pip install -r .[optional]
```
skoudoro: Why `optional` is not in a group ?

Futhermore `-r` option needs a file, this will generate an error.

skoudoro (changes requested):
Hi @DimitriPapadopoulos,

Overall, looks good and it is almost ready to go. Can you check my last comments ?

skoudoro:
Thank you for finding [dependency-groups](https://github.com/pypa/dependency-groups) package also 👍

pyproject.toml
DimitriPapadopoulos: These are optional runtime dependencies and will appear as such in PyPI.

Dependency groups are typically for developer tools, that should not appear as runtime dependencies in PyPI.

doc/devel/installation_from_source.rst
DimitriPapadopoulos: Because these are runtime dependencies, not a dependency group.

skoudoro: ok.

Can you use `-e` instead of `-r` here.  `-r` will fail. thanks

skoudoro: or just remove the `-r`.   Let's not use the `-e`.

DimitriPapadopoulos: Fixed. Sorry about that, I think was seeing `-e` instead of `-r` :smile:

skoudoro (approved):
Thank you for this work @DimitriPapadopoulos,

All good, merging

PR #4033, opened 2026-05-16
[ENH] Cython, CUDA, webgpu, metal tracking from GPUStreamlines

36000:
## Description

This ports "simplified" probabilistic tracking from GPUStreamlines. Simple trackers assume:
        (1) the entire SF can be loaded into memory (not SH);
        (2) isotropic voxels;
        (3) simplified stopping criteria (ie, threshold on a scalar map);
        (5) generic probabilistic direction getting
Simplified trackers ("simplets") are implemented in CUDA, WebGPU, Metal.

In addition, some "lessons learned" from implementing these are ported back into the default fast cython tracking. In particular, we:
        (1) run in batches of streamlines, which can be saved to TRX files directly
        (2) parallelize the initial direction finding step as well
        (3) reserve the memory for the pmf outside of the individual propagators  (reducing malloc/free calls)
     
 This allows generating tractograms from 10,000,000 seeds (20M+ streamlines) and saving them into large TRX files without using too much memory, and can still be done on a laptop. All previous features should be preserved. This new Cython implementation is slower than the GPUStreamlines numba implementation, but only slightly (say, 20-30%), and it allows for all of the things the "simplified" probabilistic trackers cannot do (anisotropic voxels, use of SH directly over SF, complicated stopping criteria). 
 
 There are these additional features in general:
      (1) includes an update to handle tracking on non-symmetric ODFs and SH with full basis
      (2) faster random seeds
      (3) Batched TRX file saving

## Motivation and Context

This makes prob. tractography much faster. Using an example HBN subject, a >20mm threshold, and GWI seeding, it generates ~25 million streamlines with an average length of 62.4mm in around 20 minutes on an AMD Ryzen AI 9 HX 370 CPU with 24 cores. This was using an ODF with 724 directions, a step size of 0.5, and a max angle of 30 degrees. This time *includes* finding peak directions from seed points, streamline generation, and file saving to TRX. 

## How Has This Been Tested?

The GPUStreamlines version has been used many times; I added this to one existing DIPY test of prob. tracking.

## Checklist

- [ x] I have read the [CONTRIBUTING](https://github.com/dipy/dipy/blob/master/.github/CONTRIBUTING.md) guidelines.
- [ x] My code follows the [DIPY coding style](https://docs.dipy.org/stable/devel/coding_style_guideline.html).
- [ x] I have added tests that cover my changes (if applicable).
- [ x] All new and existing tests pass locally.
- [x ] I have updated the documentation accordingly (if applicable).

## Type of Change

- [ x] New feature (non-breaking change which adds functionality)
- [ ] Documentation update


skoudoro:
Thank you for this! 

Let us know when it is ready for review !

36000:
@skoudoro this is ready for review. I think this can be a template for how we add other accelerators from GPUStreamlines.

skoudoro:
Thank you @36000 !  I will look deeper into it this week and get back to you asap

dipy/tracking/jit/numba_njit/generate_streamlines_numba.py
```diff
@@ -0,0 +1,291 @@
+import math
+import random
+
+import numpy as np
+
+from dipy.tracking.jit.numba_njit.tracking_helpers import trilinear_interp_generator
+
+
+def genStreamlinesMergeProb_generator(
```
skoudoro: C style naming, can you follow python style naming

dipy/tracking/tracker.py
```diff
@@ -243,6 +270,14 @@ def probabilistic_tracking(
         reached the stopping criterion.
     save_seeds: bool, optional
         True to return the seeds with the associated streamline.
+    use_jit : bool, optional
+        Use the Numba JIT implementation of the probabilistic tracking.
+        If None, will only use the JIT implementation if numba is installed.
+        Default: None.
```
skoudoro: default to remove

dipy/tracking/tracker.py
```diff
@@ -243,6 +270,14 @@ def probabilistic_tracking(
...
+    jit_chunk_size : int, optional
+        Number of seeds to process in each chunk when using the JIT implementation.
+        A smaller chunk size will reduce memory usage but may increase processing time.
+        Default: 25000
```
skoudoro: default to remove

dipy/tracking/tracker.py
```diff
@@ -263,22 +298,96 @@ def probabilistic_tracking(
...
-        sphere=sphere,
-        basis_type=basis_type,
-        legacy=legacy,
-        nbr_threads=nbr_threads,
-        seed_buffer_fraction=seed_buffer_fraction,
-        save_seeds=save_seeds,
-    )
+    numba, have_numba, _ = optional_package("numba")
+    if use_jit is None:
+        use_jit = have_numba
```
skoudoro: to be more pythonic: `use_jit = use_jit or have_numba`

dipy/tracking/tracker.py
```diff
@@ -263,22 +298,96 @@ def probabilistic_tracking(
...
+
+        if seed_directions is not None:
+            warn(
+                (
+                    "seed_directions are not supported when "
+                    "using the JIT probabilistic tracker. "
+                    "Set jit=False to use seed_directions."
+                ),
+                stacklevel=2,
+            )
```
skoudoro: why this is not supported ?

dipy/tracking/tracker.py
```diff
@@ -187,6 +212,8 @@ def probabilistic_tracking(
     seed_buffer_fraction=1.0,
     return_all=True,
     save_seeds=False,
+    use_jit=None,
```
skoudoro: use_jit=None silently opts users into JIT when numba is installed. Existing users running probabilistic_tracking with save_seeds=True or return_all=False will get an ArraySequence instead of a generator — different type, different memory profile.

This is a Breaking change without warning. Default use_jit=False or later in the code, set it to false if none until JIT path fully matches the non-JIT path.

dipy/tracking/tracker.py
```diff
@@ -263,22 +298,96 @@ def probabilistic_tracking(
...
+        pmf_field = np.zeros(
+            (*selected_pmf["value"].shape[:3], len(sphere.vertices)), dtype=np.float64
+        )
+        stop_map = np.zeros(selected_pmf["value"].shape[:3], dtype=np.float64)
+        for ii in range(selected_pmf["value"].shape[0]):
+            for jj in range(selected_pmf["value"].shape[1]):
+                for kk in range(selected_pmf["value"].shape[2]):
+                    idx_ = np.array([ii, jj, kk], dtype="float64")
+                    pmf_field[ii, jj, kk] = pmf_gen.get_pmf(idx_)
+                    stop_map[ii, jj, kk] = sc.check_point(idx_)
```
skoudoro, line 328: this look like a very slow path. Triple loop in python to build those map. It needs some thinking.

dipy/tracking/jit/numba/nu_tractography.py
```diff
@@ -0,0 +1,239 @@
...
+            __pval = __cval
+
+        return shDir0, slineOutOff
+
+    def _generate_streamlines(self, seeds, shDir0, slineOutOff):
+        nSlines = int(slineOutOff[-1])
+
+        slineSeed = np.full(nSlines, -1, dtype=np.int32)
+        slineLen = np.zeros(nSlines, dtype=np.int32)
+        sline = np.zeros((nSlines * self.MAX_SLINE_LEN * 2, 3), dtype=self.REAL_DTYPE)
```
skoudoro: I need to check memory impact here on laptop

dipy/tracking/jit/numba_njit/generate_streamlines_numba.py
```diff
@@ -0,0 +1,291 @@
...
+        first_step,  # float64[3]  initial direction
+        pmf_volume,  # float64[DIMX, DIMY, DIMZ, DIMT]
+        metric_map,  # float64[DIMX, DIMY, DIMZ]
+        sphere_vertices,  # float64[DIMT, 3]
+        streamline,  # float64[MAX_SLINE_LEN*2, 3]
+        pmf_scratch,  # float64[DIMT]  scratch buffer
+    ):
+        point = np.empty(3, dtype=np.float64)
+        direction = np.empty(3, dtype=np.float64)
+        new_dir = np.empty(3, dtype=np.float64)
```
skoudoro: float64 enforced, Why ? could it be float32 since user can choose float32

skoudoro (changes requested):
Hi @36000,

Thank you for this. sorry for the delay. Here an initial review /observation

Overall, can it be more functional programming than class based ? Many methods could be easily reuse somewhere else as functions. The design seems a bit over-complicated and can be simplify. ou can have a dataclass object to move around for all your needed config. 

- numba folder is missing an __init__.py
- numba_jit folder is missing an __init__py. 

2 subfolders are not needed.  `Jit` is enough already

in docstring:
- remove all the default
- make sure to mark what is optional or not.

do not forget python style instead of c-style

dipy/tracking/tracker.py
36000: I am not super familiar with the best way to do this via the DIPY API, I would like to precalculate all of these maps and handle the interpolation internally, because that is how GPUStreamlines is currently written.

dipy/tracking/jit/numba/nu_tractography.py
36000: This is minimal because all of the GPUStreamlines code is chunked by default. I think it would be roughly 600MB using the default chunk size.

dipy/tracking/tracker.py
36000: I will set it to False and also implement seed_directions, save_seeds, and return_all. Hopefully, we can make it so the two backends can return the same format.

36000:
@skoudoro how does this refactoring look? I have moved things into a more functional paradigm and removed a lot of the subfolders. This will still work to add in cuda tracking and metal tracking in the future. I also made everything return generators. I am wondering, can I also add an option to return a filepath from this function? The way pyAFQ does tractography, we save the tractogram to disk as we generate, so we can generate TRX files of 100s of GBs in size in minutes on a laptop. I changed the trx saving method a little bit to accept a generator, so it may be useful for non-JIT tracking.

skoudoro:
Hi @36000,

> how does this refactoring look? 

To me, much better !  much easier to read, follow and understand.

> I also made everything return generators.

Great !

> I am wondering, can I also add an option to return a filepath from this function?

This, I do not know yet, we do not do it at this level usually.

> I changed the trx saving method a little bit to accept a generator,

Make sense and great update.


I need to play a bit with this PR and see the performance improvement on a laptop. I will try to run than tonight or this wekend and get back to you next week.

Thanks for this work !

dipy/tracking/numba/generate_streamlines.py
```diff
@@ -0,0 +1,292 @@
...
+    DIMT,
+    SPHERE_SYMM,
+    STEP_SIZE,
+    MAX_ANGLE,
+    TC_THRESHOLD,
+    MAX_SLINE_LEN,
+    PMF_THRESHOLD_P,
+    REAL_DTYPE,
+):
+    from numba import njit, prange
```
skoudoro, line 22: import on the top with `optional_package`

dipy/tracking/numba/generate_streamlines.py
```diff
@@ -0,0 +1,292 @@
+import math
+import random
+
+import numpy as np
+
+from dipy.tracking.numba.tracking_helpers import trilinear_interp_generator
+
+
+def gen_streamlines_prob_generator(
```
skoudoro, line 9: function to document

dipy/tracking/numba/generate_streamlines.py
```diff
@@ -0,0 +1,292 @@
...
+    MAX_SLINE_LEN,
+    PMF_THRESHOLD_P,
+    REAL_DTYPE,
+):
+    from numba import njit, prange
+
+    trilinear_interp = trilinear_interp_generator(DIMX, DIMY, DIMZ, DIMT)
+
+    @njit
+    def check_point(point, metric_map):
```
skoudoro, line 27: function to document

dipy/tracking/numba/generate_streamlines.py
```diff
@@ -0,0 +1,292 @@
...
+            + wx1 * wy0 * wz0 * metric_map[ix1, iy0, iz0]
+            + wx1 * wy0 * wz1 * metric_map[ix1, iy0, iz1]
+            + wx1 * wy1 * wz0 * metric_map[ix1, iy1, iz0]
+            + wx1 * wy1 * wz1 * metric_map[ix1, iy1, iz1]
+        )
+
+        return 2 if val > TC_THRESHOLD else 3
+
+    @njit
+    def get_direction_prob_step(
```
skoudoro, line 69: function to document

dipy/tracking/numba/generate_streamlines.py
```diff
@@ -0,0 +1,292 @@
...
+                new_dir[2] = -sphere_vertices[ind_prob, 2]
+        else:
+            new_dir[0] = sphere_vertices[ind_prob, 0]
+            new_dir[1] = sphere_vertices[ind_prob, 1]
+            new_dir[2] = sphere_vertices[ind_prob, 2]
+
+        return 1
+
+    @njit
+    def tracker(
```
skoudoro, line 150: function to document

dipy/tracking/numba/generate_streamlines.py
```diff
@@ -0,0 +1,292 @@
...
+
+            if tissue_class == 0 or tissue_class == 1 or tissue_class == 3:
+                break
+
+            i += 1
+
+        return i, tissue_class
+
+    @njit(parallel=True)
+    def gen_streamlines_prob(
```
skoudoro, line 212: function to document

dipy/tracking/numba/num_streamlines.py
```diff
@@ -0,0 +1,166 @@
+import math
+
+import numpy as np
+
+from dipy.tracking.numba.tracking_helpers import trilinear_interp_generator
+
+
+def get_num_streamlines_prob_generator(
```
skoudoro, line 8: function to document

dipy/tracking/numba/num_streamlines.py
```diff
@@ -0,0 +1,166 @@
...
+    DIMZ,
+    DIMT,
+    RELATIVE_PEAK_THRESH,
+    MIN_SEPARATION_ANGLE,
+    NUM_EDGES,
+    SPHERE_SYMM,
+    PMF_THRESHOLD_P,
+    REAL_DTYPE,
+):
+    from numba import njit, prange
```
skoudoro, line 20: import on the top with `optional_package`

dipy/tracking/numba/num_streamlines.py
```diff
@@ -0,0 +1,166 @@
...
+
+    trilinear_interp = trilinear_interp_generator(DIMX, DIMY, DIMZ, DIMT)
+
+    @njit
+    def peak_directions(odf, sphere_vertices, sphere_edges, dirs_out):
```
skoudoro, line 25: function to document

dipy/tracking/numba/num_streamlines.py
```diff
@@ -0,0 +1,166 @@
...
+            if not too_close:
+                dirs_out[k, 0] = ax
+                dirs_out[k, 1] = ay
+                dirs_out[k, 2] = az
+                k += 1
+
+        return k
+
+    @njit
+    def get_direction_prob_start(
```
skoudoro, line 113: function to document

dipy/tracking/numba/num_streamlines.py
skoudoro: Also, I wonder why it needs to be reimplemented.  Is our cython version slower ? 
I need to understand this part. 

I would prefer if we could reuse a maximum of code.

dipy/tracking/numba/num_streamlines.py
```diff
@@ -0,0 +1,166 @@
...
+            if pmf_scratch[i] < abs_thresh:
+                pmf_scratch[i] = 0.0
+
+        # Step 3  peak directions
+        ndir = peak_directions(pmf_scratch, sphere_vertices, sphere_edges, dirs_out)
+
+        return ndir
+
+    @njit(parallel=True)
+    def get_num_streamlines_prob(
```
skoudoro, line 143: function to document

dipy/tracking/numba/tracker.py
```diff
@@ -0,0 +1,123 @@
...
+from dipy.tracking.numba.generate_streamlines import (
+    gen_streamlines_prob_generator,
+)
+from dipy.tracking.numba.num_streamlines import (
+    get_num_streamlines_prob_generator,
+)
+
+
+def numba_sl_generator(jit_tracker_data, seeds, seed_directions=None, nbr_threads=0):
+    import numba
```
skoudoro, line 13: use `optional_package` and needs to be on the top

dipy/tracking/numba/tracker.py
```diff
@@ -0,0 +1,123 @@
...
+from dipy.tracking.generic_jit_tracker import StreamlineChunk, streamline_generator
+from dipy.tracking.numba.generate_streamlines import (
+    gen_streamlines_prob_generator,
+)
+from dipy.tracking.numba.num_streamlines import (
+    get_num_streamlines_prob_generator,
+)
+
+
+def numba_sl_generator(jit_tracker_data, seeds, seed_directions=None, nbr_threads=0):
```
skoudoro, line 12: function to document

dipy/tracking/numba/tracking_helpers.py
```diff
@@ -0,0 +1,56 @@
+import math
+
+
+def trilinear_interp_generator(DIMX, DIMY, DIMZ, DIMT):
```
skoudoro, line 4: same document and import

skoudoro (changes requested):
Hi @36000,

Also, function documentation needs to be added, import normalize. 

Multiple function are duplicated for jit. which can bring maintenance burden and some implementation difference. I am still wondering if there is a way to reuse some part and keep the same performance.  I will study that  but if you could look into it also. 

Thanks

skoudoro:
also, can you add some benchmarks here between current implementation and this JIT numba implementation ?

skoudoro:
Hi @36000,

To follow up our discussion in Neurohackademy, do you have time for an update here ?

As a reminder, we plan a release by the end of september / beginning of october

36000:
Hello @skoudoro , thanks for the reminder. I will get a new draft of this working this week, so we can review it and merge before the big release

36000:
@neurolabusc Just to keep you posted, we are planning on deprecating GPUStreamlines and moving things into DIPY proper. Here is the PR where we are doing this. this has involved a few changes:
(1) we removed the old numba tractography and switched it to cython, merging it with DIPY's cython tracking, to reduce code duplication
(2) we removed PTT/bootstrap support (not sure anyone uses them and the code becomes much longer with them)

Otherwise, we still have your webgpu+metal support alongside cuda and cython, just only implemented for generic, probabilistic tracking. 

Our goal is to have this done by DIPY 2.0

36000:
@skoudoro I have made a lot of changes. I have removed our numba implementation and changed DIPY's cython implementation so that we do not have redundant codebases. This should speed up DIPY's default tracking. I also brought in the CUDA, webgpu, and metal trackers in a "simplet" folder described in the top comment. 

It is ready for review.

skoudoro:
> we removed PTT/bootstrap support

I am concerned by that, I need to look deeper. Can the user still use a slow version if they want ?

> I have removed our numba implementation and changed DIPY's cython implementation so that we do not have redundant codebases. 

Great !

> This should speed up DIPY's default tracking. I also brought in the CUDA, webgpu, and metal trackers

very cool, excellent work! I will look deeper into it but I already like the premises

> "simplet" folder described in the top comment.

Why "simplet" ? meaning ? strange folder name. 


Failing CI's are not related. 

I hope to get back to you by the end of the week.

36000:
> I am concerned by that, I need to look deeper. Can the user still use a slow version if they want ?

I removed it from the GPU versions I am moving in to DIPY; I did not touch the bootstrap/ptt pipelines already in DIPY. I can bring them over if you think there is a demand; I just did not think anyone was using the GPU bootstrapped tractography/GPU PTT.


> Why "simplet" ? meaning ? strange folder name.

We can change it, for me it is short for simple tractography

dipy/workflows/tracking.py
```diff
@@ -24,6 +25,25 @@
 from dipy.workflows.workflow import Workflow
 
 
+def _save_tracking_result(tracking_result, reference, out_tract, save_seeds):
```
skoudoro, line 28: Please, add docstring.

Note: this function should disappear when we have only one tracking cli

dipy/core/interpolation.pxd
```diff
@@ -3,7 +3,11 @@ cimport numpy as cnp
 from dipy.align.fused_types cimport floating, number
 
 
-cdef int trilinear_interpolate4d_c(floating[:, :, :, :] data,
+cdef int trilinear_interpolate3d_c(double[:, :, :] data,
```
skoudoro, line 6: maybe we should use floating datatype to accept `float` and `double`

dipy/tracking/simplet/metal.py
```diff
@@ -0,0 +1,180 @@
...
+        "4f8i",
+        tracker_data.max_angle,
+        tracker_data.stop_threshold,
+        tracker_data.step_size,
+        tracker_data.pmf_threshold,
+        seed & 0xFFFFFFFF,
+        (seed >> 32) & 0xFFFFFFFF,
+        int(n_seeds),
+        *tracker_data.shape,
+        tracker_data.max_steps,
```
skoudoro, line 96: this generate an error so  Use "4f2I6i" or pack via `np.uint32` like in `webgpu.py`

dipy/tracking/simplet/kernel.cu
```diff
@@ -0,0 +1,379 @@
+/* Copyright (c) 2020, NVIDIA CORPORATION. All rights reserved.
```
skoudoro, line 1: in this file, AI detected the following error/bug but not sure if it is relevant:

> ballot mask 0 (reachable: curand_uniform ∈ (0,1] → selected_cdf == last_cdf) gives indProb = low + __ffs(0) - 1 = -1 when low == 0 → sphere_vertices[-1]. Metal/WGSL guard with dimt-1 fallback; port it.

dipy/workflows/tracking.py
```diff
@@ -47,7 +67,7 @@ def run(
         save_seeds=False,
         nbr_threads=0,
         random_seed=1,
-        seed_buffer_fraction=1.0,
+        chunk_size=25000,
```
skoudoro, line 70: you will need `@deprecated_params` decorator to warn the user of this change

dipy/workflows/tracking.py
```diff
@@ -295,7 +307,7 @@ def run(
         min_wm_pve_before_stopping=0,
         nbr_threads=0,
         random_seed=1,
-        seed_buffer_fraction=1.0,
```
skoudoro, line 298: you will need `@deprecated_params` decorator to warn the user of this change

dipy/tracking/utils.py
```diff
@@ -538,45 +540,33 @@ def random_seeds_from_mask(
...
     rng = np.random.default_rng(random_seed)
     shape = mask.shape
-    mask = mask.flatten()
-    indices = np.arange(len(mask))
-    rng.shuffle(indices)
+    mask = mask.ravel()
 
-    where = [np.unravel_index(i, shape) for i in indices if mask[i] == 1]
-    num_voxels = len(where)
+    # Randomize the voxel order, then keep the voxels inside the mask
```
skoudoro, line 547: `random_seeds_from_mask` now yields different seeds for same `random_seed`. 

this is a reproducibility issue and break. Something to dig further and important to fix. 

it might be the reason of the difference of streamline between cpu and gpu

skoudoro (commented):
Hi @36000,

Sorry for the delay, still reviewing step by step to understand the code and what has been done. 

In parallel, it seems there is a memory leak with Metal version. here some stats from my macOS M1 (did not try yet the cuda backend):

│ backend │ seeds │ mode │ streamlines │ track time │  peak RSS  │ RSS after  │
│ cpu         │ 588k  │ trx       │ 507,052       │ 27.2 s        │ 2.2 GB        │ 1.8 GB       │
│ metal      │ 588k  │ trx       │ 507,160        │ 13.1 s        │ 19.4 GB 🔴 │ 16.2 GB 🔴 │
│ webgpu  │ 588k  │ trx       │ 507,160        │ 18.6 s       │ 6.5 GB        │ 2.2 GB     │


but even if we lower the memory use of Metal, it will be around 7GB for the peak which is still 3x more than the cpu. I need to check if we can do better.  

Also:
- the number of streamlines differ between cpu and GPU. I will check why, it might be normal. 
- speed doesnot look impressive, but it makes sense to me for such a small scale. it is a good news for cython version and speed will show in large scale for GPU
- wgpu slower than metal : expected and make sense to me.
-   PMF held twice — float64 copy for SimplePmfGen (seed peaks) + float32 copy for GPU. Please using `floating` type to avoid that kind of issue



Reproducibility is important. CPU +GPU should give identic result if possible with exact same seed and rng.



I will continue to dig slowly but I believe we are close to have this huge piece IN. 

Thanks for this works @36000

skoudoro:
@arokem  @deka27 @maharshi-gor @gabknight @Atharva-Shah-2298 :  can you do a round of review  please? 

At least, it would be great if you could  run it locally and give some stats. My tests was only on macOS M1 for now. 

Thank you all for the feedback

skoudoro:
Also, Can you rename `simplet` module by something more explicit please ? 

call it : `accel` or `gpu`

Same for the function/class name that contain `simple`.  replace it by `gpu`. 

e.g: 

```
 prepare_simple_tracker_data → prepare_gpu_tracker_data
simple_sl_generator → gpu_streamline_generator
_SimpleTrackerData → _GpuTrackerData
```

Atharva-Shah-2298:
I benchmarked this on AMD Ryzen 7 5700 with 32 GBs (Windows) 

│ backend │ seeds │ mode │ streamlines │ track time │ peak RSS │ RSS after │
│ cpu     │ 588k  │ trx  │ 588,000     │ 4.25 s     │ 0.44 GB  │ 0.15 GB   │
│ webgpu  │ 588k  │ trx  │ 588,000     │ 3.63 s     │ 0.60 GB  │ 0.30 GB   │

For some reason no matter what I did...for CUDA it says "CUDA error 700: illegal memory access".

36000:
Thank you for testing this. In the past things like "CUDA error 700: illegal memory access" have been caused by having conflicting driver versions, so make sure you have the dependencies specified in the cuda sections. It can get pretty annoying; that's part of the reason I wrote the Numba implementation:

```
cu13 = [
    "nvidia-cuda-runtime",
    "nvidia-curand-cu12",
    "cuda-python<14",
    "cuda-core[cu13]",
    "cuda-cccl[cu13]",
]
cu12 = [
    "nvidia-cuda-runtime-cu12",
    "nvidia-curand-cu12",
    "cuda-python<13",
    "cuda-core[cu12]",
    "cuda-cccl[cu12]",
]
```

For naming, I prefer accel to gpu, as metal is not a gpu and we may want to add different backends later.

To directly compare the cpu and accel versions, I can to implement float 32 and Philox4x32-10 in cython (they do not need to be default). Then I can set up a test where the floating-point and stop masks exactly match.

As for webgpu/metal issues, I'm afraid I will be just as clueless about them as you all; I believe they were translated from the cuda version using AI by @neurolabusc . Maybe I broke something when i removed bootstrapping/ptt from the metal version, I will look into it. Hopefully it is just a matter of things not being freed at the right time. I do not currently use a mac so I will have trouble debugging this.

PR #4035, opened 2026-05-21
Security: Unsafe Pickle Loading in load_pickle Function

tomaioo:
## Summary

Security: Unsafe Pickle Loading in load_pickle Function

## Problem

**Severity**: `Critical` | **File**: `dipy/io/pickles.py:L42`

The `load_pickle` function in `dipy/io/pickles.py` uses `pickle.load()` without any validation or restrictions on what can be deserialized. This is a classic insecure deserialization vulnerability. An attacker could craft a malicious pickle file that, when loaded, executes arbitrary code on the user's machine. This is particularly dangerous since pickle files are commonly used to store model data and other serialized objects in scientific computing workflows.

## Solution

Replace `pickle.load()` with a safer alternative. Options include: 1) Using `numpy.load()` with `allow_pickle=False` for numpy arrays, 2) Using JSON for simple data structures, 3) If pickle must be used, implement a restricted unpickler that whitelists safe classes, or 4) Add clear documentation warnings about only loading pickle files from trusted sources. Consider deprecating this function in favor of safer serialization formats.

## Changes

- `dipy/io/pickles.py` (modified)

### DIFF
--- a/dipy/io/pickles.py
+++ b/dipy/io/pickles.py
@@ -21,7 +21,7 @@ def save_pickle(fname, dix):
     >>> fd, fname = mkstemp() # make temporary file (opened, attached to fh)
     >>> d={0:{'d':1}}
     >>> save_pickle(fname, d)
-    >>> d2=load_pickle(fname)
+    >>> d2=load_pickle(fname, check_safety=False)
 
     We remove the temporary file we created for neatness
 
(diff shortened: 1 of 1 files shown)


skoudoro (commented):
Thanks for this @tomaioo 

Actually, I am not sure we should keep this function. I believe we should deprecate it and then remove it. 

For now, merging  after the CI runs

skoudoro (approved):
All good, Thanks @tomaioo , merging

Issue #4041, opened 2026-05-25
Failures in test_patch2self_v3_preserves_out_dtype_and_precision_invariance on amd64


sanvila:
### Summary

Hello. The build of this package has failed at least twice in the official Debian infrastructure due to a failing test.

https://buildd.debian.org/status/logs.php?pkg=dipy&arch=all

Note: In Debian, the arch:all autobuilder creates architecture-independent packages, but it is actually a regular amd64 machine, so we can say the failure happens (randomly) at least on the amd64 architecture.

The failing test is the same in both cases:

`FAILED dipy/denoise/tests/test_patch2self.py::test_patch2self_v3_preserves_out_dtype_and_precision_invariance`

This is the failing test in detail:

```
=================================== FAILURES ===================================
_______ test_patch2self_v3_preserves_out_dtype_and_precision_invariance ________

pytestconfig = <_pytest.config.Config object at 0x7fdc1170c1a0>, args = ()
kwargs = {'rng': Generator(PCG64) at 0x7FDC0533BD80}
rng = Generator(PCG64) at 0x7FDC0533BD80, signature = <Signature (rng)>

    def _set_random_number_generator_wrapper(pytestconfig, *args, **kwargs):
        rng = np.random.default_rng(seed_v)
        kwargs["rng"] = rng
        signature = inspect.signature(func)
        if pytestconfig and "pytestconfig" in signature.parameters.keys():
            output = func(pytestconfig, *args, **kwargs)
        else:
>           output = func(*args, **kwargs)
                     ^^^^^^^^^^^^^^^^^^^^^

dipy/testing/decorators.py:104: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

rng = Generator(PCG64) at 0x7FDC0533BD80

    @needs_sklearn
    @set_random_number_generator(2025)
    def test_patch2self_v3_preserves_out_dtype_and_precision_invariance(rng):
        shape = (10, 10, 6, 13)
        base = 30.0 + 2.0 * rng.standard_normal(shape)
        bvals = np.repeat(1000.0, shape[-1])  # all DWI
    
        data32 = base.astype(np.float32)
        data64 = base.astype(np.float64)
    
        den32 = p2s.patch2self(data32, bvals, model="ridge", alpha=1.0, version=3)
[...]

PR #4043, opened 2026-05-26
RF: Added fixes

iprajwalreddy:
This PR fixes issue mentioned in #4041. The issue was due to the way sketching is done in v3. Sketching is random, so when there were two calls being done and the random state not being set to a specific value, caused the error.


skoudoro (approved):
thank you for this @iprajwalreddy, merging

PR #4046, opened 2026-05-28
Fix single-CPU edge case in test_auto_ray_chunk_size_negative_n_jobs

### LINKED ISSUE #4034: test_auto_ray_chunk_size_negative_n_jobs does not work on single-CPU systems
### Summary

Hello. While building (the Debian package of) dipy (version 1.12.1) on a virtual machine with 1 CPU, the test in the title failed in this way:

```
=================================== FAILURES ===================================
___________________ test_auto_ray_chunk_size_negative_n_jobs ___________________

    def test_auto_ray_chunk_size_negative_n_jobs():
        # n_jobs=-1 must not collapse to 1 worker (which would yield 1 giant chunk)
        from dipy.utils.multiproc import determine_num_processes
    
        n_vox = 100_000
        chunk_neg = auto_ray_chunk_size(n_jobs=-1, n_gradients=160, n_vox=n_vox)
        chunk_pos = auto_ray_chunk_size(
            n_jobs=determine_num_processes(-1), n_gradients=160, n_vox=n_vox
        )
        assert chunk_neg == chunk_pos
        # Must produce more than one chunk
>       assert chunk_neg < n_vox
E       assert 100000 < 100000

dipy/utils/tests/test_parallel.py:132: AssertionError

```

It looks like this particular test is not really meant to be tested on single-cpu systems. If this is the case, I suggest using something like this:

`@pytest.mark.skipif(os.cpu_count() == 1, reason="not meant to be run on single-cpu systems")`

This way everybody will be able to build the package from source (Debian packages always run the tests as part of the build itself).

On the other hand, maybe this is actually a bug in the test that needs to be fixed. I don't really know.

Note: A very simple way to reproduce (if you don't have a single-cpu VM at hand) is to configure GRUB with `GRUB_CMDLINE_LINUX="nr_cpus=1"` and reboot.

Thanks.

### Platform

Debian unstable

### DIPY version

1.12.1

### Environment

_No response_

### Python version

python 3.13 and 3.14

### Additional information

_No response_

Sunny11092003:
## Description

Fixes the single-CPU edge case in
`test_auto_ray_chunk_size_negative_n_jobs`.

The previous test assumed that `n_jobs=-1`
would always result in multiple chunks, which
is not true on systems with only one available CPU.

This update adjusts the test logic to correctly
handle both:

* multi-CPU systems
* single-CPU systems

## Motivation and Context

On single-CPU environments,
`determine_num_processes(-1)` resolves to `1`,
which makes a single chunk equal to `n_vox`
valid behavior.

The previous assertion:

```python
assert chunk_neg < n_vox
```

incorrectly failed in this case.

Fixes #4034

## How Has This Been Tested?

The logic was reviewed against the existing
test behavior and issue reproduction details.

Note:
I could not run the full test suite locally because
the current development branch requires Python 3.12+,
while my local environment is Python 3.10.

## Checklist

* [x] I have read the [[CONTRIBUTING](https://github.com/dipy/dipy/blob/master/.github/CONTRIBUTING.md)](https://github.com/dipy/dipy/blob/master/.github/CONTRIBUTING.md) guidelines.
* [x] My code follows the [[DIPY coding style](https://docs.dipy.org/stable/devel/coding_style_guideline.html)](https://docs.dipy.org/stable/devel/coding_style_guideline.html).
* [x] I have added tests that cover my changes (if applicable).
* [ ] All new and existing tests pass locally.
* [ ] I have updated the documentation accordingly (if applicable).

## Type of Change

* [x] Bug fix (non-breaking change which fixes an issue)
* [ ] New feature (non-breaking change which adds functionality)
* [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
* [ ] Documentation update
* [ ] Maintenance / CI / Infrastructure

### DIFF
--- a/dipy/utils/tests/test_parallel.py
+++ b/dipy/utils/tests/test_parallel.py
@@ -119,17 +119,31 @@ def test_auto_ray_chunk_size_memory_scales_with_ram():
 
 
 def test_auto_ray_chunk_size_negative_n_jobs():
-    # n_jobs=-1 must not collapse to 1 worker (which would yield 1 giant chunk)
+    # n_jobs=-1 must behave the same as the resolved worker count
     from dipy.utils.multiproc import determine_num_processes
 
     n_vox = 100_000
-    chunk_neg = auto_ray_chunk_size(n_jobs=-1, n_gradients=160, n_vox=n_vox)
-    chunk_pos = auto_ray_chunk_size(
-        n_jobs=determine_num_processes(-1), n_gradients=160, n_vox=n_vox
+    resolved_jobs = determine_num_processes(-1)
+
+    chunk_neg = auto_ray_chunk_size(
+        n_jobs=-1,
+        n_gradients=160,
+        n_vox=n_vox,
     )
+
+    chunk_pos = auto_ray_chunk_size(
+        n_jobs=resolved_jobs,
+        n_gradients=160,
+        n_vox=n_vox,
+    )
+
     assert chunk_neg == chunk_pos
-    # Must produce more than one chunk
-    assert chunk_neg < n_vox
+
+    # Multiple chunks are only expected when more than one worker exists
+    if resolved_jobs > 1:
+        assert chunk_neg < n_vox
+    else:
+        assert chunk_neg == n_vox
 
 
 def test_auto_ray_chunk_size_more_jobs_smaller_chunks():
--- a/dipy/utils/tests/test_parallel.py
+++ b/dipy/utils/tests/test_parallel.py
@@ -140,10 +140,7 @@ def test_auto_ray_chunk_size_negative_n_jobs():
     assert chunk_neg == chunk_pos
 
     # Multiple chunks are only expected when more than one worker exists
-    if resolved_jobs > 1:
-        assert chunk_neg < n_vox
-    else:
-        assert chunk_neg == n_vox
+    assert chunk_neg <= n_vox
 
 
 def test_auto_ray_chunk_size_more_jobs_smaller_chunks():


dipy/utils/tests/test_parallel.py
```diff
@@ -119,17 +119,31 @@ def test_auto_ray_chunk_size_memory_scales_with_ram():
...
+
     assert chunk_neg == chunk_pos
-    # Must produce more than one chunk
-    assert chunk_neg < n_vox
+
+    # Multiple chunks are only expected when more than one worker exists
+    if resolved_jobs > 1:
+        assert chunk_neg < n_vox
+    else:
+        assert chunk_neg == n_vox
```
skoudoro: it might be too strict, I would recommend to use `assert chunk_neg <= n_vox`

skoudoro (commented):
Hi @Sunny11092003,

Thank you for doing this fix!. Overall, it looks good. See the tiny comment below to fix.
Then, we can merge

Sunny11092003:
Thanks for the review @skoudoro ! I've updated the assertion to assert chunk_neg <= n_vox and pushed the changes.

skoudoro:
Thanks for the update @Sunny11092003 !

Code format CIs is failing.  see error here: https://github.com/dipy/dipy/actions/runs/26618574869/job/78444114381?pr=4046#step:4:20

make sure to follow our commit message style. Please update commit message

Sunny11092003:
Thanks for the feedback @skoudoro!

I've updated the commit message to follow the project conventions and force-pushed the changes. Please let me know if there's anything else I should adjust.

Sunny11092003:
Hi @skoudoro, just following up on this PR. The requested changes have been addressed and all checks are now passing. Please let me know if any further updates are needed. Thanks!

skoudoro:
All good, merging. thanks @Sunny11092003

PR #4047, opened 2026-05-29
Update CCMetric energy after backward step

TomasGuija:
## Description

`CCMetric.compute_backward()` computed the backward CC energy but did not store it in `self.energy`, unlike `CCMetric.compute_forward()` and the other metric implementations. This PR updates `compute_backward()` to store the computed energy in `self.energy`.

It also adds tests for `CCMetric` and `SSDMetric` to verify that `get_energy()` reports the expected energy after both forward and backward metric steps.

## Motivation and Context

This issue appears to be silent in the standard SyN registration workflow and is not expected to change registration results.

In `SymmetricDiffeomorphicRegistration`, `metric.compute_forward()` is called before `metric.compute_backward()`, so `self.energy` is already initialized before `metric.get_energy()` is called. For `CCMetric`, the forward and backward steps compute different displacement fields, but both evaluate the same cross-correlation energy for the same image pair on the same grid. As a result, the stale forward energy matches the backward energy in the current SyN+CC optimization path.

However, `compute_backward()` should still update `self.energy` for consistency with the metric API and with the other metric implementations. Without this fix, direct use of `CCMetric.compute_backward()` or custom optimization code using the metric interface outside `SymmetricDiffeomorphicRegistration` could observe an unset or stale value from `get_energy()`.

## How Has This Been Tested?

Two tests were added under `dipy/align/tests/test_metrics.py`.

The tests initialize `CCMetric` and `SSDMetric` on randomly generated static and moving images, then run forward and backward steps independently. After each step, they verify that `metric.get_energy()` reports the expected energy:

- For `CCMetric`, the expected value is computed from the local cross-correlation factors, matching the energy accumulation used by `compute_cc_forward_step_2d()` and `compute_cc_backward_step_2d()`.
- For `SSDMetric`, the expected value is the sum of squared differences between the static and moving images.

## Checklist

## Type of Change

- [ ] Documentation update

### DIFF
--- a/dipy/align/tests/test_metrics.py
+++ b/dipy/align/tests/test_metrics.py
@@ -1,7 +1,12 @@
 import itertools
 
 import numpy as np
-from numpy.testing import assert_array_almost_equal, assert_array_equal, assert_raises
+from numpy.testing import (
+    assert_almost_equal,
+    assert_array_almost_equal,
+    assert_array_equal,
+    assert_raises,
+)
 from scipy import ndimage
 
 from dipy.align.metrics import CCMetric, EMMetric, SSDMetric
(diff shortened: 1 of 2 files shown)


TomasGuija:
Hello @skoudoro, I opened this PR to fix the minor bug we discussed, along with an explanation of why the bug was silent. I also added a couple of tests that would fail without the small fix. Please feel free to take a look whenever you have a moment.

dipy/align/tests/test_metrics.py
```diff
@@ -47,6 +52,83 @@ def init_metric(shape, radius):
     metric.initialize_iteration()
 
 
+def setup_metric(metric, static, moving):
+    r"""Initialize a metric with a simple pair of images for one iteration."""
```
skoudoro: `r` not needed

dipy/align/tests/test_metrics.py
```diff
@@ -47,6 +52,83 @@ def init_metric(shape, radius):
...
+    metric.set_static_image(static, None, np.ones(metric.dim), None)
+    metric.set_moving_image(moving, None, np.ones(metric.dim), None)
+    metric.initialize_iteration()
+    return metric
```
skoudoro: this return is strange and a bit redundant.

`setup_metric` modifies `metric` in-place AND `returns it`. I would do one or the other

dipy/align/tests/test_metrics.py
```diff
@@ -47,6 +52,83 @@ def init_metric(shape, radius):
...
+
+
+def cc_energy_from_factors(factors, radius):
+    r"""Compute CC energy as implemented in ``crosscorr.pyx`` step functions.
```
skoudoro: `r` not needed

skoudoro (changes requested):
Hi @TomasGuija,

thank you for this !

Overall, looks good to me. See below minor comments. 

Also, check the CIs. one is failing due to a not standard commit message. You will to update your commit message

TomasGuija:
> Hi @TomasGuija,
> 
> thank you for this !
> 
> Overall, looks good to me. See below minor comments.
> 
> Also, check the CIs. one is failing due to a not standard commit message. You will to update your commit message

Thanks for reviewing! I addressed those issues in the latest commit and renamed the first commit as well.

skoudoro (approved):
All good, merging

PR #1613, opened 2018-08-10
Parallelize diffeomorphic registration

RicciWoo:
Parallelize diffeomorphic registration using prange and local function:
1. parallelized \_compose_vector_fields_3d in vector_fields.pyx;
2. parallelized warp_3d in vector_fields.pyx.

To specify number of threads: add num_threads to SymmetricDiffeomorphicRegistration
> sdr = SymmetricDiffeomorphicRegistration(metric, level_iters, num_threads=4)

PR #1612, opened 2018-08-09
Parallelize affine registration

RicciWoo:
Parallelize affine registration using prange and local function:
1. cythonized \_gradient_* in vector_fields.pyx;
2. typed and made nogil on \_compute_pdfs_* in parzenhist.pyx;
3. parallelized \_gradient_3d using prange and local function.

To specify number of threads: give num_threads to MutualInformationMetric
> metric = MutualInformationMetric(nbins, sampling_prop, num_threads=4)

PR #1599, opened 2018-07-26
Parallelize the first part of 3d affine registration using local function

RicciWoo:
Parallelize _gradient_3d() & _sparse_gradient_3d() in vector_fields.pyx using local function.

You can profile this affine example [sl_aff_reg_3d.py](https://github.com/RicciWoo/registration/blob/master/sl_aff_reg_3d.py) to test the overall performance (multiple calls).
> python -m cProfile -o sl_aff_par_grad_fun.prof sl_aff_reg_3d.py

For testing the performance of each call: use the other repository with timer [aff-par-grad-fun-tim](https://github.com/RicciWoo/dipy/tree/aff-par-grad-fun-tim).

You can run this python code [sl_test_grad_3d.py](https://github.com/RicciWoo/registration/blob/master/sl_test_grad_3d.py) to time a single call to _gradient_3d(). This code is for running on a headless cluster, to run on local computer, please remove the virtual frame buffer "xvfbwrapper".
> python sl_test_grad_3d.py

PR #4049, opened 2026-06-08
NF: Parallelize motion correction

deka27:
## Description

  Add a `num_processes` option to `register_series`, `register_dwi_series`, and `motion_correction` (and expose it on the `MotionCorrectionFlow` workflow) to register the volumes of a series in parallel across a pool of worker processes.
  The per-volume registrations are independent, so they are fanned out with a `multiprocessing` `spawn` pool. Default is `1` (serial), so existing behavior is unchanged.

  The constant arguments (static reference volume, affines, pipeline, mask, optimizer options) are shipped to each worker **once** via a pool initializer rather than being re-pickled with every volume, and the reference-volume special-case is handled in a single shared loop used by both the serial and parallel paths.

  ## Motivation and Context

  Motion correction of a DWI series registers each volume to the reference independently and sequentially, which is slow for large series. Since the per-volume registrations don't depend on one another, they can run concurrently. Adding `num_processes` lets users trade cores for wall-clock time, with no change to the result. The `spawn` start method is used (rather than mutating the global start method via `set_start_method`) so the pool is self-contained and safe on macOS/Windows.

  ## How Has This Been Tested?

  Added `test_register_dwi_series_parallel`, which runs `motion_correction` with `num_processes=1` and `num_processes=2` on an abbreviated dataset and asserts the registered data and affines match (the parallel path must reproduce the serial result).

  ```bash
  conda run -n dipy_test python -m pytest \
    dipy/align/tests/test_api.py::test_register_dwi_series_parallel \
    dipy/align/tests/test_api.py::test_register_series \
    dipy/align/tests/test_api.py::test_register_dwi_series_and_motion_correction \
    dipy/align/tests/test_api.py::test_register_dwi_series_multi_b0 \
    dipy/align/tests/test_api.py::test_register_dwi_to_template
  ```

  All new and existing register_series / register_dwi_series / motion_correction tests pass locally.

  Checklist

  - [x] I have read the CONTRIBUTING guidelines.
  - [x] My code follows the DIPY coding style.

  Type of Change

  - [ ] Documentation update

### DIFF
--- a/dipy/align/_public.py
+++ b/dipy/align/_public.py
@@ -8,6 +8,7 @@ streamlines.
 
 import collections.abc
 from functools import partial
+import multiprocessing as mp
 import numbers
 from pathlib import Path
 import re
@@ -41,6 +42,7 @@ from dipy.testing.decorators import warning_for_keywords
 from dipy.tracking.streamline import set_number_of_points
 from dipy.tracking.utils import transform_tracking_output
 from dipy.utils.logging import logger
+from dipy.utils.multiproc import determine_num_processes
 
 __all__ = [
     "syn_registration",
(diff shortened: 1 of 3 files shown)


skoudoro:
Hi @deka27, 

Thank you for this. 

I do not know if it is a good idea. You might enter into thread racing.  image registration use already openmp which means take many thread.  Imagine you have 1000 of thread, and 1 volume already use those threads, The 2nd volume will slowdown the first one, etc... 

I Believe it is better to improve the performance of the image registration itself than paralellize per volume.  Let me know what you think.

I will reference this work that need to be finished: #1613 #1612 #1599

deka27:
Closing this one. After @skoudoro's feedback it's clear that per-volume parallelization is the wrong layer: the registration metric is already OpenMP-parallel, so a single volume already uses the cores, splitting per-volume (and pinning OMP_NUM_THREADS=1, which is process-global and would hit numpy/scipy's BLAS too) would just redistribute the same work with added overhead rather than speed things up.

PR #4052, opened 2026-06-09
BF: Remove holes and islands from SynthSeg mask

TomasGuija:
## Description

This PR cleans the binary brain mask returned by `SynthSeg.predict`.

Previously, the returned mask was computed directly from the hard segmentation as `labels > 0`. This PR applies DIPY's existing `remove_holes_and_islands` utility to that binary mask before returning it.

## Motivation and Context

`SynthSeg.predict` returns both anatomical labels and a binary brain mask. The segmentation pipeline already applies hole/island cleanup to intermediate segmentation masks, but the final returned binary mask was not explicitly postprocessed.

As a result, the returned mask could contain small disconnected components or holes. Cleaning the final mask makes it more suitable for downstream use cases such as skull stripping, masking, and registration.

## How Has This Been Tested?

I verified the change locally by:

- Running `SynthSeg.predict` on a T1-weighted image.
- Saving the returned brain mask before and after the change.
- Visually comparing the resulting masks.

| Before | After |
|---|---|
| Before mask | After mask |

## Checklist

## Type of Change

- [ ] Documentation update


TomasGuija:
I updated the PR following the discussion.

The SynthSeg `predict` method now has an explicit `return_masks` flag. By default, masks are not returned anymore. If `return_masks=True`, the method returns the brain mask in addition to the predicted labels and label dictionary.

I also added a warning when masks are requested, because the returned brain mask is generated by applying `remove_holes_and_islands` to SynthSeg's prediction. Therefore, some voxels predicted as background (`0`) may be interpreted as brain in the mask. In other words, the union of the predicted label maps is not equal to the returned brain mask.

For now, I kept `return_prob` and `return_masks` separate:

- `return_prob=True` returns the probability maps.
- `return_masks=True` returns the post-processed brain masks together with the label prediction.

Technically, both could be returned at the same time, but that introduces an additional dimensionality/space issue: the probability maps are returned in SynthSeg's internal space (`192 x 192 x 192 x n_labels`), while the labels and masks are recovered back to the original image space. To avoid adding that ambiguity in this PR, I kept the two options separate for now.

dipy/nn/torch/synthseg.py
```diff
@@ -736,13 +756,16 @@ def predict(
...
         if dim == 3:
             labels = labels[0]
-            if not return_prob:
+            if return_masks and not return_prob:
                 masks = masks[0]
 
         if return_prob:
             return labels
-        return labels, self.label_dict, masks
+        if return_masks:
```
skoudoro: this will be a bug is `return_masks=True, return_prob=True` because mask will not be initialized

dipy/nn/torch/synthseg.py
```diff
@@ -736,13 +756,16 @@ def predict(
                 ].astype("int32")
                 temp = recover_img(temp, params_list[i], order=0)
                 labels[i] = np.round(temp).astype(np.int32)
-            masks[i] = (labels[i] > 0).astype(np.int32)
+            if return_masks:
```
skoudoro: this will be a bug is `return_masks=True, return_prob=True` because mask will not be initialized

dipy/nn/torch/synthseg.py
```diff
@@ -736,13 +756,16 @@ def predict(
...
-            if not return_prob:
+            if return_masks and not return_prob:
                 masks = masks[0]
 
         if return_prob:
             return labels
-        return labels, self.label_dict, masks
+        if return_masks:
+            return labels, self.label_dict, masks
+        return labels, self.label_dict
```
skoudoro: here, we break the API by changing the number of output. it needs to be handle differently

dipy/workflows/segment.py
```diff
@@ -786,7 +786,9 @@ def run(
                         "be used for 'synthseg' brain masking."
                     )
                 vol = data[..., 0] if data.ndim == 4 else data
-                _, _, mask_volume = synthseg_model.predict(vol, affine)
+                _, _, mask_volume = synthseg_model.predict(
+                    vol, affine, return_masks=True
```
skoudoro: we have an option name `finalize_mask` below. it is redundant. It needs some thinking.

dipy/nn/tests/test_synthseg.py
```diff
@@ -65,6 +65,38 @@ def test_default_weights_batch():
...
+    with pytest.warns(UserWarning, match="The returned brain masks"):
+        labels, label_dict, masks = synthseg_model.predict(
+            input_arr, np.eye(4), return_masks=True
+        )
+
+    npt.assert_equal(labels.shape, input_arr.shape)
+    npt.assert_equal(masks.shape, input_arr.shape)
+    npt.assert_equal(labels.dtype, np.int32)
+    npt.assert_equal(masks.dtype, np.int32)
+    npt.assert_equal(np.isin(masks, [0, 1]).all(), True)
```
skoudoro: a bit complicated, use `npt.assert_array_equal`

skoudoro (changes requested):
Hi @TomasGuija,

Thank you for updating this. 

See below major comments. It needs some update

dipy/nn/torch/synthseg.py
TomasGuija: True. I solved it in one of my latest commits now. I kept a constant number of returned items for this version.

dipy/workflows/segment.py
TomasGuija: Yes, I removed that one for now, so it is only used in other paths but not on synth seg path.

dipy/nn/tests/test_synthseg.py
TomasGuija: I simplified it, it was indeed over-complicated

pjsjongsung:
Thank you! @TomasGuija I just have two concerns

1. return_prob was returning synthseg model space because it was mainly used for debugging purposes. I think now that we are returning it as a replacement of labels, I think it makes more sense to return it in original space as well. I believe it can be easily done by changing the `order` parameter in the `recover_img` function.

2. Are we making removing holes not optional now? It seems `return_mask` removes holes and islands by default. If so, is this something we want to be consistent with in other segmentation methods (i.e. median otsu)?

We can talk more on the online discussion soon.

TomasGuija:
> Thank you! @TomasGuija I just have two concerns
> 
> 1. return_prob was returning synthseg model space because it was mainly used for debugging purposes. I think now that we are returning it as a replacement of labels, I think it makes more sense to return it in original space as well. I believe it can be easily done by changing the `order` parameter in the `recover_img` function.
> 2. Are we making removing holes not optional now? It seems `return_mask` removes holes and islands by default. If so, is this something we want to be consistent with in other segmentation methods (i.e. median otsu)?
> 
> We can talk more on the online discussion soon.

Hey @pjsjongsung, thank you for the review! Yes, these are concerns I also considered while developing the PR.

1. That is right. Another possibility, since `return_prob` is mainly intended for debugging rather than regular use, would be to expose it through a separate `predict_raw()` method. The current `predict()` method could then call `predict_raw()` internally. Users who need the raw labels or probabilities could call `predict_raw()` directly, while `return_prob` could be removed from `predict()`. However, I am not sure whether this would introduce too many changes to the current API.

2. That is also true. I could make hole and island removal optional. However, I think applying it by default makes sense, since most users would probably expect `return_mask` to directly provide a complete and usable brain mask. A mask containing holes or isolated regions may require additional post-processing before it can be used.

Happy to discuss both points further in the online meeting.

TomasGuija:
@pjsjongsung @skoudoro I made a few changes after our last meeting.

We mentioned making `finalize_mask=True` by default, both in `SynthSeg.predict()` and through `BrainMaskFlow`. However, I noticed there may be a small issue when using the workflow from the CLI.

In `dipy/workflows/base.py`, Boolean arguments seem to be handled like this:

```python
if dtype is bool:
    _kwargs["action"] = "store_true"
    default_ = {arg: False}
    self.set_defaults(**default_)
```

From what I understand, this may mean that Boolean workflow arguments always default to `False` in the CLI, regardless of the default set in the `run()` method.

So we could end up with different defaults:

* `BrainMaskFlow().run(...)` would use `finalize_mask=True`.
* The CLI without `--finalize_mask` might still use `False`.

I have not checked this end-to-end though.

skoudoro:
@pjsjongsung, can you give a look. 

Not sure hat is the last update here. Can you give feedback @TomasGuija ?

Thanks!

skoudoro (approved):
Great!, LGTM, all good, thanks @TomasGuija,

merging

PR #4058, opened 2026-06-16
BF: Update for single slice only in the odf

maharshi-gor:
- Selecting correct slice.

## Description

The ODF on affine based slice changes started picking up multiple voxel if the position was in between two voxels and could reach the threshold.

## Motivation and Context

This would show you multiple ODF glyphs on a single slice. We are required to choose one.

## How Has This Been Tested?

Run FORCE_FIT and open the peaks.pam5 this would show multiple glyphs before the fix and after the fix it should only show the closest one.


skoudoro (approved):
OHBM need. Test need to be added. merging

PR #2962, opened 2023-11-02
[WIP] ODF-Fingerprinting

filipp02:
Dear Authors of DIPY,

I'd like to draw your attention to our implementation of ODF-Fingerprinting. It's a fiber reconstruction technique that identifies fiber directions by comparing the ODFs from the signal with the precomputed ODF-dictionary.
We published a few papers on this method, e.g., 
https://doi.org/10.1016/j.neuroimage.2019.05.024
https://doi.org/10.1002/mrm.29208

Our implementation uses DIPY and hopefully can be integrated with it smoothly. I am aware that one file in our repository (rdsi.py) is a fairly rudimentary translation from MATLAB and it uses a number of modules that could be replaced with others or reimplemented locally, but otherwise I hope the rest of the code would meet your standards.

I'm happy to work with you on improving our code to your satisfaction.

Best,
Patryk

PR #4062, opened 2026-06-22
NF: Add ODF-Fingerprinting (ODF-FP) reconstruction

Atharva-Shah-2298:
## Description

Adds ODF-FP (:footcite:`Baete2019`, :footcite:`Filipiak2022`), a
forward-modeling/fingerprinting reconstruction: the ODF reconstructed from the
data with GQI is matched against a precomputed dictionary of simulated ODF
"fingerprints", and the microstructure of the best-matching fingerprint is
assigned to the voxel.

## Motivation and Context

**This supersedes and closes #2962** — the original 80+-commit branch is
condensed into 5 clean commits on top of current `master`. Now that FORCE uses an optimized search engine, ODFFP has been made so much faster. I have also managed to cache the rotation matrix because it came to my notice that its a O(vertex) operation which will be less than O(Nvox)

## How Has This Been Tested?

- `dict_idx`/`peak_dirs` bit-identical and ODFs identical to ~1e-15 vs. the
  original method on the same sphere.
- Full editable build on current `master`; `test_odffp.py` passes; pre-commit clean. 

Some changes from original implementation.
- Sphere is now dipy native
- Posterior dictionary is left to be added
- Resampler has been changed to be faster
- multi_voxel fit decorator for batched processing
- Class inheritances broken

## Checklist

- [ X] I have read the [CONTRIBUTING](https://github.com/dipy/dipy/blob/master/.github/CONTRIBUTING.md) guidelines.
- [ X] My code follows the [DIPY coding style](https://docs.dipy.org/stable/devel/coding_style_guideline.html).
- [ X] I have added tests that cover my changes (if applicable).
- [ X] All new and existing tests pass locally.
- [ X] I have updated the documentation accordingly (if applicable).

## Type of Change

- [ X] New feature (non-breaking change which adds functionality)
- [ ] Documentation update


Atharva-Shah-2298:
Would love to hear from @filipp02 for feedback and the implementation.

Atharva-Shah-2298:
@skoudoro I think there might be a problem in the CIs because of the examples using 16gb ram for a slice. Waiting to see if the CIs fail..

filipp02:
> Would love to hear from @filipp02 for feedback and the implementation.

Excellent work, Atharva! Thank you for all the improvements you've made to the original code.

deka27:
Works as expected.

ODF-FP speedup — PR 4062 vs PR 2962 (Stanford HARDI, 1M dict, 16 threads)

  | Scenario | Stage | PR 2962 | PR 4062 | Speedup |
  |---|---|--:|--:|--:|
  | Full brain (145,841 vox) | Reconstruction | 1108.8 s | 252.2 s | **4.40×** |
  | Full brain | Total (gen + recon) | 1180.0 s | 335.3 s | 3.52× |
  | Single slice (4,102 vox) | Reconstruction | 36.4 s | 7.4 s | **4.90×** |
  | Both | Dict generate (one-time) | ~67–71 s | ~82–83 s | 0.85× |

Atharva-Shah-2298:
Can expect ~10 X speedup from original PR. ODF trace matching now moved to SH space with fp32 which reduces the parameters ~9 times (Still bit identical). The core issue that remains is that there are 99% 3 fibers in the dictionary and that causes some areas like CC to have 3 fibers on default penalty or directly 0 fibers on higher penalties. Might want to look into resampling the parameters or introduce dispersion @filipp02 .

dipy/reconst/odffp.py
```diff
@@ -0,0 +1,771 @@
...
+from dipy.reconst.gqi import GeneralizedQSamplingModel
+from dipy.reconst.multi_voxel import multi_voxel_fit
+from dipy.reconst.odffp_matching import accumulate_block, finalize_match
+from dipy.reconst.shm import real_sh_descoteaux, sf_to_sh, sh_to_sf, smooth_pinv
+
+DEFAULT_RECON_EDGE = 1.2
+DEFAULT_DICT_EDGE = 1.2
+
+DEFAULT_FIT_PENALTY = 1e-5
+MAX_FIT_PENALTY = 0.1
```
skoudoro: Those variables are used only once. 

I do not think there are needed to be global, they can just be in the init class of `OdffpDictionary`

dipy/reconst/odffp.py
```diff
@@ -0,0 +1,771 @@
...
+# Number of dictionary fingerprints matched per block. The matching is streamed
+# over blocks so the full (n_voxels x n_dict) similarity is never materialized.
+# A large block keeps the BLAS matmul efficient (one big GEMM beats many small
+# ones) while still bounding the (n_voxels x block) tile for very large
+# dictionaries; ~5e5 runs at materialize speed at roughly half the peak memory.
+MATCH_BLOCK_SIZE = 524288
+
+
+def _default_sphere():
+    return get_sphere(name="repulsion724")
```
skoudoro: what is the goal of this? 

I believe we should avoid this. DIPY has a default sphere already (`from dipy.data import default_sphere`).

Why not using it ? it is confusing.

dipy/reconst/odffp.py
```diff
@@ -0,0 +1,771 @@
...
+
+    These reproduce :func:`~dipy.reconst.shm.sf_to_sh`/``sh_to_sf`` with no
+    regularization, so that ``inv_B`` and ``B`` match the SH fit and evaluation
+    used by :meth:`OdffpModel.resample_odf`.
+    """
+    basis, _, order = real_sh_descoteaux(
+        sh_order_max, sphere.theta, sphere.phi, legacy=False
+    )
+    inv_basis = smooth_pinv(basis, np.sqrt(0.0) * (-order * (order + 1)))
+    return basis, inv_basis
```
skoudoro: if it reproduce, improve original code with a smaller PR.  Let's not increase maintenance

dipy/reconst/odffp.py
```diff
@@ -0,0 +1,771 @@
...
+    if sin_sqr != 0:
+        skew = np.array(
+            [
+                [0, -cross[2], cross[1]],
+                [cross[2], 0, -cross[0]],
+                [-cross[1], cross[0], 0],
+            ]
+        )
+        rotation += skew + np.dot(skew, skew) * (1 - np.dot(direction, pole)) / sin_sqr
+    return rotation
```
skoudoro: Not sure we should keep this here. Thinking..

dipy/reconst/odffp.py
```diff
@@ -0,0 +1,771 @@
...
+        self.peak_dirs = None
+        self.micro = None
+        self.ratio = None
+        self.peaks_per_voxel = None
+        if dict_file is not None:
+            self.load(dict_file)
+
+    # -- signal simulation ----------------------------------------------------
+
+    @staticmethod
```
skoudoro: why static ? creation a function if it is does not need to be in a class

dipy/reconst/odffp.py
```diff
@@ -0,0 +1,771 @@
...
+        return np.array([np.min(values), np.max(values)])
+
+    def _random_fraction_volumes(self, p_iso, p_fib, peaks_per_voxel):
+        fraction_volumes = np.zeros(peaks_per_voxel + 1)
+
+        # Lower bounds are hard limits; variability stays in [0, p_random_max].
+        p_random_max = 1 - (p_iso[0] + peaks_per_voxel * p_fib[0])
+        p_random = np.hstack(
+            (
+                np.random.uniform(0, p_iso[1] - p_iso[0]),
```
skoudoro: does this will give different results depending on the seeds? CI's will not like it and generate strange failure.

dipy/reconst/odffp.py
```diff
@@ -0,0 +1,771 @@
...
+        # Penalty group of each fingerprint (negative -> ignored in matching).
+        n_fibers = dictionary.peaks_per_voxel
+        group = np.where(n_fibers < 0, -1, np.maximum(0, n_fibers - 1))
+        self._group = np.ascontiguousarray(group, dtype=np.intp)
+        self._n_groups = int(self._group.max()) + 1
+
+        # Resampling operators, cached by main-peak vertex across voxels/fits.
+        self._operators = {}
+
+    @staticmethod
```
skoudoro: same comment for static above

dipy/reconst/odffp.py
```diff
@@ -0,0 +1,771 @@
...
+        )  # (n_dict, n_sh)
+
+        # Penalty group of each fingerprint (negative -> ignored in matching).
+        n_fibers = dictionary.peaks_per_voxel
+        group = np.where(n_fibers < 0, -1, np.maximum(0, n_fibers - 1))
+        self._group = np.ascontiguousarray(group, dtype=np.intp)
+        self._n_groups = int(self._group.max()) + 1
+
+        # Resampling operators, cached by main-peak vertex across voxels/fits.
+        self._operators = {}
```
skoudoro, line 672: this cache is edited in parallel. For sure it will create corrupt data and incorrect result. 

either you need to lock (1thread at the time can write on it ) or you do not cache or redesign differently

dipy/reconst/tests/test_odffp.py
```diff
@@ -0,0 +1,247 @@
...
+    odf_dict = OdffpDictionary(gtab)
+    odf_dict.generate(
+        dict_size=dict_size, max_peaks_num=max_peaks_num, max_chunk_size=dict_size
+    )
+    return odf_dict
+
+
+def _single_voxel(gtab, mevals, angles, fractions):
+    signal, _ = multi_tensor(gtab, mevals, angles=angles, fractions=fractions, snr=None)
+    return signal
```
skoudoro: I see some helper function that can be reused in other place. so they need to move somewhere else and check if they do not exist already

dipy/reconst/odffp.py
```diff
@@ -0,0 +1,771 @@
...
+# ones) while still bounding the (n_voxels x block) tile for very large
+# dictionaries; ~5e5 runs at materialize speed at roughly half the peak memory.
+MATCH_BLOCK_SIZE = 524288
+
+
+def _default_sphere():
+    return get_sphere(name="repulsion724")
+
+
+def _rotation_to_pole(direction, pole=(0.0, 0.0, 1.0)):
```
skoudoro: can you check when  when direction is exactly antipodal to pole. 

something weird seems to happen

dipy/reconst/odffp_matching.pyx
```diff
@@ -0,0 +1,176 @@
...
+
+    with nogil, parallel():
+        for v in prange(n_vox, schedule='static'):
+            best_score = -INFINITY
+            best_j = -1
+            for g in range(n_groups):
+                if group_idx[v, g] < 0:
+                    continue
+                if penalty > 0:
+                    score = log(group_best[v, g]) - 2.0 * penalty * g
```
skoudoro, line 139: this can have `score=nan` if `drop_negative_odf=False` (constructor option). 

log(negative) so something to fix because it will always fails best_score and have higher impact.

I would check `group_best[v, g]`

skoudoro (changes requested):
Hi @Atharva-Shah-2298,

Thank you for updating this. I did not try to run the code yet. trying to look at the overall design first. See below my comments

furthermore:
- docstring is missing on every function. To add
- remove all `staticmethod`. it means it can be a simple function that can be reused by something else
- make sure to not duplicate code, reinvent what is already existing and bring more maintenance than needed. Currently, this code can be a lot simplified it seems. 

Overall this is quite a large PR. Thank you for this work again

dipy/reconst/odffp.py
Atharva-Shah-2298: Agreed!. Put in appropriate places

dipy/reconst/odffp.py
Atharva-Shah-2298: The default sphere is an issue as it is a hemisphere so I have to use the sphere for default sphere that is the repulsion724 so now importing that directly.

dipy/reconst/odffp.py
Atharva-Shah-2298: Agreed. Slight variation. Baked directly into the code

dipy/reconst/odffp.py
Atharva-Shah-2298: Written out after copy should not make this happen now.

dipy/reconst/tests/test_odffp.py
Atharva-Shah-2298: Just reused a different testcase import from other module

dipy/reconst/odffp_matching.pyx
Atharva-Shah-2298: Modified to have a fallback for only the positive ones.

dipy/reconst/odffp.py
Atharva-Shah-2298: This can be either in core.geometry or a function here. Your call!

arokem:
Just catching up here. This looks great, and this is probably a discussion for separate work, but while you are at it, I was wondering: could this method have a notion of a separate "predict" function, similar to the prediction methods that we have in other recon methods?

Atharva-Shah-2298:
> Just catching up here. This looks great, and this is probably a discussion for separate work, but while you are at it, I was wondering: could this method have a notion of a separate "predict" function, similar to the prediction methods that we have in other recon methods?

Ofcourse, Its just exposing the signal index. We can do it but I am not sure what do the authors think about this. @filipp02 Are there any caveats or issues that you know with getting the SM signal out as a prediction?

Atharva-Shah-2298:
> Just catching up here. This looks great, and this is probably a discussion for separate work, but while you are at it, I was wondering: could this method have a notion of a separate "predict" function, similar to the prediction methods that we have in other recon methods?

@arokem One caveat I think of is that this is a gtab specific method so it will diverge from other recon methods on consuming different gtabs.  But the typical signal prediction can be done.

skoudoro:
Any update @filipp02 and @Atharva-Shah-2298 ? 

IT would be great to have this on the future release planned end of this month(September) or beginning of October

Atharva-Shah-2298:
We just have to take a call on the predict function and I am done from my side. We can wait for @filipp02 to see if there is any issue with the current implementation

arokem:
If that's the only thing holding this up then I would not wait for it and I would suggest punting that to future work when someone actually needs this prediction functionality.

Atharva-Shah-2298:
> If that's the only thing holding this up then I would not wait for it and I would suggest punting that to future work when someone actually needs this prediction functionality.

Waiting for the original authors ( @filipp02 and Steven Baete) of the method to comment on the implementation if any and then we can go ahead and merge it.

Atharva-Shah-2298:
Okay so I added predict function. According to me the dict should give you a signal projection of biophysical space so in theory it should be okay. The workflow is in as well. Thanks for the reviews @arokem @skoudoro @deka27 & the author @filipp02

skoudoro:
is there any additional comments here? @arokem  ? @deka27 ? @filipp02 ?

I will try a last time locally and if no additional comments, I will merge by the end of the week.

dipy/reconst/odffp_matching.pyx
```diff
@@ -0,0 +1,186 @@
...
+                best_j = fallback_j
+            best[v] = best_j
+
+    if num_threads is not None:
+        restore_default_num_threads()
+
+    return np.asarray(best)
+
+
+def select_best_match(double[:, ::1] similarity, cnp.npy_intp[::1] group,
```
skoudoro: This function is used only by the tests. 

not use and bring more maintenance.... Can you remove and the associated tests?

doc/examples/reconst_odffp.py
```diff
@@ -0,0 +1,150 @@
...
+# Build the dictionary of ODF fingerprints. ``OdffpDictionary`` simulates the
+# diffusion signal of random multi-compartment configurations on a symmetric
+# sphere and reconstructs their ODFs with GQI. In practice the dictionary holds
+# on the order of one million fingerprints, which is the default; generating it
+# takes about a minute and a few gigabytes of memory.
+
+sphere = get_sphere(name="repulsion724")
+
+odf_dict = OdffpDictionary(gtab, sphere=sphere)
+odf_dict.generate()
```
skoudoro, line 71: Note to myself:  check time and size here. might not be reasonable for a tutorial ( odf 2.9 GB f64 ??)

skoudoro (commented):
Here is my last review. 

Now I just need multiple run before the final approve and merge

doc/examples/reconst_odffp.py
Atharva-Shah-2298: The only way to tackle this is use less number of elements in the dictionary or use a smaller sphere but if FORCE was fine, I would hope this is too?

maharshi-gor: Or use a segment of the brain?
You cand do something like this?
https://docs.dipy.org/1.12.0/examples_built/reconstruction/reconst_sfm.html

Atharva-Shah-2298: I would rather not. Users not so familiar with reconstruction might understand better if they see in total a slice for better  context

doc/examples/reconst_odffp.py
```diff
@@ -0,0 +1,150 @@
...
+# coefficients. The stored ODFs and the peak values follow the quantitative
+# anisotropy convention: the isotropic floor of every ODF is removed and the
+# volume is scaled so that the largest peak is 1, so that the size of an ODF
+# reflects its anisotropy. It is written to a PAM5 file with ``save_pam``.
+
+from dipy.io.peaks import save_pam
+
+peaks = odffp_peaks(fit)
+save_pam("odffp_peaks.pam5", peaks, affine=affine)
+
```
maharshi-gor: Just check you already have odf coefficients in pam5 file. Or just store the coefficient in nii.gz as that is how handle them we do not use direct sphere based odfs.


```
from dipy.viz.skyline.app import skyline

skyline(
    visualizer_type="stealth",
    sh_coeffs=[(sh_coeffs, affine, "odf")],
    out_dir="doc/examples/_static",
    out_stealth_png="odf_glyphs.png",
)
```

This much code should store the odf image.

Atharva-Shah-2298: So keeping legacy=False works fine for dipy-side reconstruction, but skyline's default "descoteaux07" path is legacy=True, so the stored coefficients would render incorrectly in skyline.

doc/examples/reconst_odffp.py
Atharva-Shah-2298: I think should be resolved with legacy SH update in the output. Didnt do it in the viz to not break the tracking methods as they stil rely on legacy SH basis.

doc/examples/reconst_odffp.py
Atharva-Shah-2298: Done @maharshi-gor . Resolve if satisfactory

PR #4067, opened 2026-06-25
MI implementation for SyN registration.

TomasGuija:
## Description

This PR represents one of my main efforts for GSoC 2026: implementing the Mutual Information (MI) metric for SyN registration.

## Motivation and Context

DIPY's SyN registration already supports several similarity metrics, such as CC, EM, and SSD. However, Mutual Information, one of the most important metrics for multimodal medical image registration and already available in other registration frameworks, was still missing for SyN.

In this PR, I worked on a functional implementation of the MI metric for DIPY's SyN registration.

A new `SimilarityMetric` class, named `MIMetric`, was added in `metrics.py`, providing an interface for using the new implementation with `SymmetricDiffeomorphicRegistration`. In addition, a new local-support path was implemented in `parzenhist.pyx`, providing the necessary methods to compute the MI derivative between two images with respect to a dense displacement field.

## How Has This Been Tested?

While further benchmarking is still planned, following the work available in my benchmarking branch [`registration-benchmark`](https://github.com/TomasGuija/dipy/tree/registration-benchmark), I tested the implementation on a single multimodal image pair: T1 and B0 images obtained using DIPY's `read_syn_data()`.

Both images were skull-stripped using SynthSeg and rigidly prealigned before the comparison. The quantitative comparison then focused exclusively on the SyN registration step.

The quantitative results are summarized in the following table:

| Method | NCC | NMI | Runtime (s) | Runtime (min) |
|---|---:|---:|---:|---:|
| Baseline | -0.44152 | 1.07815 | — | — |
| ANTs MI-SyN | -0.59649 | 1.11984 | 121.27 | 2.02 |
| DIPY MI-SyN | -0.58943 | 1.12253 | 292.91 | 4.88 |

So far, the implementation appears to work as expected. It is slower than ANTs in this initial experiment, although it remains to be studied whether this difference comes from the metric implementation itself or from differences in the SyN optimization algorithms.

## Checklist

## Type of Change

- [ ] Documentation update

### DIFF
--- a/dipy/align/metrics.py
+++ b/dipy/align/metrics.py
@@ -261,8 +261,16 @@ class MIMetric(SimilarityMetric):
         Computes image gradients in physical coordinates and prepares the
         histogram quantities needed to evaluate mutual information updates.
         """
+        if self.static_image.dtype != self.moving_image.dtype:
+            raise ValueError("Static and moving images must have the same dtype.")
+
+        dtype = np.asarray(self.moving_image).dtype
+
+        if dtype not in (np.float32, np.float64):
+            raise ValueError("MIMetric expects float32 or float64 images.")
+
         self.gradient_moving = np.empty(
-            shape=self.moving_image.shape + (self.dim,), dtype=np.float32
+            shape=self.moving_image.shape + (self.dim,), dtype=dtype
         )
         for i, grad in enumerate(gradient(self.moving_image)):
             self.gradient_moving[..., i] = grad
@@ -274,7 +282,7 @@ class MIMetric(SimilarityMetric):
             self.reorient_vector_field(self.gradient_moving, self.moving_direction)
 
         self.gradient_static = np.empty(
-            shape=self.static_image.shape + (self.dim,), dtype=np.float32
+            shape=self.static_image.shape + (self.dim,), dtype=dtype
         )
         for i, grad in enumerate(gradient(self.static_image)):
             self.gradient_static[..., i] = grad
--- a/dipy/align/metrics.py
+++ b/dipy/align/metrics.py
@@ -9,6 +9,7 @@ from scipy import ndimage
 from dipy.align import (
     crosscorr as cc,
     expectmax as em,
+    parzenhist as ph,
     sumsqdiff as ssd,
     vector_fields as vfu,
 )
@@ -210,6 +211,150 @@ class SimilarityMetric:
         image matches the current static image. Lower energy is typically
         considered better (optimizer minimizes).
         """
+
+
+class MIMetric(SimilarityMetric):
+    @warning_for_keywords()
+    def __init__(self, dim, *, nbins=32, smooth=0.0):
+        r"""Mutual Information metric for SyN registration.
+
+        Similarity metric based on a Parzen-window estimate of the joint
+        intensity distribution. The metric computes dense displacement update
+        fields through the ``SimilarityMetric`` API.
+
+        Parameters
+        ----------
+        dim : int (either 2 or 3)
+            The dimension of the image domain.
+        nbins : int, optional
+            Number of bins to use for the joint intensity histogram.
+        smooth : float, optional
+            Standard deviation of the Gaussian smoothing kernel to apply to
+            update fields before they are returned.
+        """
+        super().__init__(dim)
+        self.nbins = nbins
+        self.smooth = smooth
+        self.energy = np.inf
+        self.gradient_static = None
+        self.gradient_moving = None
+        self.forward_histogram = ph.ParzenJointHistogram(nbins)
+        self.backward_histogram = ph.ParzenJointHistogram(nbins)
+        self._connect_functions()
+
+    def _connect_functions(self):
+        r"""Assign the methods to be called according to the image dimension
+
+        Assigns the appropriate functions to be called for vector field
+        reorientation according to the dimension of the input images.
+        """
+        if self.dim == 2:
+            self.reorient_vector_field = vfu.reorient_vector_field_2d
+        elif self.dim == 3:
+            self.reorient_vector_field = vfu.reorient_vector_field_3d
+        else:
+            raise ValueError(f"MI Metric not defined for dim. {self.dim}")
+
+    def initialize_iteration(self):
+        r"""Prepares the metric to compute one displacement field iteration.
+
+        Computes image gradients in physical coordinates and prepares the
+        histogram quantities needed to evaluate mutual information updates.
+        """
+        self.gradient_moving = np.empty(
+            shape=self.moving_image.shape + (self.dim,), dtype=np.float32
+        )
+        for i, grad in enumerate(gradient(self.moving_image)):
+            self.gradient_moving[..., i] = grad
+
+        # Convert moving image's gradient field from voxel to physical space
+        if self.moving_spacing is not None:
+            self.gradient_moving /= self.moving_spacing
+        if self.moving_direction is not None:
+            self.reorient_vector_field(self.gradient_moving, self.moving_direction)
+
+        self.gradient_static = np.empty(
+            shape=self.static_image.shape + (self.dim,), dtype=np.float32
+        )
+        for i, grad in enumerate(gradient(self.static_image)):
+            self.gradient_static[..., i] = grad
+
+        # Convert static image's gradient field from voxel to physical space
+        if self.static_spacing is not None:
+            self.gradient_static /= self.static_spacing
+        if self.static_direction is not None:
+            self.reorient_vector_field(self.gradient_static, self.static_direction)
+
+        self.forward_histogram.setup(self.static_image, self.moving_image)
+        self.forward_histogram.update_pdfs_dense(self.static_image, self.moving_image)
+        self.forward_histogram.mi_weights = np.zeros_like(self.forward_histogram.joint)
+        self.energy = self.forward_histogram.compute_mi(local_support=True)
+
+        self.backward_histogram.setup(self.moving_image, self.static_image)
+        self.backward_histogram.update_pdfs_dense(self.moving_image, self.static_image)
+        self.backward_histogram.mi_weights = np.zeros_like(
+            self.backward_histogram.joint
+        )
+        self.backward_histogram.compute_mi(local_support=True)
+
+    def compute_forward(self):
+        r"""Computes one step bringing the moving image towards the static.
+
+        Computes the update displacement field to be used for registration of
+        the moving image towards the static image.
+        """
+        displacement = np.zeros(
+            shape=self.static_image.shape + (self.dim,),
+            dtype=self.gradient_moving.dtype,
+        )
+        self.forward_histogram.compute_dense_mi_update(
+            self.static_image,
+            self.moving_image,
+            self.gradient_moving,
+            displacement,
+        )
+        for i in range(self.dim):
+            displacement[..., i] = ndimage.gaussian_filter(
+                displacement[..., i], self.smooth
+            )
+        return displacement
+
+    def compute_backward(self):
+        r"""Computes one step bringing the static image towards the moving.
+
+        Computes the update displacement field to be used for registration of
+        the static image towards the moving image.
+        """
+        displacement = np.zeros(
+            shape=self.static_image.shape + (self.dim,),
+            dtype=self.gradient_static.dtype,
+        )
+        self.backward_histogram.compute_dense_mi_update(
+            self.moving_image,
+            self.static_image,
+            self.gradient_static,
+            displacement,
+        )
+        for i in range(self.dim):
+            displacement[..., i] = ndimage.gaussian_filter(
+                displacement[..., i], self.smooth
+            )
+        return displacement
+
+    def get_energy(self):
+        r"""The numerical value assigned by this metric to the current image pair
+
+        Returns the Mutual Information energy computed for the current image
+        pair.
+        """
+        return self.energy
+
+    def free_iteration(self):
+        r"""Frees the resources allocated during initialization"""
+        del self.gradient_static
+        del self.gradient_moving
+        self.forward_histogram.mi_weights = None
+        self.backward_histogram.mi_weights = None
 
 
 class CCMetric(SimilarityMetric):
--- a/dipy/align/metrics.py
+++ b/dipy/align/metrics.py
@@ -253,8 +253,15 @@ class MIMetric(SimilarityMetric):
         Computes image gradients in physical coordinates and prepares the
         histogram quantities needed to evaluate mutual information updates.
         """
+        if self.static_image.dtype != self.moving_image.dtype:
+            raise ValueError("Static and moving images must have the same dtype.")
+
+        dtype = np.asarray(self.moving_image).dtype
+        if dtype not in (np.float32, np.float64):
+            raise ValueError("MIMetric expects float32 or float64 images.")
+
         self.gradient_moving = np.empty(
-            shape=self.moving_image.shape + (self.dim,), dtype=np.float32
+            shape=self.moving_image.shape + (self.dim,), dtype=dtype
         )
         for i, grad in enumerate(gradient(self.moving_image)):
             self.gradient_moving[..., i] = grad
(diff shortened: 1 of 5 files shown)


dipy/align/metrics.py
```diff
@@ -212,6 +213,143 @@ def get_energy(self):
         """
 
 
+class MIMetric(SimilarityMetric):
+    @warning_for_keywords()
```
skoudoro: not needed

TomasGuija:
Hey @skoudoro, I made a few changes, including:

* Removing the larger four-way `dtype`/dimension dispatch from `compute_dense_mi_update`
* Removing the unnecessary keyword warning
* Adding some extra tests (coming soon, still working on them)

While working on this, I noticed a couple of related things.

I introduced a new dtype branch in `update_pdfs_dense`, but it is very similar to the existing ones in `update_gradient_dense` and `update_gradient_sparse`. Since those are already part of the current implementation, I thought it would be better to leave them unchanged for now. I only refactored `compute_dense_mi_update`, as that was the largest and hardest dispatch block to read.

I also noticed that some docstrings use hardcoded 3D shapes even though the methods support both 2D and 3D. However, this is also present in existing methods. For example, `update_gradient_dense` supports both dimensions, but its docstring only describes the 3D case.

Would it make sense to handle all of those docstring updates together in a separate documentation PR?

TomasGuija:
Hey @skoudoro, something else I noticed. Last meeting, we discussed some apparent normalization differences between the 2D and 3D implementations of `_compute_pdfs_dense_and_local_derivatives`.

I found that these differences are inherited from the existing `_compute_pdfs_dense_2d`, `_compute_pdfs_dense_3d`, and sparse implementations in the current DIPY codebase:

- Dense 2D normalizes the joint PDF using `valid_points`.
- Dense 3D and sparse normalize it using `total_sum`.
- The static marginal is normalized using `valid_points` in all cases.

In principle, `total_sum` and `valid_points` should be equal because each valid sample’s cubic B-spline contributions should sum to one.

I could investigate this further and propose a small normalization cleanup, either in this PR or in a separate one. `total_sum` appears potentially redundant, so it may be possible to remove it, but I would first verify the assumptions. So it's up to you: I can change the normalization everywhere in `parzenhist.pyx` on this PR, or create a minimal separate one for that specific fix. Anyhow, I think semantically everything is correct as it is right now, only a bit inconsistent.

dipy/align/parzenhist.pyx
```diff
@@ -1245,6 +1601,347 @@ cdef _joint_pdf_gradient_sparse_3d(double[:] theta, Transform transform,
...
+    """
+    cdef:
+        cnp.npy_intp nrows = update_field.shape[0]
+        cnp.npy_intp ncols = update_field.shape[1]
+        cnp.npy_intp i, j, offset_id, idx, r, c, valid_points
+        double weight, norm_factor
+
+    update_field[...] = 0
+
+    with nogil:
```
skoudoro: why you need `nogil` ? why not at the function level with `noexcept nogil`?

dipy/align/parzenhist.pyx
```diff
@@ -1245,6 +1601,347 @@ cdef _joint_pdf_gradient_sparse_3d(double[:] theta, Transform transform,
...
+    cdef:
+        cnp.npy_intp nslices = update_field.shape[0]
+        cnp.npy_intp nrows = update_field.shape[1]
+        cnp.npy_intp ncols = update_field.shape[2]
+        cnp.npy_intp k, i, j, offset_id, idx, r, c, valid_points
+        double weight, norm_factor
+
+    update_field[...] = 0
+
+    with nogil:
```
skoudoro: same question as above

dipy/align/parzenhist.pyx
```diff
@@ -1292,6 +1989,51 @@ def compute_parzen_mi(double[:, :] joint,
     return metric_value
 
 
+def compute_parzen_mi_weights(double[:, :] joint,
```
skoudoro, line 2022: I need to check for performance but this function could call `compute_parzen_mi`

skoudoro (changes requested):
Hi @TomasGuija, 

Overall, looks good.

See below few comments.

Also, overall, I prefer to avoid `cpdef`. They do not always do a good job. make sure to have only `cdef` or `def`, easier to identify

skoudoro:
> Would it make sense to handle all of those docstring updates together in a separate documentation PR?

yep, make sense

> I could investigate this further and propose a small normalization cleanup, either in this PR or in a separate one. total_sum appears potentially redundant, so it may be possible to remove it, but I would first verify the assumptions. So it's up to you: I can change the normalization everywhere in parzenhist.pyx on this PR, or create a minimal separate one for that specific fix. Anyhow, I think semantically everything is correct as it is right now, only a bit inconsistent.

Separate PR, fine, thanks !

dipy/align/parzenhist.pyx
TomasGuija: Both functions do compute the same MI value and factors, but while ``compute_parzen_mi`` accumulates the affine parameter gradient, ``compute_parzen_mi_weights`` stores  the same factors for the voxelwise SyN update. Directly calling ``compute_parzen_mi`` would introduce another unnecessary loop as it is right now, but a possible approach would be to add an optional ``mi_weights`` output, so the same method can be used for both the affine and the SyN path. I think that makes the most sense.

skoudoro: make sense and agree

dipy/align/metrics.py
```diff
@@ -212,6 +213,142 @@ def get_energy(self):
...
+    def compute_forward(self):
+        r"""Computes one step bringing the moving image towards the static.
+
+        Computes the update displacement field to be used for registration of
+        the moving image towards the static image.
+        """
+        displacement = np.zeros(
+            shape=self.static_image.shape + (self.dim,),
+            dtype=self.gradient_moving.dtype,
+        )
```
skoudoro: this should slowdown the whole process. You reallocate an array at every iteration which is a lot. 

it might be the reason of your difference with Ants. 

Create a persitent buffer like `self.joint/self.smarginal/self.mmarginal` in parzen histogram, which are set up once and can be reshape if needed. 

you will save time for sure.

dipy/align/metrics.py
```diff
@@ -212,6 +213,142 @@ def get_energy(self):
...
+    def compute_backward(self):
+        r"""Computes one step bringing the static image towards the moving.
+
+        Computes the update displacement field to be used for registration of
+        the static image towards the moving image.
+        """
+        displacement = np.zeros(
+            shape=self.static_image.shape + (self.dim,),
+            dtype=self.gradient_static.dtype,
+        )
```
skoudoro: this should slowdown the whole process. You reallocate an array at every iteration which is a lot. 

it might be the reason of your difference with Ants. 

Create a persitent buffer like `self.joint/self.smarginal/self.mmarginal` in parzen histogram, which are set up once and can be reshape if needed. 

you will save time for sure.

skoudoro (changes requested):
Hi @TomasGuija 

Overall, looks good ! excited to add this feature to our registration framework! 

See below the last change to do

dipy/align/parzenhist.pyx
TomasGuija: Hey, just implemented that. What I did is, I added `mi_weights` as an optional output of `compute_parzen_mi`, so the MI value, affine gradient, and bin weights can all be computed by the same method. 

I kept `compute_parzen_mi_weights` as a small wrapper:

```python
def compute_parzen_mi_weights(joint, smarginal, mmarginal, mi_weights):
    return compute_parzen_mi(
        joint, None, smarginal, mmarginal, None, mi_weights
    )
```

The wrapper does not duplicate any computation. It only provides a cleaner interface for the SyN path, avoiding calls containing several positional `None` arguments while preserving the existing `compute_parzen_mi` argument order.

dipy/align/metrics.py
TomasGuija: Hey, just saw to this. I think the solution is clean enough. The code now uses persistent buffers stored in `MIMetric`. The metric keeps separate forward and backward buffers and reuses those across iterations. They are only reallocated when the image shape changes between different pyramid levels, or if the input dtype changes (I don't think that should happen usually, but I decided to add it as a guard to prevent possible errors).

dipy/align/parzenhist.pyx
```diff
@@ -356,6 +382,88 @@ class ParzenJointHistogram:
...
+        if static.dtype not in [np.float32, np.float64]:
+            raise ValueError("Images must be of dtype float32 or float64.")
+
+        if not self.setup_called:
+            self.setup(static, moving, smask=smask, mmask=mmask)
+
+        n_offsets = 2 * self.padding
+        local_derivative_by_parzen_bin = np.zeros(
+            (n_offsets,) + moving.shape + (dim,), dtype=mgradient.dtype)
+        joint_pdf_index = np.zeros(moving.shape, dtype=np.intp)
```
skoudoro: same problem here. new allocation at every iteration for `local_derivative_by_parzen_bin` and `joint_pdf_index`. 

Please, update like before:  cache buffer on ParzenJointHistogram, reallocate only on shape/dtype change

skoudoro (changes requested):
Hi @TomasGuija, 

My last comment on this PR below.

After that, please update your benchmark and table to see if those update had an impact. 

Thanks !

dipy/align/metrics.py
```diff
@@ -212,6 +213,149 @@ def get_energy(self):
...
+        the moving image towards the static image.
+        """
+        displacement = self.forward_displacement
+        self.forward_histogram.compute_dense_mi_update(
+            self.static_image,
+            self.moving_image,
+            self.gradient_moving,
+            displacement,
+        )
+        self.energy = self.forward_histogram.metric_val
```
skoudoro: I have a doubt here.

I wonder if it is not `self.energy = -self.forward_histogram.metric_val`.

it seems that MI increases as fit improves. However, the other metrics decrease toward convergence.

Check `_get_energy_derivative`. 

also, check at the imaffine version:
https://github.com/dipy/dipy/blob/90af8ba0934e4175157960f3163373187e32c963/dipy/align/imaffine.py#L851

dipy/align/metrics.py
```diff
@@ -212,6 +213,149 @@ def get_energy(self):
...
+        the static image towards the moving image.
+        """
+        displacement = self.backward_displacement
+        self.backward_histogram.compute_dense_mi_update(
+            self.moving_image,
+            self.static_image,
+            self.gradient_static,
+            displacement,
+        )
+        self.energy = self.backward_histogram.metric_val
```
skoudoro: same doubt here.

I wonder if it is not `self.energy = -self.backward_histogram.metric_val`.

it seems that MI increases as fit improves. However, the other metrics decrease toward convergence.

Check `_get_energy_derivative`. 

Also, check at the imaffine version:
https://github.com/dipy/dipy/blob/90af8ba0934e4175157960f3163373187e32c963/dipy/align/imaffine.py#L851

skoudoro (changes requested):
Hi @TomasGuija,

Overall, it looks good and I believe it is almost ready to be merged. Before that, can you re-run some benchmark? 

It would be great to see if the update above had an impact or not. Thank you for your feedback.

Also, I have a doubt about 2 lines, it would be great to check them because it might be a problematic bug. See below.

dipy/align/tests/test_imwarp.py
```diff
@@ -709,6 +652,98 @@ def get_synthetic_warped_circle(nslices):
...
+    b = 0.1
+    m = 4
+
+    image = np.load(fname)
+    moving, static = get_warped_stacked_image(image, nslices, b, m)
+
+    # Configure the metric
+    metric = metrics.MIMetric(2, nbins=16, smooth=1.0)
+
+    # Configure and run the Optimizer
```
pjsjongsung, line 730: I get there are some other tests that are like this, but is there a reason behind the 2D test parameters for the SyN algorithm using default but 3D test parameters are set?

dipy/align/metrics.py
TomasGuija: Absolutely, good catch. I just updated it along with the related tests. I think it only affects convergence, which is why the optimization was working despite the bug. 

In relation to this, I believe in the case of cross correlation, the cython code already returns the negative energy, while the mutual infomation parzenhist script comptues the original MI value. For cross correlation we use the computed energy as it is, while as you pointed out, for MI we do have to negate it. 

I think it would be nice to be coherent in the near future and make both frameworks use the same convention, otherwise it may be a bit confusing, mostly for future contributors.

dipy/align/tests/test_imwarp.py
TomasGuija: Good point. Not really that I know of. I just mirrored the existing tests for CC. However, even though the 3D parameters are set, I just realized they are mostly the defaults. Only `opt_tol` differs, which I don't think should actually affect the test much. I think we just might as well set all parameters to default for all tests.

TomasGuija:
> Hi @TomasGuija,
> 
> Overall, it looks good and I believe it is almost ready to be merged. Before that, can you re-run some benchmark?
> 
> It would be great to see if the update above had an impact or not. Thank you for your feedback.
> 
> Also, I have a doubt about 2 lines, it would be great to check them because it might be a problematic bug. See below.

Hey, I saw to those changes, you can view my other comments. As for benchmarking, I will get to that now and tell you the results. I'm a bit busy lately with a deadline coming soon, but I'm sure I will find the time to run this.

dipy/align/metrics.py
```diff
@@ -212,6 +213,149 @@ def get_energy(self):
...
+        else:
+            raise ValueError(f"MI Metric not defined for dim. {self.dim}")
+
+    def initialize_iteration(self):
+        r"""Prepares the metric to compute one displacement field iteration.
+
+        Computes image gradients in physical coordinates and prepares the
+        histogram quantities needed to evaluate mutual information updates.
+        """
+        if self.static_image.dtype != self.moving_image.dtype:
```
pjsjongsung: I assume this followed other metrics, but it feels strange we are comparing the dtype via the metric. Shouldn't this be a common case for all metrics, hence the error has to be dealt with when initializing SyN instead?

TomasGuija:
I’ve fixed the issues we discussed and retrieved the cluster experiment results. The benchmark completed successfully for 31 image pairs. The remaining cases did not complete for several reasons, primarily minor data-loading issues in the benchmarking code and intermittent connectivity problems with the NAS, as discussed during the meeting.

The dataset used for this experiment was introduced in my [Fifth Coding Week Blog Post](https://dipy.org/posts/2026/2026_06_29_Tomas.html). The experiment was ran under the [mi-benchmarking branch](https://github.com/TomasGuija/dipy/tree/mi-benchmarking) using the configuration built for MI SyN registration.

The results below are reported as mean ± standard deviation across the 31 successfully processed pairs.

| Metric | Baseline | ANTs | DIPY |
|---|---:|---:|---:|
| NCC ↑ | 0.3892 ± 0.0640 | 0.5938 ± 0.0422 | **0.6040 ± 0.0434** |
| NMI ↑ | 1.0218 ± 0.0065 | 1.0478 ± 0.0089 | **1.0514 ± 0.0112** |
| Mean-label Dice ↑ | 0.5136 ± 0.0536 | 0.5740 ± 0.0406 | **0.5851 ± 0.0353** |
| Mean-label Jaccard ↑ | 0.3763 ± 0.0497 | 0.4386 ± 0.0383 | **0.4506 ± 0.0318** |
| SyN execution time ↓ | — | **45.21 ± 7.24 s** | 109.63 ± 24.19 s |

Overall, DIPY achieved slightly higher mean registration-quality scores than ANTs across all the evaluated metrics. However, ANTs was approximately 2.4 times faster.

The benchmarking branch included the latest registration optimization PRs (#4073 and #4083). However, the cluster job requested only one CPU per task, and I did not explicitly configure `num_threads` or log the effective OpenMP thread count. I believe the timing difference was most probably related to the CPU-threading configuration, although I cannot confirm. There may also be a remaining bottleneck in the MI-specific computations in parzenhist.pyx, which are not currently parallelized. Further performance investigation could therefore be left for a future PR.

So far, the results suggest that MI registration is working well in terms of registration quality. Further performance optimization could therefore be considered separately in a future PR. Also, I will try to run this again on the cluster trying to make sure the latest optimizations are working, as soon as I have the cluster working properly again.

skoudoro (approved):
Thank you for this work @TomasGuija.

All good.

Merging

PR #4069, opened 2026-06-30
ENH: Faster FORCE fitting (Ray load-balancing + threaded post-processing) and example polish

Atharva-Shah-2298:
## Description

Three small, independent improvements to FORCE

1. ** auto_ray_chunk_size ** now emits several chunks per worker instead of one
   (`n_vox // (n_jobs * 8)`), so the Ray scheduler can load-balance.

2. **Parallelized the per-microstructure post-processing.** The FWHM-KDE ambiguity
   dominates serial fit time and is single-threaded numpy; since numpy releases the
   GIL on its large reductions, the per-parameter loop is fanned over a thread pool.
   It stays single-threaded inside Ray workers (where process-level parallelism
   already uses the cores). Output is unchanged. On a full HCP brain this gave ~1.6x
   faster serial fitting with bit-identical maps.

3. **Docs example:** Added a note that the Ray engine is faster and requires Ray to be
   installed; use family-consistent colormaps (DTI: viridis, PVE: gray, NODDI:
   inferno) and add per-microstructure uncertainty/ambiguity maps so that the user knows what all we get as an output.

## Motivation and Context

FORCE fitting was dominated by two avoidable serial costs:

- With the Ray engine, the chunk heuristic produced exactly one chunk per worker, so
  the run finished only when the single slowest worker did — no slack for the
  scheduler to rebalance. Producing several chunks per worker lets faster workers
  absorb the slack and also lowers peak memory per in-flight chunk.
- The per-microstructure uncertainty/ambiguity (FWHM-KDE) is ~half of the serial
  fit time and ran single-threaded while the rest of the cores sat idle.

## How Has This Been Tested?

- `pytest dipy/reconst/tests/test_force.py` — 10 passed.
- `pytest dipy/utils/tests/test_parallel.py` — 9 passed (existing `auto_ray_chunk_size`
  assertions still hold).
- **Threaded post-processing is bit-identical**: fit with `postproc` single-threaded
  vs threaded gives max |Δ| = 0.0 across all uncertainty/ambiguity maps; confirmed it
  falls back to 1 thread inside a Ray worker.

## Checklist

- [ X] I have read the [CONTRIBUTING](https://github.com/dipy/dipy/blob/master/.github/CONTRIBUTING.md) guidelines.
- [ X] My code follows the [DIPY coding style](https://docs.dipy.org/stable/devel/coding_style_guideline.html).
- [ X] I have added tests that cover my changes (if applicable).
- [X ] All new and existing tests pass locally.
- [X ] I have updated the documentation accordingly (if applicable).

## Type of Change

- [ X] Bug fix (non-breaking change which fixes an issue)
- [ X] Documentation update

### DIFF
--- a/dipy/reconst/force.py
+++ b/dipy/reconst/force.py
@@ -1,3 +1,4 @@
+from concurrent.futures import ThreadPoolExecutor
 import json
 import os
 from pathlib import Path
@@ -19,6 +20,10 @@ from dipy.sims.force import (
     save_force_simulations,
 )
 from dipy.utils.logging import logger
+from dipy.utils.multiproc import get_usable_cpu_affinity
+from dipy.utils.optpkg import optional_package
+
+ray, has_ray, _ = optional_package("ray")
 
 # Named constants
 EPSILON = 1e-12
@@ -1087,16 +1092,33 @@ class FORCEModel(ReconstModel):
             params_arrays["ambiguity"] = A
             params_arrays["entropy"] = np.full(n_vox, np.nan, dtype=np.float32)
 
-        # Per-microstructure uncertainty and ambiguity from posterior density
+        # Per-microstructure uncertainty and ambiguity from posterior density.
+        # The FWHM-KDE ambiguity is the dominant serial cost; numpy releases the
+        # GIL on its large reductions, so fan the per-parameter loop over threads.
         prior_ranges = getattr(self, "_prior_ranges", {})
-        for param in MICRO_PARAMS:
-            if param in d and param in prior_ranges:
-                vals = d[param][neighbors].astype(np.float32)
-                u, a = compute_microstructure_uncertainty_ambiguity(
-                    vals, W, prior_ranges[param]
-                )
-                params_arrays[f"uncertainty_{param}"] = u
-                params_arrays[f"ambiguity_{param}"] = a
+        todo = [p for p in MICRO_PARAMS if p in d and p in prior_ranges]
+
+        def _micro(param):
+            vals = d[param][neighbors].astype(np.float32)
+            return param, compute_microstructure_uncertainty_ambiguity(
+                vals, W, prior_ranges[param]
+            )
+
+        # One thread inside a Ray worker: process-level parallelism already
+        # uses the cores, so threading here would oversubscribe.
+        if has_ray and ray.is_initialized():
+            n_threads = 1
+        else:
+            n_threads = min(len(MICRO_PARAMS), len(get_usable_cpu_affinity()))
+        if n_threads > 1 and len(todo) > 1:
+            with ThreadPoolExecutor(max_workers=min(n_threads, len(todo))) as ex:
+                micro_results = list(ex.map(_micro, todo))
+        else:
+            micro_results = [_micro(p) for p in todo]
+
+        for param, (u, a) in micro_results:
+            params_arrays[f"uncertainty_{param}"] = u
+            params_arrays[f"ambiguity_{param}"] = a
 
         if kwargs.pop("_raw", False):
             return params_arrays
--- a/dipy/utils/parallel.py
+++ b/dipy/utils/parallel.py
@@ -96,10 +96,10 @@ def auto_ray_chunk_size(
 
       where ``bytes_per_voxel = n_gradients × 4 + output_bytes_per_voxel``.
 
-    * **Parallelism upper bound** — at least ``n_jobs`` chunks are produced
-      so all workers stay busy::
+    * **Parallelism upper bound** — several chunks per worker are produced so
+      the scheduler can load-balance across workers::
 
-          N_par = n_vox // n_jobs   (only applied when n_vox is given)
+          N_par = n_vox // (n_jobs * 8)   (only applied when n_vox is given)
 
     Final result: ``clamp(min(N_mem, N_par), lo=N_min)``.  Available RAM
     is detected via :func:`_available_ram` (stdlib only, no extra
@@ -113,7 +113,7 @@ def auto_ray_chunk_size(
         Number of diffusion gradients — sets the input chunk byte size.
     n_vox : int, optional
         Total number of voxels to fit.  Used to cap chunk size so that
-        at least ``n_jobs`` chunks are produced.
+        several chunks per worker are produced.
     shared_obj_nbytes : int, optional
         Total bytes already placed in the Ray object store (e.g. the
         simulation library).  Deducted from the available RAM budget.
@@ -139,7 +139,11 @@ def auto_ray_chunk_size(
     bytes_per_voxel = n_gradients * 4 + output_bytes_per_voxel
     n_mem = int(ray_budget / n_jobs / bytes_per_voxel)
 
-    n_par = (n_vox // n_jobs) if n_vox is not None else n_mem
+    # Emit several chunks per worker rather than one, so the Ray scheduler can
+    # load-balance: fast workers pull extra chunks while slow ones take fewer,
+    # instead of every worker being blocked on a single large chunk.
+    chunks_per_worker = 8
+    n_par = (n_vox // (n_jobs * chunks_per_worker)) if n_vox is not None else n_mem
 
     return max(n_min, min(n_mem, n_par))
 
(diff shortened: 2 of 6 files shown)


skoudoro:
Hi @Atharva-Shah-2298,

Did you check without `ray` in you PR to see if auto option is still relevant ?

Atharva-Shah-2298:
> Hi @Atharva-Shah-2298,
> 
> Did you check without `ray` in you PR to see if auto option is still relevant ?

Auto is just exposed to ray so yes if we select the engine it is useful but I am not sure if we want to hardcode it. I checked that lower batch sizes help a lot but then too much lowering regresses the speedup so it depends on the number of cores for optimal sweetspot and that is why I came up with this formula as I am assuming there would be atleast 8 cores.

dipy/reconst/force.py
```diff
@@ -473,6 +473,27 @@ def compute_uncertainty_ambiguity(scores):
 )
 
 
+def _postproc_n_threads():
```
skoudoro: useless function. to remove

use `ray, has_ray, _ = optional_package("ray")`

dipy/reconst/force.py
```diff
@@ -1087,16 +1108,30 @@ def fit(self, data, *, mask=None, **kwargs):
...
+
+        def _micro(param):
+            vals = d[param][neighbors].astype(np.float32)
+            return param, compute_microstructure_uncertainty_ambiguity(
+                vals, W, prior_ranges[param]
+            )
+
+        n_threads = _postproc_n_threads()
+        if n_threads > 1 and len(todo) > 1:
+            from concurrent.futures import ThreadPoolExecutor
```
skoudoro: import on the top

doc/examples/reconst_force.py
```diff
@@ -152,27 +157,35 @@
...
-    (axes[0, 0], fa_map, "FA", "gray", 0, 1),
-    (axes[0, 1], md_map, "MD", "hot", None, None),
-    (axes[0, 2], wm_fraction, "WM Fraction", "gray", 0, 1),
-    (axes[1, 0], num_fibers, "Num Fibers", "viridis", 0, 3),
-    (axes[1, 1], gm_fraction, "GM Fraction", "gray", 0, 1),
-    (axes[1, 2], csf_fraction, "CSF Fraction", "Blues", 0, 1),
-]
 
-for ax, arr, title, cmap, vmin, vmax in panels:
+def _show(ax, arr, title, cmap, vmin=None, vmax=None):
```
skoudoro: check in `dipy.viz.plotting` if we have a show or equivalent function. if not, create one and use that.

skoudoro (changes requested):
Hi @Atharva-Shah-2298,

Thanks for this. Here is the initial review.

I need to check locally also.

doc/examples/reconst_force.py
Atharva-Shah-2298: Yeah worked around with image mosaic but wont be as pretty as before. Okay for now

dipy/reconst/force.py
Atharva-Shah-2298: Done.

dipy/reconst/force.py
```diff
@@ -1087,16 +1091,33 @@ def fit(self, data, *, mask=None, **kwargs):
...
+            return param, compute_microstructure_uncertainty_ambiguity(
+                vals, W, prior_ranges[param]
+            )
+
+        # One thread inside a Ray worker: process-level parallelism already
+        # uses the cores, so threading here would oversubscribe.
+        if has_ray and ray.is_initialized():
+            n_threads = 1
+        else:
+            n_threads = min(len(MICRO_PARAMS), os.cpu_count() or 1)
```
skoudoro: this line can create oversubscription on HPC nodes, meaning too many threads so process is slowdown.

Instead of `os.cpu_count()` use `dipy.utils.multiprocs.get_usable_cpu_affinity()`.

however, this function do not exist. Please add this function:

```python
import os
import platform
import subprocess
import sys

def get_usable_cpu_affinity():
    """
    Returns a set of CPU core IDs available to the current process.
    Falls back to a full list of system cores if strict affinity tracking 
    is unsupported on the host platform.
    """
    # 1. Native Unix/Linux Affinity (Most accurate for containers)
    if hasattr(os, 'sched_getaffinity'):
        try:
            return os.sched_getaffinity(0)
        except Exception:
            pass

    # 2. Python 3.13+ Multiplatform Standard API
    if hasattr(os, 'process_cpu_count'):
        proc_count = os.process_cpu_count()
        if proc_count:
            return set(range(proc_count))

    # 3. macOS Fallback (Querying system restriction state via taskpolicy)
    if sys.platform == 'darwin':
        try:
            # Check physical cores vs logical cores if needed, 
            # otherwise fall back to the logical count
            res = subprocess.check_output(['sysctl', '-n', 'hw.ncpu']).strip()
            return set(range(int(res)))
        except Exception:
            pass

    # 4. Windows/Generic Fallback (Using multiprocessing or os)
    # On Windows, true affinity masks can be set via Win32 API, 
    # but native python defaults to the total allocated system count.
    try:
        import multiprocessing
        return set(range(multiprocessing.cpu_count()))
    except Exception:
        return set(range(os.cpu_count() or 1))

# Example usage:
def test_to_define():
    cores = get_usable_cpu_affinity()
    print(f"Platform: {platform.system()}")
    print(f"Usable CPU Core IDs: {cores}")
    print(f"Total Usable Count: {len(cores)}")
```

doc/examples/reconst_force.py
```diff
@@ -148,42 +168,116 @@
...
-    plt.colorbar(im, ax=ax, fraction=0.046)
 
-plt.tight_layout()
-plt.savefig("force_maps.png", dpi=150, bbox_inches="tight")
-# plt.show()
+def _mosaic(images, labels, kwargs, filename):
+    fig, ax = image_mosaic(images, ax_labels=labels, ax_kwargs=kwargs, figsize=(20, 5))
+    for axe in ax:
+        axe.axis("off")
+    fig.savefig(filename, dpi=150, bbox_inches="tight")
```
skoudoro: lines to remove.  use directly image mosaic (eg. filename=filename)

if you do not use filename, doc generation block, because of the `plt.show()`. Check the code:
https://github.com/dipy/dipy/blob/1f982002d959d91b0c08487eca5a79be66906aca/dipy/viz/plotting.py#L329-L332


Feel free to improve the code

skoudoro (changes requested):
Thanks for the update @Atharva-Shah-2298,

See below, the 2 last comments to address.

Thanks!

dipy/utils/parallel.py
```diff
@@ -139,7 +139,11 @@ def auto_ray_chunk_size(
     bytes_per_voxel = n_gradients * 4 + output_bytes_per_voxel
     n_mem = int(ray_budget / n_jobs / bytes_per_voxel)
 
-    n_par = (n_vox // n_jobs) if n_vox is not None else n_mem
+    # Emit several chunks per worker rather than one, so the Ray scheduler can
+    # load-balance: fast workers pull extra chunks while slow ones take fewer,
+    # instead of every worker being blocked on a single large chunk.
+    chunks_per_worker = 8
```
skoudoro, line 145: can we tune this ? Do you recommend it or we should avoid ?

skoudoro (commented):
Thanks for the quick update @Atharva-Shah-2298.

See last question below.

dipy/utils/parallel.py
Atharva-Shah-2298: I tried with many values on 3 machines, 1 with 24 cores 128GB ram, 1 with 8 cores 16 threads with 32 GB ram and 1 Mac M1 with 8 cores 8 threads with 8 gb of ram. This is the sweetspot I found. If I increase the division, the Overhead of writing and reading takes over and then if I reduce it too much, it chokes up less memory machines. I still think there is room for some intelligence here but I am not sure whats a better way than to hard code as of now.

skoudoro (approved):
ok, all green and I do not any immediate issue. it works for me so we can go ahead and merge

Thanks for this work @Atharva-Shah-2298.

merging

ps: CI's all green here but not the other PR...

PR #4071, opened 2026-07-02
OPT: avoid redundant Parzen spline-bin evaluation

TomasGuija:
## Description

This PR updates the Parzen joint histogram implementation used for mutual information by removing a redundant cubic B-spline bin evaluation.

For each sample, the current implementation evaluates moving-bin offsets `[-2, -1, 0, 1, 2]`. However, the weight assigned to each neighboring bin is computed by `_cubic_spline(q - cn)`, where `q` is the neighboring bin index and `cn` is the sample's continuous bin position. The `_cubic_spline()` function returns zero when `|q - cn| >= 2`. Since `c = floor(cn)` for in-range samples, the left-most evaluated bin `q = c - 2` is always at distance greater than or equal to `2` from `cn`, so it always receives zero weight. The implementation now iterates only over the effective nonzero support:

`range(1 - padding, padding + 1)`

which corresponds to offsets `[-1, 0, 1, 2]` when `padding = 2`.

This also removes hardcoded `2` values from the Parzen histogram loops and reuses the existing `padding` variable consistently.

## Motivation and Context

The previous code used `padding = 2`, but the histogram update loops still hardcoded the cubic B-spline radius manually through expressions such as `spline_arg = (c - 2) - cn` and `for offset in range(-2, 3)`.

This made the code less clear and caused one unnecessary spline evaluation per sample in both PDF and gradient computations.

This PR keeps the mathematical behavior unchanged for the effective support while avoiding redundant computation and making the relation between `padding` and the offset loop clearer.

## How Has This Been Tested?

I tested this by instrumenting the original implementation with debug counters in the dense 3D PDF and gradient paths. In a real MI run, the left-most offset was checked across 80,560 samples per evaluation, and its contribution was always exactly zero for both the PDF and gradient computations.

## Checklist

## Type of Change

- [ ] Documentation update

### DIFF
--- a/dipy/align/parzenhist.pyx
+++ b/dipy/align/parzenhist.pyx
@@ -66,9 +66,22 @@ class ParzenJointHistogram:
         # than one bin, we need to add extra bins to both sides of the
         # histogram to account for the contributions of the minimum and maximum
         # intensities. Padding is the number of extra bins used at each side
-        # of the histogram (a total of [2 * padding] extra bins). Since the
-        # support of the cubic spline is 5 bins (the center plus 2 bins at each
-        # side) we need a padding of 2, in the case of cubic splines.
+        # of the histogram (a total of [2 * padding] extra bins).
+        #
+        # The cubic B-spline kernel has radius 2: a sample can only affect
+        # histogram bins whose centers are less than 2 bins away from the
+        # sample's continuous bin position.
+        #
+        # Here, c is the integer bin immediately to the left of the sample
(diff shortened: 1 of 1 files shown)


Atharva-Shah-2298:
The commit makes sense to me. I have not gone into depth of the usecase and the intended downstream use. 

On other note 

________________________________ test_csd_xval _________________________________


This CI has started failing on every PR made in last few days. Need to check what changed here!

skoudoro (approved):
Thanks for this @TomasGuija,

After multiple tests, looks good to me. Merging

PR #4073, opened 2026-07-03
 OPT: Parallelize displacement-field composition

TomasGuija:
## Description

This PR parallelizes displacement-field composition in `vector_fields.pyx` using OpenMP. The implementation keeps the existing public behavior unchanged while adding `num_threads` support to the relevant composition and inversion paths used by SyN.

The main change is that the outer composition loops are executed with `prange`, with per-thread statistics accumulated independently and merged after the parallel section.

## Motivation and Context

Displacement-field composition is called repeatedly during SyN registration and fixed-point inversion. For 3D registration, this can become a meaningful runtime bottleneck.

This PR improves performance by parallelizing composition while preserving the existing output field and statistics.

End-to-end SyN timing comparison on single-sample experiments:

| Case | Pre-fix time (s) | Post-fix time (s) | Time reduction (s) | Speed-up | Time reduction (%) |
|---|---:|---:|---:|---:|---:|
| 2D | 7.206175 | 6.720559 | 0.485616 | 1.07x | 6.74% |
| 3D | 11.163576 | 8.494781 | 2.668795 | 1.31x | 23.91% |

## How Has This Been Tested?

Tested locally with added coverage for threaded vector-field composition:

- `compose_vector_fields_2d(..., num_threads=1)` vs `num_threads=2`
- `compose_vector_fields_3d(..., num_threads=1)` vs `num_threads=2`
- output displacement fields are identical
- merged statistics match the one-thread reference

Also tested SyN registration timing before and after the change using the DIPY registration examples.

## Notes

While reviewing the composition statistics, I noticed a possible mismatch between the documentation and the existing implementation: the docs describe the returned statistics as maximum, mean, and standard deviation of vector norms, while the implementation appears to compute statistics from squared norms. This PR intentionally preserves the existing behavior.

Composition statistics are currently always computed, even in paths where they are not used. A follow-up PR could add an optional `compute_stats` parameter to skip unnecessary statistic computation.

## Checklist

## Type of Change

- [ ] Documentation update

### DIFF
--- a/dipy/align/imwarp.py
+++ b/dipy/align/imwarp.py
@@ -1111,6 +1111,7 @@ class SymmetricDiffeomorphicRegistration(DiffeomorphicRegistration):
         inv_iter=20,
         inv_tol=1e-3,
         callback=None,
+        num_threads=None,
     ):
         """Symmetric Diffeomorphic Registration (SyN) Algorithm
 
@@ -1145,6 +1146,12 @@ class SymmetricDiffeomorphicRegistration(DiffeomorphicRegistration):
             a function receiving a SymmetricDiffeomorphicRegistration object
             to be called after each iteration (this optimizer will call this
             function passing self as parameter)
+        num_threads : int, optional
+            Number of threads to use for displacement field inversion. If None,
+            the value of the ``OMP_NUM_THREADS`` environment variable is used
+            if set; otherwise, all available threads are used. Negative values
+            count backward from the maximum number of threads. Zero raises an
+            error.
         """
         super().__init__(metric=metric)
         if level_iters is None:
@@ -1164,6 +1171,7 @@ class SymmetricDiffeomorphicRegistration(DiffeomorphicRegistration):
         self.full_energy_profile = []
         self.verbosity = VerbosityLevels.STATUS
         self.callback = callback
+        self.num_threads = num_threads
         self.moving_ss = None
         self.static_ss = None
         self.static_direction = None
@@ -1608,6 +1616,7 @@ class SymmetricDiffeomorphicRegistration(DiffeomorphicRegistration):
                 self.inv_iter,
                 self.inv_tol,
                 start=self.static_to_ref.backward,
+                num_threads=self.num_threads,
             )
         )
 
@@ -1620,6 +1629,7 @@ class SymmetricDiffeomorphicRegistration(DiffeomorphicRegistration):
                 self.inv_iter,
                 self.inv_tol,
                 start=self.moving_to_ref.backward,
+                num_threads=self.num_threads,
             )
         )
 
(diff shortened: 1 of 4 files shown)


Atharva-Shah-2298:
@TomasGuija. Weirdly enough single thread version I tested measured at 0.66 to 0.77x. The 2D multi threaded and 3D multithreaded worked at ~1.5X on 2 threads. All the code LGTM. Can you just address the test case coverage CI?

Thanks

Atharva-Shah-2298 (approved):

dipy/align/imwarp.py
```diff
@@ -846,13 +846,19 @@ def expand_fields(self, expand_factors, new_shape):
         self.disp_grid2world = expanded_grid2world
         self.disp_world2grid = expanded_world2grid
 
-    def compute_inversion_error(self):
+    def compute_inversion_error(self, num_threads=None):
```
skoudoro: `*` to add

dipy/align/imwarp.py
```diff
@@ -911,7 +925,7 @@ def shallow_copy(self):
         new_map.is_inverse = self.is_inverse
         return new_map
 
-    def warp_endomorphism(self, phi):
+    def warp_endomorphism(self, phi, num_threads=None):
```
skoudoro: same

dipy/align/imwarp.py
```diff
@@ -1153,6 +1189,12 @@ def __init__(
         if len(level_iters) == 0:
             raise ValueError("The iterations list cannot be empty")
 
+        if num_threads is not None:
+            if not isinstance(num_threads, int):
+                raise TypeError("num_threads must be an int or None")
+            if num_threads < 1:
+                raise ValueError("num_threads must be a positive integer or None")
```
skoudoro: use function named `determine_nb_thread`

skoudoro (changes requested):
Hi @TomasGuija, 

Small comment but this is ready to go

dipy/align/imwarp.py
TomasGuija: Regarding this comment, I removed those errors, but `determine_nb_thread` is still called later in `vector_fields.pyx`. I structured it this way mainly so that calls to the different composition methods made outside `imwarp` still determine the correct number of threads. I avoided calling it again in `imwarp` to prevent duplicating the call. That said, I can add it there as well if you would prefer.

dipy/align/imwarp.py
TomasGuija: True, I will try to embed those changes into an earliler commit so as to avoid having a commit for only that.

dipy/align/imwarp.py
skoudoro: no, all good 👍

skoudoro:
Thanks for the update,  Waiting for the CI and we can go ahead and merge

PR #4077, opened 2026-07-06
RF: Centralize shared Cython DEFs

jhlegarreta:
## Description

Move shared compile-time constants into a common Cython module and import them where needed to remove duplicate definitions and keep values consistent.

## Motivation and Context

We should not be duplicating constant definitions.

## How Has This Been Tested?

GitHub actions will test changes.

## Checklist

- [X] I have updated the documentation accordingly (if applicable).

## Type of Change

- [ ] Documentation update

Follow-up to PR #4070.

### DIFF
--- a/dipy/segment/clustering_algorithms.pyx
+++ b/dipy/segment/clustering_algorithms.pyx
@@ -2,6 +2,7 @@
 
 import itertools
 import numpy as np
+from libc.limits cimport INT_MAX
 
 from dipy.segment.cythonutils cimport Data2D, shape2tuple
 from dipy.segment.metricspeed cimport Metric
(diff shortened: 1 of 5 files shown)


jhlegarreta:
@skoudoro Have spent some time trying to make this work; not sure where the issues is. Help is appreciated.

skoudoro:
Hi @jhlegarreta,

Multiple meetings, so I will take a look later today.

but it looks like you did not update any `meson.build` files despite adding a file.  `pxd` are like `.h` in c so compiler need to find where are your headers. 

I will check later

jhlegarreta:
@skoudoro Thanks for fixing the compile errors. Two comments:
- `DTYPE` is still not centralized after your push-force: can it not be centralized?
- The commit message 
```
Drop the added `--depfile` for `constants.pxd` in
`dipy/segment/meson.build`: `extra_args` is never passed to
`cython_gen.process()` in this file, so it never reached the
compiler; wiring it up is unsafe since Cython's `-M`/`--depfile` is a
boolean flag, not one taking a path, and passing a path makes Cython
treat it as an extra source file to compile.
```

corresponds to no changes in the mentioned lines.

- Slightly related (or unrelated):
https://github.com/dipy/dipy/blob/master/dipy/utils/meson.build#L13

are never passed to the relevant commands, and that seems to happen consistently across other `meson.build` files. That change is unrelated to this PR, and would probably welcome a separate PR.

skoudoro:
Hi @jhlegarreta,

- For `DTYPE`, I think this variable is useless, and we can remove it, so inline the code (name too generic)
- sorry for commit message, updated
- agree with `extra_args`, need a specific PR

jhlegarreta:
Sounds good, Serge. Thanks for amending the commit to finish the PR.

skoudoro:
ok, merging

Issue #2376, opened 2021-04-27
Inconsistent registration results across macos (darwin) and linux


kmannislands:
## Description

Hello dipy team and thank you for all of your excellent work!

We're running into a strange issue around image registration of a DWI image to a base anatomical image (say t1). 

What we're encountering---after much debugging--is a difference in registration results when running on our mac dev machines compared to running on docker-ized linux (non-docker linux as well, actually).

I've prepared an [isolated reproduction repo](https://github.com/kmannislands/dipy-registration-repro) that is centered around registering the scans (using `RigidTransform3D` or `TranslationTransform3D`) on mac and pickling the results then running the same test on mac again (passes) and in a "vanilla" linux docker container for comparison (fails). I can share the pickle files with the imaging data if requested. Please see the `README.md` in the linked repo for reproduction steps.

## Way to reproduce
[If reporting a bug, please include the following important information:]
- [ ] Code example: sample repo found here: https://github.com/kmannislands/dipy-registration-repro Please contact me for pickled testing data if needed.
- [ ] Relevant images (if any)
- [ ] Operating system and version:
  macOs: `Darwin-19.6.0-x86_64-i386-64bit`
  linux:  `Linux-4.19.121-linuxkit-x86_64-with-debian-10.3`
- [ ] Python version 3.7.6
- [ ] dipy version: tested dipy `0.16.0` and `1.4.0` and observed the same result.
- [ ] dependency version (numpy, scipy, nibabel, h5py, cvxpy, fury) (these are for dipy 0.16.0)
    * numpy 1.18.4
    * scipy 1.5.2
    * nibabel 3.1.0
    * h5py 3.1.0
    * don't seem to have cvxpy present against dipy 0.16.0 anyways
    * same for fury

skoudoro:
Hi @kmannislands,

Thank you for your feedback, we will look into it.

> ... is a difference in registration results ...

Is it a huge difference? Can you give us approximately an order? It is normal for image registration to not get the exact same result, you often have a slight change due to the optimization. 

[this tutorial](https://dipy.org/documentation/1.4.0./examples_built/register_binary_fuzzy/#example-register-binary-fuzzy) shows you how it can be sensitive.

kmannislands:
@skoudoro hello and thanks for a speedy reply.

It's interesting because it does seem to depend on the input images. We get identical (to ~7 decimals) affine matrices on mac/linux for other registrations (dwi onto other dwi) we perform with the same approach.

[...]

PR #4078, opened 2026-07-07
NF: CC implementation for affine registration

TomasGuija:
## Description

This PR adds a new local cross-correlation metric for affine registration.

The new `CrossCorrelationMetric` follows the same optimizer-facing interface as the existing `MutualInformationMetric`, providing `setup`, `distance`, `gradient`, and `distance_and_gradient` methods for use with `AffineRegistration`.

The implementation reuses the existing cross-correlation machinery in `dipy.align.crosscorr`, including the same local CC factor computation and negative-energy convention used by the SyN registration path. The affine-specific part projects the local CC image-space gradient onto affine transform parameters through the transform Jacobian.

## Motivation and Context

DIPY currently supports mutual information for affine registration, but local cross-correlation is not available as an affine registration metric.

This change adds cross-correlation support for affine registration, which is useful for monomodal registration problems where local normalized cross-correlation is often an appropriate similarity measure.

The goal of the implementation is to keep the new metric consistent with the existing DIPY registration design, while reusing as much of the existing cross-correlation backend as possible.

## How Has This Been Tested?

I tested the new affine cross-correlation metric using the sample datasets downloaded through DIPY, following the existing affine 3D registration example structure.

The test registered the Stanford HARDI b0 volume against the SyN b0 sample volume using a center-of-mass initialization followed by translation, rigid, and affine registration stages with the new `CrossCorrelationMetric`. This provides a monomodal test case, since both images are b0 MRI volumes.

The pre-registration and post-registration metrics improved consistently across the registration pipeline:

| Stage | Global NCC ↑ |
|---|---:|
| Identity / pre-registration | 0.190317 |
| Center of mass | 0.477166 |
| Translation CC | 0.558438 |
| Rigid CC | 0.665606 |
| Affine CC | 0.675099 |

## Checklist

## Type of Change

- [ ] Documentation update

### DIFF
--- a/dipy/align/crosscorr.pyx
+++ b/dipy/align/crosscorr.pyx
@@ -355,6 +355,292 @@ def precompute_cc_factors_3d_test(floating[:, :, :] static,
 @cython.boundscheck(False)
 @cython.wraparound(False)
 @cython.cdivision(True)
+cdef double _compute_cc_2d(
+        floating[:, :, :] gradient,
+        floating[:, :, :] factors,
+        cnp.npy_intp radius,
+        bint forward_step,
+        floating[:, :, :] displacement=None,
+        double[:] metric_gradient=None,
+        double[:] theta=None,
+        Transform transform=None,
+        double[:, :] grid2world=None):
+    """Compute 2D CC energy and write the requested gradient output.
+
+    This helper contains the common computation used by SyN and affine
+    registration. If ``displacement`` is provided, it writes the dense SyN
+    update. If ``metric_gradient`` is provided, it accumulates the affine
+    parameter gradient instead. If neither is provided, it computes only the
+    CC energy. The two output buffers are not intended to be used together.
+
+    Parameters
+    ----------
+    gradient : array, shape (R, C, 2), optional
+        the gradient of the static image for a forward step, or the gradient
+        of the moving image for a backward step. May be None when only the CC
+        energy is requested.
+    factors : array, shape (R, C, 5)
+        the precomputed cross correlation terms obtained via
+        precompute_cc_factors_2d.
+    radius : int
+        the radius of the neighborhood used for the CC metric when
+        computing the factors. The energy and gradient are computed
+        only over the valid image region, excluding a boundary of
+        width radius pixels.
+    forward_step : bool
+        if True, compute the forward step. Otherwise, compute the backward
+        step. This choice is only required by SyN.
+    displacement : array, shape (R, C, 2), optional
+        buffer in which to write the dense displacement gradient for SyN. If
+        None, the dense displacement gradient is not computed.
+    metric_gradient : array, shape (n,), optional
+        array to write the affine registration gradient of the cross
+        correlation energy with respect to ``theta``. If None, the affine
+        parameter gradient is not computed.
+    theta : array, shape (n,), optional
+        current parameter vector of the affine transform. Used only for affine
+        registration and required when ``metric_gradient`` is not None.
+    transform : instance of Transform, optional
+        transform with respect to whose parameters the gradient must be
+        computed. Used only for affine registration and required when
+        ``metric_gradient`` is not None.
+    grid2world : array, shape (3, 3), optional
+        transform from static grid coordinates to the physical coordinates
+        where the affine transform Jacobian must be evaluated. Used only for
+        affine registration and required when ``metric_gradient`` is not None.
+
+    Returns
+    -------
+    energy : float
+        the cross correlation energy (data term).
+
+    """
+    cdef:
+        cnp.npy_intp nr = factors.shape[0]
+        cnp.npy_intp nc = factors.shape[1]
+        cnp.npy_intp n = 0
+        cnp.npy_intp r, c, k
+        int constant_jacobian = 0
+        double energy = 0
+        double Ii, Ji, sfm, sff, smm, localCorrelation, temp
+        double spatial_derivative
+        double[:, :] J
+        double[:] x
+
+    if metric_gradient is not None:
+        n = metric_gradient.shape[0]
+        J = np.empty((2, n), dtype=np.float64)
+        x = np.empty((2,), dtype=np.float64)
+
+    with nogil:
+        if metric_gradient is not None:
+            metric_gradient[:] = 0
+
+        for r in range(radius, nr-radius):
+            for c in range(radius, nc-radius):
+                Ii = factors[r, c, 0]
+                Ji = factors[r, c, 1]
+                sfm = factors[r, c, 2]
+                sff = factors[r, c, 3]
+                smm = factors[r, c, 4]
+                if sff == 0.0 or smm == 0.0:
+                    continue
+                localCorrelation = 0
+                if sff * smm > 1e-5:
+                    localCorrelation = sfm * sfm / (sff * smm)
+                if localCorrelation < 1:  # avoid bad values...
+                    energy -= localCorrelation
+
+                if displacement is not None or metric_gradient is not None:
+                    # The affine path always uses the backward update.
+                    if displacement is not None and forward_step:
+                        temp = 2.0 * sfm / (sff * smm) * (
+                            Ji - sfm / sff * Ii
+                        )
+                    else:
+                        temp = 2.0 * sfm / (sff * smm) * (
+                            Ii - sfm / smm * Ji
+                        )
+
+                    if displacement is not None:
+                        displacement[r, c, 0] -= temp * gradient[r, c, 0]
+                        displacement[r, c, 1] -= temp * gradient[r, c, 1]
+                    else:
+                        x[0] = _apply_affine_2d_x0(
+                            <double>r, <double>c, 1.0, grid2world
+                        )
+                        x[1] = _apply_affine_2d_x1(
+                            <double>r, <double>c, 1.0, grid2world
+                        )
+
+                        if constant_jacobian == 0:
+                            constant_jacobian = transform._jacobian(theta, x, J)
+
+                        for k in range(n):
+                            spatial_derivative = (
+                                J[0, k] * gradient[r, c, 0]
+                                + J[1, k] * gradient[r, c, 1]
+                            )
+                            metric_gradient[k] -= temp * spatial_derivative
+
+    return energy
+
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+@cython.cdivision(True)
+cdef double _compute_cc_3d(
+        floating[:, :, :, :] gradient,
+        floating[:, :, :, :] factors,
+        cnp.npy_intp radius,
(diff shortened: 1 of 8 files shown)


TomasGuija:
Hey @skoudoro,

I made the changes we discussed and added test coverage too. There are just two things still pending:

1. **Return docstrings:** We discussed making the parameter names in the `Returns` sections match the variables actually returned. However, I noticed that `MutualInformationMetric` follows the same pattern as my implementation. For example, the `distance` docstring calls the returned value `neg_mi`, while the code returns `-1 * self.metric_val`. For now, I kept the new docstrings as they are. Maybe I could open a small follow-up PR just to improve docstring consistency across these metrics?

2. **Mask support:** The new metric doesn’t support masks yet. I don’t remember whether we discussed this, or maybe I forgot to bring it up. Supporting masks involves a few design decisions, so for now passing one raises a clear `NotImplementedError`. We can talk about it in the next meeting and decide whether to implement it here or in a separate PR once this first version is merged.

That’s all for now. Looking forward to your feedback!

dipy/align/imaffine.py
```diff
@@ -892,6 +895,299 @@ def distance_and_gradient(self, params):
...
+            self.compute_affine = cc.compute_cc_affine_2d
+        elif self.dim == 3:
+            self.precompute_factors = cc.precompute_cc_factors_3d
+            self.compute_affine = cc.compute_cc_affine_3d
+        else:
+            raise ValueError(
+                f"Cross-correlation is not defined for dimension {self.dim}"
+            )
+
+    @warning_for_keywords()
```
skoudoro: to remove, Not needed for new function, only for historical one

dipy/align/crosscorr.pyx
```diff
@@ -779,3 +788,261 @@ def compute_cc_backward_step_2d(floating[:, :, :] grad_moving,
                 out[r, c, 0] -= temp * grad_moving[r, c, 0]
                 out[r, c, 1] -= temp * grad_moving[r, c, 1]
     return np.asarray(out), energy
+
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+@cython.cdivision(True)
+def compute_cc_affine_2d(
```
skoudoro, line 980: I believe this function is a duplicate of `compute_cc_backward_step_2d`. 

I think there a smarter way to reuse the existing function. Can you update?

dipy/align/crosscorr.pyx
```diff
@@ -779,3 +788,261 @@ def compute_cc_backward_step_2d(floating[:, :, :] grad_moving,
...
+                        )
+                        metric_gradient[k] -= temp * spatial_derivative
+
+    return energy
+
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+@cython.cdivision(True)
+def compute_cc_affine_3d(
```
skoudoro, line 1062: same, I believe this function is a duplicate of `compute_cc_backward_step_3d`. 

I think there a smarter way to reuse the existing function. Can you update?

dipy/align/imaffine.py
```diff
@@ -892,6 +895,299 @@ def distance_and_gradient(self, params):
...
+        static_mask : array, shape (S, R, C) or (R, C), optional
+            static image mask. Not currently supported by this metric.
+        moving_mask : array, shape (S', R', C') or (R', C'), optional
+            moving image mask. Not currently supported by this metric.
+
+        """
+        if static_mask is not None or moving_mask is not None:
+            raise NotImplementedError(
+                "Masks are not currently supported by CrossCorrelationMetric"
+            )
```
skoudoro, line 991: I confirm, implement masking version please

dipy/align/imaffine.py
```diff
@@ -892,6 +895,299 @@ def distance_and_gradient(self, params):
...
+            current affine transform.
+
+        """
+        warped_moving = self.affine_map.transform(self.moving)
+        warped_moving = np.asarray(warped_moving, dtype=np.float64)
+        factors = self.precompute_factors(self.static, warped_moving, self.radius)
+        factors = np.asarray(factors)
+        return factors
+
+    @warning_for_keywords()
```
skoudoro: not needed on new function

dipy/align/imaffine.py
```diff
@@ -892,6 +895,299 @@ def distance_and_gradient(self, params):
...
+        Parameters
+        ----------
+        params : array, shape (n,)
+            the parameter vector of the transform currently used by the metric
+            (the transform name is provided when self.setup is called), n is
+            the number of parameters of the transform
+
+        Returns
+        -------
+        cc_energy : float
```
skoudoro: You should update it. This is a mistake on `MI` metric that needs to be updated also on a new PR

dipy/align/imaffine.py
```diff
@@ -892,6 +895,299 @@ def distance_and_gradient(self, params):
...
+        Parameters
+        ----------
+        params : array, shape (n,)
+            the parameter vector of the transform currently used by the metric
+            (the transform name is provided when self.setup is called), n is
+            the number of parameters of the transform
+
+        Returns
+        -------
+        grad : array, shape (n,)
```
skoudoro: same as above

dipy/align/imaffine.py
```diff
@@ -892,6 +895,299 @@ def distance_and_gradient(self, params):
...
+        Parameters
+        ----------
+        params : array, shape (n,)
+            the parameter vector of the transform currently used by the metric
+            (the transform name is provided when self.setup is called), n is
+            the number of parameters of the transform
+
+        Returns
+        -------
+        cc_energy : float
```
skoudoro: same as above

skoudoro (changes requested):
Hi @TomasGuija,

Thank you for adding this metric !

Below some comments. Furthermore:
- Please update ImageRegistrationFlow to allow `CC`metric
- Update one tutorial or add a Tutorial to let people know that it exists

Thank you for your future update.

dipy/align/crosscorr.pyx
TomasGuija: Yes there is some duplicated code, that's true. The reason I introduced a separate affine method is that ``compute_cc_backward_step_2d/3d`` produces a dense update vector for every voxel. For affine registration, we do not need to keep that full vector field: each voxel update is immediately projected through the affine transform Jacobian and accumulated into a small parameter-gradient vector.

Calling the existing method and then projecting the derivative onto the transform parameters would work, but it would allocate and write a full displacement field and then require another pass over it for the affine projection. I think a possible way to remove duplicated code would be to extract that shared per-voxel calculation into a separate helper, and then both the SyN and affine methods can reuse it. 

We can discuss that tomorrow in the meeting.

skoudoro: yep, we can discussed it, but already +1 for this plan

dipy/align/imaffine.py
TomasGuija: Yes, I need to check those and open a new PR. I was thinking about it though, because writing the actually returned value on the docstring, such as ``self.metric_val`` doesn't seem a good idea. I could then either locally define ``cc_energy`` so things match, or following the example I found elsewhere in the codebase (in particular at [``segment/metric.py``](https://github.com/dipy/dipy/blob/master/dipy/segment/metric.py)), simply write the dtype of the returned value without explicitly naming it. In this case, it would be simply ``float`` for example.

skoudoro: yeah, just return the dtype, it simplify the maintenance and good practice

dipy/align/crosscorr.pyx
TomasGuija: Hey @skoudoro, while investigating the best way to do this, I noticed that there is already quite a lot of duplication in the current CC implementation. The SyN forward and backward functions use the same allocation, voxel loops, validity checks, energy calculation, and output operation. The only internal difference is the `temp` formula; the corresponding static or moving gradient is already provided by the caller.

Since, from what I understand, the main remaining point before this PR can be merged is simplifying the code to remove duplication, I have opened a small preliminary PR at #4088. It replaces each forward/backward pair with a shared `compute_cc_step_2d/3d` function and an explicit `forward_step` argument. `CCMetric` calls these functions directly, and the existing tests and call sites have been updated accordingly.

I would therefore propose:

1. Review and merge the new cleanup PR.
2. Rebase this branch on top of it.
3. Extend the shared structure so SyN can write its dense update while affine registration directly accumulates the parameter gradient, without allocating a dense displacement field.

I have not fully polished the third step yet, but I think it can now be implemented more cleanly on top of the shared CC functions.

What do you think?

skoudoro: fine with this strategy. Thanks !

skoudoro: #4088 has been merged @TomasGuija, you can update this PR

skoudoro:
to rebase and fix the conflict. Thanks @TomasGuija !

skoudoro:
```python

=============================== warnings summary ===============================
align/tests/test_crosscorr.py::test_compute_cc_affine_2d
  /home/runner/work/dipy/dipy/venv/lib/python3.13/site-packages/dipy/align/tests/test_crosscorr.py:51: DeprecationWarning: compute_cc_backward_step_2d is deprecated. Use compute_cc_step_2d with forward_step=False instead.
  
  * deprecated from version: 1.13
  * Will raise <class 'dipy.utils.deprecator.ExpiredDeprecationError'> as of version: 2.0
    update, expected_energy = cc.compute_cc_backward_step_2d(

align/tests/test_crosscorr.py::test_compute_cc_affine_3d
  /home/runner/work/dipy/dipy/venv/lib/python3.13/site-packages/dipy/align/tests/test_crosscorr.py:91: DeprecationWarning: compute_cc_backward_step_3d is deprecated. Use compute_cc_step_3d with forward_step=False instead.
  
  * deprecated from version: 1.13
  * Will raise <class 'dipy.utils.deprecator.ExpiredDeprecationError'> as of version: 2.0
    update, expected_energy = cc.compute_cc_backward_step_3d(

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=================================== Werrors ====================================
Warnings as errors: Activated. 
2 warnings were raised and treated as errors. 

```

TomasGuija:
Hey @skoudoro I just saw your last commit. I had also though about just relaxing the test, but didn't get enough time to study the source of these issues. Do you think this can be merged then? Should the MacOS issue be studied further too in general?

skoudoro:
> Do you think this can be merged then? Should the MacOS issue be studied further too in general?

Hi @TomasGuija,

I answer it here: https://github.com/dipy/dipy/issues/2376#issuecomment-5132964635

skoudoro (approved):
CI's error not related.  All good, Thank you for this work @TomasGuija!

merging

PR #4080, opened 2026-07-09
Flexible input for synthseg

TomasGuija:
## Description

This PR updates `SynthSeg.predict` to support flexible input sizes instead of forcing every image to a fixed `192 x 192 x 192` model space.

The preprocessing pipeline now resamples input images to 1 mm isotropic resolution, computes the required model input shape dynamically, and pads the data to the smallest shape whose dimensions are divisible by 32. The predicted labels and masks are still recovered back to the original input image space, preserving the existing output behavior for standard predictions.

This PR also adds a small utility function, `get_padded_shape`, to compute the padded model-compatible shape in a reusable and readable way.

## Motivation and Context

Currently, `SynthSeg.predict` always passes `final_size=(192, 192, 192)` during preprocessing. This makes the model input shape fixed, meaning that larger images can be cropped during preprocessing, potentially reducing the field of view available to the model. While using the current interface, I have witnessed several cases where the output predictions were cropped and did not cover the full anatomy.

This behavior is also more restrictive than what is required by the original [SynthSeg implementation](https://github.com/BBillot/SynthSeg). The official SynthSeg documentation states that, when using the optional crop argument, the crop size must be divisible by 32, but it does not require a fixed `192 x 192 x 192` input size. In fact, by default, SynthSeg processes the whole image, and cropping is only an optional mechanism for faster analysis or fitting the image into GPU memory.

This change therefore removes the hard-coded `192 x 192 x 192` assumption and instead adapts the model input size to the input image after resampling. The new behavior keeps the full resampled field of view, while still ensuring that the final model-space shape is compatible with the network architecture by padding each dimension to a multiple of 32.

## How Has This Been Tested?

I tested this by running `SynthSeg.predict` on multiple T1-weighted samples from the OASIS-2 dataset and comparing the outputs obtained from the current `master` branch against the outputs obtained from this PR branch.

I visually inspected the predicted SynthSeg labels and masks from both branches. Across many OASIS-2 samples, I observed the same behavior: the flexible-input branch preserves a larger field of view when the resampled image does not naturally fit within the fixed `192 x 192 x 192` space, while the final recovered labels and masks remain aligned with the original input image space.

A representative visual comparison is shown below:

<table>
  <tr>
    <td align="center">
      <br>
      <sub><b>Master</b>: fixed <code>192 x 192 x 192</code> preprocessing</sub>
    </td>
    <td align="center">
      <br>
      <sub><b>This PR</b>: flexible multiple-of-32 preprocessing</sub>
    </td>
  </tr>
</table>

The images were taken manually in Slicer, so small mismatches may just come from slightly different screenshot positioning.

## Checklist

## Type of Change

- [ ] Breaking change (fix or feature that would cause existing functionality not to work as expected)
- [ ] Documentation update

### DIFF
--- a/dipy/nn/torch/synthseg.py
+++ b/dipy/nn/torch/synthseg.py
@@ -638,8 +638,8 @@ class SynthSeg:
         if batch_size is None:
             batch_size = 1
 
-        transformed_images = []
-        params_list = []
+        transformed_images = [None] * len(T1)
+        params_list = [None] * len(T1)
 
         # Resample the data to 1 mm isotropic resolution.
         ori_shape = T1.shape[1:]
@@ -650,8 +650,8 @@ class SynthSeg:
                 target_voxsize=(1.0, 1.0, 1.0),
                 order=3,
             )
-            transformed_images.append(t_img)
-            params_list.append(params)
+            transformed_images[i] = t_img
+            params_list[i] = params
 
         # Pad all images to a common shape compatible with SynthSeg.
         max_shape = np.max([img.shape for img in transformed_images], axis=0)
--- a/dipy/nn/torch/synthseg.py
+++ b/dipy/nn/torch/synthseg.py
@@ -22,7 +22,9 @@ from scipy.ndimage import gaussian_filter
 
 from dipy.data import get_fnames
 from dipy.nn.utils import (
+    get_padded_shape,
     normalize,
+    pad_crop,
     recover_img,
     set_logger_level,
     transform_img,
@@ -596,7 +598,8 @@ class SynthSeg:
             has a batch dimension.
         return_prob : bool, optional
             Whether to return the probability map instead of a
-            label map. Useful for testing.
+            label map. Probability maps remain in the 1 mm isotropic model
+            space, padded to dimensions divisible by 32. Useful for testing.
 
         Returns
         -------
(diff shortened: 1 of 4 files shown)


dipy/nn/utils.py
```diff
@@ -277,6 +279,28 @@ def recover_img(image, params, *, order=3):
     return recovered
 
 
+def get_padded_shape(shape, multiple):
+    """Round each dimension of a shape up to a given multiple.
```
pjsjongsung: We are ceiling the dimensions, not rounding them

pjsjongsung:
Added a small comment. Everything else looks great! Thank you @TomasGuija

dipy/nn/utils.py
TomasGuija: Yeah that's right, I just changed it to be a bit more clearn, you can tell me whether it sounds better now.

dipy/nn/torch/synthseg.py
```diff
@@ -635,25 +638,36 @@ def predict(
...
         for i, T1_img in enumerate(T1):
             t_img, params = transform_img(
                 T1_img,
                 affine[i],
                 target_voxsize=(1.0, 1.0, 1.0),
-                final_size=(192, 192, 192),
                 order=3,
             )
+            transformed_images.append(t_img)
+            params_list.append(params)
```
skoudoro: avoid append and try to initialize shape in advance

skoudoro (approved):
LGTM, Thanks @TomasGuija 

merging

PR #4087, opened 2026-07-23
RF: simplify Parzen PDF normalization

TomasGuija:
## Description

Following what we discussed at #4067, this PR simplifies and makes the Parzen PDF normalization consistent across dense 2D, dense 3D, and sparse sampling. All implementations now normalize using `valid_points` and the redundant `total_sum` accumulation has been removed.

## Motivation and Context

During registration, the moving image is resampled on the static-image grid. If the current transformation maps a static-grid position outside the moving image, there is no real moving intensity to interpolate, so the resampler returns an artificial zero.

If the images are not normalized and zero lies outside the moving-image intensity range, this artificial value can fall outside the support of the histogram. It may then contribute little or no B-spline weight to the joint PDF. However, the sample is still included in `valid_points`, and its static intensity is still added to the static marginal. For example, if 100 samples are counted but only 50 contribute joint weight, normalization by `valid_points` gives:

```text
joint.sum() = 50 / 100 = 0.5
```

The previous `total_sum` normalization used in 3D and sparse sampling forced the joint PDF to sum to one in this situation, but the static marginal was still normalized using `valid_points`. Therefore:

```python
smarginal != joint.sum(axis=1)
```

During the standard `AffineRegistration` workflow, however, the input images are normalized to `[0, 1]`. Zero is consequently inside the histogram range, including the artificial zeros introduced outside the moving-image domain. Each counted sample then contributes a total B-spline weight of one, so:

```text
total_sum ≈ valid_points
```

Under this assumption, removing `total_sum` preserves the expected numerical behaviour while making dense 2D, dense 3D, sparse sampling, and the PDF gradients use the same normalization.

The only problem remaining is that if the metric is used directly,  without `AffineRegistration` performing normalization, the failure I described remains possible. Actually, even in standard normalization, although it maybe isn't that important, artificial zeros sampled outside the moving image are still treated as real image intensities. I think this is mostly okey, since usually 0 intensity corresponds to background though. 

A complete fix would require detecting positions outside the moving-image domain and excluding those samples from the joint PDF, marginals, and gradient. That would require a bigger PR. The scope of this one is mainly to allow a sooner merge of #4067, but it could be done in the future.

## Checklist

## Type of Change

- [ ] Documentation update

### DIFF
--- a/dipy/align/parzenhist.pyx
+++ b/dipy/align/parzenhist.pyx
@@ -636,10 +636,9 @@ cdef _compute_pdfs_dense_2d(double[:, :] static, double[:, :] moving,
         cnp.npy_intp offset, valid_points
         cnp.npy_intp i, j, r, c
         double rn, cn
-        double val, spline_arg, total_sum
+        double val, spline_arg
 
     joint[...] = 0
-    total_sum = 0
     valid_points = 0
     with nogil:
         smarginal[:] = 0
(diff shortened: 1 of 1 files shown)


skoudoro (approved):
Hi @TomasGuija,

Thank you for this!

it make sense and I am not sure why the previous code diverged

merging !

PR #4088, opened 2026-07-24
RF: unify forward and backward CC steps

TomasGuija:
## Description

This PR consolidates the existing forward and backward SyN Cross-Correlation steps into two shared functions:

- `compute_cc_step_2d`
- `compute_cc_step_3d`

The direction is selected explicitly through the `forward_step` argument. `CCMetric` and the corresponding tests have been updated to use the new functions.

## Motivation and Context

This follows the code-simplification discussions at #4078.

While investigating the duplicated code between the affine CC implementation and the existing SyN backward steps, I noticed that the SyN forward and backward implementations were themselves largely duplicated. Their allocation, voxel loops, validity checks, energy calculation, and output operation were identical. Only the derivative factor selected for `temp` differed.

This PR removes that pre-existing duplication without changing the numerical behavior. It also provides a cleaner base on which the CC-for-affine branch can later be rebased and further simplified.

For backward compatibility, since the four existing functions are directly callable, I could keep them as thin wrappers around `compute_cc_step_2d/3d`, with `forward_step` fixed to the corresponding direction, and mark them as deprecated for removal in a future release. All internal call sites would still use the new shared functions directly. I initially omitted these wrappers because the functions appear to be low-level implementation details, but I can add them if we consider them part of the supported API.

## Checklist

- [x] I have read the [[CONTRIBUTING](https://github.com/dipy/dipy/blob/master/.github/CONTRIBUTING.md)](https://github.com/dipy/dipy/blob/master/.github/CONTRIBUTING.md) guidelines.
- [x] My code follows the [[DIPY coding style](https://docs.dipy.org/stable/devel/coding_style_guideline.html)](https://docs.dipy.org/stable/devel/coding_style_guideline.html).

## Type of Change

- [ ] Documentation update

### DIFF
--- a/dipy/align/crosscorr.pyx
+++ b/dipy/align/crosscorr.pyx
@@ -1,6 +1,8 @@
 """ Utility functions used by the Cross Correlation (CC) metric """
 
 import numpy as np
+from dipy.utils.deprecator import deprecate_with_version
+
 from dipy.align.fused_types cimport floating
 cimport cython
 cimport numpy as cnp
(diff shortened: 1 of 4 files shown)


skoudoro (changes requested):
Hi @TomasGuija,

It looks good. 

However, a deprecation cycle is needed for the old function name. 

look at an exemple here: https://github.com/dipy/dipy/blob/af8f2e2b01649e9eb2c408a7a4fe700913ed0784/dipy/utils/fast_numpy.pyx#L71-L80

skoudoro (approved):
Thank you for this @TomasGuija,

LGTM, merging !

PR #4091, opened 2026-07-27
BF: fix dead code paths and a shadowed argument found 

skoudoro:
While auditing the codebase for duplicated / near-duplicated functions, several
of the duplicate pairs turned out to have **already drifted**: one copy carried a
fix or a guard the other did not. This PR collects the small, zero-risk fixes that
came out of that comparison, one commit per issue.

None of these change numerical output.

### `BF: replace non-existent np.Null with None in gqi empty-cone paths`
`dipy/reconst/gqi.py`

`patch_maximum()` and `patch_sum()` returned `np.Null` on the empty-cone path.
`np.Null` does not exist in NumPy, so both raised `AttributeError` instead of
returning. The twin function `equatorial_maximum()` — whose body is otherwise
identical — already returns `None, None` for the equivalent empty-band case.

### `BF: revive the dead --log_level CLI option in workflows`
`dipy/workflows/flow_runner.py`

`get_level()` looked up `logging._levelNames`, which was removed in Python 3.4.
The lookup always raised, the bare `except` swallowed it, and the function always
returned `logging.INFO`. Every workflow therefore ignored `--log_level`.

Replaced with `logging.getLevelNamesMapping()` (Python 3.11+; DIPY requires 3.12),
keeping the INFO fallback for unrecognised values.

### `BF: honour the startpoint argument of qtdmri generalized_crossvalidation`
`dipy/reconst/qtdmri.py`

The body overwrote its own keyword argument with `startpoint = 1e-4` on the first
line, so `startpoint=5e-4` in the signature was dead. Made `1e-4` the default and
dropped the overwrite. No caller passes `startpoint`, so this is numerically a
no-op while making the argument usable. Also added the missing parameter docs.

### `BF: stop shadowing the mask parameter inside force_peaks`
`dipy/reconst/force.py`

The ODF normalisation branch rebound the local name `mask`, clobbering the
function's own `mask` argument. Harmless today because `mask` is already consumed
by that point, but it silently breaks any later use of the parameter.

### `RF: build FORCE peaks through _pam_from_attrs`
`dipy/reconst/force.py`

`force_peaks` hand-rolled a `PeaksAndMetrics` and set 5 of its 10 attributes,
duplicating the canonical constructor in `dipy.direction.peaks`. Reuse
`_pam_from_attrs` so the attribute list lives in one place.

### Testing
`test_gqi.py`, `test_qtdmri.py`, `test_force.py`, `test_iap.py`, `test_io.py` pass.

### DIFF
--- a/dipy/reconst/force.py
+++ b/dipy/reconst/force.py
@@ -1557,10 +1557,10 @@ def force_peaks(fitted_object, *, mask=None, sh_order=8):
         ):
             v_max = np.max(odf, axis=-1, keepdims=True)
             v_min = np.min(odf, axis=-1, keepdims=True)
-            mask = v_max > 1.0
+            norm_mask = v_max > 1.0
             denom = (v_max - v_min) + 1e-12
             normalized_odf = (odf - v_min) / denom
-            odf = np.where(mask, normalized_odf, odf)
+            odf = np.where(norm_mask, normalized_odf, odf)
             res_sh = sf_to_sh(odf, default_sphere, sh_order=sh_order)
     peaks = PeaksAndMetrics()
     peaks.peak_dirs = res_dirs
--- a/dipy/reconst/force.py
+++ b/dipy/reconst/force.py
@@ -8,7 +8,7 @@ import warnings
 import numpy as np
 
 from dipy.core.gradients import unique_bvals_tolerance
-from dipy.direction.peaks import PeaksAndMetrics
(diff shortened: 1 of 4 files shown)


skoudoro:
Failed CI's in viz are expected and will not be fixed until FURY 2.0

maharshi-gor (approved):
Thank you for the work @skoudoro !

PR #4093, opened 2026-08-08
STYLE: Prefer using `pytest` temporary path fixtures in tests

jhlegarreta:
# Description

Prefer using the `pytest` `tmp_path` and `tmp_path_factory` fixtures in tests.

## Motivation and Context

- Makes the code more readable as no additional context (and the associated indentation) required by the use of `TemporaryDirectory` is needed.
- Simplifies path manipulation as they return a `pathlib.Path` object and thus, no recurrent casting to `Path` is necessary.
- Increases consistency in temporary directory naming, and thus, eases maintenance.

## How Has This Been Tested?

Will rely on CI builds.

## Checklist

## Type of Change

- [ ] Documentation update

### DIFF
--- a/dipy/align/tests/test_api.py
+++ b/dipy/align/tests/test_api.py
@@ -1,6 +1,3 @@
-from pathlib import Path
-from tempfile import TemporaryDirectory
-
 import nibabel as nib
 import numpy as np
 import numpy.testing as npt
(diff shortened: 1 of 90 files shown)


jhlegarreta:
~@skoudoro Ran out of ideas here to make the fixtures work.~

jhlegarreta:
Coverage build failure is unrelated (addressed in PR #4097).

skoudoro:
After some PRs merged, this one have mutiple conflict.

Furthermore, I need other opinions, not sure to be in favor of this PR. 

I prefer to depends on a standard library for tmp path than a 3rd party library. Even if `pytest` is solid, you never know the future. Our transition from numpy testing  + nosetests to pytest was easy and smooth because we did not heavily depends on those tools. 

Such a PR will increase our dependance to pytest when it is not really necessary.

jhlegarreta:
> After some PRs merged, this one have mutiple conflict.

Resolved.

> I prefer to depends on a standard library for tmp path than a 3rd party library. Even if pytest is solid, you never know the future. Our transition from numpy testing + nosetests to pytest was easy and smooth because we did not heavily depends on those tools.
> Such a PR will increase our dependance to pytest when it is not really necessary.

I was hoping this might no longer be an issue, but here are some additional arguments:
- Many major Python packages (e.g. Numpy, Pandas, etc.) use this feature.
- If the feature goes away, think that this small task is done a few hundred times across the code base, so wouldn't it make sense to have a fixture ourselves that does the job? A single variable or line in the tests, and no need to think about the output folder variable name (`tmppath`, `out_dir`, and a few other variants we have now), no need for an extra indentation across entire, long blocks. So, ultimately, less maintenance, and more time saving, which is highly relevant.
- The code is read many more times that it is written, so having a cleaner code with less boilerplate code and less visual hurdles (less indentation) is important.

IMHO, this feature will not go away that easily from pytest, and I do not know about any other alternative to pytest at this time so as to start asking ourselves about its future.

skoudoro:
@arokem any opinion ?

skoudoro:
Do you have any comments or opinion on that @arokem ?

arokem:
Sorry, missed your previous ping. I agree that this overall improves readability, and reduces boilerplate. +1 to merge from me.

skoudoro:
thanks for the feedback

jhlegarreta:
Had pinged Serge on the chat; should have stated it here, sorry; `test_default_Cnn1DDenoiser_flow` is failing due to some test fixture/parameterization issue that needs to be solved:
https://github.com/dipy/dipy/actions/runs/31411838388/job/93570013862

I ran out of options/exhausted my bandwidth to continue investigating fixes.

PR #4105, opened 2026-08-20
OPT: remove OpenMP lock contention in _bundle_minimum_distance

skoudoro:
## Description

_bundle_minimum_distance acquired a single global OpenMP lock inside the innermost loop, once per streamline pair, so every thread serialised on the same lock. Scaling was negative: on a 10-core machine (8 P + 2 E) a 400x400x20 evaluation took 7.6 ms on one thread and 4518 ms on the default ten, a 594x penalty for using all cores.

## Motivation and Context

Rather than lock, remove the sharing. min_j[i] accumulates in a thread-local scalar stored once per i, and each thread reduces into its own private row of min_i. No two threads write the same address, so the lock is gone entirely. min() is exactly associative and commutative on doubles, so the reduction is bit-identical to the serial result whatever the thread count or schedule; this was verified against 54 reference values from the previous implementation, spanning nine shapes and six thread counts, compared as raw bit patterns.

400x400x20 at the default thread count drops from 4518 ms to 5.1 ms. 1500x1500x20 goes from 78.4 ms serial to 14.3 ms on eight threads, so larger problems now scale positively instead of degrading. End to end, dipy_recobundles over the 30-bundle HCP atlas goes from hours to 57 seconds on a 1.3M streamline tractogram.

## How Has This Been Tested?

using existing test and test recobundles

## Checklist

## Type of Change

- [ ] Documentation update

### DIFF
--- a/dipy/align/bundlemin.pyx
+++ b/dipy/align/bundlemin.pyx
@@ -9,7 +9,7 @@ cimport numpy as cnp
 cimport safe_openmp as openmp
 from safe_openmp cimport have_openmp
 
-from cython.parallel import prange
+from cython.parallel import prange, threadid
 from libc.stdlib cimport malloc, free
 from libc.math cimport sqrt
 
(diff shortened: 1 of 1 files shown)


Garyfallidis:
Close to 1000X speedup ? DIPY for ever !!!

Garyfallidis:
Great catch! Merging ...!

PR #4123, opened 2026-08-23
STYLE: Delete unused variables from `core` module Cython code

jhlegarreta:
## Description

Delete unused variables from `core` module Cython code.

Fixes:
```
/home/runner/work/dipy/dipy/dipy/core/interpolation.pyx:190:16:
'values' defined but unused (try prefixing with underscore)
```

and other similar warnings raised in:
https://github.com/dipy/dipy/actions/runs/32512937731/job/96868025924?pr=4111

## Motivation and Context

Remove all style warnings from Cython code.

## How Has This Been Tested?

Relying on GHA CIs.

## Checklist

## Type of Change

- [ ] Documentation update

### DIFF
--- a/dipy/core/interpolation.pyx
+++ b/dipy/core/interpolation.pyx
@@ -187,7 +187,6 @@ def map_coordinates_trilinear_iso(cnp.ndarray[double, ndim=3] data,
     """
     cdef:
         double w[8]
-        double values[24]
         cnp.npy_intp index[24]
         cnp.npy_intp off, i, j
         double *ds=<double *> cnp.PyArray_DATA(data)


dipy/core/interpolation.pyx
```diff
@@ -187,7 +187,6 @@ def map_coordinates_trilinear_iso(cnp.ndarray[double, ndim=3] data,
     """
     cdef:
         double w[8]
-        double values[24]
```
skoudoro, line 190: The code is using a variable name `value` no defined but `values` is defined. 

Something to check deeper... I do not think we should remove it. we should just updates the variable name

skoudoro (changes requested):
Hi @jhlegarreta,

I do not think this variables should be removed.

Can you look into it.

Thanks

dipy/core/interpolation.pyx
skoudoro: Checked the history — withdrawing my objection, your removal is correct. merging

PR #4130, opened 2026-08-24
BF: Make FORCE simulation generation reproducible via a seed

### LINKED ISSUE #4092: DIPY FORCE simulation reproducibility
### Summary

`dipy.sims.force.generate_force_simulations` cannot be made reproducible through the public API. Two calls with identical arguments produce different simulation libraries, which changes downstream FORCE parameter estimates. There are two separate defects.

### 1. The documented reproducible path is unreachable

`dipy/sims/force.py:108`:

```python
def init_worker(base_seed=None):
    """Initialize a ProcessPoolExecutor worker with a unique RNG state.

    With ``base_seed=None``, seed = PID + high-resolution time so every
    worker has a different stream. With ``initargs=(base_seed,)`` for
    reproducibility, seed = base_seed + PID so workers differ but the run
    is reproducible for a fixed worker count.
    """
```

But `initargs` is never passed. Both pool constructions omit it — `dipy/sims/force.py:627` and `:680`:

```python
with ProcessPoolExecutor(
    max_workers=num_cpus, initializer=init_worker, mp_context=ctx
)
```

and the single-process paths at `:607` and `:657` call `init_worker()` bare. There is also no `seed` parameter on `generate_force_simulations`, so a caller cannot reach the `base_seed` branch at all. Every run therefore takes the `base_seed=None` path:

```python
seed = (os.getpid() * (2**16) + int(time.perf_counter_ns())) % (2**32)
```

### 2. Even if it were wired, `+ os.getpid()` defeats it

```python
seed = (int(base_seed) + os.getpid()) % (2**32)
```

PIDs are not stable across runs, so identical `base_seed` values still yield different streams on every invocation. The docstring's claim that "the run is reproducible for a fixed worker count" does not hold as implemented.

### Reproducing

```python
import numpy as np
from dipy.sims.force import generate_force_simulations

b = np.array([0.] * 3 + [1000.] * 12)
r = np.random.default_rng(0)
v = r.normal(size=(15, 3)); v /= np.linalg.norm(v, axis=1, keepdims=True); v[:3] = 0
g = gradient_table(b, bvecs=v, b0_threshold=50.)

a = generate_force_simulations(g, num_simulations=2000, num_cpus=2, verbose=False)
c = generate_force_simulations(g, num_simulations=2000, num_cpus=2, verbose=False)
print(np.allclose(a["signals"], c["signals"]))   # False, with no way to make it True
```

### Why it matters — measured impact

We are running FORCE as one arm of a test–retest repeatability study, where the estimator must be a fixed deterministic map or the irreproducibility is absorbed into the within-subject variance term of ICC.

Two libraries built from identical inputs (500,000 simulations, single-shell b = 1000 with 32 directions, `use_posterior=False`, Rician noise at SNR 18, 7,200 synthetic voxels spanning free water 0.05–0.50):

| quantity | value |
|---|---|
| free-water difference between the two libraries, per voxel | **RMSE 0.082** |
| the same estimator's total error against ground truth | RMSE 0.105 |
| implied library-sampling contribution to error **variance** | **≈ 30 %** |
| difference in the ROI-median free water | 0.0025 |

So roughly a third of FORCE's voxelwise error variance is library sampling rather than data or model, and a voxelwise free-water map is only weakly reproducible even given identical input data and an identical gradient table.

`save_force_simulations` / `load_force_simulations` provide a workaround by persisting the artefact, which is what we adopted. But that only preserves a library once built; it cannot reproduce one, so results are not reproducible from source on another machine.

### Suggested fix

Add a `seed` parameter to `generate_force_simulations` and thread it through, but **seed per task rather than per worker** so reproducibility does not depend on `num_cpus`. Work is already dispatched in batches (`batch_size=1000`), so each batch has a natural index. Deriving each batch's RNG from `(seed, batch_index)` — e.g. `np.random.default_rng([seed, batch_index])` — gives:

- identical output for identical `seed`, on any machine
- independence from `num_cpus`, which the current `base_seed + PID` scheme cannot offer even in principle
- no reliance on global `np.random.seed` / `random.seed` state in workers

`init_worker`'s docstring would then need updating, since the "reproducible for a fixed worker count" caveat would no longer apply.

### Platform

WSL2 (on Windows 11 x86_64), Ubuntu 22.04.1 LTS

### DIPY version

 dipy 1.12.1

### Environment

_No response_

### Python version

Python 3.14

### Additional information

_No response_

Atharva-Shah-2298:
## Description

`generate_force_simulations` could not be made reproducible through the public API: the documented `base_seed` path of `init_worker` was never wired up, and its `base_seed + PID` scheme would not have been reproducible across runs even if it were.

## Motivation and Context

Fixes #4092.

- **`dipy/sims/force.py`**: `generate_force_simulations` gains a `seed` parameter (default `2298`). Seeding is per *batch*, not per worker: each batch reseeds itself from a stream derived via `np.random.SeedSequence([seed, batch_index])` at the top of `_generate_batch_worker`. Identical seeds therefore yield bit-identical libraries independently of `num_cpus` or worker scheduling, while every batch still draws distinct random values. `seed=None` keeps the previous fresh-entropy behavior. The dead `base_seed` branch of `init_worker` is removed and its docstring corrected.
- **`dipy/reconst/force.py`**: `FORCEModel.generate` gains `seed=2298` and threads it through. The simulation cache registry records the seed as part of the cache key (`_seed_matches`); legacy entries without a `seed` field were not generated reproducibly, so they only match `seed=None` requests.
- **`dipy/workflows/reconst.py`**: `--seed` exposed on the FORCE workflow.

## How Has This Been Tested?

Ran the issue's reproducer: two identical calls now return identical libraries, `num_cpus=1` vs `num_cpus=4` are bit-identical, different seeds diverge, and within a run all 2000 signal rows are unique (no batch is a clone of another, guarding against the same seed being propagated to every worker).

New tests cover: same-seed/default-seed reproducibility, different-seed divergence, `seed=None` entropy, CPU-count independence (1 vs 2 workers), distinct per-batch streams, and cache keying on the seed (hit on repeat, miss on a different seed, legacy-entry handling).

## Checklist

- [X] I have added tests that cover my changes (if applicable).
- [X] All new and existing tests pass locally.
- [X] I have updated the documentation accordingly (if applicable).

## Type of Change

- [X] New feature (non-breaking change which adds functionality)
- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update

### DIFF
--- a/dipy/reconst/force.py
+++ b/dipy/reconst/force.py
@@ -18,6 +18,7 @@ from dipy.sims._force_core import (
     DEFAULT_TWO_FIBER_MIN_ANGLE,
 )
 from dipy.sims.force import (
+    DEFAULT_FORCE_SEED,
     DEFAULT_NUM_ODI_VALUES,
     DEFAULT_ODI_RANGE,
     generate_force_simulations,
@@ -266,6 +267,7 @@ def _find_cached_simulation(
         Path to the cached ``.npz`` file, or None if no match found.
     """
     registry = _load_cache_registry(cache_dir)
+    legacy_seed_cache_reported = False
     for entry in registry:
         if entry["num_simulations"] != num_simulations:
             continue
@@ -277,6 +279,15 @@ def _find_cached_simulation(
             continue
         if not _min_angles_match(entry, min_crossing_angles):
             continue
+        if "seed" not in entry and seed is not None:
+            candidate = cache_dir / entry["filename"]
+            if candidate.exists() and not legacy_seed_cache_reported:
+                logger.info(
+                    f"Cached FORCE simulations at {candidate} predate seed "
+                    "tracking and cannot guarantee reproducibility. A new "
+                    "seeded library will be generated."
+                )
+                legacy_seed_cache_reported = True
         if not _seed_matches(entry, seed):
             continue
         candidate = cache_dir / entry["filename"]
@@ -810,7 +821,7 @@ class FORCEModel(ReconstModel):
         num_simulations=500000,
         output_path=None,
         num_cpus=1,
-        seed=2298,
+        seed=DEFAULT_FORCE_SEED,
         wm_threshold=0.5,
         tortuosity=False,
         odi_range=DEFAULT_ODI_RANGE,
@@ -851,8 +862,8 @@ class FORCEModel(ReconstModel):
             Seed for reproducible library generation (see
             :func:`~dipy.sims.force.generate_force_simulations`). Identical
             seeds yield identical libraries for any ``num_cpus``. The seed
-            is part of the cache key. When None, every run uses fresh
-            entropy.
+            is part of the cache key. When None, each newly generated library
+            uses fresh entropy; a matching cached library may still be reused.
         wm_threshold : float, optional
             Minimum WM fraction to include fiber labels.
         tortuosity : bool, optional
--- a/dipy/reconst/force.py
+++ b/dipy/reconst/force.py
@@ -761,7 +761,7 @@ class FORCEModel(ReconstModel):
     ):
         r"""
         FORCE (FORward matching for Complex microstructure Estimation) model
-        :footcite:p:`Shah2025`.
+        :footcite:p:`Shah2026`.
 
         FORCE is a forward modeling paradigm that reframes how diffusion data
         is analyzed. Instead of inverting the measured signal, FORCE simulates
--- a/dipy/reconst/force.py
+++ b/dipy/reconst/force.py
@@ -216,6 +216,19 @@ def _min_angles_match(entry, min_crossing_angles):
     )
 
 
+def _seed_matches(entry, seed):
+    """Check whether a registry *entry*'s generation seed matches the request.
+
+    Entries written before the seed became part of the cache key carry no
+    ``seed`` field; like entries generated with ``seed=None``, they only
+    match requests with ``seed=None``.
+    """
+    entry_seed = entry.get("seed")
+    if entry_seed is None or seed is None:
+        return entry_seed is None and seed is None
+    return int(entry_seed) == int(seed)
+
+
 def _find_cached_simulation(
     cache_dir,
     gtab,
@@ -224,6 +237,7 @@ def _find_cached_simulation(
     odi_range,
     num_odi_values,
     min_crossing_angles,
+    seed,
 ):
     """Search the registry for a simulation matching the given parameters.
 
@@ -243,6 +257,8 @@ def _find_cached_simulation(
         Resolved number of ODI grid points.
     min_crossing_angles : tuple
         ``(two_fiber_min_angle, three_fiber_min_angle)`` in degrees.
+    seed : int or None
+        Generation seed requested for the library.
 
     Returns
     -------
@@ -261,6 +277,8 @@ def _find_cached_simulation(
             continue
         if not _min_angles_match(entry, min_crossing_angles):
             continue
+        if not _seed_matches(entry, seed):
+            continue
         candidate = cache_dir / entry["filename"]
         if candidate.exists():
             return str(candidate)
@@ -276,6 +294,7 @@ def _register_cached_simulation(
     odi_range,
     num_odi_values,
     min_crossing_angles,
+    seed,
 ):
     """Add a new entry to the cache registry.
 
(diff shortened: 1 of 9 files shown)


dipy/reconst/tests/test_force.py
```diff
@@ -351,6 +352,21 @@ def test_odi_grid_matches_legacy_and_exact():
     assert not _odi_grid_matches(entry, (0.01, 0.3), 19)  # diff range, same grid
 
 
+def test_seed_matches_legacy_and_exact():
```
Atharva-Shah-2298, line 356: This will change as FORCE becomes more mature. Currently a stop gap for the legacy key entries

dipy/sims/force.py
```diff
@@ -681,10 +698,11 @@ def create_mm(_name, dt, shape):
     if num_cpus == 1:
         # Serial in-process path — no subprocess spawned, safe on all platforms
         init_worker()
-        for start_idx, bs in batch_specs:
+        for start_idx, bs, batch_seed in batch_specs:
```
skoudoro: adds  `state = np.random.get_state()`  and `np.random.set_state(state)` before and after the loop because  `np.random.seed(batch_seed)` resets the user's global NumPy RNG. This will have an impact on the user code for stuff not related to us.

dipy/sims/force.py
```diff
@@ -681,10 +698,11 @@ def create_mm(_name, dt, shape):
     if num_cpus == 1:
         # Serial in-process path — no subprocess spawned, safe on all platforms
         init_worker()
```
skoudoro: if seed is not `None`, `init_worker` is useless.

dipy/sims/force.py
```diff
@@ -146,27 +146,21 @@ def smallest_shell_bval(bvals, *, b0_threshold=50, shell_tolerance=50, n=1):
     return min_shells, shell_mask
 
 
-def init_worker(base_seed=None):
+def init_worker():
```
skoudoro: in theory, since FORCE has already been release, you need a deprecation cycle when removing a parameter. 
so you need to add `@deprecated_params("base_seed", None, since="1.13.0", until="2.0.0")`

dipy/reconst/force.py
```diff
@@ -788,6 +810,7 @@ def generate(
         num_simulations=500000,
         output_path=None,
         num_cpus=1,
+        seed=2298,
```
skoudoro: magic number hard to track. it appears 3 times. if someone change it in one place, what are the consequences ?

I recommend a global variable that you use in the 3 places.

dipy/reconst/force.py
```diff
@@ -216,6 +216,19 @@ def _min_angles_match(entry, min_crossing_angles):
...
 
 
+def _seed_matches(entry, seed):
+    """Check whether a registry *entry*'s generation seed matches the request.
+
+    Entries written before the seed became part of the cache key carry no
+    ``seed`` field; like entries generated with ``seed=None``, they only
+    match requests with ``seed=None``.
+    """
+    entry_seed = entry.get("seed")
```
skoudoro, line 227: current user from the last release do not have `seed`. 

So the simulation will re-run and they will not understand why and why it is so long. 

add a warning or logging.info to explain

skoudoro (changes requested):
Hi @Atharva-Shah-2298,

Thank for this update. Overall, it looks good.

See below some comments. 

Also, ""every run uses fresh entropy" sentence in the docsstring is not anymore true. I believe you should update. 

Thanks

skoudoro:
also a rebase or merge from master would be great ! Thanks !

dipy/sims/force.py
Atharva-Shah-2298: I think I checked within the omp scope but made it tighter.

skoudoro (commented):
Hi @Atharva-Shah-2298,

Please rebase and fix conflict in this PR

Atharva-Shah-2298:
A small issue with the year of the submission that I had missed. The latest preprint year prefix has to be changed as well and made the style consistent with the other bib entries. Didnt want to open a seperate PR for this so just a small text change as well.

skoudoro (approved):
ok, LGTM

Thank you for the fix @Atharva-Shah-2298.

Merging

PR #4110, opened 2026-08-21
STYLE: Honor object in variable name

jhlegarreta:
## Description

Honor object in variable name: use `sft` as a name for a `StatefulTractogram` object. Increases consistency with the rest of the code.

## Motivation and Context

Style.

## How Has This Been Tested?

Relying on GHA CIs.

## Checklist

- [X] I have read the [CONTRIBUTING](https://github.com/dipy/dipy/blob/master/.github/CONTRIBUTING.md) guidelines.
- [X] My code follows the [DIPY coding style](https://docs.dipy.org/stable/devel/coding_style_guideline.html).
- [ ] I have added tests that cover my changes (if applicable).
- [ ] All new and existing tests pass locally.
- [ ] I have updated the documentation accordingly (if applicable).

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Maintenance / CI / Infrastructure

PR #4143, opened 2026-08-30
BF: Hide the injected `rng` from the pytest-visible signature

dylanpulver:
## Description

`set_random_number_generator()` advertises the wrapped test's full signature to `pytest` (`decorators.py:113`), `rng` included — the parameter the decorator injects. `pytest` reads that as a fixture request, so a test declaring `rng` without a default errors at setup with `fixture 'rng' not found`. This hides `rng` and leaves the test's own fixtures to resolve.

## Motivation and Context

`DIPY Test` has failed on every push to `master` since #4093 merged on 2026-08-19 — 8 of those 9 runs for this reason, the 9th a runner shutdown. One error out of 1,566 collected tests, 19 of 20 jobs green.

#4093 gave `rng` a default in 214 signatures so `pytest` would ignore it, and missed two of the 235 call sites: `test_default_Cnn1DDenoiser_flow(pytestconfig, rng)` and `test_horizon_flow(tmp_path, rng)`. Only `coverage` installs TensorFlow, which is why one cell of twenty caught it. The second site is latent — `optional_package("fury", max_version="1.0.0")` rejects the FURY 2.0.0 CI installs, so it never runs.

`set_random_number_generator` is also the only decorator in that module with no test of its own.

## How Has This Been Tested?

Four tests added; three fail on `master` (two `fixture 'rng' not found`, one on the signature assertion), the fourth controls the `rng=None` form.

I also diffed the `pytest`-visible signature of all 232 decorated tests, `master` vs branch: `rng` disappears everywhere, nothing else moves.

Scoped local runs, macOS / Python 3.13: 746 passed / 0 failed on `master`, 750 / 0 here, per-package counts identical bar the four new tests. `align`, and parts of `tracking`, `io` and `workflows`, stall on data fetches here on unmodified `master`; neither TensorFlow nor FURY installs, so neither real call site runs locally.

Adding `rng=None` to those two tests would also green the CI — happy to swap.

## Checklist

- [ ] All new and existing tests pass locally. *(scoped — see above; `dipy/align` cannot run on this machine)*

## Type of Change

- [ ] Documentation update

### DIFF
--- a/dipy/testing/decorators.py
+++ b/dipy/testing/decorators.py
@@ -96,21 +96,19 @@ def set_random_number_generator(seed_v=1234):
     def _set_random_number_generator(func):
         @wraps(func)
         def _set_random_number_generator_wrapper(*args, **kwargs):
-            rng = np.random.default_rng(seed_v)
-            kwargs["rng"] = rng
-
-            sig = inspect.signature(func)
-            params = sig.parameters
-
-            # If the wrapped test expects pytestconfig, preserve it as the
-            # first positional argument when present.
-            if "pytestconfig" in params and args:
-                return func(*args, **kwargs)
-
+            kwargs["rng"] = np.random.default_rng(seed_v)
             return func(*args, **kwargs)
 
-        _set_random_number_generator_wrapper.__wrapped__ = func
-        _set_random_number_generator_wrapper.__signature__ = inspect.signature(func)
+        # `pytest` builds a test's fixture request from the signature it can
+        # see, so the wrapper must advertise the parameters the test owns
+        # (`tmp_path`, `pytestconfig`, ...) while hiding `rng`, which this
+        # decorator supplies. Leaving `rng` visible makes `pytest` look for a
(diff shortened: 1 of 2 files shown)


skoudoro:
Thank you for doing this @dylanpulver!

I will check that on Monday or Tuesday (back from vacation, depending on priority/deadline). 

I am surprised that there is not a simpler way to fix that. 

but if it is the way to go, we will go ahead and merge it. 

we will keep you updated

skoudoro:
also @jhlegarreta, it seems to fails since your PR. Can you check that also ?

jhlegarreta:
I am aware of the fact that this has been failing since my PR (see https://github.com/dipy/dipy/pull/4093#issuecomment-5347263532 and https://github.com/dipy/dipy/pull/4110#issuecomment-5373637329), but my attempts to fix this did simply not work. I am fine with this solution. Thanks for doing this @dylanpulver.

skoudoro (approved):
I took time to look deeper and all good. 

Thank you @dylanpulver! merging

Issue #3933, opened 2026-03-31
Multiple bugs in `dipy/data/fetcher.py`: race condition, silent failures, broken unzip, and dead test code


ISHAN9876:
### Summary

Encountered these bugs when I was reading and understanding the underlying codebase for issue #3928 
Before writing tests its utmost important to ensure codebase stability. 

Static analysis and test review of `dipy/data/fetcher.py` and `dipy/data/tests/tests_fetcher.py` revealed 12 bugs - ranging from a silent data-loss race condition to a test that has never run in the project's history. Each is reproducible without a network connection. Minimal examples are included for each.

<img width="558" height="620" alt="Image" src="https://github.com/user-attachments/assets/19809742-7f3e-4ec3-a70d-8e5671743f2c" />

### Platform

Ubuntu 24.04

### DIPY version

1.12.0

### Python version

3.12.3

### Additional information

Additional test bugs: mkstemp fd never closed (Windows PermissionError), hardcoded ports 8000/8001 (CI collision), os.chdir not in try/finally (working directory permanently changed on failure), bare except swallows all failures silently.

kunalKumar-13:
I went through this list against current `master` while working on a separate bug in the same file, so here is where each one actually stands six months on. Line numbers are from `master` today.

**Already fixed** — three of the twelve have been closed out since March:

| # | | evidence |
|---|---|---|
| 1 | `fetch_data` `makedirs` TOCTOU | `fetcher.py:429` is now `os.makedirs(folder, exist_ok=True)` |
| 2 | `except (FetcherError, Exception)` | no longer present anywhere in the file |
| 5 | `check_md5` docstring names `md5` | docstring and signature both say `stored_md5` |

**Still live** — eight of them:

| # | | evidence |
|---|---|---|
| 3 | plain `.gz` always raises | `fetcher.py:555` — a `.gz` whose stem is not `.tar` falls to `raise ValueError("File extension is not recognized")`; there is still no `gzip.open` path |
| 6 | `fetch_tissue_data(include_optional=...)` ignored | accepted at `:2620`, never read anywhere in the body |
| 7 | `fetch_tissue_data` `makedirs` race | `:2640` is a bare `os.makedirs(folder)` inside `if not folder.exists()` — the same shape that was fixed at `:429` for `fetch_data`, just not here |
| 8 | `target_tractrogram_hcp` typo | `:2259` — double `r`, while the fetcher is `fetch_target_tractogram_hcp` |
[...]

PR #4146, opened 2026-09-01
Restrict tar extraction to the destination directory

asaraog:
One of the extraction problems from #3933.

`_make_fetcher` extracts downloaded tar archives with a bare `tarfile.extractall()`. Members are not checked against the destination, so an archive entry like `../escaped.txt` is written **outside** the target folder:

```python
with tarfile.open(tar_path) as tf:
    tf.extractall(path=out)              # current behaviour
# -> escaped.txt now exists in out's parent

with tarfile.open(tar_path) as tf:
    tf.extractall(path=out, filter="data")   # this PR
# -> tarfile.OutsideDestinationError: '../escaped.txt' would be extracted to ...
# -> only good.txt lands in out/
```

`filter="data"` is the extraction filter Python added for exactly this (CVE-2007-4559); it becomes the default in 3.14, and dipy already requires >=3.12, so no version guard is needed. The fetchers pull archives from remote URLs, so this is the path that matters.

Also switched the tar and zip handles to context managers — both were closed by hand, so an error mid-extraction leaked the handle.

Scoped deliberately: #3933 lists a dozen findings across `fetcher.py` and its tests, and the test-side ones (mkstemp handles, hardcoded ports, `os.chdir` without `try/finally`) already look fixed on main — `_free_port_server` binds port 0 and the `check_md5` tests close their descriptors. Happy to take more of the list if you point me at what is still live.

`ruff check` and `ruff format --check` clean on the changed file under the pinned 0.15.12.

### DIFF
--- a/dipy/data/fetcher.py
+++ b/dipy/data/fetcher.py
@@ -505,8 +505,8 @@ def _make_fetcher(
         A message to print to screen when fetching takes place. Default (None)
         is to print nothing
     unzip : bool, optional
-        Whether to unzip the file(s) after downloading them. Supports zip, gz,
-        and tar.gz files.
+        Whether to unzip the file(s) after downloading them. Supports zip,
+        tar.gz, and tar.bz2 files.
     use_headers : bool, optional
         Whether to use headers when downloading files.
 
@@ -537,16 +537,14 @@ def _make_fetcher(
                 p = Path(f)
                 if p.suffix in (".gz", ".bz2"):
                     if p.with_suffix("").suffix == ".tar":
-                        ar = tarfile.open(Path(folder) / f)
-                        ar.extractall(path=folder)
-                        ar.close()
+                        with tarfile.open(Path(folder) / f) as ar:
+                            ar.extractall(path=folder, filter="data")
                     else:
                         raise ValueError("File extension is not recognized")
                 elif p.suffix == ".zip":
-                    z = zipfile.ZipFile(Path(folder) / f, "r")
-                    files[str(f)] += (tuple(z.namelist()),)
-                    z.extractall(folder)
-                    z.close()
+                    with zipfile.ZipFile(Path(folder) / f, "r") as z:
+                        files[str(f)] += (tuple(z.namelist()),)
+                        z.extractall(folder)
                 else:
                     raise ValueError("File extension is not recognized")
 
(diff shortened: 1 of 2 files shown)


dipy/data/fetcher.py
```diff
@@ -537,16 +537,16 @@ def fetcher(*, include_optional=False):
                 p = Path(f)
                 if p.suffix in (".gz", ".bz2"):
                     if p.with_suffix("").suffix == ".tar":
-                        ar = tarfile.open(Path(folder) / f)
-                        ar.extractall(path=folder)
-                        ar.close()
+                        # filter="data" refuses members that would escape the
+                        # destination directory; it becomes the default in 3.14
```
skoudoro: comment to remove

skoudoro (changes requested):
Hi @asaraog, 

Thank you for doing this. 

Please, can you add a regression test for the CVE-2007-4559 fix?

Also, Can you update the docstring to say that we support  "zip, tar.gz, and tar.bz2" since a bare `.gz` would fail.

Thanks

dipy/data/fetcher.py
asaraog: Removed, thanks.

asaraog:
Thanks for the quick review @skoudoro — all three done.

- **Regression test**: `test_make_fetcher_tar_member_cannot_escape_folder` in `dipy/data/tests/test_fetcher.py`. It builds a `.tar.gz` containing a `../escaped.txt` member, serves it through the existing local test server, and runs a real `_make_fetcher(..., unzip=True)`. Verified both ways — on `master` it fails (the member lands outside the destination), with the fix it raises `OutsideDestinationError` and nothing escapes. `pytest dipy/data/tests/test_fetcher.py`: 35 passed.
- **Docstring**: now reads "Supports zip, tar.gz, and tar.bz2 files" — you are right that a bare `.gz` hits the `ValueError`.
- **Inline comment**: removed.

skoudoro:
Hi @asaraog,

one CI is failing because of that (see below). Can you update commit message to follow our standard ? 

image

Thanks!

asaraog:
Squashed into one commit with the BF: prefix. Passes tools/hooks/commit-msg.py locally. Thanks for the pointer.

skoudoro (approved):
Thank you for this work @asaraog, merging

PR #4166, opened 2026-09-10
BF: clean up the failed download when raise_on_error is True

asaraog:
fetch_data cleans up the failed download only when raise_on_error is False; the default True path re-raises and leaves the bad file on disk.


skoudoro (approved):
LGTM, thanks @asaraog, merging

PR #4167, opened 2026-09-10
BF: Bias field unmask output

skoudoro:
## Description

Fixes the regression-based bias field correction (`bias_field_correction`, `polynomial_bias_field_dwi`, `dipy_bias_field_correction`) so it produces a plausible field on real DWI data.

- The corrected volume is no longer passed through `applymask`. The brain mask only restricts the regression; the field is extrapolated outside the mask so background voxels are corrected, not zeroed.
- The regression now runs inside the N4 histogram-sharpening loop (Tustison et al. 2010). Without it, a direct fit of the log b0 explains grey/white matter contrast as a bias field.
- The B-spline fit gains a bending-energy penalty on its control lattice (`smoothness`).
- Design and Gram matrices are built once per pyramid level and reused across sharpening iterations; the B-spline field is evaluated as a separable tensor product. B-spline runtime on HARDI193 drops from 20 s to 1.8 s.
- `method="auto"` is deprecated (removal after 1.15.0). The workflow default moves from `auto` to `bspline`.

## Motivation and Context

On the Sherbrooke `HARDI193` dataset the output was skull-stripped and the field spanned 0.7 to 2.2 inside the brain (up to 3e8 outside), tracking anatomy rather than coil sensitivity. After this PR the field spans 0.8 to 1.3, correlates at 0.96 with SimpleITK N4 (previously 0.62), and the corrected GM/WM ratio matches N4 within 0.01. On Stanford HARDI the correlation goes from 0.94 to 0.98.

`auto` chose between methods by coefficient of variation, which decreases whenever tissue contrast is absorbed into the field, so it favoured the wrong result.

## How Has This Been Tested?

- `pytest dipy/denoise/tests/test_bias_correction.py` (46 tests, 12 new: background preservation, `zero_background`, integer clipping, histogram sharpening on a two-tissue phantom, bending penalty properties, separable evaluator vs design matrix, plan reuse, deprecation warning).
- `pytest dipy/workflows/tests/test_nn.py -k biasfield`.
- `doc/examples/bias_correction_dwi.py` run end to end with SimpleITK installed.
- Manual comparison against SimpleITK N4 on HARDI193 and Stanford HARDI (field range, correlation, GM/WM ratio, runtime).

## Checklist

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Documentation update


skoudoro:
image

dipy/denoise/tests/test_bias_correction.py
```diff
@@ -699,3 +712,302 @@ def test_dtype_preservation():
...
+    affine = (2.0 * ii - 0.5 * jj + 3.0 * kk + 1.0).ravel()
+    np.testing.assert_allclose(affine @ penalty @ affine, 0.0, atol=1e-10)
+    curved = (ii**2).astype(float).ravel()
+    assert curved @ penalty @ curved > 0
+
+    # Axes with fewer than 3 control points contribute nothing
+    assert np.all(_bending_penalty(n_control=(2, 2, 2)) == 0)
+
+
+def test_sharpening_separates_tissue_from_bias():
```
Atharva-Shah-2298, line 889: test_sharpening_separates_tissue_from_bias only covers method="poly" but now bspline is the default. Should we add a test for bspline as well so that we can understand regression for that as well?

Atharva-Shah-2298 (approved):
Works as intended. LGTM!

dipy/denoise/bias_correction.py
```diff
@@ -999,9 +1274,10 @@ def bias_field_correction(
 
         - ``"poly"``: Legendre polynomial regression — fast, low-parameter.
         - ``"bspline"``: Cubic B-spline regression — more flexible.
-        - ``"auto"``: Run both methods and return the one with lower
-          Coefficient of Variation within the brain mask. The chosen method
-          is logged at INFO level.
+        - ``"auto"``: Deprecated since 1.13.0, removed after 1.15.0. Runs
```
maharshi-gor, line 1277: Make sure to put 2.0.0

maharshi-gor:
Hey @skoudoro I agree with @Atharva-Shah-2298 and there is one place where deprecation of auto will be done 1.15.0.

Please check those otherwise LGTM!

maharshi-gor (approved):

PR #4168, opened 2026-09-11
BF: ODF slice alignment fix

maharshi-gor:
## Description

Fix ODF (SH glyph) rendering alignment for subjects whose affine matrix contains rotations. Two independent bugs caused misalignment:

1. **Shader direction transform inconsistency**: Several functions in the WGSL ray-marching shader evaluated SH coefficients using world-space directions instead of transforming them to local/voxel space via `world_transform_inv`. The bracketing/bisection functions correctly transformed directions in their main code paths, but fallback paths, Newton refinement (`evaluate_surface_analytic`), finite-difference normals (`evaluate_implicit`), the final intersection result, and directional RGB coloring all operated in world space. This caused incorrect ODF shapes, normals, and colors when the affine contained rotations.

2. **Slice plane coordinate space mismatch**: The image slicer and peaks slicer use **world-axis-aligned** plane tests (flat slices in world space regardless of affine rotation), but the ODF slicer used **voxel-index matching** (integer comparison of per-glyph voxel indices against a target index). With a rotated affine, voxel-aligned slices appear tilted in world space, causing the ODF slice to visually separate from the co-registered image and peaks slices.

### Changes

**WGSL shader** (`sh_billboard.wgsl`):
- Added `world_transform_inv` direction transforms in `surface_difference_fast` (fallback), `surface_difference` (fallback), `evaluate_surface_analytic` (both paths), `evaluate_implicit` (both paths), and `find_surface_intersection` (non-Hermite final result).
- Fragment shader now computes `local_color_dir` for directional RGB coloring and debug visualization modes.
- Replaced integer voxel-index slice visibility test with world-space distance-to-plane test using tolerance `abs(world_transform[i][i]) * 0.5`, matching the approach used by Fury's peaks slicer.

**Python** (`sh_slicer.py`):
- `SHGlyph3D.set_slices()` now passes world-space positions directly to shader uniforms when an affine is present, instead of converting to voxel indices. Without an affine, voxel-space clipping is preserved.

**Python** (`sh_billboard.py`):
- Removed `s_slice_indices` buffer binding and creation — no longer needed since slice visibility is determined by world-space position, not voxel indices.

## Motivation and Context
When loading diffusion MRI data from subjects with non-trivial affine matrices (containing rotations and axis flips, not just scaling and translation), the ODF glyphs rendered by Skyline did not align with the co-registered image slices and peak overlays. The ODFs appeared on tilted slice planes and their shapes/colors were distorted. This is because the affine's rotation component was not consistently handled in two parts of the rendering pipeline: the GPU shader's SH evaluation and the CPU-side slice plane selection logic.

The user's affine that exposed the issue:
```
[[ -2.5    0.08   0.07  113.64]
 [  0.07   2.45  -0.49 -104.41]
 [  0.08   0.49   2.45  -31.5 ]
 [  0.     0.     0.     1.   ]]
```

## How Has This Been Tested?

- Updated existing unit tests that verified voxel-coordinate mapping to instead verify world-coordinate pass-through behavior.
- Added new tests:
  - `test_sh_glyph_set_slices_passes_world_positions_with_rotated_affine` — uses the real rotated affine from the reported case, confirms world positions reach the shader uniforms unchanged.
  - `test_sh_glyph_set_slices_does_not_clip_with_affine` — verifies out-of-range world positions are not clipped when an affine is present (the shader's tolerance-based plane test handles bounds).
  - `test_sh_glyph_set_slices_clips_to_volume_without_affine` — verifies backward compatibility: voxel-space clipping is preserved when no affine is set.
  - `test_sph_glyph_billboard_sliced_has_no_slice_indices_buffer` — confirms the removed buffer is no longer created.
- Visual verification needed with a subject whose affine contains rotations, checking that ODF slices are flat and aligned with image/peaks slices.

## Checklist

## Type of Change

- [ ] Documentation update

### DIFF
--- a/dipy/viz/skyline/render/sh_billboard.py
+++ b/dipy/viz/skyline/render/sh_billboard.py
@@ -589,16 +589,6 @@ class BillboardSphGlyphShader(MeshShader):
             )
         }
 
-        # Per-glyph slice indices (for sliced rendering)
-        slice_buf = getattr(wobject, "slice_indices_buffer", None)
-        if slice_buf is not None:
-            coeff_bindings[1] = Binding(
-                "s_slice_indices",
-                "buffer/read_only_storage",
-                slice_buf,
-                "VERTEX",
-            )
-
         self.define_bindings(2, coeff_bindings)
         bindings[2] = coeff_bindings
 
@@ -1420,7 +1410,6 @@ def sph_glyph_billboard_sliced(
     obj._sh_mapping_mode = mapping_mode
     obj._sh_requested_lut_res = lut_res
 
-    obj.slice_indices_buffer = Buffer(voxel_coords.ravel().astype(np.int32))
     obj.material.n_coeffs = material_n_coeffs
 
     enable_octahedral_lut(
--- a/dipy/viz/skyline/render/sh_slicer.py
+++ b/dipy/viz/skyline/render/sh_slicer.py
@@ -14,7 +14,6 @@ from dipy.viz.skyline.render.renderer import (
     slice_slider_bounds,
     slice_slider_values_from_state,
     slice_state_from_slider_values,
-    voxel_values_from_slice_state,
 )
 from dipy.viz.skyline.render.sh_billboard import sph_glyph_billboard_sliced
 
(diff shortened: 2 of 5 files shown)


Atharva-Shah-2298:
image
 Works as expected!

Atharva-Shah-2298 (approved):
The code looks good to me. Dont understand the WGSL formatting.

skoudoro (approved):
Looks ok, Some shader files need some cleaning but it can be done on a new PR

thank you @maharshi-gor,

merging

PR #4179, opened 2026-09-17
BF: Robust RGB based data adaption in SKYLINE

### LINKED ISSUE #4175: SKYLINE frozen on wrong button click
### Summary

If I open a 3d image and click on RGB button ----> because it is not 4d, it gets stuck and Skyline becomes unusable because the error then takes over the thread

### Platform

Windows 11 X86-64

### DIPY version

dipy master

### Environment

_No response_

### Python version

Python 3.13

### Additional information

_No response_

maharshi-gor:
Closes #4175

## Description

Adds tri-state RGB support to Skyline's `Image3D` volume slicer. Three new IO utilities (`has_rgb_dtype`, `is_rgb_compatible_data`, `unpack_rgb_array`) handle structured NIfTI DT_RGB24 detection, value-based RGB compatibility checking, and structured-to-plain array conversion. The `rgb` parameter changes from `bool` to `bool | None` across the full stack: `None` auto-detects from structured dtype, `True` forces RGB mode, `False` disables it entirely. The UI RGB toggle now appears only when data has RGB-compatible values. `IntrospectiveArgumentParser` learns `BooleanOptionalAction` for `bool or None` parameters, enabling `--rgb` / `--no-rgb` CLI flags.

## Motivation and Context

In master, `Image3D` accepts `rgb=False` as a plain boolean — the user must know upfront whether their file is RGB and explicitly pass `rgb=True`. Structured NIfTI DT_RGB24 files (produced by `decfa`) arrive as structured dtypes that require unpacking before rendering. Without auto-detection, these files display as broken scalar volumes unless the user explicitly passes `--rgb`. On the CLI, `--rgb` was a valueless flag with no way to explicitly suppress RGB mode.

This PR makes RGB handling automatic for the common case (structured DT_RGB24 files) while keeping explicit control available. The UI toggle is restricted to volumes whose values are actually RGB-compatible, preventing it from appearing on DWI scalar data that happens to have 3 directions.

## How Has This Been Tested?

- `pytest dipy/io/tests/test_utils.py -x -q` — `test_has_rgb_dtype` (structured dtype detection), `test_is_rgb_compatible_data` (shape, uint8, integer range, float range, NaN, structured exclusion, empty array), `test_unpack_rgb_array` (RGB/RGBA unpacking, non-RGB passthrough)
- `pytest dipy/viz/skyline/render/tests/test_image.py -x -q` — structured auto-detection, RGBA auto-detection, channel value preservation, explicit `rgb=True` override, plain uint8 shows toggle, `rgb=False` suppresses structured detection, `rgb=True` on 3D scalar raises, DWI-range volume not rgb_capable, incompatible values raise on force
- `pytest dipy/workflows/tests/test_iap.py -x -q` — tri-state `BooleanOptionalAction` (absent→None, `--tristate`→True, `--no-tristate`→False, ordinary `--flag`→True store_true unchanged)
- Pre-commit (ruff check, ruff format, codespell) passes on all changed files

## Checklist

## Type of Change

- [ ] Documentation update


dipy/viz/skyline/render/image.py
```diff
@@ -71,8 +72,10 @@ def create_image_visualization(
         Callback function to be called after rendering.
     opacity : int, optional
         Opacity of the volume rendering.
-    rgb : bool, optional
-        Whether the image is RGB
+    rgb : bool or None, optional
+        ``None`` (default): auto-detect from structured NIfTI ``DT_RGB24``
```
skoudoro: default to remove

dipy/viz/skyline/render/image.py
```diff
@@ -178,9 +183,10 @@ def __init__(
             Callback used to request a render/update.
         opacity : int, optional
             Slice opacity in percent, expected in ``[0, 100]``.
-        rgb : bool, optional
-            Interpret a 4D volume as RGB/RGBA channels when True.
-            Colormap and directional-volume controls are ignored in this mode.
+        rgb : bool or None, optional
+            ``None`` (default): auto-detect from structured NIfTI ``DT_RGB24``
```
skoudoro: default to remove

dipy/viz/skyline/app.py
```diff
@@ -81,9 +81,10 @@ class Skyline:
         Value for ``cluster length thr``.
     buan_pvals : str, optional
         Value for ``buan pvals``.
-    rgb : bool, optional
-        Interpret a 4D volume as RGB/RGBA channels when True.
-        Colormap and directional-volume controls are ignored in this mode.
+    rgb : bool or None, optional
+        ``None`` (default): auto-detect from structured NIfTI ``DT_RGB24``
```
skoudoro: default to remove

dipy/io/utils.py
```diff
@@ -94,6 +94,106 @@ def make5d(data):
...
+
+    Returns
+    -------
+    bool
+        ``True`` when *data* carries a structured RGB/RGBA dtype.
+    """
+    return data.dtype.names is not None and tuple(data.dtype.names) in _RGB_FIELD_NAMES
+
+
+def is_rgb_data(data):
```
skoudoro: useless code..... any reason ? 

this is called only in the tests.  to remove and remove associated tests

dipy/viz/skyline/render/image.py
```diff
@@ -190,15 +196,15 @@ def __init__(
         """
         self.dwi = volume
         self.affine = affine
+        if rgb is None:
+            rgb = has_rgb_dtype(self.dwi)
+        self.dwi = unpack_rgb_array(self.dwi)
+        self._rgb_capable = self.dwi.ndim == 4 and self.dwi.shape[-1] in (3, 4)
```
skoudoro: does it mean it accept an rgb slice only ?

dipy/viz/skyline/app.py
```diff
@@ -1289,8 +1293,10 @@ def skyline(
         Clusters with average length less than ``cluster_length_thr`` in mm will be
         hidden. If None, it will show all cluster above the 25th percentile of the
         cluster length distribution.
-    rgb : bool, optional
-        Enable the colors in the image if 4D data with RGB/RGBA channels.
+    rgb : bool or None, optional
+        ``None`` (default): auto-detect from structured NIfTI ``DT_RGB24``
```
skoudoro: default to remove

dipy/workflows/viz.py
```diff
@@ -98,7 +98,8 @@ def run(
         stealth : bool, optional
             Do not use interactive mode just save figure.
         rgb : bool, optional
-            Enable the colors in the image if 4D data with RGB/RGBA channels.
+            Pass ``--rgb`` to force RGB mode.  Omitting the flag enables
```
skoudoro: How do you get RGB=FALSE here ?  it is either `true` or `None` and not `true/false`. I do not get that part. what is the benefit of `None`

skoudoro (changes requested):
Hi @maharshi-gor,

Thank you for this. See below some comment. 

Also, in the description, you talk about gitignore,but I do not get it since there is no gitignore here. Maybe description need an update or some files has been forgotten

dipy/workflows/viz.py
maharshi-gor: You are right on getting the False part in workflow.

To make you understand there are 3 stages of rgb.

1. Force True
2. None: Auto-detect.
3. Force False.

But, it might be tricky to put False in bool. Do you have any suggestions for that?

@skoudoro

dipy/viz/skyline/render/image.py
maharshi-gor: No, this means it can be rgb data so, we show switcher for the same.

dipy/io/utils.py
maharshi-gor: I was planning to use it for something else. Let me add it later when I introduce those changes.

dipy/viz/skyline/render/image.py
skoudoro: so explain me `shape=3` I do not understand

maharshi-gor: If the last channel is either 3 or 4 that would indicate that it can be rgb or rgba. 

But, I have a better alternative than this which I will push in a bit.

skoudoro:
Not sure how the flags works, can you explain?

image

skoudoro:
ok, I got it. it seems you just need to handle the scary value error for the user

skoudoro:
Can you fix the tests @maharshi-gor ?

thanks

maharshi-gor:
Hey @skoudoro, Thank you for the feedback, Should be ready to go!

skoudoro:
ok, let me check a last time

dipy/viz/skyline/render/image.py
```diff
@@ -482,7 +489,7 @@ def render_widgets(self):
         if changed:
             self._synchronize = new
 
-        if self.dwi.ndim == 4 and self.dwi.shape[-1] in (3, 4):
+        if self._rgb_capable:
```
skoudoro: look like a bug here.   if `rgb=False` but `rgb_capable`, it will toggle the button...

Please, can you check. you can do it via a test.

dipy/viz/skyline/render/image.py
```diff
@@ -190,16 +196,17 @@ def __init__(
...
-            raise ValueError(
-                "When specifying rgb=True, the last dimension of the volume "
-                "must be 3 (RGB) or 4 (RGBA)."
+        if rgb is None:
+            rgb = has_rgb_dtype(self.dwi)
+        self.dwi = unpack_rgb_array(self.dwi)
+        self._rgb_capable = is_rgb_compatible_data(self.dwi)
+
+        if rgb and not self._rgb_capable:
+            logger.error(
```
skoudoro, line 206: the `raise ValueError` was correct since we are at the API level. 

I was saying that it should be catch at the workflows level for better message and not scare the user.

skoudoro (changes requested):
Hi @maharshi-gor,

See below some comment

dipy/viz/skyline/render/image.py
maharshi-gor: We will never go back to the workflow level.  As the visualization will be over if we go back. Also now there is asynchronous file loading which makes it even more difficult. At max we can be at the app.py where we load the files. If that is fine.

maharshi-gor: We might need to stick to logger based messages in the case of SKYLINE.

skoudoro: ok, it might need a refactor in a new PR. 

I will put more thinking later.

skoudoro (approved):
Failing CI's are unrelated. 

Thank you for this work @maharshi-gor, merging

PR #4185, opened 2026-09-21
Document vecs2 in vector_cosine

VenishPaneliya:
`vector_cosine`'s Parameters section lists `vecs1` twice, so `vecs2` has no entry of its own and the rendered docs show a duplicated parameter:

```
    Parameters
    ----------
    vecs1 : (N, R) or (R,) array_like
       N vectors (as rows) or single vector.  Vectors have R elements.
    vecs1 : (N, R) or (R,) array_like
       N vectors (as rows) or single vector.  Vectors have R elements.
       It should be possible to broadcast `vecs1` against `vecs2`
```

The second entry is clearly meant to be `vecs2` — its own description refers to broadcasting "`vecs1` against `vecs2`", and the signature is `vector_cosine(vecs1, vecs2)`.

After the change numpydoc parses the section as `['vecs1', 'vecs2']`, matching the signature exactly.

Docs only, no behaviour change. `dipy/core/tests/test_geometry.py` passes (16 tests), and `ruff check` reports the same 2 pre-existing findings in this file with and without the change.


skoudoro (approved):
LGTM, thanks @VenishPaneliya

PR #4187, opened 2026-09-21
BF: save NIfTI-compliant color FA and fix dipy_info crash

maharshi-gor:
## Description

Migrates color-FA (DEC FA) NIfTI output in the DTI, DKI, and FW-DTI reconstruction workflows — plus the `reconst_dti.py` tutorial — from a plain uint8 4D array to the NIfTI-compliant structured `DT_RGB24` format via `decfa()`, and fixes `dipy_info` so it no longer crashes on such files. Also vectorizes the previously per-voxel-loop `decfa`/`decfa_to_float` helpers.

## Motivation and Context

`save_nifti(orgb, np.array(255 * RGB, "uint8"), affine)` wrote color-FA as a plain 4D array instead of the NIfTI-standard structured RGB dtype (intent code 1001) that `dipy.io.utils.decfa()` already exists to produce. Separately, `dipy_info` (`IoInfoFlow`) crashed with `TypeError: cannot perform reduce with flexible type` on structured-dtype RGB NIfTI files, since stats/percentile calls can't reduce over a structured dtype.

## How Has This Been Tested?

Extended `test_decfa` with a vectorization + round-trip case; added `test_save_nifti_decfa` (new `dipy/io/tests/test_image.py`, registered in `meson.build`) and `test_io_info_rgb` (exercising the former crash path directly); fixed two existing workflow tests whose RGB shape assertions broke under the new structured dtype.

```
pytest dipy/io/tests/test_utils.py dipy/io/tests/test_image.py dipy/io/tests/test_io.py \
       dipy/workflows/tests/test_io.py dipy/workflows/tests/test_reconst_dti.py \
       dipy/workflows/tests/test_reconst_dki.py -q
# 62 passed
```

Independently reviewed by a separate reviewer agent — APPROVE, re-confirmed after a post-review docstring wording fix (fresh re-read of the file + fresh test run). `ruff check`: 44 pre-existing violations remain in touched files; each individually verified to fall outside this change's diff hunks — no new violations introduced.

## Checklist

## Type of Change

- [ ] Documentation update


dipy/io/utils.py
```diff
@@ -229,9 +229,9 @@ def decfa(img_orig, *, scale=False):
     if scale:
         data_orig = (data_orig * 255).astype("uint8")
 
-    for ii in np.ndindex(img_orig.shape[:3]):
-        val = data_orig[ii]
-        out_data[ii] = (val[0], val[1], val[2])
+    out_data["R"] = data_orig[..., 0]
+    out_data["G"] = data_orig[..., 1]
+    out_data["B"] = data_orig[..., 2]
```
arokem, line 234: Does this have the same effect as the replaced code? 

I agree that it's clearer this way, but just want to make sure that I understand what the change entails.

arokem (commented):
I _think_ this is good, but honestly a bit confused. I seem to remember testing nifti files with the code that is being changed on tools that expected tensor dtype, and that worked. But it's been a while. Does it really not work as is?

dipy/io/utils.py
maharshi-gor: Yes I just vectorized it. I did not update any existing tests so that can be treated that it conserves the structure.

maharshi-gor:
> I _think_ this is good, but honestly a bit confused. I seem to remember testing nifti files with the code that is being changed on tools that expected tensor dtype, and that worked. But it's been a while. Does it really not work as is?

Hey @arokem can you elaborate on this? I am not able to follow.

Thanks!

arokem:
Is current master actually broken? In what way?

maharshi-gor:
So, few months back we dipy_info got re-implemented to cover a lot more information with a better printing structure by @jhlegarreta . If I use the externally created rgb file and it fails (So, files generated from DIPY would have worked on dipy_info as they never incorporated the codec. I realized it as I just used a file which was converted from DICOM to NIfTI). Same happened with SKYLINE. We just merged the skyline PR.

Also, on the other note it helps the viewer to quickly pick up that the file is rgb than doing the guess work of channels and dtype. So, in general better. 

I saw the function already present in the codebase written by you and not being used anywhere. So thought of updating it making DIPY align with the standard.

I hope I answered your concern @arokem

maharshi-gor:
> Is current master actually broken? In what way?

Yes, outside rgb file which follow NIfTI standard would fail. But usually we do not see the issue as we only use it in viewer in theory not in any step of pipeline.

arokem:
Still not sure I understand, but if someone else reviews this and thinks its fine, I would be fine merging this, because it looks correct.

dipy/io/image.py
```diff
@@ -136,6 +149,10 @@ def save_nifti(fname, data, affine, *, hdr=None, dtype=None):
 
     kwargs = {"dtype": dtype} if NIBABEL_4_0_0_PLUS else {}
     result_img = nib.Nifti1Image(data, affine, header=hdr, **kwargs)
+
+    if decfa:
+        result_img = _decfa(result_img, scale=True)
```
skoudoro: why `True` is harcoded  and we enforce it ?  

user can not change it. maybe he want this `False`.

dipy/io/image.py
```diff
@@ -105,6 +106,18 @@ def save_nifti(fname, data, affine, *, hdr=None, dtype=None):
...
+        If True, encode ``data`` as a NIfTI-compliant directional-encoded
+        color FA (DEC FA) image with structured ``DT_RGB24`` dtype and
+        intent code 1001. ``data`` must be a 4D array whose last dimension
+        has 3 elements (R, G, B) in the 0-1 range; values are scaled to
+        0-255 internally. ``hdr``, if given, is used as the starting
+        header, but its data dtype and dimensions are overwritten by the
+        DEC FA encoding. ``dtype`` has no effect on the saved file in this
+        case: the encoder reads ``data`` directly (unaffected by any
+        dtype set on the intermediate image) and always writes the
+        structured RGB dtype.
```
skoudoro: Can we simplify the description ?

Do we really need  4 lines explaining `dtype` is ignored?

dipy/io/image.py
```diff
@@ -2,6 +2,7 @@
 import numpy as np
 from packaging.version import Version
 
+from dipy.io.utils import decfa as _decfa
```
skoudoro: `_decfa` name confusing and error prone. Why did you need this understand? 

maybe the function arguments need a rename and not this

skoudoro (changes requested):
Hi @maharshi-gor,

Thank you for this. I understand the goal here.

Overall, looks good.

I am just afraid that  `load_nifti_data(rgb) * 1.0` breaks now due to the change of type.

I need to check some tutorial like `reconst_dti.py` and if it is the case, it will need fix and user warning.

maharshi-gor:
You concern is valid. But, I have not seen any place which is doing what you asked with rgb.

So for the case of load_nifti_data should we automatically unpack rgb to regular dtype and load_nifti  will preserve the structure @skoudoro ?

dipy/io/image.py
maharshi-gor: My idea was so land at decfa standard implementation and for more flexibility the user can use the decfa method itself.

maharshi-gor:
Hello @skoudoro ,

I have addressed all the concerns. You can take a look again.

PR #4188, opened 2026-09-21
MNT: Line ending fix

maharshi-gor:
## Description

`dipy/reconst/tests/test_shore_odf.py` had crept in with CRLF line endings, which was tripping the mixed-line-ending pre-commit hook on unrelated PRs that happened to touch that file. This PR normalizes the file to LF and closes the    
 gap that let it happen again:                                                                                                                                                                                                             
                                                                                                                                                                                                                                           
 - Added * text=auto eol=lf to .gitattributes so Git normalizes all text files to LF on checkout/commit.                                                                                                                                   
 - Passed args: ['--fix=lf'] to the mixed-line-ending pre-commit hook so it auto-fixes CRLF to LF instead of just flagging it.                                                                                                             
 - Re-saved dipy/reconst/tests/test_shore_odf.py with LF endings (content is unchanged — git diff -b shows no differences).

## Motivation and Context

CRLF endings in a tracked file caused spurious mixed-line-ending hook failures for contributors editing that file on unrelated changes. Enforcing `eol=lf` repo-wide and having the hook auto-fix rather than only detect prevents this     
from recurring for any file.    

## How Has This Been Tested?

 - git diff -b master..HEAD -- dipy/reconst/tests/test_shore_odf.py confirms the file content is byte-identical aside from line endings.                                                                                                   
 - pre-commit run mixed-line-ending --all-files runs clean with the new `--fix=lf` argument

## Checklist

<!-- Please check all that apply. -->

- [ ] I have added tests that cover my changes (if applicable). — whitespace/tooling-only change; verified manually per above, no new test warranted.
- [x] All new and existing tests pass locally. — no code paths changed; line-ending normalization is content-neutral.
- [ ] I have updated the documentation accordingly (if applicable). — no user-facing documentation affected.

## Type of Change

<!-- Check the relevant option(s). -->

- [ ] Documentation update


.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+# STYLE: Normalize all line endings to LF
+cbeab34e2c390569ffbb99ef5594f5846ba150a6
```
skoudoro: I do not understand the addition of this file and not sure it is needed. it seems it will bring more maintenance than value

skoudoro (changes requested):
Hi @maharshi-gor,

Thank you for this. 

Strange description, please update it in a more human and relevant way. 

> "852 tracked text files renormalized"

What is that ? or where is it ?

> "previously 608 w/crlf"

same question ?

maharshi-gor:
Hello @skoudoro Thank you for reviewing this.

Both the comments are actually linked. Currently you are seeing much less changes as the commit is put in blame ignore. I did not want to update the authorship of the original author on the places where they have the expertise.

Given that this fix touched almost every file. I have started this file, where we can put SHA of the commit which just does the style commit on the entire repo. Otherwise it will show almost all the files which windows users would have touched.

.git-blame-ignore-revs
maharshi-gor: You are correct I thought it will be a huge update on codebase but DIPY is pretty much on the LF. I have removed this.

maharshi-gor:
Hey @skoudoro Updated the description and removed the .git-blame-ignore-revs file.

you can take a look again.

Thank you!

skoudoro (approved):
Thank you for this addition and the update.

All good, merging

